PR #558, opened 2026-04-18
Replace csgrs crate with a local implementation of csg

Ricardo-C-Oliveira:
The core2 cargo was [deprecated](https://crates.io/crates/core2) on 4/15, this crate is a dependency for [csgrs](https://github.com/timschmidt/csgrs). Claude suggested that given how csgrs is being used, having a local implementation is cheap and simple.

## Summary by CodeRabbit

* **New Features**
  * Added constructive solid geometry operations: union, difference, intersection for polygon meshes.

* **Refactor**
  * Geometry pipeline moved from an external library to an internal BSP-based implementation, simplifying processing.

* **Chores**
  * Removed the external geometry processing dependency.

* **Improvements**
  * More robust handling of degenerate/invalid triangles and normals during conversions.


louistrue:
Thx, @Ricardo-C-Oliveira

PR #562, opened 2026-04-19
fix(geometry): subtract voids per sub-mesh so multi-layer walls keep layer colors (#541)

### LINKED ISSUE #541: multi layer compound walls - materials not correctly read
Attaching a simple project . the south wall is not correctly displayed with materials. it suddenly gets transparency etc:

Intended visual:

Image

How its rendered in ifclite:

Image

[Project1.zip](https://github.com/user-attachments/files/26546952/Project1.zip)

louistrue:
## Summary

Walls that use direct `IfcStyledItem` bindings for each extrusion item — the classic "air / insulation / brick" multi-layer wall — rendered as a single grey block in the viewer as soon as the element had any opening (window, door). Ordering or material-set tweaks from PR #560 had no effect here because **those layers never became sub-meshes in the first place**: `has_openings → process_element_with_voids` collapses every representation item into one merged, void-subtracted mesh.

This PR adds the missing path.

## What changed

- `rust/geometry/src/router/voids.rs`
  - Split `process_element_with_voids` into a thin wrapper plus a new `apply_voids_to_mesh` helper that takes an already-built mesh and applies the full opening/clipping-plane chain. Behavior of the existing function is preserved.
  - Add `process_element_with_submeshes_and_voids`: collect one sub-mesh per `IfcShapeRepresentation` item, then run `apply_voids_to_mesh` on each. Each sub-mesh keeps its original `geometry_id` so the color resolver can look up the per-item `IfcStyledItem`.
- `rust/wasm-bindings/src/api/gpu_meshes.rs`
  - The three `has_openings` branches that drive `parseMeshes` / `parseMeshesAsync` (simple phase + complex phase + streaming) now try the per-sub-mesh void path first. They fall back to the merged-mesh path only when it yields no sub-meshes or errors out — so nothing regresses for walls without layered bindings.
- `rust/geometry/tests/submesh_voids_test.rs`
  - Three new tests: (1) a three-layer wall with an opening keeps three sub-meshes with the original extrusion IDs; (2) CSG actually runs (triangle count changes vs. the uncut path); (3) the empty-return contract when the void index has no entry.

## Why this complements #560

#560 fixes per-layer color assignment when material layers flow through `IfcMaterialLayerSet`. This PR fixes the case where the element already has direct `IfcStyledItem` bindings per extrusion but the void path threw the sub-mesh boundaries away. Together they cover the two common "multi-layer wall" authoring patterns.

## Test plan

- [x] `cargo test -p ifc-lite-geometry --test submesh_voids_test` — 3/3 pass
- [x] `cargo test -p ifc-lite-geometry --test csg_void_test` — no regression on existing void behavior
- [x] `cargo build -p ifc-lite-wasm` — clean (2 pre-existing warnings, not introduced here)
- [x] Manual viewer check against Project1.ifc reporter file: the south wall's air / insulation / brick layers render with their individual surface styles after opening subtraction

Closes #541

https://claude.ai/code


louistrue:
Addressed @chatgpt-codex-connector's P2 feedback on per-submesh recomputation in d51df00.

Split `apply_voids_to_mesh` into:
- `build_void_context(element, opening_ids, decoder) -> VoidContext` — expensive per-element work: `classify_openings` (which re-meshes every `IfcOpeningElement` via `process_element`), `merge_rectangular_openings`, and clipping-plane extraction/transform.
- `apply_void_context(mesh, &ctx) -> Mesh` — cheap per-mesh work: bounds, extend rect openings along the extrusion axis (needs per-submesh wall_min/wall_max), batched rectangular clip, non-rect CSG, and clip-plane pass.

`process_element_with_submeshes_and_voids` now builds the context once and reuses it across every sub-mesh, eliminating the N× classification cost on multi-layer walls with windows. The single-mesh `apply_voids_to_mesh` remains as a thin wrapper so `process_element_with_voids` behavior is unchanged. All three `submesh_voids_test` cases and `csg_void_test` still pass. WASM bundle regenerated (1003K).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #597, opened 2026-04-29
feat(renderer): transparencyOverrides for X-Ray mode

### LINKED ISSUE #593: Transparency/Visibility
Hey, love all the work you're doing. Just wondering if there is transparency settings/features within the renderer? If not, should I set about developing one to merge in?

joepaddock-uk:
Closes #593. Adds `RenderOptions.transparencyOverrides?: Map<expressId, alpha>`.

Per-frame routing — no `batch.color` mutation, IFC-declared alphas stay baked. Affected meshes/batches route through the existing transparent pipeline via two helpers (`alphaForMesh`, `alphaForBatch`); per-batch result is memoised in a static `WeakMap<overrides, WeakMap<batch, number>>` on `Renderer` so steady-state is O(1) per batch (cache GCs naturally on overrides identity change).

Selected meshes are exempt at the pre-write site (`isSelected ? mesh.color[3] : alphaForMesh(...)`); for batches, the existing highlight pass renders selected meshes opaquely on top of any transparent geometry — selection stays fully readable.

Includes a changeset (`@ifc-lite/renderer`: minor).

## Summary by CodeRabbit

* **New Features**
  * Added per-entity transparency overrides for per-frame X‑Ray alpha control.
  * Non-selected meshes/batches render using the override alpha via the transparent pipeline.
  * Selected meshes remain opaque to preserve highlights.
  * Mixed batches resolve to the minimum override alpha and re-render highlights correctly.

### DIFF
--- a/.changeset/transparency-overrides.md
+++ b/.changeset/transparency-overrides.md
@@ -1,0 +1,28 @@
+---
+"@ifc-lite/renderer": minor
+---
+
+Add `transparencyOverrides?: Map<expressId, alpha>` to `RenderOptions` for
+per-frame alpha control (X-Ray mode).
+
+Non-selected meshes/batches whose `expressId` appears in the map render at the
+override alpha through the existing transparent pipeline. Selected meshes are
+exempt so highlight rendering stays opaque. Mixed batches (some entries
+overridden, some not) take the minimum override alpha — the selection
+highlight pass then re-renders selected meshes opaque on top, so the user sees
+selection in full while the rest fades.
+
+Use case: viewers that want a true see-through "X-Ray" effect (selection visible
+through ghosted geometry) instead of fully hiding non-selected elements via
+`isolatedIds`.
+
+Per-batch alpha resolution walks `batch.expressIds` per frame. For typical batch
+sizes the cost is well below noise vs. the GPU work, and callers supply a fresh
+Map when contents change (same convention as `hiddenIds`/`isolatedIds`). Routing
+is purely per-frame — no mutation of `batch.color`, so IFC-declared alpha baked
+into cached batches stays untouched.
+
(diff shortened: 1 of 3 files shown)


louistrue:
@joepaddock-uk Thanks for the patch, this looks clean. Regarding the CodeRabbit feedback above:

1. Mutable Map cache staling is valid, please fix. txBatchCache keys by Map identity, so if anything calls overrides.set(...) on the same instance, the inner WeakMap keeps serving stale alphas while alphaForMesh reads fresh. Don't bother with a version field: either just doc it ("supply a new Map when contents change", same convention as hiddenIds/isolatedIds), or drop the cache entirely and walk batch.expressIds per frame. For typical batch sizes the cost is well below noise vs. the GPU work, and the bug class disappears. Personally I'd just delete the cache.

2. Federated IDs: not for this PR. CodeRabbit is technically right but hiddenIds/isolatedIds/selectedIds are all Set<number> keyed by bare expressId with selectedModelIndex doing the disambiguation. Switching one option to globalId while leaving the rest legacy makes the API inconsistent. I'll open a separate issue to do all of them at once. Keep Map<number, number> here.

3. Partial sub-batch pipeline is valid too, please fix. This one's a real correctness bug: at line 1227 isTransparent is still computed from color[3], but renderBatch writes the resolved override alpha into the uniform. So X-Ray combined with hide/isolate routes through the opaque pipeline (depth-write on, no blending) with a transparent alpha won't render see-through. CodeRabbit's suggested diff (alphaForBatch(subBatch, subBatch.color[3])) seems like the right fix.

If you can push (1) and (3), happy to merge. 🙏

louistrue:
Thx @joepaddock-uk 🙌

PR #610, opened 2026-04-30
fix(geometry): swept-disk follow-ups + WASM rebuild for #606

louistrue:
## Summary

Three follow-ups to [#606](https://github.com/louistrue/ifc-lite/pull/606) (`feat(geometry): honor IfcSweptDiskSolid StartParam/EndParam`):

- **CodeRabbit review feedback.** Junction-point dedup in `get_composite_curve_points_trimmed` is now coordinate-aware. The previous unconditional `skip(1)` distorted directrices whose adjacent composite-curve segments don't share a coordinate-identical junction vertex (model drift, mismatched cartesian points). Drop the first point only when it's within `1e-6` of the last point already collected. Regression test with non-coincident junctions added.
- **PR description "unrelated bug spotted while testing".** `SweptDiskSolidProcessor` re-picked `up` per cross-section based on `tangent.x.abs() < 0.9`; consecutive tangents that straddled the threshold flipped the sign of `perp1`, swapping vertex `j` to the opposite angular position on the next ring — visible as a twisted / flat-ribbon tube at L-bends (and the rebar hooks the PR called out). Replaced the per-iteration `up` selection with a Rotation-Minimising Frame: pick `up` once for the first sample, then propagate via Rodrigues rotation from `prev_tangent` onto `curr_tangent`. Three new unit tests cover straight-line invariance, 90° L-bend non-flip, and degenerate-input handling.
- **WASM rebuild.** Release CI installs `wasm-pack` but never invokes `scripts/build-wasm.sh`, and `packages/wasm` has no build script — it just publishes the checked-in `pkg/*` binaries. So #606's trim-param fix was not actually reaching consumers; the binary on npm was still pre-#606. Ran `scripts/build-wasm.sh` and committed the regenerated `pkg/ifc-lite_bg.wasm` + `pkg/ifc-lite.{js,d.ts}` + `pkg/ifc-lite_bg.wasm.d.ts` so the next `@ifc-lite/wasm` patch release ships #606 + the two source fixes above. `Cargo.lock` syncs to the workspace's 2.2.0 version (overdue from the last release).

The two source-only changes ship as `.changeset/swept-disk-followups.md` (`@ifc-lite/wasm: patch`).

## Out of scope (capture for follow-up issues if desired)

From the #606 PR description's "Known limitations" — left for separate PRs because each is its own scope:

- `IfcLine`, `IfcCircle`, `IfcEllipse`, `IfcTrimmedCurve`, `IfcBSplineCurve` directrices still ignore `StartParam` / `EndParam`. Their parameterisations are length / angle / knot-based and need their own samplers.
- Fractional lerp uses uniform-vertex parameter, not arc-length. Accurate for piecewise-linear segments (the rebar case) but an approximation for curved parents.
- `SweptDiskSolid.InnerRadius` (hollow-tube) is still ignored. Pre-existing.

## Test plan

- [x] `cargo test -p ifc-lite-geometry --lib 'profiles::tests'` — 18 pass (1 new: `test_composite_curve_trim_keeps_non_coincident_junction`)
- [x] `cargo test -p ifc-lite-geometry --lib 'processors::swept'` — 3 pass (`rmf_is_constant_on_a_straight_line`, `rmf_does_not_flip_at_sharp_bends`, `rmf_handles_degenerate_inputs`)
- [x] `cargo test -p ifc-lite-geometry --lib` — 124 pass, 1 fail (`extrusion::tests::test_circular_profile_detection` — pre-existing on main, unrelated)
- [x] `pnpm --filter @ifc-lite/wasm test` — packaging test green against rebuilt artifacts
- [x] `pnpm test:esm` — 26 packages green, no entrypoint regressions
- [ ] Visual regression on a rebar-heavy IFC (not run locally — would need a Tekla/Revit fixture; the L-bend RMF test exercises the same code path as a unit test)


louistrue:
@joepaddock-uk care to test with your tekla rebar files? Hope i didnt mess anything up 😅 https://ifc-lite-git-fix-swept-disk-followups-ltplus.vercel.app/

joepaddock-uk:
@louistrue working with my test models, can reconfirm when next version is pushed in my own app too!

PR #656, opened 2026-05-12
fix: reject overlapping geometry streams

maxkrut:
## Summary
- reject overlapping browser/WASM `GeometryProcessor` streaming runs before they can re-enter IFCLite internals
- cover both mesh and instanced streaming entrypoints with the same module-level guard
- add a regression test that starts one stream, verifies the second stream fails with a controlled JS error, then confirms the guard is released after cancellation
- add a patch changeset for `@ifc-lite/geometry`

## Repro
In React/Vite dev mode, StrictMode or quick modal remounts can start a second `GeometryProcessor.processStreaming(...)` run while the previous WASM-backed stream is still initializing or parsing. With large IFC files this can surface as a Rust/WASM `RefCell already borrowed` panic instead of a recoverable JS error.

## Test plan
- `corepack pnpm --filter @ifc-lite/data build`
- `corepack pnpm --filter @ifc-lite/geometry test`
- `corepack pnpm --filter @ifc-lite/geometry build`
- `corepack pnpm --filter @ifc-lite/geometry exec tsc --noEmit`
- `git diff --check`

---
<a href="https://app.blacksmith.sh/louistrue/codesmith/ifc-lite/pr/656"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

## Summary by CodeRabbit

* **Bug Fixes**
  * Geometry processor now rejects concurrent WASM streaming operations, throwing an error when a second streaming run is initiated while a previous operation is still active.

[Review Change Stack](https://app.coderabbit.ai/change-stack/louistrue/ifc-lite/pull/656)

### DIFF
--- a/.changeset/cold-walls-wait.md
+++ b/.changeset/cold-walls-wait.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Reject overlapping WASM streaming geometry runs with a controlled JavaScript error before re-entering the processor.
--- a/packages/geometry/src/geometry-processor-streaming.test.ts
+++ b/packages/geometry/src/geometry-processor-streaming.test.ts
@@ -179,6 +179,37 @@ describe('GeometryProcessor byte streaming fallback', () => {
     expect((completeEvent?.coordinateInfo as { buildingRotation?: number })?.buildingRotation).toBe(0.5);
   });
 
+  it('rejects overlapping WASM streaming runs before re-entering the processor', async () => {
+    const firstGeometry = new GeometryProcessor();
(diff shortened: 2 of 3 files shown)


louistrue:
Th, @maxkrut

Issue #674, opened 2026-05-14
broken door handle geom


louistrue:
[door.ifc.txt](https://github.com/user-attachments/files/27745590/door.ifc.txt)

> 
> It is still not fixed: 
> 
> <img width="609" height="555" alt="Image" src="https://github.com/user-attachments/assets/b9933473-94c6-4c35-a8ec-35e8dd27a749" /> 

 _Originally posted by @denizmaral in [#604](https://github.com/louistrue/ifc-lite/issues/604#issuecomment-4442751913)_

louistrue:
Reopening — PR #793 fixed the surface-of-revolution sign-preservation bug, but the door handle is still visibly broken in the viewer after deploy.

The lever bar (long horizontal cylinder) renders detached from the rosette (round disc), with a small stem pointing down-and-away from the rosette instead of horizontally toward the bar. Looks like a different primitive than the SoR — possibly the IfcExtrudedAreaSolid or IfcAdvancedBrep that forms the connecting stem has its direction/orientation computed wrong.

@louistrue reposted the screenshot showing the residual defect after the deploy.

Investigating which sub-shell of the IfcDoor produces the stem and what geometry primitive backs it.

PR #801, opened 2026-05-25
fix(geometry): resolve glass-vs-metal-trim Z-fight on Revit-exported doors (#674)

louistrue:
## Root cause

Revit-exported door fixtures ship the glass pane and the metal-trim
sub-shells with their interface faces sitting on *exactly* the same
plane (in `tests/models/various/issue-604-door.ifc` the glass brep
#1227 spans `y=[150, 155]` while metal-trim #2326 ends at `y=150` and
#1778 starts at `y=155`, with the glass XZ footprint nested inside the
trim's). Coplanar same-pixel faces fight for the depth buffer, which
shows up in the rendered viewer as the striped/banded flicker across
the door's vision panel.

## Fix layer

Picked **(c) element-level coplanar dedup** — added a small
`dedupe_coplanar_submeshes` pass that runs inside
`process_element_with_submeshes` *before* the placement transform.
For each pair of sub-meshes belonging to the same door/window it:

1. Checks whether their local-frame AABBs share an axis-aligned plane
   (`a.max[axis] ≈ b.min[axis]` or `a.min[axis] ≈ b.max[axis]`).
2. Checks whether their 2D projection on the other two axes overlaps
   with positive area (so a touching edge alone is ignored).
3. Nudges the smaller-by-vertex-count sub-mesh inward by
   `eps = 5e-5 × max_sub_mesh_extent` (≈ 0.05–0.1 mm at typical IFC
   product scales) on that face plane.

Why this layer rather than the alternatives:

- **(a) geometry epsilon-shrink on the glass only.** Cleanest in
  theory but requires deciding "is this brep glass?" — there is no
  inverse-index from `IfcAdvancedBrep` to `IfcStyledItem` in
  `ifc-lite-core`, so detection would mean scanning every styled item
  on every brep. The dedup achieves the same physical effect (glass
  ends up at 5 − 2ε mm) without an inverse-lookup.
- **(b) / (d) material-driven render bias / polygon offset.** Larger
  surface area (renderer plumbing + WebGPU pipeline changes) and ships
  the wrong inversion of responsibility — the geometry layer would
  emit Z-fighting meshes and ask the renderer to paper over it.
- **(e) CSG cleanup.** Overkill and risks regressing other fixtures.

The dedup uses AABB face heuristics (no per-triangle plane analysis),
runs in O(N²) over the sub-meshes of a single element, and is a no-op
when N < 2 or no plane pairing is found.

## Before / after AABBs (issue-604 door, raw mm, local frame)

```
Pre-fix                                          Post-fix
Glass    #1227  y=[150.00, 155.00]   ⟶   Glass    #1227  y=[150.105, 154.895]
Metal-pt #1778  y=[155.00, 181.00]   ⟶   Metal-pt #1778  y=[155.000, 181.000]
Metal-pt #2326  y=[124.00, 150.00]   ⟶   Metal-pt #2326  y=[124.000, 150.000]
```

Glass thickness 5 mm → 4.79 mm (Δ = 2 × 0.105 mm = sub-pixel). Both
trim sub-shells stay put — they have ~800 verts vs glass's 24 so the
dedup picks glass to shrink.

## Verification

- `cargo test -p ifc-lite-geometry --tests` — **177 + 5 new = 182 passed,
  0 failed** (4 new zfight unit tests under `router::zfight::tests` +
  1 new regression `glass_submesh_is_inset_to_break_coplanar_zfight`).
- `cargo test -p ifc-lite-geometry --tests --features manifold-csg` —
  same.
- Per-element diff vs IfcOpenShell on **AC20-FZK-Haus** and
  **C20-Institute-Var-2** via the correctness harness:
  - FZK: 3 of 83 elements changed (1 door, 2 windows). Max delta
    0.100 mm — exactly the dedup ε.
  - Institute: 1 of 702 elements changed (1 door). Max delta 0.105 mm.
  - All deltas are below `diff.py`'s 0.1 mm voxel pitch, so IoU vs IOS
    stays unchanged. No new elements drop out, no extents collapse.
- `bash scripts/build-wasm.sh` — built clean (1.3 M, no new warnings).

## Test plan

- [x] `cargo test -p ifc-lite-geometry --tests`
- [x] `cargo test -p ifc-lite-geometry --tests --features manifold-csg`
- [x] Per-element correctness harness on FZK + Institute (no semantic regressions)
- [x] WASM build (`scripts/build-wasm.sh`)
- [ ] Visual confirmation in the deployed preview that the door's glass panel no longer flickers

### DIFF
--- a/rust/geometry/src/router/mod.rs
+++ b/rust/geometry/src/router/mod.rs
@@ -13,6 +13,7 @@ mod processing;
 mod transforms;
 mod voids;
 mod voids_2d;
+mod zfight;
 
 #[cfg(test)]
 mod tests;
--- a/rust/geometry/src/router/processing.rs
+++ b/rust/geometry/src/router/processing.rs
@@ -612,6 +612,10 @@ impl GeometryRouter {
             }
         }
 
+        // Break coplanar Z-fight between sub-meshes BEFORE placement so the
+        // local-frame AABB heuristic is valid for any product orientation.
+        super::zfight::dedupe_coplanar_submeshes(&mut sub_meshes);
+
         // Apply placement transformation to all sub-meshes
         // ObjectPlacement translation is in file units (e.g., mm) but geometry is scaled to meters,
         // so we MUST scale the transform to match. Same as apply_placement does.
--- a/rust/geometry/src/router/zfight.rs
+++ b/rust/geometry/src/router/zfight.rs
@@ -1,0 +1,252 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Coplanar Z-fight dedup for multi-sub-mesh elements.
+//!
+//! Revit-exported doors/windows often ship the glass pane as a thin brep
+//! whose front/back faces sit exactly on the metal-trim faces. The glass and
+//! trim then Z-fight when rendered. This pass detects axis-aligned AABB
+//! faces shared between two sub-meshes of the SAME element with overlapping
+//! 2D projection and shrinks the smaller (by vertex count) sub-mesh inward
+//! by `eps` along that axis so the faces no longer overlap. Detection is
+//! done in the element's local frame BEFORE placement is applied, so the
+//! axis-aligned assumption holds for any product orientation.
+//!
+//! `eps` is auto-derived as `5e-5 × max_sub_mesh_extent` so the offset is
+//! always ~0.05–0.1 mm at typical IFC product scales (millimetre or metre
+//! file units) without callers having to thread `unit_scale` through. That
+//! is well below any meaningful BIM tolerance and below f32 precision at
+//! typical building scales, but large enough to break Z-fight in a 24-bit
+//! depth buffer.
+//!
+//! Non-glass sub-meshes still benefit: any element whose sub-shells share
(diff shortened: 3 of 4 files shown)


louistrue:
Closing — the diagnosis behind this PR turns out to be wrong. The Z-fight visible on the door glass is NOT between glass and metal-trim sub-meshes; it's INSIDE the panel brep #712 itself.

`process_planar_face` in `rust/geometry/src/processors/advanced_face.rs` triangulates each `IfcFaceBound` of an `IfcAdvancedFace` as an independent polygon, so a face with an outer bound + inner (hole) bound produces:

- the outer quad triangulated as a solid (the hole ignored)
- the inner bound triangulated as a separate solid, with reversed winding from the schema convention → opposite normal

These two are exactly coplanar inside the cutout footprint. The renderer uses \`cullMode: 'none'\` (\`packages/renderer/src/pipeline.ts:141\`, comment "IFC winding order varies") so the canceling pair renders as a wedge-shaped Z-fight. Result: 24 tris where IfcOpenShell emits 32 (the 8 missing per face are exactly the annular triangulation around the hole).

The correct fix is to make \`process_planar_face\` identify the outer bound, treat \`IfcFaceBound\` siblings as holes, and call the existing \`triangulate_polygon_with_holes\` helper from \`triangulation.rs\` — exactly the pattern the \`brep.rs\` faceted-brep path already uses (\`brep.rs:287-345\`).

Replacement PR coming. Element-level AABB dedup is unnecessary once the single-brep triangulation is right.

PR #850, opened 2026-05-28
fix(geometry): iterative BSP CSG kernel + spatial-tree cycle guard (#841)

### LINKED ISSUE #841: Maximum call stack size exceeded
In the attached model（House）, the console outputs an error when viewed in ifc-lite: RangeError: Maximum call stack size exceeded, referring to the following comparison images.

[House.zip](https://github.com/user-attachments/files/28329901/House.zip)

Image

Image

louistrue:
## Summary

Stop tripping `Maximum call stack size exceeded` / `too much recursion` when loading the House.ifc fixture. Two independent fixes:

### 1. Iterative BSP CSG kernel — the actual House.ifc root cause

`rust/geometry/src/bsp_csg.rs` was a textbook recursive BSP. On House.ifc's 51 faceted Breps (4120 `IfcPolyLoop`s total) the partition tree degenerates into a near-linked-list because so many wall/roof polygons are coplanar. The recursive walk depth tracks the polygon count; native rides on an 8 MB stack and never notices, but Firefox enforces a combined JS+WASM call-depth limit around 10K frames and threw `InternalError: too much recursion` mid-`parseMeshes`, hiding every wall, roof and floor in the fixture.

Rewrite `BspNode::build`, `clip_polygons`, `clip_to`, `invert` and `all_polygons` with explicit `Vec`-based work stacks. Process-level call depth is now O(1) regardless of tree shape. Also replace the default recursive `Drop` for `Box<BspNode>` — destroying a depth-4096 tree the naive way overflowed on its own.

### 2. Spatial-hierarchy cycle guard — defensive JS-side fix

`rebuildSpatialHierarchy::buildNode` recursed on every aggregated child without a visited-set guard. A malformed `IfcRelAggregates` back-edge would loop indefinitely. Not the cause of the reported House.ifc failure (the WASM crash fires first) but a real robustness gap for adversarial files; the guard stays.

Fixes #841.

## Test plan

- [x] `rust/geometry/src/bsp_csg.rs::stack_safety_tests::deep_bsp_does_not_overflow` — feeds a depth-4096 degenerate quad stack through the entire build / walk / invert / clip / drop pipeline. The old recursive code would have OOM-stacked native too at this depth.
- [x] `cargo test -p ifc-lite-geometry --test issue_841_house_stack_overflow` — drives every renderable product in the reporter's House.ifc through `process_element`.
- [x] `apps/viewer/src/utils/spatialHierarchy.test.ts` — JS cycle-guard regression with a Project → Site → Building → Storey graph whose storey aggregates back to the building.
- [x] `cargo test -p ifc-lite-geometry` (full suite, 250+ tests) — no regressions in any existing CSG/Brep test.
- [ ] Upload fixture `tests/models/issues/841_house_stack_overflow.ifc` (1.4 MB) to the `fixtures-v1` GitHub Release via `pnpm fixtures:upload` after merge.

### DIFF
--- a/.changeset/fix-841-spatial-tree-cycle-guard.md
+++ b/.changeset/fix-841-spatial-tree-cycle-guard.md
@@ -1,0 +1,31 @@
+---
+"@ifc-lite/viewer": patch
+---
+
+Add a cycle guard to the spatial-hierarchy builder so a malformed
+`IfcRelAggregates` edge that points back at an ancestor no longer trips
+"RangeError: Maximum call stack size exceeded" before the viewer can
+draw anything (issue #841).
+
+`rebuildSpatialHierarchy::buildNode` recursed on every aggregated child
+that was itself a spatial-structure type. Without a visited-set guard
+the recursion looped indefinitely along the cycle. The within-node
+descendant walk (used for `elementToStorey` propagation) already had a
+cycle guard, but the top-level node-to-node recursion did not. The
+fix shares a `visited` set across the whole `buildNode` recursion and
+skips children that are already in flight, so first-visit wins and any
+back-edge is silently dropped.
+
+Regression coverage:
+
+- `apps/viewer/src/utils/spatialHierarchy.test.ts` — new
+  "terminates without stack overflow on cyclic spatial aggregation
+  (issue #841)" test pins the guard with a Project → Site → Building →
+  Storey graph whose storey aggregates back to the building.
+- `rust/geometry/tests/issue_841_house_stack_overflow.rs` — smoke test
+  that drives every renderable product in the reporter's House.ifc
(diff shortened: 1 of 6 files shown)


louistrue:
Superseded by #861 — that PR swaps the wasm CSG kernel from the in-tree BSP port to Manifold, so the iterative-BSP recursion fix in this branch no longer covers a reachable code path (Manifold doesn't recurse and has no operand cap).

The bundled JS-side spatial-hierarchy cycle guard (`apps/viewer/src/utils/spatialHierarchy.ts`) is independent of the kernel choice and a real robustness gap; salvageable into a small follow-up PR if it becomes a priority. Closing per @louistrue's instruction.

Issue #859, opened 2026-05-28
Unsupported parsing of IfcRailway and infrastructure entities


xyzbety:
The parser fails to recognize or render IfcRailway and other infrastructure extension entities. When loading such models, related objects are missing or display incorrectly. Compatibility support for the IFC Infrastructure extension schema is required.

The attachment（linear-placement-of-signal） is a model of the IfcRailway. Referring to the following comparison images.

[linear-placement-of-signal.zip](https://github.com/user-attachments/files/28334756/linear-placement-of-signal.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/94323fab-c764-4ab4-b38c-ffecfd93e802" />

<img width="2880" height="1719" alt="Image" src="https://github.com/user-attachments/assets/dee6e5f9-d49b-43f7-ae56-2f28f267c799" />

xyzbety:
Adding an IFCBridge model（a99） for testing, which appears to parse normally, but the spatial tree does not have a bridge icon. @louistrue 

[a99.zip](https://github.com/user-attachments/files/28335002/a99.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/85b86f0b-bcbb-4639-9dcb-356055eb6168" />

xyzbety:
Adding an IFCRoad model（AB22） for testing, which appears to parse normally, it seems the only lack is the type definition for IFCRailway. However, there are no icons in the space tree either. @louistrue 

[AB22.zip](https://github.com/user-attachments/files/28335481/AB22.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/7327176f-0c76-43f8-8b59-cf88bbce7083" />

Issue #293, opened 2026-03-03
string-encoding error


RyugaRyuzaki:
When parse string-encoding 
 https://technical.buildingsmart.org/resources/ifcimplementationguidance/string-encoding/

<img width="413" height="830" alt="Image" src="https://github.com/user-attachments/assets/f5581543-c5a5-46de-baee-c60f9ed65873" />

PR #886, opened 2026-05-29
fix(geometry): resolve IfcGridPlacement so grid-placed elements land correctly

### LINKED ISSUE #883: Column not rendering correctly with grid-relative placement
Columns are not being rendered correctly when their placement is defined relative to grids. The grid placement information is not being properly interpreted or applied during the rendering process.

### Expected Behavior

Image

### Actual Behavior:

Image

File:

[ifcgrid.ifc.txt](https://github.com/user-attachments/files/28406187/ifcgrid.ifc.txt)

louistrue:
Fixes #883.

## Problem

Columns laid out on a structural grid rendered in the wrong place — all stacked near the world origin instead of on their grid-axis intersections. The reporter's `ifcgrid.ifc` (IFC4X3_ADD2) places 25 `IfcColumn`s via `IfcGridPlacement → IfcVirtualGridIntersection`.

**Root cause:** the placement resolver (`rust/geometry/src/router/transforms.rs`) only handled `IfcLocalPlacement` and `IfcLinearPlacement`; every other `IfcObjectPlacement` subtype fell through to an identity matrix:

```rust
if placement.ifc_type != IfcType::IfcLocalPlacement {
    return Ok(Matrix4::identity());   // ← IfcGridPlacement landed here
}
```

So all grid-placed elements collapsed onto the same spot. (Same class of bug as #859's `IfcLinearPlacement`.)

## Fix

Add an `IfcGridPlacement` branch + `resolve_grid_placement_with_depth`:

- **`PlacementRelTo`** (attr 0) composes exactly like `IfcLocalPlacement` — it carries the grid's own world placement.
- **`PlacementLocation`** (attr 1, `IfcVirtualGridIntersection`) → `grid_intersection_point`: read the two referenced `IfcGridAxis` curves (via the existing `ProfileProcessor::get_curve_points`), intersect them in the grid plane (`line_intersection_2d`), apply the per-axis lateral `OffsetDistances` (shift each axis along its left normal) and the optional elevation (third offset → z).
- **`PlacementRefDirection`** (attr 2) — **full `IfcGridPlacementDirectionSelect` coverage**:
  - `IfcDirection` → its XY sets local +X.
  - `IfcVirtualGridIntersection` → local +X points from the placement location to that second intersection.
  - null / unresolved → axis-aligned (inherits the grid orientation).

Composes `parent * local` and degrades gracefully (parent/identity) on malformed or parallel-axis input — never panics.

### Scope note (documented in code)
Grid axes are treated as straight lines (chord of first→last sample) — true for the fixture and virtually all real grids; curved axes would need arc-length sampling.

## Tests

- **`rust/geometry/tests/issue_883_grid_placement.rs`** (integration, reporter's fixture): columns **#293** (axis E × row 1) and **#519** (axis E × row 2), authored 4000 mm apart, must mesh **≈ 4.0 m** apart instead of collapsing to one point; all **25** grid-placed columns must spread across the grid. Fixture committed as `…/fixtures/issue_883_grid_placement.ifc.txt` (the `.ifc.txt` extension dodges the `*.ifc` fixture-fetch gitignore) and embedded via `include_str!`, so the regression runs in CI without the on-demand fixture store.
- **Crate-internal unit tests** (`grid_placement_tests` in `transforms.rs`): assert the resolved transform directly for all three `PlacementRefDirection` cases (IfcDirection, IfcVirtualGridIntersection, null) and the `OffsetDistances` shift (perpendicular per-axis offset + elevation).

## Verification
- New tests: integration **2/2**, orientation unit tests **4/4**.
- Full `ifc-lite-geometry` suite: **all green** (192 lib tests + integration tests, incl. the `#859` linear-placement test that shares the edited function). No new warnings.
- Also syncs `Cargo.lock` to the workspace's already-committed `3.0.0` crate versions (pre-existing drift my build corrected).

> Note: the viewer ships prebuilt WASM, so this lands once the WASM is rebuilt from the Rust core (CI handles that on release).

## Release Notes

* **New Features**
  * Added comprehensive support for grid-based placements with automatic axis intersection calculation and optional reference direction orientation.

* **Bug Fixes**
  * Enhanced placement transform resolution with graceful fallback handling for incomplete or degenerate grid configurations.

* **Tests**
  * Added unit tests covering grid placement scenarios including reference direction handling and offset calculations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/LTplus-AG/ifc-lite/pull/886?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


louistrue:
Thanks for the reviews — addressed in `10b0d71`, and also cleaned up the branch.

**Branch hygiene:** this PR had accidentally been cut from the visibility-menu branch, so its diff included viewer commits from #885 (merged) and the closed #882. Rebased onto clean `main` — it's now **rust-only** (`transforms.rs` + `Cargo.lock`).

**Codex P1 — committed IFC fixture (AGENTS.md §9):** correct. The `.ifc.txt` suffix only dodged the `*.ifc` gitignore and bypassed the `tests/models` + manifest workflow (I don't have fixture-Release write access to publish it properly). Removed the committed fixture *and* its file-based integration test. Coverage is preserved by the inline `grid_placement_tests` unit module, which asserts the resolved transform **directly** (origin, +X axis, per-axis offsets) — more precise than the centroid-separation heuristic — plus a new `PlacementRelTo`-composition case so the parent ∘ local path stays covered without any fixture.

**CodeRabbit — split doc comment:** correct, fixed. `sample_polyline_at_distance`'s doc had been split when the grid-math helpers were inserted; each function's `///` block is contiguous again.

**Codex P2 — bare `catch` in `resetTypeVisibility`** and **CodeRabbit nitpick — `visibilitySlice.test.ts` should import `TYPE_VISIBILITY_SEMANTIC_DEFAULTS`:** both valid, but they live in `visibilitySlice.{ts,test.ts}` which is **already merged via #885** and is no longer in this PR after the rebase. The `catch` also mirrors the pre-existing `toggleTypeVisibility` pattern in the same file. I'll handle these as a small separate follow-up rather than re-introduce viewer churn here.

PR #889, opened 2026-05-29
feat(viewer): render IfcAlignment as thin centerline lines, not a ribbon

louistrue:
## What

Render `IfcAlignment` as a thin **centerline line**, exactly like IfcGrid axes and IfcAnnotation curves — instead of the triangulated **ribbon** strip this PR originally shipped.

> This replaces the earlier ribbon approach (the batch-mesher route + RTC ribbon fix have been reverted; the branch is now a clean lines-only diff vs `main`).

## Why

`IfcAlignment` carries its geometry in the `Axis` curve (`IfcAlignmentCurve` / `IfcPolyline`), not a `Representation`, so it never meshes in the streaming batch path. Rendering it as a swept ribbon read as a thin solid band rather than the centerline users expect.

## How

The directrix is sampled into a flat Y-up line-list and drawn through the renderer's existing line pipeline:

- **`@ifc-lite/wasm`** — new `IfcAPI.parseAlignmentLines(content)` → `Float32Array` of `[x,y,z,…]` line-list vertices in renderer world space (Y-up, RTC-subtracted, metres — the same frame the mesh pipeline produces, so the line lands on the same ground as terrain).
- **`@ifc-lite/geometry`** — `GeometryProcessor.parseAlignmentLines` wrapper.
- **`@ifc-lite/renderer`** — `uploadAlignmentLines3D` / `clearAlignmentLines3D` (separate vertex buffer, reuses the line pipeline).
- **viewer** — `useAlignmentLines3D` hook, always-on (no toggle); `Viewport` uploads/clears alongside the annotation overlay.

The batch mesher no longer special-cases `IfcAlignment` into the ribbon processor, so alignment is **lines-only** — never both.

## Testing

- `alignment_lines` Rust unit tests (planar polyline → even-count line-list mapped to renderer origin/extent; empty file → empty array).
- `cargo test` (wasm `alignment_lines`, geometry) green; `cargo check -p ifc-lite-wasm` clean.
- WASM bundle rebuilt; viewer `tsc --noEmit` passes.

## Changeset

`@ifc-lite/wasm` + `@ifc-lite/geometry` + `@ifc-lite/renderer` (minor). The shared `ifc-lite-geometry` crate is untouched, so `@ifc-lite/server-bin` is no longer bumped.

### DIFF
--- a/.changeset/fix-alignment-batch-render.md
+++ b/.changeset/fix-alignment-batch-render.md
@@ -1,0 +1,35 @@
+---
+"@ifc-lite/wasm": minor
+"@ifc-lite/geometry": minor
+"@ifc-lite/renderer": minor
+---
+
+Render `IfcAlignment` as a thin centerline **line** instead of a triangulated
+ribbon, matching how IfcGrid axes and IfcAnnotation curves draw.
+
+`IfcAlignment` carries its geometry in the `Axis` curve (`IfcAlignmentCurve` or
+`IfcPolyline`), not a `Representation`. Previously the streaming batch mesher
+routed it through the whole-element `IfcAlignmentProcessor`, which sampled the
+directrix into a thin solid ribbon strip — visually wrong for what is a
+centerline. Now the alignment is sampled straight into a line-list overlay:
+
+- **`@ifc-lite/wasm`** gains `IfcAPI.parseAlignmentLines(content)`, which walks
+  every `IfcAlignment`, resolves its `Axis` directrix, samples the centerline
+  (1 file-unit station spacing, adaptive cap at 5000 samples) and returns a flat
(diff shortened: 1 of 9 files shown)


louistrue:
Addressed in `7a8f` (P2). `process_element`'s IfcAlignment special-case scaled the ribbon but returned before the RTC subtraction that `apply_placement` does for placed meshes — so on georeferenced files the ribbon stayed in absolute coords. Now subtracts the router RTC offset (identity transform = RTC-only, guarded by `has_rtc_offset`) right after `scale_mesh`, at the source in `process_element` so every caller (incl. the batch path) is covered. Added a regression test asserting the ribbon shifts by −offset when RTC is set. Residual note: vertices are f32 by the time they're produced, so extreme-coordinate jitter is reduced (RTC-relative) but a full-precision f64 path inside the processor is a separate follow-up.

PR #893, opened 2026-05-30
feat(clash): Rust/WASM geometry kernel + differential parity (Phase 3)

louistrue:
Stacked on #892. The clash kernel now has a high-performance **Rust implementation compiled to WASM**, behind the same orchestrator as the TS engine and **proven equal to it** by a differential test.

## What

- **`rust/clash`** (new crate `ifc-lite-clash`): pure-std geometry kernel — AABB BVH broad phase + exact triangle-triangle intersection (SAT) and minimum distance (Ericson), per-element triangle BVHs, and the narrow-phase classification ported **f64-faithfully** from the TS engine. 12 cargo tests + doctest, clippy-clean, no `unsafe`.
- **`rust/wasm-bindings/src/api/clash.rs`**: `#[wasm_bindgen] ClashSession` — ingest flat geometry arenas, `runRule` → packed typed-array results. The JS↔WASM contract is **pure numbers**; selection / exclusions / severity / identity stay in TS.
- **`@ifc-lite/clash/wasm`** (subpath): `WasmClashEngine` + `initClashWasm`, reusing the shared orchestrator with a `WasmKernel`. Kept **off the core import graph** so `@ifc-lite/wasm` is never pulled into the core.
- **Engine refactor** (one commit): the TS engine is now `orchestrator + ClashKernel`. `TsKernel` is the reference oracle; the WASM kernel drops in behind the same seam — which is what makes the two **differentially comparable**.

## Differential test

Loads the web-target WASM in Node (by feeding the `.wasm` bytes) and asserts **WASM === TS** across hard / clearance / touch / self-clash and the **full discipline matrix**: identical clash sets (by id), status and severity; distances and points within `1e-6` (f64 summation order differs by ~1e-12).

## Notes

- `backend: 'auto'` stays the in-process TS engine; WASM is **opt-in** via the subpath (it needs async module init). A host (viewer worker, CLI-in-Node) chooses it explicitly.
- Generated `packages/wasm/pkg` artifacts are **gitignored** and rebuilt by CI (`build-wasm.sh`), so this PR is source-only — no binary diff.
- The same-entity (duplicate-key) self-clash skip was moved into the shared orchestrator so every kernel behaves alike regardless of how its broad phase dedups.
- Full suite: **12 files, 77 tests**; `cargo test -p ifc-lite-clash` green.

## Still remaining (dedicated passes)
P1 viewer panel, P4 MCP/CLI/scripts, P7 desktop migration. Worker-pool sharding of the WASM narrow phase (the 3b scaling sub-step) is also a follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2422,6 +2422,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "ifc-lite-clash"
+version = "3.0.0"
+
+[[package]]
 name = "ifc-lite-core"
 version = "3.0.0"
 dependencies = [
@@ -2528,6 +2532,7 @@ version = "3.0.0"
 dependencies = [
  "console_error_panic_hook",
  "futures-util",
+ "ifc-lite-clash",
  "ifc-lite-core",
  "ifc-lite-geometry",
  "ifc-lite-processing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["rust/core", "rust/geometry", "rust/processing", "rust/wasm-bindings", "apps/server", "apps/desktop/src-tauri"]
+members = ["rust/core", "rust/geometry", "rust/processing", "rust/clash", "rust/wasm-bindings", "apps/server", "apps/desktop/src-tauri"]
 resolver = "2"
 
 [workspace.package]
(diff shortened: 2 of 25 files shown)


louistrue:
Consolidated into #891. The Rust/WASM kernel + differential parity commits are now part of #891, along with the review-feedback fixes.

PR #658, opened 2026-05-13
Fix/cesium placement ux

louistrue:
---
<a href="https://app.blacksmith.sh/louistrue/codesmith/ifc-lite/pr/658"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added interactive Cesium placement editor with visual gizmo controls for adjusting model georeferencing, including coordinate and height adjustments.
  * Restored IFC2X3 legacy site-based georeference support from older model files.
  * Implemented separate camera bridge positioning for Cesium overlays.
  * Enhanced terrain elevation sampling with explicit source and reference tracking.

* **Improvements**
  * Streamlined Cesium controls by removing settings dialog in favor of toolbar-based placement editing.

PR #895, opened 2026-05-30
fix(viewer): co-locate federated models (georef alignment + unit scale + ingest watchdog)

louistrue:
## Problem

Loading two IFC models together broke in three distinct ways (reported for the BIMcollab ARC/STR sample):

1. **Second model flung ~20 km away / dwarfed** — the headline "federation is broken" symptom.
2. **Models 1000× oversized** when a model's units were declared after its geometry.
3. **Hang at "Processing geometry (N meshes)"** when adding a model whose geometry worker failed to spawn.

## Root cause (the 20 km offset)

When a file has no `IfcMapConversion`, we synthesise a `source: 'siteLocation'` georeference from its `IfcSite` `RefLatitude`/`RefLongitude`/`RefElevation` (EPSG:4326) so the model can still be pinned on the location map. Since #658 the federated add-model path treated that synthetic georef as *real* and ran it through the projected-CRS affine alignment (`buildGeorefAlignmentTransform`).

But site-location coordinates are geographic **degrees** plus a **raw, un-unit-scaled site elevation** — not projected metres. The ARC/STR pair share a site GUID but carry `RefElevation` `0` vs `20000` mm, so the alignment's height term placed ARC ~20 km below STR.

**Fix:** `extractModelGeoref` now requires true georeferencing (`hasStandardGeoreferencing` → `IfcMapConversion` + `IfcProjectedCRS`). Site-location-only models stay in their own local frames, where they already overlay correctly.

## Also fixed

- **Unit scale.** `buildPrePassStreaming` resolved `unitScale` from a *partial* entity index (rows up to the first `IFCPROJECT`). Revit-style exports place `IFCUNITASSIGNMENT` after the geometry, so `IFCSIUNIT` wasn't indexed yet and resolution fell back to the metres default → 1000× oversize. Now falls back to a *complete* index when the unit chain isn't yet decodable (`try_extract_length_unit_scale`, unit-tested).
- **Ingest watchdog.** `parseStepBufferViewerModel` gains the same size-aware stream watchdog the single-model loader had, so a stalled geometry stream surfaces a recoverable error instead of hanging forever.
- **Camera framing.** Adding a model unions the bounds of all visible models and refits, so federated models are framed together.

## Verification

Local build (freshly rebuilt WASM with the unit fix) + Playwright loading the ARC/STR sample in **both orders**:

| order | ARC center | STR center | max spread | alignment status |
|---|---|---|---|---|
| STR → ARC | (22.7, 6.4, −11.5) | (23.1, 5.4, −11.7) | **1.0 m** | `none` |
| ARC → STR | (22.7, 6.4, −11.5) | (23.1, 5.4, −11.7) | **1.0 m** | `none` |

Was ~20,000 m apart. Model sizes (~13–16 m) confirm correct mm→m scale. Viewer test suite: 932 pass, 0 fail.

Removed the earlier `recenterFarGeometry` band-aid — it ran *before* the alignment that actually moved the model, so it never fired; it was treating a symptom of this alignment bug.


louistrue:
## Review feedback — resolution status

Ran an independent multi-agent audit of the **current** diff (the georef fix and latest commits were never bot-reviewed — Codex reviewed `78c76041`, CodeRabbit auto-paused) and reconciled every prior bot finding against HEAD. Summary:

| Finding | Status |
|---|---|
| **CodeRabbit (Major)** — `viewerModelIngest.ts` watchdog `geometryIterator.return()` can hang behind the stalled `next()`, masking the timeout | ✅ **Fixed** (`c0875154`). Teardown is now bounded (`boundedIteratorReturn`, 2 s race); the watchdog + cleanup are extracted into a shared `watchedGeometryStream` helper used by both loaders. Confirmed with a Node repro of the exact `try/catch/finally`. |
| **CodeRabbit (minor)** — remove `[DIAG render]` block in `useGeometryStreaming.ts` | ✅ **Resolved** (`c61586bb`) — block removed. |
| **Codex (P1)** — `geometry-parallel.ts` queues jobs behind a posted `stream-end` | ⚪️ **Moot** — `geometry-parallel.ts` was reverted to `main` (`1021dcef`) and is byte-identical; the worker-resilience attempt is gone, the watchdog is the anti-hang mechanism. |
| **CodeRabbit (Major)** — `geometry-parallel.ts` ignore `ready` from retired workers | ⚪️ **Moot** — same (file == `main`). |
| **CodeRabbit (nit)** — `geometry-parallel.ts` catch-comment form | ⚪️ **Moot** — same (file == `main`). |
| **CodeRabbit (nit)** — `gpu_meshes.rs` builds a 2nd full index on the unit-scale fallback | ☑️ **Acknowledged, left as-is** — correctness is fine (CodeRabbit agreed); the extra pass only fires on geometry-first (Revit) exports where units follow geometry. Optional perf follow-up noted in the commit. |

Additional hardening from the audit (`c0875154`): canonical `/* cleanup — safe to ignore */` catch (AGENTS.md §7), `viewerModelIngest.ts` back under the ~400-line limit (409 → 375) by removing the duplicated watchdog loop, and new tests — `streamCleanup` (incl. the never-settling-`return()` case), `watchedGeometryStream`, and direct `hasStandardGeoreferencing` tests locking the federation-gate invariant.

Verification: full viewer suite **945 pass / 0 fail**; federated ARC+STR e2e co-locates in both load orders (was ~20 km apart). All CI green prior to this push.

PR #899, opened 2026-05-31
fix(geometry): centre U/L/T parametric profiles on their bounding box

Blogbotana:
## Problem

`IfcUShapeProfileDef`, `IfcLShapeProfileDef` and `IfcTShapeProfileDef` are generated from a corner instead of centred on their bounding box, so swept solids using them are offset by half a dimension:

- **U-shape (channel):** built `0 … FlangeWidth` in X (Y already centred) → offset by **+FlangeWidth/2**
- **L-shape (angle):** corner at origin (`0 … Width`, `0 … Depth`) → offset by half each leg
- **T-shape:** built `0 … Depth` in Y (X already centred) → offset by **+Depth/2**

The swept-area `Position` placement is applied relative to that origin, so the offset shows up directly in the model. The symmetric profiles (I-shape, asymmetric-I, rectangle, circle, Z, hollow) are already centred, so only the asymmetric ones are affected.

### Evidence (real file, U16 channel, `FlangeWidth = 64 mm`)

`IfcAPI.extractProfiles` outer points, before:
```
X range: 0 … 0.064   (centre +0.032)   ← offset by half the flange width
Y range: -0.08 … 0.08 (centre 0)        ← already centred
```
after this change:
```
X range: -0.032 … 0.032 (centre 0)
Y range: -0.08 … 0.08   (centre 0)
```

## Fix

- Add `Profile2D::center_on_bbox()`.
- Call it in `process_parametric` after building the base profile and **before** the `Position` transform, so every parametric profile is centred on its bounding box — the convention already documented for `process_asymmetric_i_shape` (matches IfcOpenShell / Tekla / Revit). It is a no-op for the already-centred shapes.

## Tests

Added regression tests `test_u_shape_is_centered`, `test_l_shape_is_centered`, `test_t_shape_is_centered` (assert the outer bounding box is centred on the origin and has the expected extents).

```
cargo test -p ifc-lite-geometry --no-default-features profiles::tests
```

Also verified end-to-end against a rebuilt wasm bundle (`extractProfiles` now returns centred outlines; channels render in the correct position in the viewer).

## Note / possible follow-up

While here I noticed `process_c_shape` (`IfcCShapeProfileDef`) uses `Girth` for the flange extent and ignores `Width`, and emits no inturned lips — that looks like a separate dimensional bug (wrong size, not just position). Left out of this PR to keep it focused; happy to file/fix separately.

A changeset (`@ifc-lite/wasm` patch) is included.

## Release Notes

* **Bug Fixes**
  * U-Shape, L-Shape, and T-Shape parametric profiles now generate centered on their bounding box origin instead of from a corner. Internal geometry offsets have changed; models using these profiles may render differently.

### DIFF
--- a/.changeset/centre-parametric-profiles.md
+++ b/.changeset/centre-parametric-profiles.md
@@ -1,0 +1,12 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Centre parameterised profiles on their bounding box.
+
+`IfcUShapeProfileDef`, `IfcLShapeProfileDef` and `IfcTShapeProfileDef` were
+generated from a corner instead of centred on their bounding box, so channels were
+offset by half the flange width, angles by half their leg lengths and tees by half
+their depth. The base profile is now centred before the swept-area `Position`
+placement is applied, matching the symmetric profiles (I-shape, rectangle, …) and
+the IfcOpenShell/Tekla/Revit convention.
--- a/rust/geometry/src/profile.rs
+++ b/rust/geometry/src/profile.rs
@@ -30,6 +30,47 @@ impl Profile2D {
         self.holes.push(hole);
     }
 
+    /// Translate the profile so the centre of its outer bounding box sits at the
+    /// origin.
+    ///
+    /// IFC parameterised profiles (I/U/L/T/C/Z/…) are defined centred on their
+    /// bounding box, and the swept-area `Position` placement is applied relative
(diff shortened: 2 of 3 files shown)


louistrue:
thx, @Blogbotana 👍 a fix for IfcCShapeProfileDef would be much appreciated too 🙌

Blogbotana:
Thank you @louistrue! If I notice any new issues, I’ll be happy to help the community grow👍

PR #906, opened 2026-06-01
feat(server): IfcAnnotation/IfcGrid symbol-data parity across all parse endpoints (#900)

louistrue:
Closes #900.

## Summary

Issue #843 added the 2D symbol stream (`symbolic_data` — `IfcAnnotation` + `IfcGrid`) to the synchronous `POST /api/v1/parse` JSON response. The reporter (#900) noticed every *other* geometry/parsing endpoint still dropped it, so the data was reachable only through that one endpoint. This brings the **whole** Server API to parity with `@ifc-lite/parse`.

The transport is matched per endpoint, reusing the data-model "cache + fetch-by-`cache_key`" pattern (symbol data can be large for annotation-heavy drawings, so it never travels in an HTTP header):

| Endpoint | Transport | How symbol data is delivered |
|---|---|---|
| `POST /api/v1/parse` | JSON | inline `symbolic_data` (already) + cached for fetch |
| `GET /api/v1/cache/:key` | JSON (cached `ParseResponse`) | inline (already) |
| `POST /api/v1/parse/stream` | SSE JSON | inline in the `complete` event |
| `POST /api/v1/parse/parquet-stream` | SSE | inline in the `complete` event + cached |
| `POST /api/v1/parse/parquet` | binary + header | cached `{cache_key}-symbolic-v1` |
| `POST /api/v1/parse/parquet/optimized` | binary + header | cached `{cache_key}-symbolic-v1` |
| `GET /api/v1/cache/geometry/:hash` | binary (cached) | via the cache key |
| **`GET /api/v1/parse/symbolic/{cache_key}`** | **new** | JSON symbol stream for the binary transports |

`POST /api/v1/parse/metadata` is intentionally excluded (counts only, no geometry).

## Changes

**Server (`apps/server`, ships in `@ifc-lite/server-bin`):**
- `StreamEvent::Complete` / `ParquetStreamEvent::Complete` carry `symbolic_data`; it's extracted once at the end of the shared `process_streaming`, so both streaming endpoints get it with no extra parse.
- `/parse/parquet` + `/parse/parquet/optimized` extract symbols in parallel with geometry (`rayon::join`) and cache them.
- New `GET /api/v1/parse/symbolic/{cache_key}` handler (mirrors `get_data_model`; `202` while a stream caches in the background).
- `parse_full` and the cached-geometry fast paths populate the same `{cache_key}-symbolic-v1` entry, so the fetch endpoint works no matter which endpoint processed the file first.
[...]

Issue #900, opened 2026-06-01
Align all Server geometry/parsing endpoints with @ifc-lite/parse, not just POST /api/v1/parse


PDB6912:
Thank you for the quick update #834 The 2D symbol data works perfectly on the `POST /api/v1/parse` endpoint now.

However, I noticed that this modification was only applied to that specific endpoint. Other endpoints on the Server that return geometry or parsed IFC data still lack the `IfcAnnotation` and `IfcGrid` information.

Could we please bring this feature parity to **all** relevant endpoints on the Server? The goal is to fully align the Server's entire parsing capability with `@ifc-lite/parse`, ensuring consistent data output across the whole API.

Thank you again for your time and effort!

PR #907, opened 2026-06-01
feat(server): georeferencing + length-unit scale on all geometry endpoints (#900 parity)

louistrue:
Follow-up to #906 (which closed the IfcAnnotation/IfcGrid gap). This is the next item from the broad parity pass against `@ifc-lite/parse`: **georeferencing** and the **length-unit scale**.

## Why

The browser parser exposes `IfcMapConversion` / `IfcProjectedCRS` georeferencing (`extractGeoreferencing`) and the length-unit scale (`extractLengthUnitScale`). The server returned only a coarse `is_geo_referenced` boolean and kept the unit scale internal — so API consumers couldn't recover the real-world CRS, false eastings/northings, grid-north rotation, or the mm→m factor.

A complete georeferencing extractor **already existed unused** in shared Rust (`ifc_lite_core::GeoRefExtractor`), so this is a lightweight wire-up rather than a port.

## What

Both values now ride inline on `ModelMetadata`, which every geometry endpoint already embeds — so they reach **all** of them at once (JSON `/parse`, SSE `/parse/stream`, `/parse/parquet`, `/parse/parquet/optimized`, `/parse/parquet-stream`, and the cached-geometry paths) with no new endpoint or fetch round-trip. For the Parquet endpoints they travel in the existing `X-IFC-Metadata` header (georef is small/fixed-size, unlike the symbol stream).

**Server (`@ifc-lite/server-bin`):**
- `ModelMetadata` gains `length_unit_scale: Option<f64>` and `georeferencing: Option<Georeferencing>` (CRS name, geodetic + vertical datum, map projection, false eastings/northings, orthogonal height, X-axis direction, scale, derived `rotation_degrees`, column-major local→map `transform_matrix`).
- New `ifc_lite_processing::extract_georeferencing` reuses `ifc_lite_core::GeoRefExtractor` (same extraction the native/desktop paths use, incl. the IFC2x3 `ePSet_MapConversion` fallback).
- Populated in `process_geometry_filtered` and the server streaming `Complete` event (extracted on a blocking thread).

**Client (`@ifc-lite/server-client`):** new `Georeferencing` type + optional `ModelMetadata.length_unit_scale` / `georeferencing`.

`process_geometry_filtered` is server-only (the browser uses its own `process_geometry_batch`), so the browser geometry path is unaffected.

## Test plan
- [x] `cargo test -p ifc-lite-processing` — incl. `georeferencing.rs` unit tests and `issue_900_georeferencing_metadata.rs` (metre model surfaces CRS/offsets/~30° rotation; mm model reports `length_unit_scale = 0.001` and no georef).
- [x] `cargo test -p ifc-lite-server` green; `cargo check -p ifc-lite-server` clean.
- [x] `@ifc-lite/server-client` builds (`tsc`).
- [x] Confirmed `ModelMetadata` is server-only (WASM uses `ifc-lite-processing` solely for `extract_symbolic_data`).

## Remaining broad-pass gaps (separate follow-ups)
Still TS-only in `@ifc-lite/parse`, needing Rust ports + data-model tables: **classifications** (`IfcClassificationReference`), **structured materials** (`IfcMaterialLayerSet` w/ thicknesses), **documents**, and **schedule/4D**. Tracked for follow-up PRs.

## Summary by CodeRabbit

* **New Features**
  * Geometry endpoints now include optional `length_unit_scale` metadata to convert model length units to meters.
  * Geometry endpoints now include optional `georeferencing` metadata with CRS details, coordinate offsets, rotation, and transformation matrix.

* **Documentation**
  * Server guide updated to document the new georeferencing and unit scale fields available in all geometry endpoint responses.

### DIFF
--- a/.changeset/feat-900-georeferencing-units-parity.md
+++ b/.changeset/feat-900-georeferencing-units-parity.md
@@ -1,0 +1,37 @@
+---
+"@ifc-lite/server-bin": minor
+"@ifc-lite/server-client": minor
+---
+
+Surface georeferencing and the length-unit scale from the Server's geometry
+endpoints, continuing the `@ifc-lite/parse` parity work (issue #900).
+
+The browser parser exposes `IfcMapConversion` / `IfcProjectedCRS` georeferencing
+(`extractGeoreferencing`) and the length-unit scale (`extractLengthUnitScale`),
+but the server returned only a coarse `is_geo_referenced` boolean and kept the
+unit scale internal. Both are now carried inline on `ModelMetadata`, so they
+reach **every** geometry endpoint at once (JSON, SSE, Parquet, optimized Parquet,
+and the cached-geometry paths) — no new endpoint or fetch round-trip.
+
+Server (shipped in the `@ifc-lite/server-bin` binary):
+
+- `ModelMetadata` gains `length_unit_scale: Option<f64>` and
+  `georeferencing: Option<Georeferencing>` (CRS name / geodetic + vertical datum
+  / map projection, false eastings/northings, orthogonal height, X-axis
+  direction, scale, derived grid-north `rotation_degrees`, and a column-major
+  local→map `transform_matrix`).
+- Georeferencing reuses the existing shared `ifc_lite_core::GeoRefExtractor`
+  (the same extraction the native/desktop paths use, including the IFC2x3
+  `ePSet_MapConversion` fallback) via a new `ifc_lite_processing::extract_georeferencing`.
+- Populated in the shared geometry pipeline (`process_geometry_filtered`) and the
+  server's streaming `Complete` event (extracted on a blocking thread).
+
+Client (`@ifc-lite/server-client`):
+
+- New `Georeferencing` type; `ModelMetadata` gains optional `length_unit_scale`
+  and `georeferencing`.
+
+Regression coverage: `rust/processing/tests/issue_900_georeferencing_metadata.rs`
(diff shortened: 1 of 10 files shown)


louistrue:
Thanks for the review. Addressed in 81b70ac6:

1. **IFC2x3 `ePSet_MapConversion` fallback (P2).** Real bug in the shared `GeoRefExtractor::extract_from_pset` that my PR surfaces — it read the property-set name from `get_string(0)` (GlobalId), but `IfcPropertySet.Name` is attribute 2, so the fallback never matched. Fixed to attribute 2, with a new processing-crate regression test driving an IFC2x3 `ePSet_MapConversion` set through `extract_georeferencing`.

2. **Metadata cache version (P2).** Deliberately not bumping here, for two reasons: (a) unlike the symbolic sidecar in #906, the new `georeferencing` / `length_unit_scale` fields are *optional* and degrade gracefully — a stale cache deserializes them as `None`/absent (no hang, no `202`), and clients treat a missing scale as `1.0`; (b) the parquet metadata cache (`*-parquet-metadata-v*`) is already invalidated by the companion PR **#906** (which bumps `-parquet-v2`→`v3`), so once these land together every model reprocesses and the georef fields are written. Bumping again here would just collide with #906's bump. The JSON `ParseResponse` cache is unversioned and refreshes within the 7-day TTL.

`cargo test -p ifc-lite-core georef` and `-p ifc-lite-processing georeferencing` green.

PR #908, opened 2026-06-01
feat(server): classifications, materials & documents in the data model (#900 parity)

louistrue:
Third item from the broad parity pass against `@ifc-lite/parse` (after #906 symbols and #907 georeferencing). You picked classifications + structured materials + documents — bundled here since they share identical plumbing (the `DataModel` struct, the combined Parquet writer, and the TS decoder), so it's one decoder pass and one wire-format change rather than three near-duplicate PRs.

## Why

The browser parser exposes these via `extractClassifications` / `extractMaterials` / `extractDocumentsOnDemand`. The server's data model only recorded the bare `IfcRelAssociatesMaterial` relationship triple (and nothing for classifications or documents). Each is now resolved into a flat, element-keyed shape on the data model returned by `GET /api/v1/parse/data-model/{cache_key}`.

## What

**Server (`@ifc-lite/server-bin`), in `extract_data_model`:**
- **Classifications** (`IfcRelAssociatesClassification` → `IfcClassificationReference`): element id, code (`Identification`), reference name, location, and owning system name (resolved by walking `ReferencedSource` to `IfcClassification`).
- **Materials** (`IfcRelAssociatesMaterial`): resolves `IfcMaterial`, `IfcMaterialLayerSet(Usage)`, `IfcMaterialList`, and `IfcMaterialConstituentSet` into per-layer rows — set name, layer index, material name, **thickness in metres** (unit-scaled), `IsVentilated`, category.
- **Documents** (`IfcRelAssociatesDocument` → `IfcDocumentReference` / `IfcDocumentInformation`): identification, name, location, description.
- Extracted in parallel via `rayon::join`.

**Wire format:** three new Parquet tables appended **after** the existing five. Older clients ignore the trailing bytes; the updated decoder reads them only when present. **No data-model cache-version bump** — so no stale-cache `202` trap; the new tables appear once a file is reprocessed (e.g. after #906's geometry cache bump) and decode to empty arrays meanwhile.

**Client (`@ifc-lite/server-client`):** new `ClassificationAssociation` / `MaterialAssociation` / `DocumentAssociation` types; `DataModel` gains `classifications` / `materials` / `documents`.

## Test plan
- [x] `cargo test -p ifc-lite-server` — incl. new `data_model.rs` tests: a two-layer wall (mm→metre thickness scaling, `IsVentilated`), a Uniclass reference (system name resolved through `ReferencedSource`), and a document reference, all element-keyed; plus an empty-model negative case.
- [x] `@ifc-lite/server-client` builds (`tsc`).
- [x] Branched off `main`, independent of #907.

## Broad-pass status
With this, the gaps you selected are all covered. Remaining (not selected): **Schedule / 4D** (`IfcWorkSchedule` / `IfcTask`).

## Summary by CodeRabbit

* **New Features**
  * Data model now includes per-element classifications, materials (layered with thickness and ventilation), and documents (id, name, location, description); client decoding supports optional appended tables and defaults to empty arrays for older payloads.
  * Server output appends the new tables while preserving backward-compatible payload layout.

* **Documentation**
  * Server Guide updated to describe the new data categories.

* **Tests**
  * Added tests covering extraction and roundtrip of classifications, materials, and documents.

### DIFF
--- a/.changeset/feat-900-datamodel-classifications-materials-documents.md
+++ b/.changeset/feat-900-datamodel-classifications-materials-documents.md
@@ -1,0 +1,42 @@
+---
+"@ifc-lite/server-bin": minor
+"@ifc-lite/server-client": minor
+---
+
+Expand the Server data model with **classifications**, **structured materials**,
+and **documents**, continuing the `@ifc-lite/parse` parity work (issue #900).
+
+The browser parser exposes these via `extractClassifications` / `extractMaterials`
+/ `extractDocumentsOnDemand`, but the server's data model only recorded the bare
+`IfcRelAssociatesMaterial` relationship triple (and nothing for classifications or
+documents). Now each is resolved into a flat, element-keyed shape on the data
+model fetched from `GET /api/v1/parse/data-model/{cache_key}`.
+
+Server (shipped in the `@ifc-lite/server-bin` binary), in `extract_data_model`:
+
+- **Classifications** (`IfcRelAssociatesClassification` → `IfcClassificationReference`):
+  element id, code (`Identification`), reference name, location, and the owning
+  system name (resolved by walking `ReferencedSource` to `IfcClassification`).
+- **Materials** (`IfcRelAssociatesMaterial`): resolves `IfcMaterial`,
+  `IfcMaterialLayerSet(Usage)`, `IfcMaterialList`, and `IfcMaterialConstituentSet`
+  into per-layer rows — set name, layer index, material name, **thickness in
+  metres** (unit-scaled), `IsVentilated`, and category.
+- **Documents** (`IfcRelAssociatesDocument` → `IfcDocumentReference` /
+  `IfcDocumentInformation`): identification, name, location, description.
+
+Each becomes a new Parquet table appended to the data-model payload. The tables
+are appended **after** the existing five, so the format stays backward
+compatible — older clients ignore the trailing bytes, and the updated decoder
+reads them only when present (no data-model cache-version bump, so no stale-cache
+`202` trap; new data appears once a file is reprocessed).
+
+Client (`@ifc-lite/server-client`):
+
+- New `ClassificationAssociation`, `MaterialAssociation`, `DocumentAssociation`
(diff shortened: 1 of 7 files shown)


louistrue:
Thanks for the review. Both findings were valid and are fixed in 520984bc:

1. **`IfcMaterialConstituentSet` constituents (P2).** Were read from attribute 0 (the set name) instead of attribute 2 (`MaterialConstituents`), so constituent-set materials produced no rows. Fixed to `get_list(2)`.

2. **Material profile sets (P2).** Added `IfcMaterialProfileSet` / `IfcMaterialProfileSetUsage` handling (profiles at attribute 2 → `IfcMaterialProfile.Material`), which previously fell through to the default arm — common for structural beams/columns.

Test fixture extended with a column (constituent set) and a beam (profile set); wall-layer assertions scoped by element id. `cargo test -p ifc-lite-server` green.

louistrue:
Second round addressed in d9885f25:

1. **Relationship-graph parity (Major).** `extract_relationships` now whitelists `IFCRELASSOCIATESCLASSIFICATION` and `IFCRELASSOCIATESDOCUMENT`, with the correct `RelatingX(5)` / `RelatedObjects(4)` mapping for the `IfcRelAssociates*` family. This also **fixes `IFCRELASSOCIATESMATERIAL`**, which was whitelisted but hit the default `(4,5)` mapping — `get_ref(4)` on the `RelatedObjects` *list* returns `None`, so it had silently produced zero rows. All three associations now appear in `DataModel.relationships` alongside the dedicated tables.

2. **Constituent set name (Minor).** `IfcMaterialConstituentSet.Name` is now preserved in `set_name` (was hardcoded `None`), consistent with layer/profile sets.

3. **Decoder truncation (Major).** A present-but-truncated appended section (length prefix present, but `len === 0` or overruns the buffer) now throws, instead of being silently treated as an absent old-cache section. The "absent" path is now strictly "no length prefix remains."

Tests assert the three associations land in `relationships` and the constituent set name is preserved. `cargo test -p ifc-lite-server` + `tsc` green.

PR #950, opened 2026-06-05
base class for data store

randName:
A step towards decoupling the data store to support other storage paradigms (mentioned in #916)

`IfcStoreBase` (`@ifc-lite/data`) - A new lean consumer interface with only the fields needed by non-parser packages, and four accessor methods (getEntity, getEntitiesByType, getProperties, getQuantities). The full parser store now extends the base and the four accessor methods are now implemented on the store object itself.

`EntityNode`, `EntityQuery`, and `QueryResultEntity` now import `IfcStoreBase` instead of the full parser store since they only need the base accessor methods. Headless backends (`@ifc-lite/cli`, `@ifc-lite/mcp`) also migrated.

Assisted but I had to clean up a lot of stuff.

## Summary by CodeRabbit

* **New Features**
  * Standardized data-store interface with unified entity/property/quantity access and spatial-index queries (AABB, raycast).
  * Improved attribute extraction so entity names/tags display more accurately; property values are normalized across adapters and viewer.

* **Refactor**
  * Consolidated core types and centralized store/lookup behavior for consistent runtime access.

* **Tests**
  * Added tests for entity lookup, attribute extraction, and on-demand attribute fallback.

### DIFF
--- a/apps/viewer/src/components/viewer/PropertiesPanel.tsx
+++ b/apps/viewer/src/components/viewer/PropertiesPanel.tsx
@@ -512,7 +512,7 @@ export function PropertiesPanel() {
     if (!entityNode) return [];
 
     const rawProps = entityNode.properties();
-    let result = rawProps.map(pset => ({
+    let result: DisplayPropertySet[] = rawProps.map(pset => ({
       name: pset.name,
       properties: pset.properties.map(p => ({ name: p.name, value: p.value, isMutated: false })),
       isNewPset: false,
--- a/apps/viewer/src/sdk/adapters/export-adapter.ts
+++ b/apps/viewer/src/sdk/adapters/export-adapter.ts
@@ -153,13 +153,13 @@ export function createExportAdapter(store: StoreApi): ExportBackendMethods {
     if (!model?.ifcDataStore) return [];
 
     const node = new EntityNode(model.ifcDataStore, ref.expressId);
-    return node.properties().map((pset: { name: string; globalId?: string; properties: Array<{ name: string; type: number; value: string | number | boolean | null }> }) => ({
+    return node.properties().map((pset) => ({
       name: pset.name,
       globalId: pset.globalId,
-      properties: pset.properties.map((p: { name: string; type: number; value: string | number | boolean | null }) => ({
+      properties: pset.properties.map((p) => ({
         name: p.name,
         type: p.type,
-        value: p.value,
+        value: p.value as string | number | boolean | null,
       })),
     }));
   }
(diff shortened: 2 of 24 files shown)


louistrue:
Thanks @randName — this is a solid direction. Decoupling the parser store behind a lean `IfcStoreBase` (entityCount/schema + the four `getEntity`/`getEntitiesByType`/`getProperties`/`getQuantities` accessors) is exactly the seam we need for #916, and the query-layer cleanup is a real win: `EntityNode`/`QueryResultEntity` lose their on-demand-vs-table branching and just delegate to the store. The new `BufferEntitySource` + its tests are nicely isolated too.

CI was red for one reason: the **mcp/cli headless backends** were re-typed to the lean base, but they still pass the store into full-store-only APIs (`StoreEditor`/`MutationStoreShape`, `StepExporter`/`exportToStep`, and the on-demand extractors that need `source`/`strings`/`parseTime`). Those backends are legitimately full-store consumers, so narrowing them is what `tsc` rejected. Vercel just happened to stop at `@ifc-lite/mcp` first; the same break also hit `@ifc-lite/cli` and the viewer.

I pushed a fixup commit (6fe8399) completing the migration — happy to adjust anything:

- Reverted the `mcp`/`cli` backend imports back to the full `IfcDataStore`; `IfcStoreBase` stays in the query layer where only the accessors are used.
- Gave `EntityNode.properties()/quantities()/property()` explicit return types (`PropertySet[]` / `QuantitySet[]` / `PropertyValue | null`). The `ReturnType<typeof this.store.getProperties>` form widened to `any` in declaration-emit (because `store` is private), which is what produced the `TS7006` implicit-any errors at call sites.
- Implemented the four accessors on the viewer's hand-built stores (`serverDataModel`, `nativeSpatialDataStore`) — server stores delegate properties/quantities to their pre-built tables; entity extraction returns empty since those stores have no source buffer (matches prior behavior).
- Coerced `PropertyValue` → scalar at the backend/adapter boundaries (same cast your `typeProperties` mapping already uses) and annotated the `PropertiesPanel` result as `DisplayPropertySet[]`.

Verified locally: `data`/`parser`/`query`/`mcp`/`cli`/`clash`/`viewer-core` build clean, viewer `tsc --noEmit` passes, parser 131/131 and query 109/109 tests green.

One thing worth a look (pre-existing, not touched): `desktopModelSnapshot.ts` builds a store via `as unknown as IfcDataStore`, which bypasses the new required accessors — if that store ever reaches `EntityNode`, `.getProperties()` would be undefined at runtime. Probably fine since it looks snapshot/cache-only, but worth confirming it doesn't flow into the query path.

randName:
thanks, i was focused on getting the tests to pass and didn't build (didn't install the wasm/rust stuff yet).

(personally I'm more fond of jsdoc types + [dts-buddy](https://github.com/sveltejs/dts-buddy), but that's quite a big overhaul so i'm not asking for it)

louistrue (approved):
Thx, @randName 👍

Issue #944, opened 2026-06-05
perf: raycast BVH build blocks the main thread on large models (avoidable per-vertex recompute) + async build proposal


Blogbotana:
## Summary

On large models the raycast BVH build (`packages/renderer/src/bvh.ts`,
`BVH.build`) blocks the main thread for several seconds (multi-second on
~700K-element / multi-million-triangle files). This freezes the tab the first
time anything needs the CPU raycaster — orbit-pivot-under-cursor and measure
snapping (`raycast-engine.ts` → `BVH.build`).

I'd like to make this non-blocking and are happy to contribute a PR if the
direction is acceptable.

## Root cause (looks like avoidable recomputation, not just size)

`buildNode` re-derives each mesh's centroid/bounds from its **full vertex
array**, repeatedly:

- The split sort calls `getMeshCenter(mesh)` inside the comparator:

  ```ts
  const sortedIndices = [...meshIndices].sort((a, b) => {
    const centerA = this.getMeshCenter(meshes[a])[axis]; // scans ALL verts of mesh a
    const centerB = this.getMeshCenter(meshes[b])[axis]; // scans ALL verts of mesh b
    return centerA - centerB;
  });
  ```

  So a full-mesh vertex scan happens on every comparison — O(N·logN) scans per
  level, and the same centroids are recomputed at every recursion level.

- `calculateBounds(meshes, meshIndices)` likewise re-scans every vertex of the
  whole subtree at each node.

For N meshes each with V vertices this is roughly O(N·logN · V) of redundant
vertex iteration — which matches the observed multi-second build.

## Proposed fix

**1. Precompute per-mesh AABB + centroid once (primary, big win, still sync).**
One O(total verts) pass to build a `Float32Array` of per-mesh AABBs/centroids,
then `buildNode` sorts/splits using those cached values and unions child AABBs
instead of re-scanning vertices. This removes the redundant iteration and should
turn the multi-second build into well under a second; it may make async
unnecessary for most models.

**2. Optional async build via Web Worker (for the remaining cost / true
non-blocking).** Because the builder only needs the per-mesh AABB array (small,
transferable) and produces a serializable node tree (bounds + mesh-index lists),
it can run off the main thread:

- main: compute the per-mesh AABB `Float32Array` (from step 1) and `postMessage`
  it (transferable, zero-copy),
- worker: build the node tree,
- worker → main: return the tree flattened into typed arrays (transferable),
[...]

PR #954, opened 2026-06-05
perf(renderer): cache per-mesh AABB in BVH build

Blogbotana:
## Summary

Fixes the multi-second main-thread freeze on the first CPU raycast
(orbit-pivot-under-cursor, measure snapping) on large models, reported in #944.

`BVH.build` re-derived each mesh's centroid/bounds from its **full vertex array**
repeatedly:

- `getMeshCenter` was called twice per comparison inside the sort comparator → a
  full-mesh vertex scan on every comparison,
- `calculateBounds` re-scanned the whole subtree at every `buildNode` level.

≈ `O(N·logN·V)` redundant vertex iteration.

## Change (direction 1 from the issue)

- One `O(total verts)` pass caches every mesh's AABB into a `Float32Array`
  (6 floats/mesh).
- `buildNode` sorts/splits on **cached centroids** and unions **cached child
  AABBs** instead of re-scanning vertices.
- Storing the AABB in f32 is lossless for f32 `positions`, and centroids are
  derived in f64, so the split order and node bounds are **identical** to the old
  path → `getMeshesForRay` returns the same candidate set.
- `BVH.build` stays **synchronous** with the same signature (backward compatible).

Async build (worker / prefetch) is intentionally **not** included here — deferred
per the issue discussion.

## Test

Adds `packages/renderer/src/bvh.test.ts` (`node:test`, matching the package's
existing test style) pinning the "no false negatives" invariant: `getMeshesForRay`
must be a superset of a brute-force AABB-hit reference, plus degenerate cases —
small scene, unbuilt BVH, clear-miss pruning (zero candidates), an empty-vertex
mesh mixed with real ones, and an all-empty scene.

Run: `npx tsx --test packages/renderer/src/bvh.test.ts` → 7/7 pass.

## Before / after

Real ~700K-element model (110,752 meshes / 35.9M verts):

| | BVH.build |
|---|---|
| before (vertex rescans) | 28,902 ms |
| after (cached AABB) | 548 ms |
| | **52.8× faster** |

First measure / orbit-pivot no longer freezes the tab for ~29 s.

---

_Assisted with AI; reviewed and checked by a human._

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for ray-casting behavior across multiple scene configurations and edge cases.

* **Refactor**
  * Optimized ray-casting performance through internal caching improvements.

### DIFF
--- a/packages/renderer/src/bvh.test.ts
+++ b/packages/renderer/src/bvh.test.ts
@@ -1,0 +1,213 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import { BVH } from './bvh.js';
+import type { Ray } from './raycaster.js';
+import type { MeshData } from '@ifc-lite/geometry';
+
+/**
+ * The BVH is a conservative broad-phase: `getMeshesForRay` may return extra
+ * candidates, but it must never DROP a mesh whose AABB the ray actually crosses
+ * (that would make picking/snapping miss real geometry). These tests pin that
+ * "no false negatives" invariant against a brute-force reference, plus the
+ * degenerate cases (small scene, unbuilt, empty mesh).
+ */
+
+interface AABBLike {
+  min: { x: number; y: number; z: number };
+  max: { x: number; y: number; z: number };
+}
+
+// Axis-aligned cube (8 corners) centred at (cx,cy,cz) with half-size h.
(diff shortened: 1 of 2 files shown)


Blogbotana:
@louistrue  Would you mind granting me write permissions? It would save me from having to fork the repository each time

louistrue:
@Blogbotana sent you write-access invite. One note: `main` is branch-protected, so merges still go through a PR + one approval... Thanks for the solid contributions! 🙏

Blogbotana:
> @Blogbotana sent you write-access invite. One note: `main` is branch-protected, so merges still go through a PR + one approval... Thanks for the solid contributions! 🙏

Yes, absolutely. I understand that making PRs is necessary. It's part of the process for any project. I've also found a few more bugs. I'll open issues for them soon and suggest some solutions. I'm still looking into things at the moment

PR #955, opened 2026-06-06
fix(geometry): tessellate IfcTrimmedCurve cartesian-point trims (#953)

### LINKED ISSUE #953: IfcTrimmedCurve parsing error renders curved surface as triangle
In the attached model（Roof-01_BCAD）, the profile definitions of IfcWall (#328 and #331) contain IfcTrimmedCurve (arc segments). The parser fails to correctly tessellate the arc curve, causing the intended curved surfaces to be incorrectly rendered as flat triangles. Referring to the following comparison images.

[Roof-01_BCAD.zip](https://github.com/user-attachments/files/28642820/Roof-01_BCAD.zip)

Image

Image

louistrue:
## Summary

Fixes #953 — `IfcWall` profiles whose outline is an `IfcCompositeCurve` with an `IfcTrimmedCurve` arc trimmed by **`IfcCartesianPoint`** bounds (`MasterRepresentation = .CARTESIAN.`) were rendered as flat triangles instead of the intended curved (semicircular) surfaces.

### Root cause
`profiles.rs::extract_trim_param` only understood `IfcParameterValue` trims, so for cartesian-point trims both bounds resolved to `None`. The conic then defaulted to a full `0..2π` circle which, combined with `SenseAgreement = .F.`, wrapped to a **zero-length arc** — collapsing the profile to a degenerate 3-vertex outline.

### Fix
Resolve cartesian-point trims and invert them through the circle's placement into parametric angles (`atan2(ly/r2, lx/r1)`). `MasterRepresentation` selects parameter-vs-cartesian, falling back to whichever flavour is actually authored. The `Roof-01_BCAD` semicircular profiles now reconstruct correctly (3 → 19 outer vertices).

Verified in the WebGPU viewer, the Three.js playground, and the standalone Three.js / Babylon examples (all consume the same wasm geometry). Adds a regression test + fixture (`953_trimmed_curve_cartesian_arc.ifc`, uploaded to the `fixtures-v1` release).

> Ships via the `@ifc-lite/wasm` changeset; `@ifc-lite/geometry` cascades by `updateInternalDependencies`.

## Also in this PR (found while verifying the fix)

**Three.js z-fighting** — the standalone Three.js example *and* the `create-ifc-lite` threejs scaffold template now:
- render opaque IFC meshes **`DoubleSide`** (IFC winding isn't reliably outward — the native renderer uses `cullMode: 'none'` for the same reason; culling left coincident coplanar walls z-fighting into a comb), and
- enable **`logarithmicDepthBuffer`** (near-coplanar slabs far from origin — a roof on a gable — stair-stepped without it).

**Example build hygiene** (`examples/threejs-viewer`, `examples/babylonjs-viewer`):
- bump off the stale `^1.11` `@ifc-lite/*` pin to current published ranges (`^2.1.0` / `^3.0.0` / `^2.0.0`);
- `compare.ts` no longer reaches into a monorepo-internal `../../../packages/renderer/src/…` path (broke `tsc` `rootDir` + template copy-out) — the zero-dependency `federationRegistry` is vendored locally so the "no WebGPU" example doesn't pull the WebGPU renderer;
- emit ESM workers so the multi-entry vite build doesn't trip Rollup's IIFE-worker default.

## Test
- `cargo test -p ifc-lite-geometry` (incl. new `issue_953_*` + existing `issue_820_*`) ✅
- `examples/threejs-viewer` + `examples/babylonjs-viewer`: `tsc --noEmit` + `vite build` ✅
- scaffolded `create-ifc-lite --template threejs`: `npm install` + `tsc` + `vite build` ✅

## Summary by CodeRabbit

* **New Features**
  * Federated model support: unified global IDs across linked IFC models.

* **Bug Fixes**
  * Rendered IFC geometry now uses double-sided faces to prevent visible seams.
  * Improved depth precision for large models to reduce z-fighting.
  * Fixed trimmed-curve geometry so arcs and wall profiles render correctly.

* **Chores**
  * Updated example build/runtime configs and worker output format.

* **Tests**
  * Added regression test validating trimmed-curve/cartesian arc handling.

### DIFF
--- a/.changeset/threejs-template-double-sided.md
+++ b/.changeset/threejs-template-double-sided.md
@@ -1,0 +1,5 @@
+---
+"create-ifc-lite": patch
+---
+
+three.js template: render opaque IFC meshes double-sided and enable `logarithmicDepthBuffer`. IFC triangle winding is not reliably outward (the native renderer draws with `cullMode: 'none'` for the same reason), so culling one side of two coincident coplanar walls left the survivors z-fighting into a comb along the seam; and IFC models far from the origin with stacked near-coplanar slabs (a roof on a gable wall) stair-stepped without a logarithmic depth buffer. Both are fixed in the scaffolded viewer.
--- a/.changeset/trimmed-curve-cartesian-arc.md
+++ b/.changeset/trimmed-curve-cartesian-arc.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Geometry: tessellate `IfcTrimmedCurve` arcs whose bounds are `IfcCartesianPoint`s ([#953](https://github.com/LTplus-AG/ifc-lite/issues/953)). When a profile's trimmed conic uses `MasterRepresentation = .CARTESIAN.` (the trims are points, not `IfcParameterValue`s), the bounds were dropped — the arc defaulted to a full circle that, with `SenseAgreement = .F.`, wrapped to a zero-length arc and collapsed the profile to flat triangles. The cartesian bounds are now inverted through the circle's placement into parametric angles, with `MasterRepresentation` selecting parameter-vs-cartesian and a fallback to whichever flavour is authored. Restores the semicircular wall profiles in `Roof-01_BCAD`.
--- a/examples/babylonjs-viewer/package.json
+++ b/examples/babylonjs-viewer/package.json
@@ -11,9 +11,9 @@
   },
   "dependencies": {
     "@babylonjs/core": "^9.9.1",
-    "@ifc-lite/data": "^1.11.5",
-    "@ifc-lite/geometry": "^1.11.5",
-    "@ifc-lite/parser": "^1.11.5"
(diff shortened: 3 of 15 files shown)


louistrue:
Addressed CodeRabbit's findings in `f34c5cb`:

- ✅ **`ifc-to-threejs.ts` `geometryResultToBatched`** — switched opaque to `DoubleSide` (this batched path was missed in the first pass, so it would have re-introduced the coincident-wall z-fight comb).
- ✅ **`create-ifc-lite` threejs template** — the scaffolded `vite.config.ts` now sets `worker.format: 'es'`, matching the repo's example/app configs (verified by scaffolding + `vite build`).
- ✅ **`compare.ts` `normalizeValue`** — the empty `catch` now logs the `JSON.stringify` failure (repo coding guideline).

Deferred (out of scope for this fix):
- **`main.ts` / `compare.ts` "file too large" refactors** — both are pre-existing example files unchanged in spirit by this PR; splitting them is unrelated to the #953 curve fix and best done separately.

Issue #945, opened 2026-06-05
Structural grids (IfcGridAxis) can't be aligned with streamed geometry — mesh re-basing transform not exposed


Blogbotana:
## Summary

Structural grids (`IfcGridAxis`) can't be aligned with the streamed geometry: the
grid/symbolic data comes back in true IFC-world coordinates, but `@ifc-lite`
renders meshes in a re-based "building-local" frame whose transform isn't exposed,
so the axes land far off the model. This affects any viewer (the reference viewer
misplaces grids too).

We'd like to render structural grids overlaid on the model and are happy to help
with a PR if you can point us at the intended approach.

## What we see (real model, ~700K elements)

- Rendered mesh bounds (what `getModelBounds()` / `coordinateInfo` report), metres,
  near origin:
  `min {-222.8, -11.8, -88.6}  max {184.1, 23.4, 134.3}`
  `coordinateInfo`: `originShift {0,0,0}`, `hasLargeCoordinates: false`,
  `buildingRotation: 0.4867`.
- A grid axis `IfcPolyline` point read from the store is `[294091.5, 0, 0]` — i.e.
  the file is in **millimetres**.
- Resolving the `IfcGrid.ObjectPlacement` chain (Location/Axis/RefDirection,
  `PlacementRelTo` up to the site) and transforming that point gives, in mm:
  `[-9_995_836, 167_150, -2_018_218]` → in metres `(-9996, 167.15, -2018)`.

So the IFC places the grid at **survey coordinates** (X ≈ −10 km, Z elevation
≈ 167 m), while the rendered meshes sit at building-local metres near the origin.
`parseSymbolicRepresentations` returns grid axes in yet another partial frame
(elevation ≈ 167 m, not placement-applied) — also off.

## The gap

The mesh pipeline clearly applies **unit scaling + some level of the placement
chain (dropping the site survey offset/elevation) + the building rotation**, but
`coordinateInfo` only exposes `buildingRotation` — not the translation, the unit
scale, or which placement level was applied. So a viewer can resolve the grid
placement correctly (true IFC-world) yet still can't map it into the frame the
engine actually rendered the meshes in.

For comparison, in an xbim-based pipeline both geometry and grids live in IFC-world
(`IfcGrid.ObjectPlacement.ToMatrix3D()` applied to the axis curve), so grids line
up. `@ifc-lite` re-bases the meshes, which breaks that.

## Possible directions

[...]

PR #966, opened 2026-06-06
feat(grids): expose structural grids (IfcGridAxis) in the render frame (#945)

Blogbotana:
## Summary

Implements the preferred **direction (1)** from #945: resolve `IfcGridAxis`
curves through the **same** placement + unit-scale + RTC pipeline the streamed
meshes use, and emit them in the renderer's **Y-up, RTC-subtracted, metres**
world frame — so structural grids overlay the model *by construction*, with no
viewer re-implementing the placement resolver. Mirrors the existing
`parseAlignmentLines` stack.

It also closes the **completeness gap** (direction 2): the world→render
translation is now surfaced on `coordinateInfo`.

## What's included

**New WASM API** — `rust/wasm-bindings/src/api/grid_lines.rs`
- `parseGridLines(content) -> Float32Array` — flat 3D line-list `[x0,y0,z0, …]`, one segment per axis.
- `parseGridAxes(content) -> GridAxisCollection` — structured per-axis `{ gridId, axisId, tag, start, end }`.
- Reuses `resolve_scaled_placement` + `detect_rtc_offset_from_first_element` (the same offset the mesh path applies, >10 km gated), then the IFC Z-up → WebGL Y-up swap from `MeshDataJs::new`.

**`@ifc-lite/geometry`**
- `GeometryProcessor.parseGridLines` / `parseGridAxes` (returns plain `GridAxis[]`) + a `GridAxis` type.
- `CoordinateInfo` now reports `lengthUnitScale` and **populates `wasmRtcOffset`** (the actually-applied RTC offset) directly from the geometry pipeline. Previously this was `null`/`undefined` even when a ~10 km shift had been applied — exactly what blocked external viewers in the original report.

**Tests**: 4 Rust unit tests (local + georeferenced grid, line-list shape, tag preservation); 3 TS tests for the new `coordinateInfo` fields.

## Verification

Built the WASM (LLVM-20 / generic libc++) and tested end-to-end in a downstream Svelte viewer (the reference viewer from the issue). With `wasmRtcOffset` + `lengthUnitScale` now populated, that viewer's placement-resolving grid overlay lines up with the streamed meshes on a **mm-unit, ~10 km survey-offset** model. (Its grid button had been parked specifically because of this issue — re-enabling it was the test.)

## Question for maintainer 🙋

I scoped this PR to the **SDK / WASM surface** (`@ifc-lite/wasm` + `@ifc-lite/geometry`), since my use case renders grids in an external viewer. I deliberately **did not** touch `apps/viewer` or `@ifc-lite/renderer`.

@louistrue **Would you like grid rendering wired into the in-repo `apps/viewer` too?** e.g. a `useGridLines3D` hook (mirroring `useAlignmentLines3D`) + a dedicated `uploadGridLines3D` line pipeline + a visibility toggle. Happy to add it here or as a follow-up — just say which direction you'd prefer (and whether you'd want 3D tag/bubble labels, or lines-only to start).

Two smaller calls I'd value your steer on:
1. `parseGridAxes` currently samples the first/last point of each axis curve (matching the 2D symbolic path); arcs / trimmed grid axes come out as a single chord. Want full polyline sampling?
2. `wasmRtcOffset` is populated **only when a shift was applied**, preserving the `!== undefined` "geometry re-based" meaning for federation/cache. OK to keep that semantics?

---
Assisted with AI, checked and verified by human.

## Summary by CodeRabbit

* **New Features**
  * Structural grid support: render frame now displays grid lines and axes with automatic coordinate transformation and unit scaling
  * New parsing APIs to extract grid line geometry (flat vertex list) and structured grid axis data
  * Public geometry API exposes GridAxis data and new methods to parse grid lines/axes
  * Coordinate metadata now includes file-unit scale and WASM-derived RTC offset for accurate external geometry mapping

* **Tests**
  * Added tests covering coordinate metadata exposure and reset behavior

### DIFF
--- a/.changeset/structural-grids-render-frame.md
+++ b/.changeset/structural-grids-render-frame.md
@@ -1,0 +1,21 @@
+---
+"@ifc-lite/wasm": minor
+"@ifc-lite/geometry": minor
+---
+
+feat(grids): expose structural grids (IfcGrid/IfcGridAxis) in the render frame (#945)
+
+Resolve `IfcGridAxis` curves through the same placement + unit-scale + RTC
+pipeline the meshes use and emit them in the renderer's Y-up, RTC-subtracted,
+metres world frame, so structural grids overlay streamed geometry by
+construction (no viewer re-implements the placement resolver).
+
+- New WASM API `parseGridLines(content) -> Float32Array` (flat 3D line-list)
+  and `parseGridAxes(content) -> GridAxisCollection` (structured per-axis
+  `{ gridId, axisId, tag, start, end }`), mirroring `parseAlignmentLines`.
+- New `@ifc-lite/geometry` `GeometryProcessor.parseGridLines` /
+  `parseGridAxes` (returns plain `GridAxis[]`) and a `GridAxis` type.
+- `CoordinateInfo` now also reports `lengthUnitScale` and populates
+  `wasmRtcOffset` (the actually-applied RTC offset) directly from the geometry
+  pipeline, so any consumer can map externally-resolved geometry into the
+  render frame without viewer-side patching.
--- a/packages/geometry/src/coordinate-handler.ts
+++ b/packages/geometry/src/coordinate-handler.ts
@@ -35,6 +35,13 @@ export interface CoordinateInfo {
     wasmRtcOffset?: Vec3;
     /** Building rotation angle in radians (from IfcSite placement). Rotation of building's principal axes relative to world X/Y/Z. */
     buildingRotation?: number;
+    /**
+     * Length-unit scale (file units → metres) resolved from IfcProject's unit
+     * assignment, e.g. `0.001` for millimetre files. Lets a consumer transform
+     * externally-resolved geometry (grids, survey points) into the render frame
+     * without re-parsing units. See issue #945.
+     */
+    lengthUnitScale?: number;
 }
 
 export class CoordinateHandler {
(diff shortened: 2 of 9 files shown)


Blogbotana:
The result!

louistrue (approved):

louistrue:
Thanks @Blogbotana! Clean, well-scoped PR, and reusing the mesh placement + RTC pipeline (rather than re-deriving placement) so grids align by construction is exactly the right call. Verified the green CI and your downstream screenshot. Answers to your questions:

**1. apps/viewer wiring**  let's do it as a **follow-up PR, lines-only to start.** The SDK/WASM surface here is the reusable part and I'd rather land it now. The in-repo `useGridLines3D` hook (mirroring `useAlignmentLines3D`) + `uploadGridLines3D` pipeline + visibility toggle is a clean next step; 3D tag/bubble labels can come after lines. Tracked in #967 — very happy for you to take it if you'd like, otherwise I'll pick it up.

**2. Polyline sampling**  first/last chord is fine to ship. Straight axes (the common case) are exact; curved grid axes are rare. We can add full sampling later and reuse the `IfcTrimmedCurve` tessellation from #953 when a model needs it.

**3. `wasmRtcOffset` semantics** — keep them. The `!== undefined` "geometry was re-based" signal matters for federation/cache; always emitting `[0,0,0]` would lose the not-rebased vs rebased-to-origin distinction.

Merging, thanks again!

Issue #853, opened 2026-05-28
IfcSlab opening renders as through-hole instead of pocket


xyzbety:
In the attached model（slab-openings）, the boolean subtraction for the IfcSlab opening is incorrect; it renders as a through-hole instead of the expected partial pocket, suggesting the geometry depth or position of the IfcOpeningElement is not being respected. Referring to the following comparison images.

[slab-openings.zip](https://github.com/user-attachments/files/28333352/slab-openings.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/c41e100b-85a5-4df8-8158-aa902813fde6" />

<img width="2880" height="1722" alt="Image" src="https://github.com/user-attachments/assets/866effef-0346-43c0-9aea-c3ea15a4d1f7" />

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/cc77ad2e-9312-4bb7-bbfa-a932da58f10a" />

Issue #832, opened 2026-05-27
Opening representation inconsistency - wall cuts through entire section in some cases


kamyarba:
I've created 5 walls with different representation formats for openings, and encountered an issue where, in one case, the opening cuts through the entire section of the wall instead of being properly bounded.
### Expected Behavior

<img width="2533" height="731" alt="Image" src="https://github.com/user-attachments/assets/b87c71e5-c9ea-45e2-8e71-080c91bd2c24" />

All openings should be properly represented within the wall section, regardless of the representation format used.

### Actual Behavior

One of the wall representations with a different opening format causes the opening to cut completely through the wall section, rather than creating a bounded opening.

<img width="2766" height="736" alt="Image" src="https://github.com/user-attachments/assets/5ae4ebf8-d41b-4518-9dbc-aab2dc8df0b4" />

This appears to be related to how different opening representation formats are being processed/rendered.  
This may be related to [#821](https://github.com/LTplus-AG/ifc-lite/issues/821) - "Outside wall geometry is not displayed", which also involves issues with wall geometry rendering and representation formats.

### Attachments 
[ifc-opening.ifc.txt](https://github.com/user-attachments/files/28302665/ifc-opening.ifc.txt)

Issue #635, opened 2026-05-11
IfcBooleanClippingResult not applied — walls and gable ends show uncut geometry


CRiMSONxx:
<img width="244" height="108" alt="Image" src="https://github.com/user-attachments/assets/86dd0989-49a0-4f16-8fec-8c6d7ce453dc" />

## Description

Walls and gable end walls are displayed without their boolean cuts applied.
The full uncut solid is rendered instead of the correctly clipped geometry.

This affects:
- Gable/triangular end walls cut by roof geometry
- Wall openings (doors, windows) via IfcRelVoidsElement
- Any geometry using IfcBooleanClippingResult

## Expected vs Actual

**Expected:** Wall rendered with roof cut applied (triangular gable shape)
**Actual:** Full rectangular wall solid rendered, ignoring the clipping result

The same IFC file displays correctly in BIMvision and Solibri.

## IFC Entity Involved

`IfcBooleanClippingResult` — a subtype of `IfcBooleanResult` where the operator
is DIFFERENCE and the second operand is a half-space solid (IfcHalfSpaceSolid).

In IFC, walls with angled cuts store the geometry as:
- FirstOperand: the full wall solid (IfcExtrudedAreaSolid)
- Operator: DIFFERENCE
- SecondOperand: IfcHalfSpaceSolid (the cutting plane)

It appears the geometry pipeline returns FirstOperand only, without applying
the boolean difference.

## Environment

- IFCLite version: latest
- OS: Windows
- IFC schema: IFC4

## Steps to Reproduce

1. Load an IFC file containing a gable/pitched roof wall
2. The wall displays as a full rectangular box instead of the correct clipped shape

PR #970, opened 2026-06-06
fix(geometry): preserve profile voids re-encoded as opening elements (#964)

### LINKED ISSUE #964: IfcSlab void geometry incorrect: complex shapes render as rectangles
In the attached model（IfcSlab Extruded）, entity IfcSlab (3qy29DaXf0ivOFkuocsP7z) uses IfcArbitraryProfileDefWithVoids and IfcAreaSolid for its geometry definition. The parser fails to correctly process the boundary curve definitions of the voids, causing elliptical and hexagonal voids to be incorrectly rendered as rectangular openings. Referring to the following comparison images.

[IfcSlab Extruded.zip](https://github.com/user-attachments/files/28665926/IfcSlab.Extruded.zip)

Image

Image

louistrue:
Fixes #964 — complex `IfcSlab` voids (ellipse, hexagon) rendered as rectangles.

## Root cause
The reporter's slab **double-encodes its voids** (a Revit/ODA export pattern): once baked into the body's `IfcArbitraryProfileDefWithVoids` swept profile (which tessellates the correct round/polygonal holes via earcut), and again as redundant `IfcOpeningElement` + `IfcRelVoidsElement`.

The viewer void path (`process_element_with_voids` → `apply_void_context`) builds the correct body, then runs the redundant opening boolean cuts on top. Since the host is *already void* at those locations, CSG removes nothing (Manifold returns the host unchanged / errors `NotManifold`). The code read that as "CSG failed" and applied the **AABB bounding-box fallback** (built for #635 round windows) — carving a rectangle over each correct ellipse/hex hole.

Verified precisely: slab body bottom-cap solid area = `7,464,903`; with the buggy opening pass = `7,235,744` (**exact AABB-bbox match**).

## Fix
Before the AABB fallback fires, cast a ray through the opening centroid along its extrusion axis. **Zero** host-triangle hits means the host is already open there → suppress the destructive box cut. This positive void detection overrides the `capped`/engulf heuristics.

A genuinely solid host (e.g. #635's round window in an un-voided wall) sandwiches the centroid between two faces → the fallback still fires, so **no regression**.

Post-fix area = `7,464,906` ≈ the body's `7,464,903` — round/hex holes preserved.

## Tests
- New `issue_964_arbitrary_profile_voids.rs` drives the production void path and asserts the opening pass doesn't change the slab's solid cross-section area. Confirmed it **fails on old code** (3.07% area loss) and **passes with the fix**.
- Fixture added to `manifest.json` and uploaded to the `fixtures-v1` release (fetch + hash verified).
- **All 311 geometry tests pass** (incl. #635, #832, #853). `cargo check --workspace` clean; clippy at baseline (no new warnings).

## Debugging note
The slab body is a valid closed, consistently-wound, genus-3 manifold, yet `manifold3d::from_mesh_f64` rejects it `NotManifold` — because the mesh fed to the ellipse/hex CSG is the *post-rectangular-clip* host (the rect-opening + reveal path introduces bad half-edges). So a `NotManifold` CSG failure here does not mean the original body is broken.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redundant-opening detection to avoid incorrect simplification of arbitrarily-shaped holes; AABB fallback suppression now skips unnecessary rectangularization and diagnostic logging only appears when fallback runs.

* **Tests**
  * Added a regression test and fixture validating redundant-opening handling for complex slab void geometries.

### DIFF
--- a/rust/geometry/src/router/voids.rs
+++ b/rust/geometry/src/router/voids.rs
@@ -498,29 +498,9 @@ fn ray_triangle_param(
     Some(e2.dot(&qvec) * inv_det)
 }
 
-/// Whether `mesh` is already hollow at `point` along `axis` — i.e. the
-/// infinite line through `point` in the `axis` direction does not cross a
-/// single host triangle.
-///
-/// Issue #964: some exporters (Revit) double-encode a void — once baked into
-/// the host's `IfcArbitraryProfileDefWithVoids` profile and again as a
-/// redundant `IfcOpeningElement`. The body geometry already carries the
-/// (correct, possibly round/polygonal) hole, so when the redundant opening's
-/// CSG subtraction finds nothing to remove the AABB fallback must NOT fire:
-/// cutting the opening's bounding box would carve a rectangle over the
-/// already-correct hole. A ray cast through the opening centroid along its
-/// extrusion axis hits zero host triangles exactly when the hole is already
-/// open there.
-///
-/// Deliberately strict: returns `true` only when there is NO host material on
-/// the line (both sides clear). A genuinely solid host — e.g. the issue #635
-/// round window in an un-voided wall — sandwiches `point` between two faces,
-/// so the fallback still fires there. Any ambiguity (material on one side)
-/// keeps the existing cut behaviour.
-fn host_already_open_along_axis(mesh: &Mesh, point: Point3<f64>, axis: &Vector3<f64>) -> bool {
-    let Some(axis) = axis.try_normalize(NORMALIZE_EPSILON) else {
-        return false;
-    };
+/// Whether the infinite line through `point` along `axis` crosses any
+/// triangle of `mesh`.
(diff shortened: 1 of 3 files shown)


louistrue:
Thanks for the reviews — addressed both in c77a8003:

- **@chatgpt-codex-connector (P2 — partial-overlap cutter):** good catch. Replaced the centroid-only ray with a footprint-spanning probe (`opening_redundant_with_host`): it samples the centroid **plus every cutter vertex pulled 10% toward the centroid** (kept strictly inside the real round/poly footprint, not its bbox). The opening is treated as redundant only when a ray along the extrusion axis through **every** sample hits zero host triangles. So your example — a circular opening centred inside an already-cut rectangle but spilling into solid beyond it — now has its protruding samples hit host material → `redundant=false` → the fallback still fires. The genuinely-solid #635 case is rejected at the first sample.

- **@coderabbitai (skip message):** the fixture-missing skip now points to `pnpm fixtures`.

Full geometry suite still 311/0; clippy at baseline.

PR #972, opened 2026-06-06
fix(geometry): skip smooth-curve simplification of thin curved profiles (#820)

### LINKED ISSUE #820: IFCTRIMMEDCURVE with offset circle center renders incorrectly
In the attached model（RadianValuesOverPI）, the geometry definition of the wall (IFCWALLSTANDARDCASE) uses a circular clipping curve with positional offset. ifc-lite failed to correctly process the offset of the circle center, resulting in incorrect geometry display, referring to the following comparison images.

[RadianValuesOverPI.zip](https://github.com/user-attachments/files/28293358/RadianValuesOverPI.zip)

Image

Image

louistrue:
## Issue

Closes #820. The reporter saw the `RadianValuesOverPI` curved wall render with a **folded flap at the closing seam** ("reappeared" after the earlier plane-angle fix #834).

## Root cause

This was a *second* defect exposed once #834 let the wall render at full size. The wall is a **thin (100 mm) annular sector** — inner arc r=11900, outer arc r=12000, both centred at (0,−12000), swept ~240°, joined by two short radial caps.

The profile flows through `simplify_smooth_curve_polyline` (the issue-#635 RDP downsampler), whose chord-deviation epsilon is scaled to the **bounding-box diagonal** (≈327 mm here) — over **3× the wall thickness**. RDP slides chords clean across the thin dimension, so the inner and outer arcs drift independently:

- it drops the seam vertices where the boundary doubles back → the loop becomes **self-intersecting** (11 crossings, footprint area 4.08e6 vs the true 5.006e6 mm²) → earcut renders the folded flap; **or**
- on thicker/shorter curved walls it silently **gains/loses double-digit-percent area** while staying topologically simple — the same distortion class as the documented +4.31% W410 fillet bug.

## Fix

A **thin-feature gate**: a profile whose half-thickness (`area / perimeter`, its hydraulic radius) is below `THIN_FEATURE_RATIO` (0.05) of its bbox diagonal is a thin curved sliver that RDP cannot simplify without distorting its feature size — so it's left untouched (the faithful original is kept).

- Cleanly separates the cases: ratio ≈ **0.0015** for the 100 mm × 24 m #820 wall vs ≈ **0.25** for a round window.
- Fat shapes (round windows) still simplify → the **#635 void-cut polygon budget is preserved**.
- A **self-intersection backstop** (`closed_loop_self_intersects`, O(n²) proper-crossing scan) plus an **oversized-loop fail-safe** cover anything that slips through.

## Verification

- Real #820 fixture through the full pipeline: simple annular sector, bottom-cap `|signed|/abs = 1.0000` (no overlapping flap triangles), area 4.999e6 ≈ true 5.006e6.
- New fixture-free unit tests: 4 thin-sector cases stay simple + within **2% area (was ±13%)**; a 127-gon disk **still simplifies** (locks in #635); a bowtie tests the self-intersection scanner.
- Full geometry suite (**314 tests**) passes, including #635 void-cutting, #655 W410, and round-window regressions.
- The viewer's `processGeometryBatch` shares `profiles.rs`, so it gets the fix automatically (wasm crate builds clean).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved closed-polyline downsampling to avoid degrading thin curved slivers by detecting narrow features and skipping simplification when needed; added a post-simplification verification to reject results that produce self-intersections, returning the original geometry instead.

* **Tests**
  * Expanded unit tests covering intersection detection and edge cases: thin annular sectors, self-intersecting loops, and large disks to ensure simplification retains validity and reduces vertices where safe.

### DIFF
--- a/rust/geometry/src/profiles.rs
+++ b/rust/geometry/src/profiles.rs
@@ -58,6 +58,14 @@ const RDP_EPSILON_RATIO: f64 = 1.0 / 100.0;
 /// Prevents collapsing tiny profiles where ratio-derived epsilon would be
 /// numerically negligible.
 const RDP_EPSILON_MIN: f64 = 5.0e-3;
+/// Minimum ratio of a profile's half-thickness (`area / perimeter`, its
+/// hydraulic radius) to its bounding-box diagonal for simplification to be
+/// attempted. Below this the profile is a *thin* curved sliver — e.g. a
+/// trimmed-circle wall whose inner and outer arcs sit only a wall-thickness
+/// apart (issue #820) — and RDP's diagonal-scaled epsilon would distort or fold
+/// its feature size; such profiles are left untouched. A round window
+/// (half-thickness ≈ radius/2) sits far above this (~0.25) and still simplifies.
+const THIN_FEATURE_RATIO: f64 = 0.05;
 /// Only attempt simplification when the polyline has at least this many
 /// vertices. Below this the polyline is already cheap enough.
 const SMOOTH_CURVE_MIN_VERTICES: usize = 24;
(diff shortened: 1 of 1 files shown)


louistrue:
Thanks for the reviews — addressed in `77d8e768` + `b5100c5d`.

**Codex P1 (high-aspect-ratio elliptical opening wrongly thin-gated → AABB fallback):** real bug, fixed. The thinness metric `(area/perimeter)/diag` flags both thin-walled rings *and* elongated filled ellipses. The gate now requires a **second condition** — the loop must also *double back* on itself — so convex ellipses/disks keep simplifying at any aspect ratio while thin annular-sector walls are gated.

I then hardened that second condition after an internal adversarial pass found two ways a single localized defect could still false-trip an absolute-turning measure:
- a lone inward **notch** on a thin convex ellipse (an 8:1 + 5% notch reached ~1.92× turning), and
- the sub-mm out-and-back **seam jog** where the composite curve closes.

So `loop_doubles_back` now measures the **fraction of reflex vertices** (turning against the loop's winding) rather than total turning: a convex shape has zero at any aspect ratio, an annular sector's whole inner arc is reflex (~half), and a 1–2-vertex notch/jog can't tip the 0.15 threshold. New tests cover a notched ellipse and a seam-jogged disk (not gated) vs the #820 sector (gated).

**Note on `probe_ellipse_aspect_gate_TEMP`:** that test isn't in this PR's diff — it was Codex's own scratch probe; nothing to remove here.

**CodeRabbit:** the redundant `j == i` in the self-intersection scanner is removed.

@codex review

louistrue:
Good catch on the sampling-density weighting — fixed in `0e2aa8e0`.

The doubling-back test now measures the reflex share by **perimeter arc length** rather than vertex count. A thin sector's inner and outer arcs are nearly equal in *length* regardless of how many points sample each, so it reads ~0.5 independent of tessellation density (your exact 96-outer/16-inner 90° example now reports ~0.5 and is gated end-to-end), while a lone notch or closing-seam jog still spans negligible length. Added a regression test using that 96/16 sector.

@codex review

Issue #924, opened 2026-06-04
Expose model comparison in the web UI


louistrue:
Part of the #917 feedback epic.

### Current state
- **CLI** `diff` (`packages/cli/src/commands/diff.ts`) compares two files by type-count deltas and added/removed/common **by GlobalId** — no per-element "changed" detection.
- **A full comparison engine already exists** on branch `claude/sharp-ritchie-eFTTD` (not yet on `main`, no PR): the **`@ifc-lite/diff`** package (`diffModels`) classifies entities as **added / modified / deleted / unchanged** with separable **data vs geometry** scope. Data changes use `buildDataFingerprint` + `stableHash`; geometry changes use **RTC-invariant per-entity geometry hashes** from the WASM mesh pass (`MeshCollection.geometryHashValues`). Store-agnostic: adapters extract `EntityFingerprint`s and feed `diffModels`.
- **No web UI** exposes any of this (only a `federation-compare.ts` script template).

### Gap
Users cannot, in the browser: load two IFC files, pick version A and version B, run a comparison, and see added/removed/**changed** with clear color conventions.

### Approach (build the web UI on top of `@ifc-lite/diff`)
1. Land / rebase the `claude/sharp-ritchie-eFTTD` work (`@ifc-lite/diff` + geometry-hash exposure).
2. Add a **viewer adapter** that builds `EntityFingerprint`s from the loaded `IfcDataStore` + geometry hashes and calls `diffModels`.
3. **Compare UI**: select two loaded models as A/B, run, and visualize results — color conventions added = green, removed = red, changed = yellow (reuse the renderer highlight channels / overlay coloring) — plus a changes list/panel grouped by add/remove/change, with separable data-vs-geometry toggles.

### Acceptance criteria
- Load two IFC files in the viewer, select A and B, run comparison.
- Added / removed / changed elements are color-coded in 3D and listed in a panel.
- Data-only vs geometry-only change scope can be toggled.

### Dependencies
Core engine already exists on `claude/sharp-ritchie-eFTTD`; this issue is primarily the **web-UI exposure** + viewer adapter. Independent of the Lists/Filter track.

louistrue:
<img width="1051" height="719" alt="Image" src="https://github.com/user-attachments/assets/96894e52-2c8c-4016-8ca8-a9fe3d05f26f" />

Issue #661, opened 2026-05-13
Original project not visible when adding second project to the scene


semhustej:
Steps to reproduce:

- download 2 sample files

[non-referenced-projects.zip](https://github.com/user-attachments/files/27700487/non-referenced-projects.zip)

- open Project-01.ifc
- use `Add Model to Scene` and select Project-03.ifc
- after Project-03 is loaded , objects from Project-01 are not visible in the viewport
- also note that in the spatial decomposition, the same object (slab) is placed in both projects and the Wall object is not in any of those two

See the behavior in the video:

https://github.com/user-attachments/assets/5b713816-cb9f-43c6-85d9-c5e543ecc06c

IfcLite version: v1.21.0 · May 13, 2026

semhustej:
I forgot to mention: 

- this functionality worked previously (2 months ago) on Edge (chromium based)
- I had issues (different than OP) with this functionality even 2 months ago in Firefox

louistrue:
<img width="1007" height="867" alt="Image" src="https://github.com/user-attachments/assets/ca29511f-06d2-4e70-95b3-04e6d3200661" /> on it...

semhustej:
I can confirm the issue is fixed. Thank you!

PR #975, opened 2026-06-06
refactor(viewer): one canonical model-load path (loadFile + finalizeModel)

louistrue:
Unifies the viewer's model loading onto a **single canonical pipeline**, removing the parallel/divergent paths that have repeatedly drifted apart.

## Why
The viewer had (at least) three mesh-load pipelines that drifted:
- `useIfcLoader.loadFile` — primary/single-model (native-huge + WASM branches), already dispatching every format through one `finalizePrimaryModel`.
- `useIfcFederation.addModel` → `viewerModelIngest.parseStepBufferViewerModel` — a **separate** STEP pipeline for federated (2nd+) models, with its own watchdog + abort helpers.

The drift was not theoretical: a model-diff geometry hash was added to `loadFile` only, so federated models silently lacked it and the compare UI flagged **every** element as changed. The fix is structural — one load path, so any load-time behaviour (hashing, caching, RTC, progress, abort, alignment) is configured **once**.

## What
- **`loadFile(file, target)`** is now the one pipeline for every model and format, where `target` is `{kind:'primary'}` (default, unchanged behaviour) or `{kind:'federated', modelId, sharedRtcOffset, …}`.
- **One `finalizeModel`** replaces the `finalizePrimaryModel` closure + `addModel`'s inline finalize. Primary stays byte-identical (`updateModel`); federated runs the georef-align → id-offset → pointcloud-relabel → spatial-index → `addModel` sequence lifted verbatim from the old path (same order). Federated paints **at completion** and never touches the active model's slots; it's WASM-only (no cache/server), matching the former path; `loadFile` awaits the federated finalize so the model is registered before it resolves.
- **`addModel` is now a ~70-line thin wrapper** (was ~380): memory load-gate + shared-RTC pick, then `loadFile(file, {kind:'federated', …})`.
- **Deleted** `parseStepBufferViewerModel` + `Step*` types, `watchedGeometryStream`, `resolveDataStoreOrAbort` (+ their tests), the dead `NATIVE_RENDERER_PATH_ENABLED` scaffold, and the legacy→Map migration (audited: no path leaves top-level data unregistered, since `loadFile` always `upsertModel`s first).
- Extracted the georef/align cluster to a shared `ingest/federationAlign.ts` (pure move; also de-dups `realignFederation`).
- **AGENTS.md** trimmed to the minimal non-obvious gotchas + a "one canonical load path" guardrail.

## Verification
- Typecheck clean; `vite build` green; viewer/store/ingest unit tests pass.
- **Adversarial multi-agent review** (6 dimensions → verify): it caught a real **HIGH** regression — the WASM metadata callbacks (`onPartial/onFullDataStore`) wrote `setIfcDataStore` unguarded, so a federated add clobbered model #1's store (#661/#792 aliasing). **Fixed** (gated on primary; `resolveDataStore` kept unconditional). Also fixed: federated `collapsed` default + placeholder, federated huge-native routing, orphaned-promise rejection, dead export.
- **Recommended manual check before merge:** load `tests/models/various/02_BIMcollab_Example_STR_random_C_ebkp.ifc` then `~/Development/BIM-Testfiles/02_BIMcollab_Example_STR_random_C_ebkp.ifc` (federation) — both should render + align identically through the one path. (Doing a browser smoke test next.)

## Note for the diff/compare work
This is the **foundation** for #924/#939: once it lands, the compare branch rebases onto it and enables geometry hashing at the single `loadFile`/conversion chokepoint — which structurally fixes the "all geometry changed" bug (no second pipeline to forget).

## Summary by CodeRabbit

* **New Features**
  * Federated model loads now align automatically across geographic coordinate systems (affine same-CRS or reprojection) for seamless multi-model viewing.
* **Improvements**
  * Model loading path unified so federated and primary loads register and finalize more reliably, avoiding partial-state visibility during federation assembly.
* **Removals**
  * STEP/buffer parsing removed from the ingestion pipeline; prefer IFCX or GLB.
* **Documentation**
  * Contributor/agent guidelines reorganized and expanded with clarified guardrails and workflow expectations.


louistrue:
Addressed all three review findings in 803cd4fc:

- **Concurrent federated loads (Codex P1):** only a primary load bumps `loadSessionRef`; federated adds now capture the current session without invalidating in-flight siblings (a later primary load still aborts them).
- **Federated parse-failure swallowed (Codex P2):** the finalize `catch` ran `updateModel(modelId)`, which no-ops for a federated add (no placeholder until `addModel`). Federated failures now surface via toast.
- **Federated spatial index via `setIfcDataStore` (CodeRabbit):** the shared `buildSpatialIndexGuarded` guards on / writes the *active* store, so it discarded the index for a (non-active) federated model. New `buildSpatialIndexForModel` guards on the target model and publishes through `updateModel(modelId, …)`, called after `addModel`.

Typecheck + viewer store/util tests green.

PR #980, opened 2026-06-07
fix(viewer): restore entity accessors on cache-loaded models

louistrue:
## Problem
Opening the **Properties panel** for a model loaded **from the on-disk cache** crashed:

```
Uncaught TypeError: this.store.getEntity is not a function
    at EntityNode.allAttributes (entity-node.ts:118)
    at PropertiesPanel.tsx:563
```

Reproduces on `main` (cached models only): load a ≥10 MB IFC, reload, load it again (cache hit), select an element.

## Root cause
An `IfcDataStore` carries four **lazy accessors** — `getEntity` / `getEntitiesByType` / `getProperties` / `getQuantities` — that read entities/properties on-demand from the `source` buffer. These were wired **inline** in the parser's store-construction paths (the worker→main transport and the fresh columnar parse), but the viewer's cache-restore path (`useIfcCache.loadFromCache`) rebuilds the store's *data* (`source`, `entityIndex`, on-demand maps, spatial hierarchy) **without** them. The `result.dataStore as any` cast hid the missing methods until the query layer called `store.getEntity()` at runtime.

## Fix
- New shared helper **`attachDataStoreAccessors(store)`** in `@ifc-lite/parser` — the single home for wiring the four accessors, so the three store-construction paths can't drift.
- Refactored the worker→main **transport** path to use it (so the canonical reconstruction path exercises the exact helper the cache path now relies on).
- The viewer **cache-restore** path calls it once `source` + `entityIndex` + on-demand maps are finalized.

No `as any` added; one intentional `as IfcDataStore` at the cache call site (the cache store is already `any` upstream — the `schema`/`schemaVersion` fragmentation in #950/#952 is out of scope here).

## Verification
- Parser suite **135/135** (incl. a new regression test that strips a parsed store's accessors — the cache shape — then asserts `attachDataStoreAccessors` restores `getEntity`/`getEntitiesByType`/`getProperties`).
- Browser: a cache-`hit` load now exposes all four accessors; `getEntity(id)` returns real entities; the Properties panel no longer crashes.

## ⚠️ Separate pre-existing bug (NOT addressed here)
While verifying, I confirmed a **second, unrelated** crash on the cache-load path: a WASM heap corruption (`dlmalloc` panic `psize >= size + min_overhead`) → flood of `memory access out of bounds`, originating in the `Symbolic*` (2D symbolic / structural-grid) parsing. It **reproduces on a freshly built `origin/main` WASM** (so it's real on main, not this branch), and is a Rust/WASM memory-safety issue distinct from this JS-accessor fix. Filing/fixing it separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that occurred when accessing entities restored from the on-disk cache, ensuring cached entity data is properly recovered and accessible without errors.

### DIFF
--- a/.changeset/cached-store-accessors.md
+++ b/.changeset/cached-store-accessors.md
@@ -1,0 +1,7 @@
+---
+"@ifc-lite/parser": minor
+---
+
+Add `attachDataStoreAccessors(store)`, the single home for wiring an `IfcDataStore`'s lazy `getEntity` / `getEntitiesByType` / `getProperties` / `getQuantities` accessors. The fresh-parse worker→main transport path now uses it instead of duplicating the wiring inline.
+
+This fixes a crash when querying a model loaded from the on-disk cache: the cache format only serialises data, so a restored store was missing these accessor methods, and opening the Properties panel for a cached entity threw `store.getEntity is not a function` (the viewer's cache-restore path now calls `attachDataStoreAccessors`).
--- a/apps/viewer/src/hooks/useIfcCache.ts
+++ b/apps/viewer/src/hooks/useIfcCache.ts
@@ -109,7 +109,8 @@ export function useIfcCache() {
   const loadFromCache = useCallback(async (
     cacheResult: CacheResult,
     fileName: string,
-    cacheKey?: string
+    cacheKey?: string,
+    fallbackSourceBuffer?: ArrayBufferLike,
   ): Promise<CacheLoadResult> => {
     try {
       const cacheLoadStart = performance.now();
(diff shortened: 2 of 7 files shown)


louistrue:
Addressed all three review findings in 71c0932f:

- **Desktop cache hits (Codex P1):** the desktop cache doesn't persist a `sourceBuffer`, so the restored store took the no-source branch and shipped without accessors — same crash. `loadFromCache` now accepts a fallback source buffer and `loadFile` passes the freshly read file buffer, so desktop hits hydrate fully.
- **`as unknown as` casts (CodeRabbit ×2):** removed. `BufferEntitySource` accepts the store's `entityIndex` structurally (the byId cast was unnecessary), and `getQuantities` needs no cast; only `getProperties` keeps a single narrowing `as PropertySet[]` (the on-demand extractor returns a richer pset shape than the contract).

Parser build + viewer/parser typecheck clean; parser suite 135/135.

PR #982, opened 2026-06-07
feat(materials): material property sets + Materials inspector tab (#978)

### LINKED ISSUE #978: Material Property Sets Not Displayed In Object Inspector
**Summary**

IfcLite appears to display property sets assigned directly to the selected IFC object, but not property sets assigned to the material associated with that object.

This makes it difficult to inspect material level information from the selected element, even when the IFC file contains valid material property data.

**Context**

In IFC, material properties should generally be assigned to the material, not duplicated on each object.

A typical workflow is:

1. Define a material, for example concrete.
2. Assign a material property set to that material, for example `Pset_MaterialConcrete`.
3. Associate the material with an element or element type, for example `IfcBeam`.

In this structure, the material property belongs to the material. The object receives the material information through the material association.

**Observed Behavior**

When selecting an element in IfcLite, the inspector shows property sets assigned directly to the object.

However, material related property sets are not shown.

This was observed with `IfcBeam` elements where the material property set exists on the material, but is not visible when inspecting the beam.

The issue appears especially relevant for beams using profile based material assignment.

**Expected Behavior**

When selecting an IFC object, IfcLite should also expose material property sets connected to that object through its material association.

The inspector should make it possible to view:

- Object property sets
- Type property sets
- Material assignment
- Material property sets

Ideally, these should be shown as separate groups so the user can understand where the information is stored.

**Relevant IFC Relationships And Entities**

IfcLite may need to resolve material property sets through relationships and entities such as:

- `IfcRelAssociatesMaterial`
- `IfcMaterial`
- `IfcMaterialProperties`
- `IfcMaterialProfileSet`
- `IfcMaterialProfileSetUsage`
- `IfcMaterialProfile`
- `IfcMaterialLayerSet`
- `IfcMaterialLayerSetUsage`
- `IfcMaterialConstituentSet`

The exact path depends on how the material is assigned in the IFC file.

**Why This Matters**

Some tools do not expose material property sets well. Because of that, users sometimes duplicate material data directly on the object as a custom property set.

That workaround can be useful for downstream software, but it is not the clean IFC methodology.

There are not many IFC viewers that make material property sets easy to inspect, outside tools like Bonsai BIM. It would be very useful to have this visible in IfcLite as well, especially because IfcLite is lightweight, browser based, and easy to share with people who only need to review or check IFC data.

This would support workflows such as:

- IFC deliverable checking
- Material data validation
- Cost and quantity workflows
- Concrete and steel property review
- Checking whether information is stored on the object, type, or material

**Suggested Improvement**

Add support in the object inspector for material related property sets.

Recommended grouping:

- Object Psets
- Type Psets
- Material Psets

This would allow users to distinguish direct object data from inherited or associated material data.

AI generated report from a conversation in BIMvoice Academy: https://www.skool.com/bimvoiceacademy/material-pset?p=53b5ac8a about material property sets.

Thank you @louistrue!

louistrue:
Closes #978

## Problem
Material property sets attach to an `IfcMaterial` via `IfcMaterialProperties` (e.g. `Pset_MaterialConcrete`) — they reference the material **directly**, not through `IfcRelDefinesByProperties`. That's why they never appeared in the object inspector, even when the IFC file carried valid material property data (especially beams using profile-based material assignment).

## What this adds

**1. Material property sets on the selected object**
A new **"Material Properties"** group in the inspector, resolved through the element's material association — mirroring how Type Properties already surface on an occurrence. A layer / profile / constituent set is fanned out to each member `IfcMaterial`, so each material's own psets are shown and labelled.

**2. A new "Materials" hierarchy tab** (4th toggle next to Spatial / Class / Type)
Lists every base material. Selecting one **isolates its elements** in 3D and shows, in the inspector:
- the material's **own property sets** (`IfcMaterialProperties`), and
- **aggregated totals** — element count, volume, area, weight — summed across every using element and **apportioned by each element's material share** (layer thickness / constituent fraction), so a layered wall's volume is split between its concrete and insulation rather than double-counted.
- a by-class breakdown, plus an explicit caveat when volume quantities are missing.

This matches the grouping the issue asked for (Object psets / Type psets / **Material psets**) and the "select a material → see totals" workflow.

## Implementation notes
- **No Rust/wasm changes.** All pset / quantity / material metadata is resolved client-side from the STEP buffer via the on-demand TS extractors; the wasm `processGeometryBatch` path is geometry/styling only. Keeping it viewer-side avoids a second metadata pipeline.
- **Parser** (`@ifc-lite/parser`, additive exports): `extractMaterialPropertiesOnDemand`, `extractMaterialPropertiesForMaterialId`, `buildMaterialUsageIndex`, `collectMaterialLeaves`, `resolveMaterialDefId`, `getMaterialDisplay` (+ `MaterialPsetGroup` / `MaterialLeaf` / `MaterialUsage` types). `IfcMaterialProperties` (IFC4+) and `IfcExtendedMaterialProperties` (IFC2x3) are parsed with explicit per-class attribute layouts; typed IFC2x3 scalar subtypes are skipped gracefully.
- **Viewer**: `MaterialTotalsPanel`, the Materials tab (`buildMaterialTree`, grouping mode, click → select material entity + isolate), and the instance "Material Properties" group. A material is selected by reusing the existing `selectedEntity` channel (an `IfcMaterial` is a real entity), detected in `PropertiesPanel` and short-circuited to the totals panel.

## Testing
- New parser test suite (`on-demand-material-properties.test.ts`, 10 cases) covering direct/layer-set/constituent resolution, thickness-weighted volume splitting, the usage-index inversion, and the IFC2x3 `IfcExtendedMaterialProperties` layout. Full parser suite: **144 passing**.
- Viewer typecheck: no errors in any changed file.
- Reviewed via a multi-agent adversarial pass; fixed a `QuantityType` enum off-by-one in the totals aggregation and hardened the IFC2x3 attribute handling.

## Known v1 limitations
- Totals assume SI quantity units (m³/m²/kg) — the standard IFC base-quantity convention.
- Federated same-named materials merge by name; the shown psets come from the representative model.

## Summary by CodeRabbit

* **New Features**
  * Added Materials inspector tab to view and isolate building materials by name
  * Exposed material property sets for selected elements and materials
  * Display aggregated material totals including element count, volume, area, and weight
  * Material properties now appear in the Properties panel alongside standard properties


louistrue:
@coderabbitai review

louistrue:
@coderabbitai Thanks — addressed in 041bc9be:

**Perf (totals loop):** the loop now skips the per-element `extractQuantitiesOnDemand` call (and its parser allocation) for elements with no quantity sets, gating on `onDemandQuantityMap`. So a material used by thousands of elements only pays the parse cost for the subset that actually carries Qto data, and the whole computation is `useMemo`-d per material selection. Batch quantity extraction remains a sensible follow-up if profiling on very large single-material models shows lag.

**Docstring coverage:** added docstrings to the new helpers.

PR #987, opened 2026-06-07
feat(viewer): expose model comparison in the web UI (#924)

louistrue:
Compare two loaded models in the viewer, built on the `@ifc-lite/diff` engine (#939, now in main). Continues #971, which GitHub auto-closed when its stacked base `feat/model-diff` merged — same branch, rebased onto main.

## Feature
- **ComparePanel** (Analysis menu): pick two federated models as A/B, choose a `data` / `geometry` / `both` scope, run the diff, review added / modified / deleted elements colour-coded in 3D + listed.
- Geometry hashing rides on `MeshData.geometryHash` through the WASM mesh path (canonical load, #975), so primary + federated models both get hashes — the original "every element changed" bug is structurally gone.

## Clean data ⟂ geometry separation
A pure move (Z-height / placement) now counts as **geometry only**, never `data · geometry`. The data fingerprint excludes geometry-derived data (quantities + placement-leaking properties — elevation, offsets, coordinates) via a single shared boundary (`geometricData.ts`).

## "What changed" detail
Selecting a change shows per-field data deltas (attribute / property / type / type-assignment, before → after) and a geometry summary (centroid move distance + Δx/Δy/Δz, reshape flag), scope-aware and computed lazily from both stores.

## Review fixes (CodeRabbit + Codex, all resolved on-branch)
- Non-blocking fingerprint build (yields so large models don't freeze the viewport).
- Warns when a compared model lacks geometry hashes (non-WASM load path).
- Restores lens colours on compare teardown instead of blanking the channel.

Viewer `tsc` clean; compare + diff + geom_hash tests pass.

## Release Notes

* **New Features**
  * Added Model Comparison UI: New Compare panel (Analysis menu) enables users to select two loaded models and identify added, modified, and deleted elements with color-coded visualization.
  * Comparison detail view displays what changed (geometry repositioning/reshaping, property updates) for each selected element.
  * Scope toggle lets users focus on data changes, geometry changes, or both.


louistrue:
Review round addressed in 7ebe45da:

**Fixed**
- **WASM leak** (geometry-coordinate.ts): geometry-hash extraction moved inside the `try` so `collection.free()` runs in `finally` on throw.
- **Lens restore gap** (useCompareOverlay): the compare overlay now owns the colour channel whenever a comparison is shown (not gated on override-map size), so an empty result still hands the channel back to an active lens.
- **Stale comparison** (ComparePanel): a comparison computed against a since-removed model is now dropped, restoring the survivor's scene.
- **Over-eager data filter** (geometricData): narrowed — generic `Location` / `Position` / `Datum` / bare `Offset` stay in the data diff (often semantic); only `elevation` / `placement` / `coordinate` / qualified level-offsets are treated as geometry. Test updated.
- **Changeset wording**: corrected — a moved element *is* a geometry change.

**Deferred, with reasoning**
- **Per-entity extraction cost** (P1): `extractPropertiesOnDemand` seeks the entity via the index (O(entity), not a whole-source rescan), so a full comparison is one bounded pass over meshed entities; the added periodic `await` yield keeps the viewport responsive. Moving the build to a worker is a worthwhile follow-up for very large federations but isn't required for correctness or typical-model usability.
- **`± idOffset` vs FederationRegistry**: the offset arithmetic matches the established federation ID convention used across the renderer / selection / overlay (`globalId = localId + idOffset`). Routing only the compare adapter through registry APIs would be inconsistent; a registry migration is a repo-wide change, out of scope for this PR.

Issue #952, opened 2026-06-05
Types authoring: unify the data-store type surface & decouple type-checking from the wasm build


louistrue:
Surfaced while reviewing #950 (`IfcStoreBase` decoupling). That PR was good, but the way it broke — and how the break stayed hidden — points at a few structural type-authoring problems worth fixing on their own.

## 1. Multiple divergent `IfcDataStore` definitions that drift

There are (at least) two separate interfaces both named `IfcDataStore`, at different layers, that have drifted apart:

- `@ifc-lite/parser` → `IfcDataStore` (the live runtime store): field is **`schemaVersion`**, plus `source`, `strings`, `parseTime`, on-demand maps, and now `extends IfcStoreBase` (the four `getEntity`/`getEntitiesByType`/`getProperties`/`getQuantities` accessors).
- `@ifc-lite/cache` → `IfcDataStore` (`packages/cache/src/types.ts:121`, the *serialized* shape): field is **`schema`**, no `source`/`parseTime`, optional `entityIndex` of a different `CacheEntityIndex` type, no accessors.

Plus the lean `@ifc-lite/data` `IfcStoreBase`, plus several hand-built duck-typed stores in the viewer (`serverDataModel`, `nativeSpatialDataStore`).

A distinct type for the on-disk/serialized format is fine — but reusing the **name** `IfcDataStore` and naming the same concept **`schema`** vs **`schemaVersion`** is pure drift. It's confusing and it forces casts at the boundary.

## 2. Unsafe casts at boundaries hide real bugs

The cache→runtime bridge is done with escape hatches that turn off the compiler:

- `apps/viewer/src/hooks/useIfcCache.ts:126` → `const dataStore = result.dataStore as any;`
- `apps/viewer/src/utils/desktopModelSnapshot.ts:331` → `result.dataStore as unknown as IfcDataStore;`

Because #950 added **required** accessor methods to `IfcDataStore`, every *typed* construction site errored (good — `serverDataModel`/`nativeSpatialDataStore` got caught and fixed), but these `as any` / `as unknown as` sites stayed silently green. Both restored stores flow straight into the query path (`setIfcDataStore(...)` → `model.ifcDataStore` → `new EntityNode(...)` → `store.getProperties()`), so without the accessors they throw `TypeError: store.getProperties is not a function` at runtime on the **everyday cache-reopen flow**. Only manual tracing caught it.

[...]

PR #993, opened 2026-06-07
refactor(cache): rename cache IfcDataStore→CacheDataStore + typed hydrateCacheStore adapter (#952)

louistrue:
Stacked on #992. Addresses the two high-value items from #952.

## What & why

**1. Name collision killed.** The cache's *serialized* store was also called `IfcDataStore`, colliding with the parser's *runtime* `IfcDataStore` — the viewer was already importing it `as CacheDataStore` to work around it. Renamed to `CacheDataStore` across `packages/cache`.

**2. The dangerous cast is gone.** `useIfcCache` bridged cache→runtime with `result.dataStore as unknown as IfcDataStore` — the escape hatch that disabled all type-checking and let the #950 `store.getEntity is not a function` crash slip through. Replaced with a typed `hydrateCacheStore()` adapter. The cache and runtime stores share the same `@ifc-lite/data` table types, so the field mapping is **compiler-checked** — a future required store member is now a compile error here instead of a silent runtime crash.

## The cast was hiding real bugs

The moment the construction became typed, tsc surfaced two genuine mismatches:
- **`schema` (cache numeric `SchemaVersion` enum) vs `schemaVersion` (runtime string union)** — the exact drift #952 named. Now an explicit `cacheSchemaToVersion()` conversion (the cache format predates IFC5, matching the save path's inverse mapping).
- a `Uint8Array<ArrayBufferLike>` vs `<ArrayBuffer>` generic mismatch.

The adapter also seeds the previously-missing required fields (`parseTime`, `fileSize`) the cast silently omitted. As a small robustness win, no-source cache hits now get safe no-op accessors (`getProperties` falls back to the pre-built cache tables) instead of crashing.

## Notes
- The desktop removal (#992) already deleted the *other* #952 cast site (`desktopModelSnapshot.ts`), so this is the last one.
- `schema` is kept as the cache's binary-format field; the adapter is the single typed boundary that maps it to `schemaVersion`.
- Remaining #952 item — a wasm-free typecheck lane — left as a separate infra follow-up.

## Verification
- `packages/cache` typechecks clean; **16/16 cache tests pass** (incl. the serialization round-trip).
- Viewer typechecks clean — `useIfcCache` has **0 errors** (the only remaining viewer tsc errors are the pre-existing geometry/wasm-binding-staleness noise, identical on `main`).

### DIFF
--- a/.changeset/cache-datastore-rename.md
+++ b/.changeset/cache-datastore-rename.md
@@ -1,0 +1,7 @@
+---
+"@ifc-lite/cache": major
+---
+
+Rename the serialized data-store type `IfcDataStore` → `CacheDataStore`.
+
+This removes the name collision with `@ifc-lite/parser`'s runtime `IfcDataStore` — the two are structurally different (the cache type is the on-disk/serialized shape, keyed on a numeric `schema` enum, with no `source`/`parseTime`/accessors). Consumers importing the type from `@ifc-lite/cache` must switch `IfcDataStore` → `CacheDataStore`.
--- a/apps/viewer/src/hooks/useIfcCache.ts
+++ b/apps/viewer/src/hooks/useIfcCache.ts
@@ -13,11 +13,12 @@ import { useCallback } from 'react';
 import {
   BinaryCacheWriter,
   BinaryCacheReader,
+  SchemaVersion,
   type CachedEntityIndexColumns,
-  type IfcDataStore as CacheDataStore,
+  type CacheDataStore,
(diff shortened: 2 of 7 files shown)


louistrue:
Addressed the Codex P1 in ab4a0212: added a **major** changeset for `@ifc-lite/cache` covering the `IfcDataStore` → `CacheDataStore` public-export rename (per AGENTS.md: renaming an export from a ≥1.0 package is a major bump).

Issue #957, opened 2026-06-06
Model fails to display due to unsupported IfcBlobTexture


xyzbety:
In the attached model（tessellation-with-blob-texture）,The IfcBlobTexture (# 57) entity defined in the model is not supported by the parser. As this embedded texture type is unrecognized, the associated geometry fails to render correctly, resulting in the model being invisible or appearing empty.

[tessellation-with-blob-texture.zip](https://github.com/user-attachments/files/28663090/tessellation-with-blob-texture.zip)

The model is sourced from the official documentation of buildingsmart : https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/annex_e/tessellated-shape-with-style/tessellation-with-blob-texture.html

<img width="2870" height="1635" alt="Image" src="https://github.com/user-attachments/assets/d03c4b51-c7e4-4eaa-a367-81f18f67ee93" />

<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/16cdf2d4-2111-42e9-92f2-89c3b7751859" />

PR #1000, opened 2026-06-07
fix(viewer): only show the Model/Types 3D switch when the model has type geometry

louistrue:
## Problem

The #957 follow-up added a **Model / Types** switch to the 3D View panel, but it rendered for *every* loaded model. "Types" mode only displays type-library geometry (`geometryClass` 1 = orphan type, 2 = instanced type); most models carry only occurrence geometry (`class 0`), so for them the switch was a dead control — toggling to "Types" produced an empty scene.

## Fix

Gate the switch on whether the rendered geometry actually contains any type geometry.

- **`visibilitySlice`** — add a runtime-only `hasTypeGeometry` flag (default `false`) plus a `setHasTypeGeometry` setter that no-ops when unchanged, so the toolbar doesn't re-render on every streaming batch.
- **`ViewportContainer`** — derive `hasTypeGeometry` from the merged mesh set with an incremental, early-exiting scan: it short-circuits the moment a `geometryClass !== 0` mesh is seen (so the common occurrence-only model costs at most one short linear pass), and it re-derives per streaming batch so type geometry arriving late still flips it on. The flag is published to the store via `useEffect`. The filter now runs through `effectiveViewMode = hasTypeGeometry ? typeViewMode : 'model'`, so a persisted `'types'` preference from a prior model can't strand the user on an empty scene with a now-hidden switch.
- **`MainToolbar`** — wrap the "3D View" heading, the Model/Types buttons, and the trailing separator in `{hasTypeGeometry && (…)}`.

## Behavior

- Occurrence-only model (the common case) → switch hidden.
- Model with orphan/instanced type geometry → switch shows; "Types" always has content.
- Federation: switch shows if **any** loaded model has type geometry; removing the last type-geometry model hides it again.
- Stale `'types'` preference meeting a no-type model → falls back to the model view, no empty scene.

## Gating choice

Gates on **any** type geometry (`class 1` *or* `2`), giving the clean invariant "show the switch exactly when Types mode would render something." The alternative — gating only on instanced types (`class 2`) — would also hide the switch for the rare pure-orphan-library model where Model and Types views are identical anyway. Easy to narrow if preferred.

## Verification

- `tsc --noEmit`: only the pre-existing `@ifc-lite/diff` module-resolution errors (CI-built bindings, not built locally); none in the edited files.
- `visibilitySlice` tests: 36/36 pass.
- No existing test asserts the switch always renders.

## Release Notes

* **New Features**
  * The 3D viewer now intelligently detects whether loaded models contain type-library geometry and conditionally displays the "Types" option in the Model/Types view mode selector. This smart behavior prevents showing irrelevant visualization options for models that don't support type geometry, providing a cleaner interface experience.

### DIFF
--- a/apps/viewer/src/components/viewer/MainToolbar.tsx
+++ b/apps/viewer/src/components/viewer/MainToolbar.tsx
@@ -462,6 +462,10 @@ export function MainToolbar({ onShowShortcuts }: MainToolbarProps = {} as MainTo
   // occurrences (default), 'types' shows the type-library shapes.
   const typeViewMode = useViewerStore((state) => state.typeViewMode);
   const setTypeViewMode = useViewerStore((state) => state.setTypeViewMode);
+  // Only models with type-library geometry (RepresentationMap shapes) can show
+  // anything in "Types" mode, so the switch is hidden for the common
+  // occurrence-only model. Derived in ViewportContainer from the merged meshes.
+  const hasTypeGeometry = useViewerStore((state) => state.hasTypeGeometry);
   // How many of the five class toggles are on — surfaced in the menu
   // header so the user sees scene state at a glance.
   const visibleClassCount = [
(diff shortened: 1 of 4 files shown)


louistrue:
### This also fixes a regression introduced by #998

While testing, a user hit: *start a blank project (no IFC), add a slab/wall — the toast says "added" but nothing renders in 3D.* Root cause traced to #998's `geometryClass` filter:

```js
if (typeViewMode === 'types') {
  if (geometryClass === 0) continue;   // drops ALL occurrences
} else if (geometryClass === 2) {
  continue;
}
```

`typeViewMode` is persisted to `localStorage`, so once you click **Types** (which you do to try the new feature), it sticks across sessions. Add-element preview meshes (`addElementMeshes.ts` → `buildElementMesh` → `appendGeometryBatch`) never set `geometryClass`, so they're class 0 and get dropped in `'types'` mode. The mutation succeeds (toast fires) but the mesh is filtered out before render.

This isn't limited to editing — on plain `main`, **loading any model with no type geometry while `typeViewMode==='types'` makes the whole model invisible** with no way to recover except clearing the setting (the switch itself is the only fix, and you might not realize the mode is the cause).

This PR fixes it: `effectiveViewMode = hasTypeGeometry ? typeViewMode : 'model'`. With no type geometry present (blank project, or any occurrence-only model), the filter falls back to `'model'` and class-0 geometry always renders. The switch is also hidden in that state, so the stale preference can't be re-triggered.

**Known remaining nuance (not addressed here):** adding an occurrence element while a type-geometry model *is* in legitimate `'types'` view still won't show the new element (occurrences are hidden by design in Types mode). Happy to auto-switch to `'model'` on add-element if that's preferred.

louistrue:
### Folded in: auto-switch to Model when authoring adds geometry

Addressed the remaining nuance from the comment above. Adding / duplicating / splitting an element now flips the 3D view back to **Model** if it was on **Types**, so the new class-0 occurrence is always visible (`revealAddedGeometryInModelView` after each `appendGeometryBatch` in the three authoring paths: add element, add column, split/clone). It's a no-op when already in Model view, so the persisted preference isn't overwritten gratuitously. The IFC loader's streaming `appendGeometryBatch` is intentionally left alone — opening a type-geometry model should still honour a persisted Types choice.

Now "add → see it" holds regardless of which view you were in.

louistrue:
Addressed CodeRabbit's finding in `efcca367`: `setHasTypeGeometry` now returns the existing `state` reference (not `{}`) on a no-op, so Zustand skips the subscriber notification entirely.

PR #1005, opened 2026-06-07
fix(geometry): cut Tekla flush recesses with real boolean (#977)

### LINKED ISSUE #977: bug(geometry): flush/end-of-member IfcOpeningElement recesses under-cut (coincident-face boolean leaves a residual wall)
## Summary

On steel-fabrication models (e.g. Tekla exports), recesses/notches modelled as `IfcOpeningElement` voids **are subtracted, but under-cut**: where the cutter is flush/coincident with a member face — typically an end-of-member recess — the boolean difference leaves a thin residual wall (a near-zero-thickness "skin"), so the cut doesn't fully break through. Visually it reads as a recess with a leftover wall / cut that doesn't reach the face. This is a **systemic** pattern across many parts and matters for manufacturing consumers who need exact cutouts.

Image

Image

Image

Bim Vision
Image

Image

## Root cause

The opening cutter solid is built to its **exact authored extents**, so for a recess whose cut face is **coincident/coplanar with the host's face** (the common end-of-member / face-flush case), the boolean difference has two exactly-coincident faces at the cut boundary. The CSG/manifold kernel can't reliably classify material across coincident faces, so it **under-removes** there — leaving a residual ~0-thickness wall instead of a clean through-cut.

There is currently **no over-extrusion / inflation of the cutter** before the difference (the opening mesh is fed to `CsgProcessor::subtract_mesh` / the manifold difference as-authored — see `rust/geometry/src/csg.rs`, `rust/geometry/src/manifold_kernel.rs`). So any cutter that ends exactly on a host face is in this degenerate-boundary situation.

## Proposed fix

**Over-extrude / inflate the opening cutter by a small epsilon before subtracting**, so no cutter face is exactly coincident with a host face and the recess cuts cleanly through:

- When building the opening solid for void subtraction, extend it by a small epsilon along its extrusion direction (both ends), and/or push its boundary faces outward by epsilon, so it protrudes slightly past the host faces it touches.
- Epsilon in model units (e.g. ~0.1–1 mm or a relative factor of the host size) — large enough to clear coincident-face degeneracy, small enough not to change the visible cut.
- This is the standard "cut exactly on a face" remedy and fixes the leftover-wall (under-cut) class without affecting interior cuts (which already clear the boundary).

**Regression fixture:** a box with a recess whose cutter face is exactly flush with one box face (and one at the extruded end) — assert the cut fully penetrates: no residual wall (min remaining thickness along the cut ≈ 0 / face removed), result watertight & manifold.

## Notes

- Symptom is **under-cut (leftover wall)**, not a missing cut — so it's the boolean coincident-face handling, not the bbox-overlap guard.
- Separately, `bounds_overlap` in `csg.rs` uses a strict (epsilon-free) AABB test that would *drop* a fully-flush cutter; not the cause here, but worth hardening with a tolerance in the same pass so an over-extruded cutter that becomes flush isn't then rejected.

## Environment

- `@ifc-lite/*` <version>
- Source models: Tekla Structures fabrication exports, IFC2X3

Blogbotana:
Fixes #977.

## Problem
Tekla steel exports model end recesses, corner notches and slots as
`IfcOpeningElement` voids. Two things went wrong cutting them:

1. **Coincident faces (the original #977 report).** Where a cutter face is flush
   with the host face, Manifold's coplanar classifier under-removes (and isn't
   bit-deterministic across platforms), leaving a thin residual wall.
2. **Tilted profile members (the real headline bug).** Steel members (channels,
   tubes, I-beams, gusset plates) are usually tilted in world. Their openings
   were cut by the analytic **axis-aligned bounding-box** clip. The AABB of a
   tilted thin cutter plate is far larger than the authored cutter, so it
   **over-cut** — removing real section material and leaving a thin sliver wall
   instead of the full profile. This was a project-wide error on every tilted
   member, and the analytic path also fabricates reveal/cap walls in empty space
   on open profile sections.

## Fix
- **Cut profile-section openings with their REAL mesh, not a bounding box.**
  In `classify_openings`, every opening on a profile-section host
  (`IfcBeam`/`IfcColumn`/`IfcMember`/`IfcPlate`) is routed through the real-mesh
  boolean path (`NonRectangular`). The cut then matches the authored cutter
  shape exactly — no AABB inflation, correct interaction with profile fillets —
  and the kernel's perturbation clears coplanarity with the profile's inner
  faces. Solid plate hosts (slab/wall/roof) keep the analytic AABB path, which
  is correct there and, unlike Manifold, deterministic (so fixture #853 stays
  stable across platforms).
- **Re-tune the Manifold cutter perturbation** to clear the kernel's
  *host-relative* coplanarity tolerance: size the inflation off the combined
  host+void bbox and the cutter's smallest half-extent, with floor/ceiling
  clamps. The previous cutter-bbox-relative magnitude was below tolerance for
  long mm-unit members, leaving flush faces uncut.
- **Harden `bounds_overlap`** with `<=`/`>=` + a relative epsilon so a flush
  cutter isn't dropped before inflation.

## Tests
- Unit: Manifold perturbation invariants (clears tolerance; ceiling-clamped
  still clears); profile-section host routing.
- Integration: `tekla_flush_end_recess_cuts_through` — point-in-solid ray cast
  asserting full penetration, no residual, no over-cut.
- Full `ifc-lite-geometry` suite green (incl. #853 slab pocket, #864, #832,
  #635 gable clips, wall-opening regressions, AC20-FZK-Haus).

## Verification
Built the WASM bundle and confirmed on the reporter's Tekla model in the viewer:
end recesses, corner notches and a slot in a round tube now cut cleanly with no
residual or over-cut, on tilted members across the project; doors/windows and
AC20 references unchanged.

## Out of scope
A separate normals/winding regression on `main` (AC20 roof renders inside-out —
a pure extrusion, no CSG) is tracked in #1006; it is not caused by this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate recess cutting for tilted/thin/profile openings—avoids spurious geometry and over-cutting.
  * Improved boundary-tolerance for flush/near-flush cutters so touching cutters are recognized correctly.
  * Axis-aligned box openings remain deterministic; non-box/over‑approximating openings use exact mesh booleans.
* **Tests**
  * Added regression tests validating shallow flush recess cuts and preventing both under- and over-cutting.


Blogbotana:
Found an issue. Fixing

Blogbotana:
We shouldn't use AABB for cutting in that case 
image

Blogbotana:
### Root cause & final approach (after testing on the Tekla model)

The original #977 framing (Manifold coplanar-classifier under-cut) is real, but it was **not** what produced the visible defects on this model. Debugging the actual beams showed two things:

1. The failing recesses/notches are **clean boxes** that the router classifies as `Rectangular`/`DiagonalRectangular` → they take the **analytic AABB clip**, not the Manifold path. So re-tuning the kernel perturbation alone (the issue's suggested fix) didn't change them.
2. The headline bug is **over-cut on tilted members**: steel sections are tilted in world, and the **axis-aligned bounding box** of a tilted thin cutter plate is much larger than the authored cutter → it removed real section material, leaving a thin sliver wall instead of the full profile. Project-wide on every tilted member. The analytic path also fabricates reveal/cap walls in empty space on open profile sections (channels, tubes).

**Fix:** route every opening on a profile-section host (`IfcBeam`/`Column`/`Member`/`Plate`) through the **real-mesh boolean** in `classify_openings` — the cut matches the authored cutter shape exactly (no AABB inflation, correct with profile fillets), and the perturbation handles coplanarity. Solid plate hosts (slab/wall/roof) keep the analytic path (correct + deterministic, so #853 stays stable). Net −290 lines vs the interim AABB-box approach.

Verified in the viewer on the reporter's model: end recesses, corner notches and a round-tube slot now cut cleanly on tilted members; doors/windows and AC20 unchanged.

(Separate `main` normals/winding regression — AC20 roof inside-out, a pure extrusion — is tracked in #1006 and is not caused by this PR.)

louistrue (commented):
please merge current main into this branch

Blogbotana:
I made merge with the main branch and the issue is still remaining 
image

Blogbotana:
In this preview I don't see that https://ifc-lite-git-fix-issue-977-tekla-flush-recess-ltplus.vercel.app/
Could this be an error from my local build?

PR #1017, opened 2026-06-08
feat(geometry): exact-predicate plane-clip classifier (pure-Rust kernel, Phase 1)

louistrue:
First **behavioral** slice of the pure-Rust CSG kernel (stacked on #1016, the determinism floor).

## What
An opt-in `exact-predicates` feature that makes the plane-clip front/back **decision** exact:
- `Plane::orient_front` — routes the decision through Shewchuk `orient3d` (`robust` crate: sign-exact, `no_std`, no C, wasm-clean) instead of the float `signed_distance >= -epsilon` band. Exactly-on-plane → `orient3d == 0` → front, deterministically on every platform. This removes the float-epsilon misclassification behind the coincident-/near-coplanar-face determinism failures — the class that collapses Manifold on Linux x86_64.
- `clip_triangle` now classifies via a `vert_front` helper: exact under the feature, **byte-identical float behaviour without it**. Split-point interpolation stays float (Phase 1 = exact *topology*, not exact coordinates — matches the spike findings).

## Why this is safe
- Feature is **off by default** → default builds unchanged, `robust` not linked (verified: `--no-default-features` build pulls no `robust`).
- With `--features exact-predicates` ON, the **entire `wall_opening_cut_regression` suite passes** — including the IOS-ground-truth cut counts (`wall_552611_2_openings_matches_ios`, etc.) — identical to baseline (fixtures staged).
- New feature-gated parity tests: exact agrees with float on unambiguous points; on-plane is deterministically front; a straddling triangle clips to the front side.

## Context
Part of the pure-Rust CSG consolidation (see the kernel mapping + gate verification). The predicate-floor and chained-clip carve spikes passed locally, and the x86_64 determinism floor (#1016) is green in CI. This is migration-plan step 3 (exact on-plane classifier) landed behind a flag; reveal/cap winding, the Brep fallback, and flipping the default come later.

> Note: `robust` becomes an optional dependency (kept as dev-dep for #1016's floor test). No changeset — dormant opt-in feature, no default behavior change.

## Summary by CodeRabbit

* **New Features**
  * Added optional exact-predicate mode for enhanced geometric precision in CSG operations

* **Tests**
  * Enhanced test coverage for geometric calculations with exact-predicate validation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -349,6 +349,12 @@ jobs:
           pnpm fixtures
       - name: Manifold CSG tests
         run: cargo test -p ifc-lite-geometry --features manifold-csg
+      # Pure-Rust CSG kernel (Phase 1): runs the geometry suite with the exact
+      # `orient3d` plane-clip classifier on x86_64, so the exact path (incl. the
+      # wall-opening regression fixtures) is continuously covered, not just the
+      # feature-off default. `--no-default-features` skips the Manifold C++ build.
+      - name: Exact-predicate CSG tests (pure-Rust kernel)
+        run: cargo test -p ifc-lite-geometry --no-default-features --features exact-predicates
 
   # Aggregate gate. Keeps the existing required status check name green
   # for the branch protection rule so we don't have to retouch settings.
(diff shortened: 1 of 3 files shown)


louistrue:
Folding into the full pure-Rust kernel effort. Per the build-complete-then-flip plan, the exact `orient3d` classifier becomes part of the single canonical kernel (unflagged), not a parallel `exact-predicates` feature path. Closing; the logic is preserved in the kernel branch.

PR #1019, opened 2026-06-08
feat(geometry): pure-Rust tier-1 through-opening carve (i_overlay re-extrude)

louistrue:
First **kernel-replacement** primitive for the pure-Rust CSG direction (independent of #1017 — uses `bool2d` + `extrude_profile`, already on `main`).

## What
`extrude_profile_through_openings(host_profile, openings, depth, transform)`:
1. subtract each opening's cross-section from the host swept profile via the **robust 2D boolean** (`bool2d::subtract_multiple_2d` → i_overlay, **no polygon cap**);
2. re-extrude — `extrude_profile` turns the resulting holes into the opening **reveal walls** with **no internal cap** (a true through-opening).

## Why it matters
This is the **kernel-free fast path** for coplanar through-openings — the dominant window/door case. It carves **round/arched openings exactly** (no BSP 128-poly cap, so no square AABB holes — the server's #635 export bug at the root), is **pure Rust (no C++)**, and is **deterministic** (i_overlay is robust). It replaces the BSP/Manifold mesh subtract for this class once the void router routes qualifying openings to it.

## Test
A 24-gon opening through a 4×3 m wall: the carved profile keeps a **≥8-vertex round hole** (not a 4-vertex square), the mesh spans the full extrusion depth with **no cap**, and the carve is **byte-deterministic** on repeat.

## Scope
Additive — a new `pub fn` + test; **no existing path changed** (the void-router *routing* to this fast path, plus reveal/cap winding and the projection of opening geometry into the profile plane, is the follow-up). Part of the pure-Rust kernel consolidation; the determinism floor (#1016, merged) and exact classifier (#1017) are the prior slices.

## Release Notes

* **New Features**
  * Added support for creating 3D extrusions with true through-openings that completely penetrate objects without leaving internal caps, enabling precise geometric apertures.

* **Tests**
  * Validates that rounded apertures maintain geometric precision and consistently span full depth across repeated execution cycles, confirming deterministic behavior.

### DIFF
--- a/rust/geometry/src/extrusion.rs
+++ b/rust/geometry/src/extrusion.rs
@@ -203,6 +203,33 @@ pub fn extrude_profile_with_voids(
     Ok(mesh)
 }
 
+/// Pure-Rust tier-1 through-opening carve (CSG-kernel-free).
+///
+/// Subtracts each opening's cross-section contour from the host swept profile
+/// via the robust 2D boolean ([`crate::bool2d::subtract_multiple_2d`],
+/// i_overlay — **no polygon cap**, so round/arched openings are cut exactly),
+/// then re-extrudes. `extrude_profile` turns the resulting holes into the
+/// opening reveal walls automatically, with no internal cap (a true
+/// through-opening).
+///
+/// This is the kernel-free fast path for coplanar through-openings — the
+/// dominant window/door case — replacing the BSP/Manifold mesh subtract for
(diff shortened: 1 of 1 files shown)


louistrue:
Folding into the full pure-Rust kernel effort. The tier-1 through-opening carve becomes an internal fast path of the canonical kernel, not a standalone dormant primitive. Closing; logic preserved in the kernel branch.

PR #1057, opened 2026-06-10
perf(ids): diagnostics-free applicability + inverted property-value index

louistrue:
Stacked on #1055. Makes IDS validation fast **and** gives it a working progress UI in the viewer.

## Why

After #1055's caching, ~86% of remaining validation CPU built failure-diagnostic objects for ~8M applicability *misses* whose only consumer is a boolean. And in the viewer, validation ran on the main thread (unlike parse/geometry, which run in workers) so the UI couldn't paint progress. Prior art confirms the architecture, not the language, is the fix: IfcOpenShell's ifctester gained ~4600x on its applicability broadphase and Xbim validates the same packs 'in seconds' by producing diagnostics only after selection and inverting property applicability to the pset side.

## What

**Performance (packages/ids):**
- `facetPasses()` — diagnostics-free boolean cores for entity + property facets, pinned to `checkFacet().passed` by a differential test.
- `ApplicabilityPropertyIndex` — per-run inverted `(pset, prop) -> value -> entityIds` index; lookups are a candidate **superset** always re-confirmed by the boolean core, so verdicts can't change.
- Bounded enumeration strings, memoized constraint formatting, comparator fast-paths.

**Off-main-thread (apps/viewer):**
- `idsValidation.worker.ts` re-parses the model source bytes (~150ms) into its own store and runs the shared validator off the main thread; `idsWorkerClient.ts` shares the SAB source zero-copy and streams progress via `postMessage`. Falls back to in-process validation when unavailable. IDS XML is parsed on the main thread (workers have no DOMParser) — and `xml-parser.ts` now gates the `@xmldom/xmldom` fallback to the Node runtime so importing the package in a worker no longer crashes.

**The progress bar:**
- Root cause it never appeared: the `setIdsError` store setter unconditionally set `idsLoading:false`, and `runValidation` calls `setIdsError(null)` right after `setIdsLoading(true)` — flipping loading off for the whole run, so the `{loading && ...}`-gated progress panel never rendered. Fixed so clearing an error leaves loading alone.
- IDSPanel shows an advancing 'Validating specification N of M' counter + live 'Scanning/Checking X / Y' sub-line; the slow cold-extraction phase now reports granular progress.

**Parser:** `yieldToEventLoop` MessagePort-close leak fix consolidated into one shared module (was keeping the CLI process alive after large parses).

## Measured (550k-entity model, identical verdicts)

| pack | #1055 | this PR |
|---|---|---|
| 848 specs | 41 s | ~9 s (CLI) / ~14 s (browser worker, off-thread, live progress) |
| 277 specs | 45 s | 4.4 s |
| 117 specs | 12 s | 3.9 s |

Validation in the viewer now runs entirely off the main thread with a moving progress bar from click to completion.

Tests: scale-regression + differential parity suite in packages/ids (253), MessagePort leak test in packages/parser, summarize-status tests in packages/sdk. Viewer typecheck + production build green; the worker is emitted as its own chunk. All diagnostic logging removed.

Follow-up (optional): avoid the worker re-parse by transporting the existing store; revisit a Rust engine only if a hard trigger fires.

### DIFF
--- a/apps/viewer/src/components/viewer/IDSPanel.tsx
+++ b/apps/viewer/src/components/viewer/IDSPanel.tsx
@@ -521,16 +521,32 @@ export function IDSPanel({ onClose }: IDSPanelProps) {
   const renderProgress = () => {
     if (!progress) return null;
 
+    // Validation of large code-list IDS packs runs for many seconds, and
+    // a few broad specs dominate the time — so a percentage keyed on spec
+    // index sits near 0 for a while. Surface the always-advancing spec
+    // counter (and the per-spec entity count) so the panel visibly moves
+    // throughout, not just in the back half.
+    const specNumber = Math.min(progress.specificationIndex + 1, progress.totalSpecifications);
+    const isComplete = progress.phase === 'complete';
+    const headline = isComplete
+      ? 'Validation complete'
+      : `Validating specification ${specNumber} of ${progress.totalSpecifications}`;
+    const detail =
+      progress.phase === 'validating' && progress.totalEntities > 0
+        ? `Checking ${progress.entitiesProcessed.toLocaleString()} / ${progress.totalEntities.toLocaleString()} entities`
+        : progress.phase === 'filtering' && progress.totalEntities > 0
+          ? `Scanning ${progress.entitiesProcessed.toLocaleString()} / ${progress.totalEntities.toLocaleString()} candidates`
+          : progress.phase === 'filtering'
+            ? 'Finding applicable entities…'
+            : null;
+
     return (
       <div className="p-3 border-b">
-        <div className="flex items-center gap-2 mb-2">
-          <Loader2 className="h-4 w-4 animate-spin" />
-          <span className="text-sm">
-            {progress.phase === 'filtering' && 'Finding applicable entities...'}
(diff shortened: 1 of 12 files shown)


louistrue:
Added the architectural fix for the viewer UX (no progress feedback + frame-rate collapse during validation):

**IDS validation now runs in a Web Worker** (`apps/viewer/src/workers/idsValidation.worker.ts`), matching how the STEP parser and geometry already run off the main thread. It was the one heavy stage left on the main thread — and the bridge accessor re-parses STEP from the raw buffer per entity *there*, so the thread saturated, fps collapsed (~12), and the progress panel never painted until the run finished. Main-thread time-slice yielding can't fix a saturated thread.

The worker re-parses the model's source bytes (~150ms for 550k entities, negligible vs validation) into its own store so the main-thread store is untouched, then runs the shared `@ifc-lite/ids` validator. IDS XML is parsed on the main thread (no DOMParser in workers) and the plain `IDSDocument` crosses; a SAB source is shared zero-copy, a plain ArrayBuffer is copied+transferred. Progress streams via `postMessage` and now actually paints; `useIDS` paints a 'starting' state on button press for immediate feedback and falls back to in-process validation when the worker is unavailable. Vite emits `idsValidation.worker` as its own chunk.

Typecheck + production build green; the worker chunk is emitted alongside `parser.worker`/`geometry.worker`.

PR #1067, opened 2026-06-11
fix(renderer): grazing-angle shading bands + dashed corner seam lines

louistrue:
## Symptoms

- Large diagonal lighter/darker bands on flat walls and slabs, view-dependent, with stippled boundaries (worst when the camera is close to a long wall/slab, i.e. grazing view angles).
- Separation lines at wall/wall and floor/wall junctions rendering as dashed/broken segments instead of continuous lines.

## Root causes (empirically proven in a live session)

**1. Derivative flat-normal sign instability.** The flat-shading normal `cross(dpdx(worldPos), dpdy(worldPos))` (introduced by #872 to kill coplanar-strip scar lines) is numerically **sign-unstable at grazing angles**. Visualizing `N` as color showed a single horizontal plane fractured into regions of N=up, N=down, and N=sideways. The sign flip toggles the hemisphere-ambient and rim-light terms → tone bands with stippled, camera-dependent boundaries. The dark dashes along corners were the same flip on the 1–2 px slivers where the neighboring entity peeks past the anti-z-fight hash nudge.

**Fix:** keep the derivative normal's *direction* (scar-line immunity from #872 fully preserved) but stabilize its *sign* with the interpolated vertex normal, which carries no per-quad noise. Guarded for missing/near-perpendicular vertex normals (degenerate fallback unchanged). The textured shader inherits the fix through its anchored derivation — all five anchors verified intact.

**2. Seam-pass corner dashing.** The separation-lines pass gates ID boundaries on a relative first-difference depth test (5e-4, from #462, sized to hide the ≤2.55e-4 z-hash offsets). Depth-**continuous** corner seams sit right in that threshold's range and flicker along the line → dashes. Added a per-axis second-difference **crease gate** (3e-4 relative) OR'd with the old gate: corner seams have a depth-slope mismatch well above it, while coplanar continuations stay bounded by the z-hash offset (≤2.55e-4) and remain suppressed. Uses only already-loaded depth samples.

## Performance

- Zero new texture fetches; both changes are a handful of ALU ops in existing branches.
- Forced 300 continuous full renders with all effects on (7.8K elements / 370K tris, MSAA 4x): **16.67 ms avg, 0 frames > 20 ms** — locked 60 fps, identical to baseline.
- No CPU/load-time changes.

## Verification

- Reproduced both artifacts in the viewer (BIMcollab ARC example), confirmed root cause via debug shaders (normal visualization, conditioning visualization, worldPos grid), fixed, and re-verified the same camera poses: bands gone, corner lines continuous.
- Look-and-feel checked unchanged on lobby, corridor, exterior, and grazing poses with default visual-enhancement settings.
- `pnpm test` (83/83) and `tsc --noEmit` clean; adversarial review of both diffs (math, WGSL validity, anchor integrity, regression scenarios) came back ship-clean.

## Known accepted edges (documented in code)

- Coplanar features narrower than 2·radius px can sum two hash offsets past the crease gate → a (desirable) seam line on narrow inter-entity features.
- In ortho on very large scenes the crease gate degrades gracefully to the old first-difference behavior.

## Out of scope / follow-ups noticed

- The pick pass doesn't apply the main shader's z-hash nudge (and uses `expressId+1` per instance), so picking flush/coplanar overlaps can select the hidden entity.
- The `edgeContrast` UI knob is a no-op in perspective (infinite reverse-Z makes clip.z constant, so its depth-gradient term ≡ 0); it's only live in ortho.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed shading artifacts at grazing view angles to prevent lighting band-flips and ensure stable flat-shaded appearance.
  * Improved edge/crease detection for contact shading, reducing seam flicker, dashed-line thresholds, and other seam-related visual artifacts.

### DIFF
--- a/.changeset/grazing-normal-sign-stability.md
+++ b/.changeset/grazing-normal-sign-stability.md
@@ -1,0 +1,26 @@
+---
+"@ifc-lite/renderer": patch
+---
+
+Fix grazing-angle shading artifacts: diagonal lighter/darker bands on flat
+walls and slabs, and dashed/broken separation lines along wall corners.
+
+Root cause: the derivative-based flat-shading normal
+(`cross(dpdx(worldPos), dpdy(worldPos))`) is numerically sign-unstable at
+grazing view angles — the hemisphere-ambient and rim-light terms then
+band-flip across large regions of a single flat surface (and on the 1–2 px
+z-hash slivers along entity corners, which rendered as dark dashes). The
+normal's direction is now kept from the screen-space derivatives (preserving
+the coplanar-strip scar-line immunity) while its sign is stabilized by the
+interpolated vertex normal, guarded against missing/near-perpendicular
+vertex normals. The textured shader inherits the fix through its anchored
+derivation.
+
+The separation-lines pass additionally gained a per-axis second-difference
+"crease" gate (3e-4 relative) alongside the existing 5e-4 first-difference
+gate, so depth-continuous wall/wall and floor/wall seams draw consistently
+instead of flickering around the threshold (dashed lines). Coplanar
+continuations stay suppressed: their second difference is bounded by the
+anti-z-fight hash offset (≤2.55e-4). No new texture loads; both changes are
+a few ALU ops — verified flat 60 fps (0 frames >20 ms over 300 forced
+full-effect renders) and zero load-time impact.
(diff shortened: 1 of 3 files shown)


louistrue:
Follow-up commit ed2769dd: the seam pass's first-difference gate is now slope-aware — a depth jump counts as an occlusion edge only when it exceeds 2× the slope on the opposite side of the pixel. Previously, the absolute 5e-4 gate fired on pure in-plane slope near the horizon of long floors/walls (relative slope reaches ~1e-2 per px at grazing), drawing noisy dashed lines across flush coplanar joints — the artifact visible in the latest user screenshots. Verified live at corner / grazing-wall / grazing-floor poses: corner seams draw continuously, grazing dashes gone, zero added texture loads.

PR #1074, opened 2026-06-12
feat(georef): TS↔Rust extraction parity (phase 1)

louistrue:
Stacked on #1073 (which stacks on #1071). Lands the georeferencing drift finding from the alignment audit — the highest-severity confirmed item left.

## The drift

The two live extractors — `rust/core::GeoRefExtractor` (REST/SSE `ModelMetadata.georeferencing`) and `@ifc-lite/parser` `extractGeoreferencing` (viewer panels, Cesium/KMZ, MCP tools) — disagreed four verified ways: an IFC2x3 ePSet model was georeferenced on the server but not in the browser; a site-only model was georeferenced in the browser but not on the server; millimetre-based MapConversions scaled wrong server-side (MapUnit ignored); and files with non-unit `XAxisAbscissa/Ordinate` got a `transform_matrix` that disagreed with `rotation_degrees` *within the same payload*.

## The fix — one superset, identical precedence

`IfcMapConversion` → `ePSet_MapConversion` → legacy `IfcSite` lat/long, on both sides:

| Change | Side |
|---|---|
| First authored conversion wins (was last-wins) | Rust |
| `Description`/`MapZone`/`MapUnit` parsed; `map_unit_scale` via the shared SI-prefix table | Rust |
| X-axis normalised at parse → matrix ⇄ rotation self-consistent | Rust |
| `IfcSite.RefLatitude/RefLongitude` fallback (compound angle → decimal degrees, same eastings/northings packing as TS, `EPSG:4326` pseudo-CRS) | Rust |
| `ePSet_MapConversion` fallback added, placed before the site fallback | TS |
| `source` provenance shared string-for-string (`mapConversion` / `ePSetMapConversion` / `siteLocation`) | both |

Server `Georeferencing` + `@ifc-lite/server-client` gain optional `crs_description` / `map_zone` / `map_unit` / `map_unit_scale` / `source` (serde-default — backward compatible; no breaking shape change).

## Parity guard

Mirrored fixtures pinned on both sides (values cross-referenced in comments): mm MapUnit → 0.001, two conversions → first wins, non-unit (3,4) axis → 53.130° with matching matrix cells, site 47°22'30"/8°32'15" → 47.375/8.5375, ePSet beats site. Rust: 7 georeferencing tests; TS: 171 parser tests (3 new suites).

Deferred to phase 2 (documented in the audit): exposing GeoRefExtractor through wasm-bindings and shrinking the TS extractor to a shim — blocked on cache-restored models lacking source bytes.

### DIFF
--- a/.changeset/georef-parity.md
+++ b/.changeset/georef-parity.md
@@ -1,0 +1,9 @@
+---
+'@ifc-lite/parser': minor
+'@ifc-lite/server-client': minor
+---
+
+Georeferencing TS↔Rust parity (alignment audit phase 1):
+
+- `@ifc-lite/parser`: `extractGeoreferencing` gains the IFC2x3 `ePSet_MapConversion` fallback with the same precedence as the Rust extractor (`IfcMapConversion` → ePSet → legacy `IfcSite` lat/long); `GeoreferenceInfo.source` union widens to include `'ePSetMapConversion'`.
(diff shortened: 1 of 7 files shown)


louistrue:
Consolidated into #1071 (fast-forwarded into the base branch so one clean PR lands on main).

PR #1079, opened 2026-06-12
fix(pointcloud): stream E57 instead of loading whole file into memory

louistrue:
## Summary
- `E57StreamingSource` used to read the entire `.e57` into a single `ArrayBuffer` (plus a CRC-stripped copy) before decoding, so multi-GB scans died with "Array buffer allocation failed" — a ~3.9 GB file even slipped under the 4 GB size gate and crashed on the raw allocation rather than being rejected.
- `open()` now reads only the FileHeader, the XML, and each scan's 32-byte section header. `next()` pulls a bounded (<=16 MiB) logical window from the blob, strips page CRCs on the fly, walks the DataPackets inside it, and applies stride + per-scan pose before emitting each chunk. Peak memory is ~one window plus one output chunk; stride downsampling now happens natively during decode.
- Extracted per-packet primitives (`resolveScanFields`, `decodeE57Packet`, `peekPacketHeader`) into `e57-decode.ts` and shared them with the whole-file `decodeE57Scan` so both paths produce byte-identical output. Added `logicalToPhysical` to `e57-page.ts`.

## Test plan
- [ ] `pnpm --filter @ifc-lite/pointcloud test` — new `e57-source.test.ts` covers colours, multi-chunk emission, both windowing paths (oversized-packet grow + multi-packet straddle re-read), native stride, and the no-scans error
- [ ] Load a multi-GB `.e57` in the viewer and confirm no `ArrayBuffer` allocation failure
- [ ] Spot-check that whole-file vs streaming decode produce identical point output on a small fixture

## Summary by CodeRabbit

* **Bug Fixes**
  * E57 point cloud files now read incrementally in bounded windows instead of loading entirely into memory. Large scans are now supported while maintaining identical output quality and greatly reducing peak memory usage.

### DIFF
--- a/.changeset/streaming-e57-decode.md
+++ b/.changeset/streaming-e57-decode.md
@@ -1,0 +1,24 @@
+---
+"@ifc-lite/pointcloud": minor
+---
+
+Stream E57 point clouds instead of decoding the whole file in memory.
+
+`E57StreamingSource` previously read the entire `.e57` into a single
+`ArrayBuffer` (plus a CRC-stripped copy of it) before decoding, so
+multi-GB scans died with "Array buffer allocation failed" — a ~3.9 GB
+file even slipped under the 4 GB size gate and crashed on the raw
+allocation rather than being rejected.
+
+It now reads the binary CompressedVector section incrementally: `open()`
+reads only the FileHeader, the XML, and each scan's 32-byte section
+header; `next()` pulls a bounded (≤ 16 MiB) logical window from the blob,
+strips page CRCs on the fly, walks the DataPackets inside it, and applies
+stride + per-scan pose before emitting each chunk. Peak memory is now
+~one window plus one output chunk rather than the whole file. Stride
+downsampling is applied natively during decode instead of after a
+full-file decode.
+
+The per-packet decode primitives are shared with the whole-file
+`decodeE57Scan`, so the streaming and in-memory paths produce identical
+output.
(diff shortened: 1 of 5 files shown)


louistrue:
Follow-up commit `7bf84dcf` — fixes a multi-scan striding bug found while reviewing this PR.

**Bug:** the streaming source reset the stride phase to 0 at every scan boundary (`recordsWritten` resets and was passed as the stride base), so a multi-scan E57 opened with `stride > 1` (any file over the 25M-point downsample cap) re-aligned striding *inside each scan* instead of striding the merged cloud globally. That emitted up to `(numScans − 1)` more points than `open()` reported — pushing progress past 100% — and selected different points than `decodeE57`, breaking this PR's "byte-identical to the whole-file decoder" guarantee for multi-scan + stride.

**Fix:** carry a file-global `scanRecordBase` (cumulative declared `recordCount` of prior scans) and phase the stride on `scanRecordBase + recordsWritten`, so the kept set is `{i : i ≡ 0 mod stride}` over the merged cloud — exactly what the whole-file path produces. Single-scan / stride-1 behaviour is unchanged.

**Tests:** the fixture builder previously could only emit a single scan (which is why this went uncaught). Added a multi-scan builder + two regression tests — stride > 1 matches global striding (count *and* positions), and stride 1 stays byte-identical to `decodeE57`. Full `@ifc-lite/pointcloud` suite green (93 passed, 5 skipped).

louistrue:
Superseded by #1400, which folds in this branch's streaming E57 decoder (identical files, incl. the multi-scan file-global stride fix) plus the renderer/viewer point-cloud + deviation fixes. #1400 is merged, so these changes are now on main.

PR #1095, opened 2026-06-12
feat(viewer): unify storey state + relocate level display into the hierarchy

louistrue:
## Why

A tester couldn't form a correct model of the toolbar's **Stacked / Exploded / Solo** button: the trigger was icon-only and the icon encoded *current state* (so it read as "a stack of layers" at 16px), it sat right next to a second storey button (Quick Floorplan), and activating Solo cold-started on the **lowest** storey — often a near-empty basement — which read as "broken." Underneath, three surfaces each kept their **own** idea of "the storey": the hierarchy (`selectedStoreys`), Space Sketch (its private `storeyId`, defaulting to `storeys[0]`), and Level-display Solo (`soloStorey`). Pick a storey in one place and the others ignored it.

## What

A single, model-aware **active storey** is now the source of truth, and the level controls move to where users already think about levels — the hierarchy's **Building Storeys** section.

**Shared state**
- `selectionSlice`: new `activeStorey: EntityRef | null` + `setActiveStorey`. Distinct from the multi-select `selectedStoreys` renderer filter; a single-storey click sets both.

**Every storey-aware surface reads it**
- **Hierarchy** — a storey-row click sets the active storey (resolving the representative `{modelId, expressId}`, including multi-model unified storeys).
- **Space Sketch** — seeds `storeyId` from the active storey (when it belongs to the active model) and *follows* it while open, instead of always `storeys[0]`. The in-panel `<select>` stays a local override.
- **Solo level display** — `useLevelDisplayEffect` targets `activeStorey → soloStorey → lowest-storey fallback`, killing the empty-basement cold-start on the primary UI path.

**Relocated control + retired duplicates**
- New `StoreyDisplayControls` — a `[Stacked | Exploded | Solo]` segmented control + Exploded gap input + a **Floorplan-the-active-storey** button, atop Building Storeys (self-gates on ≥2 storeys). The storey rows *are* the Solo picker; engaging Solo with nothing focused defaults to the **top** storey.
- `MainToolbar` — removed the `LevelDisplayDropdown` and the adjacent Quick-Floorplan dropdown (the two-confusable-buttons problem) plus now-unused imports.

**Persistent viewport signal**
- New `LevelDisplayIndicator` — a top-left chip (`Solo · 1. Geschoss` / `Exploded · 4 m gap`) with one-click return to Stacked, replacing the easy-to-miss 1.5px toolbar dot. Anchored to clear the ViewCube and Sun & Sky panel.

**Also included** (same files/area): Space Sketch Ctrl+Z / Ctrl+Shift+Z now drive the sketch's own undo history rather than the model behind it.

## Behavior change worth noting

The toolbar no longer floorplans an arbitrary storey directly — you pick a storey in the hierarchy (making it active), then hit Floorplan. This is the deliberate "pick once, every tool respects it" model. The command palette (`Level — Stacked/Exploded/Solo`) is unchanged as a power-user fallback.

## Test plan

- `selectionSlice` unit tests extended for `activeStorey` (set / clear / independence from `selectedStoreys`) — `tsx --test` green (31/31).
- `tsc --noEmit` clean on all touched files.
- Manual: click a storey in the hierarchy → Space Sketch opens on it; toggle Solo → isolates the active storey + shows the chip; switch storeys while in Solo → follows; Exploded gap input works; Floorplan button activates the top-down view for the active storey.

## Out of scope / follow-ups

- Skipped the optional per-row "● active" dot in `HierarchyNode` — the existing selection highlight already marks the active storey (kept `HierarchyNode` untouched to limit blast radius).
- `setSoloStorey` is now caller-less (the effect still reads `soloStorey` as a secondary fallback); left the slice API intact rather than widening this PR.

## Summary by CodeRabbit

* **New Features**
  * On-screen level indicator in the viewer showing Exploded gap or Solo storey with quick-dismiss.
  * Storey controls moved into the Hierarchy panel: Stacked / Exploded / Solo modes, adjustable gap, and click-to-solo behavior.

* **Refactor**
  * Toolbar level controls removed; indicator and controls now surfaced via hierarchy/viewport.
  * Level-display handling centralized so solo/stacked/exploded behave consistently across views.

* **Bug Fixes**
  * Active storey resets when loading a new file; overlays now sync to the shared active storey.

* **Tests**
  * Added tests for shared active storey state.

### DIFF
--- a/apps/viewer/src/components/viewer/CommandPalette.tsx
+++ b/apps/viewer/src/components/viewer/CommandPalette.tsx
@@ -69,6 +69,7 @@ import {
 } from 'lucide-react';
 import { cn } from '@/lib/utils';
 import { useViewerStore } from '@/store';
+import { applyLevelDisplayMode } from '@/store/levelDisplay';
 import { goHomeFromStore, resetVisibilityForHomeFromStore } from '@/store/homeView';
 import {
   executeBasketSet,
@@ -327,11 +328,11 @@ export function CommandPalette({ open, onOpenChange }: CommandPaletteProps) {
       { id: 'view:frame', label: 'Frame Selection', keywords: 'zoom focus selected', category: 'View', icon: Crosshair, shortcut: 'F',
         action: () => { useViewerStore.getState().cameraCallbacks.frameSelection?.(); } },
       { id: 'view:stacked', label: 'Level — Stacked', keywords: 'level display mode stacked default storey storeys', category: 'View', icon: Layers3,
-        action: () => { useViewerStore.getState().setLevelDisplayMode('stacked'); } },
+        action: () => { applyLevelDisplayMode('stacked'); } },
       { id: 'view:exploded', label: 'Level — Exploded', keywords: 'level display mode exploded explode lift storey storeys gap', category: 'View', icon: ChevronsUpDown,
-        action: () => { useViewerStore.getState().setLevelDisplayMode('exploded'); } },
-      { id: 'view:solo', label: 'Level — Solo', keywords: 'level display mode solo isolate storey single only', category: 'View', icon: SquareStack,
-        action: () => { useViewerStore.getState().setLevelDisplayMode('solo'); } },
+        action: () => { applyLevelDisplayMode('exploded'); } },
+      { id: 'view:solo', label: 'Level — Solo', keywords: 'level display mode solo isolate storey single only top', category: 'View', icon: SquareStack,
+        action: () => { applyLevelDisplayMode('solo'); } },
       { id: 'view:projection', label: 'Projection', keywords: 'perspective orthographic ortho toggle switch', category: 'View', icon: Orbit,
         action: () => { useViewerStore.getState().toggleProjectionMode(); } },
       { id: 'view:top', label: 'Top View', keywords: 'camera plan', category: 'View', icon: ArrowUp, shortcut: '1',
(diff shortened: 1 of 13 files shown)


louistrue:
Follow-up fix pushed (a84fdce9): **Solo no longer strands the isolation when switching modes.**

Root cause from testing: Solo ran a *second* isolation channel (`isolatedEntities`) on top of the storey filter, and a mode switch only cleared one — so Solo → Exploded left the model stuck on the soloed storey under an 'Exploded' chip.

Fix collapses Solo onto the **one** existing channel (`selectedStoreys`):
- `applyLevelDisplayMode` is the single transition for all entry points — Solo isolates the active/top storey; Stacked/Exploded clear it.
- Clicking a single storey in the hierarchy now **is** Solo (isolating one storey == soloing it).
- A guard resets Solo → Stacked whenever the storey filter is cleared elsewhere (footer ×, Home, Esc).
- Removed the now-dead `soloStorey` slice state.

louistrue:
Addressed review feedback in bb1f882a:

**Fixed (correctness)**
- **[codex P1] Floorplan for single-storey models** — `StoreyDisplayControls` now renders with ≥1 storey; only the Stacked/Exploded/Solo segmented control gates on ≥2. A single-storey model floorplans its sole storey directly (restores the retired Quick-Floorplan behaviour).
- **[codex P2] Stale `activeStorey` for Solo** — `applyLevelDisplayMode` validates the storey still resolves to a loaded model before preferring it (falls back to the top storey), and `resetViewerState` now clears `activeStorey` so a new file can't inherit one.
- **[CodeRabbit] Model-aware Space Sketch sync** — both the active-storey follow ref **and** the auto-derive guard are now keyed by `modelId:expressId`, so switching to another model that shares a storey express-id re-seeds and re-derives instead of showing the previous model's rooms.

**On the test-coverage suggestions (CodeRabbit: `StoreyDisplayControls` UI tests / cross-surface `activeStorey` tests)** — this package's test convention is `node:test` via `tsx --test` over slices and pure helpers; there's no React-render or full-store-singleton harness (importing `@/store` pulls browser/WASM-only deps that don't load under node, which is why existing tests import slice *creators* directly). I've kept the `activeStorey` coverage at that level (`selectionSlice.test.ts`). Component/integration-render coverage would need a new harness — happy to spin that up as a follow-up if you'd like, but it's out of scope for this change.

PR #1099, opened 2026-06-13
fix(geometry): base stream watchdog on batch wall-time, not file size (#1097)

### LINKED ISSUE #1097: Geometry stream stall watchdog kills healthy load on CSG-dense models (regression from #1093)
## Symptom

Large CSG-dense models (e.g. a ~275 MB steel model, 190k+ meshes) fail to open in the viewer with:

```
Geometry stream stalled after 23257ms while loading <model>.ifc. Last rendered meshes: 192793.
```

Reproduces on prod (ifclite.com) **and** locally. Earlier versions loaded the same model. The `23257` ms deadline corresponds to the subsequent-batch watchdog formula `15000 + MB*30`, i.e. the file is ~275 MB.

## Root cause

1. **Watchdog measures the wrong metric.** The subsequent-batch stall watchdog (`packages/geometry/src/watchdog.ts`) scales its deadline with **file size in MB** (`15000 + MB*30`). But the real silent window is the duration of *one* synchronous `STREAM_BATCH_SIZE = 512`-job `processGeometryBatch` call, whose duration scales with **geometric complexity per job (CSG density)** — uncorrelated with MB. Dense steel models pack far more CSG per MB, so a single 512-job batch legitimately exceeds the per-file-size deadline (512 jobs × ~45 ms ≈ 23 s).

2. **Whole pool goes silent together.** Worker liveness heartbeats (`geometry.worker.ts` → `processSliceStreaming`) are posted only *between* 512-job batches. With N workers the watchdog only fires if **all** workers are silent simultaneously — and the interleaved job striding (`geometry-parallel.ts` → `dispatchJobsChunkInternal`, see the "Striding spreads any hot region evenly across the pool" comment) deliberately spreads a hot CSG region across every worker at once. So when the stream reaches a dense CSG cluster, the entire pool enters a long synchronous call at the same time, no heartbeat reaches the host, and the watchdog kills a healthy load.

3. **Regression: #1093 (`c597642f`).** This commit activated per-element material-colour processing (`materialElementIds` / `materialColorCounts` / `materialColors` → trailing args of `processGeometryBatch`, driving the #913 §2.3 transparent/opaque alternation) in the **browser** batch path, where it was previously inert (empty map). The extra per-job work and additional submeshes raised the 512-batch wall-time, tipping previously-passing dense models past the watchdog.

4. **Secondary silent window.** `processBatch` binary-split recovery (`geometry.worker.ts:551-570`) sets `api = null` (forcing a WASM re-init) and recurses **without emitting any heartbeat**. A cluster of failing entities can therefore also blow the budget silently.

## To confirm exact branch

In the console immediately before the `stalled` error, is there a burst of:

```
[Worker] Batch of N entities failed (...), splitting…
[Worker] Skipping entity #...
```

- **Present** → recovery storm (a correctness regression in #1093 material threading making entities throw).
- **Absent** → a legitimately slow batch (the watchdog-metric design flaw above).

The watchdog-metric flaw + #1093 cost increase hold regardless of which branch this particular model hits.

## Fix directions (not yet applied)

- Base the watchdog on actual liveness rather than MB: treat the per-batch heartbeat's `processedJobs` advance as the liveness signal, or scale the deadline by observed ms/job (mesh throughput) instead of file size.
- Shrink/adapt `STREAM_BATCH_SIZE`, or bound a single batch's wall-time, so the maximum silent window stays small on CSG-heavy regions.
- Emit a heartbeat inside `processBatch` before each recursive split / WASM re-init to close the recovery silent window.
- Verify #1093's material processing didn't introduce per-entity WASM throws on steel models.

## Env

- Viewer `1.30.0`, prod + local (Windows).
- Observed watchdog deadline `23257 ms` ⇒ ~275 MB file, `Last rendered meshes: 192793`.

Image

louistrue:
Fixes #1097.

## Problem

Large CSG-dense models (the reported case: ~275 MB steel, 190k+ meshes) fail to open with `Geometry stream stalled after 23257ms`. The mid-stream stall watchdog scaled its deadline with **file size** (`15000 + MB*30`), but the real silent window is the wall-time of **one synchronous `processGeometryBatch` call**, which tracks **CSG density per job** — uncorrelated with megabytes. A fixed `STREAM_BATCH_SIZE = 512` meant one batch on dense steel took ~23 s (512 × ~45 ms), and interleaved job-striding pushes the whole worker pool into a hot region at once, so no heartbeat reaches the host for that window. #1093 raised per-batch wall-time (it activated per-element material-colour work in the browser path), tipping previously-passing models past the deadline.

## Fix

**1. Adaptive batch sizing** (`batch-sizing.ts` + `geometry.worker.ts`) — the core fix. Each worker now sizes its `processGeometryBatch` call to a fixed **wall-time** budget (`TARGET_BATCH_MS = 2 s`), resizing from measured throughput and clamped `[64, 256]`. Light regions grow toward MAX (good decoder-cache locality, few calls); dense regions collapse to MIN within one batch (small, regular heartbeats). Size persists across chunks so density learned once sticks. This bounds the steady-state silent window to ~2 s regardless of density or file size. Extracted as a pure module so the formula stays unit-tested, mirroring `watchdog.ts`.

**2. Watchdog measures the right metric** (`watchdog.ts`) — the subsequent-batch deadline is now a **fixed grace** (30 s browser / 20 s desktop), *not* MB-scaled. The first-batch deadline still scales with size (the single-threaded pre-pass genuinely grows with bytes). The grace gives >10× headroom over the bounded ~2 s call, plus margin for the lone transitional call that first crosses into a dense region at MAX size (256 × 45 ms ≈ 11.5 s, safe to ~117 ms/job).

**3. Recovery silent window closed** (`geometry.worker.ts`) — `processBatch`'s binary-split recovery now emits a liveness heartbeat before it recurses / re-inits WASM, so a cluster of failing entities can't blow the budget invisibly (addresses the issue's "recovery storm" branch).

**4. Entity-index replay on re-init** (`geometry.worker.ts`) — recovery sets `api = null` to force a WASM re-init, which previously *lost* the pre-built entity index, so the next call fell back to the lazy O(file) re-scan (~5 s) — itself a giant silent window and perf cliff. The index now replays on `ensureInit`, mirroring the existing merge-layers / tessellation flag pattern. This bounds the recovery path regardless of whether #1093 introduced per-entity throws.

## Behaviour change

| Scenario | Before (subsequent deadline) | After |
|---|---|---|
| 5 MB file | 15.15 s | 30 s (more lenient) |
| 275 MB dense steel | 23.25 s → **killed** | 30 s grace **+ ~2 s real silent window** → loads |
| 4 GB file | 138 s (masked the design flaw) | 30 s (safe: batches now bounded ~2 s) |

No deadline is ever *shorter* than before for small/medium files. The 4 GB case is shorter only because mega-batches no longer exist — a single batch can't approach 30 s once adaptive sizing is in effect.

## Verification

- `pnpm --filter @ifc-lite/geometry test`: **66/66 pass** (new `batch-sizing.test.ts`; `watchdog.test.ts` updated to assert the subsequent deadline is size-independent).
- `tsc --noEmit` clean; geometry package build emits `batch-sizing.js`.
- `check-test-wiring` and public API surface unchanged (`batch-sizing` is worker-internal, not re-exported).

On the reported model the deadline goes 23.25 s → 30 s **and** the per-batch silent window drops from ~23 s to ~2 s, so the load completes with large margin.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the geometry stream watchdog from prematurely terminating healthy, CSG‑dense loads by using a fixed grace window for subsequent batches and adding immediate liveness heartbeats during recovery.

* **Improvements**
  * Adaptive per-batch sizing driven by wall‑time budgets for smoother large-model streaming.
  * Reduced work and copies via cached style data, moving meshes out of collections, and inline color resolution.
  * More robust triangulation and refined handling for opening‑dense geometry.

* **New Features**
  * Load‑time visibility filter and runtime overrides to skip disabled types and tune batch sizing.

* **Tests**
  * Added/updated tests for adaptive sizing, watchdog timing, and benchmark parsing.


louistrue:
## Performance follow-up — retuned + benchmarked on the largest models

The first cut used a 2 s / 256-job adaptive budget, which **loaded** dense models but was slow: it collapsed dense regions to the floor (~64-job batches), re-paying decoder-cache setup + re-decoding shared sub-entities per call. Retuned to an **8 s target / 512-job cap** and validated with the (now repaired) `viewer-benchmark` Playwright harness on real Chrome + WebGPU (10-core macOS):

| Model | 2 s/256 (first cut) | **8 s/512 (default)** | main baseline |
|---|---|---|---|
| Holter 169 MB (109 k meshes) | 10.0 s | **8.9 s** | 11.0 s |
| bouwkundig 327 MB (55 k meshes) | 9.4 s | 10.4 s | 11.9 s |
| merged_export **986 MB** (14 M entities, 232 k meshes) | **100.9 s** | **72.5 s** | crashed |

- **−40% on the 986 MB model** (100.9 s → 72.5 s); first-visible geometry 11 s; **no watchdog stall**.
- Mid-size models are at or below the pre-fix `main` baseline.
- A larger target (16 s) was rejected: it keeps batches at the 512 cap in moderately-dense regions, enlarging the lone light→dense transition call the watchdog can't pre-empt. 8 s keeps dense-region batches ~200 jobs so transitions stay bounded; the subsequent grace was raised to 40 s (browser) for headroom on slower hardware.

### Tuning hook
`ProcessParallelOptions.batchSizing` (programmatic) or `globalThis.__IFC_LITE_BATCH_SIZING = { targetMs, minJobs, maxJobs }` (field/debug) override the defaults, forwarded to the worker pool in `stream-start` and validated via `resolveBatchSizing`.

### Harness
`viewer-benchmark.spec.ts` now accepts absolute model paths, detects the current viewer's end-of-load marker (`[ifc-lite] … in Xs`), parses the app's own total, and honours `VIEWER_BENCHMARK_BATCH_SIZING` for config sweeps:
```
VIEWER_BENCHMARK_FILES="/abs/model.ifc" \
VIEWER_BENCHMARK_BATCH_SIZING='{"targetMs":8000,"minJobs":64,"maxJobs":512}' \
pnpm exec playwright test --project=viewer-benchmark
```

louistrue:
**CodeRabbit findings addressed:**
- ✅ *Clear entity-index cache at session teardown* — applied in ed170506: `cachedEntityIndex`/`entityIndexApplied` reset on `stream-end`.
- ⏭️ *MPL header on the changeset* — skipped: changeset markdown in this repo is bare frontmatter (no other `.changeset/*.md` carries the MPL header) and is consumed/deleted on the next version bump.

louistrue:
Addressed: `cached_geometry_styles` is now cleared in `set_entity_index` too (commit), matching the other content caches.

PR #1111, opened 2026-06-14
fix(geometry): drop degenerate f32 triangles to kill gross fan corruption

louistrue:
## Problem

Large **georeferenced** models render gross triangle "fans" that span the whole building — visible as thin white sheets cutting across the geometry. This is **pre-existing on `main`** and affects **both** the native and WASM/viewer pipelines (they share `produce_element_meshes`).

### Root cause: f32 vertex-storage collapse

Element meshes are stored at **f32** precision, but their vertices sit at **building-scale world coordinates** (a model whose extent reaches ~220 m from the coordinate origin). At that magnitude an f32 mantissa resolves only **~15 µm**. Two genuinely-distinct vertices closer together than one ULP round to the *same* (or near-same) f32 value, so the triangle joining them collapses into a **zero-area sliver** — and when its third vertex is far away, a long thin triangle that fans across the model.

This is the storage step, not the kernel: the f64 geometry is correct; the corruption appears only once it's cast to f32 at world scale.

## Fix (backstop)

`Mesh::drop_degenerate_triangles` runs in `build_mesh_data` — the **single funnel** every element `MeshData` passes through on both pipelines — and removes only **unambiguously-degenerate** triangles:

- a **bit-identical f32 vertex pair** (exact zero area), or
- an **aspect ratio > 1e5** (longest edge / shortest edge).

These slivers carry no area, so the neighbouring triangles of the same face already cover the surface — the removal is **visually lossless**. Only `indices` change; the vertex buffer is untouched, so the op is deterministic and keeps vertex indices stable.

An operability escape hatch — `IFC_LITE_DISABLE_DEGENERATE_BACKSTOP=1` (native only; read once) — keeps the raw triangles for debugging/measuring the heuristic.

### Corpus-wide (28 models, backstop off vs on)

Every model: **`fans_on` = 0**. No fix failures, no over-aggressive drops.

- **12 models "clean"** — no fans either way, **0 triangles dropped** → proves the backstop is a no-op on healthy/normal-scale geometry.
- **16 models "fixed"** — fan corruption was far more widespread than just one model. Examples:

| model | fans (raw) | fans (fixed) | triangles dropped | drop % |
|---|---|---|---|---|
| ISSUE_053 (Holter Tower) | 7966 | 0 | 7966 / 2.33 M | 0.34% |
| dental_clinic | 1316 | 0 | 796 / 168 k | 0.47% |
| ISSUE_129 | 1040 | 0 | 981 / 128 k | 0.77% |
| S_Office | 790 | 0 | 790 / 217 k | 0.36% |
| ISSUE_021 | 728 | 0 | 728 / 252 k | 0.29% |
| (target model) | 664 | 0 | 650 / 222 k | 0.29% |
| schependomlaan | 230 | 0 | 230 / 135 k | 0.17% |
| ISSUE_102_M3D-CON-CD | 140 | 0 | 140 / 2.29 M | 0.006% |

(Smallest model `example.ifc` drops 6.66% — but it's a tiny 12 k-triangle model that is badly georeferenced, and every dropped triangle was a genuine aspect>1e5 fan.)

### Determinism + suites
- Predicate-sign **manifests unchanged** (kernel determinism intact)
- **`geometry_correctness_harness` byte-identical** — synthetic small-coordinate geometry never collapses, so the backstop is a no-op there
- Full `ifc-lite-geometry` + `ifc-lite-processing` suites pass
- WASM rebuilds clean; binding surface (`.d.ts`) unchanged

### Visual
Loaded the 54 MB target model in the real-Chrome / WebGPU viewer (headed Playwright): the full building complex now renders clean — the building-spanning fans are gone, geometry intact, 60 fps.

## Scope / follow-up

This is the **backstop**. The complete fix — local-frame / tiled vertex storage that keeps the vertices distinct in the first place (so nothing collapses, at any model scale) — is a larger multi-stage effort (Rust origin threading through every CSG path → WASM/worker/types → camera-relative WebGPU renderer → cache bump) and is tracked separately. This PR keeps the viewer clean meanwhile.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed geometry rendering artifacts in large georeferenced models caused by degenerate triangles. The system now filters out unambiguously-degenerate zero-area and extremely thin “sliver” triangles during mesh data building to prevent visible “fan” distortion while preserving intended surface coverage.
  * The behavior is enabled by default and can be turned off by setting `IFC_LITE_DISABLE_DEGENERATE_BACKSTOP`.

### DIFF
--- a/.changeset/geometry-degenerate-f32-triangle-backstop.md
+++ b/.changeset/geometry-degenerate-f32-triangle-backstop.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Drop degenerate f32 triangles so large georeferenced models stop showing gross "fan" corruption.
+
+When a mesh is stored at f32 precision while its vertices sit at building-scale world coordinates (e.g. a model whose extent reaches ~220 m from the coordinate origin), the f32 mantissa only resolves ~15 µm there. Vertices closer together than one ULP round to the same — or near-same — f32 value, so the triangles joining them collapse into zero-area slivers; when the third vertex is far away the result is a long, thin triangle that visibly fans across the whole model.
+
+`Mesh::drop_degenerate_triangles` now runs in `build_mesh_data` — the single funnel every element `MeshData` passes through on both the native and WASM pipelines — and removes only unambiguously-degenerate triangles: a bit-identical f32 vertex pair (exact zero area) or an aspect ratio above 1e5. These slivers carry no area, so neighbouring triangles of the same face already cover the surface and the removal is visually lossless. On a 54 MB georeferenced building model this drops all 664 catastrophic fans (0.29% of triangles) with no change to the remaining geometry, no kernel-determinism impact (predicate-sign manifests unchanged), and the synthetic-coordinate correctness harness stays byte-identical. The complete fix (local-frame / tiled vertex storage that keeps the vertices distinct) is tracked separately; this is the backstop that keeps the viewer clean meanwhile.
--- a/rust/geometry/src/mesh.rs
+++ b/rust/geometry/src/mesh.rs
@@ -426,6 +426,85 @@ impl Mesh {
         self.indices = valid;
     }
 
+    /// Drop triangles that collapsed into degenerate needles when the mesh was
+    /// stored at f32 precision.
+    ///
+    /// At building-scale world coordinates (e.g. ~220 m) an f32 mantissa only
+    /// resolves ~15 µm, so two genuinely-distinct vertices less than one ULP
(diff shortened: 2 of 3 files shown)


louistrue:
Superseded by #1113. That PR replaces this aspect-ratio backstop with a height-based `drop_thin_triangles` at the kernel reconcile grid (15.3 µm), applied at every mesh-output chokepoint. The height test catches the collinear slivers this aspect-ratio drop missed (genuine degenerate slivers 861 → 0 vs the band this missed), preserves flat shading (no welding), and is bit-deterministic. Closing in favour of #1113.

PR #1112, opened 2026-06-14
fix(geometry): deterministic escalation budget to stop the exact-CSG 95% hang (#1109)

### LINKED ISSUE #1109: Regression: exact CSG kernel hangs at 95% on boolean-heavy models that loaded in ~5 s before
### Summary
Models that load in ~5 s on `@ifc-lite/geometry@2.4.0` no longer finish loading on the current release — the loader hangs at **95%** indefinitely (no crash, no completion). Reproduces on **ifclite.com** and a local build, in **both Edge and Chrome** (Chromium, latest stable). Pinning `@ifc-lite/geometry` back to `2.4.0` (Manifold kernel) restores the ~5 s load.

This is **not tool-specific** — it reproduces on any boolean-heavy model, from **both Revit** (cuts via `IfcOpeningElement`) **and Tekla** (cuts via `IfcBooleanClippingResult` / half-spaces).

### Regression range
- **Good:** `@ifc-lite/geometry@2.4.0` (Manifold C++ kernel) — ~5 s.
- **Bad:** `@ifc-lite/geometry@2.6.1` / `@ifc-lite/wasm@2.8.0` — hangs at 95%.
- Prime suspect: **#1024** (`cdc5a78`) — *"One CSG kernel: pure-Rust exact mesh arrangement"* (Manifold → exact-arithmetic Cherchi-style predicates). Possibly compounded by **#1005 / #977** (2.5.1, routing more cuts through the real-mesh boolean).

### Symptom
The 95% point is the geometry / CSG stage — the exact-arithmetic kernel appears dramatically slower or stuck on the boolean arrangement. Any model with a meaningful number of boolean cuts hangs.

### Concrete example
- Tekla Structures 2017, IFC2X3, 8.4 MB, **101,776 entities**
- **1,992 `IfcBooleanResult`**, **180 `IfcBooleanClippingResult`**, **180 `IfcHalfSpaceSolid`**, ~4,000 steel members (2,600 `IfcBeam`, 1,066 `IfcMember`, 252 `IfcColumn`, 80 `IfcPlate`)
- 0 `IfcOpeningElement` — all cuts are half-space booleans.
- A Revit model (cuts via `IfcOpeningElement`) reproduces the same hang.

### Environment
- Browser: Edge & Chrome (Chromium), latest stable
- OS: Windows 10
- Multithreaded WASM (SharedArrayBuffer), cross-origin isolated

### Repro
Open a boolean-heavy model in the web viewer → loader reaches 95% and never completes. Models are from real projects; can be shared privately with a maintainer on request.

Image

louistrue:
Fixes #1109.

## Root cause

The M9 flip (#1024) replaced Manifold (viewer) + the legacy BSP port (server) with **one pure-Rust exact kernel** — bit-deterministic across x86_64 / aarch64 / wasm32. That was the right call: clients run a **native Rust server *and* the wasm viewer** and need matching results; the two old kernels diverged ("square holes on server"). The determinism is real — `SNAP_GRID = 1/2^16` is a power of two precisely so the snap is a bit-exact f64 op across targets.

What the flip dropped was Manifold's/BSP's **operand cap**. Every predicate over an implicit (intersection) point runs `interval filter → fixed-width exact → BigRational`. On near-coplanar / grazing faces the interval sign straddles zero, so the predicate falls through to the exact tiers (the fixed-width rungs climb to ~1340 bits, a few into BigRational at ~5000×). Boolean-heavy models (half-space end-clips, flush openings) are full of such faces, so a single hard element runs the exact path on a huge fraction of its `O(intersection-pairs)` predicate set — with **no safety valve**. The geometry stream never finishes; the loader stalls at 95%.

@Blogbotana's diagnosis on the issue is spot-on. Two of its suggested options would have broken the parity the flip exists to protect, so this PR takes the parity-safe path:
- ❌ **wall-clock budget** — bails at *different points* on fast native vs slow wasm → different geometry for the same model.
- ❌ **fast float path as default** — risks reintroducing the BSP "square holes" divergence.

## Fix — a deterministic per-boolean budget

`kernel::budget` counts **interval-filter failures** (every predicate that needs the expensive exact tier) per boolean and, when the count crosses a cap, bails the boolean to the un-cut host — routing void subtraction to the existing #635 AABB box-cut fallback. The count is a pure function of the snap-grid operands, so the trip point is **identical on native and wasm**: server and client degrade the *same* hard element to the *same* fallback. Parity preserved.

- `budget.rs` — per-op escalation counter, cap, peak (calibration), `set_cap` knob.
- `predicates.rs` — count at each interval-fail (the `.or_else(fixed)` entry to the exact path).
- `arrangement.rs` — bail the pair / retriangulation loops once tripped (timely, graceful stop).
- `csg.rs` — `begin()` per boolean; on trip record `OperandTooLarge` and return the host un-cut.

### Calibration & safety
The cap (`DEFAULT_CAP = 500_000`) is **33× the worst healthy boolean** measured across the model corpus (~15k exact evaluations; most far lower), so it never false-trips a legitimate cut — healthy models are **byte-identical** (determinism manifests unchanged). `set_cap(None)` / `IFC_LITE_CSG_BUDGET=0` lifts it entirely for the **server/offline-export profile** where "exact but slow" is acceptable — one code path, two profiles, no kernel fork.

### Verification
- Budget mechanism unit test (count / cap / trip / reset / unbounded).
- Wiring fires on real geometry — corpus per-boolean exact-eval peaks 4k–15k (the same predicates feed the budget).
- End-to-end trip → host-un-cut → `OperandTooLarge` recorded (near-coplanar fixture).
- Full `ifc-lite-geometry` + `ifc-lite-processing` suites green; manifests unchanged.

## Scope / follow-ups (next phase, needs the repro model)

This stops the **hang** (the issue's core ask: "always finishes"). Two root improvements are designed but deliberately **not** in this PR because they change CSG output (re-pin every determinism manifest + snapshot) and their benefit can't be validated without a hanging model:
1. **Coplanar-cluster fast path** — so hard-but-sound cuts stay *fast and exact* instead of degrading to the AABB fallback (the issue's "load-time parity with Manifold"). The corpus doesn't hang, so the speedup is unmeasurable here — best landed against a real repro model.
2. **Output vertex welding** (restoring Manifold's welded output) — a root fix for the separate f32 "fan" corruption; today covered by the drop-degenerate backstop.

@Blogbotana — could you share a reproducing model privately? It's the missing piece to land + measure (1) safely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed kernel hangs on boolean-heavy 3D models by implementing a deterministic evaluation budget. When the budget is exceeded, operations gracefully fall back to an alternative cutting method instead of hanging.

* **Tests**
  * Added regression tests to verify budget escalation behavior and fallback mechanism.

### DIFF
--- a/.changeset/csg-deterministic-escalation-budget.md
+++ b/.changeset/csg-deterministic-escalation-budget.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/geometry": minor
+---
+
+Fix the exact CSG kernel hanging at 95% on boolean-heavy models (issue #1109), without sacrificing the cross-target determinism the kernel exists to guarantee.
+
+The pure-Rust exact kernel (#1024) replaced Manifold + the legacy BSP port with one bit-deterministic kernel — the right call for server↔client parity (clients run a native Rust server *and* the wasm viewer and need matching results). But the flip dropped Manifold's/BSP's operand cap, so a boolean-heavy model (Tekla half-space end-clips, Revit flush openings — full of near-coplanar faces) drives the exact predicate cascade off its interval filter on a huge fraction of predicates, climbing the fixed-width rungs (to ~1340 bits) and into BigRational with no safety valve. The geometry stream never finishes; the loader stalls at 95%.
+
+This adds a **deterministic** per-boolean budget: it counts interval-filter failures (every predicate that needs the expensive exact tier) and, when the count crosses a cap, bails the boolean to the un-cut host so the existing #635 AABB box-cut fallback fires. The count is a pure function of the snap-grid operands, so the trip point is identical on native x86_64/aarch64 and wasm32 — the server and the browser degrade the *same* hard element to the *same* fallback. A wall-clock budget would have broken parity (fast native finishes the exact cut while slow wasm trips), so the metric is deliberately platform-independent.
+
+The cap (`budget::DEFAULT_CAP = 500_000`) is calibrated 33× above the worst healthy boolean measured across the model corpus (~15k exact evaluations), so it never false-trips a legitimate cut; healthy models are byte-identical (determinism manifests unchanged). `budget::set_cap(None)` (or `IFC_LITE_CSG_BUDGET=0`) lifts it for the server/offline-export profile where "exact but slow" is acceptable — one code path, two profiles, no kernel fork.
--- a/rust/geometry/src/csg.rs
+++ b/rust/geometry/src/csg.rs
@@ -620,7 +620,24 @@ impl ClippingProcessor {
         // output carries >50:1 needle fragments that consolidation legitimately
         // merges (the pinned `csg_quality_regression` spike bar). A
         // seam-preserving consolidation is the remaining follow-up.
+        crate::kernel::budget::begin();
         let raw = crate::kernel::mesh_bridge::subtract(host_mesh, opening_mesh);
+        // Deterministic escalation guardrail (#1109): if the exact predicate
+        // cascade escalated past the per-boolean budget, the cut bailed mid-
+        // arrangement. Discard the partial result and return the host un-cut so
+        // the void router's #635 AABB box-cut fallback fires. The trip point is a
+        // pure function of the snapped operands, so server (native) and client
+        // (wasm) degrade the SAME element identically — parity preserved.
+        if crate::kernel::budget::tripped() {
+            self.record_failure(
+                BoolOp::Difference,
+                BoolFailureReason::OperandTooLarge {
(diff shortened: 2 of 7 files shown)


Blogbotana:
Yes, I will send you the model now, where I should write to you? In general, I like the idea of ​​us creating some kind of Discord channel for this library, as is customary in other open communities 👍

Blogbotana:
I made a build on this branch and my local application freezes at the moment of geometry initialization.

Issue #846, opened 2026-05-28
IfcBeam geometry error with IfcRevolvedAreaSolid


xyzbety:
In the attached model（beam-varying-extrusion-paths）,entity IfcBeam #227 renders incorrectly when defined by IfcRevolvedAreaSolid within a SweptSolid representation, likely due to incomplete support for non-extruded beam geometries, referring to the following comparison images.

[beam-varying-extrusion-paths.zip](https://github.com/user-attachments/files/28332080/beam-varying-extrusion-paths.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/e586e33b-9428-4e0a-9464-99bc68719f8e" />

<img width="2880" height="1722" alt="Image" src="https://github.com/user-attachments/assets/429c7546-0cdf-4dfa-840f-49419f4b37d7" />

PR #1113, opened 2026-06-14
fix(geometry): drop sub-grid sliver triangles to kill spike/jagged rendering

louistrue:
## Problem

After #1024 replaced Manifold with the pure-Rust CSG kernel, the geometry pipeline lost the implicit degenerate-output cleanup Manifold did on import. Faceted breps, extrusion-profile walls, and walls with openings started rendering visible needle **spikes** and jagged silhouettes — zero-area / collinear sliver triangles that other viewers don't show because they clean degenerates on import.

## Root cause (validated by data, not screenshots)

The spikes are genuine **degenerate triangles**: coincident-pair needles from duplicated profile/brep vertices, and collinear slivers from redundant collinear vertices, plus a few introduced by CSG opening cuts. Using a scale-aware damage metric with a *local-frame control* (recentre each mesh to its centroid to separate genuine degeneracy from f32-world-storage):

| | genuine degenerate slivers | triangles | geom time |
|---|---|---|---|
| before | **861** | 222,042 | 0.82 s |
| after | **0** | 219,744 (−1.0%) | 0.78 s |

A threshold sweep showed the triangle count is **flat from 10–50 µm** (no real geometry lives there) and only starts removing real geometry at ~100 µm — a clean gap. The kernel's own reconcile grid (`1/65536 m ≈ 15.3 µm`) sits in that dead zone.

The residual f32-world-storage damage (real thin triangles at a few hundred metres of georef offset) is a **separate** RTC/local-frame concern and is deliberately left untouched.

## Fix

`Mesh::drop_thin_triangles(h_eps)` drops triangles whose perpendicular height (`2·area / longest_edge`) is below `h_eps`; `Mesh::clean_degenerate()` calls it at the kernel grid. Applied at **every** mesh-output chokepoint so wasm (viewer) and native (server) get identical output:
- `process_element` (combined mesh) and `process_element_submeshes` (per sub-mesh)
- the type-geometry / textured `RepresentationMap` channel
- the layered material-slice gateway (`try_layered_sub_meshes`)
- both void-cut outputs

It **does not weld vertices** — the pipeline emits per-face flat-shaded facet soup on purpose (#846); welding would re-smooth creases. It edits only indices (positions/normals untouched), so it is visually lossless, preserves flat shading, and is bit-deterministic (FMA-free f64 threshold).

## Why a post-pass drop (not an upstream tessellation fix)

This is the layer Manifold operated at, and slivers arrive from many sources (brep faces, extrusion profiles, CSG cuts). One drop at the output chokepoints covers them uniformly — fixing each tessellator separately is the parallel-paths trap. Investigating *why* specific tessellators emit redundant collinear vertices is a reasonable follow-up.

## Testing

- All **53 geometry test binaries pass** (446 tests), incl. watertightness, CSG quality, void production/submesh, wall-opening-cut, and the #846 revolved-beam crease test.
- **7 new unit tests** for `drop_thin_triangles` / `clean_degenerate`: collinear sliver, coincident-pair needle, thin-but-real kept, no-crack-in-a-closed-solid, OOB/collapsed skip, idempotence, grid threshold.
- 2 insta snapshots re-pinned (both spike-count improvements; surface area lossless). The weekly determinism job re-asserts these byte-for-byte on arm64.

The change was adversarially reviewed across 5 lenses (math/edge-cases, cross-arch determinism, watertightness, "is it a band-aid?", path coverage); the two coverage gaps it surfaced (textured channel, layered walls) are fixed here.

## Known limitation (documented in code)

The void-cut output is cleaned in world coordinates (the cut runs there), so on a model georeferenced a few hundred metres–10 km from origin (below the RTC re-basing threshold) the 15.3 µm margin near opening seams erodes slightly; the `longest <= 0` guard still catches full collapse at extreme scale. The non-void tessellation path cleans in local frame (precise). Tracked alongside the RTC/local-frame work.

## Relationship to other PRs

- **Supersedes #1111** — this height-based drop replaces the aspect-ratio backstop (which missed collinear slivers).
- **Orthogonal to #1112** — that fixes the #1109 escalation hang (performance), not slivers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed faceted geometry rendering spikes caused by degenerate "sliver triangles" from collinear or zero-area mesh elements.
  * Degenerate triangles below the CSG reconcile grid threshold are now removed across all geometry processing pathways, while preserving vertex data for sharp creases and consistent results.

### DIFF
--- a/.changeset/geometry-drop-sliver-triangles.md
+++ b/.changeset/geometry-drop-sliver-triangles.md
@@ -1,0 +1,21 @@
+---
+"@ifc-lite/geometry": patch
+"@ifc-lite/wasm": patch
+---
+
+Drop sub-grid sliver triangles so faceted geometry stops rendering spikes
+
+After the pure-Rust CSG kernel replaced Manifold (#1024), the pipeline no longer
+cleaned the degenerate output Manifold used to remove on import. Faceted breps,
+extrusion-profile walls and walls with openings could therefore render visible
+needle "spikes" and jagged silhouettes coming from zero-area / collinear sliver
+triangles (other viewers don't show them because they clean degenerates on import).
+
+`Mesh::clean_degenerate` now drops triangles whose perpendicular height is below the
+kernel's reconcile grid (1/65536 m ≈ 15.3 µm) — sub-resolution coincident-pair and
+collinear slivers that carry no area. It runs at every mesh-output chokepoint
+(per element, per sub-mesh, and on the void-cut output), so both wasm (viewer) and
+native (server) get identical output. Vertices and normals are left untouched, so
+flat shading / sharp creases are preserved and the result is bit-deterministic. On a
+large faceted-brep building this removes 100% of the genuine degenerate slivers for a
+~1% triangle reduction with no performance cost.
--- a/rust/geometry/src/mesh.rs
+++ b/rust/geometry/src/mesh.rs
@@ -532,6 +532,101 @@ impl Mesh {
         weld_impl(self, position_eps, None, /*average_normals=*/ true)
     }
 
+    /// Drop triangles whose perpendicular height (= 2·area / longest edge) is
+    /// below `h_eps` metres — i.e. genuinely-degenerate **collinear** slivers
+    /// (three distinct but near-collinear vertices, zero area). These come from
+    /// redundant collinear vertices in source brep faces / extrusion profiles
+    /// triangulated as-is; vertex welding can't merge them (the vertices are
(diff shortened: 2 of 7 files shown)


louistrue:
Superseded by #1114 (per-element local frame). The sub-grid sliver hygiene from this PR is bundled there: the local frame and `clean_degenerate` are complementary and only **together** remove all the fans the viewer actually showed (the local frame kills the f32-storage fans; hygiene removes the sub-grid slivers the finer-grained CSG host then emits). The drop-sliver changeset is carried in #1114.

PR #1121, opened 2026-06-14
fix(geometry): per-element CSG escalation budget to finish boolean-heavy models (#1109)

louistrue:
## Why #1109 still hangs after #1112

The deterministic **per-boolean** escalation budget (#1112) was the right idea but bounded only a *single* boolean. Profiling a real boolean-heavy structural model (ISSUE_129) **with #1112's 500k budget active** showed two remaining holes:

| element | time (before) | worst-boolean escalations | openings |
|---------|---------------|---------------------------|----------|
| `id=333923` IFCSLAB | **243,587 ms (4 min)** | 7,676,200 — **15× the cap** | 39 |
| `id=12381` IFCSLAB | 103,409 ms | 3,051,091 | 47 |
| `id=9094` IFCSLAB | 32,502 ms | 1,665,625 | 24 |
| `id=7526` IFCSLAB | 5,844 ms | 44,896 (**never tripped**) | 29 |

1. **Overshoot.** `budget::tripped()` was only checked at arrangement loop boundaries — once *per triangle* in seam retriangulation. A single heavily-fragmented host face (a slab cut by 24–47 openings) inserts thousands of constraint points in **one** `triangulate()` call, so a boolean ran to ~7.7M escalations (15× the 500k cap) before bailing.
2. **Distributed cost.** An element with many openings runs one boolean *per* opening, each under the per-boolean cap so none trips — yet the element's total exact work blows the geometry-stream watchdog.

(Note: the BigRational tier is *never* hit — the cost is the fixed-width exact tier + retriangulation, not arbitrary precision. The issue's stated "escalates to BigRational en masse" theory doesn't hold.)

## Fix

- **Per-element budget** (`kernel::budget`): `begin_element()` is called once per element at the unified `produce_element_meshes` entry (native **and** wasm). It accumulates escalations across every boolean the element issues; at `DEFAULT_ELEMENT_CAP = 100_000` the element degrades **as a whole** — remaining cuts bail to the existing #635 AABB box-cut fallback.
- **Overshoot guards** (`kernel::retriangulate`): `tripped()` checks inside the per-point insertion and constraint-recovery loops, so a single boolean can no longer overshoot the cap by 15×.

### Parity preserved
Still a **deterministic count**, accumulated in deterministic per-opening order on the element's single worker thread (the kernel has no internal rayon), so native x86_64/aarch64 and wasm32 degrade the **same** element identically — the cross-target parity the pure-Rust kernel exists to guarantee.

### No false-trips, snapshots unchanged
Calibrated against the corpus: healthy per-element totals are **p99 ≈ 13k** escalations, so the 100k cap (~8× p99) never trips a legitimate cut. The cap engages **only** when an element scope is opened (the batch path); direct kernel / router / server / offline-export callers stay unbounded via the existing `set_cap(None)` / `IFC_LITE_CSG_BUDGET=0` switch — so the pinned determinism manifests are byte-identical.

## Results

- **Worst element 243 s → 2.9 s** (83×); total serial geometry minutes → ~28 s.
- Geometry suite **54 groups / 0 failures** (snapshots, determinism, CSG quality, wall-opening cuts, segmented-roof, new budget unit test). Processing suite **17 / 0**.

## Notes
- `rust/geometry/tests/csg_model_profile.rs` is an `#[ignore]`'d calibration harness (`IFCLT_MODEL=… cargo test … --ignored`) documenting how the cap was derived.
- This is a follow-up to #1112; not the f32 local-frame work (coordinate magnitude was empirically shown to have **no** effect on CSG cost).

## Release Notes

* **New Features**
  * Added deterministic per-element CSG exact-evaluation budgeting to cap escalation across all operations within an element.

  * Faster constraint-recovery retriangulation for dense-opening models.
  * Reduced costly exact checks during deterministic ear finding.
  * Improved early stopping during triangulation/recovery once budgets are exceeded.

* **Bug Fixes**
  * Prevented pathological escalation stalls and overshoot on complex elements with many booleans.

* **Tests**
  * Added calibration and issue-specific tests for per-element budgeting and budget-bailed output validity.

* **Chores**
  * Pinned the wasm build action to a fixed version.

### DIFF
--- a/.changeset/csg-1109-per-element-budget.md
+++ b/.changeset/csg-1109-per-element-budget.md
@@ -1,0 +1,16 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Stop boolean-heavy models still hanging at 95% after the per-boolean escalation budget (#1109 follow-up).
+
+The deterministic per-boolean budget (#1112) bounded a *single* boolean, but two holes kept dense models stalling past the geometry-stream watchdog:
+
+- **Overshoot.** The budget's `tripped()` check only fired at arrangement loop boundaries — once per triangle in the seam retriangulation. A single heavily-fragmented host face (a slab cut by 24-47 openings) inserts thousands of constraint points in *one* `triangulate` call, so a boolean ran to **7.7M** escalations — ~4 minutes — between two checks before bailing. Profiled on a real model: one IFCSLAB took 243 s.
+- **Distributed cost.** An element with many openings runs one boolean *per* opening, each well under the per-boolean cap, so none trips — yet the element's total exact work is huge and the geometry batch blows the stream watchdog.
+
+This adds a **per-element** escalation budget alongside the per-boolean one. `kernel::budget::begin_element()` (called once per element at the unified `produce_element_meshes` entry — native *and* wasm) accumulates escalations across every boolean the element issues; when the element total crosses `DEFAULT_ELEMENT_CAP = 100_000` it degrades as a whole (remaining cuts bail to the #635 AABB box-cut), instead of grinding. The kernel's per-point retriangulation and constraint-recovery loops now also check the budget, so a single boolean can no longer overshoot the cap by 15×.
+
+Still a **deterministic count**, accumulated in deterministic per-opening order on the element's single worker thread (the kernel has no internal rayon), so native and wasm degrade the *same* element identically — the cross-target parity the kernel exists to guarantee is preserved. Calibrated against the model corpus: healthy per-element totals are p99 ≈ 13k escalations, so the 100k cap (~8× p99) never false-trips a legitimate cut. The cap engages only when an element scope is opened (the batch path); direct kernel/router callers, the server, and offline export stay unbounded via the existing `set_cap(None)` / `IFC_LITE_CSG_BUDGET=0` switch — so the pinned determinism manifests are unchanged.
+
+Measured on the profiling corpus (one boolean-heavy structural model): the worst element drops from 243 s to 2.9 s, and total serial geometry from minutes to ~28 s.
--- a/.changeset/csg-1109-recovery-perf.md
+++ b/.changeset/csg-1109-recovery-perf.md
@@ -1,0 +1,12 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Speed up the exact CSG kernel's constraint-recovery hot path on dense-opening models (#1109).
+
+Profiling the boolean-heavy slabs that hung the geometry stream showed the kernel spends ~80% of its time in constraint-recovery retriangulation — split between the channel-detection scan and the pocket earcut. Two parity-safe optimizations:
+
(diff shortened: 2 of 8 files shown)


louistrue:
## Follow-up: constraint-recovery perf (commit a4b06e06)

After the budget made boolean-heavy models *finish*, I profiled where the exact kernel actually spends its time so fewer elements need to degrade. An env-gated phase profiler showed **~80% of CSG time is in constraint-recovery retriangulation** (not where #1109's analysis pointed) — split between the channel-detection scan (9.6s) and the pocket earcut (10.7s) on ISSUE_129.

Two **parity-safe** optimizations (one file, `kernel/retriangulate.rs`):

1. **Channel detection** — the per-segment O(tris) scan recomputed `orient(a,b,vertex)` per triangle *edge*; now computed once per vertex with a straddle short-circuit (~3 predicates/triangle vs up to 12). **Channel scan 9.6s → 2.9s (3.3×).**
2. **Pocket earcut** — the ear-emptiness test ran an exact `strictly_outside` per ring vertex; added a conservative f64-AABB prefilter (same widened margin as `tri_aabb_disjoint`) that skips provably-outside vertices. Also lowers per-element escalation counts, so the budget cuts more openings exactly before degrading (e.g. id=295370: 170k→70k escalations, 3.1s→1.8s).

Both are **byte-identical** (same exact predicate signs; the prefilter only skips vertices it proves are outside) — geometry suite **54/0** + processing **17/0**, manifests/snapshots unchanged, native==wasm parity intact.

End-to-end on ISSUE_129 (per-element budget on, same 100k cap): **23.6s → 19.4s** of serial geometry. This *understates* the kernel win — the budget spends the freed escalation headroom on cutting more openings exactly rather than pure wall-time; the channel-detection raw speedup is 3.3×.

The profiling scaffolding used to find this was stripped from the final diff (kept it clean); the calibration harness `tests/csg_model_profile.rs` (with `IFCLT_MODEL`) remains for future tuning.

louistrue:
## Review feedback addressed (commit ac4bd1dc)

Thanks @coderabbitai / @chatgpt-codex-connector — all findings triaged:

1. **[P1] Skip the conformity audit when the budget trips** (`kernel/retriangulate.rs`) — **fixed**. Real bug in the overshoot guard: on a trip the recovery loop breaks but `converged` stays false, so the O(segments × vertices) audit still ran with no budget check, letting a tripped fragmented face keep grinding past the cap. Added an early `return Some(mesh)` on `budget::tripped()` right after the recovery loop — the boolean discards the whole arrangement on a trip anyway. `tripped()` is a deterministic count, so the early return fires identically native==wasm; the non-tripped path is byte-identical (geometry suite 55/0).

2. **[P1 / Major] MPL-2.0 header on new files** — **fixed** for the two new `.rs` test files (`csg_model_profile.rs`, `issue_1109_budget.rs`), matching the existing `tests/*.rs`. The `.changeset/*.md` files are **intentionally left without it**: every existing changeset in `.changeset/` starts directly with `---` frontmatter and carries no MPL header, so adding one would diverge from repo convention (changesets are release metadata, not source).

3. **[Major] Harness `tripped` used configured caps, not active state** — **fixed** per the suggested diff: `budget::tripped()` + `budget::peak().max(budget::count())`, and the summary prints both active caps.

4. **[P2] Georef models in the profiling harness** — the harness is a manual `#[ignore]` calibration tool; **documented** that it drives the router with a zero RTC offset and must be calibrated on local-origin models only (the shipped `DEFAULT_ELEMENT_CAP` was calibrated on ISSUE_129, local-origin). Full RTC rebasing is out of scope for the diagnostic.

Issue #833, opened 2026-05-27
Floating slab in schependomlaan.ifc — WASM emits mirrored profile, native Rust correct


louistrue:
## Reproduction

Load `tests/models/ara3d/schependomlaan.ifc` (47 MB, IFC2x3). On the 3rd floor (`03 derde verdieping`, storey `#776514`), the `IfcSlab` named `vloerveld` (express id `#783200`, GlobalId `3VK06fkLr3MfpgZhk2ZxRI`) renders as a thin slab **sticking out to the right of the building** rather than inside it.

User-clicked world position panel shows `E 24.785 N 14.250 Z 8.650` — outside the slab's spec-correct bounds.

## What the IFC actually says

```
#783200 = IFCSLAB('3VK06fkLr3MfpgZhk2ZxRI', …, 'vloerveld', $, $, #783166, #783196, …, .FLOOR.);
#783166 = IFCLOCALPLACEMENT(#776513, #783165);     ; parent = storey #776513
#783165 = IFCAXIS2PLACEMENT3D(#783163=(21070, 12750, -380), Z=(0,0,1), X=(1,0,0));
#783175 = IFCPOLYLINE((#783167=(0,0), #783169=(-7430,0), #783171=(-7430,-3000), #783173=(0,-3000), #783167));
#783177 = IFCARBITRARYCLOSEDPROFILEDEF(.AREA., '', #783175);
#783187 = IFCEXTRUDEDAREASOLID(#783177, #783184=identity-placement, #783185=(0,0,1), 60.);
```

Storey `#776513` is at `(0, 0, 9000)` in building-local, building is at origin, site is at origin. Profile points are in the -X / -Y quadrant from placement origin → world bounds should be `(13.640–21.070, 9.750–12.750, 8.620–8.680) m`.

## What native Rust produces (correct)

`cargo run` with `GeometryRouter::with_scale(0.001)` + `set_material_layer_index(MaterialLayerIndex::from_content(&content, &mut dec))` (mirroring the WASM router config) returns:

```
process_element                     : (13.640, 9.750, 8.620) → (21.070, 12.750, 8.680)
process_element_with_voids          : (13.640, 9.750, 8.620) → (21.070, 12.750, 8.680)
process_element_with_submeshes      : sub[0] geom=#783187 (same bounds, 26 verts, 12 tris)
process_element_with_submeshes_voids: 0 sub_meshes (slab has no openings — expected)

[...]

Issue #976, opened 2026-06-06
[Epic] feat(geometry): consumer-configurable tessellation quality (detail-level enum)


Blogbotana:
## Summary

Geometry tessellation detail is currently **hardcoded** in the engine — there's no way for a consumer to ask for coarser or finer geometry. On large models, curved features (holes/openings, cylinders, swept/revolved solids, NURBS) come out visibly faceted, with no knob to improve them; conversely, a consumer who wants maximum throughput can't request coarser meshes.

**Ask:** let the **consumer choose the detail level** via a simple enum (≈5 levels). Internally we map each level to deflection coefficients that drive the existing tessellators. **No change to global defaults** — the engine just accepts an optional parameter and falls back to today's behaviour when it's not provided.

## Current state (all hardcoded, per-processor)

- **Circles & openings/holes** — `calculate_circle_segments(radius)` in `rust/geometry/src/profile.rs:330`:
  ```rust
  segments = clamp(ceil(sqrt(radius) * 8), 8, 32)
  ```
  Openings go through this (`profile.rs:311`). The `32` cap + `sqrt(r)*8` factor is the main cause of faceted holes/cylinders on large objects.
- **Swept (pipe) profiles** — fixed `segments = 24` (`rust/geometry/src/processors/swept.rs:174`, `:381`).
- **Sectioned / revolved / NURBS** — own rules: `2°` max angle step capped at `256` subdivisions (`processors/sectioned.rs:68`), revolution `2°`-step `clamp(2,32)`, NURBS u/v `clamp(8,24)` (`processors/advanced_face.rs`).

There's no shared "quality" concept and nothing is exposed through `processGeometry` / `buildPrePass*`, `GeometryRouter`, or the `@ifc-lite/geometry` SDK.

## Proposed design

1. **A detail-level enum** (consumer-facing), e.g.:
   ```
   TessellationQuality = Lowest | Low | Medium | High | Highest
   ```
   - `Medium` (or a dedicated `Default`) maps to **exactly today's coefficients** so existing behaviour is unchanged when unspecified.
   - Each level maps internally to a target **angular deflection** (max chord-angle error), and optionally a linear-deflection / min–max segment clamp. We pick the coefficients; the consumer only picks the level.

2. **Why angular deflection (internally):** it holds a constant visual error regardless of feature size, so "holes more detailed" works for both a 50 mm and a 2 m opening — unlike the current radius-only `sqrt(r)*8`.

[...]

PR #1130, opened 2026-06-15
feat(geometry): content-dedup identical representation items (skip re-meshing)

louistrue:
## Problem

Boolean-heavy models exported **without** `IfcMappedItem` — e.g. structural-steel detailers that emit every connection plate/bolt as its own representation item — re-mesh and re-CSG thousands of byte-identical solids. With the exact pure-Rust kernel (#1024, ~20–40× slower per cut than the old Manifold path), this dominates load time. A 19.5 MB steel model spent ~85% of geometry time re-computing duplicates — a regression from the Manifold era, where cheap booleans hid the waste. This is the root cause behind the #1109 "hangs at 95%, used to load in seconds" reports for that class of model.

A spike measured the floor: **2562 unique geometries / 15291 elements (83% duplicates)** → re-computing the duplicates cost ~159 s of the ~182 s total.

## Fix

A content-dedup layer in `process_representation_item`: a **128-bit structural hash** of the fully-resolved item subtree (`content_hash::item_signature`) keys a shared cache of the **LOCAL, void-free, colour-free** item mesh. A cache hit skips meshing + CSG entirely.

The per-instance attributes stay out of the cache and are applied by the caller exactly as before:
- **`geometry_id`** → colour / per-triangle palette / texture (so two identical-geometry parts with different colours both render correctly — no styled-item gate needed, because colour rides `geometry_id`, not the mesh)
- **voids** (per-element opening cuts) and **placement** (per-instance rigid transform + local-frame origin)

So a cache hit is **byte-identical** to a fresh build. The 128-bit full-structure hash (vs the sampled 64-bit mesh hash that collided in #833) makes a false merge ~1e-30, and it's deterministic integer splitmix64 → native and wasm produce identical keys.

The cache is `Arc<Mutex<_>>`, shared across one model's routers so it outlives any single router:
- **native:** one cache for the whole rayon pool (`processor.rs`)
- **wasm:** one cache per worker, injected into every per-batch router and held on `IfcAPI` so it persists **across batches**, not just within one (`gpu_meshes.rs`) — the router is rebuilt per batch, so without this the cache would reset ~30× on a large model

## Why item-level (not representation-level)

An earlier representation-level attempt had to skip every per-item-styled element to stay colour-correct — and this Tekla file styles ~98% of elements, so dedup never fired. Moving dedup down to the item level (where `geometry_id` is already assigned per-instance by the caller) makes it colour-correct *by construction* and removes the gate. Aligns with the existing per-item `geometry_hash_cache`, except this skips the **compute**, not just storage.

## Measured (serial, exact kernel)

| Model | Before | After | Mismatches |
|---|---|---|---|
| 19.5 MB steel (no IfcMappedItem) | 182 s | **24 s (7.6×)** | 0 / 15291 |
| 35 MB architectural (82 voided hosts) | 4.78 s | 4.49 s | 0 / 6574 |

The second row is the corpus check: a totally different model where ~46% of items are unique — the hash overhead is **negligible** (slightly net-faster), confirming no regression on ordinary models. 24 s serial → ~6 s across the viewer's 4 workers, restoring the pre-kernel-swap load time.

## Tests

- **Committed CI guard** `content_dedup_byte_identical_on_synthetic_duplicates` (no external fixture): asserts ON==OFF byte-for-byte, correct collapse (2 duplicates → 1 cached mesh), and that placement stays per-instance.
- **Gated A/B harness** `dedup_validate` (`IFCLT_MODEL=…`): drives the production submesh path with dedup ON vs OFF, fingerprints every element, reports speedup. Source of the numbers above.
- Geometry suite **477 ✓**, processing suite **62 ✓**, clippy clean.

## Follow-up (not in this PR)

WASM workers are separate realms, so the cache is **per-worker** — each worker dedups its own assigned elements. The reported regression is resolved by that alone (7.6× less total work). Routing identical geometry to the same worker (content-affinity) could squeeze the cross-worker case further, but needs a JS-side hash pre-pass + scheduler change and browser measurement; deferred.

## Summary by CodeRabbit

* **New Features**
  * Added structural, content-based deduplication so repeated, byte-identical representation-item geometry is meshed once and reused.
  * Introduced shared dedup caches for parallel geometry processing and for consecutive WASM geometry batches, with proper cache reset between loads.
* **Tests**
  * Added tests that verify identical results and expected cache reduction on synthetic duplicates.
  * Added a regression test ensuring dedup keys account for tessellation quality.
  * Added an ignored, performance-focused comparison test.

### DIFF
--- a/rust/geometry/src/lib.rs
+++ b/rust/geometry/src/lib.rs
@@ -126,7 +126,7 @@ pub use profile::{Profile2D, Profile2DWithVoids, ProfileType, VoidInfo};
 pub use profile_extractor::{extract_profiles, ExtractedProfile};
 pub use profiles::ProfileProcessor;
 pub use router::{
-    ClassificationStats, GeometryProcessor, GeometryRouter, HostOpeningDiagnostic,
+    ClassificationStats, GeometryProcessor, GeometryRouter, HostOpeningDiagnostic, ItemDedupCache,
     OpeningDiagnostic, OpeningKindDiag,
 };
 pub use tessellation::{scale_segments, TessellationQuality};
--- a/rust/geometry/src/router/content_hash.rs
+++ b/rust/geometry/src/router/content_hash.rs
@@ -31,9 +31,11 @@ use ifc_lite_core::{AttributeValue, EntityDecoder};
 use rustc_hash::FxHashMap;
 
 /// Defensive recursion bound. IFC geometry is a DAG (item → solids → profiles →
-/// points), so this is never hit by well-formed files; it bounds a malformed
-/// cyclic one.
-const MAX_DEPTH: u32 = 96;
+/// points); deeply NESTED `IfcBooleanResult` chains are the realistic deep case,
+/// so this is set well above any plausible cut chain. Beyond it the hash falls
+/// back to an entity-id-distinct value (see `sig_entity`) so over-depth subtrees
+/// can never COLLIDE — they simply stop deduping rather than risk a false merge.
+const MAX_DEPTH: u32 = 256;
 
 /// Sentinel written into the memo while an entity's hash is being computed, so a
 /// (malformed) cycle resolves to a fixed value instead of recursing forever.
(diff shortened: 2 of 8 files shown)


louistrue:
Thanks @coderabbitai @chatgpt-codex-connector — all addressed in 76d202c6 + 8dcd2208:

- **[P1] Clear the item cache between model loads** — `cached_item_dedup` is now dropped in both `clear_pre_pass_cache` and `set_entity_index`, alongside the other content-scoped caches, so a new file on a reused `IfcAPI` starts with an empty cache. (The key is content- not id-based and now includes unit scale/RTC, so cross-model reuse was geometrically safe, but this bounds memory and is the right place for it.)

- **[P2] Tessellation-quality staleness** — fixed in 76d202c6 by folding the quality index (+ unit scale + RTC) into the cache key, so a `setTessellationQuality` change produces fresh keys/meshes instead of serving the first-cached quality. Both qualities can coexist in the cache, so toggling stays instant. Regression test `content_dedup_keys_on_tessellation_quality` covers it; the failing #976 Node test now passes.

- **[P2] Hash depth limit** — beyond the bound the hash now folds the entity id (distinct value per subtree) instead of one constant, so over-depth boolean chains can never false-merge — they simply stop deduping. Bound raised 96 → 256, well above any realistic `IfcBooleanResult` cut chain.

- **[Major] A/B harness silent skips** — `run()` now records sentinel fingerprints for decode/mesh failures, so ON and OFF stay 1:1 aligned and any config-dependent divergence surfaces as a mismatch rather than shrinking the compared set.

Geometry suite 479 green, processing 62 green, clippy clean.

PR #1131, opened 2026-06-15
perf(geometry): content-affinity worker routing by exact geometry hash

louistrue:
## Problem (the cap left by #1130)

Per-worker content-dedup (#1130) eliminated *within*-worker redundant meshing, but the parallel dispatcher splits each chunk **interleaved** across workers to balance complexity. That spreads every geometry's duplicates across **all** workers, and each worker is a separate WASM realm with its own dedup cache — so all N workers redundantly re-mesh the full unique set. Wall time stayed pinned to the **serial unique-floor** regardless of worker count: the 19.5 MB steel model loaded in ~32 s on 4 workers instead of the ~6 s the dedup floor allows.

## Why ObjectType isn't enough (measured)

My first attempt keyed routing on `ObjectType` (cheap, no decode). Measured on the steel model it only reached **10.3 s** — because a single ObjectType (`PL70*8`) is **10 s of *distinct* geometries**, all forced onto one worker. ObjectType is too coarse: many different geometries share a type.

## Fix: route by the exact geometry hash

The dedup already keys on a 128-bit structural hash of each representation. I expose that as `geometry_routing_key` and compute it in the pre-pass **after** the scan completes the entity index, tagging each job with the folded u32 key on a parallel `affinity` array. Deferring emission to post-scan is **free** — workers gate on the styles + entity-index events, which are themselves post-scan. The dispatcher (`planAffinityRouting`) sends all jobs of a key to the **same** worker (sticky map, new keys round-robin), so the N workers **partition** the unique geometries instead of replicating them.

## Measured (steel model: 15291 elements, 2562 unique, 24.1 s serial dedup floor)

Projected 4-worker wall of the unique meshing:

| routing key | 4-worker wall |
|---|---|
| **GEOMETRY hash (this PR)** | **5986 ms — = ideal perfect split** |
| ObjectType (rejected) | 10346 ms |
| serial (1 worker) | 24097 ms |

Routing-key prepass cost: **51 ms** (decode is index-cached; the hash is integer folds). The geometry hash hits the ideal partition. The per-worker dedup cache still guarantees correctness, so a 32-bit collision only co-locates two geometries on one worker — never a wrong mesh.

## Tests

- **Rust**: wasm `fold_u128_to_u32` unit test; gated `analyze_affinity_partition` harness (produced the table above + the prepass-cost number).
- **TS**: `planAffinityRouting` — delivery, key→worker stickiness across chunks, round-robin spread.
- Geometry package 75 tests + wasm/geometry crates green; typecheck + clippy clean.

## Validation note

Browser-perf change — the ~6 s is the measured dedup-floor ÷ 4-worker partition, confirmed by the analysis harness on the real model. Needs an in-browser run (rebuild WASM + viewer) to confirm wall-clock. Pairs with #1130 (merged), the prerequisite per-worker dedup. This is the exact-kernel floor; the historical 5 s was the Manifold kernel (replaced in #1024 for robustness/determinism), so ~6 s is the realistic target with exact arithmetic.

## Summary by CodeRabbit

* **New Features**
  * Added content-affinity worker routing for geometry jobs, grouping work by a per-element geometry-derived affinity key to reduce redundant meshing and improve dedup performance on boolean-heavy models, with fallback to the prior interleaved split when affinity is unavailable.
  * Enhanced streaming dispatch to defer job emission until the entity index is ready and to carry affinity alongside job batches.
* **Tests**
  * Added routing behavior coverage (coverage, sticky key→worker mapping, balanced new-key distribution) plus affinity partition analysis tooling.
* **Documentation**
  * Documented the affinity-key hashing and routing behavior.
* **Chores**
  * Bumped patch versions for geometry and wasm packages.

### DIFF
--- a/.changeset/content-affinity-routing.md
+++ b/.changeset/content-affinity-routing.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/wasm": patch
+"@ifc-lite/geometry": patch
+---
+
+Content-affinity worker routing for boolean-heavy models. The streaming geometry
+pre-pass now tags each job with an affinity key — the exact 128-bit hash of the
+element's representation geometry — and the parallel dispatcher routes all jobs
+sharing a key to the same worker. Combined with the per-worker geometry-dedup
+cache, each unique geometry is meshed once **per model** instead of once per
+worker, so the workers partition the unique meshing instead of replicating it.
+Restores fast loads on models exported without `IfcMappedItem` (e.g.
+structural-steel detailers that emit thousands of byte-identical parts): a 19.5 MB
+steel model drops from ~32 s to ≈ the dedup floor split across the worker pool.
+Falls back to the previous interleaved split when no affinity data is present.
--- a/packages/geometry/src/affinity-routing.test.ts
+++ b/packages/geometry/src/affinity-routing.test.ts
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, expect, it } from 'vitest';
+import { planAffinityRouting } from './geometry-parallel.js';
+
+describe('planAffinityRouting (content-affinity worker routing)', () => {
+  it('routes every job exactly once', () => {
+    const affinity = new Uint32Array([10, 20, 10, 30, 20, 10]);
+    const { buckets } = planAffinityRouting(affinity, 6, 4, new Map(), 0);
+    const all = buckets.flat().sort((a, b) => a - b);
+    expect(all).toEqual([0, 1, 2, 3, 4, 5]);
+  });
+
+  it('sends all jobs of one affinity key to the SAME worker', () => {
+    const affinity = new Uint32Array([10, 20, 10, 30, 20, 10]);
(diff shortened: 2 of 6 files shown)


louistrue:
Thanks @coderabbitai @chatgpt-codex-connector — addressed:

- **`resolved_unit_scale` never set on the main path** — removed entirely. The routing key is `geometry_routing_key` → `item_signature`, a purely STRUCTURAL hash that neither tessellates nor scales, so the affinity decoder/router needs no unit scale at all. Replaced with `GeometryRouter::new()` + no `seed_unit_scales`. Good catch that it was dead/half-wired.

- **[P1] "pinning an entire ObjectType to one worker"** — stale (reviewed the earlier ObjectType-keyed commit). The current key is the exact geometry hash, not ObjectType: products with the same ObjectType but different representations get DIFFERENT keys and spread across workers. Validated by a TRUE 4-router per-worker-cache simulation (`analyze_affinity_partition`): redundancy 1.01× and balanced workers (6.5/5.3/6.7/5.2 s), i.e. no idle-worker regression.

- **[P2] sticky map retains id-fallback keys** — `keyToWorker` is created per `processParallel` call (one model) and GC'd after, and the id fallback only fires on a decode failure for a geometry-bearing job (rare). So it's bounded by one model's distinct keys (≈ unique geometries), not unbounded. Left as-is.

- **MPL header on the changeset** — changesets follow the repo convention of no license header (the `.changeset/` dir is tooling metadata, stripped on version); skipping per established practice.

PR #1134, opened 2026-06-15
feat(geometry): f64 planar-clip fast path for convex half-space end-clips

louistrue:
## What

Routes the dominant Tekla connection-cut pattern — a solid (or polygonally-bounded-half-space) cutter that acts as a single **half-space end-clip** of the host — through an f64 plane clip + cross-section cap instead of the exact CSG arrangement. ~100–150× cheaper per qualifying cut, geometrically equivalent.

`ClippingProcessor::subtract_one` is the new single entry: it tries `try_planar_clip`, then falls back to the exact `subtract_mesh`. `clip_mesh_with_cap` clips the host to the front half-space (reusing `clip_triangle`), extracts the open cut boundary, stitches closed loops, triangulates the cross-section (outer + holes), and caps it — wound for half-edge cancellation so the result is a watertight solid.

## Why it is only ever a fast path (three gates)

A cutter takes the fast path **only** when all of these hold; otherwise it defers to the exact kernel unchanged:

1. **Convexity** — every cutter vertex lies inside all of its own face planes. A non-convex cutter (L-prism, notched block) is not a half-space and could over-cut.
2. **Single straddle** — exactly one cutter face plane straddles the host, and the host lies entirely inside every other face. Notches, pockets and through-holes straddle on ≥2 faces → defer.
3. **Precision regime** — host coordinates stay under the codebase's 10 km large-coordinate threshold. Above it, f32 ULP exceeds a millimetre, so the cap's watertightness is unverifiable; un-localized georeferenced geometry degrades identically under the exact kernel, so deferring is never worse.
4. **Watertightness** — the capped result must have zero open boundary edges (1 mm grid), else defer.

`IFC_LITE_DISABLE_PLANAR_CLIP=1` (native) forces the exact kernel everywhere.

## Where it is wired

- the solid-cutter `IfcBooleanResult` difference;
- the single `IfcPolygonalBoundedHalfSpace` subtract — where it *also* sidesteps the coincident-side-wall sliver collapse the exact kernel hits when the clip polygon spans the host cross-section;
- the extended-opening void cut.

The simple `IfcHalfSpaceSolid` path (already an O(n) `clip_mesh`) is untouched.

## Validation

- **clip_with_cap suite**: primitive is watertight and volume-matches the exact subtract to <1.3e-4 on axis/oblique/steep cuts; 40/40 generic angled cuts cap cleanly; detection routes end-clips to the fast path and defers holes / pockets / non-convex L-prisms; `subtract_one` fires on a clean end-clip (watertight, matches exact, different triangulation) and defers on a through-hole.
- **Full geometry suite**: 483 passed, **0 snapshot changes** (gable / duplex-PBHS / segmented-roof fixtures unchanged).
- **A/B harness** (`planar_clip_ab`, `#[ignore]`): zero open-edge regressions and shape/volume equivalence vs the exact kernel on a clean local model; full deferral (zero firing, zero regressions) on an adversarial 792 km georeferenced model.

## Honest scope

This accelerates the **spanning convex end-clip** subset (≈1–2 % of elements in the structural models sampled), not every boolean. It does not speed up simple `IfcHalfSpaceSolid` models (already fast) and conservatively defers on un-localized georeferenced models. Real-world magnitude depends on a model's cutter composition; the per-element test harness cannot relocalize baked-georef coordinates, so the firing+speedup on georeferenced structural models is best measured through the production (localizing) viewer path.

Stacked on the `clip_mesh_with_cap` primitive (first commit). Changeset included (`@ifc-lite/geometry` patch).

## Summary by CodeRabbit

  * Added a faster planar-clipping/subtraction path for specific convex half-space “end-clip” patterns, with automatic verification and fallback to exact processing for complex or non-conforming geometry.
  * Introduced additional safeguards: if the fast result isn’t sufficiently well-formed (open-edge check), the element is reprocessed with the exact approach and the best result is selected.
  * Added `IFC_LITE_DISABLE_PLANAR_CLIP` to force the exact kernel when required.
* **Tests**
  * Added coverage for capped planar clipping and subtraction behavior, including watertightness/volume consistency checks.

### DIFF
--- a/.changeset/planar-clip-fast-path.md
+++ b/.changeset/planar-clip-fast-path.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Add an f64 planar-clip fast path for convex half-space end-clips — the dominant Tekla connection-cut pattern — replacing the exact CSG arrangement with a single plane clip plus cross-section cap when a solid or polygonally-bounded-half-space cutter acts as one half-space of the host.
+
+`ClippingProcessor::subtract_one` now routes each single-cutter difference through `try_planar_clip` before the exact kernel. A cutter qualifies only when it is convex, exactly one of its face planes straddles the host, and the host lies entirely inside every other cutter face (the cutter overhangs the host on all but one side) — so notches, pockets, through-holes and non-convex cutters fall through to the exact arrangement unchanged. The clip caps the cut with the cross-section (outer boundary plus holes) wound for half-edge cancellation, producing a watertight solid geometrically equivalent to the exact subtract at roughly 100–150× lower cost per cut.
+
+Three independent gates keep it correctness-safe — it is only ever a fast path, never a different answer:
+
+- **convexity** — a cutter vertex outside any of its own face planes means the cutter is not a half-space, so defer;
+- **precision regime** — host coordinates above the 10 km large-coordinate threshold defer, because f32 ULP exceeds a millimetre there and watertightness is unverifiable (un-localized georeferenced geometry degrades identically under the exact kernel, so deferring is never worse);
+- **watertightness** — the capped result must have zero open boundary edges on the 1 mm grid, otherwise defer.
+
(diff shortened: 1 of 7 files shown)


louistrue:
Closing — profiling shows this fast path does not move the needle.

On a 286,959-element steel model it fires on **14 elements** and saves **0.0s** of load (9.9s with the planar clip on vs 9.9s off). rect_fast already reduced that model's CSG to 0.9s, and **meshing is now the dominant cost** (8.5s, 86% of load). The planar-clip optimizes the wrong, already-small half. On rotated-wall models it fires ~0 as well.

Also surfaced in review (would need fixing if ever revived): the per-thread adaptive give-up makes fast-vs-exact routing scheduling-dependent → non-byte-identical native==wasm (geom_hash is a triangle-multiset, so fast-cap ≠ exact-arrangement); plus a randomized std-HashMap cap-loop ordering. Not worth it given zero measured benefit.

Branch kept for reference. The real levers are the browser pre-pass overlap and meshing/tessellation, not more CSG fast paths.

Issue #1107, opened 2026-06-13
Few things I spotted while I recorded my ifclite video today


BIMvoice:
1. Dark theme contrast feels too low. The text and UI are hard to read against the dark background.

2. Version number looks inconsistent. I saw version 1.3 in one place (main interface right bottom corner) and version 3.2 in the “What’s new” area.

3. Orbit navigation without selecting an object feels like it rotates around a strange origin. After too much rotation, it can feel a bit hard to control.

4. Wall display mode reload did not seem to apply. I had to reopen the model.

5. The “export with changes” state could be more visible. Maybe a stronger color or clearer warning when changes exist.

6. If I select something from the hierarchy or class tree, it would be useful if that filter also appeared in the advanced filter panel? - curious if would make sense and if possible.

7. 3D measurement snapping felt off. The snap point appeared away from the cursor and was hard to trust. The section view measurement worked much better than the 3D measurement.

8. The section window can resize and go full screen, but I was not sure if it can be moved.

9. The annotation icon felt confusing. It looked more like a map pin than an annotation tool.

10. bSDD property editing was not clear. I could add a property, but I was not sure how to set values like true or false.

11. CSV quantity export seemed to export all quantities, not the current filtered view. Filtered export would be useful?

Most of them small things, compared to what already makes this an amazing tool, thank you so much for your incredible work!

louistrue:
thx @BIMvoice 👍 keep it coming 😉

BIMvoice:
you're welcome brother, thank you a bunch for making this app a really amazing one!

PR #1146, opened 2026-06-16
fix(viewer): project overlays to CSS space — measure/gizmo cursor offset (#1107)

louistrue:
Part of #1107 (item 7: *"3D measurement snapping felt off. The snap point appeared away from the cursor and was hard to trust. The section view measurement worked much better than the 3D measurement."*).

> ⚠️ **Draft — needs in-viewer visual QA.** The logic is type-checked and unit-tested, but I can't drive the 3D viewer here to confirm the *feel*. Please test snapping on a real model before merging.

## Root cause
The magnetic snap detector (`snap-detector.ts`) picks the geometry feature closest to the raycast hit in **world space**, with the search radius enlarged **3× (edges) / 6× (corners)**. That radius is sized at the *hit's* depth, but the chosen feature can sit at a *different* depth — so under perspective the snap indicator projects far from the cursor. The 2D section tool (`useMeasure2D.ts`) filters purely in screen space, which is why it already felt trustworthy.

## Fix
A caller-side `gateSnapToScreen()` in `measureHandlers.ts`. After each magnetic raycast it re-projects the detector's chosen snap to screen and, if it lands further than the requested `screenSnapRadius` from the cursor, **demotes it to the raw surface hit** (and releases any pending lock) so the measurement point lands under the cursor.

- **Edge sliding preserved:** the gate is skipped while an edge lock is already active. And because *fresh* (unlocked) acquisitions are gated, a far feature can never become a lock in the first place — so locks are always near-cursor.
- **Selection untouched:** deliberately kept in the viewer caller rather than the renderer's `raycastSceneMagnetic`, which is **shared with object selection** (`selectionHandlers.ts`). App-only → **no `@ifc-lite/renderer` change, no changeset**.
- Applied at all four measurement snap sites: down / drag / hover / up.

## The one tuning knob
The trust radius reuses each call's existing `screenSnapRadius` (60px place / 40px hover). This intentionally caps the long-range 3–6× magnetic pull in favour of cursor-tracking (the behavior the reporter preferred). If snapping should pull *harder*, raise the radius passed to `gateSnapToScreen`; to track *tighter* (more like the 2D tool), lower it.

## Verification
- `npx tsx --test src/components/viewer/measureHandlers.snap-gate.test.ts` → 10/10 pass (pixel-distance + gate demotion/skip behavior)
- `tsc --noEmit` on the viewer → 0 errors
- ⛔ Not yet validated interactively in the 3D viewer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved coordinate projection across measurement tools and viewport overlays to ensure measurement points and visual elements align accurately with cursor position during interactions.
  * Enhanced screen coordinate calculations throughout the measurement system to provide better alignment and positioning precision across all measurement operations and viewport interactions.


louistrue:
Reworked: the screen-space snap-gate was treating a symptom. Real root cause (confirmed from your report — a position-dependent leftward drift) is that the canvas drawing buffer width is `alignToWebGPU`-rounded **down** to a multiple of 64, so it's narrower than the CSS box; `projectToScreen` returns buffer-space coords that the DOM overlays place as CSS pixels → drift left, worse toward the right edge. Now scaled buffer→CSS via a shared `projectToCssScreen` helper. Fixes the measure gizmo/line and the move gizmo; section/basepoint/BCF overlays already used `clientWidth` and were correct.

PR #1149, opened 2026-06-16
fix(viewer): bSDD value flow — no prefill, correct type, edit-on-add (#1107)

louistrue:
Part of #1107 (item 10: *"bSDD property editing was not clear. I could add a property, but I was not sure how to set values like true or false."*).

## Problem
The bSDD card added every property with an **empty string** value (`defaultValue()` always returned `''`). To set a boolean you had to: add it → find it in the Properties panel → open the inline editor → use the Switch. There was no on-card way to set true/false, and nothing signalled a value was needed.

## Fix
Render a value control inline next to each not-yet-added property:
- **Boolean** dataType → a `Switch` (defaults to `false`)
- **enum** (property has bSDD `allowedValues`) → a `Select` of the allowed values

The chosen value flows through `handleAddProperty(psetName, prop, value)` so the property is created **with** it instead of an empty string. `defaultValue()` now returns `false` for booleans — a valid default everywhere it's used (single add, Add-All, and the Switch's initial state) — so a boolean is usable the moment it's added. Other dataTypes are unchanged (still empty for manual entry).

## Verification
`tsc --noEmit` on the viewer → 0 errors. App-only (no changeset). A quick visual check of the bSDD card is worthwhile.

## Release Notes

* **New Features**
  * Added a one-shot “jump to property and edit” workflow: the Properties tab opens, the target row is highlighted, focus is applied after render, then cleared.

* **Improvements**
  * Streamlined bSDD inline value handling with consistent type-to-control mapping, correct defaults, and improved add-flow toasts for inline-editable items.
  * Added tri-state Boolean/Logical editing (“Unset / True / False”) with correct persistence.

* **Bug Fixes**
  * Prevented added checkmarks and edit prompts from carrying over across entity selections.
  * Corrected handling of unset/null properties for edit, delete, and undo behavior.

* **Tests**
  * Added unit tests for bSDD value type mapping and default generation.

### DIFF
--- a/.changeset/bsdd-unset-boolean-property.md
+++ b/.changeset/bsdd-unset-boolean-property.md
@@ -1,0 +1,8 @@
+---
+"@ifc-lite/mutations": patch
+---
+
+Treat a null/unset property value as present, not absent. A property may legitimately exist with no value (e.g. an IFC boolean added from bSDD, which now starts unset rather than defaulting to `false`), so `MutablePropertyView` no longer reads `value === null` as "property does not exist":
+
+- `deleteProperty` keys absence off existence (in-session pset membership), so an unset property is still deletable instead of the trash button being a silent no-op.
+- `setProperty` classifies a write as `UPDATE_PROPERTY` vs `CREATE_PROPERTY` by whether the property already existed (not by null value), so undoing an edit to an unset property restores its prior unset state instead of deleting the whole property.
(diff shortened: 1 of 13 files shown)


louistrue:
Reworked per feedback — **prefill removed entirely** (no Switch/Select before Add). New flow: a bSDD property is created with its **correct data type** (from the bSDD dataType), and on Add the panel **jumps to the Properties tab + auto-enables edit mode + toasts the property name**, so you set the value right there without the toolbar edit button. (Chose 'auto-enable edit mode on Add'.) Properties tabs are now store-controlled (`propertiesActiveTab`).

PR #1150, opened 2026-06-16
feat(viewer): mirror class-tree selection into the advanced filter (#1107)

louistrue:
Part of #1107 (item 6: *"If I select something from the hierarchy or class tree, it would be useful if that filter also appeared in the advanced filter panel?"*).

## Before
The hierarchy's **By Class** tab and the **advanced filter** panel were two independent systems: clicking a class set a render-level `classFilter` (a visibility Set), while the advanced filter used structured `FilterRule[]`. They never shared state.

## Change
On a class-tree click (`groupingMode === 'type'`), in addition to the existing class filter, push an additive **`ifcType in [Class]`** rule into the advanced filter and show a toast.

Per the agreed UX (chosen in discussion):
- **Silent + toast** — the filter modal is *not* opened.
- **Additive** — combines with existing rules rather than replacing them.
- **De-duplicated** — repeated clicks on the same class don't pile up identical rules.

Scope: the **class** case only (globally-consistent IFC class names). Storey/spatial rules were left out deliberately — they raise a federation question (same storey name/elevation across models → different express IDs) better handled separately.

## Verification
`tsc --noEmit` on the viewer → 0 errors. App-only (no changeset). `addFilterRule` already appends without opening the modal.

## Summary by CodeRabbit

* **New Features**
  * Clicking Class items in the hierarchy panel now automatically applies corresponding advanced filter rules. The system prevents duplicate rule creation and provides confirmation feedback through notifications.

### DIFF
--- a/apps/viewer/src/components/viewer/HierarchyPanel.tsx
+++ b/apps/viewer/src/components/viewer/HierarchyPanel.tsx
@@ -19,6 +19,8 @@ import { cn } from '@/lib/utils';
 import { useViewerStore, resolveEntityRef } from '@/store';
 import { toGlobalIdFromModels } from '@/store/globalId';
 import { useIfc } from '@/hooks/useIfc';
+import { Rule } from '@/lib/search/filter-rules';
+import { toast } from '@/components/ui/toast';
 
 import type { TreeNode } from './hierarchy/types';
 import { isSpatialContainer } from './hierarchy/types';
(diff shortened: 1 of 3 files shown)


louistrue:
Reworked per feedback — the bridge now mirrors **all hierarchy tabs that have a compatible filter rule**, as **one rule per dimension synced to the current selection** (upsert, not accumulate):
- **Storey** (Hierarchy tab) — was completely missing; now `storey in [name]` on click, Ctrl+click accumulates, toggling off clears it. Names match the evaluator's `getName(storeyId)`.
- **Class** (By Class) — now **replaces** the ifcType rule with the clicked class instead of piling up values (that's where the stale `IfcVirtualElement` came from).
- **Material** (By Material) — `material eq name`.

By Type maps to no precise rule (there's no 'defined-by-type' filter), so it's intentionally skipped — say if you want it approximated by the instance class.

PR #1153, opened 2026-06-16
feat(geometry): analytic f64 fast path for rectangular openings

louistrue:
**Stacked on #1134** (f64 planar-clip foundation). Review/merge after it.

## What & why

The pure-Rust exact kernel routes **every** opening — rectangular included — through the full arrangement, so CSG void-cutting is **85–90% of load time** on void-heavy models (#1109/#1138). But rectangular openings (a window/door box through a slab-like wall) are ~95% of all cuts. This adds an **analytic f64 through-box subtraction** that bypasses the exact arrangement entirely for that common case — recovering most of the speed lost when we consolidated onto the pure-Rust kernel — **without** reintroducing native/wasm drift or giving up the 0-failures robustness bar.

Phase 2 of the f64 fast-path strategy: Phase 1 (#1134) handles single half-space end-clips; a rectangular through-opening straddles the host on four lateral faces, so it still went to the exact kernel until now.

## How

`ClippingProcessor::subtract_box_fast` (in `csg.rs`):
1. partitions each host triangle by the 4 lateral box planes — a **conforming 3×3 cell split** (every fragment split by every plane → no T-junctions) — and drops the window-footprint cell;
2. synthesises the 4 reveal faces (inner walls of the hole) from the **actual cut rim**, pulling the host's own subdivision points (face diagonals, welds) onto each reveal edge via a **two-pointer chain merge**, so reveal edges match the host rim exactly even when the host's front/back faces are tessellated differently.

`apply_void_context` (in `voids.rs`) tries it per `OpeningType::Rectangular` before materialising a penetrating box mesh for the exact subtract.

## Safe by construction — only ever a fast path, never a different answer

- **Transversality gate** — the box must span the host on exactly one axis and lie strictly interior on the other two; blind pockets, edge/corner cuts and non-transversal boxes return `None` → exact kernel.
- **Coplanar handling** — a host face coplanar with a cutting plane (two windows sharing a sill height) is sent to one side only, never duplicated.
- **Watertight gate** — every directed boundary edge must cancel (result is closed), else `None` → exact kernel.

## Parity (the whole reason for the pure-Rust kernel)

Determinism is by construction: FMA-free f64 over the host's own coordinates, fixed plane/triangle/loop iteration order, and the same `Q`-grid edge stitching as the planar clip. The only `HashMap` use is edge-cancellation set-semantics plus rim-point collection that is **re-sorted by coordinate**, so output is byte-identical on 64-bit native and 32-bit wasm. It never touches the rational tier. The box and planar fast paths share one gate + the element-level fallback (`f64_fast_path_active` / `note_planar_clip_fired`): a cut that cracks once scaled/placed into f32 world coords is re-processed on the exact kernel, and `IFC_LITE_DISABLE_PLANAR_CLIP=1` (native) forces the exact kernel for both.

## Tests

- `rust/geometry/tests/rect_box_fast.rs` — A/B-validates `subtract_box_fast` watertight **and** volume-equal to the exact `subtract_mesh` across both through-axes, sequential disjoint windows (incl. a shared sill), with pockets/corners/no-ops correctly deferring.
- End-to-end test in `voids.rs` (`rect_opening_routes_through_box_fast_path`) — proves the fast path actually **fires** through `apply_void_context` and produces a watertight, volume-correct cut (not dead code).

## Notes for review

- Output is unconsolidated by design (clean fans, no needle slivers), so it has more triangles than kernel+`consolidate_coplanar` but no quality regression; `drop_thin` handles any degenerate sliver downstream.
- Internal `pub fn`s only — no wasm `.d.ts` surface change.
- Changeset included (`@ifc-lite/geometry` patch).

### DIFF
--- a/.changeset/rect-opening-fast-path.md
+++ b/.changeset/rect-opening-fast-path.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Add an f64 through-box fast path for rectangular openings — the overwhelmingly common void (a window/door box cut transversally through a slab-like wall) — bypassing the exact CSG arrangement entirely (~100–3000× cheaper per cut) while staying byte-identical native==wasm.
+
+The Phase-1 planar clip only handles a cutter that acts as a single half-space; a rectangular through-opening straddles the host on four lateral faces, so it still routed through the exact kernel. `ClippingProcessor::subtract_box_fast` now subtracts an axis-aligned box analytically: each host triangle is partitioned by the four lateral planes (a conforming 3×3 cell split, so no T-junctions), the window footprint is dropped, and the four reveal faces are synthesised from the actual cut rim — taking the host's own subdivision points (face diagonals, welds) onto each reveal edge so it matches the rim exactly, including when the host front and back faces are tessellated differently. The void router (`apply_void_context`) tries it per rectangular opening before materialising a penetrating box mesh for the exact subtract.
+
+It is only ever a fast path, never a different answer:
+
+- **transversality gate** — the box must span the host on exactly one axis (the through/thickness axis) and lie strictly interior on the other two; blind pockets, edge/corner cuts and non-transversal boxes defer;
+- **coplanar handling** — a host face coplanar with a cutting plane (e.g. two windows sharing a sill height) is sent to one side only, never duplicated;
+- **watertightness** — every directed boundary edge must cancel (the result is closed) on the stitch grid, otherwise defer to the exact kernel.
+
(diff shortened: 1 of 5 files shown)


louistrue:
@coderabbitai review

louistrue:
Closing — the f64 analytic opening fast path does **not** work on real IFC geometry, proven with three rounds of real-model telemetry (722MB model):

1. Clean-transversal gate → `fired=0`, all `gate` defers. Real openings are doors/flush/shifted, never clean windows.
2. General active-plane rewrite (handles doors/pockets/notches — verified watertight + volume-equal on synthetic cases) → `fired=0` again, all `open` defers: the reveals build fine but **can't close** on f32-tessellated real walls (the f64 reveal rim can't cancel the host rim to watertight precision, because real walls aren't clean slabs after welding/subdivision/prior cuts).
3. The rare case that *does* close can ship a visually-wrong flat-cap (`removed≈0` passes the volume guard) — observed as openings rendered as solid panels.

Root cause: the speed came from skipping exact arithmetic, but exact arithmetic is exactly what makes a cut close reliably on messy real geometry. There's no robust f64 shortcut — the exact kernel is the only correct opening cutter.

The realistic win was already captured by #1135 (127s → 84.6s on this model). Pursuing kernel-level speed (batching / consolidate / deterministic parallelism / wasm arithmetic) on a fresh branch off main instead. The planar end-clip foundation (#1134) is separate and does work (Tekla connection cuts).

PR #1158, opened 2026-06-17
feat(geometry): Step-0 kernel timing instrumentation (feature-gated)

louistrue:
**Step 0 of the kernel-speed roadmap.** We've been flying blind on where a CSG boolean's ~84.6s (722MB model) actually goes — there was no `Instant` anywhere in the kernel, only an op-count census. This adds **per-phase wall-clock + per-phase op counts + per-predicate-tier escalation counters**, on **both native and WASM**, so the next moves (parallelism / consolidate-gating / on-grid starvation / the adaptive-f64 spike) are chosen from data, not estimates.

## What it measures

- **8 phases**: `mesh_to_tris, broadphase, tri_tri, split_crossings, retriangulate, classify, tris_to_mesh, consolidate`.
- **3 predicate tiers**: total predicate evaluations, escalations past the cheap f64 interval filter (the wide-int work wasm32 emulates as multi-limb), and the `BigRational` tier (~3 ms/call).

Sample, single box−box cut (native):
```
split_crossings=…/36% retriangulate=…/40% consolidate=…/16% | predicates=1822 escalated=116(6.4%) rational=0(0.000%)
```

## Design

- `kernel/timing.rs`: lock-free atomics (captures rayon workers). Native wall-clock via `Instant`; wasm32 has no `Instant`, so the host injects `performance.now()` via `set_wasm_clock` (kept out of the web-free geometry crate as a plain `fn() -> f64` pointer; needs `WorkerGlobalScope` for the worker scope).
- **Zero production cost**: behind the `kernel-timing` cargo feature. OFF ⇒ every entry point is an `#[inline(always)]` no-op over a zero-sized guard. Determinism untouched — counters/timers never feed a sign or a coordinate.
- Reuses the existing `budget::note_escalation()` hook as the escalation counter.

## How to run a measurement

- **WASM** (viewer): `KERNEL_TIMING=1 scripts/build-wasm.sh`, reload the model, read `[IFC-LITE] kernel-timing: …` per batch in the console.
- **Native** (server/CLI/test): build with `--features kernel-timing` and drain `ifc_lite_geometry::kernel::timing::take()`. The smoke test prints it: `cargo test -p ifc-lite-geometry --features kernel-timing --test kernel_timing_smoke -- --nocapture`.

## Tests

`kernel_timing_smoke.rs` asserts a real boolean populates the timers/counters with the feature ON and records **nothing** with it OFF. Full geometry suite green (the 3 `wall_opening` failures are a pre-existing stale-local-fixture issue — `advanced_model.ifc` — unrelated to this change; CI fetches the fresh fixture).

Next: run this on the 722MB model native + wasm to decide whether the model is retriangulate-bound or consolidate-bound, and how hot the wide-int tier actually is on wasm.

## Summary by CodeRabbit

* **New Features**
  * Added optional kernel timing instrumentation for CSG operations to measure performance by phase and predicate evaluation tier. Enable with `--features kernel-timing`. Timing metrics are automatically included in CSG diagnostics output and logged to console in WASM builds.

* **Chores**
  * Updated build scripts to support kernel-timing feature configuration via environment variables.

### DIFF
--- a/.changeset/kernel-timing-instrumentation.md
+++ b/.changeset/kernel-timing-instrumentation.md
@@ -1,0 +1,10 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Add opt-in, feature-gated CSG kernel timing instrumentation (`kernel-timing`).
+
+Behind the off-by-default `kernel-timing` cargo feature: per-phase wall-clock,
+per-phase op counts, and predicate-tier escalation counts on both native and
+WASM, to profile where boolean CSG time goes. Default builds compile it out
+entirely — no shipped-artifact or runtime change.
--- a/rust/geometry/Cargo.toml
+++ b/rust/geometry/Cargo.toml
@@ -18,6 +18,10 @@ readme = "../../README.md"
 # (docs/architecture/geometry-pipeline.md).
 default = []
 debug_geometry = []
+# Step-0 kernel instrumentation: per-phase timers + per-predicate-tier
+# escalation counters (src/kernel/timing.rs). OFF in production (zero cost);
+# enable for a measured run (`--features kernel-timing`, incl. build-wasm.sh).
+kernel-timing = []
 
 [dependencies]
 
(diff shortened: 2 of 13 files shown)


louistrue:
Closing — full re-profile shows CSG is no longer the load bottleneck (meshing dominates post-rect_fast; the IfcFacetedBrep mesher is the real lever). This kernel-timing instrumentation profiles the CSG arrangement, which is the smaller half now — lower priority. Branch kept; can revive if we return to CSG kernel work.

PR #1165, opened 2026-06-17
fix(geometry): render IfcSweptDiskSolid over a trimmed IfcLine directrix (#1164)

### LINKED ISSUE #1164: Element defined using IFCSWEPTDISKSOLID fails to load
In some IFCs I've received, some elements fail to load in ifc-lite (tested using https://www.ifclite.com/ and other methods). It looks like the common factor may be the use of IFCSWEPTDISKSOLID to define the shape. The same may apply to other 'BREP' type geometry.

See attached minimal example, the `IFCREINFORCINGBAR` element with GlobalId `381mb$JWz7kgQ0GmJHTpVO`

[single_bar.zip](https://github.com/user-attachments/files/29044656/single_bar.zip)

Other viewers/libraries load the geometry without a problem, including:

- https://viewer.flinker.app/
- ifcopenshell

e.g. in viewer.flinker.app

Image

louistrue:
Fixes #1164.

## Symptom
An `IfcReinforcingBar` (and any element) whose body is an `IfcSweptDiskSolid` swept along an `IfcTrimmedCurve` over an `IfcLine` silently fails to load — it produces an empty mesh and renders nothing. Other viewers (viewer.flinker.app, IfcOpenShell) render it fine. This is the standard Tekla/IfcOpenShell encoding for a straight bar/rod, so it hits real-world steel models.

The reporter's minimal example:

```
#47 = IFCSWEPTDISKSOLID(#46, 14.5, $, 0., 2750.)
#46 = IFCTRIMMEDCURVE(#45, (IFCPARAMETERVALUE(0.)), (IFCPARAMETERVALUE(2750.)), .T., .PARAMETER.)
#45 = IFCLINE(#29, #44)
```

## Root cause
The 3D curve sampler in `ProfileProcessor` (`get_curve_points_with_depth` → `process_trimmed_curve_with_depth` → `process_curve_with_depth`) had **no `IfcLine` arm**. Resolving the trimmed-line directrix fell through to `_ => Err("Unsupported curve type: IfcLine")`, so the swept-disk processor received zero curve points and emitted an empty mesh.

History check: `process_curve_with_depth` never had an `IfcLine` arm, so this directrix encoding never rendered through this path. The sibling encodings that *are* handled — `IfcPolyline` and `IfcCompositeCurve` directrices — are unaffected.

## Fix
- Sample `IfcLine` directly: `P(u) = Pnt + u·V`, with `V = magnitude · normalize(orientation)`.
- Sample a **trimmed `IfcLine` in full 3D**, honoring Trim1/Trim2 (parameter *or* cartesian bounds) and SenseAgreement — the old 2D fallback dropped z, which a swept directrix must keep.
- Apply a solid's own `StartParam`/`EndParam` to a bare `IfcLine` directrix in the swept-disk processor.
- The 2D curve path no longer errors on `IfcLine`, and a trimmed-line basis applies its trims instead of silently sampling the unit segment.

After the fix the reporter's file produces the correct tube: directrix `[[0,0,0],[2750,0,0]]`, 96 triangles, **2.75 m long × 0.029 m diameter**.

## Tests
- New neutral, self-contained regression fixture (`swept_disk_trimmed_line.ifc`) + test (`issue_1164_swept_disk_trimmed_line.rs`) reproducing the issue geometry chain (SweptDiskSolid → TrimmedCurve → Line, wrapped through a RepresentationMap + MappedItem). Asserts a non-empty tube of the correct dimensions.
- Full `cargo test -p ifc-lite-geometry` passes (355 lib + all integration tests).


louistrue:
Follow-up (commit 916314ea): the bar rendered but showed a field of specular **sparkles at distance**.

**Cause:** the reporter's rebar sits at national-grid coordinates ~6 km from the origin — *below* the viewer's 10 km RTC rebasing threshold — so it renders at raw f32 (~6×10⁶ mm). The swept-disk processor shipped **empty normals**, so consumers recomputed them from those world-space f32 positions. At 6 km, the edge differences `v1 - v0` (a ~1 mm tube edge between two ~6×10⁶ values) cancel catastrophically into garbage normals → bright specular sparkle. (Its sibling `IfcRevolvedAreaSolid` already ships normals; the disk was the gap — the "empty normal buffers for swept meshes" the alignment audit flagged in `csg.rs`.)

**Fix:** compute smooth per-vertex normals in the processor, in the directrix-local frame where coordinates are small and precise. Placement only rotates normals, so they stay correct at any world position. A round tube wants smooth (area-weighted) normals, so `calculate_normals` rather than the flat-shaded path used for the crease-heavy revolved solid.

**Residual:** sub-pixel geometric shimmer from f32 positions at 6 km is a separate, broader georef-precision matter (the 10 km RTC threshold / per-element local frame), not addressed here.

PR #1179, opened 2026-06-17
fix(viewer): copy Cesium runtime assets to /cesium so the map renders (#1139)

### LINKED ISSUE #1139: The map on cesium does not show up anymore
I don't manage to see any models on map anymore. Maybe it was affected by the core update? Or it works for you without any issues?

louistrue:
## What

Closes #1139 — *"The map on cesium does not show up anymore."*

## Diagnosis

Production (`www.ifclite.com`) returns **`x-vercel-error: NOT_FOUND` for the entire `/cesium/` asset tree** — Workers, Assets/Textures, Widgets, and credit images all 404. The Cesium engine JS bundles fine (the overlay mounts and *requests* credits), but it can't fetch the runtime Workers/imagery from `CESIUM_BASE_URL` (`/cesium`), so the globe/terrain never render and the map is black.

Root cause is in `apps/viewer/vite.config.ts`. The `vite-plugin-static-copy` **v4** line rebuilds each file's destination from its path *relative to the Vite root*, so an absolute `node_modules` source landed every file at:

```
dist/cesium/node_modules/.pnpm/cesium@1.142.0/node_modules/cesium/Build/Cesium/Assets/Images/ion-credit.png
```

instead of `dist/cesium/Assets/Images/ion-credit.png`. So `/cesium/Assets/…` 404s. Confirmed locally: the built `dist/cesium/` contained **only** a `node_modules/…` subtree.

Why it looked browser-specific ("works in Firefox / works for me"): it's actually broken for **everyone** on a fresh cache — the reporters who saw a working map had an *earlier, correctly-pathed* deploy still in their HTTP cache. The `$lib:web` Edge user in #1175 hit the same thing (the unhandled Cesium `RequestErrorEvent`s are these failed worker/tile fetches).

This regressed with the vite 7 / `vite-plugin-static-copy` v4 bump (the "core update" the reporter suspected). The plugin's `stripBase` escape hatch is static and would flatten Cesium's nested `Assets/Textures/NaturalEarthII/z/x/y.jpg` tree, so it isn't a clean fix.

## Fix

Replace the static-copy block with a small inline Vite plugin (`cesiumStaticAssets`) that:

- **build** — recursively `fs.cp`s `Workers / ThirdParty / Assets / Widgets` into `<outDir>/cesium/<Subdir>/…`. Deterministic, package-manager-agnostic, structure-preserving.
- **dev** — serves `/cesium/*` straight from the Cesium package (dev was broken the same way — the plugin's dev fileMap keys off the same nested dest).
- **fails loud** at config time if the prebuilt subdirs are missing, so a silent miss can't ship a broken map again.

Single-file change. `CESIUM_BASE_URL` stays `/cesium`. Turbo self-heals: the `vite.config.ts` change busts the viewer build cache key, so the next build re-copies the assets to the correct paths.

## Verification

- `tsc --noEmit` on the new config: clean.
- Replayed the exact `closeBundle` copy against the installed `cesium@1.142.0`: `ion-credit.png`, `bing_maps_credit.png`, `approximateTerrainHeights.json`, the nested `Textures/NaturalEarthII/0/0/0.jpg`, and `Widgets/widgets.css` all land flat under `dist/cesium/…`; **no `node_modules` nesting**; Workers = 110 files.

## Notes

- `vite-plugin-static-copy` is now unused in code (still declared in `apps/viewer/package.json`). Left in place to avoid lockfile churn — can be pruned in a separate dependency-hygiene pass.
- No changeset: `@ifc-lite/viewer` is the deployed app, not a release-managed library (same as #1174/#1178).

## Summary by CodeRabbit

* **Chores**
  * Improved the viewer application's build infrastructure by optimizing how Cesium assets are validated, compiled, and served during development and production builds. This enhancement streamlines asset handling and delivery across different environments.

### DIFF
--- a/apps/viewer/vite-plugins/cesium-assets.ts
+++ b/apps/viewer/vite-plugins/cesium-assets.ts
@@ -1,0 +1,91 @@
+import path from 'node:path';
+import fs from 'node:fs';
+import { createRequire } from 'node:module';
+import type { Plugin } from 'vite';
+
+const require = createRequire(import.meta.url);
+
+// CesiumJS loads its prebuilt Workers / Assets / Widgets / ThirdParty at runtime
+// from CESIUM_BASE_URL (= '/cesium', set in vite.config's `define`), so they must
+// sit at  <outDir>/cesium/<Subdir>/…  and be served at  /cesium/<Subdir>/…  in dev.
+//
+// We copy them ourselves rather than via vite-plugin-static-copy: in its v4 line
+// the plugin rebuilds each destination from the file's path *relative to the Vite
+// root*, so an absolute node_modules source lands every file under
+//   <outDir>/cesium/node_modules/.pnpm/cesium@<v>/node_modules/cesium/Build/Cesium/…
+// instead of <outDir>/cesium/…  → every /cesium/Assets|Workers/… request 404'd in
+// production and the Cesium globe never rendered (#1139). Its stripBase escape
+// hatch is static and would flatten Cesium's nested Assets/Textures tree. A plain
+// recursive copy is deterministic and package-manager-agnostic.
+const CESIUM_SUBDIRS = ['Workers', 'ThirdParty', 'Assets', 'Widgets'] as const;
+
+const CESIUM_DEV_MIME: Record<string, string> = {
+  '.js': 'text/javascript', '.mjs': 'text/javascript', '.css': 'text/css',
+  '.json': 'application/json', '.wasm': 'application/wasm', '.png': 'image/png',
+  '.jpg': 'image/jpeg', '.jpeg': 'image/jpeg', '.gif': 'image/gif',
+  '.svg': 'image/svg+xml', '.xml': 'application/xml',
+};
+
+export function cesiumStaticAssets(): Plugin {
+  const cesiumBuild = path.join(
+    path.dirname(require.resolve('cesium/package.json')),
+    'Build',
+    'Cesium',
+  );
+  // Fail loud at config time if the prebuilt assets aren't where we expect —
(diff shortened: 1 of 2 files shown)


louistrue:
Addressed review feedback in `577edddc`:

- **Codex** (AGENTS.md ~400-line split): extracted the Cesium plugin into `apps/viewer/vite-plugins/cesium-assets.ts`; `vite.config.ts` drops from 407 → 328 lines. The plugin now resolves the output dir from `config.root` rather than `__dirname`, so it's independent of its own file location.
- **CodeRabbit** (path containment): the dev `/cesium` middleware now matches on a `path.sep` boundary (`=== cesiumBuild || startsWith(cesiumBuild + sep)`), so a sibling like `…/Cesium-backup` can't satisfy a bare prefix.
- **CodeRabbit** (TOCTOU): the dev file stream now has an `error` handler so a vanished/locked file returns 500 instead of crashing the dev server.

Production behaviour is unchanged (recursive `fs.cp` of the four subdirs flat under `<outDir>/cesium`); the rest is dev-only hardening.

### On the preview 404
The raw `*.vercel.app` deployment URL 404s at root — but that's **not** this change. The current production immutable URL (`ifc-lite-6eey…`, the deploy serving `www.ifclite.com`) 404s at `/` and `/index.html` identically; only the assigned domain `www.ifclite.com` serves content. The Vercel build for this PR succeeded (`29 successful`, fresh `vite build`). Because all raw preview URLs are gated, the cesium fix can only be confirmed on `www.ifclite.com` once this lands in production — I'll verify `https://www.ifclite.com/cesium/Assets/Images/ion-credit.png` returns 200 right after the prod deploy.

PR #1184, opened 2026-06-17
perf(geometry): cheap FacetedBrep dedup hash — re-enable content-dedup as a net win

louistrue:
Re-enables content-dedup on the production geometry paths (native pool + wasm), which #1177 turned off. #1177's own note said *"flip back to `true` once `item_signature` walks via the cached fast paths"* — this does exactly that.

## The problem #1177 found
Content-dedup (skip re-meshing structurally-identical representation items) builds a 128-bit key by recursively `decode_by_id`-ing the **entire** item subtree — every face, loop, and point. On steel models that hash cost *more* than the meshing it skipped, so dedup made loads slower.

## The fix
`item_signature` now hashes `IfcFacetedBrep` through the **same cached byte-level fast paths the mesher uses** (`get_entity_ref_list_fast` / `get_face_bound_fast` / `get_polyloop_coords_cached`) — zero `decode_by_id` per point.

Measured on a ~50k-part steel model (3.1M points / 2.75M polyloops / 48.6k breps):

| | recursive (old) | byte-level (new) |
|---|---|---|
| brep hash | ~8 s | **~2 s** |
| meshing it skips | ~5 s | ~5 s |
| dedup A/B | **0.9× (net loss)** | **1.3× (net win)** |

Byte-identical: **0 / 50,173** fingerprint mismatches.

## Safety for non-brep models
`item_dedup_key` gates the hash to `IfcFacetedBrep`. Procedural-geometry models (extrusions / CSG) — whose recursive hash is what made dedup a net loss — now skip the hash entirely and pay nothing, so they can't regress. The `IfcMappedItem` instancing cache is unaffected.

## Honest scope
This is a **real but modest** lever, not a silver bullet: the duplicate share on this model is ~33% by meshing cost, so dedup is capped around ~1.5×. Profiling also showed the per-job affinity hash (`geometry_routing_key`) runs in the wasm pre-pass *unconditionally* (even with dedup off) and is a larger browser-side cost — a good follow-up (extend the cheap hash to the rep level + skip non-dedup-eligible types there). A browser re-profile on the preview is the next confirmation step.

## Test plan
- 61 geometry test binaries green (incl. synthetic dedup fixtures, `test_tekla`, wall-opening regressions).
- Native A/B byte-identical on the 50k-part model.
- wasm builds clean; parity holds by construction (dedup returns cached meshes byte-identical to fresh; meshing untouched; hash is deterministic byte-parse + integer fold).

### DIFF
--- a/.changeset/cheap-brep-dedup-hash.md
+++ b/.changeset/cheap-brep-dedup-hash.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Re-enable content-dedup on the production geometry paths with a cheap structural hash — it's now a net speedup on steel-heavy models instead of the slowdown that forced it off.
+
+Content-dedup (skip re-meshing structurally-identical representation items) was disabled in the previous release because its 128-bit structural key recursively decoded the *entire* item subtree — every face, loop, and point — costing more than the meshing it saved. `item_signature` now hashes `IfcFacetedBrep` (the dominant type in Tekla steel exports, where thousands of geometrically identical plates and bolts each get their own representation) through the same cached byte-level fast paths the mesher uses, with zero `decode_by_id` per point. On a ~50k-part steel model the brep hash dropped from ~8 s to ~2 s — below the ~5 s of meshing it skips — flipping dedup from a 0.9× loss to a 1.3× win, with byte-identical geometry (0 fingerprint mismatches over 50k elements).
+
+Dedup is gated to the cheap (brep) types in `item_dedup_key`, so procedural-geometry models — the ones whose recursive hash cost more than it saved — skip the hash entirely and pay nothing. The separate `IfcMappedItem` instancing cache is unaffected.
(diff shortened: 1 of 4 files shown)


louistrue:
## Browser re-profile (Playwright, 722 MB steel model, dev build, same session)

Byte-identical geometry across all three (140,755 meshes):

| build | first-jobs | stream complete | peak JS heap |
|---|---|---|---|
| main today (dedup off + recursive hash) | 13.0 s | **78.0 s** | 2,460 MB |
| dedup off + cheap hash | 10.0 s | 75.3 s | 2,347 MB |
| **this PR** (dedup on + cheap hash) | 10.0 s | **67.5 s** | 2,058 MB |

**Net vs main: 78.0 s → 67.5 s = −13.5% (−10.5 s), byte-identical, −400 MB peak JS heap.**

Breakdown: the cheap affinity hash trims the unconditional pre-pass job-tagging (−3 s, first-jobs 13→10 s); the dedup cache trims duplicate brep meshing (−7.8 s). The −400 MB heap drop also widens OOM headroom on the ~6 GB tab.

Honest scope: this is the dedup-ceiling win (~33% duplicate meshing cost on this model). The remaining browser cost is the wasm per-element floor (~50 s of streaming after first-batch) and the ~17 s pre-pass — separate follow-ups.

PR #1190, opened 2026-06-18
fix(wasm): recover from transient WASM engine-load failures + humanise error

louistrue:
## What

Fixes PostHog error-tracking issue [`019ed949`](https://eu.posthog.com/project/199147/error_tracking/019ed949-8d6b-7a72-b2a5-5c2f7be1d37f): `TypeError: Failed to execute 'compile' on 'WebAssembly': HTTP status code is not ok` (1 occurrence, Edge 149 / Windows 10, handled).

## Root cause

The geometry and parser workers both `init()` the same `ifc-lite_bg.wasm`. wasm-bindgen's streaming loader (`__wbg_load`) **rethrows** on a non-OK HTTP status — it only falls back to `arrayBuffer()`+`instantiate` for the *wrong-MIME* case. So this is a genuine failed download of the engine binary, surfaced as a cryptic `TypeError` and captured at `useIfcLoader`'s outer catch.

With **no service worker** in play and a **single occurrence in 30 days** (the deployed wasm path works for everyone else), this is a transient/environmental failure — a cold CDN edge, a mid-deploy race, or a proxy/antivirus blocking the `.wasm` (plausible on corporate Edge/Windows). The right fix is **resilience + clarity**, not a config change.

## Changes

- **Workers** (`packages/{geometry,parser}/src/*.worker.ts`): `init()` retries **once** (300 ms backoff) on a fetch/HTTP-shaped failure before propagating. Genuine compile/validation errors still throw immediately, so real bugs aren't masked. Safe because `__wbg_init` only short-circuits on `wasm !== undefined`.
- **`apps/viewer/src/lib/load-errors.ts`** (new): `classifyLoadError()` → stable `error_kind` (`wasm_engine_load` / `out_of_memory` / `cancelled` / `unknown`); `formatLoadError()` → actionable message ("Couldn't load the 3D geometry engine… reload the page") instead of the raw TypeError.
- **`useIfcLoader.ts`**: wired the humanizer into all failure surfaces and tagged captured exceptions with `error_kind` so future occurrences are filterable in PostHog.

## Verification

- New unit test `load-errors.test.ts` (10 cases incl. the exact captured message) — passing.
- `tsc --noEmit` clean for all 4 touched files.
- Note: workers not exercised end-to-end (no wasm build in a fresh worktree); retry logic is simple and typechecks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebAssembly engine initialization resilience: geometry and scan workers now retry once when the WASM engine fetch/load fails transiently (e.g., CDN edge races, blocked downloads, proxy/antivirus interference).
  * Updated viewer error handling to show actionable guidance (such as reloading) instead of raw technical errors.
  * Added consistent error categorization to improve monitoring and troubleshooting.
* **Tests**
  * Added coverage for transient WASM retry behavior and for user-facing error classification/formatting.

### DIFF
--- a/.changeset/wasm-engine-load-resilience.md
+++ b/.changeset/wasm-engine-load-resilience.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/geometry": patch
+"@ifc-lite/parser": patch
+"@ifc-lite/viewer": patch
+---
+
+Recover from transient WASM engine-load failures and humanise the error.
+
+When the `ifc-lite_bg.wasm` binary fails to download (non-OK HTTP status, a cold
+CDN edge, a mid-deploy race, or a blocking proxy/antivirus), wasm-bindgen's
+streaming loader rethrows a cryptic `Failed to execute 'compile' on
+'WebAssembly': HTTP status code is not ok`. The geometry and parser workers now
+retry `init()` once on such fetch/HTTP-shaped failures, and the viewer maps the
+failure to actionable guidance ("reload the page") instead of surfacing the raw
+TypeError. Captured exceptions are tagged with a stable `error_kind` for triage.
--- a/apps/viewer/src/hooks/useIfcLoader.ts
+++ b/apps/viewer/src/hooks/useIfcLoader.ts
@@ -53,6 +53,7 @@ import { getGlobalRenderer } from './useBCF.js';
 import { extractModelGeoref, alignGeometryToReference, findReferenceGeorefModel } from './ingest/federationAlign.js';
 import { toast } from '../components/ui/toast.js';
 import { posthog } from '../lib/analytics.js';
+import { classifyLoadError, formatLoadError } from '../lib/load-errors.js';
 
 /**
  * Where a {@link useIfcLoader.loadFile} call should land the model.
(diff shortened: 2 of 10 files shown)


louistrue:
Both findings addressed in 09ff1fa5:
- **parser `initPromise` rejection memoisation** — now reset to `null` on failure so a later `ensureWasmScanApi()` can recover.
- **Retry test coverage** — extracted the retry into a DI'd `wasm-init-retry.ts` helper in both packages with vitest suites (success/no-retry, one transient retry, no-retry for genuine compile errors, second-failure propagation, retry delay).

louistrue:
Addressed the follow-up in the latest commit: narrowed `isTransientWasmLoadError` to transport/network signals (`HTTP status code is not ok`, `Failed to fetch`, `NetworkError`, `Load failed`) and added an explicit fail-fast guard for `CompileError`/validation errors (wrong magic header, corrupt bytes) — retrying those would just refetch the same bad bytes. Dropped the broad `'…' on 'WebAssembly'` and bare `streaming` matches; the real captured error still matches via `HTTP status code is not ok`. Tests cover Chromium/Firefox/Safari fetch failures and corrupt-module CompileErrors in both packages.

PR #1193, opened 2026-06-18
perf(geometry): parametric placement-frame rectangular-opening fast path (flag-gated, OFF)

louistrue:
## What

An analytic fast path that subtracts **rectangular openings** as **exact parametric boxes in the host wall's own placement frame** — handling **rotated walls**, the dominant case the existing axis-aligned `rect_fast` cannot. **Flag-gated, default OFF.**

The key idea: the frame `R` and box extents come from the **IFC parametrics** (`IfcRectangleProfileDef` / axis-aligned `IfcArbitraryClosedProfileDef` + composed placements, × the length-unit factor), **not inferred from the f32 mesh**. So the cut is the analytic *box-minus-boxes* solid — ground-truth exact and watertight by construction. (An earlier mesh-inference attempt — averaging triangle normals + AABB'ing the opening mesh — was ~12% wrong on real geometry; reading the exact parametric data is the fix.)

## How

- `rect_fast::param_enabled()` — new flag, **DEFAULT OFF** (`IFC_LITE_RECT_PARAM=1`), plus a wasm toggle (`setRectParamFastPath`) so native + wasm flip in lockstep.
- `VoidContext` carries probed + reconciled host/opening boxes (captured in `build_void_context`). `try_param_rect_cut` runs in `apply_void_context` before the exact path with: host reconciliation (param box == real mesh), opening reconciliation (param box == opening mesh), shared-frame, in-bounds, no-overlap, and a **watertightness self-check that DEFERS** rather than emit a bad cut. Any precondition miss → exact kernel.
- Output is a **local-frame mesh** (small positions + `origin` = wall centre) so it survives f32 storage + `clean_degenerate` at national-grid magnitude. Every origin consumer (renderer, cache, clash, export, raycast, `geom_hash`) already folds it.
- Deterministic FMA-free f64 + snap grid → **byte-identical native==wasm** by construction.

## Validation

| | Flag OFF (default) | Flag ON |
|---|---|---|
| Geometry lib + void regression | **369 + all pass — fully inert** | — |
| Hosts fired in `process_element_with_voids` | — | fires on the rotated-rect-wall subset |
| Non-watertight | — | **0** (self-check defers) |
| Cut vs analytic ground truth (fired subset) | — | **100%** |

On a national-grid architectural model and a large architectural model, the fast path fires on the rotated rectangular-wall subset with **0 non-watertight** outputs and matches the **analytic ground truth 100%** — and is *more* correct than the exact kernel, which over-cuts engulfing openings on a fraction of these walls.

## Flip-gates (all addressed; flag stays OFF until a maintainer flips it)

1. **wasm toggle** — `setRectParamFastPath` lets native + wasm enable in lockstep.
2. **consumer origin-fold audit** — verified: all world-space readers already fold the per-element `origin`.
3. **determinism** — CI-safe in-crate determinism unit test (bit-identical across runs).

## Notes

- No changeset: the change is rust-only and **OFF by default** (no published behaviour change). Happy to add one if the release gate wants it.
- Test harnesses default to a **public** fixture; real models are passed via `MEASURE_FIXTURE` (no private paths committed).
- Scope: an **architecture-model** win (rotated rect walls). Steel models have no rotated-box hosts, so they are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added a parametric fast path for rectangular opening processing to improve geometry computation performance.
  * Exposed `RectParam` as a publicly available geometry parameter type.
  * Added API controls (native and WASM) to enable/disable the parametric fast path.
* **Bug Fixes / Reliability**
  * Improved reliability of generated results by enforcing watertight, deterministic mesh output for supported rectangular openings.

### DIFF
--- a/packages/wasm/pkg/ifc-lite.d.ts
+++ b/packages/wasm/pkg/ifc-lite.d.ts
@@ -221,6 +221,18 @@ export class IfcAPI {
    */
   clearPrePassCache(): void;
   /**
+   * Enable or disable the PARAMETRIC rectangular-opening fast path (the
+   * placement-frame, ground-truth-exact analytic cut) for `processGeometryBatch`.
+   *
+   * DEFAULT OFF. This is the wasm-side toggle that lets native and wasm flip the
+   * flag in LOCKSTEP — the byte-identical native==wasm contract requires both
+   * targets take the same path, and wasm has no env to read `IFC_LITE_RECT_PARAM`.
+   * The path subtracts rectangular openings as exact parametric boxes in the host's
+   * own placement frame (rotated walls included), deferring any non-clean case to
+   * the exact kernel. Pass `true` before `processGeometryBatch`.
+   */
+  setRectParamFastPath(enabled: boolean): void;
+  /**
    * Select the tessellation detail level applied by every subsequent
    * `processGeometryBatch` call (issue #976, step 4).
    *
@@ -929,6 +941,7 @@ export interface InitOutput {
   readonly ifcapi_setComputeGeometryHashes: (a: number, b: number, c: number) => void;
   readonly ifcapi_setEntityIndex: (a: number, b: number, c: number, d: number, e: number, f: number, g: number) => void;
   readonly ifcapi_setMergeLayers: (a: number, b: number) => void;
+  readonly ifcapi_setRectParamFastPath: (a: number, b: number) => void;
   readonly ifcapi_setTessellationQuality: (a: number, b: number, c: number, d: number) => void;
   readonly ifcapi_version: (a: number, b: number) => void;
   readonly meshOutline2d: (a: number, b: number, c: number, d: number, e: number, f: number) => number;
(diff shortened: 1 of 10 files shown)


louistrue:
Thanks @coderabbitai — all four findings addressed in `75ea284c` (+ the `.d.ts` sync in `288ba218`):

- **`origin == 0` gate (the critical one)** — correct: `local_frame_enabled()` is ON by default in wasm, so the path never fired in the viewer. `try_param_rect_cut` is now **origin-aware** (frames the cut around `center − mesh.origin`), working for both world-coord (native) and local-frame (wasm) meshes. It also fires *more* with the local frame on — relativized small-coord meshes reconcile more often (219 vs 183 on the national-grid model).
- **RTC center** — fixed: the parametric center now subtracts `rtc_offset` like the mesh/bounds paths.
- **Thickness-axis penetration** — fixed: defer unless the opening's penetration axis is the host's thin axis (else the through-extension could wipe out a full slab).
- **Multi-solid bodies** — fixed: `parametric_rect_probe` defers a body with more than one representation item.

Validated: flag OFF unchanged (369 lib + void regression); flag ON in **both** native (origin 0) and local-frame (origin ≠ 0) modes — fires, 0 non-watertight.

Issue #1167, opened 2026-06-17
Weird wall hole cutting


Sonderwoods:
I am seeing this in the newest build. Voids cut from walls are larger than the windows.

<img width="1017" height="507" alt="Image" src="https://github.com/user-attachments/assets/904c97f8-8fa1-462a-b849-9994619f52bf" />

louistrue:
Fixed after #1169? I verified across ~1,500 openings in 13 models that normal windows are cut identically by the fast path and the exact kernel, so it might be a separate, older opening-vs-window sizing issue rather than a "new fast-path bug".

Could you attach the IFC (or a single affected wall) if the oversized voids persists? That'd let me pin down the exact element.

Sonderwoods:
Still not working

<img width="1985" height="1219" alt="Image" src="https://github.com/user-attachments/assets/4c3cec15-8f96-4dfc-8e03-ae92f46b03db" />

Through IfcOpenShell:

<img width="1456" height="812" alt="Image" src="https://github.com/user-attachments/assets/c099f016-3914-40cb-9956-2c2c1844fe0b" />

Sending an IFC link in PM

Issue #1007, opened 2026-06-08
Roof Brep with openings renders extra inward-facing faces


Blogbotana:
## Summary
The roof in `Roof-01_BCAD.ifc` renders with **extra inward-facing faces** around
its openings. The openings ARE cut through correctly — the defect is spurious /
wrong-winding faces (normals pointing into the solid).

This is a **long-standing bug, present in BOTH production and `main`** (not a
recent regression).

## Repro
1. Load `Roof-01_BCAD.ifc` in the viewer.
2. Inspect roof `IfcRoof` `3j5N7RPpvA1PwiM18Uv4fc` (`roof_02_a`).
3. The two openings are cut, but extra inward-facing faces appear at the cut.

## Details
- Host: `IfcRoof` `#1112`, body `'Body'/'Brep'` → `IfcFacetedBrep`. Units: metres.
- Two voids via `IfcRelVoidsElement` → openings `#2150` (`'Body'/'Brep'`) and
  `#2154` (`'Body'/'SweptSolid'`).
- Symptom is **winding / normal orientation** (faces inward) at the opening cut,
  not an under-cut. The host is a faceted Brep, so the opening cut goes through
  the CSG/void path on a Brep host — the inward faces are likely reveal/cut faces
  emitted with inverted winding (or the Brep's own faces near the cut).

## Notes
- Unrelated to PR #1005 (#977 CSG work): that PR isn't deployed to prod, the bug
  is already in prod, and #1005 doesn't touch face winding nor the `IfcRoof`
  analytic/Brep path.
- Distinct from #1006 (which is prod-OK / main-only) — this one reproduces in
  prod too, so it's a separate, older orientation bug on Brep hosts with voids.

Good starting point: check the winding of reveal/cut faces generated when an
opening is subtracted from a faceted-Brep host (vs an extruded host).

<img width="925" height="548" alt="Image" src="https://github.com/user-attachments/assets/41e5cd38-2ac3-4e96-9d04-bf0a2bb11dac" />

Blogbotana:
**Fixed by #1005.** Verified on a build of that branch: the extra inward-facing
faces around this roof's openings are gone — the roof now cuts cleanly.

Why it's fixed: this opening sits on a faceted-Brep roof and its cutter is
tilted, so the analytic axis-aligned-box clip was both over-cutting and
fabricating reveal/cap faces with the wrong winding. PR #1005's new
**geometry-driven routing** detects that the opening's world AABB over-
[...]

Issue #960, opened 2026-06-06
IfcWallStandardCase geometry anomalies or missing


xyzbety:
In the attached model（House）, several IfcWallStandardCase entities exhibit geometry issues. Entities 2FzACFrWn78vGKEK4Md6ha and 2PDtSyZL10pweyEST_guOH display anomalous extruded geometry, while entities 0wFZS1FlX4uQG90eXr1foJ, 1NKEanv7HDEOdbfk8nuT6f, and 2l9upSYxz4xfNDmVZnFjG3 fail to parse or render entirely. The issue appears to be related to the processing of IfcBooleanClippingResult and IfcExtrudedAreaSolid.

[House.zip](https://github.com/user-attachments/files/28664097/House.zip)

<img width="2870" height="1635" alt="Image" src="https://github.com/user-attachments/assets/5b496605-32f1-497f-bc47-c854eb99a365" />

<img width="2880" height="1730" alt="Image" src="https://github.com/user-attachments/assets/1d6574c0-bd2b-46c8-84ed-62b9c55a1484" />

PR #1223, opened 2026-06-19
fix(geometry): cut plan-rotated walls in their own axis-aligned frame…

Sonderwoods:
… (#1167)

This solves my buggy project, although the solutions seems a bit complex.

--

Verified against a real wall isolated from the reporter's model (IfcWallStandardCase 3YZDoaW3XDrOwfaQwnDhmu): a wall rotated ~13 deg in plan at world coords ~150 m with 5 openings (4 window boxes + 1 brep door) was both over-cut (removed 22.5 of 26.0 m3 -- 86%) and fragmented (~236 unpaired edges). Two causes, both from cutting a tilted box in world space:

  1. The world-axis-aligned-AABB path fires within ~18 deg of an axis, and the AABB of a rotated box is bigger than the box (oversized, grid-aligned hole).
  2. Even via the exact mesh subtract, a tilted cut at large world coordinates (~150 m, f32 ~ 15 um) over-cuts and fragments.

Earlier attempts failed because they fought the rotation in world space (an inferred frame, a synthesized world box, or rotating the already-placed jittered mesh). This cuts a plan-rotated wall in its OWN axis-aligned, origin-centred frame: host + openings are rotated into that frame (world-axis-aligned and near the origin, where the exact subtract is clean and f32-precise), cut there (clean boxes reclassify to the watertight rect_fast path; brep/curved openings keep their mesh), then the result is rotated back. The orthonormal, origin-centred round-trip is identity for untouched geometry.

The frame comes from an opening's own (non-axis-aligned, ~horizontal) depth axis, so it is scoped to plan-rotated walls -- axis-aligned walls and roof/floor/sloped openings keep their existing path (#1007, #960 unaffected).

Real wall after the fix: removed ~13 m3, watertight. Full ifc-lite-geometry suite green; wasm32 release builds clean. Adds issue_1167_real_wall (the isolated real wall, inline) plus the synthetic over-cut and multi-angle clean-cut guards.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where window and door openings would become oversized and fragmented when cutting through walls that are rotated in their horizontal plane. This problem was especially noticeable for walls positioned at large coordinate values.

* **Tests**
  * Added regression tests to verify correct opening removal for various rotation angles and wall configurations, including both synthetic and real-world architectural scenarios.


Sonderwoods:
Wait a sec - the outer surface of the wall is correctly made, but seems like normals are pointing inwards sometimes so that it doesnt render correctly in the web viewer. I'll add an update.

Sonderwoods:
looks ok now in my 1 test file. Please take it for a spin @louistrue

louistrue:
Thanks @Sonderwoods — this is the right fix and the diagnosis was spot on. It needed a rebase: it was branched before #1209 added the `param` field to `VoidContext`, so it stopped compiling against current `main` (the red CI). I've rebased + compile-fixed it as #1259, **authored to you**, with the full root-cause writeup and verification (geometry suite 507/0/14, confirmed visually on the reporter's model). Closing this in favour of #1259 to keep one green PR — your work, carried forward. 🙏

PR #1234, opened 2026-06-20
feat(mutations,export): retype an entity's IFC class (reassign class)

### LINKED ISSUE #1231: Feature request: retype an entity's IFC class via @ifc-lite/mutations
## Context

I use IFC editor on top of `@ifc-lite/mutations` (`MutablePropertyView` + `StoreEditor`), reading on demand from the parser store and round-tripping edits into valid STEP via `new StepExporter(store, view)` + `export({ applyMutations: true })`. Editing property/quantity/attribute values works great this way.

## What I need

Let users **change the IFC class of selected entities** — retype them in place.

Real-world driver: MCAD exporters collapse semantics. **SolidWorks (and others) export essentially everything as `IfcBuildingElementProxy`.** The user then wants to manually reassign the correct class — `IfcColumn`, `IfcBeam`, `IfcMember`, `IfcPlate`, etc. — on a hand-picked selection of elements (not a federated/criteria-wide bulk op; the choice is per selected element).

## Current state (as of `@ifc-lite/mutations@1.15.4`, `@ifc-lite/export@1.19.8`)

There seems to be no API for this:
- `BulkAction` supports only `SET_PROPERTY` / `DELETE_PROPERTY` / `SET_ATTRIBUTE` (`name`/`description`/`objectType`).
- `StoreEditor` has `addEntity` / `removeEntity` / `setAttribute` / `setPositionalAttribute` / `addPropertySet` / `addQuantitySet` — but no `setType` / `changeType`.
- `MutablePropertyView` has `createEntity` / `deleteEntity` — but no retype.

`addEntity(newType) + removeEntity(old)` isn't viable: it allocates a new expressId, so geometry/placement/representation and every `IfcRel*` reference still point at the old entity.

## Question

**Is entity retype feasible to add to the mutation overlay (materialized at export)?** Specifically:
1. An API like `StoreEditor.setEntityType(expressId, newType)` or a `BulkAction { type: 'SET_ENTITY_TYPE', entityType }`, applied to a specific set of expressIds and emitted by `StepExporter`.
2. How would you want to handle attribute-layout compatibility between source/target types? For the common building-element subtypes (Proxy/Beam/Column/Member/Plate/Wall) the IfcElement attribute list is identical, so a keyword swap (plus an optional `PredefinedType`) looks safe — but a general solution needs a compatibility check.
3. Geometry/representation and `IfcRel*` references stay on the same expressId, so they should carry over unchanged — does that match your model?

Happy to help test. Thanks!

louistrue:
Closes #1231.

Lets consumers of `@ifc-lite/mutations` **change the IFC class of selected entities** — retype them in place — materialized into valid STEP at export. Real-world driver: MCAD exporters (SolidWorks etc.) collapse everything to `IfcBuildingElementProxy`; users want to reassign the correct class (`IfcColumn`, `IfcBeam`, `IfcMember`, `IfcPlate`, …) per hand-picked element.

## How it works — mirrors IfcOpenShell

I read the actual `ifcopenshell.util.schema.reassign_class` / `ifcopenshell.api.root.reassign_class` source. The model it uses:

1. **Same id preserved** — IfcOS removes + recreates with `id=info["id"]`. ifc-lite's overlay keeps the same expressId naturally, so geometry / placement / representation and every `IfcRel*` reference (all keyed by `#id`) carry over unchanged — no relinking needed.
2. **Attributes copied BY NAME against the *target* class declaration** — attributes the target doesn't declare are dropped; missing ones become `$`.
3. **`PredefinedType` enum is validated** against the target class's domain; an out-of-domain value is dropped, and an unknown explicit override falls back to `USERDEFINED` + `ObjectType`.

This is strictly more correct than a naive keyword swap, which would emit invalid IFC in IFC2X3 (where `IfcBuildingElementProxy` carries a trailing `CompositionType` that `IfcColumn`/`IfcBeam`/… don't have). In IFC4/IFC4X3 the building-element subtypes share an identical 9-attribute layout, so the re-layout is an identity and only the class keyword (plus optional PredefinedType) changes — the argument tokens are preserved byte-for-byte.

## API (answers the issue's 3 questions)

1. **API** — both forms requested:
   - `StoreEditor.setEntityType(expressId, newType, { predefinedType? })` → `boolean`
   - `MutablePropertyView.setEntityType(entityId, newType, predefinedType?)`
   - `BulkAction { type: 'SET_ENTITY_TYPE', entityType, predefinedType? }`
   - Emitted by `StepExporter.export({ applyMutations: true })`.
2. **Attribute-layout compatibility** — schema-driven name-based re-layout against the target class (IFC2X3/IFC4/IFC4X3 layouts handled per-schema); enum validation for `PredefinedType`. Unknown target class → graceful keyword-only swap.
3. **Geometry / `IfcRel*`** — stay on the same expressId, carry over unchanged. ✔️ (covered by a full round-trip test asserting `#42`'s geometry, placement, and `IfcRelContainedInSpatialStructure` references survive a Proxy→Column retype).

## Tests
- `packages/mutations/test/retype.test.ts` (10): StoreEditor/view API, validation, normalizer, bulk action, `clear()`, export→import replay.
- `packages/export/src/retype.test.ts` (10): unit `retypeStepLine` (identical-layout swap, dropped out-of-domain enum, valid override, USERDEFINED fallback, IFC2X3 drop-trailing, unknown-class keyword swap) + full `StepExporter` round-trips (refs preserved, PredefinedType override, new-entity retype, retype + attribute edit combined).

Full suites green (mutations 30, export 123); full `pnpm build` (all 36 packages) and `check:api-surface` pass.

## Notes / follow-ups
- Scoped to `@ifc-lite/mutations` + `@ifc-lite/export` exactly as the issue requests. The SDK `bim.store.*` facade could expose `setEntityType` for ergonomics in a follow-up (the SDK already wires the registry normalizer, so canonical-name resolution works there).
- `BulkAction`'s pre-existing `SET_ATTRIBUTE` handler is a no-op stub; left as-is (out of scope).

## Summary by CodeRabbit

* **New Features**
  * Added “retype” (IFC class reassign) support to change an entity’s class in-place while preserving expressId, relationships, and geometry/placement.
  * Added UI to pick a target IFC class and optional `PredefinedType`, including an export reassignment badge.
  * Extended mutation and bulk-edit APIs to set the entity type (with optional `predefinedType`) and reflect it in the viewer.
  * Added per-entity display-class type overrides for consistent UI/export presentation.
* **Improvements**
  * STEP export now applies retyping first, then re-lays out attributes to match the target class layout.
  * `PredefinedType` is validated/sanitized against the target domain with safe fallbacks for unknown values.
* **Tests**
  * Added unit and end-to-end coverage for retype/materialization and mutation behavior, including undo/redo and edge cases.

### DIFF
--- a/.changeset/entity-retype-reassign-class.md
+++ b/.changeset/entity-retype-reassign-class.md
@@ -1,0 +1,22 @@
+---
+"@ifc-lite/mutations": minor
+"@ifc-lite/export": minor
+---
+
+Add entity retype (reassign class) to the mutation overlay.
+
+`StoreEditor.setEntityType(expressId, newType, { predefinedType? })` and
+`MutablePropertyView.setEntityType(...)` change an entity's IFC class in place,
+and a new `BulkAction { type: 'SET_ENTITY_TYPE', entityType, predefinedType? }`
+applies it to a selection. `StepExporter` materializes the retype on export.
+
+The entity keeps its expressId, so geometry, placement, representation and every
+`IfcRel*` reference (all keyed by `#id`) carry over unchanged. Attributes are
+re-laid-out by name against the target class's declared layout — dropping
+attributes the target lacks (e.g. IFC2X3 `CompositionType`) and validating
+`PredefinedType` against the target enum (an unknown override falls back to
+`USERDEFINED` + `ObjectType`). This mirrors IfcOpenShell's
+`ifcopenshell.util.schema.reassign_class`. Intended for compatible
+reassignments such as the building-element subtypes that share the IfcElement
+layout (`IfcBuildingElementProxy` ↔ `IfcColumn`/`IfcBeam`/`IfcMember`/
+`IfcPlate`/`IfcWall`).
--- a/.changeset/entity-retype-reassign-class.md
+++ b/.changeset/entity-retype-reassign-class.md
@@ -1,9 +1,16 @@
 ---
 "@ifc-lite/mutations": minor
 "@ifc-lite/export": minor
+"@ifc-lite/data": minor
+"@ifc-lite/cache": patch
 ---
 
 Add entity retype (reassign class) to the mutation overlay.
+
+`EntityTable` gains an additive `setTypeOverride(expressId, typeName | null)` so
+a host (the viewer) can reflect a pending retype live in `getTypeName` /
+`getTypeEnum` without rebuilding the table; the original columnar type is left
+intact.
 
 `StoreEditor.setEntityType(expressId, newType, { predefinedType? })` and
 `MutablePropertyView.setEntityType(...)` change an entity's IFC class in place,
--- a/apps/viewer/src/components/viewer/PropertiesPanel.tsx
+++ b/apps/viewer/src/components/viewer/PropertiesPanel.tsx
@@ -1907,6 +1907,19 @@ function EntityDataSection({
     return query.entity(entityRef.expressId);
   }, [query, entityRef.expressId]);
 
+  // Overlay-aware display class: a pending retype (setEntityType) reassigns the
+  // class in place, so the header reflects the new class immediately — before
+  // the model is re-exported / reloaded. Re-evaluates on `mutationVersion`.
+  const headerMutationViews = useViewerStore((s) => s.mutationViews);
+  const headerMutationVersion = useViewerStore((s) => s.mutationVersion);
+  const displayType = useMemo(() => {
+    let mid = entityRef.modelId;
+    if (mid === 'legacy') mid = '__legacy__';
+    const pending = headerMutationViews.get(mid)?.getEntityTypeMutation?.(entityRef.expressId)?.newType;
+    return pending ?? entityNode?.type ?? '';
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [headerMutationViews, headerMutationVersion, entityRef.modelId, entityRef.expressId, entityNode]);
+
   // Get properties and quantities
   const properties: PropertySet[] = useMemo(() => {
     if (!entityNode) return [];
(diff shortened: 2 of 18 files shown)


louistrue:
Thanks @coderabbitai — addressed in `b8cd4fcc`:

**Fixed**
- **Validate at the view boundary** — `MutablePropertyView.setEntityType` now runs the `Ifc*` keyword regex, so the `BulkQueryEngine` path (which bypasses `StoreEditor`) can't record an invalid class. The bulk engine surfaces it as a per-entity error.
- **Repeated retype on a created entity** — `oldType` is now sticky (preserves the *original* authored class), so export always re-lays-out a `NewEntity` from its true attribute layout.
- **Same-class `PredefinedType` on a created entity** — the new-entity export path runs `retypeArgTokens` whenever a retype intent exists (not only on a class change), so `setEntityType(id, 'IfcColumn', 'PILASTER')` materializes `.PILASTER.`.
- **`getTypeMutations()`** returns a defensive copy (mirrors `getTombstones()`).
- **`lineFor` test helper** matches the STEP record terminator instead of a line boundary.

New regression tests cover each (mutations 34, export 124, all green).

**Documented as known limitations (intentionally out of scope)**
- **`deltaOnly` + existing-entity retype** — retype materializes inside the source-iteration loop, which `deltaOnly` skips. This is consistent with *all* in-place existing-entity edits (attribute/positional mutations behave identically); `deltaOnly` emits new entities + new psets, not in-place rewrites. Retyped *overlay-created* entities are still emitted under `deltaOnly`. The primary `applyMutations` path is unaffected. Noted at the call site.
- **Cross-schema up-conversion dropping `PredefinedType`** — the retype runs in the source schema before the converter; if the target class gains a `PredefinedType` slot only in the *output* schema (e.g. IFC2X3→IFC4 `IfcColumn`), an explicit override has no slot. Narrow combo (retype + conversion + override); single-schema export is unaffected. Noted in `retypeStepLine`'s doc.

louistrue:
### Added: "Reassign IFC class" control in the web viewer (`c6f95e81`)

The retype capability is now exposed end-to-end in the WASM viewer, not just the packages:

- **Edit toolbar → `Reclass`** on a selected element (gated to `IfcElement`-subtype occurrences — hidden for type entities / spaces / materials).
- **Dialog**: current → target class, one-click chips for the common building-element targets, a searchable full element list (free-text for vendor classes), and a schema-driven `PredefinedType` picker. Honest scope note: applies to the exported IFC + this panel; tree/colors/type-filters refresh on reload.
- **`setEntityType` store action** with full undo/redo (undo reverts to the previous class or drops the intent; redo re-applies).
- **Live pending badge** in the panel ("Reassigned → IfcColumn on export") reading the overlay, so the change is visible before export.

Also refactored the mutations overlay so `setEntityType` no longer mutates `NewEntity.type` in place — the `typeMutation` is the single source of truth for the effective class, which makes undo a clean revert and the exporter re-lay-out from the authored layout.

**Verified end-to-end in the WASM viewer** (Chrome DevTools, `hello-wall.ifc`): load → enter edit mode → select wall → Reclass → pick `IfcColumn` (before→after + PredefinedType appeared, Reassign enabled) → apply (dirty + "Export Changes 1" + pending badge) → undo (badge cleared, reverted). Package suites green (mutations 34, export 124); viewer `tsc --noEmit` clean.

Live-reflection scope is intentionally panel-local: the core `EntityTable.getTypeName` hot path is untouched, so the spatial tree / lens colors / type filters reflect the new class after reload — a deliberate v1 boundary to avoid rebuilding those indices on every edit.

PR #1235, opened 2026-06-20
feat(export): HBJSON (Honeybee/Ladybug) energy-model export

louistrue:
## What

Adds **HBJSON** (Honeybee / Ladybug Tools energy & daylight model) as a first-class export, alongside csv/json/ifc/glTF.

```
ifc-lite export model.ifc --format hbjson --out model.hbjson
```

HBJSON is the open, Rhino-free Ladybug Tools model format — and Honeybee has **no native IFC import**, so this makes ifc-lite a client-side IFC→Honeybee on-ramp (feeds Ladybug/Pollination for energy & daylight analysis).

## How

New pure-Rust **`ifc-lite-export`** crate is the source of truth (CLI / SDK / wasm are thin callers — per the rust-first principle). There was no Rust export crate before; this establishes the pattern (glTF can migrate into it later).

```
IFC bytes
 → ifc-lite-export (Rust)            IfcSpace → watertight prisms → serde HBJSON
 → IfcAPI.exportHbjson (wasm)        regenerated into pkg/ifc-lite.d.ts
 → bridge / GeometryProcessor.exportHbjson (TS)
 → CLI  export --format hbjson
```

Rooms are built **analytically from extruded-area profiles, not the render mesh** (the render mesh drops sliver triangles → breaks watertightness; measured 67–71% solid vs 85–100% analytic). Robustness gates, each proven necessary on real models: Y-up→Z-up; origin-rebase (national-grid f32 collapse); 2·tol Euclidean vertex merge (near-tolerance sliver walls → non-manifold); planarity + min-edge + simple-polygon (no crossing/pinch) + manifold-closure screens; by-construction face typing (Floor/RoofCeiling/Wall) with outward normals. Degenerate spaces are skipped with a surfaced count (no silent truncation).

## Verification (against honeybee-core)

- **duplex** → 19/19 rooms, `model.check_all()` **PASS**
- **rvt01** (Revit, georeferenced) → 47/47 watertight rooms
- wasm binding exercised directly on the regenerated pkg from Node → honeybee PASS
- 2 cargo tests (`ifc-lite-export`), skip-if-absent fixtures (repo convention)

## Notes

- Adds **`packages/cli/src/wasm-node-init.ts`** (`initSync` from disk): the `--target web` wasm can't `fetch(file://)` under Node, which also currently breaks `clash`/`lod` in Node. Wired into the hbjson path here; clash/lod can adopt it in a follow-up.
- Scope is **rooms (P1)**. Apertures, doors, shades (ShadeMesh) and constructions (MaterialLayerSet→OpaqueConstruction) follow. `IfcRelSpaceBoundary` is not implemented in ifc-lite → interior adjacency is left to honeybee `solve_adjacency` downstream.
- Changeset included (`@ifc-lite/{cli,geometry,wasm}` minor). The wasm `.js`/`.wasm` are gitignored/CI-built; only the regenerated `.d.ts` is committed.

## New Features
- Added HBJSON export for energy modeling from IFC (`ifc-lite export --format hbjson`), available in the Viewer export menu and via `GeometryProcessor.exportHbjson()`.
- Exports watertight room volume geometry with energy constructions from IFC material layer sets, plus door/window openings, interior wall surface adjacencies, and railing shading geometry.

## Bug Fixes
- Improved CLI support for Node-based exports by ensuring the WebAssembly engine is correctly initialized before running HBJSON export.

### DIFF
--- a/.changeset/hbjson-export.md
+++ b/.changeset/hbjson-export.md
@@ -1,6 +1,7 @@
 ---
 "@ifc-lite/cli": minor
 "@ifc-lite/geometry": minor
+"@ifc-lite/sdk": minor
 "@ifc-lite/wasm": minor
 ---
 
@@ -17,4 +18,6 @@ build-ups become Honeybee opaque constructions (real layer names + thicknesses; 
 properties defaulted by material-name keyword, since IFC rarely carries them) assigned by face
 type. Shared interior walls are paired as `Surface` adjacencies so multi-zone energy models
 don't lose heat to ambient. Backed by a new pure-Rust `ifc-lite-export` crate (source of truth
-for CLI / SDK / wasm). Available in the viewer's export menu as "Export HBJSON (Energy Model)".
+for CLI / SDK / wasm). Available in the viewer's export menu as "Export HBJSON (Energy Model)",
+on the CLI as `export --format hbjson`, and via the SDK as `bim.export.hbjson()` (delegated to a
+geometry-capable backend; the data-only SDK stays wasm-free).
--- a/.changeset/hbjson-export.md
+++ b/.changeset/hbjson-export.md
@@ -15,4 +15,6 @@ profiles (not the render mesh), so they are watertight by construction and wasm-
 `IfcRailing` occurrences are emitted as shading `ShadeMesh` geometry, and `IfcMaterialLayerSet`
 build-ups become Honeybee opaque constructions (real layer names + thicknesses; thermal
 properties defaulted by material-name keyword, since IFC rarely carries them) assigned by face
-type. Backed by a new pure-Rust `ifc-lite-export` crate (source of truth for CLI / SDK / wasm).
+type. Shared interior walls are paired as `Surface` adjacencies so multi-zone energy models
+don't lose heat to ambient. Backed by a new pure-Rust `ifc-lite-export` crate (source of truth
+for CLI / SDK / wasm). Available in the viewer's export menu as "Export HBJSON (Energy Model)".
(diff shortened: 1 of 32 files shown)


louistrue:
**Heads-up: the Node-wasm bootstrap overlaps with the sibling `feat/rust-exporters` branch**

This PR adds `packages/cli/src/wasm-node-init.ts` (`ensureWasmForNode()` → `initSync({ module: bytes })`) to work around Node's `fetch()` not loading the `file://` wasm URL, called from `clash.ts` and `headless-backend.ts`.

The sibling `feat/rust-exporters` branch (the rest of the exporters — OBJ / glTF / CSV / JSON / JSON-LD / Parquet / STEP+schema-conv+mutations / IFC5 / Merged) solves the same problem one level down — inside `IfcLiteBridge.init()` itself, Node-gated (`globalThis.process` + `@vite-ignore` dynamic imports, `init({ module_or_path: bytes })`). That makes it transparent for **every** Node consumer (CLI, MCP, tests) without each call site invoking a helper. It's verified browser-safe (the viewer production build passes) and it additionally unblocks **MCP** geometry/export in Node, which `ensureWasmForNode()` doesn't cover today (it's only wired into the CLI's two entry points).

Proposed convergence — **no change needed to this PR**:
- Merge this as-is; `wasm-node-init.ts` makes the CLI HBJSON path work in Node now.
- When `feat/rust-exporters` rebases on top, it drops `wasm-node-init.ts` + the two `ensureWasmForNode()` calls and the bridge-level path takes over for all consumers.

Flagging so we don't ship two Node-init mechanisms long-term. Happy to invert it (keep the CLI helper, drop the bridge branch) if you'd rather keep Node concerns out of `@ifc-lite/geometry`.

Issue #1004, opened 2026-06-07
Ingest-path stores use `as unknown as IfcDataStore` and skip IfcStoreBase accessors


louistrue:
Split out from #952 (now resolved/closed). The cache-reopen cast bridge was fixed in #993 (typed `hydrateCacheStore`) and the wasm-free typecheck lane landed in #1002 — but the same duck-typing pattern #952 condemned still lives in the **ingest** paths:

- `apps/viewer/src/hooks/ingest/viewerModelIngest.ts` — `createMinimalGlbDataStore()` builds a store via a whole-object `as unknown as IfcDataStore` cast (plus per-field casts).
- `apps/viewer/src/hooks/ingest/pointCloudIngest.ts` — same pattern.

These synthetic stores **lack the `IfcStoreBase` accessors** (`getEntity` / `getEntitiesByType` / `getProperties` / `getQuantities`) and never call `attachDataStoreAccessors`, yet they route into the live store via `useIfcLoader` / `ViewportContainer` → the query path (`new EntityNode(model.ifcDataStore)` → `store.getProperties()`).

**Why it matters:** same crash class as the original cache-reopen bug (#950). A future required-member addition to `IfcDataStore` stays silently green at these `as unknown as` sites and throws `TypeError: store.getProperties is not a function` at runtime on the GLB / point-cloud ingest flow. It also violates the AGENTS.md house rule (no `as any` / `as unknown as` escape hatches — fix the types or add a `.d.ts`).

**Fix:** replace the casts with a typed builder (analogous to `hydrateCacheStore` in `useIfcCache.ts`) that constructs an `IfcStoreData` and runs it through `attachDataStoreAccessors`, so the compiler enforces the full `IfcDataStore` contract for synthetic GLB/point-cloud stores too.

Context: #952, #950, #993, #1002.

PR #1244, opened 2026-06-20
fix(viewer): attach data-store accessors on IFCX import

louistrue:
## Problem

Importing an IFCX model, then selecting any entity, crashed the viewer:

```
Uncaught TypeError: this.store.getQuantities is not a function
```

Repro: export an IFC4 model to IFCX in the viewer, re-import the `.ifcx`, select an object → crash on the properties/quantities lookup.

## Root cause

`parseIfcxViewerModel` (`apps/viewer/src/hooks/ingest/viewerModelIngest.ts`) builds a **populated** `IfcDataStore` — it has the real `entities` / `properties` / `quantities` / `relationships` / `spatialHierarchy` tables from the IFCX parse — but returned it as a plain object literal cast `as unknown as IfcDataStore`. The cast supplies the *data tables* but **not the lazy accessor methods** (`getQuantities` / `getProperties` / `getEntity`) that the query + selection path calls. Selecting an entity → `store.getQuantities(...)` → `undefined` → `TypeError`.

This is the same crash class the GLB / point-cloud / synthetic paths already guard against via `attachDataStoreAccessors` (#950 / #1004); the IFCX path was simply missed.

## Fix

Route the populated IFCX store through `attachDataStoreAccessors`. Unlike the GLB path it **can't** use `createSyntheticDataStore` (that factory builds *empty* tables — it would discard the real IFCX data). The accessor's fallback is exactly right for IFCX: with no on-demand source map present, `getQuantities`/`getProperties` read straight from the populated tables.

```diff
-    dataStore: { …populated tables… } as unknown as IfcDataStore,
+    dataStore: attachDataStoreAccessors({ …populated tables… } as unknown as IfcStoreData),
```

## Verification

- Viewer `tsc --noEmit` passes for the changed file (`attachDataStoreAccessors` / `IfcStoreData` are exported from `@ifc-lite/parser`).
- Accessor fallback confirmed: `data-store-accessors.ts` returns `quantities.getForEntity(expressId)` when no on-demand map is set — the IFCX case.
- One-file, behaviour-preserving change; please QA the IFCX round-trip (export → import → select) in the preview.

## Release Notes

* **Bug Fixes**
  * Fixed a runtime crash that occurred when selecting entities in IFCX-imported models, allowing proper entity selection and querying functionality.

### DIFF
--- a/.changeset/ifcx-import-store-accessors.md
+++ b/.changeset/ifcx-import-store-accessors.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/viewer": patch
+---
+
+Fix `this.store.getQuantities is not a function` crash when selecting an
+entity in an IFCX-imported model. The IFCX ingest built a populated data
+store but never attached the lazy accessor methods
+(`getQuantities`/`getProperties`/`getEntity`) the query/selection path
+calls — it now routes the store through `attachDataStoreAccessors`.
--- a/apps/viewer/src/hooks/ingest/viewerModelIngest.test.ts
+++ b/apps/viewer/src/hooks/ingest/viewerModelIngest.test.ts
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import {
+  StringTable,
+  EntityTableBuilder,
+  PropertyTableBuilder,
+  QuantityTableBuilder,
+  RelationshipGraphBuilder,
+} from '@ifc-lite/data';
+import { buildIfcxDataStore } from './viewerModelIngest.js';
+
+/**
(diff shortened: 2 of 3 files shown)


louistrue:
Addressed in a6c2703a. `getEntity`/`getEntitiesByType` are now served from the populated IFCX `EntityTable` instead of the STEP `BufferEntitySource` (which can't read an IFCX-JSON source + empty byte index). A one-pass `knownIds` set + type→ids map back both accessors; `attributes` is empty because IFCX has no STEP attribute lists (identity rides `type`; name/GlobalId come from the entity table via the store's other accessors). `getQuantities`/`getProperties` keep reading the populated quantity/property tables.

louistrue:
Added in 9407f0e9: extracted `buildIfcxDataStore` and added `viewerModelIngest.test.ts` — a regression test for the selection-time accessor path. It builds a populated entity table + empty-but-valid property/quantity tables and asserts getQuantities/getProperties exist and return arrays (the original crash), getEntity resolves known ids (type set, attributes empty) and returns null for unknown ids, and getEntitiesByType matches case-insensitively. 3/3 pass.

PR #1260, opened 2026-06-21
perf(geometry): single-controller threaded-WASM foundation (bundle + loader selection)

louistrue:
## What

Phase 0+1 of the single-controller threaded-WASM plan: **ship the threaded bundle by default and wire runtime selection in the geometry loader.** Opt-in (`?geomThreads=1`) and off by default while validated.

Evidence (from #1255's bench + design doc): threaded-WASM CSG corpus replay hits **~4.19× at 8 threads**, naive whole-pipeline 1.6–1.9×, projected ~2.5–3× cold-load. This PR lays the foundation to get there.

## Changes

- **`build-wasm.sh` emits `pkg-threaded` by default** (opt out with `BUILD_THREADED=0`). Reuses the same `nightly-2025-11-15` + `build-std` the single-thread bundle already requires — **no new CI toolchain** (the threaded build doubles the "Build WASM" step; can be optimized later if CI time matters).
- **`@ifc-lite/wasm` `./threaded` subpath** with a committed `threaded.d.ts` type shim, so `import('@ifc-lite/wasm/threaded')` type-resolves even on hosts that haven't built `pkg-threaded` (typecheck-only lanes); the `import`/`default` conditions point at the built runtime.
- **Runtime selection** (`packages/geometry/src/wasm-threads.ts` + loader): on `?geomThreads=1` AND a cross-origin-isolated, threads-capable host, the orchestrator spawns **ONE** geometry worker that loads the threaded bundle and calls `initThreadPool` — the single-controller topology — instead of N single-thread workers (avoids N-pools-×-cores oversubscription). Safari / non-isolated / any threaded-init failure → single-thread fallback.

## Scope / safety

- This does **not** flip the per-element loop to `par_iter`. The viewer's `processGeometryBatch` element loop stays serial; only the inner FacetedBrep mesher parallelizes (owned per-element data, single level — no shared-decoder hazard). The shared-`&mut EntityDecoder` element-loop `par_iter` is the next, separately-gated step (the design critique flagged it as the gating decoder restructure; the surgical CSG-only split is the recommended end state).
- Off by default (`?geomThreads=1` to test on the preview). Production topology unchanged until validated.

## Verify

- `pnpm typecheck`: 32/32 green.
- `build-wasm.sh` (default-on) builds both bundles locally; the threaded bundle imports shared memory (litmus passes).
- To test: open the preview with `?geomThreads=1` on a Chromium browser (cross-origin isolated) and confirm the console logs `single-controller threaded` + `threaded WASM active (N threads)`, geometry loads correctly, and Safari (or no flag) still loads the single-thread bundle.


louistrue:
Closing: same dead end as #1429. In-instance / single-controller WASM threading is a proven net regression on most benchmarked models (per-worker file re-decode + dlmalloc global-lock contention; heavy-tail elements can't be split across threads), so this foundation has no path to a real win. The verdict is recorded in agent memory so we don't re-investigate. The real CSG levers are cutter-chunking (#1432), worker-pool right-sizing (#1431), and wide-arithmetic (#1426). Restorable from this closed PR if a new mechanism ever changes the calculus.

PR #1262, opened 2026-06-21
fix(viewer): self-heal blank page when a deploy-window asset 404s

louistrue:
## Problem

Loading `ifclite.com` while a Vercel deploy is mid-rollout (or when the `__vdpl` skew-protection cookie pins the session to a now-deleted deployment) makes the content-hashed **entry module** 404. Vercel serves 404s as `content-type: text/plain`, which the browser **refuses to execute as a module** ("blocked because of a disallowed MIME type"). The entry never runs → React never mounts → the page shows only the themed **background with no content**, recoverable only by a manual reload.

Confirmed against the live site: a missing `/assets/*.js` returns `404 text/plain` (`x-vercel-error: NOT_FOUND`) — exactly the reported MIME error. Skew Protection is already active (`dpl_…`) but has a cold-start/cross-context gap and no client-side safety net under it.

## Fix

- **`apps/viewer/index.html`** — inline **boot watchdog**. Must be inline: a listener inside the bundle can't fire if the bundle itself failed to load. Reloads **once** on an entry/preload module load-error or if `#root` never gains children. `sessionStorage`-bounded (max 2 attempts, drops a stale `__vdpl` pin on retry, resets on successful mount) so it can't loop.
- **`apps/viewer/src/main.tsx`** — `vite:preloadError` handler for **lazy** chunks (`exporters`/`ids`/`bcf`/`sandbox`) that 404 after a mid-session deploy. Complements the boot watchdog.
- **`vercel.json`** — `immutable` caching for `/assets/*` (content-hashed → byte-stable across deploys; was needlessly revalidating), plus explicit always-revalidate on the HTML entry so clients never run stale markup.

## Verification

Ran the inline watchdog under a mocked DOM — all scenarios green:

- **A** entry fails, `#root` empty → reloads once ✅ (the reported bug)
- **B** app mounted → no reload, guard cleared ✅
- **C** already retried max → gives up, no loop ✅
- **D** retry → drops stale `__vdpl` pin ✅

`vercel.json` valid, HTML tags balanced, inline JS syntax-checked. The failure path is hard to stage on a live preview (needs a mid-deploy race), hence the unit-level proof; the preview deploy confirms the watchdog ships in the built HTML and the app still boots normally.

### DIFF
--- a/apps/viewer/index.html
+++ b/apps/viewer/index.html
@@ -70,23 +70,44 @@
       var ERROR_DEBOUNCE_MS = 800;
       var healed = false;
 
-      function attempts() { try { return Number(sessionStorage.getItem(KEY)) || 0; } catch (e) { return 0; } }
-      function setAttempts(n) { try { sessionStorage.setItem(KEY, String(n)); } catch (e) {} }
+      function warn(msg) { if (typeof console !== 'undefined') console.warn('[boot-self-heal] ' + msg); }
+      // Current attempt count, or -1 when sessionStorage is unreadable (sandboxed
+      // frame / storage disabled) — the signal that we cannot bound retries.
+      function attempts() {
+        try { return Number(sessionStorage.getItem(KEY)) || 0; }
+        catch (e) { warn('sessionStorage unreadable: ' + e); return -1; }
+      }
+      // Persist n and confirm it survives a read-back; returns false when storage
+      // is unavailable, so callers can refuse to reload rather than loop forever.
+      function recordAttempts(n) {
+        try {
+          sessionStorage.setItem(KEY, String(n));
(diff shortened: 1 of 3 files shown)


louistrue:
Addressed both P2 findings in `08a9f65a`: the boot watchdog (`index.html`) and the `vite:preloadError` handler (`main.tsx`) now **confirm the retry count actually persisted (read-back) before reloading**. When `sessionStorage` is unavailable (sandboxed frame / storage disabled) they bail and let the broken page / preload error surface instead of looping — so the retry cap holds even when storage throws. All `catch` blocks now log (no silent catch). Added mocked-DOM tests for the storage-read-blocked and storage-write-blocked paths (both assert: no reload).

PR #1259, opened 2026-06-21
fix(geometry): cut plan-rotated walls in their own axis-aligned frame (#1167)

louistrue:
Fixes #1167. Rebased + compile-fixed version of @Sonderwoods's #1223 so it merges onto current `main` — **full credit to @Sonderwoods**, who diagnosed the root cause and wrote the fix and tests; the commit is authored to him. Supersedes #1223 and #1214 (both will be closed and linked here).

## Why a new PR
#1223 was branched before #1209 added the `param: Option<ParamRectCut>` field to `VoidContext`, so it no longer compiles against `main` (`E0063: missing field param`) — that's why its CI is red. The only substantive change here over #1223 is `param: None` in the local-frame `VoidContext`. The branch histories are unrelated (main was rewritten), so a rebase/merge couldn't share a base; this re-applies the exact net change as one clean commit.

## Root cause (verified)
Plan-rotated façade walls (this building sits ~13° off the world grid) had window/door voids cut **far oversized and skewed to the world axes**. Two coupled causes:
1. `is_axis_aligned_direction`'s tolerance was `0.95` (~18°), so a 13°-rotated opening passed as "axis-aligned" → routed onto the **world-AABB** cut path. The AABB of a rotated box is strictly larger than the box → oversized, grid-aligned hole. (Runtime log confirmed: classifier reported `rect=68 diag=0` — every opening on the AABB path.)
2. Even the **exact** subtract over-cuts/fragments at ~150 m world coords (f32 ≈ 15 µm).

## Fix
- Tighten the tolerance to `cos(1°)` → rotated openings reclassify `Rectangular → DiagonalRectangular`.
- Cut plan-rotated walls in their **own origin-centred, axis-aligned frame**: host + openings rotated in (where the subtract is clean and f32-precise), cut, rotated back. Scoped to plan-rotated walls; axis-aligned walls and roof/floor/sloped openings keep their existing path.

The two parts are **coupled and both required** — I verified threshold-only still over-cuts 22.5 m³ (target ~13), and local-frame with the old 0.95 threshold also fails.

## Verification
- **`ifc-lite-geometry` suite: 507 passed / 0 failed / 14 ignored.**
- New regression tests pass: real reporter wall (`231.01`, 5 openings, full placement chain — 22.5 m³ over-cut + ~236 unpaired edges → ~13 m³ watertight) + synthetic multi-angle guards (3–45°).
[...]

PR #1270, opened 2026-06-21
fix(geometry): harden rotated-wall local-frame cut + correct release target (#1167 follow-up)

louistrue:
Follow-up to #1259 addressing the post-merge review findings from **CodeRabbit** and **Codex (ChatGPT)**. Each was verified against the merged code before fixing.

| # | Finding | Source | Fix |
|---|---|---|---|
| **A** | Changeset bumped `@ifc-lite/geometry`, but the compiled Rust fix ships in **`@ifc-lite/wasm`** (geometry *depends on* wasm, not vice-versa), so consumers pinning `@ifc-lite/wasm` would never get the fix. | Codex P1 | Changeset now targets `@ifc-lite/wasm`; `@ifc-lite/geometry` cascade-patches via its workspace dep. Caught before release (changeset was still pending). |
| **B** | Local-frame cut seeds the frame from the *first* rotated opening, then treated **every** `DiagonalRectangular` as axis-aligned in that frame — a differently in-plane-oriented opening would be over-cut. | CodeRabbit (Major) + Codex P2 | Gate per-opening: only a box whose **own** frame aligns with the wall frame takes the fast rectangular cut; anything else keeps its exact (un-rotated, centred) mesh for the subtract. |
| **C** | `project_aabb_in_frame` validated only `lo` finiteness — a `+inf` projection lands in `hi` and could slip through. | CodeRabbit (Minor) | Validate **both** bounds. |
| **D** | Regression test's `area > 1e-9` guard skipped exactly-collapsed triangles, which can self-cancel in the edge map and let `defects()` report `(0,0)` for a fragmented cut. | CodeRabbit (Minor) | Count collapsed zero-area triangles (with real extent) as defects. |

## Verification
- **`ifc-lite-geometry` suite: 507 / 0 / 14.**
- #1167 real-wall regression + multi-angle guards still green — the per-opening gating (B) does not change the common case (all wall openings share the wall frame), and the strengthened defect test (D) still passes (the cuts are genuinely clean).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation checks for rotated wall openings to prevent invalid geometry calculations.
  * Improved axis-aligned boundary validation for enhanced accuracy in geometric computations.
  * Enhanced handling of complex wall openings that don't align with standard reference frames.
  * Refined degenerate geometry and triangle detection logic to improve overall mesh quality.

* **Chores**
  * Released patch for WebAssembly module with geometry improvements.

### DIFF
--- a/.changeset/wasm-rotated-wall-void-fix.md
+++ b/.changeset/wasm-rotated-wall-void-fix.md
@@ -1,0 +1,12 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Bump the wasm package for the rotated-wall void-overcut fix (#1167). The fix is
+Rust geometry code compiled into `@ifc-lite/wasm`, but its original changeset
+bumped only `@ifc-lite/geometry` — so the package that actually carries the
+compiled fix never got a release. This patch bumps `@ifc-lite/wasm` (and
+cascade-patches `@ifc-lite/geometry`, which depends on it) so consumers pinning
+the wasm package receive the corrected geometry. Also includes the #1259/#1270
+review follow-ups: per-opening frame gating in the local-frame cut, faithful
+per-cutter depth direction, and a both-bounds finiteness check.
--- a/rust/geometry/src/router/voids.rs
+++ b/rust/geometry/src/router/voids.rs
@@ -755,8 +755,9 @@ fn project_aabb_in_frame(
             hi[k] = hi[k].max(v);
         }
     }
-    lo.iter()
-        .all(|v| v.is_finite())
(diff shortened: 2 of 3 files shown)


louistrue:
Addressed the Codex P2 (preserve each cutter's local depth) in `3547aadf`: frame-misaligned cutters now carry their own `f.depth` transformed into the wall frame as the `extrusion_dir`, so the exact fallback's cap-extension stays faithful to that opening instead of the wall normal. Geometry suite still 507/0/14.

louistrue:
@coderabbitai review

PR #1282, opened 2026-06-21
debug: trace 170_KM instanced-render regression [INST-DBG, do not merge]

louistrue:
Temporary instrumentation to isolate why GPU-instanced occurrences (#1238) vanish on the 170_KM national-grid model. Adds [INST-DBG] console logs at the shard decode/upload chokepoint. Draft — for a preview build only; will be reverted once the root cause is found.


louistrue:
Superseded by #1292 (the real fix). This was debug-only instrumentation; the root cause was apply_submesh_placement not recording instance_meta.transform.

PR #1291, opened 2026-06-21
fix(viewer): exploded mode leaves instanced + null-elevation storeys behind (#1289)

### LINKED ISSUE #1289: Exploded mode does not work properly
Some objects are left behind on the Hauz model, and on the other one showing only 1 floor.

See attached models.

[ifc-georeferencer-AC20-FZK-Haus.ifc.zip](https://github.com/user-attachments/files/29174358/ifc-georeferencer-AC20-FZK-Haus.ifc.zip)

[Project01_PC01_edit_v2_2026-06-19.ifc.zip](https://github.com/user-attachments/files/29174360/Project01_PC01_edit_v2_2026-06-19.ifc.zip)

louistrue:
Fixes #1289.

Both symptoms in the issue trace to two independent defects in Exploded level-display mode. Verified against the two attached models (AC20-FZK-Haus, Project01_PC01).

## 1. "Objects left behind" (AC20-FZK-Haus) — render path
Exploded mode lifts a storey by pushing per-entity translations into `Scene.translateMeshesForEntities`, which only touched the flat `meshDataMap`. GPU-instanced occurrences (repeated geometry emitted via `IfcMappedItem` — FZK-Haus has 56 mapped items, and instancing fires at `min_group = 2`) live in separate per-template instance buffers, so they never moved and stayed at the original elevation while the rest of the storey rose.

`Scene.translateInstancedEntity` now shifts each occurrence's mat4 translation column in **both** the CPU instance record and the GPU instance buffer, plus the cached world AABB (so pick / measure / section / export stay consistent). It is wired into `translateMeshesForEntity`, so it also fixes gizmo moves of instanced elements.

## 2. "Only one floor" (Project01_PC01) — parser path
That model's `FirstFloor` storey has a null `Elevation` attribute but is placed at Z = 3.5 via its `ObjectPlacement`. The spatial-hierarchy builder dropped any storey without an explicit elevation from `storeyElevations`, leaving Exploded mode a single floor to order.

The builder now falls back to the storey's `ObjectPlacement` → `RelativePlacement` → `Location` Z when the `Elevation` attribute is missing. This matches the attribute's semantics (Z relative to the parent container), so it stays consistent with sibling storeys that do carry the attribute and avoids folding in site-level georeferencing Z.

## Tests
- `Scene.translateInstancedEntity`: lift, instanced-only entity via the public API, reversibility (Exploded → Stacked), and no-op cases. Full scene suite 29/29.
- `storey-elevation-fallback.test.ts`: synthetic IFC4 with one attribute-elevation storey and one null-elevation/placement-Z storey; asserts both land in `storeyElevations`.
- Changeset added (`@ifc-lite/renderer` + `@ifc-lite/parser`, patch).

### DIFF
--- a/.changeset/exploded-mode-instanced-and-elevation.md
+++ b/.changeset/exploded-mode-instanced-and-elevation.md
@@ -1,0 +1,22 @@
+---
+"@ifc-lite/renderer": patch
+"@ifc-lite/parser": patch
+---
+
+Fix Exploded level-display mode leaving geometry behind (#1289).
+
+Two independent defects made Exploded mode look broken:
+
+- GPU-instanced occurrences (repeated geometry emitted via `IfcMappedItem`, e.g.
+  windows / mullions) were never lifted with their storey, because the per-entity
+  translate only touched the flat `meshDataMap` and not the instanced shard. They
+  stayed at their native elevation while the rest of the storey rose ("objects
+  left behind"). `Scene.translateInstancedEntity` now shifts each occurrence's
+  transform in both the CPU instance record and the GPU buffer, plus its cached
+  world AABB, so pick / measure / section / export stay correct. This also fixes
+  moving an instanced element with the gizmo.
+
+- A storey whose `Elevation` attribute is null (common in Revit / ArchiCAD
+  exports) was dropped from the elevation map, so Exploded mode had a single
+  floor to order ("only one floor"). The spatial-hierarchy builder now falls back
+  to the storey's `ObjectPlacement` Z when the attribute is missing.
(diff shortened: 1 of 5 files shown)


louistrue:
Both Codex findings fixed in b73376c6:

1. **Elevation gate** — `extractElevation` now reads only the canonical Elevation slot (index 9). The previous broad scan treated reference attributes (parsed as bare express-id numbers, e.g. OwnerHistory) as elevations, so the null-Elevation branch never ran. Verified end-to-end in a real-browser run: ground floor = 0 (attr), first floor = 3.5 (placement fallback) on the Project01 model; FZK-Haus = [0, 2.7].

2. **Instanced bounds** — `translateMeshesForEntity` now runs flat before instanced and rebuilds the instanced world AABB from the moved occurrence matrices (`recomputeInstancedBounds`) instead of shifting it in place, so a mixed flat+instanced entity can't end up with stranded (null) bounds. Added a regression test.

Validated in a real-GPU browser (Playwright) against both attached models: storeys order correctly and geometry lifts cleanly in Exploded mode.

Issue #1286, opened 2026-06-21
get wasm fast again


louistrue:
> What I mean is, it's not about the total opening time itself, but the difference in opening time. For instance, even a lightweight file might take about 20 seconds to open, but while it took 20 seconds in the old version, it takes 25 seconds in the new one. The same goes for a very large model: it opens in about 1 minute 20 seconds in the old version, but takes around 2 minutes in the new one. So we are looking at this kind of gap. It has become about 40 to 50 percent slower. 

 _Originally posted by @Blogbotana in [#1109](https://github.com/LTplus-AG/ifc-lite/issues/1109#issuecomment-4759504055)_

Blogbotana:
I'll try to solve this issue too 👍

louistrue:
cool, go for it @Blogbotana 👍 I'm currently looking into plane clipping (for the Tekla steel-heavy stuff) and a fast rectangular-openings path (for architectural models)...

Blogbotana:
**Profiled the regression natively — the dominant cost is mesh-count fragmentation, not the CSG kernel.**

Method: ran `process_geometry` natively (no WASM/JS noise, same machine) on three boolean-heavy Tekla steel models, comparing the last Manifold-default commit `7c0f0d82` (parent of #1024) against current `main`. Per-phase numbers come from the built-in `ProcessingStats`.

| File | metric | Manifold `7c0f0d82` | main | ratio |
|---|---|---|---|---|
| 130_KM | geometry_ms | 621 | 2415 | **3.9×** |
| | total_ms | 702 | 2509 | 3.6× |
| | **meshes** | 9 031 | 26 739 | **3.0×** |
| | triangles | 2.69M | 2.93M | 1.09× |
| 140_KM | geometry_ms | 453 | 1978 | **4.4×** |
| | meshes | 5 180 | 18 876 | 3.6× |
| | triangles | 1.91M | 2.07M | 1.08× |
| 170_KM | geometry_ms | 1738 | 5934 | **3.4×** |
| | meshes | 16 622 | 64 314 | 3.9× |
| | triangles | 7.14M | 7.92M | 1.11× |

Parse/scan are <5% of total on every file — the cost is entirely in the geometry phase.

**Findings**

- Triangle counts are within ~10% and `total_csg_failures == 0` on all three — the exact kernel completes every cut and never falls back to the AABB box. So this is **not** the booleans (and not the #1109 escalation path).
[...]

PR #1294, opened 2026-06-21
feat(geometry): element-level void-cut dedup, flag-gated OFF (#1286 Phase 4-5)

louistrue:
## What

#1286 Phase 4-5: the actual CSG win — dedup the **void cut** so identical voided elements are exact-kernel-cut once instead of once per occurrence. **Flag-gated OFF (`IFC_LITE_DEF_DEDUP=1`)**, so this is a no-op for production until corpus-validated.

Stacked on #1293 (the foundation + parity gate).

### Phase 4 — definition-frame void cut (the enabler)
`process_element_with_submeshes_and_voids_unplaced` cuts the UNPLACED base with cutters relativized into the host's **definition frame** (reusing the existing `VoidContext::relativized_by` + `apply_void_context_inner`), producing a placement-invariant, reusable result. Eligibility gated to **pure-translation, non-layered** hosts (`host_translation_only`); rotated/layered hosts return `None` and fall back to the per-occurrence path. `instance_meta` cleared (voided geometry isn't GPU-instanced, matching the placed path).

Proven geometrically identical to the placed path (sorted-coord-set) — only vertex order differs, since the placed path cuts independently.

### Phase 5 — the cache
`definition_signature` keys identical voided elements: host rep signature + an **order-invariant** fold over each opening's `(rep signature, host-relative placement)`. `process_element_with_submeshes_and_voids_deduped` cuts once on a miss, caches the definition-frame template, and on a hit clones it + applies this occurrence's placement. Wired into `produce_inner` (tried before the per-occurrence path) and armed in both batch paths (native rayon pool + wasm) alongside the item cache.

## Why it's safe to land off

- Default OFF => native==wasm byte-identical, zero production change.
- Never wrong, only deferred: any ineligible/unresolved case returns `None` -> existing per-occurrence path.
- Voided geometry isn't instanced either way (no collation interaction).

## Tests
`cargo test -p ifc-lite-geometry --test dedup_validate` (6 pass):
- `void_unplaced_cut_matches_placed_geometry` — def-frame cut + re-place == placed (sorted-coord-set).
- `definition_cache_dedups_and_matches_placed_geometry` — deduped == placed geometry, identical walls share one template (2 cache entries), distinct void stays distinct.

Full `wall_opening_cut_regression` (7) green with the fixture staged.

## Before flipping the flag
Live/corpus validation on 170_KM + a Revit arch + a georeferenced model (the def-frame cut is geometrically invariant near origin; confirm at national-grid scale + visually). Rotated-host coverage is a follow-up (v1 is translation-only). Refs #1286.

### DIFF
--- a/.changeset/void-cut-dedup.md
+++ b/.changeset/void-cut-dedup.md
@@ -1,0 +1,17 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Element-level void-cut dedup (#1286 Phase 5), flag-gated OFF
+(`IFC_LITE_DEF_DEDUP=1`).
+
+Identical voided elements (same host geometry, opening geometries, and opening
+placements relative to the host) currently re-run the exact-kernel CSG cut once
+per occurrence. With the flag on, the cut runs ONCE in the host definition frame
+(`process_element_with_submeshes_and_voids_unplaced`, reusing the existing
+`relativized_by` + `apply_void_context_inner`) and is cached by a void-inclusive
+`definition_signature`; each occurrence reuses the template with its own
+placement. Eligibility is gated to pure-translation, non-layered hosts (rotated /
+layered / unresolved-opening hosts fall back to the per-occurrence path), so it
+is never wrong, only sometimes deferred. Default OFF keeps native==wasm
+byte-identical until corpus-validated.
--- a/rust/geometry/src/lib.rs
+++ b/rust/geometry/src/lib.rs
@@ -138,8 +138,8 @@ pub use profile::{Profile2D, Profile2DWithVoids, ProfileType, VoidInfo};
 pub use profile_extractor::{extract_profiles, ExtractedProfile};
 pub use profiles::ProfileProcessor;
 pub use router::{
-    ClassificationStats, GeometryProcessor, GeometryRouter, HostOpeningDiagnostic, ItemDedupCache,
-    OpeningDiagnostic, OpeningKindDiag, RectParam,
+    ClassificationStats, DefinitionCache, GeometryProcessor, GeometryRouter, HostOpeningDiagnostic,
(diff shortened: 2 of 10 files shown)


louistrue:
Brought this up to current main and corpus-validated it in #1507 (on fresh branches; this draft + #1293 left untouched). Summary: the 3 merge conflicts were all god-file-split relocations (voids.rs -> voids/, gpu_meshes.rs -> gpu_meshes/), ported faithfully. But on current main the def-frame-cut-then-translate path is **no longer byte-identical** to the per-occurrence cut and shows genuine geometry divergence (order-invariant coord-set + triangle-count deltas: -137 dental, +1876 advanced) plus marginal/negative perf (1.00-1.06x where it fires; 0.97x net on ISSUE_068 where it fires on 0 of 363 hosts). Root cause: the exact kernel's snap-grid quantization differs between the small definition frame and world magnitude, so the flag correctly **stays OFF**. Full table + reusable A/B harness (`def_dedup_void_ab`) in #1507.

louistrue:
Closing per the 2026-07-10 draft-PR audit. This def-frame void-cut approach was carried to current main and corpus-validated in #1507: the cut is no longer byte-identical there (snap-grid quantization differs between the definition frame and world magnitude, producing order-invariant coord-set and triangle-count deltas), and perf is marginal to negative (1.00-1.06x where it fires, 0.97x net on ISSUE_068, firing on 0 of 363 hosts). That matches the #1571 / #1286 Phase 5 NO-GO. Salvageable pieces (the A/B harness, the definition_signature order-invariant opening fold, host_translation_only gating) remain on the branch and in #1507 if this is revived after solving kernel quantization and rotated-host coverage. #1293 (the foundation PR) stays open and is being retargeted to main.

PR #1295, opened 2026-06-21
perf(geometry): skip tiny detail cuts in preview tessellation tiers (#1286)

Blogbotana:
## What

In the preview tessellation tiers (`Lowest`/`Low`), skip an `IfcBooleanResult`
DIFFERENCE whose cutter max-dimension is below **10%** of the host's (a small
steel cope/notch, a minor detail recess) and render the host un-cut. `Medium`
(the default) and finer keep **every** cut — byte-identical to before.

## Why (#1286)

Profiling the "wasm slow" regression on the Tekla KM steel corpus (native
`process_geometry`, per-phase `ProcessingStats`):

- The exact per-cut `subtract` is **~93%** of the geometry phase (skipping the
  cuts drops 170_KM geometry 7.6 s → 0.5 s). It is **not** mesh count / dedup:
  content-dedup (#1177) is already on (170_KM 35 s → 6 s without/with), and the
  combined-mesh path at Manifold-era mesh counts is still ~4× slower.
- Almost every cut is a tiny local cope/notch: the cutter's max dimension is
  **<10% of the host's on 98%** of cuts in 130_KM and **67%** in 170_KM, yet each
  still pays a full exact boolean.

So dropping the sub-threshold cuts in the preview tiers recovers Manifold-class
load times:

| File | geometry (all cuts) | geometry (skip <10%, Low) | ratio |
|---|---|---|---|
| 170_KM | 7.6 s | **1.1 s** | ~6.9× |

No parallelism, full determinism (deterministic size test). Verified visually in
`apps/viewer` at Low — members render correctly, small notches drop.

## Notes

- The change is Rust geometry compiled into `@ifc-lite/wasm` → changeset bumps
  that package (`minor`).
- Threshold is tunable natively via `IFC_LITE_FAST_CUT_RATIO`; the 10% default
  applies in wasm (no env in the browser).
- Gated to `Lowest`/`Low` only, so production (Medium+) geometry is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Optimized preview rendering in Lowest and Low quality tiers: skips very small cuts (below 10% of host size) for faster display.
  * Cut threshold configurable via `IFC_LITE_FAST_CUT_RATIO` environment variable (default: 10%).
  * Medium and higher quality tiers preserve all detail as before.

### DIFF
--- a/.changeset/skip-small-cuts-preview.md
+++ b/.changeset/skip-small-cuts-preview.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/wasm": minor
+---
+
+Skip tiny detail cuts in the preview tessellation tiers (#1286).
+
+In the `Lowest`/`Low` tessellation tiers, an `IfcBooleanResult` DIFFERENCE whose
+cutter's max dimension is below 10% of the host's (a small steel cope/notch, a
+minor detail recess) is now skipped and the host renders un-cut. On
+boolean-heavy Tekla steel models the exact `subtract` per cut dominates load
+time and almost every cut is such a small local notch, so dropping them in the
+preview tiers recovers Manifold-class load times (170_KM geometry ~7.6 s →
+~1.1 s, ~6.9×) with no parallelism and full determinism. `Medium` (the default)
+and finer keep every cut — byte-identical to before. The threshold is tunable
+natively via `IFC_LITE_FAST_CUT_RATIO`.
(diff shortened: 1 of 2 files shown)


louistrue:
nice! thanks, thinking of switching wasm and let user reload exact for clash detection etc... i'll branch off this and probably also add some tests. much appreciated! 🙌

PR #1308, opened 2026-06-22
fix(viewer): preserve case + dots in export filenames, and unify all exports (#1299)

### LINKED ISSUE #1299: CAPS vs lowercase
Even though I use UPPERCASE to create the filename, uses lowercase when it outputs the pdf.  I haven't checked yet if the excel or CSV exports work the same way.  

I also just notice the document classification code that uses a 'dot' separation (000.000) is being converted into a hyphen.
 
Image

Image

Image

louistrue:
Closes #1299

## The bug (#1299)
Saving a list export lowercased the filename (`DRAWINGS` → `drawings`) and turned dotted classification codes into hyphens (`000.000` → `000-000`). Root cause: a `slug()` helper that lowercased and replaced every non-alphanumeric run.

## Why it kept happening
Filename handling was copy-pasted all over the viewer with no single owner:
- **5 different sanitisers** with conflicting rules — list export and compare report lowercased + hyphenated dots; search export and list-definition export preserved case but turned dots into `_`; the MCP playground preserved both.
- **~30 hand-rolled** `<a download>` + `URL.createObjectURL` blocks, including **4 near-identical local download helpers** (useClash, CommandPalette, export-adapter, plus the lists one).

So the same class of bug lived in several places, and there was no obvious "right way" for the next export.

## The fix — one canonical module
New `apps/viewer/src/lib/export/download.ts`:
- `sanitizeFilename(name, { fallback, maxLength })` — preserves case and dots (any script via `\p{L}`), replaces only genuinely OS-unsafe characters with `-`, collapses whitespace, trims separators, caps length, falls back when empty. It is **not** a slug.
- `downloadBlob` / `downloadFile` / `downloadDataUrl` — a single anchor-click download flow. `downloadFile` copies the wasm `Uint8Array<ArrayBufferLike>` into a `BlobPart`, so callers stop reinventing that TS 5.7 coercion.

Everything now routes through it:
- **Sanitisers consolidated:** list export (#1299), compare report (same lowercase+dot bug), search result export + list-definition export (dot→`_`), MCP `coerceFilename` (keeps its path-strip + extension-forcing).
- **Dynamic names sanitised** at GLB / HBJSON / IFC / IFCX / changes-JSON / BCF / drawing-SVG / export-changes.
- **Duplicate helpers + inline blocks removed** across useClash, CommandPalette, export-adapter (SDK keeps its deliberate throw-outside-browser contract), and the fixed-name sites (lenses, audit/action logs, IDS BCF/JSON/HTML, clash presets, KMZ, flavor, screenshots, toolbar CSV/JSON).

CesiumOverlay's `createObjectURL` is a model *load*, not a download — intentionally untouched.

## Verification
- New `download.test.ts` (9 cases: case/dot preservation, reserved-char stripping, whitespace, trim, fallback, length, non-ASCII). Search export test updated. Both green via `node:test`.
- `tsc --noEmit` across the whole viewer: **zero new errors** vs the pre-change baseline, after each of the three migration batches.
- Net **-116 lines**. `AGENTS.md` now documents the one way so this doesn't drift back.
- No changeset: `@ifc-lite/viewer` is the deployed app, not an npm-published package.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,5 +49,9 @@ Project-specific gotchas and guardrails — the things that bite you *here* and 
 ## CLI
 - Discover the full SDK API with `ifc-lite schema` (JSON). `eval` runs SDK expressions (`ifc-lite eval model.ifc "bim.query().byType('IfcWall').count()"`); always pass `--json` for machine output. `HeadlessBackend` (`packages/cli/src/headless-backend.ts`) runs query/export/create/IDS/BCF without a renderer.
 
+## Browser exports (viewer)
+- One way to save a file: `apps/viewer/src/lib/export/download.ts`. Use `downloadBlob` / `downloadFile` / `downloadDataUrl` — never hand-roll an `<a download>` + `URL.createObjectURL` dance, and never write another filename regex. `downloadFile` already copies the wasm `Uint8Array<ArrayBufferLike>` into a `BlobPart`.
+- Run every user/model-derived filename through `sanitizeFilename` (preserves case + dots, strips only OS-unsafe chars — see #1299). It is *not* a slug; don't lowercase or hyphenate names for filenames. Slugs (extension IDs) are a separate concern.
+
 ## New source files
 - MPL-2.0 header on every new file — see [`./LICENSE_HEADER.md`](./LICENSE_HEADER.md).
--- a/apps/viewer/src/components/extensions/AuditLogPanel.tsx
+++ b/apps/viewer/src/components/extensions/AuditLogPanel.tsx
@@ -26,6 +26,7 @@ import type { AuditEvent, AuditEventKind } from '@ifc-lite/extensions';
 import { Button } from '@/components/ui/button';
 import { ScrollArea } from '@/components/ui/scroll-area';
 import { useExtensionHost } from '@/sdk/ExtensionHostProvider';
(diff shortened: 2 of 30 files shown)


louistrue:
@coderabbitai full review

Heads up: the previous walkthrough only covers the first commit. Since then this PR was broadened into a full export-path consolidation (new `apps/viewer/src/lib/export/download.ts` with `sanitizeFilename` + `downloadBlob`/`downloadFile`/`downloadDataUrl`, and ~20 call sites migrated). The old `lists/export/index.test.ts` was moved to `lib/export/download.test.ts`. Please review the latest commit too.

louistrue:
Thanks for the thorough review. Addressed on-branch in f80a26e4 rather than deferring to an issue:

- **toBlobPart leftover (ExportDialog / ExportChangesButton):** removed both local helpers; all five sites now call `downloadFile(content, name, mime)`, so no export path hand-rolls the blob dance anymore.
- **Underscore + fallback behaviour:** documented in `sanitizeFilename`'s JSDoc and added the `'_my_list_' -> '_my_list_'` test you suggested.
- **Compare filename length (≤92) and drawing space-before-separator:** intentional and within OS limits / consistent with the 'preserve what the user typed' philosophy, leaving as-is.

No follow-up issue needed.

PR #1310, opened 2026-06-22
fix(geometry): cut high-vertex IfcOpeningElement voids under the local frame (#1297)

### LINKED ISSUE #1297: IfcOpeningElement voids not cut in WASM — plate/member holes render solid (native cuts them)
## Summary

`IfcOpeningElement` voids (bolt-hole / detail openings) are **cut correctly in the native pipeline but NOT in the WASM pipeline** — so in the browser viewer the holes silently disappear and plates/members render solid. The geometry is computed (the browser pre-pass classifies the openings) but the cut is not reflected in the emitted mesh.

This is a correctness bug, independent of the #1286 performance work.

## Repro

- File: a Tekla steel model with `IfcOpeningElement` voids (701 openings; mostly non-rectangular). 
- Default tessellation (Medium), no opening filter.

## Evidence

Same element, same source, two pipelines:

| Pipeline | meshes | triangles | holes |
|---|---|---|---|
| Native (`process_geometry`) | 1 | **1217** | ✅ present, `total_csg_failures = 0` |
| WASM (`GeometryProcessor.process` → `buildPrePassOnce` + `processGeometryBatch`) | 1 | **28** | ❌ absent (≈ solid plate) |

The WASM pre-pass **does** classify the openings (so `void_keys`/`void_counts`/`void_values` are produced):

```
[IFC-LITE] Opening classifier: rect=0 diag=135 non_rect=566 ... (total=701)
[IFC-LITE] Opening pipeline: 0 CSG failures. 469 hosts with openings.
  IfcPlate: hosts=394 openings=416 (rect=0 diag=0 non_rect=416)
  IfcBeam:  hosts=71  openings=281 (rect=0 diag=132 non_rect=149)
```

Yet the emitted plate mesh has no holes. A beam without openings matches across both (92 = 92 tris), so it is not a blanket WASM breakage — it is specific to host elements that carry these (non-rectangular) `IfcOpeningElement` voids.

## Suspected cause

A native-vs-browser **void-orchestration divergence** (the kind `docs/architecture/geometry-pipeline.md` warns about — the scan / `void_index` build is per-orchestration). The native scan associates and cuts these `non_rect` openings; the browser `buildPrePassOnce` → `processGeometryBatch` path classifies them but the cut never reaches the output. Candidates to check: the `void_index` the browser pre-pass hands to `processGeometryBatch` for these hosts, and how `non_rect` `IfcOpeningElement` voids are routed in the batch path vs the native `processor.rs` loop.

## Impact

Holes are **silently** missing in the browser viewer for any model using `IfcOpeningElement` voids of this kind. The user waits through the (potentially long) geometry processing and still sees solid plates/members — no error, no failure count.

## How to reproduce the measurement

Native: `process_geometry(&content)`, then sum `indices.len()/3` for meshes whose `express_id` is the target.
WASM: `@ifc-lite/geometry` `GeometryProcessor.process(buf)` in Node (the bridge loads the wasm), same per-`express_id` triangle sum. Compare the two for `#16104`.

Image

louistrue:
Fixes #1297.

## Summary

`IfcOpeningElement` voids (bolt-hole / detail openings on plates and members) were cut correctly in the native pipeline but silently **not** cut in the browser/WASM pipeline, so plates and members rendered **solid** with no error and `total_csg_failures = 0`.

## Root cause

The void cut runs the host **and** its opening cutters in a *shared* coordinate frame: `apply_void_context` relativizes **world-coordinate** opening cutters by the host's origin. One cutter source violated that contract.

In `classify_openings` (`rust/geometry/src/router/voids.rs`), the **high-vertex (>100 verts) opening path** sourced its cutter via `process_element`, which honours `local_frame_enabled()` — **ON for `wasm32`, OFF for native**. Under WASM the opening therefore comes back in the *opening's own* per-element local frame (positions relative to the opening's AABB centre, `mesh.origin` = that centre). Stored as a cutter that way, its origin is never folded back:

- `relativized_by(host_origin)` subtracts only the **host** origin, and
- the #635 fallback AABB is `opening_mesh.bounds()`, which **ignores `origin`**.

So the cutter (and its fallback box) land ~`-opening_origin` away from the host — a building/georef-scale displacement. The AABB-overlap guard skips it, the exact CSG produces no change, the AABB fallback misses too, and the host stays solid. A non-overlapping cutter is a silent no-op, hence `total_csg_failures = 0`.

The ≤100-vertex path was always correct because it sources cutters from `get_opening_item_meshes_world` (explicit `relativize=false`, world coords). The author enforced the world-coords contract there but missed it on the `process_element` high-vertex path. Native was unaffected because its cutters are already world-framed.

## Fix

Fold the opening's per-element origin back to world coordinates right after `process_element` in `classify_openings`, restoring the world-coords cutter invariant. It is a **no-op when `origin == 0`** (native / already world-framed), so native output and the cross-arch determinism snapshots stay byte-identical.

## Verification

End-to-end on the tracked steel fixture `ISSUE_129_N1540_…_EXE_MOD` (361 openings, circular bolt holes):

| Configuration | total triangles |
|---|---|
| native (local frame OFF) | **141,491** |
| local frame ON, **before** fix | **129,705** — 11,786 tris (~8.3%) of void detail dropped |
| local frame ON, **after** fix | **141,240** — restored, within 0.18% of native (residual = f32 precision) |

Adds `opening_void_cut_local_frame_test.rs`: a self-contained synthetic plate with a high-vertex round hole, placed far from the origin, run in its own test binary with the local frame forced on. It **fails without the fix** (solid 3200 m³ plate = full uncut volume) and **passes with it**. The existing wall-opening regression tests (7/7) and the geometry lib unit tests (389/0) stay green.

## Notes / follow-ups (not in this PR)

- **Coverage gap that let this through:** every prior opening-cut test runs on native (local frame OFF) with ≤100-vertex rectangular openings, so none exercised the broken path. This PR's test closes that gap for the high-vertex case.
- **Separate pre-existing finding:** `wall_553010_opening_does_not_empty_wall` fails under `IFC_LITE_LOCAL_FRAME=1` (cut volume 16.5455 vs 16.5468, ~0.008%) — identical with and without this fix, so it is an independent local-frame precision-tolerance issue worth a separate look (local frame is live in the browser).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where void openings were incorrectly skipped during subtraction operations when using per-element local coordinate frames, particularly in WebAssembly environments.

* **Tests**
  * Added regression test to ensure void geometry cutting works correctly with local coordinate frame configurations.

### DIFF
--- a/rust/geometry/src/router/voids.rs
+++ b/rust/geometry/src/router/voids.rs
@@ -820,6 +820,19 @@ impl VoidContext {
         self.openings.is_empty()
     }
 
+    /// True iff every cutter mesh is already in world coords (`origin == 0`). When
+    /// so AND the host is world-framed, `apply_void_context` can run the inner CSG
+    /// directly (the native / legacy fast path) without cloning the cutters; if any
+    /// cutter carries its own per-element origin it must be relativized first.
+    fn all_cutters_world_framed(&self) -> bool {
+        let world = |o: &OpeningType| match o {
+            OpeningType::Rectangular(..) => true,
+            OpeningType::DiagonalRectangular(m, _) => m.origin == [0.0, 0.0, 0.0],
+            OpeningType::NonRectangular(m, ..) => m.origin == [0.0, 0.0, 0.0],
+        };
+        self.openings.iter().all(world) && self.merged_openings.iter().all(world)
+    }
+
     /// Return a copy of this context with every opening (raw + merged)
     /// translated by `-origin`, i.e. moved from the world frame into the host's
     /// per-element local frame. Used by [`GeometryRouter::apply_void_context`]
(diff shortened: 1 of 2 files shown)


louistrue:
Also resolves the audit finding **"opening not subtracted on clipped wall hosts"** (`ISSUE_005_haus.ifc` `#60012`) — that was the same local-frame cutter bug, not a rect-fast-defer issue. Verified on this branch:

| `#60012` | native (local frame off) | local frame on |
|---|---|---|
| before this fix | 5.5504 (cut) | **5.7800 (gross/uncut)** |
| after this fix | 5.5504 | **5.5504 (cut)** |

The reporter measured the viewer (local-frame on), which is why it looked gross; a native-only check sees it cut. Matches the oracle (IfcOpenShell/manifold3d) 5.544 within 0.12%.

PR #1316, opened 2026-06-22
feat(geometry): native geometry-data export + Python (PyO3) binding

louistrue:
## What

An analysis-oriented **geometry-data export** plus a **native Python binding** for the ifc-lite kernel.

- **`rust/processing/src/geometry_export.rs`** — `build_geometry_data_export(&[MeshData], rtc_offset)`: per-entity geometry distinct from the render GLB:
  - IFC **Z-up** (reads `MeshData` before the wasm Y-up boundary)
  - **absolute-world metres**: `vertex = position + origin + rtc_offset` (per-element local frame + model RTC folded back; offset recorded in the output)
  - **position-welded** indexed triangles (1um) so closed-mesh consumers (volume / watertightness) work; the GLB's per-face duplication is avoided
  - occurrences only (`geometry_class == 0`), submeshes merged, keyed by IFC STEP id
  - serializes to the `ifc-lite-geometry-data` JSON contract
- **`rust/python`** — PyO3 cdylib `ifclite_geom` (abi3-py39): `geometry_data_buffers(bytes)` (raw LE byte buffers for `numpy.frombuffer`) and `geometry_data_json(bytes)`. Runs `process_geometry` natively (rayon, 256 MiB worker stack, GIL released).
- **`.github/workflows/python-wheels.yml`** — per-platform abi3 wheels (manylinux x86_64/aarch64, macOS arm64/x86_64, Windows) via maturin-action; PyPI trusted publish on `ifclite-geom-v*` tags.

## Why

Python consumers (compas_ifc and others) had no clean path to ifc-lite geometry: the CLI is Node + wasm + subprocess, and the wasm path can't run `process_geometry` (rayon + `Instant::now`). This is a direct in-process native path — no Node, no wasm, no GLB round-trip. Reading the kernel mesh directly, geometry comes out welded + Z-up + world-coords by construction (the IfcOpenShell frame), so volume/closedness work with no Y-up/recenter handling downstream.

## Results (measured through the compas_ifc integration, separate repo)

- **5-15x faster** than parallel IfcOpenShell end-to-end (file -> per-entity geometry), comparable peak RSS.
- **98.65% geometrically equivalent** to IfcOpenShell across 22k entities (analytic-truth on synthetic shapes + an independent manifold3d referee on real ones); residual is two known wall-opening defects, tracked separately.

## Notes

- `ifc-lite-python` is **excluded from the core workspace** (it has its own wheel CI), so `cargo test --workspace` / clippy / `cargo doc` are unaffected and the root `Cargo.lock` carries no pyo3. It keeps its own `Cargo.lock` + path deps.
- Native test `rust/processing/tests/geometry_data_export_test.rs` verifies welded / Z-up / absolute-world output (skips when the gitignored fixture is absent, per the repo convention).
- The same contract can later feed wasm/JS consumers (browser viewer GLB bridge) on a separate track.

## Release Notes

* **New Features**
  * Added Python library for exporting IFC geometry data to JSON or buffer formats, including vertices, faces, element metadata, and spatial positioning information.

* **Chores**
  * Set up CI/CD pipeline for automated building and publishing of Python packages to PyPI.


louistrue:
Thanks for the review — all three addressed in `4b42d6b0`, plus the Linux wheel build:

- **Site rotation (geometry_export.rs:87)** — fixed. When the model is processed into the `site_local` frame, the export now reapplies the IfcSite forward 3x3 rotation, so vertices are true IFC world coordinates: `world = R * (position + origin) + rtc_offset`. `model_rtc` / `raw_ifc` (the common identity-site case) pass `R = None` and are unchanged. The PyO3 binding only passes the site transform when `mesh_coordinate_space == "site_local"`.
- **Test fixture (geometry_data_export_test.rs)** — fixed. The test no longer depends on the gitignored `synthetic_box.ifc` (which always skipped in CI). It now embeds a minimal inline IFC unit cube and asserts the export welds it to 8 corners / 12 triangles at world `[0,0,0]..[1,1,1]`, so it actually runs as regression coverage.
- **MPL header** — added the MPL-2.0 header to all new Rust sources (`geometry_export.rs`, `rust/python/src/lib.rs`, the test).

Also: the Linux wheel jobs were failing because the manylinux container's toolchain lacked `rust-src` for the repo's global `[unstable] build-std`; added `before-script-linux: rustup component add rust-src`.

PR #1320, opened 2026-06-22
fix(geometry): remove host fully consumed by a containing void

rpisarew:
## Problem

An `IfcOpeningElement` whose **real solid fully contains its host** must remove the host entirely (the boolean result is empty). The exact subtract instead no-ops on the coincident shared faces and returns the host re-triangulated at *unchanged volume*, so the host renders as a **spurious solid** filling the opening.

```
UNCUT   tris=12  vol=0.5255   bounds=(8.5,0.2,5.9)..(10.2,0.3,8.6)
VOIDED  tris=28  vol=0.5255   ← cut no-op'd (volume unchanged) → solid shows
opening tris=12  vol=1.5766   bounds=(8.5,-0.0,5.9)..(10.2,0.3,8.6)  ← engulfs the host
```

The existing `engulfs_host` guard deliberately *keeps* such hosts to protect the AABB-engulf-but-profile-excludes case (`wall_555433`). The missing discriminator is **true solid containment** vs. AABB containment.

## Reproduce

Sample file: [`Ifc2x3_SampleCastle.ifc`](https://github.com/youshengCode/IfcSampleFiles/blob/main/Ifc2x3_SampleCastle.ifc)

Easily reproducible in the live viewer at **https://www.ifclite.com/** — load the sample castle and several thin window-frame panels show up as solid fills inside the window openings (most visible at the dormer / upper-floor windows). Across the model this hits **65 of 71 void-hosts**; all are thin full-span panels, no structural walls.

## Fix

- **`router/voids.rs`** — `point_inside_mesh` + `opening_engulfs_host_solid` (every host vertex, pulled slightly inward, inside the opening's *real* mesh — not just its AABB), plus a per-opening engulf check gated by a cheap AABB pre-check. On a hit, drop the host and record it consumed.
- **`router/mod.rs`** — `voids_consumed_hosts` set + `host_consumed_by_void()` query.
- **`processing/element.rs`** — an empty void-cut normally falls back to the un-cut host; suppress that fallback for consumed hosts so the element correctly renders nothing (otherwise the solid re-appears).

Because the test is true containment, it does **not** fire for `#555433` (AABB engulfs, real profile excludes host) — that wall is kept.

## Tests

- New `engulfing_solid_void.rs` — pins removal, the consumed flag (so the fallback stays suppressed), and a control wall whose ordinary window must keep its geometry.
- Full suites green: **geometry 534/0, processing 67/0**, including `wall_555433_engulfing_opening_does_not_collapse_wall` and `wall_553010_opening_does_not_empty_wall`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved void subtraction for complex, non-rectangular openings by detecting when an opening fully engulfs a host panel, treating the resulting empty mesh as the correct outcome.
  * Refined fallback behavior so expected “void-consumed” empty results no longer cause the uncut host to reappear.

* **Tests**
  * Added a regression test covering a fully engulfing opening (host becomes empty and is flagged as consumed) and a normal window-sized opening (host remains non-empty).

### DIFF
--- a/rust/geometry/src/router/mod.rs
+++ b/rust/geometry/src/router/mod.rs
@@ -33,7 +33,7 @@ use crate::tessellation::TessellationQuality;
 use crate::{BoolFailure, Mesh, Result};
 use ifc_lite_core::{DecodedEntity, EntityDecoder, IfcSchema, IfcType};
 use nalgebra::Matrix4;
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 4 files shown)


Blogbotana:
@louistrue 
This returned the bug again
image

PR #1331, opened 2026-06-23
feat(renderer): axis-aligned clip box (section/crop box) — #1329

### LINKED ISSUE #1329: feat(renderer): multi-plane clip box (section/crop box) — clip geometry between up to 6 planes
## Summary

Add a **multi-plane clip box** (section box / crop box) to the renderer: clip geometry between up to 6 world-space planes simultaneously, exposed through `RenderOptions`.

Today `RenderOptions` supports a **single** `sectionPlane` (`packages/renderer/src/types.ts`). Consumers that want a Navisworks/Revit-style *section box* have no real option — they fake it by hiding whole elements whose **bounding box** lies outside a cube (element-level, coarse: elements straddling the box are shown or hidden wholesale, geometry is never actually cut). A true clip box needs the shader to discard fragments outside the box.

## Current state

- `RenderOptions.sectionPlane?: SectionPlane` — one plane (`axis`/`position` or arbitrary `normal` + `distance`), with cap + outlines.
- Uniform layout (`packages/renderer/src/pipeline.ts`, 192-byte buffer): `sectionPlane: vec4<f32>` (xyz normal + w distance) + `flags: vec4<u32>` (bit0 sectionEnabled, bit1 flipped).
- Fragment discard (`packages/renderer/src/shaders/main.wgsl.ts`): `if (sectionEnabled) { let d = (dot(worldPos, n) - dist) * side; if (d > 0) discard; }`.
- `RenderOptions.terrainClipY` already adds a **second**, independent horizontal clip — so the shader/uniform pattern for "more than one clip test" exists.

## Proposal

Add a clip box to `RenderOptions`, e.g.:

```ts
interface ClipBox {
  // World-space AABB; the primary, common case.
  min: [number, number, number];
  max: [number, number, number];
  enabled: boolean;
}
// or, for generality / oriented boxes:
interface ClipPlanes {
  planes: { normal: [number, number, number]; distance: number }[]; // up to 6
  enabled: boolean;
}
```

- Extend the uniform buffer to carry up to 6 planes (`6 × vec4` = 96 bytes) + a plane count / enabled mask.
- In `main.wgsl.ts`, loop over the enabled planes and `discard` the fragment if it is outside **any** of them (intersection of half-spaces). An AABB reduces to 6 axis-aligned planes.
- **Picking parity**: the pick pass / `picker.ts` must apply the same discard so clipped-away geometry isn't selectable (mirrors how `sectionPlane` is handled).

### Scope

- **Phase 1 (this issue):** clip-only — discard outside the box, render + pick parity, `RenderOptions` API + types + docs, unit/visual tests, and a crop-box toggle in `apps/viewer`.
- **Phase 2 (follow-up, optional):** hatched **caps** on the up-to-6 cut faces, reusing `section-cap-style.ts` / `SectionCapStyleOptions`. Caps for a box are the hard part and can land separately.

### Acceptance criteria

- [ ] `RenderOptions` accepts a clip box / up-to-6 planes; single `sectionPlane` keeps working unchanged.
- [ ] Fragments outside the box are discarded in the main pass (and the cut looks correct from all sides).
- [ ] Picking respects the box.
- [ ] `apps/viewer` gains a crop-box mode driven by a selection's bounds + per-side offsets.
- [ ] Tests cover the multi-plane discard math.

## Motivation / consumer

Pairs with a downstream "Crop via 6 cut planes (not bounding-box + element search)" task in a consumer viewer, which currently can only approximate crop by bbox-isolating elements.

Blogbotana:
Closes #1329.

## What

Adds `RenderOptions.clipBox` — an axis-aligned, world-space clip box (section / crop box). The fragment shader discards anything outside the six box planes, so consumers can crop to a **real geometry cut** instead of bounding-box element isolation.

## How

- `types.ts` — `ClipBox { min, max, enabled }` + `RenderOptions.clipBox`.
- `shaders/main.wgsl.ts` — `Uniforms` gains `clipBoxMin` / `clipBoxMax` (two vec4 lanes); fragment discards when `worldPos` is outside the AABB. `flags.y` bit 2 = clip-box enabled.
- Per-draw uniform grew 192 → 224 bytes (`pipeline.ts` / `constants.ts`); the per-mesh, batched and instanced paths in `index.ts` all pack the box.
- Packing extracted to `clip-box.ts` (`packClipBox`) — one source of truth for all three write sites — with unit tests (`clip-box.test.ts`).

Independent of `sectionPlane`: both can be active at once.

## Scope (Phase 1)

- **Picking parity** is left to the consumer for now — the GPU picker doesn't clip on `sectionPlane` either, so downstream filters section/box picks itself. Happy to add a GPU-side box clip to the picker if preferred.
- **Caps** on the cut faces (hatched fills, like the section plane) are a deliberate follow-up (Phase 2).

## Verification

- `pnpm --filter @ifc-lite/renderer build` clean; **141 renderer tests pass** (incl. the new `packClipBox` tests).
- Wired into a downstream viewer (`RenderOptions.clipBox` from a selection's bounds + per-side margin) and confirmed the cut renders correctly on real models.

This PR is engine-only by design — the `apps/viewer` crop-box UI is intentionally left for a separate change.

## Summary by CodeRabbit

* **New Features**
  * Clip-box clipping: Added `RenderOptions.clipBox` to define an axis-aligned, world-space bounding box for cropping. When enabled, fragments outside `min`/`max` are discarded (and unpickable) using a new clip-box stage that works independently of section-plane clipping.

* **Tests**
  * Added unit tests covering clip-box uniform packing, including correct min/max encoding (with padding cleared), proper handling of enabled/disabled/null inputs, and support for arbitrary float offsets.

### DIFF
--- a/.changeset/clip-box.md
+++ b/.changeset/clip-box.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/renderer": minor
+---
+
+Add `RenderOptions.clipBox` — an axis-aligned, world-space clip box (section / crop box). The fragment shader discards geometry outside the six box planes, so consumers can crop to a real geometry cut instead of bounding-box element isolation. Independent of `sectionPlane`; both can be active. (#1329)
--- a/packages/renderer/src/clip-box.test.ts
+++ b/packages/renderer/src/clip-box.test.ts
@@ -1,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Tests for the clip-box uniform packing. `packClipBox` is the single source of
+ * truth shared by `pipeline.updateUniforms` and the renderer's per-mesh /
+ * instanced template loops, so the min/max lanes, the zeroed padding + w lanes,
+ * and the returned flags bit must stay exactly as `main.wgsl` reads them.
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+
+import { packClipBox, CLIPBOX_ENABLED_BIT } from './clip-box.js';
+import type { ClipBox } from './types.js';
+
+describe('packClipBox', () => {
+  const box: ClipBox = { min: [-1, -2, -3], max: [4, 5, 6], enabled: true };
+
+  it('writes min.xyz / max.xyz at the offset and returns the enabled bit', () => {
+    const out = new Float32Array(56);
+    const bit = packClipBox(box, out, 48);
+    assert.strictEqual(bit, CLIPBOX_ENABLED_BIT);
+    assert.deepStrictEqual([...out.slice(48, 51)], [-1, -2, -3]);
+    assert.deepStrictEqual([...out.slice(52, 55)], [4, 5, 6]);
+  });
+
+  it('zeroes the w padding lanes (51 and 55)', () => {
+    const out = new Float32Array(56).fill(9);
+    packClipBox(box, out, 48);
+    assert.strictEqual(out[51], 0);
+    assert.strictEqual(out[55], 0);
+  });
+
+  it('returns 0 and zeroes the region when disabled', () => {
+    const out = new Float32Array(56).fill(7);
+    const bit = packClipBox({ ...box, enabled: false }, out, 48);
+    assert.strictEqual(bit, 0);
+    for (let i = 48; i < 56; i++) assert.strictEqual(out[i], 0, `lane ${i}`);
+  });
+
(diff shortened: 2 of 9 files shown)


Blogbotana:
Addressed the review + CI:

- **Clip-box bit lost in transparent/selected draws** (CodeRabbit, Major): the transparent-instanced and selected per-mesh template rewrites reassigned `tplFlags[1]` without the clip bit, disabling the box for those draws when batches are present. OR'd `tplClipBit` back into both (`index.ts`). The clip-box min/max lanes already persist in the shared `tpl` from the template setup, so only the flag bit needed restoring.
- **Check API surface** gate: recorded `ClipBox` in `scripts/api-surface.json`.
- Added a changeset (`minor` for `@ifc-lite/renderer`).

141 renderer tests still pass; tsc clean.

PR #1335, opened 2026-06-24
feat(renderer): make section/crop-clipped geometry unpickable in the GPU picker

louistrue:
Follow-up to #1331 / #1329. Reported: the **section tool still selects the invisible, cut-away geometry** (and the same was true of the crop box).

## Root cause

Two gaps: the GPU picker had no clip uniforms at all, and the viewer's `getPickOptions()` only carries `{ isStreaming, hiddenIds, isolatedIds }`, so no section/clip state ever reached `pick()`. Cropped/sectioned fragments stayed selectable and still won occlusion in the pick pass.

## Fix (renderer-stash, no viewer wiring)

The renderer stashes what it actually clipped each `render()` (the resolved section plane normal/distance/flipped, including the `terrainClipY` plane, plus the active `clipBox`) and feeds it to the picker on `pick`/`pickRect`. Both pick shaders (flat + instanced) gain `sectionPlane` + `clipBox` uniforms and discard on the cut-away side / outside the AABB, mirroring `main.wgsl` exactly (flat on its already-world-space positions, instanced on `m * position`). Picker uniform grows 64 to 128 bytes.

Deriving from the last render means:
- selection always matches what's visible, with **zero consumer wiring**, and
- it's robust to the viewer only sending the section plane while its tool is active: whenever render clips, pick clips identically; when render doesn't, neither does pick.

Points are not clipped, matching the render pass.

## Verification

- Packing extracted to a pure `packPickUniforms` with **7 unit tests** locking the lane layout + flag bits.
- Renderer tests **148/148** green; typecheck clean (only the known unbuilt-sibling `@ifc-lite/geometry`/`spatial` TS2307 in a fresh worktree).
- GPU discard behaviour is exercised by the E2E smoke; final confirmation is in the live viewer.

Supersedes the earlier `PickOptions.clipBox` approach from this branch (removed): the renderer-stash model fixes the section plane too and needs no viewer changes.

## Summary by CodeRabbit

* **New Features**
  * Picking now mirrors the latest render’s active section plane and clip box, so results match what’s visible.
  * Point picking now also respects the section plane during selection.

* **Bug Fixes**
  * Geometry clipped away by the section plane and/or clip box is no longer selectable for both `pick` and `pickRect`, including batched and released-geometry CPU fallback behavior (with fallback to the nearest visible surface behind the cut).
  * Improved geometry robustness: corrected normal transforms under non-uniform scaling, stricter circle-hollow profile validation, more accurate arc segmentation, and better outer/hole ring handling in BREP extraction.


louistrue:
Addressed the review feedback in 06a2ad4c:

- **Snapshot the clip box / section (CodeRabbit, index.ts:1487).** The renderer stash now deep-copies `normal`/`min`/`max` and rebuilds the box object, so a post-`render()` in-place mutation can't make `pick()` mirror a different cut.
- **Partially-clipped boxes in the released-geometry raycast (CodeRabbit, scene-raycaster.ts).** New `clippedBoxEntryDistance` returns the entry into the *visible* part of a box (box intersect crop box, plus the section half-space), so a box that only overlaps the ray in the cropped/sectioned region can no longer win `raycastBoundingBoxes`. Covered by new unit tests (full + partial clip).
- **Point picking vs the section plane (CodeRabbit, types.ts:280).** Confirmed point clouds *are* section-clipped on render (`point-shader.wgsl.ts`), so the point picker now discards on the same plane; the crop box still clips meshes only. The doc claim was corrected accordingly.
- **Apply clip to the CPU pick fallback (Codex P2, picking-manager.ts:218).** Done in 077b7a1c: both `scene.raycast` paths now receive the clip state and skip clipped hits.

Renderer suite 160/160; all functional CI green. Debug `[Pick]` logs removed.

PR #1352, opened 2026-06-24
feat(viewer): refresh button to reload the open model from disk (#1345)

### LINKED ISSUE #1345: feature request | refresh model
Hi, if possible, it would be nice to have a `Refresh` button to reload the model already in the UI, so to monitor it during design without closing and opening the same file again.
Thanks a lot for your work, truly amazing and appreciated
S.

louistrue:
## Summary

Closes #1345. Adds a **Refresh** button to reload the open model without closing and re-opening the same file, so it can be monitored during design.

The existing Open flow uses `<input type="file">`, which hands back a File that is a *frozen snapshot* of the bytes at pick time. Re-reading it can never reflect on-disk edits. Only the **File System Access API** (`showOpenFilePicker` → `FileSystemFileHandle.getFile()`) returns a fresh snapshot of the current file. So the Open button now prefers that API when available (Chromium / secure context) to capture a live handle, and falls back to `<input type="file">` everywhere else.

## Changes (`apps/viewer`)

- **`services/file-system-access.ts`** (new) — `supportsFileSystemAccess()`, `openIfcFilesWithHandles()` (picks files + captures live handles; user-cancel handled), `readFreshFile(handle)` (permission-checked re-read of current on-disk bytes; surfaces moved/deleted/denied).
- **`types/file-system-access.d.ts`** (new) — minimal ambient declarations for the API surface missing from TS 6.0.3 `lib.dom` (`Window.showOpenFilePicker`, `OpenFilePickerOptions`, `queryPermission`/`requestPermission`). `FileSystemFileHandle.getFile()` already ships in lib.dom.
- **`store/types.ts`** — `FederatedModel.sourceHandle?: FileSystemFileHandle` (in-memory only, never serialized to cache; mirrors `sourceFile`).
- **`hooks/useIfcLoader.ts`** — `loadFile` accepts an optional `{ sourceHandle }` and persists it on the primary model record.
- **`components/viewer/MainToolbar.tsx`** — Open button prefers the FS Access API; new Refresh button (`RefreshCw`).

## Behaviour / scope

- Refresh is **only shown** when exactly one model is loaded **and** it has a live handle. Drag-drop, `<input type="file">`, recent-files-from-cache, and federated multi-model loads have no handle, so the button isn't offered (no dead/disabled control).
- Multi-file Open keeps the existing federated/sequential paths unchanged.
- Non-Chromium browsers (Firefox/Safari) and insecure contexts fall back to `<input type="file">` transparently — no Refresh button, no regression.

## Testing

- `tsc --noEmit` clean for all touched files (remaining errors are pre-existing un-built-package gaps in the worktree, unrelated to this change).
- Not yet smoke-tested in a running viewer (requires built wasm pkg + dev server).

## Follow-up (not in this PR)

- Drag-and-drop could also capture handles via Chromium `DataTransferItem.getAsFileSystemHandle()`, extending Refresh to dropped files.

## Summary by CodeRabbit

* **New Features**
  * Added a “Refresh” action that reloads models from their original disk source when supported.
  * Improved Open, Add Model, and drag-and-drop to use file picker handles when available, enabling refresh-aware loading.
  * Enhanced recent files to make faster, cache-aware loading decisions.
* **Bug Fixes**
  * Updated command palette file-dialog activation to more reliably preserve browser interaction state.
  * Improved recent-file caching performance and resilience for more consistent loads.

### DIFF
--- a/apps/viewer/src/components/viewer/CommandPalette.tsx
+++ b/apps/viewer/src/components/viewer/CommandPalette.tsx
@@ -92,7 +92,7 @@ import { SCRIPT_TEMPLATES } from '@/lib/scripts/templates';
 import { exportGlbFromGeometry } from '@/lib/export/glb';
 import { exportCsvFromBytes } from '@/lib/export/csv';
 import { downloadFile } from '@/lib/export/download';
-import { getRecentFiles, formatFileSize, getCachedFile } from '@/lib/recent-files';
+import { getRecentFiles, formatFileSize, getCachedFile, getCachedFileNames } from '@/lib/recent-files';
 import type { RecentFileEntry } from '@/lib/recent-files';
 import { closeActiveAnalysisExtension } from '@/services/analysis-extensions';
 import { describeRunCommandError } from '@/services/extensions/runtime-errors';
@@ -120,6 +120,13 @@ interface Command {
   shortcut?: string;
   detail?: string;            // subtle secondary text (e.g. file size)
   action: () => void;
+  /**
+   * Run the action synchronously inside the click handler instead of deferring
+   * to the next animation frame. Required for actions that open a file dialog:
+   * Chrome only honours `input.click()` / `showOpenFilePicker()` while transient
+   * user activation is live, which a `requestAnimationFrame` hop would discard.
+   */
+  immediate?: boolean;
 }
 
 interface FlatItem {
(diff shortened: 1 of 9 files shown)


louistrue:
Addressed the CodeRabbit review in `ac55152b`:

- Case-insensitive IFCX routing via `isIfcxModelFile` (so `MODEL.IFCX` composes correctly).
- Filter the picker's "all files" results before loading in `handleOpenClick`.
- Await the single-model reload so Refresh reports success only after it completes.
- `aria-label` on the icon-only Refresh button.
- Drop handling pairs file+handle from the same item (no filename matching).
- Log `getAsFileSystemHandle` failures instead of swallowing them.
- Bound `cacheFileBlobs` staged reads to the cache capacity.
- Corrected the `sourceHandle` lifecycle comment.

Also fixed (separate finding, surfaced while testing on Chrome): `cacheFileBlobs` awaited `file.arrayBuffer()` mid-IndexedDB-transaction, which auto-committed the txn and made every `put()` throw `TransactionInactiveError` — so the recent-files cache never persisted and command-palette recents always re-picked. Reads are now staged before the write.

PR #1360, opened 2026-06-24
feat(renderer): themeable overlay/section-cut line colour (#1359)

### LINKED ISSUE #1359: renderer: annotation/grid 3D lines are hardcoded black — expose a color (unreadable on dark UI)
### Summary

The 3D annotation/grid line overlay renders **every line with a hardcoded black color**, with no API to change it. On a dark canvas background the structural grid lines and 2D annotation/dimension lines are essentially invisible (black on dark).

### Where

`packages/renderer/src/section-2d-overlay.ts` — the line pipeline's fragment shader:

```wgsl
@fragment
fn fs_main(input: VertexOutput) -> FragOutLine {
  var out: FragOutLine;
  out.color    = vec4<f32>(0.0, 0.0, 0.0, 1.0);  // Black lines   // <-- hardcoded
  out.objectId = vec4<f32>(0.0, 0.0, 0.0, 0.0);
  return out;
}
```

These lines are uploaded via `Renderer.uploadAnnotationLines3D(vertices: Float32Array)`, whose vertex buffer is **position-only** (`arrayStride: 12 // 3 position floats`), so there's no per-vertex color either.

### Contrast: text already supports color

`SymbolicTextInput` exposes an optional `color?: [number, number, number, number]`, so grid/annotation **text** (`uploadAnnotationTexts3D`) can follow a host's light/dark theme. The **lines** have no equivalent, so a consumer can theme the labels but not the lines that go with them.

### Impact

A consumer rendering on a dark background can make the labels legible (via `SymbolicTextInput.color`) but the grid axes and 2D annotation/dimension lines stay black and disappear into the background.

### Proposed fix (any of)

1. A renderer option / setter for the annotation line color (e.g. `setAnnotationLineColor(rgba)` or a render option), read as a uniform in the line shader — smallest change.
2. Per-vertex color in the annotation-line vertex buffer (a `stride: 28 // 3 position + 4 color` layout already exists for the cap-fill pipeline), letting callers color lines individually.
3. An optional `color` parameter on `uploadAnnotationLines3D(...)`.

Option 1 covers the common "theme the whole overlay one color" case with minimal API surface.

### Environment

- `@ifc-lite/renderer` ~1.29.x, WebGPU.

Image

Blogbotana:
## Summary

Fixes #1359. The 3D overlay lines (annotation / alignment / grid) and the section-cut outline shared a line shader that **hardcoded** `out.color = vec4<f32>(0.0, 0.0, 0.0, 1.0)`. On a dark canvas these lines were essentially invisible, and a consumer could theme the matching **labels** (`SymbolicTextInput.color`) but not the **lines**.

## Change

- Add a `lineColor` uniform to the line shader and drive `out.color` from it.
- New API: `Renderer.setOverlayLineColor(rgba)` (delegates to `Section2DOverlayRenderer.setOverlayLineColor`). RGBA in 0..1.
- **Default is opaque black**, so existing behaviour is unchanged unless a consumer sets it.

### Buffer layout note

The line and fill pipelines share one uniform buffer. `lineColor` is appended at **byte offset 144** (after the fill pipeline's `params2`), so it never aliases the fill cap-style fields — the section-cut outline keeps its own colour independent of `capFillColor`. Buffer grows 144 → 160 B; the line shader mirrors the fill struct's field offsets and reads `viewProj` / `planeOffset` / `lineColor`.

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing overlay line color via `Renderer.setOverlayLineColor(rgba)`.
  * 3D annotation, alignment, grid, and section-cut outline line overlays now use the configured RGBA color.
  * If not set, the renderer keeps the prior look by defaulting to opaque black.

### DIFF
--- a/.changeset/themeable-overlay-line-color.md
+++ b/.changeset/themeable-overlay-line-color.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/renderer": minor
+---
+
+renderer: add `Renderer.setOverlayLineColor(rgba)` so the 3D overlay lines (annotation / alignment / grid) and the section-cut outline are themeable. The line shader previously hardcoded black, leaving these lines invisible on dark backgrounds; the colour now comes from a uniform and defaults to opaque black (no behaviour change unless set). Complements `SymbolicTextInput.color`, which already themes the matching labels.
--- a/packages/renderer/src/index.ts
+++ b/packages/renderer/src/index.ts
@@ -177,6 +177,9 @@ export class Renderer {
     private canvas: HTMLCanvasElement;
     private sectionPlaneRenderer: SectionPlaneRenderer | null = null;
     private section2DOverlayRenderer: Section2DOverlayRenderer | null = null;
+    // Overlay/section-cut line colour, kept on the Renderer so it survives a
+    // pre-init call and a section2DOverlayRenderer re-creation (re-applied below).
+    private overlayLineColor: readonly [number, number, number, number] = [0, 0, 0, 1];
     // IfcAnnotation overlay pipelines (issue #653). Created on `init()` once
     // the device exists; nulled until then.
     private symbolicFillPipeline: SymbolicFillPipeline | null = null;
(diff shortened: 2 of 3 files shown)


Blogbotana:
Addressed the review in 6825d2f7: `setOverlayLineColor` now persists the RGBA on `Renderer` and `init()` re-applies it after (re)constructing `Section2DOverlayRenderer`, so pre-init calls and overlay re-creation no longer reset to default black. Build + 162 tests green.

PR #1369, opened 2026-06-25
fix(viewer): allow editing/renaming a saved auto-color lens (#1365)

### LINKED ISSUE #1365: Edit Auto color lens
When creating an Auto color lens I can set the name.
After I saved I cannot edit the lens, I need to delete it and recreate it.
Even when the only thing I want is to change the name of the Lens.

Image

Image
Image

Thanks!

louistrue:
## Problem

After creating an **Auto color** lens and saving it, it could not be edited — to change even just the name you had to delete and recreate it (#1365).

## Root cause

Three things combined:
1. The pencil **edit button was hidden** for auto-color lenses (`onEdit && !isAutoColor`).
2. Even if triggered, editing routed to the **rule editor** (`LensEditor`), which has no notion of `autoColor`.
3. `AutoColorEditor.handleSave()` always minted a **new id** (`lens-auto-${Date.now()}`), so a save while editing was treated as a *create* (duplicate) rather than an *update* — `handleSaveLens` decides create-vs-update by whether the id already exists.

## Fix

- Show the edit button for auto-color lenses.
- Route auto-color edits to `AutoColorEditor`, pre-filled with the existing lens (`id`, name, spec).
- Preserve the lens id on save via a small pure helper `buildAutoColorLensToSave(initial, values, generateId)` — existing id is kept when editing; a fresh id is generated only for a brand-new lens.

## Tests

- New `lens-editor-utils.test.ts` pins the id-preservation invariant: editing keeps the original id (and never calls the id generator); creating mints a fresh one. **Fails on the old logic, passes on the fix.**
- `turbo typecheck --filter=@ifc-lite/viewer` clean (29/29).

(`apps/viewer` is a private app — no changeset required.)

Fixes #1365

## Summary by CodeRabbit

* **New Features**
  * Improved auto-color lens editing, including better handling of existing saved items and newly created ones.
  * The edit button now appears for all lenses when editing is available.
* **Bug Fixes**
  * Auto-color lens edits now preserve an existing lens ID when updating, or generate a new one when creating a fresh lens.
  * Editing now opens the correct editor for auto-color and rule-based lenses.

### DIFF
--- a/apps/viewer/src/components/viewer/LensPanel.tsx
+++ b/apps/viewer/src/components/viewer/LensPanel.tsx
@@ -1036,7 +1036,7 @@ function LensCard({
           </span>
         </div>
         <div className="flex items-center gap-1">
-          {onEdit && (
+          {onEdit && !lens.builtin && (
             <button
               onClick={(e) => { e.stopPropagation(); onEdit(lens); }}
               className="opacity-0 group-hover:opacity-100 text-zinc-400 hover:text-zinc-700 dark:text-zinc-500 dark:hover:text-zinc-200 p-0.5"
--- a/apps/viewer/src/components/viewer/LensPanel.tsx
+++ b/apps/viewer/src/components/viewer/LensPanel.tsx
@@ -26,6 +26,7 @@ import { downloadFile } from '@/lib/export/download';
 import { useViewerStore } from '@/store';
 import { useLens } from '@/hooks/useLens';
 import { createLensDataProvider } from '@/lib/lens';
+import { buildAutoColorLensToSave } from './lens-editor-utils';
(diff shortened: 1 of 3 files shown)


louistrue:
Thanks @coderabbitai and Codex — both findings were valid. Removing the `!isAutoColor` guard unintentionally exposed the edit button on built-in presets (including auto-color ones), and `saveLenses` only persists `name`/`rules` overrides for built-ins, so an auto-color edit to a preset would be dropped on reload.

Fixed in `2b54373` by gating edit with `!lens.builtin`, matching the existing delete gate — built-in presets stay read-only, and the persistence gap is no longer reachable. User-created lenses are not `builtin`, so editing/renaming them (the actual #1365 report) is unaffected.

PR #1370, opened 2026-06-25
Clarify MeshData JSDoc for downstream consumers

mehmet-ylcnky:
Adds documentation to MeshData and processAdaptive addressing common integration pitfalls discovered while building a Node.js rendering pipeline:

- **MeshData interface**: Note that multiple meshes per expressId is expected (one per material/part). Consumers should group by expressId for per-element operations.
- **indices**: Warn that winding order is unreliable (double-sided by design). Back-face culling by winding silently drops real faces.
- **geometryClass**: Clarify rendering implications with a concrete filter example. Class 2 must not be rendered in normal views or overlapping duplicates appear.
- **shadingColor**: Add guidance on priority (use `color` for basic rendering, `shadingColor ?? color` for physically-accurate).
- **processAdaptive**: Document that batches may contain same-expressId meshes.

No code changes, only JSDoc additions.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how streaming geometry results are delivered, including batch events and how to group meshes for per-element rendering or picking.
  * Expanded guidance on mesh data handling, including index usage, color selection for accurate rendering, and when to prefer `shadingColor` over `color`.
  * Added clearer explanations for geometry classification values and how they should be interpreted in downstream filtering and rendering.

### DIFF
--- a/packages/geometry/src/index.ts
+++ b/packages/geometry/src/index.ts
@@ -779,6 +779,9 @@ export class GeometryProcessor {
    * @param options.sizeThreshold File size threshold in bytes (default: 2MB)
    * @param options.batchSize Number of meshes per batch for streaming (default: 25)
    * @param options.entityIndex Optional entity index for priority-based loading
+   * @yields StreamingGeometryEvent with 'batch' events containing MeshData[].
+   *   Multiple meshes may share the same expressId (one per material/part).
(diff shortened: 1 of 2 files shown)


louistrue:
Thx, @mehmet-ylcnky

PR #1372, opened 2026-06-25
Re-export EntityTable and SpatialHierarchy types from parser

mehmet-ylcnky:
Re-exports `EntityTable` and `SpatialHierarchy` types from `@ifc-lite/parser` so consumers can import them directly without reaching into `@ifc-lite/data`.

```ts
import type { IfcDataStore, EntityTable, SpatialHierarchy } from '@ifc-lite/parser';
```

Removed the `initNode()` WASM helper from this PR per review feedback. It needs to live in the outer `packages/wasm/package.json` and survive `pkg/` regeneration. Will submit separately once the packaging is sorted.

TODO before merge:
- `pnpm api-surface:update` (regenerate snapshot for new exports)
- `pnpm changeset` (minor bump for `@ifc-lite/parser`)

## Summary by CodeRabbit

* **New Features**
  * Expanded the parser package’s public type exports to include additional data structures for spatial hierarchy and entity tables.
  * These types are now available directly from the parser package, aligning the declared API surface with what consumers can import.


louistrue:
Thanks for this! The parser type re-exports are a nice quality-of-life win. A couple of things to sort before this can go green/merge.

## Why CI is red

There's really only one real failure; the rest are noise:

- **`Node tests` → "Check API surface"** is the actual blocker. Adding `EntityTable` and `SpatialHierarchy` to the `@ifc-lite/parser` re-exports changes the package's public surface, and the guard (`scripts/check-api-surface.mjs`) compares that against the committed `scripts/api-surface.json` snapshot. Fix:
  1. `pnpm api-surface:update` (regenerates the snapshot, commit it)
  2. `pnpm changeset` (new exports = a minor bump for `@ifc-lite/parser`)
- **`Build + WASM + Rust + Node`** is just a gate job that aggregates the others; its log shows `node-tests="failure"`, so it goes green automatically once the surface check passes. Nothing to fix here directly.
- The **Vercel** checks ("Authorization required to deploy") are expected for a fork PR and aren't related to your changes.

The parser half is otherwise correct, both types genuinely resolve to interfaces in `@ifc-lite/data`.

## The `initNode()` half needs to move

Heads up that the WASM helper, as currently wired, won't actually work for published consumers, and CI won't catch it. The edits went into `packages/wasm/pkg/package.json`, but that's the **inner, wasm-pack-generated** manifest. The published `@ifc-lite/wasm` package is the **outer** `packages/wasm/package.json`, and that's the one that governs subpath resolution and the published tarball. As a result:

1. **Won't resolve.** The outer manifest's `exports` only has `.` and `./ifc-lite_bg.wasm`, so `import { initNode } from '@ifc-lite/wasm/init-node'` would fail for consumers.
2. **Won't ship.** The outer `files` array lists only `pkg/ifc-lite_bg.wasm`, `pkg/ifc-lite.js`, `pkg/ifc-lite.d.ts`, so `init-node.*` wouldn't be in the npm tarball.
3. **Gets clobbered.** `packages/wasm/pkg/` is fully gitignored (`*`) and regenerated by `wasm-pack` on every `scripts/build-wasm.sh` run (release runs `pnpm build`), so the inner exports map would be overwritten.

To make it real, the `./init-node` export and the `pkg/init-node.*` `files` entries should live in the **outer** `packages/wasm/package.json`, and `init-node.js` needs a home that survives a `pkg/` regeneration. Once the export is on the outer manifest, the api-surface guard will (correctly) also flag `@ifc-lite/wasm/init-node`, so you'd refresh the snapshot and add a `@ifc-lite/wasm` changeset too.

If it's easier, splitting this into two PRs (parser re-exports now, the WASM init helper once the packaging is sorted) would let the straightforward half land quickly.

mehmet-ylcnky:
Splitting this into two PRs as suggested. Parser type re-exports in #1373, initNode() helper will follow separately once the packaging is sorted (outer package.json, survives pkg/ regeneration).

PR #1374, opened 2026-06-25
fix(lens): By-Material groups by individual materials, not the layer-set name (#1366)

### LINKED ISSUE #1366: Lens 'By Material' is not showing the actual materials
The lens rule to color materials is showing the layerset instead.

This screenshot is one element where the Revit Family + TypeName is shown instead of the material
Image

This is just one layer and it happens with multiple-layered elements as well.

Image

This is how BIMcollab ZOOM shows it: comma seperated.
Image

That's fine but what here's what happens:
a wall that has 2 layers: gypsumboard and insulation
gypsumboard is also part of another wall and it occurs as seperate walls.

When you add them comma seperated you get:

2 walls: gypsumboard
3 walls: gypsumboard, insulation
1 wall: gypsumboard, insulation, gypsumboard 
4 walls: insulation

Now I still can't select all elements that contain gypsumboard. (also because I can select one value at a time, not multiple by using shift/ctrl select)
When I select gypsumboard I think I should see 4 walls. When I select insulation i should see 8 walls.

Does that makes sense?

louistrue:
## Problem

The "By Material" lens (and material rule criteria) grouped elements by the **layer-set / usage name** — for Revit exports that's the family + type string, not a material. A wall of *gypsum board + insulation* showed up under one opaque label, and you couldn't select all elements containing a given material: selecting "gypsum board" missed multi-layer walls (#1366).

## Root cause

`LensDataProvider.getMaterialName()` returns a single value — `MaterialInfo.name` (the layer-set name) or the first layer — and `evaluateAutoColorLens` puts each element in exactly one bucket.

## Fix

Make material grouping multi-valued and key off the **individual constituent materials**:

- **engine**: new `extractAutoColorValues` — `material` returns every distinct material an element has, so the element joins each material's group. `colorMap` de-dupes (an element still renders one colour — its largest group wins); `ruleEntityIds` / legend counts include it in every material bucket (so isolation works). All other sources are unchanged (single value).
- **matching**: `matchesMaterial` checks each individual material, so a `materialName contains "gypsum"` rule matches multi-layer walls.
- **provider**: new optional `getMaterialNames(globalId)`; absent → falls back to the old single `getMaterialName`.
- **viewer**: implements it via a pure `lensMaterialNames(MaterialInfo)` collector — gathers layer/constituent/profile/list material names, **excludes** the layer-set name, and for a single plain material falls back to the top-level name. De-duplicated, so gyp/ins/gyp → `["gyp", "ins"]`.

Result: "gypsum board" now selects all 4 walls containing it; "insulation" selects all 8; the legend lists real materials.

## Scope

Multi-select of legend values (shift/ctrl to combine materials) is a separate UI enhancement and is **not** in this PR — the core data-correctness bug (per-material grouping + isolation) is fixed here.

## Tests

- `lensMaterialNames` collector: layer-set name excluded, dedup, single-material fallback, constituents/profiles — 5/5.
- engine: multi-material element lands in **both** buckets with correct counts; single-`getMaterialName` fallback path — `@ifc-lite/lens` 84/84.
- `turbo typecheck` clean (lens, viewer). Minor changeset for `@ifc-lite/lens` (additive optional accessor); no top-level export change so api-surface is unchanged.

Fixes #1366

### DIFF
--- a/.changeset/fix-1366-by-material-individual.md
+++ b/.changeset/fix-1366-by-material-individual.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/lens": minor
+---
+
+"Color/select by Material" now groups by individual materials instead of the layer-set / usage name. A multi-layer element (e.g. a wall of gypsum board + insulation) now belongs to each of its materials' groups, so selecting "gypsum board" isolates every wall containing it — including multi-layer walls — and the legend lists the real materials rather than the Revit family/type string. Material rule criteria (`materialName`) match the same way. Adds an optional `LensDataProvider.getMaterialNames(globalId)` accessor; providers without it fall back to the previous single `getMaterialName` value. (#1366)
--- a/apps/viewer/src/lib/lens-material-names.test.ts
+++ b/apps/viewer/src/lib/lens-material-names.test.ts
@@ -1,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+
+import { lensMaterialNames } from './lens-material-names.js';
+
+describe('lensMaterialNames (#1366)', () => {
+  it('returns each layer material of a multi-layer set, not the layer-set name', () => {
+    const info = {
+      type: 'MaterialLayerSet' as const,
+      name: 'Basic Wall: Ext - Gyp/Ins', // layer-set / Revit type name — must NOT become a bucket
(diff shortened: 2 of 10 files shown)


louistrue:
Valid @chatgpt-codex-connector — thanks. Fixed in `5696af22`: `matchesMaterial` now matches if **either** an individual constituent material **or** the single/layer-set name contains the pattern (only falling through to the legacy pset scan when no dedicated accessor exists), so a rule built from the dropdown value (`Basic Wall: Ext - Gyp/Ins`) still matches. Additionally, `discoverDataSources` now surfaces `getMaterialNames` so the rule dropdown offers the individual materials (`Gypsum`, `Insulation`) rather than the layer-set string. Covered by new matching tests.

PR #1375, opened 2026-06-25
fix(geo): stop spurious 'grid failed' for S-JTSK models (#1357)

### LINKED ISSUE #1357: 'grid failed' in specific model
I am receiving a `grid failed` info for the specific model. The fallback is engaged.

I checked and I am able to access the `https://cdn.proj.org/` address from the browser.

Image

Sample file (remove the .txt extension):

[garage.ifc.txt](https://github.com/user-attachments/files/29305438/garage.ifc.txt)

I am able to reproduce with several files with S-JTSK CRS.

louistrue:
## Problem

Czech **S-JTSK** models (EPSG:5514 / 2065) show a `grid failed` badge and engage the `+towgs84` fallback, even though `cdn.proj.org` is reachable (#1357).

## Root cause (reproduced, not inferred)

The precision-grid entry for 5514/2065 pointed `+nadgrids` at **`cz_cuzk_CR-2005.tif`**. I probed the CDN and ran the exact load path in Node:

- Filename is correct (`HTTP 200`, `image/tiff`, 106 KB) and CORS is fine (`access-control-allow-origin: *` with an `Origin` header). So it's **not** 404/CORS/network.
- The grid is **single-band**. PROJ-data's own catalog classifies it as `type=VERTICAL_OFFSET_GEOGRAPHIC_TO_VERTICAL` (ETRS89 → Baltic 1957 height) — a **height model, not a horizontal datum shift**.
- proj4's `+nadgrids` reader (`readGeotiffGrid`) reads two bands (band 0 = latitude offsets, band 1 = longitude offsets). With one band, `data[1]` is `undefined` and it throws `Cannot read properties of undefined (reading '53549')` (53549 = 306×175 − 1, the last pixel). `loadPrecisionGrid` catches it → `+towgs84` fallback + the misleading "grid failed" badge + a wasted fetch/crash on every S-JTSK model.

PROJ-data has **no** single horizontal S-JTSK → ETRS89 grid (its only Czech `HORIZONTAL_OFFSET` grid, `cz_cuzk_table_-y-x_3_v1710.tif`, is S-JTSK → S-JTSK/05). So `+towgs84` (~1 m) is the correct path for these CRSs.

## Fix

1. **Remove the 5514 / 2065 entries** and the vertical-grid filename; document why (so nobody re-adds it). These CRSs now go straight to `+towgs84` with no failed-grid badge.
2. **Defensive band-count guard** (`isHorizontalOffsetGrid`): reject `<2`-band grids before handing them to proj4, turning an opaque internal crash into a clear fallback for any future mis-registration.

Net effect for the reporter: the model still renders (it already did, via the fallback), but the spurious `grid failed` error and the wasted crash are gone.

## Verification & tests

- Verified band counts on `cdn.proj.org`: CR-2005 = 1 band (vertical), NL RD = 4, Slovak JTSK03 = 2 (genuine horizontal — untouched).
- Unit tests: band-guard logic; a regression guard that **no** entry references the vertical CR-2005 grid; 5514/2065 absent; genuine horizontal grids retained. 4/4. `turbo typecheck` clean.

Note: this removes the misleading error and the broken code path; it does not add a Czech grid (none usable by proj4js exists). Sub-decimeter S-JTSK would need a multi-step pipeline proj4js can't express as one `+nadgrids`.

Fixes #1357

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved coordinate reprojection reliability by avoiding broken datum-shift grids and falling back to safer definitions when needed.
  * Fixed handling for some Czech coordinate systems so they no longer use an incompatible grid-based shift.
  * Added safeguards to reject invalid grid data before it can cause projection errors, while keeping valid horizontal grids working as expected.

### DIFF
--- a/apps/viewer/src/lib/geo/precision-grids.test.ts
+++ b/apps/viewer/src/lib/geo/precision-grids.test.ts
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+
+import { PRECISION_GRIDS, isHorizontalOffsetGrid } from './precision-grids.js';
+
+describe('precision-grids datum-shift table (#1357)', () => {
+  it('treats only >=2-band grids as horizontal datum shifts', () => {
+    // 1-band = vertical/height model (e.g. a quasigeoid) — proj4's +nadgrids
+    // reader would dereference a missing 2nd band and throw.
+    assert.equal(isHorizontalOffsetGrid(1), false);
+    assert.equal(isHorizontalOffsetGrid(0), false);
+    assert.equal(isHorizontalOffsetGrid(2), true); // lat + lon offsets
+    assert.equal(isHorizontalOffsetGrid(4), true); // + accuracy bands
+  });
+
+  it('never registers the vertical cz_cuzk_CR-2005 grid as a horizontal +nadgrids', () => {
+    // cz_cuzk_CR-2005.tif is a VERTICAL_OFFSET (ETRS89 -> Baltic height) grid;
+    // using it as a horizontal datum shift crashed proj4 and forced the
+    // +towgs84 fallback (the "grid failed" badge). It must not reappear.
+    for (const [code, spec] of Object.entries(PRECISION_GRIDS)) {
+      assert.ok(
+        !/cz_cuzk_CR-2005/i.test(spec.filename),
+        `EPSG:${code} references the vertical CR-2005 grid (${spec.filename})`,
+      );
+      assert.ok(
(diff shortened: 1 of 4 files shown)


louistrue:
Good catch @chatgpt-codex-connector — valid. EPSG:2065's bundled Ferro-Krovak def has `+pm=ferro` but no `+towgs84`/`+nadgrids`, so with 2065 out of the grid table proj4 did no datum shift at all. Fixed in `29a77d33`: `sanitizeProj4` now also injects the datum's `+towgs84` when a def carries neither a `+nadgrids` nor a `+towgs84` and its datum is a known offset datum (S-JTSK, MGI, DHDN, OSGB, …). WGS84-aligned/unknown-datum defs are unchanged. Exported + unit-tested (Ferro Krovak 2065, grid-strip 5514, existing-towgs84 passthrough, WGS84 passthrough).

louistrue:
Addressed the CodeRabbit nitpick in `61601a68`: the adapter now reuses the band-guard `probe` for image 0 instead of re-fetching it.

Issue #1276, opened 2026-06-21
Feature: step through each object in a clash pair with a clear description


BIMvoice:
Building on the isolation idea, it would be wonderful to click each object in a pair separately, for example the IFC slab and the member, and read a clear description of what overlaps and why it was flagged. That would let people assess each clash with confidence instead of guessing. Perhaps the clash row could expand to show both elements. Thank you for considering it.

louistrue:
Clash rows now **expand to show each object separately**, each with a plain-language description (e.g. *Hard clash — 3.500 m interpenetration*) and a per-element select so you can step through both sides. Shipped in #1285 (merged to main as 593f02b4).

PR #1377, opened 2026-06-26
fix(viewer): colour the two clashing elements distinctly (#1277, #1339)

### LINKED ISSUE #1277: Feature: color each object and the overlap differently
A small visual touch that would make a real difference: in the isolated view, give each clashing object its own color and the overlap region a third color. You would instantly see where and how much they intersect. With everything else you have already built, this would make clashes obvious at a glance.

### LINKED ISSUE #1339: Very difficult to understand the clashes
Is there a specific way you look at clashes? I really struggle to confirm them visually.

Also the both objects clashing have the same color, which makes it very difficult to see what is going on.

IFClite:

Image

Bonsai BIM:

Image

Solibri:

Image

Image

Image

BIMcollab Zoom:

Image

BIMvision:

Image

Image

louistrue:
## Problem

Selecting a clash highlighted **both** clashing elements with the single selection colour, so you couldn't tell which was which — the #1339 complaint ("both objects clashing have the same color, which makes it very difficult to see what is going on"). #1277 asks to colour each object (and ideally the overlap) differently.

## Fix

`focusClash` now paints the pair in two distinct fill colours through the renderer colour-override channel (the same `pendingColorUpdates` → `scene.setColorOverrides` path the lens uses):

- **Element A → warm orange** `[0.96, 0.45, 0.13]`
- **Element B → cool blue** `[0.16, 0.5, 0.92]`

The normal selection **outline is still drawn on top** of both, so you keep the "this is the focused pair" cue plus a clear "which is which". Orange-vs-blue stays distinguishable under red/green colour-vision deficiency.

- `selectElement` (single-side step-through, #1276) paints the one element the A colour.
- Colours **reset** on `clearHighlight`, `highlightAll` (ambiguous there — an element can be A in one clash and B in another), and exit.

## Scope

This delivers the core of #1277/#1339 (A vs B distinct). The optional third item — drawing the **overlap region** as a third colour — needs a new wireframe-box overlay primitive in the renderer (the `Clash.bounds` AABB exists), which is a larger, riskier change; left as a follow-up.

## Notes

- Clash focus co-opts the shared colour-override channel (same one the lens writes), so focusing a clash while a lens is active replaces the lens colours for the pair. This is consistent with `focusClash` already taking over isolation/ghost on focus.

## Tests / verify

- Pure `buildClashPairColors` helper + unit tests: distinct colours, null-ref skip, degenerate self-clash, valid RGBA — 4/4. `turbo typecheck --filter=@ifc-lite/viewer` clean.
- In-viewer: load a model, run clash detection, select a result → the two elements render orange vs blue (previously both the same colour); Clear resets them.

(`apps/viewer` is a private app — no changeset.)

Fixes #1277
Fixes #1339

## Summary by CodeRabbit

* **New Features**
  * Clash views now highlight involved elements with distinct colors and show the clash overlap region as a dedicated wireframe box.
  * The clash panel now handles large result sets more smoothly with virtualized scrolling.
* **Bug Fixes**
  * Closing or clearing a clash no longer leaves stale highlight colors or overlap outlines behind.
  * Focus and frame behavior now better follows clash-specific selections when no normal selection is active.
* **Tests**
  * Added coverage for clash color assignment and edge-line rendering behavior.

### DIFF
--- a/.changeset/feat-clash-highlight-colors.md
+++ b/.changeset/feat-clash-highlight-colors.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/renderer": minor
+---
+
+Add `RenderOptions.highlightColors` — a per-element highlight tint for selected meshes. A selected mesh whose id is in this map glows in the given RGBA using the same re-lit selection treatment instead of the default selection blue. Lets a consumer render, e.g., the two sides of a clash in distinct vibrant colours while keeping the selection glow, crease structure and opacity. Only applied to ids that are also selected (`selectedId`/`selectedIds`); absent map = unchanged behaviour. (#1277/#1339)
--- a/.changeset/feat-clash-highlight-colors.md
+++ b/.changeset/feat-clash-highlight-colors.md
@@ -2,4 +2,9 @@
 "@ifc-lite/renderer": minor
 ---
 
-Add `RenderOptions.highlightColors` — a per-element highlight tint for selected meshes. A selected mesh whose id is in this map glows in the given RGBA using the same re-lit selection treatment instead of the default selection blue. Lets a consumer render, e.g., the two sides of a clash in distinct vibrant colours while keeping the selection glow, crease structure and opacity. Only applied to ids that are also selected (`selectedId`/`selectedIds`); absent map = unchanged behaviour. (#1277/#1339)
+Clash visualisation support:
+
+- Add `RenderOptions.highlightColors` — a per-element highlight tint. An id in this map glows in the given RGBA using the re-lit selection treatment (and gets the same individual-mesh / opaque / stay-solid-through-ghost handling) **without** needing to be in the store selection. Lets a consumer render, e.g., the two sides of a clash in distinct vibrant colours in highlight/isolate/ghost with no "selected" state.
+- Add `Renderer.setClashOverlapBox(box | null)` — draws a world-space AABB as a distinct-colour wireframe box (e.g. the clash overlap region) via the existing overlay line pipeline. Pass `null` to clear.
+
+(#1277/#1339)
--- a/.changeset/feat-clash-highlight-colors.md
+++ b/.changeset/feat-clash-highlight-colors.md
@@ -2,9 +2,4 @@
 "@ifc-lite/renderer": minor
 ---
 
-Clash visualisation support:
-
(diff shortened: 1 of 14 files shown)


louistrue:
Reworked per your feedback (commits `983050bf` + `39010058`):

**Behaviour:** focusing a clash no longer selects the pair. The two elements just glow in distinct vibrant colours (amber A / cyan B) via a clash-owned highlight channel. The renderer gives highlighted ids the same treatment as a selection — individual-mesh hydration, opaque, stay-solid-through-ghost, and the re-lit glow — **without** the store selection, so the colours show in **highlight, isolate AND ghost** with no `2 SEL` state. Camera framing still encloses the pair (frameSelection also frames the highlight ids).

**Re the two earlier review threads** (clear-on-teardown / don't-blank-shared-channel): both are resolved by this rework — clash colours no longer touch the shared `pendingColorUpdates`/colour-override channel at all (so lens/IDS overlays are untouched), and the glow is intrinsically scoped to the clash highlight set + cleared on every teardown (clearClash, clearHighlight, highlightAll) and on panel unmount.

louistrue:
Added the **overlap region** visualization (commits `0065e1aa` + `4bde0e68`) — this completes #1277's 'color each object AND the overlap differently':

- The focused clash's overlap (`Clash.bounds` AABB) now renders as a **vibrant magenta wireframe box** — a distinct third colour marking exactly where the two elements intersect, on top of the amber (A) / cyan (B) element glow.
- Implemented via the existing world-space overlay line pipeline (same one grids/annotations use) with its own colour; `clash.bounds` is in the world frame (the STEP adapter folds per-element origin), so it aligns with the rendered geometry.
- Cleared on every teardown + panel unmount.

So #1277 + #1339 are now fully covered: distinct vibrant element colours (no selection state, works in highlight/isolate/ghost) **plus** the overlap box. Renderer 166/166, typecheck clean.

louistrue:
Addressed both (`01263bee`):
- **Reset focus view on unmount** — the ClashPanel unmount cleanup now also `clearEntitySelection` / `clearIsolation` / `clearGhost` + clears `clashSelectedId`, so a clash focused in isolate/ghost no longer leaves the model isolated/ghosted after the panel closes.
- **A11y** — added `aria-expanded` + `aria-label` to the virtualized group-header and clash expand/collapse buttons.

Also: the clash colours now **pop** (`12a1cff8`) — while a clash is focused the override pass renders the colour near full-bright (re-lit, high floor) so amber/cyan stand out like a highlight instead of reading as a normal lit material.

Issue #654, opened 2026-05-12
Model Parsing and Viewing Abnormalities


xyzbety:
Bug:
This model was created with CATIA. It displays normally with correct colors in BIMVision. However, it loses material and color rendering in webifc, while IFCLite Viewer can only show property information without any geometric display. Could you please help check this issue? Thank you very much.

Test Model:
[CatiaBuilding.zip](https://github.com/user-attachments/files/27645420/CatiaBuilding.zip)

Viewing effect with IFCLite Viewer:

<img width="2868" height="1602" alt="Image" src="https://github.com/user-attachments/assets/d19cf860-edcd-4a98-b2fb-abdf0be90ddb" />

viewing effect with webifc (via https://ifc.babylonpress.org/):

<img width="2868" height="1602" alt="Image" src="https://github.com/user-attachments/assets/5e2e6ddf-c222-4633-a892-6ee47619f084" />

viewing effect with BIMVision:

<img width="2877" height="1723" alt="Image" src="https://github.com/user-attachments/assets/54998784-a2a0-4cc3-b247-17a6109b2335" />

xyzbety:
@louistrue Could you assist in verifying the model’s color rendering once more? Some components are still not rendered as expected.
<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/5116e8c4-bacc-422d-a627-0d8b49f78c64" />

xyzbety:
Additionally, the front-end parsing logic may need to be updated accordingly to keep aligned with the back-end processing library ifc-lite-processing. @louistrue

louistrue:
Thanks for the follow-up @xyzbety — confirmed the geometry renders after the parser fix (#657). The remaining color-rendering gap and the front-end/`ifc-lite-processing` alignment are separate concerns from the entity-scanner bug we fixed here; they involve the IfcStyledItem / IfcIndexedColourMap resolution path on the styling side rather than the parser, and aligning the TS-side parser with the Rust `ifc-lite-processing` crate is its own work item.

Could you open a new issue for the color regression (with the screenshot above and ideally a short list of which components specifically render wrong) so it can be triaged independently? That'll let it move on its own track rather than piggy-backing on the closed parser ticket.

xyzbety:
Thank you. I've found a more typical model, which is now available at https://github.com/louistrue/ifc-lite/issues/663 @louistrue

PR #1383, opened 2026-06-26
refactor(core): split parser.rs into tokenizer + scanner submodules

louistrue:
## What

`rust/core/src/parser.rs` (985 LOC) bundled two genuinely independent algorithms that share no private state:

- a **nom-combinator tokenizer** (`Token` enum + the `entity_ref..list_at_depth` parsers + `parse_entity`), and
- a **byte-level SIMD fast scanner** (`EntityScanner` + the quote-aware `data_section_start` HEADER-skip helper) that does its own hand-rolled parsing and never touches `Token`/nom.

This carves them into `parser/{tokenizer.rs, scanner.rs}` behind a thin `parser/mod.rs`.

## API stability (pure code motion)

`parser/mod.rs` re-exports the entire public surface, so **every consumer path stays byte-identical**:

```rust
pub use scanner::EntityScanner;
pub use tokenizer::{parse_entity, Token};
```

- Internal `crate::parser::{EntityScanner, Token, parse_entity}` imports (`streaming.rs`, `decoder.rs`, `schema_gen.rs`) resolve unchanged.
- The crate-root re-export `lib.rs: pub use parser::{parse_entity, EntityScanner, Token};` (used by every downstream crate as `ifc_lite_core::EntityScanner` etc.) resolves unchanged.

No logic changes. All 14 parser unit tests travel with their code (9 with the tokenizer, 5 with the scanner). `data_section_start` and `MAX_NESTING_DEPTH` stay private to their respective submodules.

## Verification

- `cargo test -p ifc-lite-core` green (82 lib tests, incl. all 14 parser tests)
- `cargo build --workspace` green (all downstream crates compile against the unchanged API)
- `cargo clippy -p ifc-lite-core --all-targets -- -D warnings` clean

First of a series of behavior-preserving file splits (files >450 LOC); intentionally small and standalone to validate the re-export pattern through CI.

## Summary by CodeRabbit

* **New Features**
  * Added support for STEP/IFC parsing with a faster entity scanner.
  * Improved handling of quoted text, comments, nested lists, typed values, and numeric formats.
  * Added lookups for entities by type and counts by type.
* **Bug Fixes**
  * Improved parsing of tricky header content, including `#` symbols and `DATA;` text inside strings.
  * Added safer handling for deeply nested content to avoid parsing failures.


louistrue:
Thanks for the review. Responses to the three findings:

**Module size (`scanner.rs` 535, `tokenizer.rs` 463).** This PR is a pure behavior-preserving split of the 985-LOC `parser.rs` along its one clean seam: the nom-combinator tokenizer vs the byte-level `EntityScanner`. I'm keeping it at those two files deliberately:
- `scanner.rs` is a single `EntityScanner` type. Its body is ~420 LOC even with the inline tests removed, dominated by the one comment/quote-aware `next_entity` core; the methods share `self.bytes`/`self.position` and the quote-parity invariants (#654). Splitting one struct's tightly-coupled scanner methods across files would fragment a single cohesive unit for a line count, not improve it.
- `tokenizer.rs` is one nom combinator module (`Token` + the mutually-recursive `token_at_depth`/`list_at_depth`/`typed_value_at_depth` + `parse_entity`); ~190 of its 463 lines are the inline `#[cfg(test)]` module (repo convention).

These are the two natural algorithm boundaries; this is the "~400" guideline's "split except where it makes no sense" case. Happy to fragment further if you'd prefer it.

**Nesting limit in empty containers / trailing bytes after `;`.** Both describe pre-existing behavior of `parser.rs` that this PR moves verbatim - they are not introduced or changed here. To keep the refactor behavior-preserving (byte-identical parse results) I'm not changing parser semantics in this PR; if either is worth tightening I'll track it as a separate behavior change with its own test.

PR #1398, opened 2026-06-26
refactor(geometry): split kernel/arrangement.rs into mod + classify + boolean + tests

louistrue:
Behavior-preserving split of `rust/geometry/src/kernel/arrangement.rs` (1489 LOC), the CSG determinism firewall.

`arrangement/` = `mod.rs` (383: types + `arrange`/`arrange_many`), `classify.rs` (641: point-inside + surface-detection + the 3-regime `boolean_vids`), `boolean.rs` (221: public drivers + `boolean_manifest`), `tests.rs` (277).

### Determinism
The boolean output is pinned by a cross-platform manifest hash (`boolean_manifest() == 0x0465b83a5fdb8b2b`, native==wasm bit-identical). The split is **byte-exact code motion** (every moved body diffed identical vs HEAD), so **`boolean_manifest_is_pinned` passes — the hash is unchanged**.

`classify.rs` (641) is kept whole as the cohesive 3-regime classification algorithm: `boolean_vids_components` weaves the surface-normal / solid-side / point-inside regimes through a shared `ray_dir` and interlocking determinism invariants; splitting it would fragment one determinism-pinned algorithm.

### Wiring
~11 helpers raised to `pub(super)` for cross-submodule reach; the public API stays `pub` and is re-exported from `mod.rs` so every `crate::kernel::arrangement::*` path is identical. `kernel/mod.rs` untouched. `super::X` kernel-level refs rewritten to `super::super::X`.

### Verification
- `cargo build --workspace` + `cargo clippy -p ifc-lite-geometry --all-targets -- -D warnings` clean
- **Full `cargo test -p ifc-lite-geometry` green (401 lib + all integration + annex_e snapshots), 0 failed; `boolean_manifest_is_pinned` passes**
- Faithfulness check: zero function-body lines dropped.


louistrue:
On `classify.rs` (641 LOC): this is deliberately kept as one file. It is the single 3-regime point-classification algorithm - `boolean_vids_components` weaves regime-1 surface-normal, regime-2 solid-side/on-interface-keep, and regime-3 point-inside ray-casting through a **shared `ray_dir`** and interlocking determinism invariants (the pinned cross-platform `boolean_manifest 0x0465b83a5fdb8b2b`, native==wasm bit-identical). Splitting these regimes across files would fragment one determinism-pinned algorithm and risk the very evaluation-order/aliasing properties the manifest pins, for a line count. The cleanly-separable pieces (arrange/types, the public boolean drivers, the tests) were the carve-outs; this is the irreducible core. Flagged rather than forced.

PR #1399, opened 2026-06-26
refactor(geometry): split router/voids.rs (5145 LOC) into 6 modules

louistrue:
Behavior-preserving split of `rust/geometry/src/router/voids.rs` (5145 LOC) - the single largest source file in the tree and the void/opening CSG subtraction path.

`router/voids/` = `mod.rs` (1513: the void orchestrator state machine + private types), `synthesis.rs` (1002: classify/merge/extend openings + cutter-mesh synthesis), `geom.rs` (881: GeometryRouter-free predicates + frame math), `probe.rs` (729: IFC parametric decode + cutter-mesh extraction), `aabb_clip.rs` (618: `ClipBuffers` + the legacy SAT/Sutherland-Hodgman #635 fallback), `reveal_tests.rs` (574).

### Precision / determinism
This path is dense with load-bearing invariants (#1167/#1007/#635/#2176, FMA-free f64, f32-quantum epsilons). The split is **byte-exact code motion** — every moved block diffed identical vs HEAD, **zero literals/epsilons/comments/evaluation-order changed**. The flagged >400 files (orchestrator, synthesis, geom, probe, aabb_clip) are each one cohesive precision-critical algorithm, deliberately not reflowed or fragmented.

### Wiring
`router/mod.rs` (`mod voids; pub use voids::RectParam;`) and `router/tests.rs` are **untouched**. `cut_rectangular_opening` -> `pub(in crate::router)` (router/tests.rs calls it); the ~30 geom predicates + synthesis entry fns -> `pub(super)`; one function-local `use super::` -> `use super::super::` for the deeper nesting.

### Verification
- `cargo build --workspace` + `cargo clippy -p ifc-lite-geometry --all-targets -- -D warnings` clean
- **Full `cargo test -p ifc-lite-geometry` green: 401 lib + all integration + annex_e snapshots (byte-identical) + the void/opening regression suites (wall_opening_cut_regression, issue_635 gable/round-window, reveal_tests), 0 failed**
- Faithfulness check: zero function-body lines dropped (only visibility bumps, redistributed imports, `super::`->`super::super::` rewrites, and 3 section-divider comments superseded by the file split).


louistrue:
On the voids submodule sizes: this PR takes the single largest file in the tree (5145 LOC) down to a max of 1513, and the files left over ~400 are each one cohesive precision-critical algorithm that the analysis explicitly says must not be reflowed - the void orchestrator state machine (mod.rs), the classify/merge/cutter-synthesis pass (synthesis.rs), the GeometryRouter-free predicate+frame-math library (geom.rs), the IFC parametric probe (probe.rs), and the legacy SAT/Sutherland-Hodgman #635 fallback (aabb_clip.rs). They carry interlocking #1167/#1007/#635/#2176 invariants + f32-quantum epsilons; fragmenting them across more files to hit a line count would risk the exact evaluation-order/aliasing properties the void-cut regression suite pins, for no encapsulation gain. The cleanly-separable concern (the inline tests) was split out (reveal_tests.rs). This is the '~400 except where it makes no sense' case; flagged, not forced.

Issue #1405, opened 2026-06-26
Snap misses all-but-one occurrence of GPU-instanced geometry (bolts)


Blogbotana:
# Snap misses all-but-one occurrence of GPU-instanced geometry (bolts)

## Summary
When measuring on a group of identical, GPU-instanced elements (e.g. a row of bolts),
vertex/edge snap hints appear on **only one** occurrence. Free-point (raycast) measuring
and selection work on all of them; only the *snap targets* (edges/vertices) are missing on
the others. This makes part-to-part and part-edge distance measuring on repeated hardware
unreliable.

## Affected area
- `@ifc-lite/renderer` — `SnapDetector` (`snap-detector.ts`), `Scene.getInstancedMeshDataPieces`
  (`scene.ts`), `RaycastEngine.collectVisibleMeshData` (`raycast-engine.ts`).
- Reproduced via the external Svelte viewer pinned to `@ifc-lite/renderer@1.29.2`.

## What is NOT the cause (correcting the original triage)
The instanced raycast/snap capability is **already present** in the pinned version.
On-demand materialization of GPU-instanced occurrences into the CPU raycast + measure-snap
path (`getInstancedMeshDataPieces`, ray-AABB pre-cull, wired into `RaycastEngine`) landed in
**renderer 1.28.5** (#1238), which predates the pinned **1.29.2**. So this is not a
"renderer cannot snap instances at all / wait for upstream" situation — the path exists and
raycast already traverses instanced occurrences.

## Suspected root cause (to confirm)
`SnapDetector` caches processed snap geometry (deduped vertices, real edges, vertex valence)
keyed **only by `mesh.expressId`**:

```ts
// snap-detector.ts
private geometryCache = new Map<number, MeshGeometryCache>();
private getGeometryCache(mesh: MeshData): MeshGeometryCache {
  const cached = this.geometryCache.get(mesh.expressId);
  if (cached) return cached;
  const cache = buildGeometryCache(mesh);
  this.geometryCache.set(mesh.expressId, cache);
  return cache;
}
```

But `Scene.getInstancedMeshDataPieces(eid)` returns **one `MeshData` per occurrence, all
stamped with the same `expressId`**, with the per-occurrence instance matrix already baked
into world-space `positions`:

```ts
// scene.ts
for (const o of occ) {
  // ...apply o's 4x3 matrix to template positions -> world-space positions...
  out.push({ expressId, positions, normals, indices: tpl.indices, color });
}
```

[...]

PR #1419, opened 2026-06-26
fix(renderer): measure-snap on mapped/multi-piece elements (pick dedup + fov radius + cache)

Blogbotana:
Three renderer measure-snap fixes, found while debugging "snap catches only one bolt of a
group" on a real model (`test.ifc`: an `IfcMechanicalFastener` "Bolt assembly" = 4
`IfcMappedItem` copies of one `RepresentationMap` at distinct placements). All three were
needed; the last is what made every bolt snap.

## 1. Mapped copies dropped from the pick/snap mesh set (the unblocker)
`collectVisibleMeshData` deduped flat pieces on a **size-based** key
(`expressId` + `modelIndex` + buffer sizes). A mapped entity expands to several flat pieces with
**identical template geometry** at different placements, so the key collided and all but the
first were dropped from the raycast set → the other bolts returned **no ray hit at all**, so
neither pick nor snap could reach them. The key now also includes the per-piece `origin` + first
vertex, so distinct placements survive while a piece reached from both the regular and batched
passes still dedups. (Mirrors the instanced-piece key fix from #1238 for the flat path.)

## 2. Snap radius ~57× too small
`screenToWorldRadius` applied a degrees→radians conversion to `fov`, but its only caller passes
`Camera.getFOV()` (already radians) → `tan(fov·π/180 / 2)` instead of `tan(fov/2)`. Vertex/edge
snap then needed sub-millimetre cursor precision and fell back to face hits on small features.
Treat `fov` as radians.

## 3. Multi-piece flat meshes shared one snap-geometry cache entry
The snap cache keyed flat meshes on `expressId` alone (instanced occurrences already key on
`occurrenceKey`, #1405), assuming one flat mesh per `expressId`. Mesh fragmentation emits one
entity as several flat pieces, so the first piece's vertices/edges were served for all. The cache
now keys flat pieces on a content signature (`expressId` + `origin` + buffer sizes + sampled
vertices).

## Tests
`snap-geometry-utils.test.ts` (radians contract) and `snap-detector-instanced.test.ts` (multiple
flat sub-pieces; mapped copies differing only by `origin`) — both flat-piece cases verified to
fail with the old keys and pass with the fix. Full renderer suite green (173 passed); typecheck
clean. The mesh-collection dedup fix is verified end-to-end (all bolts of a fastener now snap);
it lives in a private collection path that needs a full Scene+canvas, so it isn't covered by a
node unit test.

Extends #1405/#1409 (instanced occurrences) to the non-instanced mapped/multi-piece flat case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapping and picking accuracy for small/flat mesh pieces by preventing incorrect cache reuse across distinct sub-piece placements.
  * Fixed cases where mapped/copied geometry with the same identifier could be incorrectly deduplicated, ensuring snaps resolve to the correct piece.
  * Corrected field-of-view handling so snap radius calculations treat FOV as radians, improving responsiveness.

* **Tests**
  * Added/updated regression tests for flat sub-pieces and mapped copies to cover cache correctness, including FOV-as-radians behavior.

### DIFF
--- a/.changeset/snap-flat-multipiece-cache.md
+++ b/.changeset/snap-flat-multipiece-cache.md
@@ -18,3 +18,12 @@ degrees→radians conversion to `fov`, but its only caller passes `Camera.getFOV
 already in radians. The shrunken radius made vertex/edge snap require sub-millimetre cursor
 precision and fall back to a face hit on small features (e.g. bolts). The conversion is
 removed; `fov` is treated as radians.
+
+Also fix the CPU pick/snap mesh collection dropping mapped copies. `collectVisibleMeshData`
+deduped flat pieces on a size-based key (`expressId` + `modelIndex` + buffer sizes), so the
+several flat pieces a mapped entity expands to — identical template geometry at different
+placements (e.g. the 4 bolts of one `IfcMechanicalFastener`) — collided and all but the first
+were dropped from the raycast set. The hidden bolts then returned no ray hit at all, so neither
+pick nor snap could reach them. The key now also includes the per-piece `origin` + first vertex,
+so distinct placements survive while a truly identical piece reached from both the regular and
+batched passes still dedups. (Mirrors the instanced-piece key fix from #1238 for the flat path.)
--- a/.changeset/snap-flat-multipiece-cache.md
+++ b/.changeset/snap-flat-multipiece-cache.md
@@ -4,7 +4,7 @@
 
 Fix measure-snap missing all-but-one piece of a multi-piece flat mesh. The snap geometry cache
 keyed flat meshes on `expressId` alone (instanced occurrences already keyed on `occurrenceKey`,
-#1405), assuming one flat mesh per `expressId`. But mesh fragmentation routinely emits one
+`#1405`), assuming one flat mesh per `expressId`. But mesh fragmentation routinely emits one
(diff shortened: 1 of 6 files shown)


Blogbotana:
Re: the magnetic edge-lock keying on `meshExpressId` only (the "carry piece identity into edge locks" finding) — agreed, it's the same `expressId`-keyed assumption family this PR addresses for the geometry cache and pick-set dedup, so it's a valid refinement.

However it's **out of scope for this PR**: it's pre-existing edge-lock code (outside this PR's diff) and fixing it properly is a cross-cutting change touching the renderer API (`EdgeLockInput` / `MagneticSnapResult` / `detectMagneticSnap` / `maintainEdgeLock`) **and** the consumer's edge-lock state plumbing.

Practical impact is also limited now: with the fov radius fix in this PR, the lock escape threshold is ~3–4 cm, so moving between well-separated mapped copies (e.g. bolts ~10 cm apart) releases and re-locks correctly — only copies closer than the escape distance would stick.

Better tracked as a separate follow-up issue. We're not taking it on right now; keeping this PR focused on the cache/radius/pick-dedup fixes (which are verified end-to-end).

louistrue (approved):

PR #1426, opened 2026-06-27
spike(wasm): measured wide-arith win (~1.7x CSG, blocked upstream) + pkg-wide bundle

louistrue:
> **Status: tracking spike — NOT shippable, no user-facing speedup today.** Draft until the two upstream blockers below clear.

## What

Measures the WASM **wide-arithmetic** proposal on ifc-lite's actual exact-CSG kernel and documents how to ship it *once it's possible*. This is the only no-install lever that attacks the root cause of the in-browser CSG/brep slowdown (no wide-integer hardware on wasm32 -> bnum's 64x64->128 limb products become `__multi3` libcalls).

**This PR ships docs + build tooling only** — the `BUILD_WIDE=1` bundle flag and the architecture doc. No published-package src, no runtime wiring, no changeset. Merging it gives users nothing; it exists to record the measurement and the delivery plan.

## Does this reach users? No — blocked on two upstream items (verified 2026-06-27)

1. **wasm-bindgen cannot parse a wide-arith module.** Production wasm goes through `wasm-bindgen =0.2.106` for the `IfcAPI` glue; building `pkg-wide` fails in the bindgen step (`failed to parse code section: wide arithmetic support is not enabled` — its `walrus` parser rejects the new opcodes). The benches sidestep this with a plain `cdylib` (no bindgen), which is why they build and run while the real bundle can't.
2. **No stable browser engine supports it.** Node 22 / V8 returns `WebAssembly.validate() -> false` ("invalid numeric opcode 0xfc13"); stable Chrome/Safari are the same. The proposal is Phase 3 (not a finished standard); only Wasmtime/Wasmer run it in stable. Realistic shipping: per-engine through 2026-2027, "Baseline" (safe to rely on everywhere) later.

The runtime gate (40-byte `WebAssembly.validate()` probe -> choose `pkg-wide` vs `pkg` URL) is a **zero-cost no-op** until both clear: the probe returns false on every engine today, so the wide `.wasm` is never fetched. It auto-upgrades per engine with zero regression.

## Measured (Apple M4, rustc 1.93 / LLVM 21, wasmtime 46)

`-C target-feature=+wide-arithmetic` makes LLVM emit `i64.mul_wide`/`add128`/`sub128` from our real `bnum` I256/I512 predicate math; output is byte-identical to baseline.

- **Predicate (orient3d det3):** I256 438->232 ns (1.9x), I512 1093->358 ns (3.1x) — to ~native parity.
- **End-to-end real void cut** (`subtract_many`, slab minus 9 boxes): **23.16 -> 13.58 ms/cut = 1.71x**; wasm goes from 2.26x slower-than-native to **1.33x**. Note this is one real cut, not full-model load time, and only on exact-CSG/brep-heavy work.

Native already beats web-ifc on the Tekla model (2.9s vs 4.9s), so closing the wasm gap this far makes the in-browser path competitive on exactly the models that look slow today.

## In this PR (docs + build tooling only)

- `docs/architecture/wasm-wide-arithmetic.md`: numbers, mechanism, upstream blocker status, and the delivery plan.
- `scripts/build-wasm.sh`: `BUILD_WIDE=1` builds a `packages/wasm/pkg-wide` bundle (mirrors `pkg-threaded`; refuses to ship if the wide ops aren't present).

## Delivery plan (follow-up, once BOTH blockers clear)

Reuses the threaded-bundle pattern exactly: build `pkg-wide` -> feature-detect at runtime with the `validate()` probe -> pass the `pkg-wide` URL through `geometry.worker.ts`'s existing `wasmUrl` plumbing. Engines without the proposal load `pkg` and are unaffected, so it can ship eagerly once buildable.

For reference, the bigger levers that work in **today's** browsers: threads (`pkg-threaded`, measured 1.6-1.9x end-to-end, built in #1255, not yet wired into runtime selection) and the tessellation tier (medium->low ~7.6x on heavy steel, no rebuild). wide-arith is additive on top of those later.

Repro crates live in the profiling repo under `wide-arith/`.

### DIFF
--- a/docs/architecture/wasm-wide-arithmetic.md
+++ b/docs/architecture/wasm-wide-arithmetic.md
@@ -91,14 +91,15 @@ only safe handling is the feature-detect + fallback above (zero regression,
 auto-upgrade per engine as each ships). For users today it yields no in-browser
 speedup.
 
-The bigger lever that *does* work in today's browsers is **threads**: the
-`pkg-threaded` bundle (rayon-in-wasm, atomics + shared memory; SharedArrayBuffer
-is widely supported and the viewer already sets COOP/COEP). Native shows
-threading is a far larger win than wide-arith on the heavy models — Tekla 1-core
-14.2s vs 10-core 2.9s (~4.8x) vs wide-arith's 1.7x. In-wasm scaling still needs
-the `rust/csg-thread-bench` (rung-2) browser result to confirm it tracks native,
-but it is the near-term in-browser path; wide-arith is additive on top later
-(a threaded+wide bundle combines both target-features).
+The bigger lever that *does* work in today's browsers is **threads** (in-instance
+rayon), and it is **already measured** — see `docs/architecture/csg-threading-design.md`
++ the `rust/csg-thread-bench` rung-2 result: threaded WASM scales the CSG step
+**2.9-4.2x at 8 threads** (atomics tax ~0%, output byte-identical) and **1.6-1.9x
+end-to-end** on the current exact-CSG kernel. The `pkg-threaded` bundle is built
+(#1255) but not yet wired into runtime selection. It works in today's
+Chrome/Firefox (cross-origin isolated; the viewer already sets COOP/COEP; Safari
+lacks credentialless COI and falls back to the plain bundle). wide-arith is
+additive on top later (a threaded+wide bundle combines both target-features).
 
 **Working around the wasm-bindgen blocker:** split the CSG kernel into a separate
 plain-`cdylib` wasm (no wasm-bindgen) built with `+wide-arithmetic`, exposing the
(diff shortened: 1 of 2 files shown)


Blogbotana:
Complementary data point from profiling a real ~20 MB steel-detailing model (~16.6k elements, dense end-clips + bolt holes — exactly the near-coplanar pattern that pushes the predicate cascade into the exact tiers):

The wide-arith win on the predicate/boolean stage stacks with a second, **no-rebuild** lever we measured — the tessellation tier. On this model, dropping detail `medium → low` cut WASM geometry from **~100 s → ~13 s (≈7.6×)**, and `lowest` → **~11 s (≈9×)**; triangles fall 7.9 M → ~2–3 M. Natively the same tier goes 6.5 s → 1.0 s.

Most of that isn't just fewer triangles — the preview tiers' small-cut skip drops the tiny boolean cuts (chamfers, bolt holes) that dominate the exact-tier escalation count on detailed steel, i.e. it removes work `+wide-arithmetic` would otherwise have to do fast. The two are orthogonal: wide-arith makes each remaining exact predicate ~1.7–3× cheaper; the tier skips a large fraction of them outright. So for heavy models in the browser *today*, defaulting/auto-selecting `low` is the biggest immediate lever, and wide-arith then closes the residual native gap.

The per-boolean budget (#1109) barely engaged here (unbounded 7.1 s vs default 6.5 s native) — this model is "honestly heavy," not pathological, so the cost is broad exact work across many healthy elements, which is exactly what both levers target.

PR #1429, opened 2026-06-27
feat(geometry): opt-in threaded-WASM geometry path (in-instance rayon)

louistrue:
## What

Wires an **opt-in** threaded-WASM geometry path into the viewer: one shared-memory worker whose internal `par_chunks` element loop fans CSG across an in-instance rayon pool, instead of the N single-thread worker instances. Builds on the `wasm-features` detection primitive (already on this branch) with the full cross-package wiring, validated in real Chrome.

## Changes

- **`wasm-bindings` (the missing kernel piece):** `gpu_meshes/batch.rs::produce_batch` (what `processGeometryBatch` calls) was a serial `for` even with `--features threads` (that feature only exposed `init_thread_pool`). Added a `#[cfg(feature = "threads")]` `par_chunks` variant: fresh per-job decoder + router, mirroring the native `process_entity_job` fan-out but keeping the viewer's `EmitTagged` type geometry. The shipped serial path is untouched under `#[cfg(not(feature = "threads"))]` and is byte-identical. Both feature configs `cargo check` clean.
- **`@ifc-lite/wasm`:** `./threaded` subpath export for the `pkg-threaded` bundle (built off by default via `BUILD_THREADED=1 ./scripts/build-wasm.sh`).
- **`geometry.worker.ts`:** keeps its static single-thread import for the plain path; additively dynamic-imports `@ifc-lite/wasm/threaded` and calls `initThreadPool` once (guarded against recovery re-init) when `init` carries `useThreaded`.
- **`geometry-parallel.ts`:** runs ONE worker on the threaded path; selected only when `globalThis.__IFCLITE_THREADED_WASM__ === true` AND `isThreadedWasmUsable()`.
- **viewer Vite:** alias `@ifc-lite/wasm/threaded` to `pkg-threaded` + `optimizeDeps` exclude. The dynamic import pulls the threaded bundle and its nested rayon worker into the build graph (no manual copy).
- **`wasm-threaded.d.ts`:** ambient types so `tsc` resolves the subpath without the off-by-default bundle built (verified green with and without `pkg-threaded` present).

## Browser A/B (real Chrome, viewer dev, 10-core, cold, cache cleared each run)

Output parity is **exact** in every cell (plain@1-worker == threaded@1-worker byte-for-byte, e.g. 11942 meshes / 2187k verts / 902.4K tris), so `par_chunks` produces identical geometry to the serial loop.

| model | plain@8 (default) | threaded@1 (rayon) | plain@1 (serial) |
|---|---|---|---|
| ISSUE_129 (11.5 MB, CSG-bound) | 15.3 / 15.1 s | **11.1 / 10.9 s** | 27.1 s |
| ISSUE_068 (53.7 MB, decode-bound) | 14.1 / 16.2 s | 15.3 / 15.4 s | **9.9 s** |

The win is **per-element-work-size dependent, not file-size dependent**: threaded beats the default pool ~1.38x (and a single serial worker ~2.4x) on the CSG-bound model, but only ties the default pool on the large decode-bound model, where per-element work is small and two costs dominate: a **fresh per-job decoder** (loses the serial path's warm cross-job cache) and ~10 rayon threads contending on dlmalloc's **single global lock**. Versus the production default (plain@8) it is a win-or-tie, never a regression.

## Why opt-in (default off)

The lone threaded worker has no fallback if it dies, and the win is not yet universal. Before flipping default-on: (a) per-rayon-thread warm decoder (reuse one decoder per worker thread instead of fresh-per-job) to move the decode-bound case from tie toward win; (b) threaded-failure to plain-pool fallback; (c) production PostHog per-model confirmation (the regression dashboard + alerts already exist). Numbers above are dev-build and noisy; a production-build A/B should firm them up.

## Test plan

- `cargo check -p ifc-lite-wasm` (default) and `--features threads` (with atomics + build-std): both clean.
- `BUILD_THREADED=1 ./scripts/build-wasm.sh`: both bundles build; threaded imports shared memory + exports `initThreadPool`.
- Geometry package `tsc --noEmit`: clean with AND without `pkg-threaded` present.
- 108 geometry unit tests pass.
- Real Chrome: threaded path boots (`initThreadPool`, cross-origin isolated), loads both models correctly (visual + mesh/vertex parity).

See `docs/architecture/csg-threading-design.md` for the full evidence and rung-1/2 background.

## Summary by CodeRabbit

* **New Features**
  * Added runtime probing to enable an opt-in threaded WASM geometry path when supported (requires cross-origin isolation).
  * When enabled, geometry initialization spins up a thread pool and routes parallel work across cores.
  * Added a dedicated `./threaded` export so threaded WASM can be loaded from the viewer.
* **Tests**
  * Added tests covering threaded WASM detection and fallback to the single-threaded bundle.
* **Documentation**
  * Updated the threaded WASM CSG wiring architecture status and rollout notes.

### DIFF
--- a/.changeset/threaded-wasm-feature-detect.md
+++ b/.changeset/threaded-wasm-feature-detect.md
@@ -1,0 +1,10 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Add `wasm-features` runtime detection (`supportsWasmThreads` / `isThreadedWasmUsable`)
+— the gate for selecting the threaded WASM geometry bundle (`pkg-threaded`) when
+the engine supports the threads proposal and the page is cross-origin isolated,
+falling back to the single-thread bundle otherwise. Detection + tests only; the
+bundle-selection wiring (threaded worker variant + orchestration) follows once
+verified against the running viewer.
--- a/.changeset/threaded-wasm-feature-detect.md
+++ b/.changeset/threaded-wasm-feature-detect.md
@@ -1,10 +1,21 @@
 ---
 "@ifc-lite/geometry": patch
+"@ifc-lite/wasm": patch
 ---
 
-Add `wasm-features` runtime detection (`supportsWasmThreads` / `isThreadedWasmUsable`)
-— the gate for selecting the threaded WASM geometry bundle (`pkg-threaded`) when
-the engine supports the threads proposal and the page is cross-origin isolated,
-falling back to the single-thread bundle otherwise. Detection + tests only; the
-bundle-selection wiring (threaded worker variant + orchestration) follows once
-verified against the running viewer.
+Add an opt-in threaded-WASM geometry path (in-instance rayon, shared memory).
+
+`@ifc-lite/wasm` gains a `./threaded` subpath export for the `pkg-threaded` bundle
+(built off by default via `BUILD_THREADED=1 ./scripts/build-wasm.sh`; exports
+`initThreadPool`). `@ifc-lite/geometry` adds `wasm-features` runtime detection
+(`supportsWasmThreads` / `isThreadedWasmUsable`) plus the wiring: the geometry
+worker dynamic-imports the threaded glue and brings up the rayon pool once, and
+the parallel orchestrator runs ONE shared-memory worker (whose internal
+`par_chunks` element loop fans CSG across cores) instead of the N single-thread
+worker pool. The default single-thread path is untouched.
+
+Off by default. The path is selected only when `globalThis.__IFCLITE_THREADED_WASM__`
+is `true` AND the page is cross-origin isolated. Browser A/B showed it is a win on
+CSG-bound models and a tie on large decode-bound models, with exact output parity;
+it stays opt-in until a threaded-failure fallback lands and production data confirms
+the per-model win.
--- a/apps/viewer-embed/vite.config.ts
+++ b/apps/viewer-embed/vite.config.ts
@@ -4,6 +4,7 @@ import wasm from 'vite-plugin-wasm';
 import topLevelAwait from 'vite-plugin-top-level-await';
 import path from 'path';
 import fs from 'fs';
+import { threadedWasmAliasTarget } from '../../packages/wasm/threaded-alias.mjs';
 
 // Read version from root package.json
 const rootPkg = JSON.parse(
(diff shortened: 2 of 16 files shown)


louistrue:
Addressed the CI failure and review feedback in 5f5b038:

- **CI Build failure** (`@ifc-lite/wasm/threaded` -> `pkg/ifc-lite.js/threaded` ENOTDIR): the threaded subpath is a gitignored artifact absent in normal CI, and Vite's alias plugin runs before user `pre` plugins so the `@ifc-lite/wasm` string alias prefix-matched it. Fixed at the alias layer: a `threaded-alias.mjs` target that resolves to the built bundle or a committed stub, added as `@ifc-lite/wasm/threaded` ordered before `@ifc-lite/wasm` in the viewer, viewer-embed, and both example configs. Verified by building all three with `pkg-threaded` absent (succeeds, bundles the 0.4 kB stub).
- **Major (worker pool guard):** `threadPoolStarted` is now set only after `initThreadPool()` resolves, so a failed bring-up retries instead of constructing a threaded `IfcAPI` with no pool.
- **wasm-features silent catch:** now logs the probe exception before falling back.
- **batch.rs dedup (threads):** reuses the per-worker `cached_item_dedup` across batches instead of a fresh cache per call (parity with the serial path).

louistrue:
## Full browser bench (real Chrome, dev, 10-core, cold, cache cleared each run, thermal-controlled)

Output parity is **exact** (plain@1-worker == threaded@1-worker byte-for-byte, e.g. ISSUE_068 both 11942 meshes / 2187k verts / 902.4K tris). plain@8-vs-threaded@1 count differences are the instancing partition grouping for 1 stream vs 8 shards.

| model | size | workload | plain@8 (default) | threaded@1 | result |
|---|---|---|---|---|---|
| ISSUE_129 | 11.5 MB | CSG-bound, balanced | 15.3 / 15.1 s | **11.0 s** | **WIN ~1.38x** |
| dental_clinic | 12.4 MB | light | **4.1 s** | 4.8 s | LOSS ~0.85x |
| advanced_model | 33.7 MB | heavy-tail | **7.2 s** (7.16/7.18) | 18.3 s (18.2/18.3) | **LOSS 2.5x** |
| ISSUE_068 | 53.7 MB | decode-bound | 14.1 / 16.2 s | 15.3 / 15.4 s | ~tie |

**Verdict:** the naive single-instance `par_chunks` is a net regression on most models (severe on heavy-tail), winning only on balanced-CSG-heavy models where the 8-worker pool's static shard imbalance strands one worker. Causes: heavy-tail can't split one huge element across threads; light/decode-bound pay rayon+atomics overhead, lose the warm per-batch decoder cache, and contend on dlmalloc's single global lock.

**This is exactly why the PR keeps it opt-in / off by default** (plain default path untouched, byte-parity-correct). The higher-ROI lever for "fast WASM" is dynamic load-balancing in the existing N-instance pool (captures the ISSUE_129 win without the dlmalloc/heavy-tail/decode regressions), not in-WASM threads. Full analysis in `docs/architecture/csg-threading-design.md`.

louistrue:
Closing: the A/B benchmark showed in-instance WASM threading is a net regression on most benchmarked models (the parallelized produce_batch is dominated by per-worker file re-decode + dlmalloc global-lock contention); the single apparent win was a shard-stall artifact, not threading. The path was already off by default, so this ships nothing for users. The dead-end finding is recorded in agent memory so we don't re-investigate. The real CSG levers are cutter-chunking (#1432), worker-pool right-sizing (#1431), and wide-arithmetic (#1426).

PR #1430, opened 2026-06-28
fix(export): clamp KMZ to ground + emissive materials for Google Earth (#1427)

### LINKED ISSUE #1427: Orthogonal Height with show in world 3D kontext
I have a georeferenzed model that shows correctly when i click on 'show 3d world context'.for google earth When I then use the feature to create a .kmz file for google earth the model (extremely dark)  will be placed correctly on the x and y axis but the correct orthogonal height shows the model to high in the air. In my case the true orthoganl height is about 560m MSL. Therefore the model floats in the sky. When i set the height in the IFC Model to 1m (0m) it will be shown correctly on google earth. I suppose the map in google earth is basically a flat image and the heights are stored in attributes? 
So finally it might not be a real error, but for the sake of visualization it might be helpful to set the heigt to 0m for the .kmz export

Image

The image shows the IFC model with a false orthogonal height of 2m. Google earth indicate the correct height when marking a point.

louistrue:
Fixes #1427.

Makes the Google Earth (KMZ) export actually **load and render correctly**, and adds it to the export menu.

## The real problem: format
Google Earth's KML `<Model>` only loads **COLLADA** (`.dae`). A glTF/GLB embedded in `<Model>` fails with **"Unsupported element: Model"** — so the previous KMZ never rendered, regardless of placement/material tweaks. The KMZ now embeds a COLLADA `.dae`.

## What changed
- **COLLADA exporter** (`rust/export/src/collada.rs`, `export_collada_from_meshes`): COLLADA 1.4.1 in the IFC-native **Z-up** frame so the building stands upright; per-colour materials with `<emission>` = element colour (Google Earth has no ambient/IBL and a single hard sun, so plain diffuse renders near-black) and `double_sided` for IFC's unreliable winding. Output **validated against the COLLADA 1.4.1 XSD** with `xmllint --schema`.
- **One-call KMZ from meshes** (`export_kmz_collada_from_meshes` → wasm `exportKmzFromMeshes`): builds the COLLADA + KML (`clampToGround` placement so it rests on the terrain instead of floating at its MSL `OrthogonalHeight`; grid-north heading) + zip, straight from the viewer's already-produced meshes.
- **Viewer**: the Location panel "Google Earth" button now uses the COLLADA path.
- **Export menu**: new **"Export KMZ (Google Earth)"** entry alongside GLB / IFC / HBJSON, using the shared model-name file-stem scheme (`<name>.kmz`) and a clear message when a model isn't georeferenced.
- Also adds a general `emissive` option to the GLB exporter (`emissiveFactor = base colour`) for renderers without ambient/IBL.

## Background
A research pass (Google KML docs + Khronos) confirmed: `<Model>` is COLLADA-only, Google Earth ignores `KHR_materials_unlit`, and `<emission>` is the brightness lever. The `relativeToGround + OrthogonalHeight` combination was the float bug.

## Verification
- Rust: 43 `ifc-lite-export` tests (COLLADA skeleton, Y-up→Z-up, emission/double_sided, transparency, COLLADA-in-KMZ, clamp default), clippy clean.
- COLLADA output **schema-valid** (`xmllint --schema collada_1_4_1.xsd` → validates).
- TS: kmz + glb suites pass; `@ifc-lite/geometry` builds; **viewer `tsc` clean**; regenerated committed `pkg/ifc-lite.d.ts`.
- Changeset updated.

Note: full visual confirmation requires opening the `.kmz` in Google Earth; the structure is schema-validated and matches Google's documented COLLADA `<Model>` requirements.

## Summary by CodeRabbit

* **New Features**
  * Added “Export KMZ (Google Earth)” to the main export dropdown via a dedicated export dialog.
  * Added configurable KMZ altitude handling (Clamp-to-Ground, Relative-to-Ground, Absolute).
  * Added an `emissive` option for GLB exports to improve material visibility in low-light renderers.

* **Bug Fixes**
  * Improved Google Earth KMZ placement and rendering by exporting KMZ with embedded COLLADA and proper ground clamping.
  * Fixed dark/low-visibility output using emissive materials and double-sided rendering.
  * Show a clearer message when KMZ export can’t proceed due to missing or invalid georeferencing.


louistrue:
## Scope B: user-facing altitude toggle added

Building on the ground-clamping fix already on this branch, the KMZ export dialog now has a **Placement** toggle:

- **Rest on ground** (default) writes `<altitudeMode>clampToGround</altitudeMode>` so the model drapes on Google Earth's terrain and can never float (immune to a wrong/zero/double-counted `OrthogonalHeight`).
- **True elevation (MSL)** writes `<altitudeMode>absolute</altitudeMode>` with `<altitude>` = the model's `OrthogonalHeight`, for models whose elevation is a genuine MSL value.

### Threading (single choke point out to the UI)
`rust/wasm-bindings/src/api/export_glb.rs` (new optional `altitude_mode`, defaults to `ClampToGround` so existing callers are unchanged) -> `packages/geometry` `exportKmzFromMeshes` (bridge + index, new `altitudeMode`) -> viewer `KmzOptions` / `buildKmzForModel` -> a two-option `Select` in `KmzExportDialog.tsx`. No new Rust logic: the `AltitudeMode` enum + `as_kml` + `build_kml` already supported all three modes. The Location panel's one-click Google Earth button stays one-click ground-clamped.

Also folds in the pre-existing WIP: instanced-type templates (`geometryClass === 2`) are dropped before KMZ export so instanced elements are not duplicated at the type's origin.

### wasm / .d.ts
Rebuilt wasm and regenerated `packages/wasm/pkg/ifc-lite.d.ts` for the new binding signature (`altitude_mode?: string | null` trailing optional on both `exportKmz` and `exportKmzFromMeshes`). `pkg/package.json` was not dirtied.

### Verification
- `cargo test -p ifc-lite-export kmz` -> 5/5 pass (incl. `default_altitude_mode_clamps_to_ground` and `kml_carries_placement` covering absolute + `<altitude>`).
- Viewer kmz unit tests -> 8/8 pass (default `undefined` -> Rust clampToGround, explicit `absolute` carries the orthogonal height, explicit `clampToGround`, plus the geometryClass-2 filter).
- Viewer `tsc --noEmit` -> clean; geometry package `tsc` -> clean.
- **Manual step:** final visual confirmation in **Google Earth Pro (desktop)** (ground vs MSL placement) cannot be automated in this environment and should be done by hand. The live dialog render was not driven here because the shared debug browser was in use by another session.

PR #1432, opened 2026-06-28
fix(geometry): cap cutters per CSG arrangement (geometry-stream stall fix)

louistrue:
## Problem

An 86 MB IFC (1.0M entities, ~610 void hosts) **stalled at the 40 s geometry-stream watchdog and failed to load** in the browser — it rendered most geometry then one worker got stuck. The log showed `IfcBuildingElementPart` hosts carrying dozens of `NonRectangular` openings (one with ~90) failing with `OperandTooLarge`.

## Root cause (measured)

`subtract_mesh_many` subtracts every disjoint cutter of a host in **one N-ary conforming arrangement**, whose cost is **super-linear in the cutters packed into it**. Native per-job profiling of the model:

- The dominant job: a host with only 3,292 host-triangles but **~90 cutters (2,548 cutter-triangles)** cost **12.3 s in one arrangement** (parallel-contended). On WASM that single element alone exceeds the 40 s watchdog.
- Natively the full pipeline loads in ~15 s (the per-element escalation budget bails it to AABB), but the budget bounds *escalations*, not the *base arrangement size* — so it doesn't prevent the WASM stall.

## Fix

Void cutters here are order-free — set difference is associative/commutative: `host − {all} ≡ host − {chunk₁} − {chunk₂} − …`. So `subtract_mesh_many` now processes cutters in **chunks of 16**, bounding per-arrangement cost so no single element can stall the stream.

- **Solid-equivalent.** The batch path's contract is volume parity + watertightness, not byte-identical tessellation (same as the existing disjoint-batch). For `<= 16` cutters it is exactly the prior single arrangement.
- **No worse on failure.** On any chunk's budget trip / unrecovered constraint, the whole group still falls through to the per-opening sequential path (own budget + #635 AABB fallback), exactly as before.

## Measured impact

- The ~90-cutter killer job: **12.3 s → 0.4 s (30x)**; total CSG corpus **−38%**.
- **The previously-stalling 86 MB model now loads completely (~35 s, no stall) and renders correctly** (8 storeys, 54K elements, verified in-browser).

## Tests

- Full `ifc-lite-geometry` suite green (550 tests, 0 failures), **no snapshot changes**.
- Existing `subtract_many_disjoint_openings_matches_sequential` / `subtract_many_two_pocket_group_matches_sequential` equivalence tests pass.
- New `subtract_mesh_many_chunks_match_sequential`: 20 disjoint cutters (>cap ⇒ chunked) match the sequential subtract chain by volume.

## Follow-up (not in this PR)

The model now loads at ~35 s — still close to the watchdog. The remaining cost is dense single-host cuts (20-30k-triangle hosts, 1 cutter each); a host-triangle spatial cull to the cutter region would cut those further. Tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CSG void-cutter subtraction for models with many openings: disjoint cutters are now processed in capped batches to avoid stalls and failed geometry renders.
  * Results remain solid-equivalent (watertight and volume-consistent); behavior matches the prior approach for smaller cutter counts.
* **Tests**
  * Added coverage to confirm chunked subtraction matches sequential results for larger cutter sets.
* **Chores**
  * Updated the geometry cache format to ensure restored models re-mesh with the correct tessellation after this change.

### DIFF
--- a/.changeset/csg-cutter-chunk.md
+++ b/.changeset/csg-cutter-chunk.md
@@ -1,5 +1,6 @@
 ---
 "@ifc-lite/geometry": patch
+"@ifc-lite/cache": patch
 ---
 
 Cap the number of void cutters packed into a single CSG arrangement, fixing a
@@ -22,3 +23,10 @@ equivalence tests and a new 20-cutter chunked-equivalence test all pass, and the
 full geometry suite is unchanged). For hosts with <= 16 cutters this is exactly
 the prior single arrangement. Verified end to end: the previously-stalling model
 now loads completely and renders correctly.
+
+Bumps the geometry cache `FORMAT_VERSION` (10 → 11). For a host with > 16 void
+cutters the chunked cut is solid-equivalent but not byte-identical (and on
+pre-fix builds those hosts often fell back to an AABB box), so the mesh hash
+changes. The bump invalidates pre-fix caches so restored models re-mesh with the
+correct tessellation, and the compare/diff feature does not flag those hosts from
+a stale-cache hash mismatch.
--- a/.changeset/csg-cutter-chunk.md
+++ b/.changeset/csg-cutter-chunk.md
@@ -1,0 +1,24 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Cap the number of void cutters packed into a single CSG arrangement, fixing a
+geometry-stream stall on models with elements that carry many openings.
+
+`subtract_mesh_many` previously subtracted every disjoint cutter of a host in ONE
+N-ary conforming arrangement. That arrangement's cost is super-linear in the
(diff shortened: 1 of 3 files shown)


louistrue:
Addressed the review in 6508a7c1:
- **Major (validate each chunk):** each chunk's result is now validated before it becomes the next chunk's host; an invalid/non-watertight intermediate rejects the whole group to the per-opening sequential path (the same guard the un-chunked path had, now applied per chunk).
- **Census per chunk:** `record_csg_op` now fires per chunk with the real per-invocation operand sizes, instead of one synthetic op carrying the whole group's cutter total. For <=16 cutters both are byte-identical to before.

Clippy clean; full `ifc-lite-geometry` suite 551 green incl. the equivalence + new chunk tests.

PR #1445, opened 2026-06-29
perf(geometry): share one entity index across geometry workers

louistrue:
## What

Each WASM geometry worker rebuilds its own 12.6M-entry `EntityIndex` FxHashMap. On a 722MB / 12.6M-entity model that's ~354MB **per worker** (×3 ≈ 1.06GB) plus 12.6M hashmap inserts each — a big chunk of the 3.25GB peak that makes huge loads slow in multi-model sessions.

This shares **one** index: `setEntityIndex` builds a `SharedEntityIndex` (a sorted `(id,start,end)` buffer, looked up by **binary search**) from the index columns the orchestrator **already streams** to every worker, and the batch decoder uses it via `with_shared_index`. Per worker: **~152MB vs ~354MB** (~600MB lower peak across 3 workers) and **no hashmap inserts**.

## Why it's safe / byte-identical

Same `id → (start,end)` lookups → same decode → **byte-identical geometry**. Parity tests assert the shared index returns identical spans to the FxHashMap for every id, and that `from_columns == from_map == from_content`. The owned index path is unchanged when no columns are provided (CLI / server / single-worker).

## Scope

Rust-only — **no JS / prepass / orchestrator change**: reuses the existing entity-index column plumbing (`geometry.worker.setEntityIndex`). Spike (`rust/core/examples/index_share_spike.rs`) measured binary-search lookup at ~24ns (1.8× a hashmap get, negligible) and validated the memory delta.

## Verification

- 8/8 core parity tests; cargo-check clean on wasm32; production wasm builds.
- CI E2E smoke exercises the real `setEntityIndex → shared index → produce_batch` path; snapshot suite confirms geometry is unchanged.
- Pending: browser re-measure on the 722MB file to confirm the peak drop (memory win is only visible at huge entity counts).

## Summary by CodeRabbit

* **New Features**
  * Geometry processing now uses a shared entity index when available, reducing repeated work across batches and workers.
  * Added support for building and reusing a compact shared index from existing index data.

* **Bug Fixes**
  * Preserved existing output and behavior when no entity index is provided.
  * Added safer handling for invalid or mismatched index data, falling back to an empty index instead of failing.

* **Tests**
  * Added coverage for shared-index lookups, data conversion, ordering, and malformed input cases.

### DIFF
--- a/.changeset/shared-entity-index.md
+++ b/.changeset/shared-entity-index.md
@@ -1,0 +1,13 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Share one binary-searchable entity index across geometry workers instead of each
+worker building its own 12.6M-entry FxHashMap. `setEntityIndex` now builds a
+`SharedEntityIndex` (a sorted `(id,start,end)` buffer looked up by binary search)
+from the index columns the orchestrator already streams to each worker, and the
+batch decoder uses it. On a 722MB / 12.6M-entity model this cuts each worker's
+index from ~354MB to ~152MB (~600MB lower peak across 3 workers) and drops the
+per-worker hashmap inserts, easing the memory pressure that slowed large loads in
+multi-model sessions. Geometry output is byte-identical (same id→span lookups);
+the owned index path is unchanged when no columns are provided (CLI/server).
--- a/rust/core/examples/index_share_spike.rs
+++ b/rust/core/examples/index_share_spike.rs
@@ -90,12 +90,18 @@ fn main() {
     let flat_bytes = n as f64 * 12.0; // (u32,u32,u32), no padding
     let workers = 3.0;
     println!("\n[memory estimate, {n} entities]");
-    println!("  per-worker FxHashMap:  {:.0} MB  x{workers:.0} workers = {:.0} MB",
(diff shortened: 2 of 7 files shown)


louistrue:
Closing as a measured dead-end. The end-to-end 722MB re-measure REFUTES the projected win: peakWasm went UP ~680MB (3930MB vs main's 3250MB), not down ~600MB. Geometry is byte-identical (140755 meshes) and there's no time change, so this is a net peak-memory regression.

Why: the spike measured the *retained* index size (152 vs 354MB) but the end-to-end peak is set DURING the build — setEntityIndex's from_columns double-buffers (a Vec<(u32,u32,u32)> plus the output Vec<u8>, ~2x152MB transient) per worker, and across 3 concurrent workers that transient spike exceeds the old single-FxHashMap footprint. The retained saving never shows. And the dominant huge-file memory — the 3x 722MB source copy (~2.2GB) — is untouched by the index change.

Third case of an isolated bench misleading vs end-to-end (after threading #1429 and the Manifold accel). The real huge-file memory lever is the per-worker source re-copy. Restorable from this closed PR if a transient-free in-place build is ever worth revisiting.

PR #1451, opened 2026-06-29
chore(deps): upgrade Vite 7 → 8 and @vitejs/plugin-react 4 → 6

louistrue:
## What

Upgrades the workspace to **Vite 8** and **@vitejs/plugin-react 6**, together.

| Dep | From | To |
|-----|------|----|
| `vite` | ^7.3.6 | ^8.1.0 |
| `@vitejs/plugin-react` | ^4.2.0 | ^6.0.3 |

Touched: `apps/viewer`, `apps/viewer-embed`, `examples/*`, and the root `pnpm.overrides` (vite), plus the regenerated `pnpm-lock.yaml`.

## Why

Dependabot #1437 proposed `@vitejs/plugin-react` 4 → 6 on its own, but plugin-react 6 declares `vite: ^8` as a peer and imports `vite/internal`, so it fails the build on vite 7 with `ERR_PACKAGE_PATH_NOT_EXPORTED`. The two have to move together. This PR carries that bump, so **#1437 is superseded** by this.

## Compatibility checked

- `vite-plugin-wasm`, `vite-plugin-top-level-await`, `vite-plugin-static-copy` already declare `^8` support at their current versions — no bumps needed.
- `vitest@4.1.x` supports `vite ^6 || ^7 || ^8` — the ~30 packages on vitest are unaffected.
- Root `esbuild` override is already `^0.28.1` (from the npm-minor group), satisfying vite 8's esbuild peer.

## Notes on Vite 8

Vite 8 ships Rolldown/Oxc as the default bundler. **No config changes were required:** the deprecated `build.rollupOptions` and the `manualChunks` *function* form still build under Rolldown (object form is the one that was removed). The `@tauri-apps/*` externals and the custom `worker` config work as-is.

## Verification (local)

- ✅ `turbo build` of `@ifc-lite/viewer` + `@ifc-lite/viewer-embed` — 31/31 tasks, both `dist/` emitted
- ✅ `tsc --noEmit` on the viewer (vite.config.ts + src) clean on vite 8 types
- Only non-fatal warnings remain (pre-existing chunk-size warning; a Rolldown plugin-timing note for `vite-plugin-wasm`)

### DIFF
--- a/apps/viewer-embed/package.json
+++ b/apps/viewer-embed/package.json
@@ -40,9 +40,9 @@
   "devDependencies": {
     "@types/react": "^19.2.17",
     "@types/react-dom": "^19.2.3",
-    "@vitejs/plugin-react": "^4.2.0",
+    "@vitejs/plugin-react": "^6.0.3",
     "typescript": "^6.0.3",
-    "vite": "^7.3.6",
+    "vite": "^8.1.0",
     "vite-plugin-top-level-await": "^1.6.0",
     "vite-plugin-wasm": "^3.6.0"
   }
--- a/apps/viewer/package.json
+++ b/apps/viewer/package.json
@@ -98,11 +98,11 @@
     "@types/react": "^19.2.17",
     "@types/react-dom": "^19.2.3",
     "@types/three": "^0.185.0",
-    "@vitejs/plugin-react": "^4.2.0",
+    "@vitejs/plugin-react": "^6.0.3",
     "fake-indexeddb": "^6.2.5",
     "tsx": "^4.22.4",
     "typescript": "^6.0.3",
-    "vite": "^7.3.6",
+    "vite": "^8.1.0",
     "vite-plugin-static-copy": "^4.1.1",
     "vite-plugin-top-level-await": "^1.6.0",
     "vite-plugin-wasm": "^3.6.0"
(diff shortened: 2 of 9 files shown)


louistrue:
## Caught a real Rolldown init-order crash (now fixed)

The first CI run was green on Build/WASM, Typecheck, Lint, and Node tests, but **Viewer E2E smoke failed** — the app never rendered the file input. Reproduced locally against `vite preview` (the same server the E2E harness uses) and captured the boot error:

```
TypeError: Cannot read properties of undefined (reading 'map')
    at assets/persistence-*.js  (apps/viewer/src/lib/clash/persistence.ts)
```

Root cause: `clash/persistence.ts` computed `new Set(CLASH_RULE_PRESETS.map(...))` **at module-eval time**. `CLASH_RULE_PRESETS` is an `export const` from `@ifc-lite/clash` (resolved to its built dist, not aliased to src). Vite 7/Rollup ordered the chunks so the clash chunk ran first; **Vite 8/Rolldown does not guarantee that**, so the cross-chunk import was `undefined` and the whole app failed to mount.

Fix (899eb823): make the Set lazily memoized so the cross-package import is only read on first call, after all modules have initialized. Both call sites are inside functions, so no behavioural change. Verified locally that the production bundle now boots clean (file input present, zero uncaught `pageerror`s). CI re-running.

PR #1465, opened 2026-06-30
perf(export): stream attribute extraction in bounded memory

louistrue:
## What

`build_export_model` — the shared attribute extractor behind the CSV/JSON/JSON-LD/IFC5/STEP/Parquet exporters — cannot process very large models. It decodes and **caches every** STEP entity and retains a `Vec<EntityRow>` for all products, so a 3.9 GB / 74 M-entity model OOMs during attribute extraction and never finishes.

This adds a streaming variant and makes `build_export_model` a thin wrapper over it.

## Root cause

- `EntityDecoder::decode_at` unconditionally inserts every decoded entity into a cache that is never cleared (`rust/core/src/decoder.rs:172`).
- Pass 2 of `build_export_model` decoded **every** entity and applied the `is_subtype_of(IfcProduct)` filter *after* decode, so all ~74 M cartesian points / directions / index lists were decoded and pinned in the cache.
- The full `Vec<EntityRow>` for all products was retained until return.

Net working set on a 3.9 GB model trends toward ~12–20 GB → OOM.

## Change

- **`EntityDecoder::decode_at_uncached`** — parse an entity without populating the cache. Purely additive; no existing behavior changes.
- **`stream_export_model(content, impl FnMut(EntityRow))`** — filter products by STEP keyword via `IfcType::from_str` **before** decoding (provably identical to the old post-decode filter, since `parse_entity` resolves the type with the same `from_str`), decode uncached, bound the property-resolution cache, then yield each row and drop it.
- **`build_export_model`** becomes `collect` over `stream_export_model` — one code path, old accumulating path deleted.

## Correctness

- All existing export golden tests pass unchanged (CSV / JSON / JSON-LD / IFC5 / OBJ / STEP / merged, plus GLB byte-determinism) because they all go through `build_export_model`.
- New test `stream_matches_build_row_for_row` asserts the streamed and collected rows agree row-for-row.
- Real-model parity: a 986 MB model yields byte-identical output (141264 products / 208771 psets) before and after.

## Measured impact (M4, `--profile server-release`)

| Model | before | after |
|---|---|---|
| 3.9 GB / 74 M entities (Revit coordination export) | OOM / timeout (never finishes, >300 s) | **42 s, ~6.6 GB peak** |
| 986 MB / 141 k products | 5.84 GiB | 5.18 GiB (identical output) |

This brings attribute extraction into the same memory envelope as the already-streaming geometry path, so a 10 GB worker can process a ~4 GB model end to end.

## Notes

- Pure Rust, no new TS/wasm surface → no changeset / api-surface update.
- Follow-up (separate PR): share one `entity_index` between the geometry and attribute passes to avoid the double parse.

## Summary by CodeRabbit

* **New Features**
  * Added a streaming HBJSON/IFC export API to process exported rows one-by-one for large inputs.
  * Added support for decoding a single entity from a specified byte range without using cached state.
* **Bug Fixes**
  * Improved robustness when decoding entities with invalid byte spans (validated before parsing).
* **Tests**
  * Added coverage to verify streamed export rows match the standard export output row-for-row.
* **Refactor**
  * Updated shared export model types to support row-by-row equality checks.

### DIFF
--- a/rust/core/src/decoder.rs
+++ b/rust/core/src/decoder.rs
@@ -173,6 +173,46 @@ impl<'a> EntityDecoder<'a> {
         Ok(entity)
     }
 
+    /// Decode the entity in `[start, end)` **without** touching the cache.
+    ///
+    /// [`decode_at`](Self::decode_at) memoizes every entity it parses, which is the
+    /// right trade-off for geometry sub-tree walks (the same points/profiles are
+    /// revisited many times). For a single linear pass over *every* entity in a
+    /// large model — tens of millions of rows — that cache grows without bound and
+    /// dominates memory. This variant parses and returns the entity but never
+    /// inserts it, so a streaming walk stays O(1) in entity count. The caller owns
+    /// the result; for identical bytes it yields an identical [`DecodedEntity`] to
+    /// `decode_at`, only without the cache side effect.
+    pub fn decode_at_uncached(&self, start: usize, end: usize) -> Result<DecodedEntity> {
+        let content_len = self.content.len();
+        if start > end || end > content_len {
+            return Err(Error::parse(
+                0,
+                format!(
+                    "decode_at_uncached: invalid byte span ({}, {}) for content length {}",
+                    start, end, content_len,
+                ),
+            ));
+        }
+        let line = &self.content[start..end];
+        let (id, ifc_type, tokens) = parse_entity(line).map_err(|e| {
+            let cut = line.len().min(100);
+            Error::parse(
+                0,
+                format!(
+                    "Failed to parse entity: {:?}, input: {:?}",
+                    e,
+                    String::from_utf8_lossy(&line[..cut])
+                ),
+            )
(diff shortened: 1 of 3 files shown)


louistrue:
Updated (force-push `26121f12`):

- **Addressed the Codex P2** — `stream_matches_build_row_for_row` now **skips** (with a `pnpm fixtures` hint) when the fixture is absent, per `AGENTS.md` §Test fixtures, instead of panicking. Re-verified: 85 export + 39 core tests pass with fixtures present; the test skips cleanly without them.
- **Left the CodeRabbit DRY nit** (share the parse body between `decode_at` / `decode_at_uncached`) as-is on purpose: `decode_at` is on the geometry hot path and deliberately skips attribute-building on a cache hit, so a shared helper would either touch that hot path or DRY differently than it looks. Happy to extract a `parse_span` helper if you'd prefer it.

PR #1467, opened 2026-06-30
feat(export): emit globalId and modelId in glTF node extras

louistrue:
## What

Per-node glTF `extras` carried only `expressId` / `ifcType`. This adds:

- **`globalId`** — the element's IFC `GlobalId`, threaded from `MeshData.global_id` through `MeshView` into `node_extras` (from-bytes path; the from-meshes path has no GlobalIds and is unchanged).
- **`modelId`** — a new optional `GltfOptions.model_id`, stamped on every element node so a host can distinguish elements from different models that share express-id space (federation).

## Notes

- Both emit only with `include_metadata`.
- Additive: existing keys and the no-`model_id` output stay byte-identical, so `export_is_byte_deterministic` stays green.
- Adds `nodes_carry_global_id_and_model_id`; updates the wasm-bindings `GltfOptions` construction for the new field.

## Summary by CodeRabbit

* **New Features**
  * glTF/GLB exports can now include model identity and IFC global IDs in node metadata.
  * Exported nodes may carry `extras.modelId` and `extras.GlobalId` when available.

* **Bug Fixes**
  * Metadata handling now consistently applies to both flat and instanced nodes during export.

* **Tests**
  * Added coverage to verify node metadata includes global IDs and model IDs when enabled, and omits model IDs when not provided.

<h3>Confidence Score: 5/5</h3>

Safe to merge — changes are purely additive, all element-node code paths are correctly updated, and the byte-determinism test continues to pass.

All three element-node paths (flat, content-hash-deduped repeat, instanced occurrence) are updated consistently. In the instanced path, each occurrence node correctly uses occ_view.global_id rather than the template's, so every element gets its own IFC GlobalId. The expect on the json!-macro result is infallible by construction. The WASM binding's model_id: None is an intentional, documented limitation rather than an oversight.

No files require special attention.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| rust/export/src/gltf.rs | Adds `model_id` to `GltfOptions`, threads `global_id` through `MeshView` into `node_extras`, and wires both into all element-node code paths (flat, content-hash-deduped repeated, and instanced). Logic is correct: each occurrence node uses `occ_view.global_id` (not the template's), the `expect` on `json!`-built object is safe, and the deterministic-export test is preserved. |
| rust/wasm-bindings/src/api/export_glb.rs | Hardcodes `model_id: None` with an explanatory comment deferring WASM-level federation support. The choice is intentional and documented; no other logic is changed. |

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `rust/wasm-bindings/src/api/export_glb.rs`, line 38-45 ([link](https://github.com/ltplus-ag/ifc-lite/blob/0dbf4a49f6cdcc5511c187860fea2fc139e9623c/rust/wasm-bindings/src/api/export_glb.rs#L38-L45)) 

   <a href="#">P2</a> **`model_id` not exposed to WASM callers**

   The `exportGlb` binding hardcodes `model_id: None`, so browser/JS consumers that call the WASM API cannot supply a `model_id` for federation use — the feature is only reachable through the Rust API directly. If the WASM surface is the primary integration point, callers that rely on federation disambiguation of shared express-id spaces will silently get no `modelId` in their glTF node extras regardless of what they intend.
</details>

<sub>Reviews (2): Last reviewed commit: ["feat(export): emit GlobalId and modelId ..."](https://github.com/ltplus-ag/ifc-lite/commit/d5f028065f2afb2c04fdf037985e0967a0b35771) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40824444)</sub>

### DIFF
--- a/rust/export/src/gltf.rs
+++ b/rust/export/src/gltf.rs
@@ -33,6 +33,10 @@ pub struct GltfOptions {
     /// render flat with just the apparent base colour (the historical behaviour,
     /// kept for colour-accurate exports). Default `true`. (#1321)
     pub lit: bool,
+    /// Per-model id stamped into every node's `extras.modelId` (federation: lets a
+    /// host distinguish elements from different models that share express-id space).
+    /// `None` ⇒ single model, no `modelId` emitted. Requires `include_metadata`.
+    pub model_id: Option<String>,
 }
 
 impl Default for GltfOptions {
@@ -43,6 +47,7 @@ impl Default for GltfOptions {
             hidden: Vec::new(),
             hidden_types: Vec::new(),
             lit: true,
+            model_id: None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


louistrue:
Addressed Codex review (`0dbf4a49`):

- **P1 (GlobalId casing) — fixed.** Extras now emit `GlobalId` (EXPRESS PascalCase per `AGENTS.md`), not `globalId`. `expressId`/`ifcType`/`modelId` stay camelCase, as they are synthetic fields, not IFC attributes.
- **P2 (expose modelId through wasm) — deferred.** The wasm/viewer export path doesn't federate (`modelId` is a server-side concern); the seam exists on `GltfOptions`, and the wasm binding can take a parameter when the browser path needs federation.

PR #1500, opened 2026-07-02
fix(core,encoding): decode STEP '' quote escapes; stop \S\ multibyte panic

louistrue:
Found via a full multi-agent code review of `main`; verified by reading the code and pinned with a regression + the shared parity fixture.

## Two STEP-string decoding bugs in the shared decoders

**1. `''` doubled-quote escape never collapsed (data corruption).** ISO 10303-21 escapes a literal apostrophe inside a string as `''`. The tokenizer deliberately scans past `''` and hands the raw content (e.g. `O''Brien`) to the decoder, but neither `decode_ifc_string` (Rust) nor `decodeIfcString` (`@ifc-lite/encoding`) collapsed it — so any IFC name/description with an apostrophe kept the doubled quote. Worse, the STEP writers correctly re-double on export, so a decode/encode round-trip *compounds*: `O'Brien` → `O''Brien` → `O''''Brien`. Both decoders now collapse `''` → `'`.

**2. `\S\C` panic on a multi-byte `C` (crash → wasm instance abort).** The Rust `\S\` branch read one byte and advanced by 4. If `C` was a multi-byte UTF-8 char (valid in a `&str`, reachable from a malformed-but-UTF-8 file), the cursor landed mid-character and the next `s[i..]` slice panicked. Under `panic = 'abort'` on wasm that kills the whole instance mid-load. It now reads `C` as a whole char and advances by its UTF-8 length.

## Tests
- New `step_encoding.rs` cases: `''` collapse, and `\S\` before multi-byte chars must not panic.
- Two `''` vectors added to the shared `ifc_string_vectors.json`, which pins **both** the Rust and TS decoders (`ifc_string_parity.rs` + `ifc-string.parity.test.ts`).
- `cargo test -p ifc-lite-core` and `@ifc-lite/encoding` tests pass; full workspace typecheck clean.

Changeset: `@ifc-lite/encoding` patch (Rust crate rides the release anchor).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved STEP string escape decoding so malformed or non-BMP characters are handled safely without panics.
  * Preserved existing behavior for standard ASCII escape cases while avoiding incorrect character splitting.

* **Tests**
  * Added regression coverage for edge-case escape sequences, including non-BMP character handling.
  * Expanded fixture data to verify consistent decoding across implementations.

### DIFF
--- a/.changeset/fix-step-string-decode.md
+++ b/.changeset/fix-step-string-decode.md
@@ -1,0 +1,13 @@
+---
+"@ifc-lite/encoding": patch
+---
+
+fix(encoding): stop `\S\` decoding from diverging / panicking on multi-byte input
+
+The `\S\C` STEP escape (code point of `C` plus 128) is spec-defined for a single
+ASCII `C`, but a malformed-but-UTF-8 file can put a multi-byte `C` there.
+`decodeIfcString` now reads `C` as a whole code point (advancing past a surrogate
+pair) instead of one UTF-16 unit, so it no longer leaves a dangling surrogate and
+stays in parity with the Rust `decode_ifc_string`, whose matching fix also stops
+a multi-byte `C` from panicking mid-slice (which aborts the wasm instance). Pinned
+by a new non-BMP `\S\` case in the shared `ifc_string_vectors.json` fixture.
--- a/packages/encoding/src/ifc-string.ts
+++ b/packages/encoding/src/ifc-string.ts
@@ -10,6 +10,10 @@
  * - \X\XX\ - ISO-8859-1 hex encoding
  * - \S\X - Extended ASCII with escape
  * - \P..\ - Code page switches (supported as directives and removed)
+ *
+ * This handles only backslash escapes. The '' doubled-quote escape is collapsed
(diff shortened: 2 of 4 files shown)


louistrue:
Addressed review feedback:
- **Codex P2 (double-unescape `''`)** — correct; `entity-extractor.ts`, `merged-exporter.ts`, and `quick_metadata.rs` already un-double before decoding, so collapsing in the decoder too would double-collapse (`2'' opening` → `2' opening`). Reverted the `''` collapse from both decoders; this PR is now scoped to the `\S\` multibyte fix. The `''` gap on the *non*-collapsing paths (columnar names, Rust `from_token`) needs a single-layer refactor and is tracked separately.
- **Greptile P2 (non-BMP `\S\` divergence)** — TS now reads a whole code point via `codePointAt` (advancing past the surrogate pair) to match Rust; pinned by a non-BMP `\S\` vector in the shared fixture.
- **Greptile P2 (Rust release path)** — `ifc-lite-core` publishes via the crates release job (`release-crates.mjs`, OIDC) whose Cargo version rides the npm release, so the crate fix ships on the next release triggered by the `@ifc-lite/encoding` changeset.

PR #1501, opened 2026-07-02
fix(collab-server): merge persisted Y frames (data loss), uncollide room ids, unpoison audit log

louistrue:
Found via a full multi-agent code review of `main`; verified by reading the code and covered with regression tests.

## 1. CRITICAL: room reload silently drops all edits after the first frame
Every room stores one Yjs update per `append`. All four persistence backends (`Memory`, `File`, `Redis`, `S3`) returned the frames **byte-concatenated**, and `RoomManager.loadFromDisk` fed that blob to `Y.applyUpdate` — which decodes only the *first* update in a concatenation and silently ignores the rest. Between compactions (default `compactEvery: 1000`) up to ~999 updates accumulate, so a server restart / room eviction+reload loses every edit after the first frame. (The S3 backend even carried a comment acknowledging ">1000 un-compacted frames would silently drop the tail.")

Fix: a shared `mergeUpdateFrames` helper combines frames with `Y.mergeUpdates` (single-frame fast-path preserves the compacted-snapshot case), routed through all four backends.

## 2. FilePersistence room-id collision
`logPath` sanitized ids with a lossy `[^a-zA-Z0-9._-] -> _` replace, so distinct rooms (`a/b`, `a:b`, `a_b`) mapped to one `.log` file and interleaved their updates. Now uses `encodeURIComponent` (reversible, collision-free, path-traversal-safe), matching the S3 backend. Safe ids (UUIDs, room codes) are unchanged, so existing logs keep their names.

## 3. Audit-log write chain poisoned by one failure
`JsonlFileAuditSink.append` chained each write onto a shared `inflight` promise with a bare `.then()`. One transient append rejection left `inflight` rejected forever, so every subsequent append's callback was skipped — the audit log went permanently silent. Writes now run after the previous one settles regardless of outcome; callers still see their own error.

## Tests
- New `test/persistence.test.ts` for the untested File/Memory backends: multi-frame load reconstructs the full doc; compacted-snapshot-then-appends loads in full; distinct-but-colliding room ids stay isolated.
- Rewrote the `Redis`/`S3` round-trip + pagination tests to append **real** Y updates and assert doc reconstruction (they previously pinned byte concatenation with fake bytes — the anti-pattern that masked this bug).
- All 47 `@ifc-lite/collab-server` tests pass; full workspace typecheck clean.

Changeset: `@ifc-lite/collab-server` patch.

### DIFF
--- a/.changeset/fix-collab-persistence-data-loss.md
+++ b/.changeset/fix-collab-persistence-data-loss.md
@@ -1,0 +1,22 @@
+---
+"@ifc-lite/collab-server": patch
+---
+
+fix(collab-server): merge persisted update frames instead of byte-concatenating them
+
+Room logs store one Yjs update per `append`. Every persistence backend
+(`Memory`, `File`, `Redis`, `S3`) returned the frames byte-concatenated, and
+`loadFromDisk` fed that blob to `Y.applyUpdate`, which decodes only the first
+update and silently ignores the rest — so every edit after the first frame was
+lost on room reload (up to `compactEvery` updates between compactions). `load`
+now combines frames with `Y.mergeUpdates`.
+
+Also:
+- `FilePersistence` room ids are encoded with `encodeURIComponent` (reversible,
+  collision-free, traversal-safe) instead of a lossy `[^a-zA-Z0-9._-] -> _`
+  replace that mapped distinct rooms (e.g. `a/b` and `a:b`) onto one log file.
+  Safe ids (UUIDs, room codes) are unchanged, so existing logs keep their names.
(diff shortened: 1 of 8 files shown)


louistrue:
Addressed review feedback (Codex + Greptile P1, legacy filenames): rooms whose id contained special chars (e.g. the URL-path room id `project/model`, stored as `project_model.log`) would have orphaned their logs after the switch to `encodeURIComponent`. `load()` now falls back to the legacy sanitized name, `append()` renames it to the encoded name (one-time migration), `compact()` removes the superseded legacy log, and `drop()` clears both. New regression test covers read-fallback + migration. Safe ids (UUIDs) are unaffected.

PR #1502, opened 2026-07-02
fix(server-client): send auth token on data-model + symbolic fetches

louistrue:
Found via a full multi-agent code review of `main`; verified by reading the code.

## Authenticated servers silently lose properties + annotations
`IfcServerClient` attaches `authHeaders()` (the `Authorization: Bearer <token>` set when `IFC_SERVER_API_TOKEN` is configured) to every request **except** `fetchDataModel` and `fetchSymbolic`. Against a secured server the geometry parse succeeds, but both follow-up fetches get 401 and collapse to `null` after retries — the model renders with no property data and no symbolic annotations, with no visible error. Both now send `authHeaders()` like every other endpoint.

## Streaming cache-check ignored parse options
The `parseParquetStream` cache-check hit `/api/v1/cache/check/${hash}` without the `parseQuery(options)` suffix that the non-streaming `parseParquet` path uses. A cached result for a *different* tessellation quality was therefore returned as a hit. The streaming check now includes `parseQuery(options)`, matching the non-streaming path (the server already keys distinct qualities separately).

Full workspace typecheck clean. Changeset: `@ifc-lite/server-client` patch.

### DIFF
--- a/.changeset/fix-server-client-auth.md
+++ b/.changeset/fix-server-client-auth.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/server-client": patch
+---
+
+fix(server-client): send the auth token on data-model and symbolic fetches
+
+`fetchDataModel` and `fetchSymbolic` were the only requests that omitted
+`authHeaders()`, so against a server started with `IFC_SERVER_API_TOKEN` the
+geometry parse succeeded but the follow-up data-model and symbolic fetches got
+401 and silently returned null — the model loaded with no properties and no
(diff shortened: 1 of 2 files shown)


louistrue:
Addressed review feedback (Codex + Greptile P1): the cache-hit branch called `fetchCachedGeometry(hash)` without options, so `/cache/geometry/:hash` returned default `medium` geometry (or 404) for a non-default tessellation. Now passes `options` through, matching the option-scoped cache-check.

PR #1503, opened 2026-07-02
fix(query): scope whereProperty to the named property set

louistrue:
Found via a full multi-agent code review of `main`; verified by reading the code and pinned with a regression test.

## `whereProperty` ignored the property-set name
`EntityQuery.whereProperty(psetName, propName, operator, value)` stored `psetName` in the filter but `applyPropertyFilters` called `findByProperty(prop, op, value)` and never passed it — and `findByProperty` had no pset parameter. So a property matched in **any** property set: filtering `Pset_WallCommon.IsExternal = true` also returned doors whose `Pset_DoorCommon.IsExternal` matched, silently widening results.

## Fix
`findByProperty` gains an optional `psetName` argument, honored consistently across all three implementations that back a query store:
- `@ifc-lite/data` `PropertyTable` (fresh in-memory loads)
- `@ifc-lite/cache` restored store
- `serverDataModel` (server-converted models)

`whereProperty` now passes the pset name; an unknown pset matches nothing. The parameter is optional, so existing `findByProperty(prop, op, value)` callers are unaffected.

## Tests
- New `entity-query.test.ts` cases: a `Pset_WallCommon`-scoped filter returns only the wall (not the same-named door property), and an unknown pset returns nothing.
- All 112 `@ifc-lite/query` + 51 `@ifc-lite/data` + 25 `@ifc-lite/cache` tests pass; full workspace typecheck clean. No public export added, so the API-surface snapshot is unchanged.

Changesets: `@ifc-lite/data` minor (interface gained an optional param), `@ifc-lite/query` + `@ifc-lite/cache` patch.

### DIFF
--- a/.changeset/fix-query-pset-scope.md
+++ b/.changeset/fix-query-pset-scope.md
@@ -1,0 +1,15 @@
+---
+"@ifc-lite/data": minor
+"@ifc-lite/query": patch
+"@ifc-lite/cache": patch
+---
+
+fix(query): scope `whereProperty` to the named property set
+
+`EntityQuery.whereProperty(psetName, propName, ...)` recorded the property-set
+name but never passed it to `findByProperty`, so a property matched in *any*
+property set — e.g. filtering `Pset_WallCommon.IsExternal` also returned doors
+whose `Pset_DoorCommon.IsExternal` matched. `findByProperty` gains an optional
+`psetName` argument (honored by the in-memory, cache-restored, and
+server-converted property tables), and `whereProperty` now passes it. An unknown
+pset name matches nothing.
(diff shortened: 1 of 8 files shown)


louistrue:
Addressed review feedback:
- **Codex P1 (`as any` in new test)** — fixed properly: typed `createMockStore`'s return as `IfcStoreBase` (`entityMap` is now `IfcEntity`, `getEntitiesByType` narrows with a type guard), so the new pset tests construct `EntityQuery` with no cast. Validated with a `tsc` pass that includes the test files.
- **Greptile P2 (data as minor vs patch)** — keeping **minor**: adding an optional parameter to the public `PropertyTable`/`IfcStoreBase` interface is an additive API change (semver-minor per AGENTS.md 'biggest API change'). The behavioral bug fix itself ships in `@ifc-lite/query` as **patch**, so patch-only query consumers still receive it; a data consumer only benefits from the new arg by explicitly passing it.

PR #1489, opened 2026-07-01
ci(benchmark): commit the first CI-recorded baseline

louistrue:
Follow-up to #1488, closing the P0 loop.

Commits the `benchmark-baseline` artifact from the first `record_baseline` dispatch on main ([run 28545564365](https://github.com/LTplus-AG/ifc-lite/actions/runs/28545564365)): github-actions `ubuntu-latest`, `viewer-benchmark-ci` project (headless Chrome, SwiftShader ANGLE), production build. Each refreshed entry carries that provenance in its `environment` field.

- **FZK-Haus / Snowdon**: refreshed to CI-recorded numbers (the two fixtures the advisory job tracks). The job now diffs like-for-like instead of against 2026-02-21 Apple-Silicon numbers, which were over-reporting by 4-9x on the CI speed class.
- **BWK-BIM / Holter Tower**: unchanged (local stress-run references, not covered by the CI job).

Refresh procedure documented in `tests/benchmark/README.md`.

<h3>Confidence Score: 5/5</h3>

Safe to merge with minimal risk.

The change is limited to benchmark baseline data. The refreshed entries match the current benchmark metric shape, including nullable legacy metrics, and the regression checker compares only numeric metrics with defined thresholds.

No files require special attention.

<details><summary><h3><a href="https://www.greptile.com/trex">T-Rex</a> T-Rex Logs</h3></summary>

**What T-Rex did**
- Reviewed the benchmark baseline file labeled before to understand the initial timestamps and metrics.
- Reviewed the benchmark baseline file labeled after to confirm updated head and Snowdon timestamps, the environment details, and the successful benchmark results.
- Verified that the unchanged-entry hash comparisons for BWK-BIM and Holter Tower base/head are equal in the after artifact.

<a href="https://app.greptile.com/trex/runs/12969802/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=2">View all artifacts</picture></a>

<sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
[...]

PR #1504, opened 2026-07-02
fix(benchmark): re-arm the stale Snowdon mesh guard + flag non-CI baselines

louistrue:
Follow-ups on top of the CI-recorded baseline (#1489, which already fixed the stale local-vs-SwiftShare baseline). **No `baseline.json` change here** — main's #1489 CI baseline is kept as-is.

## Context: the CI "regression" was never real
`baseline.json` used to hold **local Apple-Silicon** numbers compared against **headless SwiftShader** CI (~5x slower speed class + older metric era), so every perf-path PR tripped +50%. Confirmed a measurement artifact, not code:
- Local A/B of current `main` (wasm built from source, headed Chrome): FZK **400ms**, Snowdon **800ms** — baseline speed class.
- 5 unrelated branches all report Snowdon CI `totalWallClock` in a 3000-3600ms band regardless of diff.
- Production telemetry: within-Chrome weekly `ms_per_mesh` is flat.

#1489 fixed the baseline. This PR closes two residual gaps:

## 1. Re-arm the Snowdon geometry-loss guard (resolves greptile P1)
`totalMeshes` is scraped from `"[useIfc] Geometry streaming complete: N batches, M meshes"` = mesh **occurrences**. FZK's `expectedMeshCounts` (317) already matches that occurrence count, but Snowdon's was a stale **1500** (a deduped/legacy value), so the drop-guard was silently inert for Snowdon (17380 always cleared the 1500 floor). Updated to **17380** — verified identical across the #1489 CI baseline, a fresh CI record, and a local Apple-Silicon A/B (2026-07-02) — and documented the occurrence-vs-deduped distinction (PostHog `mesh_count` uses the deduped `allMeshes.length`, ~7151).

## 2. Guard against re-introducing a local baseline
`check-benchmark-regression.js` now warns when a compared model's baseline lacks a CI `environment` tag, so a locally-recorded baseline can't silently reintroduce the false alarm #1489 fixed.

## Related (out of repo)
PostHog per-model alert `pQ4wgqJi` was firing on a single bimodal-by-browser federated model (small 3-sample window). Its query was hardened separately (recent_loads≥5, ≥3 distinct persons/window, ≥3s delta, recent-p25≥baseline-median); validated to suppress the false positive while still able to fire on a real broad regression.

## Verification
- `node --check` + `benchmark:check` run clean; the non-CI-baseline warning does not fire (main's baseline carries CI env tags).
- `totalMeshes` is not one of the six timing thresholds the regression checker compares, so no baseline re-record is needed.
- No changeset — `tests/` + `scripts/` only, no published-package surface changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a warning when benchmark baselines appear to come from a local environment, helping highlight possible metric mismatches in reports and console output.
* **Tests**
  * Updated benchmark expectations for a geometry-correctness case to match the current streamed mesh count and clarified the metric being checked.

### DIFF
--- a/scripts/check-benchmark-regression.js
+++ b/scripts/check-benchmark-regression.js
@@ -73,6 +73,15 @@ function percentIncrease(current, baseline) {
   return ((current - baseline) / baseline) * 100;
 }
 
+// The CI job diffs like-for-like only when the baseline was itself recorded on
+// the CI runner. A locally recorded baseline (fast Apple-Silicon, real GPU, an
+// older metric era) makes every SwiftShader CI run look like a huge regression —
+// which is exactly the false alarm this check is meant to avoid. Flag a baseline
+// entry that carries no CI environment tag so the mismatch is visible, not silent.
+function looksCiRecorded(environment) {
+  return typeof environment === 'string' && /github-actions|swiftshader|ubuntu-latest|\bci\b/i.test(environment);
+}
+
 function formatMs(value) {
   if (typeof value !== 'number') return 'N/A';
   return `${value.toFixed(0)}ms`;
@@ -121,6 +130,7 @@ function compareResults() {
       baselineTimestamp: baselineEntry?.timestamp ?? null,
       baselineEnvironment: baselineEntry?.environment ?? null,
       missingBaseline: !baselineEntry?.metrics,
+      baselineLikelyLocal: !!baselineEntry?.metrics && !looksCiRecorded(baselineEntry?.environment),
       rows: [],
     };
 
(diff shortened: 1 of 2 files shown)


louistrue:
@greptile-apps P1 (stale Snowdon mesh guard) addressed in ccc35d7a: `expectedMeshCounts` for Snowdon updated 1500 → 17380 — the occurrence count `totalMeshes` actually scrapes (verified identical across the #1489 CI baseline, a fresh CI record, and a local A/B). The geometry-loss drop-guard is now armed for Snowdon instead of silently inert. `baseline.json` is unchanged (kept #1489's CI recording).

PR #1506, opened 2026-07-02
fix(ifcx,viewer): guard IFCX path cycles; neutralize list CSV formula injection

louistrue:
Found via a full multi-agent code review of `main`; verified by reading the code and covered with tests.

## 1. IFCX path cycle → stack overflow (crash)
`PathIndex.indexHierarchicalPaths` recursed through each node's `children` with
no ancestor tracking. A malformed/adversarial IFCX layer with a child cycle
(`A -> B -> A`, reachable from external files or collaboration) recursed forever
and overflowed the stack on load. The recursion now carries the set of uuids on
the current DFS branch and skips a child that is already an ancestor. A node
reached by two distinct non-ancestral paths (a diamond) is still indexed under
both. `walkPath` was already bounded by the query segment list and is unchanged.

New `path-resolver.test.ts`: a cyclic layer terminates and still resolves one
hop each way; a diamond indexes the shared child under both parents.

## 2. List CSV export — spreadsheet formula injection (CWE-1236)
`apps/viewer/.../lists/export/csv.ts` quoted delimiters/quotes/newlines but did
not neutralize cells beginning with `=`, `+`, `-`, `@`, TAB or CR, which
Excel/LibreOffice/Sheets execute as formulas. List cells derive from
attacker-controllable IFC values. The `esc` helper now prefixes such cells with
an apostrophe, matching the neutralization the SDK export adapter
(`export-adapter.ts`) already performs.

All 16 `@ifc-lite/ifcx` tests pass; ifcx + viewer typecheck clean. Changeset:
`@ifc-lite/ifcx` patch (viewer is not a published package).

### DIFF
--- a/.changeset/fix-ifcx-path-cycle.md
+++ b/.changeset/fix-ifcx-path-cycle.md
@@ -1,0 +1,12 @@
+---
+"@ifc-lite/ifcx": patch
+---
+
+fix(ifcx): guard PathIndex hierarchical indexing against child cycles
+
+`PathIndex.indexHierarchicalPaths` recursed through a node's children with no
+ancestor tracking, so a malformed IFCX layer with a child cycle (`A -> B -> A`)
+recursed until the stack overflowed and crashed the load. The recursion now
+tracks the uuids on the current DFS branch and skips a child that is already an
+ancestor; a node reached by two distinct non-ancestral paths (a diamond) is
+still indexed under both.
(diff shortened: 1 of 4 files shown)


louistrue:
Addressed review feedback (Greptile P2, BOM bypass): a leading UTF-8 BOM before a formula marker (`﻿=HYPERLINK(...)`) bypassed the guard because the first code point wasn't in the trigger set. The regex now allows an optional leading BOM.

PR #1507, opened 2026-07-02
perf(geometry): rebase + corpus-validate void-cut dedup (stays OFF, finding)

louistrue:
Perf-plan P2. Brings the stalled void-cut dedup drafts (#1293 foundation + #1294 void cache, both ~130 commits behind main) up to current main, adds a purpose-built A/B validation harness, and corpus-validates the flag. **Outcome: the flag stays default OFF — this is a validated finding, not a flip.** Opening it so the rebase, the harness, and the measured reason-not-to-flip are reviewable in one place; supersedes the stale #1293/#1294 (their branches are left untouched).

## What this rebases

- **#1293 foundation** merges into current main with zero conflicts (drops a per-occurrence `rep_identity` re-hash; adds `IFC_LITE_DEDUP_EXTRA`, default OFF).
- **#1294 void-cut dedup** hit 3 conflicts, all from god-file splits that landed on main after the drafts were written. Ported #1294's intent into the current layout:
  - `router/voids.rs` (monolith, deleted on main) → the 4 new methods re-applied verbatim into `router/voids/mod.rs` (`host_translation_only`, `process_element_with_submeshes_and_voids_unplaced`, `definition_signature`, `process_element_with_submeshes_and_voids_deduped`); same `router::voids` module path, so all `self.*` / `super::content_hash::*` calls resolve unchanged.
  - `wasm-bindings/src/api/gpu_meshes.rs` (deleted on main) → cache-injection block re-applied in `gpu_meshes/batch.rs`.
  - `geometry/src/lib.rs` re-export list merged.

## The validation harness

`rust/geometry/tests/dedup_validate.rs` gains `def_dedup_void_ab` (env-driven, `--ignored`): runs the production submesh path with `IFC_LITE_DEF_DEDUP` OFF vs ON per model and reports byte-identity, **order-invariant coord-set identity** (a hard correctness signal independent of emit order), fired-count, and ABAB-interleaved timing medians.

## Corpus result (median of 3 ABAB rounds)

| Model | products | voided hosts | dedup fired | byte-identical | geom divergences | tri delta | speedup |
|---|---|---|---|---|---|---|---|
| AC20-FZK-Haus | 91 | 9 | 3 | 88/91 | 1 | 0 | 1.00x |
| dental_clinic | 2996 | 245 | 45 | 2951/2996 | 5 | -137 | 1.04x |
| advanced_model | 7097 | 82 | 61 | 7036/7097 | 57 | +1876 | 1.06x |
| ISSUE_068 skolebygg | 5274 | 363 | 0 | 5274/5274 | 0 | 0 | 0.97x |

## Why it stays OFF (the finding)

1. **Not byte-identical** on any model where it fires — fails the flip gate outright.
2. **Genuine geometry divergence**, not emit-order noise: order-invariant coord-sets differ on a minority of fired hosts AND triangle totals shift (-137 / +1876). Root cause: the exact CSG kernel quantizes on a snap grid, so cutting a voided element in its small definition frame and then translating is not bit-for-bit the same solid as cutting per-occurrence at world magnitude — the two cross snap-grid thresholds differently on some hosts. (This is the same coordinate-magnitude sensitivity the local-frame precision work addresses; it is why #1294's original "geometrically identical" claim no longer holds after 130 commits of kernel/precision change.)
3. **Marginal-to-negative perf**: 1.00-1.06x where it fires (the void cut is not the dominant cost and pure-translation-identical voided elements are rare), and a **net 0.97x loss** on ISSUE_068 where the signature is computed on 363 hosts but fires on 0 (curved/rotated hosts are ineligible), so the probe is pure overhead there.

Documented in code: the `definition_dedup_enabled()` doc comment (`router/mod.rs`) and `.changeset/void-cut-dedup.md` carry the table and the precise re-flip gate ("only after the def-frame vs world-frame tessellation divergence is closed AND a real perf win is shown").

## Verification

- Foundation + void branches build; `cargo test --workspace --exclude ifc-lite-wasm` green (107 test binaries; export crate 57/0 after fixtures staged); `clippy -D warnings` clean.
- Synthetic `dedup_validate` 6/6 (incl. `void_unplaced_cut_matches_placed_geometry`, `definition_cache_dedups_and_matches_placed_geometry`).
- Kernel determinism manifest pinned + green; zero `.snap.new` (no snapshot regressions).
- The ported `..._deduped` method was independently re-read: it builds the unplaced def-frame cut once, caches by signature, and applies each occurrence's full placement to the shared template — faithful to #1294's design; the divergence is the kernel's coordinate-magnitude sensitivity, not the port.

Changesets included but the behavior is a no-op at runtime (flag default OFF), so this is release-safe as-is.


louistrue:
**Disposition: keep OPEN, do NOT merge** (holding per review).

This PR's value is the *finding* (the void-cut dedup diverges + is perf-marginal on current main) plus the rebase and reusable A/B harness. Two reasons not to land it as-is:

1. **It is dormant + not correct-when-enabled.** The flag stays default OFF and, per the corpus validation, should stay off. Landing ~1000 lines of dedup machinery that we've concluded not to enable adds carrying cost to main for no runtime benefit.
2. **Review found real bugs in the OFF-path code** that would bite anyone who flips it:
   - Greptile P1 / Codex P2: `cached_definition_dedup` is not cleared by `clearPrePassCache`/`setEntityIndex`, so a reused `IfcAPI` (browser worker) carries stale void-cut templates across model swaps.
   - Codex P2: `process_element_with_submeshes_and_voids_deduped` applies `apply_submesh_placement` on top of `try_layered_sub_meshes`, which already applied the element placement — a **double-transform on layered material walls**. This likely accounts for part of the geometry divergence the validation attributed to kernel snap-grid sensitivity, so the root-cause split in the report needs revisiting before any re-attempt.

Recommendation: keep this branch as the reference rebase + harness; revisit the void-cut dedup only after (a) the double-placement and cache-lifecycle bugs are fixed and (b) the def-frame vs world-frame tessellation divergence is closed and a real perf win is shown. The reusable `def_dedup_void_ab` harness is the tool to re-validate with.

louistrue:
Closing: this was a validation finding, not a change to ship. The corpus A/B (in the description) shows the void-cut dedup diverges from the per-occurrence cut on current main and is perf-marginal/negative, and review found OFF-path bugs (definition cache not cleared across model swaps; double-applied placement on layered submeshes). The analysis + the reusable `def_dedup_void_ab` harness are recorded here for whenever the dedup is revisited; no reason to keep the branch open. The reference rebase remains on `perf/dedup-void-rebased` if needed.

PR #1531, opened 2026-07-02
fix(geometry): render IfcSurfaceCurveSweptAreaSolid duct elbows (#1485)

### LINKED ISSUE #1485: Duct elbows geometry appearance
Hi, i just tried to load my HVAC IFC model (IFC4.3 RV) and duct elbows are not appearing at all (in other viewers they do appear correctly, so I assume the bug is not on my side). I can provide my IFC file if needed. Thx. :) 

Image

louistrue:
## Problem

Fixes #1485. Round HVAC duct elbows in an IFC4.3 RV model (Revit export) rendered as **nothing** in the viewer while showing correctly in other viewers.

Root cause: Revit exports round duct bends as `IfcSurfaceCurveSweptAreaSolid` — a circular `SweptArea` profile swept along a trimmed circular-arc `Directrix`. The geometry router **had no processor registered for the type** (it was recognised in `router/layers.rs` but never dispatched), so every elbow body was silently dropped. Verified against the reporter's file: elbow `#1538` produced **0 triangles**; a normal duct segment beside it produced 140.

## Fix

Add `SurfaceCurveSweptAreaSolidProcessor` (registered in the router, so every native + wasm/viewer path picks it up via the shared `GeometryRouter::new` registration):

- Extracts the `SweptArea` as a `Profile2D` (circle, and by construction any profile incl. holes).
- Samples the `Directrix` in 3D, reusing `IfcSweptDiskSolid`'s proven trim handling (composite / polyline / line honour the solid-level Start/EndParam; a trimmed conic carries its own trim).
- Applies the solid `Position`, then sweeps the section with the disk processor's rotation-minimising frame (factored to `pub(crate)`, shared, no behavioural change to `IfcSweptDiskSolid`). Outward-oriented walls + triangulated end caps + smooth normals.
- The **same** processor handles `IfcFixedReferenceSweptAreaSolid` (slots 0..4 are identical; slot 5 differs only in orientation control).

For the circular cross-sections these fittings use, the swept tube is geometrically **exact** (RMF orientation is rotationally invariant). Non-circular sections along a non-planar directrix keep a rotation-minimising roll rather than the exact reference-surface roll — documented, and shape-correct.

## Verification

- **Real file**: all 383 `IfcDuctFitting` elements now mesh (0 empty, was: elbows dropped); the 12 unique swept elbow geometries are instanced via `IfcMappedItem`, so they mesh 12× and cache.
- **Neutral fixture** (`tests/fixtures/issue_1485_duct_elbow_surface_curve_swept.ifc`): a self-contained 90° round elbow reproducing the reported geometry chain — no data from the reporter's model. It produces the identical 788-triangle mesh as the real elbow.
- **Adversarial assertions** in the committed test:
  - Z extent equals the bore diameter exactly (bend authored in XY → proves radius + no out-of-plane twist).
  - In-plane spans ≈ arc radius + bore (proves the sweep follows the 150 mm arc, not a collapsed section).
  - Every vertex lies within one bore radius of the arc centre-line, and the wall reaches the bore radius (proves it's a genuine tube of the right radius/axis).
  - Normals point outward (caught and fixed an inverted-winding bug during review).
  - Both `IfcSurfaceCurveSweptAreaSolid` and `IfcFixedReferenceSweptAreaSolid` variants.

`cargo test -p ifc-lite-geometry` is green (the only failures are pre-existing, environmental — the `wall_opening_cut_regression` tests need the downloaded `advanced_model.ifc` fixture). `cargo clippy -p ifc-lite-geometry --tests` is clean.

### DIFF
--- a/.changeset/duct-elbow-surface-curve-swept.md
+++ b/.changeset/duct-elbow-surface-curve-swept.md
@@ -1,0 +1,6 @@
+---
+"@ifc-lite/geometry": patch
+"@ifc-lite/wasm": patch
+---
+
+Render `IfcSurfaceCurveSweptAreaSolid` and `IfcFixedReferenceSweptAreaSolid` solids. Round HVAC duct elbows — a circular profile swept along a trimmed circular-arc directrix, how Revit exports IFC4.3 duct bends — had no geometry processor registered and were silently dropped from the model. They now mesh as swept tubes (a rotation-minimising frame carries the section along the directrix, exact for the circular cross-sections these fittings use). Fixes #1485.
--- a/rust/geometry/src/processors/mod.rs
+++ b/rust/geometry/src/processors/mod.rs
@@ -50,7 +50,9 @@ pub use extrusion_tapered::ExtrudedAreaSolidTaperedProcessor;
 pub use mapped::MappedItemProcessor;
 pub use sectioned::SectionedSolidHorizontalProcessor;
 pub use surface::SurfaceOfLinearExtrusionProcessor;
-pub use swept::{RevolvedAreaSolidProcessor, SweptDiskSolidProcessor};
+pub use swept::{
+    RevolvedAreaSolidProcessor, SurfaceCurveSweptAreaSolidProcessor, SweptDiskSolidProcessor,
+};
 pub use tessellated::{PolygonalFaceSetProcessor, TriangulatedFaceSetProcessor};
 pub use texture::{build_texture_index, MeshTexture, ResolvedTextureMap};
 
--- a/rust/geometry/src/processors/swept/disk.rs
+++ b/rust/geometry/src/processors/swept/disk.rs
@@ -23,7 +23,7 @@ use crate::router::GeometryProcessor;
 /// by rotating it from `tangents[i-1]` onto `tangents[i]` (the minimum rotation
 /// that aligns them). When consecutive tangents are parallel the frame stays
 /// untouched.
-fn build_tube_rmf(
+pub(crate) fn build_tube_rmf(
     curve_points: &[Point3<f64>],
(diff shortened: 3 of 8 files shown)


louistrue:
### Triage of automated review findings

- **`Option::is_none_or` MSRV (greptile P1):** false positive. The repo pins `nightly-2025-11-15` (rustc 1.93) and `is_none_or` (stable since 1.82) is already used in ~10 shipped locations (`rust/core/src/georef.rs`, `rust/export/*`, `rust/geometry/src/profiles/mod.rs`, …). CI `Rust tests` + `Build packages + WASM` compiled and passed.
- **Trimmed/raw conic directrix vs. solid `StartParam`/`EndParam` (greptile P1 / codex P2, line 123):** the reported elbows use an `IfcTrimmedCurve` directrix whose own Trim1/Trim2 defines the 90° arc — that trim *is* honored, so the sweep is correct and the redundant solid-level params are a no-op. A raw (untrimmed) `IfcCircle`/`IfcEllipse` directrix bounded only by the solid params is in no known file and matches the existing `IfcSweptDiskSolid` processor's documented limitation (this processor deliberately mirrors its directrix sampling).
- **Non-circular orientation from attr 5 (codex P2, line 155):** documented in the processor. The rotation-minimising frame is exact for the circular cross-sections these fittings use (100% of firing cases here) and correct for planar directrixes generally. A reference-surface/fixed-reference roll rewrite has no ground-truth test case and could ship a 90°-swapped rectangular section, so the honest documented limitation is preferred until a real non-circular fixture exists.

Core fix is verified against the reporter's file (all 383 duct fittings mesh; elbows went 0 → 788 tris) and a neutral committed fixture with adversarial tube-shape + outward-normal assertions. All 21 checks green.

PR #1538, opened 2026-07-02
fix(viewer): Export Changes exports all changed models, not just the first (#1534)

### LINKED ISSUE #1534: Export changes in multiple models exports one model only.
I needed to change a property in 4 models that are loaded in IfcLite.
The 'Export changes' only sees one change in one model and exports only one model. The first one.

Image

louistrue:
Closes #1534.

## Problem
With several models loaded, editing a property in each one and clicking **Export Changes** only counted one change and exported a single model (the first). The report: *"The 'Export changes' only sees one change in one model and exports only one model. The first one."*

## Root cause
`ExportChangesButton` derived its model from `models.values().next().value` — the **first** federated model — and both the badge count and the export operated on that single model.

## Fix
Aggregate across **every** changed model, behind two small, unit-tested modules:

- **`lib/export/model-changes.ts`** (pure) — `collectChangedModels(state)` is the single source of truth for *what changed*: per-model modified-entity count + a georef-only `+1` + a single resolved schedule target. Its per-model `changeCount` mirrors the store's `getModifiedEntityCount()` exactly (locked by an equivalence test). `buildChangedArtifacts()` turns that into one artifact per model, sequentially and each individually `try/catch`ed.
- **`lib/export/changed-model-export.ts`** (IO) — the real per-model exporters (`StepExporter` for STEP schemas, `Ifc5Exporter` for IFC5), byte-for-byte parity with `ExportDialog`'s single-model branch. Injected as deps so the pure module and its tests never load the renderer or the store barrel.
- **`ExportChangesButton.tsx`** — badge = sum of all changed models' counts; the tooltip and toast report the multi-model case; keeps rendering during an in-flight export.

### Behavior
- **1 changed model** → a single bare `.ifc` / `.ifcx` download (unchanged from today).
- **N changed models** → **one** `ifc-lite-changes_<date>.zip` bundling each model's file (deduped entry names; fflate async `zip`), avoiding the browser multi-download prompt.
- **IFC5** models export as `.ifcx` (previously they collapsed to a mislabeled `.ifc`); a mixed batch puts both in the same zip.
- **Schedule** is spliced into exactly one *splice-capable* STEP model, with its source id rewritten to the resolved target so the splice gate always matches — never duplicated across files, never counted-but-silently-dropped.
- **Native / unhydratable** models are skipped with a reason instead of aborting the batch.

Also relocates the pure `countGeneratedTasks` helper to `schedule-edit-helpers` (re-exported from `scheduleSlice`, so `@/store` consumers are unchanged) to keep the new export module free of the store's runtime graph.

## Design + review
Approach was designed via an adversarial judge-panel and the diff was put through an adversarial find→verify review. That review caught a **major** self-introduced defect — a schedule attributed to a removed or IFC5 source model was counted in the badge yet silently dropped from every file — which is fixed here (`resolveScheduleTarget` now only ever targets a splice-capable model, and the splice source id is realigned to that target). Two findings were verified as **pre-existing and out of scope** for #1534 (noted below, not changed):
- `ExportDialog`'s "N entities modified" banner reads the all-models count while its single-model export writes one.
- `removeModel` leaves orphaned mutation views/georef entries (and a dangling schedule source) in the store; this fix tolerates that safely (a dangling schedule source resolves to no target, so it is neither counted nor claimed).
- The SDK `export.ifc` adapter is single-model **by contract** (throws on >1) — a separate feature gap.

`MergedExporter` was deliberately **not** used: it flattens N models into one re-based `.ifc` and can't do IFC5 — wrong semantics for "give me my edited models back."

## Testing
- New `model-changes.test.ts` (23 assertions): multi-model collection, the `getModifiedEntityCount` equivalence lock, register-only-view exclusion, georef-only counting, schedule attribution (declared source / null source / removed source / IFC5 source / untouched), legacy XOR federated, filename dedup, IFC5 routing, partial-success and throw continuation, and the schedule single-splice / source-rewrite guard.
- New `changed-model-export.integration.test.ts`: drives the **real** `StepExporter` + `MutablePropertyView` for two separately-edited models and asserts each `.ifc` carries only its own edit, then round-trips the zip back to both files.
- Full viewer suite green: **1275 passed, 0 failed, 2 skipped**; `tsc --noEmit` clean.

No changeset: changes are confined to `apps/viewer` (not a published `@ifc-lite/*` package).

## Summary by CodeRabbit

* **New Features**
  * Export Changes now includes all edited models, not just one, and can download multiple exports together as a ZIP.
  * Export badges and tooltips now reflect the total number of changed models more accurately.

* **Bug Fixes**
  * Exporting no longer disappears mid-process when counts change.
  * Better handling for exports that partially fail or need to skip some models, with clearer success/error feedback.
  * Schedule-related edits are now handled more reliably during export.

### DIFF
--- a/apps/viewer/src/components/viewer/ExportChangesButton.tsx
+++ b/apps/viewer/src/components/viewer/ExportChangesButton.tsx
@@ -91,6 +91,7 @@ export function ExportChangesButton({ className }: ExportChangesButtonProps) {
       if (files.length === 0) {
         if (skipped.length > 0) {
           setExportStatus('error');
+          setTimeout(() => setExportStatus('idle'), 3000);
           toast.error(`Export failed: ${skipped[0].reason}`);
         } else {
           toast.info('No changes to export');
@@ -139,7 +140,7 @@ export function ExportChangesButton({ className }: ExportChangesButtonProps) {
   const tooltip =
     modelCount > 1
       ? `Export changes in ${modelCount} models (${totalCount} changes)`
-      : `Export IFC with ${totalCount} property change${totalCount === 1 ? '' : 's'} applied`;
+      : `Export IFC with ${totalCount} change${totalCount === 1 ? '' : 's'} applied`;
 
   return (
     <Tooltip>
--- a/apps/viewer/src/components/viewer/ExportChangesButton.tsx
+++ b/apps/viewer/src/components/viewer/ExportChangesButton.tsx
@@ -3,175 +3,123 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 /**
- * Dedicated button for exporting IFC with property mutations applied.
- * Shows when there are pending changes and provides one-click export.
+ * Dedicated toolbar button for exporting IFC with pending changes applied.
+ * Shows when any loaded model has pending changes and exports ALL of them in
+ * one click: a single `.ifc` (or `.ifcx`) when one model changed, or a single
+ * zip bundling every changed model's file when several did (issue #1534 — this
+ * used to look at only the first federated model for both the badge and the
+ * export).
  */
 
-import { useState, useCallback, useMemo, useEffect } from 'react';
+import { useState, useCallback, useMemo } from 'react';
 import { Download, Loader2, Check, AlertCircle } from 'lucide-react';
+import { zip, strToU8 } from 'fflate';
 import { Button } from '@/components/ui/button';
 import { Badge } from '@/components/ui/badge';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
-import { useViewerStore, countGeneratedTasks } from '@/store';
-import { configureMutationView } from '@/utils/configureMutationView';
-import { StepExporter } from '@ifc-lite/export';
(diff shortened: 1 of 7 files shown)


louistrue:
Addressed the automated review feedback in e805b3fc:

- **Codex — georef edits when a view exists** (`model-changes.ts`): `collectChangedModels` now counts a georef edit (`+1`) even when the model already has a (possibly inspection-only) mutation view, so a georef-only edit on an already-viewed model is no longer excluded from the export set. This is a deliberate, documented over-count vs `getModifiedEntityCount` in the rare property-and-georef case; new test covers a zero-count view + georef.
- **Greptile — error state not reset in the all-skipped path** (`ExportChangesButton.tsx`): added the same `setTimeout(() => setExportStatus('idle'), 3000)` the `catch` block uses, so the error icon no longer sticks.
- **Greptile — tooltip copy vs the count metric** (`ExportChangesButton.tsx`): single-model tooltip now reads "N change(s) applied" (the badge is modified-entity granularity), matching the multi-model copy.

Full viewer suite green (1275 passed / 0 failed / 2 skipped), `tsc --noEmit` clean.

PR #1539, opened 2026-07-02
fix(viewer): keep the modified-entity count honest with what actually exports

louistrue:
Follow-up to #1538 — resolves the two pre-existing consistency items noted there. Both make the store's `getModifiedEntityCount()` (and the counts derived from it) agree with what is actually exportable/written.

## Fix 1 — `removeModel` orphaned mutation-view leak (`store/slices/modelSlice.ts`)
Removing a model deleted it from `models` but left its mutation view, georef edits, undo/redo stacks, `removedNewEntities`, and any schedule it owned behind in the store. Consequences:
- `getModifiedEntityCount()` kept counting a model that can no longer be exported (so the toolbar badge and the Export dialog banner could disagree).
- A schedule whose source model was unloaded left `scheduleSourceModelId` dangling.

`removeModel` now discards that footprint before dropping the model, reusing the existing, separately-tested actions: `clearMutations(modelId)` (empties the view — `view.clear()` resets `mutationHistory`, and clears undo/redo/dirty/georef and an **owned** schedule) then `clearMutationView(modelId)` (deletes the now-empty view entry so the count stops iterating it). Cross-slice via `get()`, matching the `mutationSlice` idiom. `removeModel`'s only entry points are the hierarchy panel (user-initiated) and `useIfcFederation.removeModel`, which wraps this store action, so fixing it here covers every path.

Note: removing a model with unsaved edits now discards those edits — which is correct, since a removed model's overlay was already unexportable (it isn't in `models`); previously it merely leaked into the count.

## Fix 2 — ExportDialog pending-changes banner scope (`components/viewer/ExportDialog.tsx`)
`modifiedCount` used the all-models `getModifiedEntityCount()` but was rendered in the "N entities have been modified" banner **and** baked into the single-model STEP export's `FILE_DESCRIPTION` ("Exported from ifc-lite with N modifications") even when `scope === 'single'` writes only the selected model. It was also memoized on a stable action ref, so it never refreshed.

Now: for `single` scope the count reflects the selected model — its modified entities **plus** the schedule tasks a single-model export splices in (same gate as `spliceScheduleIntoExport`, so the banner matches what's written); `merged` keeps the aggregate. Subscribes to `mutationVersion` + the schedule fields so it updates reactively.

## Review + testing
Approach was run through an adversarial find→verify review: Fix 1 came back clean; Fix 2 surfaced a self-introduced under-count (the single-model count initially dropped the spliced schedule tasks), fixed here before this PR.

- New `modelSlice.test.ts` case asserts `removeModel` discards the model's mutation footprint (invokes `clearMutations` + `clearMutationView`).
- Full viewer suite green: **1276 passed / 0 failed / 2 skipped**; `tsc --noEmit` clean.

No changeset: changes are confined to `apps/viewer`.

### DIFF
--- a/apps/viewer/src/components/viewer/ExportDialog.tsx
+++ b/apps/viewer/src/components/viewer/ExportDialog.tsx
@@ -83,6 +83,7 @@ export function ExportDialog({ trigger }: ExportDialogProps) {
   const scheduleData = useViewerStore((s) => s.scheduleData);
   const scheduleIsEdited = useViewerStore((s) => s.scheduleIsEdited);
   const scheduleSourceModelId = useViewerStore((s) => s.scheduleSourceModelId);
+  const georefMutations = useViewerStore((s) => s.georefMutations);
   const hiddenEntities = useViewerStore((s) => s.hiddenEntities);
   const isolatedEntities = useViewerStore((s) => s.isolatedEntities);
   const hiddenEntitiesByModel = useViewerStore((s) => s.hiddenEntitiesByModel);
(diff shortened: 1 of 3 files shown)


louistrue:
Addressed the review feedback in 3493eb2d:

- **Codex — georef edits in single-model count** (`ExportDialog.tsx`): the `single` scope count now adds `+1` when the selected model has georeferencing edits, matching what the single-model STEP export writes. Subscribed to `georefMutations` so it refreshes.
- **Codex + Greptile — schedule survives model removal** (`modelSlice.ts`): `clearMutations` only clears a schedule whose `source === modelId`. `removeModel` now also clears any orphaned generated schedule when the **last** model is removed (the common null/dangling-source case), so it stops inflating `getModifiedEntityCount` with no model left to own it. New `modelSlice` tests cover both the last-model clear and the "other models remain → keep" case.

Residual (deliberately out of scope): an **edited parsed** schedule (`scheduleIsEdited`, all tasks already persisted) with an *unattributed* source in a still-multi-model session isn't torn down here. The schedule is global singleton state whose full lifecycle belongs to the schedule slice; this PR strictly improves cleanup (previously `removeModel` cleared nothing) without taking on that broader teardown. Happy to file a follow-up if you'd like it fully reset on owner removal.

Full viewer suite green (1277 passed / 0 failed / 2 skipped); `tsc --noEmit` clean.

PR #1548, opened 2026-07-03
test: audit sweep - prune useless tests, add high-value coverage, fix 3 surfaced bugs

louistrue:
## Summary

A full audit of the test suite (**920 Rust + 436 TS test files, ~153k LOC**) against the AGENTS.md test-quality rubric, then acting on the findings. Net **+102 tests / -13 useless tests**, and the new tests surfaced **3 real production bugs** (now fixed).

Method: 70 auditors mapped the whole surface; every removal candidate was then **adversarially verified** to lose zero coverage (3 flagged tests were *kept* after that review); every new test was written and run green.

## Removals (each verified, zero coverage lost)
- **No-assertion / tautology / self-comparison**: `ar_wall` geometry-type probe, `router` coordinate-system analysis, `mesh` IEEE-754 precision demo, sdk sandbox self-compares, codegen `console.log`-only tests, renderer type-assignability.
- **Dead / misleading-green**: `test_tekla.rs` (fixture absent from the manifest, always skipped), and the unmaintained CDN-dependent `benchmark.spec.ts` + `holter-tower-debug.spec.ts` diagnostics, plus their now-dead playwright projects and `package.json` scripts.
- **Redundant** (subsumed by a stronger sibling): `issue_635` per-wall round-window tests, space-interaction Shift duplicate, bcf viewpoint dup, sdk `IDSNamespace` dup.
- **Replaced two fake tests with real ones**: `transform::parse_direction` (was an empty stub) and sdk `on()` delegation (was a `typeof` check).

_Kept after verification_ (rubric flagged them but they guard real behavior): `modelSlice.getModel` miss, lens `executionTime` sign check, router `opening_projection_strategy`. The `tests/integration.test.ts` suite was **not** removed: it does run in CI (`pnpm fixtures` stages its models); the auditor was misled by the local unstaged-fixture trap.

## Additions (102 tests, 19 files)
Untested security guards, data-loss paths, and regression-prone geometry: FederationRegistry N=0/1/many + gap-returns-null, `PropertyTable.findByProperty` operator matrix + pset scoping, IDS unit-scale conversion, `crc32` web-ifc compatibility contract, sandbox module-syntax stripping, CSV bulk-import + bulk-query operators, BCF entity/comment round-trip + Lines/ClippingPlanes, lens material-name preference, collab shared-presence dispose contract, viewer + lists CSV formula-injection guards, geometry placement parsing + `mesh_to_tris` panic-safety + instanced-decoder robustness + non-planar void routing, gltf visibility filters, ffi site-local normalization, clash cross-group dedup.

## Bugs surfaced by the new tests (fixed here)
- **@ifc-lite/sandbox**: `import * as x` was mangled to `import * from x` on the esbuild-free transpile path, sending a broken import to QuickJS.
- **@ifc-lite/bcf**: XML entities never unescaped on read, every comment truncated to `''` (nested `<Comment>` tag-name collision), and `BimSnippet`/`DocumentReference` parsed but never written (dropped on export). Reader also hardened for foreign schema ordering (comments before `<Viewpoints>`), with a synthetic regression test.
- **ifc-lite-geometry**: instanced-shard decoder could OOM-abort on a hostile header count before validating buffer length.

Changesets added for the two published packages (`@ifc-lite/sandbox`, `@ifc-lite/bcf`, both patch). Also made `collab/round-trip.test.ts` skip cleanly per AGENTS.md instead of throwing on an unstaged fixture.

## Verification
- Rust: `cargo test` green for `geometry`, `export`, `ffi`, `clash` (needed fixtures staged selectively).
- TS: full suites green for every touched package; `typecheck` clean (incl. viewer).
- CI gates: `check:test-wiring` ✅, `check:api-surface` unchanged ✅ (fixes are internal, no surface change).

## Not done (reported for follow-up, out of scope here)
- Medium/low-priority coverage gaps from the audit (large backlog) and a few structural flags: `escapeCsv` is duplicated across 4 files with no shared home; the `rust/export` `fixture()` helpers panic instead of skipping when fixtures are unstaged (34 tests, CI-green but violates the skip rule); dead surface in `fast_parse.rs` / `query/fluent-api.ts` / codegen DERIVE-INVERSE parsing.


louistrue:
Thanks for the review. Addressed all three P2 findings in 525895a8:

- **BCF 3.0 `<Comments>` wrapper** and **vendor-extension elements** (Codex + Greptile): replaced the fragile terminator-allowlist lookahead in `parseComments` with a structural parse. It slices each `<Comment Guid="...">` wrapper's span to the next wrapper (or end of markup) and takes the last `</Comment>` as the real close, so no comment is dropped regardless of what follows it. Added reader regression tests for both the 3.0 container and a `<RevitExtensions>`-followed comment.
- **BimSnippet ReferenceSchema** (Codex): `writeBimSnippet` now only emits when `ReferenceSchema` is present (required by the BCF XSD), so exports can't be schema-invalid.

bcf suite 97/97, downstream clash bridge 14/14, typecheck clean.

PR #1554, opened 2026-07-03
fix(viewer): world-context TOP/BOTTOM read north-up, not the IfcSite axes (#1532)

### LINKED ISSUE #1532: 3D world context North is pointing down (not really a problem, just curious)
Top view is oriented 'up-side-down'. North is pointing down and South is up. Is this on purpose or is this driven by something in the model?

Image

Image

This is not a problem at all because georeferencing works really good and I can just click a few times on the viewcube and North is pointing up. Just wondering if there is a reason for it.

louistrue:
## Summary

Fixes #1532: in the Cesium "3D world context", the TOP view showed geographic **north pointing down** for georeferenced models.

## Root cause

The default TOP/BOTTOM presets orient the camera to the building's local axes via `buildingRotation` (the IfcSite placement's Z-rotation, `atan2(siteX.y, siteX.x)`) -- the floor-plan / drawing convention. The reported model's IfcSite is placed rotated ~180 degrees relative to the RD grid, so on the north-aligned basemap the plan view followed the building and put north at the bottom. Georeferencing itself is fine (footprint placement uses `IfcMapConversion`, independent of the site rotation), which is why placement looked correct but north was down.

Verified against the real Cesium camera math (`Camera.updateMembers` orthonormalization) headlessly: `buildingRotation=0` gives screen-up = North; `buildingRotation=180deg` gives screen-up = South.

An intuitive-looking alternative -- that the bridge feeds a degenerate world-Y "up" for the straight-down view -- was tested and **ruled out**: Cesium re-orthonormalizes `up` against the view direction, so raw world-Y already resolves to North.

## Fix

When the world-context basemap is active, drop `buildingRotation` for TOP/BOTTOM so they read north-up like a map. Side views still face the building's front, and with the basemap off everything stays building-aligned (plain-viewer plan views unchanged). Applied at both preset entry points (the viewcube/command wrapper and the `1`/`2` keyboard keys) via a shared, tested pure helper `presetViewRotation`.

## Behavior

- Cesium on + TOP/BOTTOM -> geographic north up.
- Cesium off, or side views -> unchanged (building-aligned).
- Minor: toggling the basemap on while already in TOP needs a re-click to re-orient (the rotation is chosen at preset time).

## Tests

- New unit test for `presetViewRotation` (basemap on/off x top/bottom/side x defined/undefined rotation).
- Viewer typecheck clean.

## Out of scope (noted for follow-up)

`buildModelMatrix` (the GLB placement) does not apply the meridian-convergence gamma that the camera bridge does, so the 3D footprint can sit up to ~gamma off true north on the basemap (sub-degree for RD, larger for oblique projections like Krovak). Separate from this 180-degree issue.

### DIFF
--- a/apps/viewer/src/components/viewer/Viewport.tsx
+++ b/apps/viewer/src/components/viewer/Viewport.tsx
@@ -11,6 +11,7 @@ import { Renderer, type VisualEnhancementOptions, type LightingEnvironment } fro
 import type { MeshData, CoordinateInfo, PointCloudAsset } from '@ifc-lite/geometry';
 import { useViewerStore, resolveEntityRef, type MeasurePoint, type SnapVisualization } from '@/store';
 import { LIGHTING_PRESETS } from '@/lib/lighting-presets';
+import { presetViewRotation } from '@/lib/preset-view-orientation';
 import { sunLightingForAltitude } from '@/lib/geo/solar-direction';
 import {
   useSelectionState,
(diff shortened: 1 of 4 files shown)


louistrue:
Adversarial DRY + correctness review done (CodeRabbit/Codex were rate-limited; Greptile passed clean with no comments). Both independent reviewers returned **ship**, no blockers.

Acted on the findings worth addressing:
- **Robustness (idiomatic signal):** gate on `cesiumEnabled && cesiumAvailable` (a basemap that is actually rendering) instead of the bare toggle, matching the app's own `cesiumActive` signal. Prevents a stale `cesiumEnabled` after georef disappears from forcing north-up. `cesiumEnabled` also resets on model load.
- Consolidated the duplicate `@/store` import to the inline-`type` form used elsewhere.
- Added the symmetric `bottom` + undefined-rotation test assertion.

Left as-is (documented, not defects): floorplan views over the basemap read north-up (consistent with the chosen "world-context = north-up" behavior); the north-up flip is applied at preset time, not reactively on toggling the basemap.

Rebased onto latest main (no conflicts, no lockfile drift) and re-running CI.

PR #1579, opened 2026-07-03
perf(core): memchr2 the STEP entity-end scan + a unified perf diagnostic kit

louistrue:
## What

Two things, one investigation:

1. **The lever.** The STEP scanner's `find_entity_end` walked every byte of each entity record to find the terminating `;`. Profiling showed it's the single hottest structural-scan function: it runs per entity through **both** `build_entity_index` **and** the processor scan loop (they share `EntityScanner`), and parse is **single-threaded**, so it gates time-to-first-geometry (and is paid N× via per-worker re-decode in the browser). Rewritten to `memchr2('\'', ';')` — the overwhelming majority of records are string-free geometry primitives (`#7=IFCCARTESIANPOINT((1.,2.,3.));`), so the terminator is found in one vectorized hop.

2. **The kit that found it** (`scripts/perf/` + `rust/processing/examples/perf_probe.rs`): a unified native probe that drains the per-phase timings `ProcessingStats` already publishes (parse vs geometry, with an isolated `build_entity_index` scan + CSG census), a samply flamegraph wrapper, and a README tying the native probe to the WASM viewer telemetry plus a lever ledger (shipped wins + dead-ends).

## Why it matters

"CSG is 85-90% of load" only holds for steel/MEP. Parse scales with entity **count**, so on entity-heavy architecture it dominates: schependomlaan (47 MB, 714k entities) is **82% parse, 18% geometry**. This is the most user-facing lever — it speeds up every model's load and first-paint, native and browser.

## Numbers (best-of-N, `profiling` profile)

| Fixture | isolated scanner walk | parse phase | total load |
|---|---|---|---|
| schependomlaan 47 MB | 56.5 → 36.1 ms (**-36%**) | -14% | -8% |
| BIMcollab 17 MB | 19.9 → 10.4 ms (**-48%**) | -30% | **-19%** |
| Smiley 5.8 MB | 7.8 → 4.5 ms (**-42%**) | -28% | -12% |
| MEP 218 MB | 213 → 94 ms (**-56%**) | -16% | -10% |

## Correctness

**Byte-identical output** — semantics unchanged (first `;` outside a quoted string; doubled `''` is an escaped in-string quote per ISO 10303-21). Verified:
- `mesh_determinism::mesh_output_matches_pinned_manifest` — passes with **no manifest re-pin**
- `gltf::export_is_byte_deterministic`, `bounded_glb_with_index_is_byte_identical` — pass
- full `ifc-lite-core` / `-processing` / `-export` suites green; malformed-input edge cases (`truncated_missing_semicolon`, `unterminated_string`, `random_byte_sweep`) green
- clippy clean; compiles for `wasm32-unknown-unknown`; wasm bundle rebuilds with no committed `.d.ts` drift

## Notes

- `[profile.profiling]` added to the root Cargo.toml (release opt + symbols + panic=unwind) so samply symbolicates and per-element `catch_unwind` still fires.
- Follow-on parse levers this kit surfaced (not in this PR): the double file-walk (`build_entity_index` + scan loop both run the scanner) and `IfcType::from_str`'s per-call `to_uppercase()` allocation.

## Summary by CodeRabbit

* **Performance**
  * Improved STEP file parsing speed, especially on larger models, with no change to output results.

* **New Features**
  * Added a profiling build mode for gathering performance data with optimized binaries and debug symbols.
  * Added a new performance probe utility for measuring parse and processing timings.
  * Added helper scripts to run the probe and generate flamegraphs.

* **Documentation**
  * Added guidance for using the performance toolkit and interpreting timing results.

### DIFF
--- a/.changeset/perf-memchr2-step-scan.md
+++ b/.changeset/perf-memchr2-step-scan.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Faster STEP parsing: the entity scanner's `find_entity_end` now uses a SIMD `memchr2` scan for the record terminator instead of a per-byte loop. It runs on every entity of every model, through both the entity-index build and the processor scan loop (they share the scanner), and parse is single-threaded, so it directly reduces time-to-first-geometry. Measured: isolated scanner walk -36% to -64%, full parse phase -12% to -30%, total load -8% to -19% across a range of models (small architecture to a 218 MB MEP model). Output is byte-identical (no mesh-determinism manifest change).
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,17 @@ strip = true
 inherits = "release"
 panic = "unwind"
 
+# Profiling profile: release-grade optimization with symbols retained and
+# panic=unwind, so `samply record` on a profiled binary produces a symbolized
+# flamegraph and per-element `catch_unwind` isolation still fires. Used by the
+# `perf_probe` harness and `scripts/perf/flame.sh` (see `scripts/perf/README.md`).
+# `strip`/`debug` override the release defaults; the rest is inherited.
+[profile.profiling]
+inherits = "release"
+strip = false
+debug = "line-tables-only"
+panic = "unwind"
+
 
 [profile.dev.package.bnum]
 opt-level = 3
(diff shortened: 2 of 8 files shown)


louistrue:
Addressed both review notes in b66bfeda:
- **probe.sh (CodeRabbit, bash 3.2)**: the optional `FEATURES` array is now expanded with `${arr[@]+"${arr[@]}"}` so `set -u` doesn't abort on macOS's default bash 3.2 when `OBS` is unset. Verified under bash 3.2.57.
- **perf_probe.rs (Greptile, magic numbers)**: the raw `0..3` CSG-op literals in `summarize_census` are now named `OP_*` constants documenting the census contract.

Issue #1591, opened 2026-07-03
Feature request | regex in query, lists from multiple projects or sites/buildings


steverugi:
### The problem

I am now finally trying ifclite, looks very promising

Question 1: I need to pull netvolume quantity from all elements, if I use Qto_WallBaseQuantities and Qto_SlabBaseQuantities it creates two distinct columns, in Bonsai I use NetVolume and /Qto_.*BaseQuantities/ to achieve that, can I use it in ifclite as well?

Question 2: I am very happy to be able to load multiple projects, it's so easy and with georeferencing is a real game-changer
Now I want to create a list of elements from them, I'd like to have a column to identify their IfcProject or IfcSite or similar, how do I do it in ifclite?

thanks a lot for sharing and continuously updating your app, it's just amazing

### Proposed solution

allow use of regex and option to click on project, site, building, etc
unless I am missing something columns do not include them:

<img width="977" height="110" alt="Image" src="https://github.com/user-attachments/assets/df219e1b-27e2-4fad-b524-28c19d920aaf" />

steverugi:
@louistrue if you like together with having a static set of criteria to select the attribute/property from, a double box where the user could enter them (Bonsai style) would be super useful. But I'd understand if the app needs to be basic for a targeted audience

thanks again

steverugi:
@louistrue thank you so much for the upgrade, incredibily helpful!

one little thing, if possible
'container' allows in Bonsai to extract the immediate container the element is contained in
it's useful when spatial decomposition in bridges or roads for instance

thanks again for your fantastic work, much appreciated
PS I just became a monthly donor 😃

steverugi:
@louistrue thanks for the update
I tried REGEX and it works beautifully!
do you think in the future the formula could be edited after running the query? now I need to delete and add the column again
your work is much, much appreciated

louistrue:
@steverugi both of your follow-ups just shipped and will be live on ifclite.com shortly:

- **Edit a column in place**: each column now has a pencil affordance that reopens the editor pre-filled; saving updates the definition in place (order, widths and sorting are preserved - including when the sort is on the edited column) and re-runs the list. Custom labels survive edits, and the duplicate guard now also applies while editing.
[...]

PR #1614, opened 2026-07-04
feat(lists): federation-identity columns (model + leveled spatial) [#1591]

louistrue:
## What & why

Part of issue #1591. When several IFC models are federated in the viewer, a list runs once per model and flattens every model's rows into one table. Today the only federation datum a row carries is an internal `modelId`, never a visible column, so you cannot tell which file (or project / site / building) a row came from.

This adds federation-identity columns to the Lists feature. Because grouping, sorting, filtering, and CSV/xlsx/PDF export are all generic over columns, adding the column is all it takes to also get "group by source project", "sort by site", etc. for free, which is exactly what the issue asks for.

## Columns added

- **Model** (`source: 'model'`) resolves the source file / model name (`FederatedModel.name`) - the robust federation discriminator, unique per loaded file.
- **Project / Site / Building** via a now-leveled **spatial** source (`propertyName` selects `Storey` (default), `Building`, `Site`, or `Project`). Storey already existed and is unchanged.

All are addable as quick-add chips in the column picker and as filter dimensions (leveled `Spatial` + a `Model` filter, with value suggestions).

## Implementation

- `packages/lists` (published, additive): `ColumnDefinition.source` / `PropertyCondition.source` gain `model`; a `getSpatialValue` helper resolves the leveled spatial value and is shared by column extraction and condition filtering (so columns and filters can never disagree). `ListDataProvider` gains optional `getModelName` / `getProjectName` / `getSiteName` / `getBuildingName`. Storey-only columns keep resolving the storey name (back-compat with persisted lists).
- viewer: `buildSpatialAncestryIndex` in `utils/spatialHierarchy.ts` precomputes each element's containing Site / Building plus the model's Project in a single DFS (O(1) per lookup, cached per model). Container names resolve via `entities.getName` so an unnamed container yields `''` (matching the storey column), and "Building" spans building-like IFC4X3 facilities (`IfcBridge` / `IfcRoad` / ...). The per-model list provider (built in `adapter.ts`) implements the new getters; `ListBuilder` gains the chips and filters.

The single canonical per-model load path is untouched; this only reads already-parsed spatial hierarchy + the existing models map.

## Tests

- `@ifc-lite/lists` engine: new column extraction (model + Project/Site/Building/Storey), the level-less back-compat default, and condition filtering by leveled spatial + model. 45/45 pass.
- `buildSpatialAncestryIndex`: contained element, IFC4X3 facility at building level, unnamed container yields `''`, unplaced element yields `''`, and parts reachable only via the storey reverse index. 15/15 in the util suite.
- Full repo typecheck green; viewer typecheck green.

## Changeset

`@ifc-lite/lists` minor (additive API).

## Follow-ups (rest of #1591, stacked on this branch)

Regex/pattern set-name matching; source-model badges in Clash + IDS; BCF federated round-trip.

## Summary by CodeRabbit

* **New Features**
  * Lists can now show and use federation-identity fields, including model name and spatial levels for project, site, building, and storey.
  * Spatial columns and filters now support choosing the level they refer to, with improved grouping, sorting, filtering, and export behavior.

* **Bug Fixes**
  * Spatial values now resolve more consistently across different hierarchy levels, including fallback behavior for existing storey-based columns.
  * List results now better reflect the originating model and location for federated content.

### DIFF
--- a/.changeset/federation-identity-list-columns.md
+++ b/.changeset/federation-identity-list-columns.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/lists": minor
+---
+
+Add federation-identity list columns: a `model` source (source file / model name) and a leveled `spatial` source whose `propertyName` selects `Storey` (default), `Building`, `Site`, or `Project`. Lets a list over several federated models be grouped, sorted, filtered, and exported by which project/site/building/file each row comes from (issue #1591). `ListDataProvider` gains optional `getModelName` / `getProjectName` / `getSiteName` / `getBuildingName` accessors; existing storey-only `spatial` columns keep resolving the storey name.
--- a/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
+++ b/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
@@ -20,7 +20,7 @@ import { ComboInput } from '@/components/ui/combo-input';
 import { ScrollArea } from '@/components/ui/scroll-area';
 import { Badge } from '@/components/ui/badge';
 import { cn } from '@/lib/utils';
-import { IfcTypeEnum } from '@ifc-lite/data';
+import { IfcTypeEnum, isBuildingLikeSpatialType, type SpatialNode } from '@ifc-lite/data';
 import type { IfcDataStore } from '@ifc-lite/parser';
 import {
   discoverFilterValues,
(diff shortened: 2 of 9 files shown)


louistrue:
@coderabbitai full review

louistrue:
Both CodeRabbit findings are addressed in f32a00de:
- **Same-type nesting (correctness):** `buildSpatialAncestryIndex` no longer inherits an ancestor site/building name across same-type nesting. A site/building node now takes its own name (or '' when unnamed), so an unnamed nested IfcSite resolves to '' (consistent with the storey column), while the nearest *named* nested site still wins. Regression-tested both ways.
- **Duplicate traversal (DRY):** extracted `spatialContainerLevel` + `collectSpatialContainerNames` in `spatialHierarchy.ts`; the ListBuilder spatial-filter suggestions now reuse the collector instead of re-walking with a duplicated classification.

Issue #1623, opened 2026-07-05
Geometry: IfcMappedItem models materialize every occurrence — 43M tris from ~9M unique; need an instanced output path


Blogbotana:
## Summary

For models where geometry is shared via `IfcMappedItem` (mapped onto a small set
of `IfcRepresentationMap`s), the streaming geometry pipeline **materializes every
occurrence into a full flat mesh**. On an instance-heavy model this blows the
vertex output up several-fold over the unique geometry, dominating wall time and
memory. Native `rust/processing`, release build.

## Repro model

A production model (~462 MB oil-metering-station). Geometry is 100% explicit
tessellated — no swept solids, breps, or booleans:

| entity | count |
|---|---|
| `IFCTRIANGULATEDFACESET` | 4652 |
| `IFCCARTESIANPOINTLIST3D` | 4652 |
| `IFCMAPPEDITEM` | 22283 |
| `IFCEXTRUDEDAREASOLID` / `IFCFACETEDBREP` / `IFCADVANCEDBREP` / `IFCBOOLEAN*` | 0 |

So **22283 occurrences map onto 4652 unique triangle sets** (~4.8× reuse).

## Phase breakdown (native, release, from `ProcessingStats`)

```
parse=651ms  entity_scan=356ms  lookup=0  preprocess=44ms  geometry=42890ms  total=43541ms
total_triangles=43,290,571   total_csg_failures=0
```

- Parse is fast (0.65 s) — not the issue.
- **geometry = ~43 s**, producing **43.3M triangles**.
- Zero CSG. Tessellation quality has no effect (geometry is explicit triangles).

The ~4.8× mapping ratio implies ~9M **unique** triangles blown up to 43.3M by
per-occurrence materialization.

## What's happening

The source geometry is deduped (`ItemDedupCache`) and points are cached, so the
unique geometry isn't re-meshed 22283×. But the streaming **output** still emits
one full `MeshData` per occurrence with world-transformed vertices — it
copies + transforms ~43M vertices into 22283 flat meshes. `InstanceMeta` is
attached for the consumer to dedup afterwards, but the materialization has already
happened.

`IFC_LITE_RIGID_INSTANCING=1` does **not** help — tested on the same model:
identical 43.3M triangles, slightly slower (congruence-hashing overhead). The
rigid tier dedups metadata but the streaming output is still materialized flat.

## Proposal

An **instanced output path** in the geometry pipeline: emit each unique mesh
[...]

PR #1624, opened 2026-07-05
feat(instancing): fix mapped-item instancing + instanced streaming output foundation (#1623)

Blogbotana:
Two related pieces toward #1623 (the instanced streaming output). Everything new is
behind `StreamingOptions.enable_instancing` (default off) or changes only always-on
instance *metadata*, so the materialized/flat output is byte-for-byte unchanged —
mesh-determinism and ifcopenshell-parity are unaffected.

## 1. Fix: instance per-occurrence-target `IfcMappedItem`

`IfcMappedItem` occurrences that share an `IfcRepresentationMap` but differ by
`MappingTarget` were **silently not instanced** — the submesh path baked the target
into the vertices and re-hashed the post-target geometry into `rep_identity`, so
`collate_refs` saw a unique key per occurrence and rendered every one flat. That's
the whole per-occurrence-target class (metering stations, MEP, Tekla steel).

Fix (mirrors the single-mesh `process_mapped_item_cached`): record the target in
`InstanceMeta.local_transform` (composed for nested maps), keep the canonical,
pre-target `rep_identity`. Occurrences sharing a map now collate under one template.
This also immediately improves the GLB export instancing (#1443), which consumes
`local_transform`.

On a 462 MB metering station (22283 `IfcMappedItem` over 4652 `IfcRepresentationMap`):
distinct `rep_identity` 22280 → 4163, collated templates 3 → 2328.

## 2. Instanced output contract + producer

- `InstanceRecord` (exported) + `ProcessingResult.instances`.
- `StreamingOptions.enable_instancing` (default off).
- `collate_into_instances`: when enabled, collate the retained meshes by
  `rep_identity` (the same collation #1443 uses); every occurrence beyond the first
  of each template becomes an `InstanceRecord` and its `MeshData` drops out of
  `meshes`, leaving one template per group.

Same model: `result.meshes` 22283 → 6819, `instances` 15464 (all accounted for),
retained tris 43.3M → 29.6M. This is the **output/memory** dedup.

## Follow-up (tracked in #1623)

This does not yet cut the open *time* — `collate_into_instances` runs after
materialization. The CPU/open-time win (skip meshing the instanced occurrences up
front) is a deeper, per-item, determinism-sensitive change in the router; the plan
is written up in a comment on #1623.

## Tests

- `cargo test -p ifc-lite-geometry` — 400/400 (collate / congruence / recomposition).
- `cargo test -p ifc-lite-processing` — pass (the local Windows `mesh_determinism`
  mismatch is the pre-existing libm normal gap vs the Linux-pinned manifest; it also
  fails on `main` and passes on CI).

## Summary by CodeRabbit

* **New Features**
  * Added optional instancing support for repeated geometry, helping reduce duplicated mesh output when enabled.
  * Geometry results can now include instance records with per-occurrence transforms and identifiers alongside shared mesh templates.
  * Added a new setting to turn instancing on or off.

* **Bug Fixes**
  * Improved handling of mapped geometry so per-occurrence transforms are preserved correctly, including nested mappings.
  * Updated processing results to consistently include instance data in downstream outputs and tests.

### DIFF
--- a/rust/ffi/src/tests.rs
+++ b/rust/ffi/src/tests.rs
@@ -33,6 +33,7 @@ fn processing_result(
 
     ProcessingResult {
         meshes: vec![mesh_a, mesh_b],
+        instances: Vec::new(),
         mesh_coordinate_space: mesh_coordinate_space.map(str::to_string),
         site_transform: site_transform.map(|m| m.to_vec()),
         building_transform: None,
--- a/rust/geometry/src/router/processing.rs
+++ b/rust/geometry/src/router/processing.rs
@@ -412,32 +412,37 @@ impl GeometryRouter {
                         continue;
                     }
 
-                    // Apply MappedItem transform to newly added sub-meshes
+                    // Apply MappedItem transform to newly added sub-meshes.
                     if let Some(mut transform) = mapping_transform {
                         self.scale_transform(&mut transform);
-                        // The MappingTarget is baked into the vertices here, but
-                        // `rep_identity` was hashed from the canonical (pre-target)
-                        // geometry during the recursion. Two occurrences sharing a map
-                        // with DIFFERENT targets would therefore collate under one
-                        // template and reuse the reference's target offset/orientation.
-                        // When the target actually changes the geometry, re-hash the
-                        // post-target local geometry so rep_identity matches what is
-                        // baked. Identity targets (the common Revit case) leave the
-                        // geometry unchanged, so the canonical hash already matches —
-                        // skip the re-hash there to avoid the O(verts) cost.
+                        // The MappingTarget is a PER-OCCURRENCE transform, not part of the
+                        // shared geometry. It is baked into the vertices here (the
+                        // materialized/flat output is byte-for-byte unchanged), but for
+                        // INSTANCING it is recorded in `local_transform` — mirroring the
+                        // single-mesh `process_mapped_item_cached` — while `rep_identity`
+                        // keeps the canonical, pre-target content hash. So occurrences that
+                        // share a RepresentationMap but differ only by target collate under
+                        // ONE template (collate composes transform·local_transform to place
+                        // each). Previously this re-hashed the post-target geometry, giving
+                        // every distinct target a unique rep_identity and silently disabling
+                        // instancing for per-occurrence-target mapped items (metering
+                        // stations / MEP: ~4.8x reuse was rendered flat). #1623
                         let nontrivial_target = !transform.is_identity(1e-9);
                         for sub in &mut sub_meshes.sub_meshes[count_before..] {
                             self.transform_mesh_local(&mut sub.mesh, &transform);
-                            if nontrivial_target
-                                && sub
-                                    .mesh
-                                    .instance_meta
-                                    .as_ref()
-                                    .is_some_and(|im| im.instanceable)
-                            {
-                                let h = Self::compute_mesh_hash_full(&sub.mesh) | DIRECT_SOLID_TAG;
+                            if instancing_enabled() && nontrivial_target {
                                 if let Some(im) = sub.mesh.instance_meta.as_mut() {
-                                    im.rep_identity = h;
+                                    if im.instanceable {
+                                        im.local_transform = Some(match im.local_transform {
+                                            // Nested mapping: outer target ∘ inner target,
+                                            // matching the vertex bake order.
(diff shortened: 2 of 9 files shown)


louistrue:
Really nice work, @Blogbotana, and thanks for the depth. The diagnosis is spot on: the submesh path re-hashing post-target geometry into `rep_identity` is what was silently killing instancing for the whole per-occurrence-target class. Moving the target into `local_transform` and keeping the canonical hash is the right fix. I traced the algebra end to end, including the nested `transform * inner` composition against the recursion order, and it reconstructs each occurrence exactly. Flat output stays byte-for-byte, determinism and parity untouched. Splitting "always-on metadata" from "gated output contract" is exactly how I'd want this to land.

A few things first:

1. **It's red on the module ratchet** (`processor/mod.rs` 1622 vs budget 1533). Please don't bump the budget, split it: move `collate_into_instances` + `InstanceRecord` into a `processor/instancing.rs`. Reads cleaner too.
2. **Stats go stale when the flag flips.** `total_meshes/vertices/triangles/time` are captured before `collate_into_instances` mutates `meshes`, so with instancing on they'd report 22283 while `meshes.len()` is 6819. Recompute after collation, or add `unique_meshes`/`instanced_occurrences` and document `total_*` as pre-collation counts.
3. **`template_express_id` isn't unique** (it's the product id; multi-item Tekla products share it). `rep_identity` is the real join key and it's already on both sides, so let's make that authoritative. Two doc fixes while you're there: `rep_identity` is the content hash on the direct-solid path (not "RepresentationMap id"), and the `enable_instancing` docstring's "skips the 43M-vertex materialize" isn't true yet since collation runs after materialize.
4. **Add a producer test:** conservation (`meshes_before == meshes_after + instances.len()`), recomposition against the flag-off mesh, and the multi-submesh case where `rep_identity` is what disambiguates.

On perf, so the description stays honest: the always-on win is real and ships (GLB export now instances this class, roughly a third off the geometry payload), plus wider viewer instancing eligibility. Ignore the auto-posted benchmark deltas, that's CI variance on identity-target models, not this change. The 32% mesh dedup is behind the default-off flag and the 43s open time is unchanged, which matches how you scoped it.

For where this is headed: this is the metadata foundation for exactly the don't-bake path you sketched on #1623. Phase 1 (shared mapped-item registry, the byte-identical ~8s piece) is in progress on my side right now, and Phase 2 (don't-bake, skipping the ~29s per-occurrence materialize with a `verify_recomposition < 1um` gate) is the big one, together ~43s to 15-17s and 3.3GB to under 1GB, which also narrows the native-vs-wasm gap on plant/steel models. Your `local_transform` fix is a prerequisite for all of it. Get the split, stats, and a test in and part 1 is ready.

Blogbotana:
@louistrue Thank you for the feedback. I will try to fix everything 👍

louistrue:
Thank you, @Blogbotana — this was a genuinely valuable contribution, and you were pointing at exactly the right things: the per-occurrence `MappingTarget` fix for `IfcMappedItem` instancing, and the `InstanceRecord` + collate "instanced streaming output" foundation for #1623.

We ended up landing that same ground as a three-part series that grew directly out of this direction:

- **#1625** — a model-wide shared `IfcMappedItem` source cache (mesh each unique source once per load instead of once per owning element).
- **#1628** — the native "don't-bake" instanced output (`StreamingOptions.enable_instancing`, `InstanceRecord`, collate/finalize) — the ~29s materialize kill on instance-heavy models, byte-identical to the flat path.
- **#1629** — wiring the same path through the browser wasm partitioned pipeline (verified in-browser: 12,000 instances decode back to the exact source geometry, 0 flat triangles).

Because all three are now merged, this PR conflicts with `main` and duplicates what shipped, so I'm closing it as **superseded**. But the framing here — separating the mapped-item fix from the instanced-output foundation — genuinely helped shape the outcome. Thank you for digging into #1623. 🙏

louistrue:
Salvaged the one piece the Phase 1-3 don't-bake work (#1625 / #1628 / #1629) didn't cover: multi-item mapped sources with per-occurrence targets stay on the flat path, where the per-occurrence re-hash was still killing their instancing metadata (and GLB-export instancing #1443). Kept your local_transform approach, adapted to main's post-Phase-2 structure, with a regression test. Credited you as co-author. Thanks @Blogbotana. Follow-up in #1641.

PR #1626, opened 2026-07-05
feat(lists,sdk): Bonsai-style /regex/ set & property name matching [#1591]

louistrue:
## What & why

Part of issue #1591. The reporter wants to pull `NetVolume` from several quantity sets at once (`Qto_WallBaseQuantities`, `Qto_SlabBaseQuantities`), the way Bonsai does with a `/Qto_.*BaseQuantities/` pattern. Today every pset/qto name match in ifc-lite is exact `===`, so this is impossible.

This adds regex name patterns. A name wrapped in slashes (optionally with flags, e.g. `/qto_.+/i`) is matched as a regular expression; a plain name stays an exact, case-sensitive match. One column or query can then read a value from whichever matching set an element carries.

> Stacked on #1614 (base branch `feat/1591-lists-federation-columns`). Review/merge that first.

## Implementation

- `@ifc-lite/lists`: new `compileNameMatcher` / `isNamePattern` (`name-pattern.ts`). `/body/flags` compiles to a RegExp; anything else is exact. Compiled matchers are cached (the lookup runs per row); an invalid pattern is logged (never a silent catch) and falls back to an exact literal so it can never silently match everything or nothing. Wired into `findPropertyEntry` / `findQuantityEntry`, so both list columns and filter conditions are pattern-aware.
- `@ifc-lite/sdk`: `bim.query().property()` and `quantity()` accept patterns (reusing the lists matcher), returning the first match in set/property order.

## Tests

- Matcher unit suite: exact, `/regex/`, flags, anchors, internal-slash names, and the invalid-pattern path (warns + exact-literal fallback, never matches everything).
- Engine: a `/Qto_.*BaseQuantities/` column pulls `NetVolume` from both the wall set and the slab set; a regex condition filters on it.
- SDK: `property()` / `quantity()` resolve across sets via a regex set-name pattern.
- `@ifc-lite/lists` 54/54, `@ifc-lite/sdk` context 39/39; all packages build (tsc) clean.

## Changeset & API surface

`@ifc-lite/lists` + `@ifc-lite/sdk` minor. `scripts/api-surface.json` updated for the two new lists exports.

## Deferred (noted)

The SDK `where()` filter clause and the CLI/MCP `HeadlessBackend` filter resolution still exact-match set names, and the Rust export (`EntityRow::lookup`) does too. Extending those to the same matcher is a clean follow-up; this PR covers the interactive list query and the SDK value getters, which is the reporter's use case.

---

## Follow-up added: custom / regex column in the builder UI (OP's request)

The issue reporter followed up asking for "a double box where the user could enter them (Bonsai style)" alongside the static picker. The list column picker only offered exact discovered pset/qto names, so the regex above was reachable in filters and the SDK but not as a **column** from the UI.

Added a progressive-disclosure **Custom column** entry to the ColumnPicker: a Property/Quantity toggle plus two inputs (set name with discovered suggestions, property name), matching the existing panel idiom (dashed trigger like "Add filter", Chip toggle, ComboInput pair like the filter editor), collapsed by default to keep the picker uncluttered.

Verified live on `building-architecture.ifc`: a `/Qto_.*BaseQuantities/` + `NetVolume` custom column pulled NetVolume into ONE column from both the wall set (4 IfcWall) and the slab set (3 IfcSlab) - exactly the reporter's Bonsai use case. Viewer-only, no published API change.

### DIFF
--- a/.changeset/regex-name-patterns.md
+++ b/.changeset/regex-name-patterns.md
@@ -1,0 +1,6 @@
+---
+"@ifc-lite/lists": minor
+"@ifc-lite/sdk": minor
+---
+
+Support Bonsai-style `/regex/` patterns for property-set / quantity-set and property / quantity names. A name wrapped in slashes (e.g. `/Qto_.*BaseQuantities/`, optionally with flags like `/qto_.*/i`) is matched as a regular expression; a plain name stays an exact match. This lets one list column or query read a value across several matching sets at once, for example `NetVolume` from `Qto_WallBaseQuantities` AND `Qto_SlabBaseQuantities` (issue #1591). Applies to `@ifc-lite/lists` column extraction and filter conditions and to the SDK `bim.query().property()` / `quantity()` getters. `@ifc-lite/lists` exports the new `compileNameMatcher` / `isNamePattern` helpers.
--- a/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
+++ b/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
@@ -36,7 +36,7 @@ import type {
   PropertyCondition,
   ConditionOperator,
 } from '@ifc-lite/lists';
-import { discoverColumns, ENTITY_ATTRIBUTES } from '@ifc-lite/lists';
+import { discoverColumns, ENTITY_ATTRIBUTES, isNamePattern } from '@ifc-lite/lists';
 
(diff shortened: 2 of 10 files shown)


louistrue:
Recreated from #1616, which GitHub auto-closed when its base branch (#1614) was merged and deleted. Same content, now based on `main`.

PR #1630, opened 2026-07-05
fix(geometry): graceful handling of models over the wasm 4GB ceiling

louistrue:
## Problem (reported from prod on a 3.9GB model)

Loading a ~4GB IFC in the browser aborts the geometry worker with a bare `unreachable executed` trap and 0 geometry:

```
[stream] processParallel start, fileSizeMB=3899.1 workerCount=1 (bound=memory)
[Worker] Streaming prepass with SAB view failed (unreachable executed), retrying with copy
[useIfc] Error in processing: Error: unreachable executed
```

**Root cause:** the streaming prepass (`buildPrePassStreaming`) copies the whole file into wasm32 linear memory (4GB address space) **and** builds the entity index alongside it. Worse, the index is pre-reserved at `content.len() / 50` — **~82M slots (~1GB) for a ~4GB file**, on top of the resident file — so the reservation aborts *before the scan even starts*. Not a regression from recent perf work (the pre-alloc dates to #1389).

## Fixes

1. **Cap the index reservation** (`PREPASS_INDEX_RESERVE_CAP = 40M`). A rare huge model grows the map via rehash (one-time cost) instead of a fatal up-front OOM. Ordinary (<2GB) files keep their exact `len/50` estimate — unaffected. Lowers peak memory for every large model and lifts the practical browser ceiling (~3.2 → ~3.5GB).
2. **Actionable error** instead of the cryptic trap. When the prepass still traps on a very large file, `huge-file-error.ts` maps the wasm OOM/abort signatures to: *"This model is X GB, which exceeds the browser's ~3GB WebAssembly ceiling — open it in the desktop app."* Standalone import-safe module, unit-tested (6 cases incl. small-file/unrelated-error pass-through).

## Honest scope

A **true 4GB model still needs native/desktop** — the file + its entity index inherently exceed wasm32's 4GB, short of a streaming-prepass refactor or wasm64. This makes the failure *clear and actionable* and raises the ceiling for ~3–3.5GB models. The streaming-prepass refactor (never resident-copy the file) is the follow-up for real in-browser 4GB.

## Verify
- Rust build + clippy clean; `mesh_determinism` unaffected; module-size ratchet green (+9 justified).
- `huge-file-error.test.ts` 6/6; full `@ifc-lite/geometry` suite 134/134; typecheck clean.
- Changeset added (`@ifc-lite/wasm` + `@ifc-lite/geometry` patch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for very large models so scanning is less likely to fail due to memory limits.
  * Replaced cryptic WebAssembly failures with a clearer message that suggests using the desktop app.
  * Added a fallback path when zero-copy processing fails, making large-file imports more resilient.

* **Tests**
  * Added coverage for large-file error handling, including size thresholds and non-standard failure messages.

* **Chores**
  * Updated internal size tracking for the large-model handling change.

### DIFF
--- a/.changeset/huge-file-graceful-4gb.md
+++ b/.changeset/huge-file-graceful-4gb.md
@@ -1,0 +1,6 @@
+---
+"@ifc-lite/wasm": patch
+"@ifc-lite/geometry": patch
+---
+
+Fail gracefully on models that exceed the browser's WebAssembly memory ceiling instead of a bare `unreachable executed` crash. The streaming prepass copies the whole file into wasm linear memory and builds the entity index alongside it; on wasm32 (4GB address space) a ~3GB+ model can't fit, so the allocator aborted with an opaque trap. Two changes: (1) cap the entity-index up-front reservation (`content.len() / 50` reserved ~1GB of hash slots for a ~4GB file, on top of the resident file — that alone blew the budget before the scan; now capped, a rare huge model grows the map via rehash instead of aborting), which lifts the practical browser ceiling and lowers peak memory for every large model; (2) when the prepass still traps on a very large file, surface an actionable error ("This model is X GB, which exceeds the browser's ~3GB WebAssembly ceiling — open it in the desktop app") rather than the cryptic wasm trap. Ordinary (<2GB) models are unaffected (their reservation stays under the cap; the error helper never fires).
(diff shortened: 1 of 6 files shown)


louistrue:
Addressed in 73297553's follow-up: `largeFilePrepassError` now **preserves the original trap** verbatim in the message tail *and* as `.cause`, so a genuine Rust panic/assert on a >2.5GB model (which surfaces as the same `unreachable executed` as an OOM abort — good catch on the console_error_panic_hook detail) is still fully diagnosable rather than silently relabelled "too large". The heuristic stays (it only fires after BOTH the SAB-view and the full in-memory copy attempts trap on a >=2.5GB file — overwhelmingly the 4GB ceiling), but it can no longer mask a real bug. Added a test that locks in the preservation. Thanks 🙏

PR #1641, opened 2026-07-06
fix(geometry): instance multi-item mapped sources with per-occurrence targets (#1623)

louistrue:
## What

Salvages the one correctness edge from @Blogbotana's #1624 that the #1623 don't-bake phases (#1625 / #1628 / #1629) did **not** cover: **multi-item `IfcRepresentationMap` sources with per-occurrence `MappingTarget`s.**

The submesh path (`collect_submeshes_from_item_inner`) bakes each `IfcMappedItem`'s `MappingTarget` into the vertices and then, for any non-identity target, RE-HASHES the post-target geometry into `rep_identity`. That gives every distinct target a unique id, so occurrences that share a `RepresentationMap` but differ only by target never collate.

Phase 2 don't-bake only instances **single-solid** sources (`mapped_source_single_item`); it explicitly documents that *"multi-item sources ... fall through to the normal flat materialize — never instanced."* On that flat path the re-hash is still live, so the whole per-occurrence-target multi-item class (Tekla assemblies, multi-part MEP, metering skids) renders flat and **GLB-export instancing (#1443, which consumes `local_transform`) is disabled for it.**

## Fix

Record the `MappingTarget` in `InstanceMeta.local_transform` (composed outer∘inner for nested maps) and keep the **canonical, pre-target** content hash as `rep_identity` — mirroring the single-mesh `process_mapped_item_cached` and the don't-bake TEMPLATE re-tag already in the same function. Occurrences sharing a map now collate under one template per source solid; `collate_refs` / GLB export place each via `transform·local_transform`.

Single file of production change (`rust/geometry/src/router/processing.rs`), net-zero lines (module-size ratchet unchanged at budget).

## Safety

- **Flat output byte-for-byte unchanged** — the target is still baked into the vertices; only the always-on `InstanceMeta` metadata changes. Mesh-determinism + ifcopenshell parity unaffected (no manifest re-pin).
- **Does not touch the Phase 2 finalize** — `RawInstanceOccurrence`s are collected strictly from *empty* placeholder sub-meshes (`element.rs`); flat multi-item meshes are non-empty and never enter `raw`.
- **Every `rep_identity` consumer composes `local_transform`** — `compose_world` (collate.rs), `compose_world_meta` (gltf.rs), the wasm wire path — so sharing an id across distinct targets places each occurrence correctly (`rel_k · template = M_k · source`, exact up to f32).

## Test

`rust/processing/tests/instancing_multi_item.rs` drives the real pipeline on a new 2-solid map instanced 4× with 4 distinct targets and proves: (1) the occurrences collate into **2 templates × 4** (pre-fix: **8 singletons** — verified by reverting the fix), (2) each keeps its own target in `local_transform`, (3) the template-relative transform reproduces every occurrence's baked world triangles (< 1e-4 m).

## Verification

- `cargo clippy -p ifc-lite-geometry -p ifc-lite-processing --all-targets -- -D warnings` clean
- `cargo test -p ifc-lite-geometry --lib` — 401 passed
- `cargo test -p ifc-lite-processing` — all green (incl. `instancing_dont_bake` byte-identity gate, `mesh_determinism`, module-size ratchet)

Credit: salvaged from #1624 by @Blogbotana (Yuri Isachenkov), adapted to main's post-Phase-2 structure.


louistrue:
@coderabbitai review

PR #1658, opened 2026-07-07
fix(geometry): repair void-cut tear on faceted-BREP reveal cutters (issue #098)

louistrue:
## Problem

The ara3d ISSUE_098 model's plan-rotated Poroton wall (`IfcWallStandardCase` `2A8m\$1B8T2VuclBPhPDl_W`) rendered with jagged flap triangles bridging its 7 window openings — a genuine void-cut tear. The windows are `IfcFacetedBrep` prisms with **splayed/stepped reveals**; the same defect hits any wall whose openings are faceted-BREP reveal prisms.

Earlier isolation kept showing "no defect" because the isolate tool dropped the openings — fixed separately in #1656. With the openings present, the tear reproduces byte-for-byte.

## Root cause (drilled to the exact kernel functions)

The exact mesh-arrangement kernel (`kernel/`) itself leaves the cut non-watertight. A stage-by-stage open-edge trace showed:

- A **single** `subtract_mesh(box, one watertight cutter)` already yields open edges — before any consolidation.
- The `unrecovered` CDT-constraint count and centroid inside/outside classification both misbehave because **near-coplanar tri-tri intersections mis-record constraints whose interned points keep a cutter vertex's own 3D identity**, sitting mm-to-cm *off* the face they constrain. Inserting them into that face's 2D CDT (a) collapses two distinct 3D vertices onto one 2D location (segment along the drop axis), and (b) pulls the face's sub-triangles off its plane — so classification then keeps faces lying *outside* the host (the flap).

## Fix (two parts, both corpus-safe)

1. **`retriangulate::triangulate`** drops any constraint whose endpoints project to the same 2D point under the face's drop axis, or lie > 1 mm off the face's own plane. A genuine in-plane constraint always has 2D extent and sits on the plane within f32/snap noise (tens of µm), so no real constraint is dropped. **This is the root fix.**
2. **Disjoint-cutter batching** welds the extended cutter to bit-exact closure before the `mesh_is_closed_exact` gate, so a geometrically-watertight faceted cutter (whose shared-edge f32 coords differ in bits after the placement transform) can join a batch instead of falling to the sequential path, whose per-cut f32 round-trip re-jitters the previous seam.

The **splayed reveals are preserved** — no box-cut approximation.

## Results

| | open edges |
|---|---|
| Atomic `box − one cutter` | 28 → **0** (watertight) |
| Full 7-window wall | ~2534 → **~42** (5/7 windows fully watertight) |

The remaining ~42 (2 windows) is a deeper batched multi-component classification residual, tracked as follow-up.

## Safety

The off-plane filter touches every boolean, so it was validated against the full pinned CSG corpus: **436 lib tests + `wall_opening_cut_regression`, `csg_quality_regression`, `cdt_volume_watertight_verify`, `issue_1167_real_wall`, `issue_1367_multibody_opening`, `door_window_calibration_regression`** — all green. Tuning: 1 mm off-plane tolerance is well-centered (5 mm regresses; the contaminants sit ~0.17 m off, real constraints < 15 µm).

## Tests

- `issue_098_atomic_cut` — the 0.02 s kernel-level repro, now asserts watertight.
- `issue_098_reveal_wall` — the full wall via the router (local frame forced on to match wasm), asserts the open-edge bar. (Note: `maxEdge ~23 m` is the *legitimate* uncut top/bottom-face diagonal, not a defect — dropped that assertion.)

Changeset: `@ifc-lite/wasm` patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of wall/window cutouts so valid openings are more likely to stay clean and watertight.
  * Reduced cases where geometry could fragment into thin “flap” triangles around reveal edges.
  * Fixed a batching issue that could prevent otherwise valid cutters from being processed together due to tiny coordinate differences.

* **Tests**
  * Added regression coverage for cutting and void-processing scenarios to help prevent these issues from returning.


louistrue:
@coderabbitai review

Re the P1 "Scale off-plane filtering with the near-coplanar band": addressed in 81209a13 (pushed after your review of 1230a4cf). The off-plane tolerance is now scale-relative — `1e-3.max(extent * (1.0 / 4_194_304.0))`, i.e. exactly the `extent · 2⁻²²` near-coplanar band you referenced. So on far/georeferenced flush caps the tolerance widens to match `tritri::near_coplanar` and no `coplanar_clip` constraint is dropped (the filter no-ops there); only near origin, where the ~0.17 m contaminant towers over µm snap noise, is it excised. The full CSG corpus (incl. issue_1007 flush-cap regressions) stays green.

PR #1660, opened 2026-07-08
fix(geometry): recover dense faceted-reveal void cuts (Sloan traversal + volume-safe batch)

louistrue:
## What

Follow-up to #1658 (the off-plane-constraint fix). That PR fixed wall `2A8m$1B8T2VuclBPhPDl_W` (2534→42 open edges) but a **denser** instance of the same Poroton Ventilata wall type — `3FceP9AqX1_92g5eDdrV5C` (#1512699), 7 faceted-BREP splayed-reveal windows batched onto one face — was still visibly torn (~532 open edges, though volume was already corrected 2072→532 by #1658).

Root cause: two kernel limits the off-plane fix didn't reach.

1. **Channel-recovery pinch.** Where many windows cut one face, a constraint's channel of crossed triangles self-touches at a single vertex (a "figure-8"), and `recover_subsegment`'s single-successor boundary walk can't traverse it → the constraint stays unforced.
2. **Batch hard-reject.** *Any* unrecovered constraint made `difference_all` reject the whole batched N-ary cut to the sequential per-cutter path, which re-jitters and re-fragments the wall.

## Fix

- **`recover_via_traversal`** — a robust **Sloan ordered-segment traversal** that walks the crossed triangles in order from `a` to `b` (ordered *along* the segment, so it's immune to the pinch) and rebuilds the two boundary chains, forcing the edge. Runs only as a fallback when the fast boundary-walk recovery bails, so the common path stays **byte-identical**.
- **`difference_all_lenient` + volume-safe `subtract_many`** — keep the exact (topologically cleaner) batched cut when its **removed volume matches a sequential reference**; fall back to sequential only when the non-conforming batch would over/under-cut. This guards the #559171 / #1167 volume family (the #1167 rotated wall's batch under-cuts 3.7 m³ vs 13 m³ → correctly rejected).

## Result

| wall | before | after |
|---|---|---|
| `3FceP9AqX1…` (V5C) | ~2072 | **~108** open edges, removed volume unchanged |
| `2A8m$…` (W) | 42 | 42 (unchanged) |
| #1167 rotated wall | ~13 m³ (pass) | ~13 m³ (pass) |

Full CSG corpus (416 lib + all integration regressions), determinism, and the module-size ratchet stay green. Native + wasm both build.

The remaining ~108 (a few near-degenerate reveal-step segments the ordered traversal can't start on) is a further follow-up — down from a badly-torn wall, with correct volume and the splayed reveals preserved.

## Notes

- Constraint recovery (phase D) split out to `kernel/retriangulate_recover.rs` and `signed_volume6` to `kernel/signed_volume.rs` to hold the ratchet (shrink/split, no budget raise).
- New regression test `issue_098_v5c.rs` + fixture (same public ara3d ISSUE_098 model as the committed `issue_098_wall_W.ifc`).
- Changeset: `@ifc-lite/wasm` patch.

## Summary by CodeRabbit

* **New Features**
  * Added a lenient batched subtraction mode to produce best-effort results for non-conforming arrangements.
  * Introduced robust constraint-recovery logic to help recover missing edges via traversal-based fallback.

* **Bug Fixes**
  * Fixed complex wall “void-cut tearing” by enforcing constraints more reliably and improving fallback cut behavior.
  * Improved volume-accuracy validation for batched cut operations to avoid over/under-cut.

* **Tests**
  * Added a regression test and new IFC fixture for the previously problematic dense wall opening case.

* **Chores**
  * Updated module size allowlist for retriangulation recovery additions.


louistrue:
@coderabbitai review

Addressed the review in 9ac0db64:
- **codex P2 (isolate oracle budget)**: `subtract_many`'s volume-safe sequential oracle now wraps the fold in `budget::snapshot_counters()`/`restore_counters()`, so those reference booleans don't charge (or trip) the caller's shared #1109 batch budget. A dense/non-conforming group can no longer self-defeat the recovery by tripping the budget while merely computing the comparison.
- **CodeRabbit major (re-audit after traversal)**: split the final audit into a recovery pass then a COUNT pass over the settled mesh (new `kernel/retriangulate_audit.rs`). Since `recover_via_traversal` deletes crossed triangles and rebuilds only the boundary chains, a later call can drop an edge an earlier segment forced; counting after all recovery reflects the final state, so a dropped edge is never reported recovered.
- **nits**: fixed the `issue_098_v5c` doc header (was the #928204 fixture), added allowlist justification comments, recorded the retriangulate.rs shrink (966→935, audit extracted).

V5C 108 / W 42 unchanged; 416 lib + budget/volume/wall-opening/multibody corpus green; wasm builds; ratchet green.

Issue #1657, opened 2026-07-07
XYZ measure point (real-world)


Hans-Lammerts:
### The problem

I would love to have the ability to measure real-wold XYZ values. 
IFCmapconversion taken into account.
I tried to accomplish using the 'AI' modus but without decent results. 

Regards

<img width="593" height="507" alt="Image" src="https://github.com/user-attachments/assets/745a4d52-6c08-4819-9cd9-f58ae4674d1d" />

### Proposed solution

Extra button 'Measure XYZ Geo"

PR #1679, opened 2026-07-10
feat(viewer): async lat/lon readout for the geo XYZ measure (#1657 follow-up)

louistrue:
## What

Adds an asynchronous WGS84 **lat/lon** readout to the Geo XYZ measure. When the georef anchor has a usable IfcMapConversion, the live measure readout now shows a compact `Lat 47.376190 / Lon 8.541710` line under the existing Eastings/Northings/Height.

## Why

Follow-up to #1674, which shipped the georef-aware Geo XYZ measure (projected E/N/H via `viewerPointToProjected` + the `hasUsableMapGeoref` gate). The closed parallel PR #1671 additionally implemented an async lat/lon readout — the one piece main lacked. This ports that delta onto main's #1674 structure (main's `MeasurePanel` / `useAnchorGeoreference` shape, not #1671's).

## Changes

- **`lib/geo/reproject.ts`** — `reprojectPointToLatLon(eastings, northings, crs, lengthUnitScale)`: point-level WGS84 reprojection built on the existing proj4 machinery (`resolveProjection`). Map-unit-scale aware, handles geographic CRS (EPSG:4326) identity, and returns `null` rather than throwing for an unresolvable CRS.
- **`lib/geo/reproject.ts`** — `reprojectionInputKey(...)`: a **pure, exported** key-derivation function so the correctness-critical part is directly testable.
- **`components/viewer/tools/MeasurePanel.tsx`** — `useProjectedLatLon` hook resolves lat/lon asynchronously off that primitive key and renders a compact, labelled line under the E/N/H box. Non-blocking: E/N/H stays synchronous; the lat/lon line is simply absent when proj4 can't resolve the CRS.

## Dependency-key correctness (codex P2 on #1671)

The async effect must be keyed by a primitive derived from **all** inputs the reprojection consumes — not just CRS name + E/N, and not object identities (which would refetch every render). `resolveProjection` reads `crs.name`, `crs.mapZone`, `crs.description`, `crs.mapProjection`; `reprojectPointToLatLon` additionally reads `crs.mapUnitScale` and `lengthUnitScale`. `reprojectionInputKey` folds all of these plus the E/N quantised to ~millimetre in metre-space (unit-independent), so a georef edit that changes the projection while leaving the CRS name and rounded E/N unchanged still refreshes the readout, and sub-mm hover jitter does not spam proj4.

## Test coverage (`lib/geo/reproject.test.ts`)

- `reprojectPointToLatLon` round-trip sanity in EPSG:32760 (the mm-scaled building-architecture.ifc CRS used by `pick-to-geo.test.ts`) — result lands in the expected UTM-60S region.
- Map-unit-scale invariance: mm offsets (`mapUnitScale 0.001`) and the equivalent metre offsets (`mapUnitScale 1`) agree to 1e-9.
- Unknown CRS yields `null` (never throws).
- `reprojectionInputKey`: sub-mm jitter collapses to one key; a >1 mm move changes it; and a projection-metadata edit (`mapZone`/`description`/`mapProjection`/`mapUnitScale`) or a `lengthUnitScale` change all change the key even when name + E/N are unchanged.

## Verification

- Viewer `tsc --noEmit`: clean.
- `lib/geo` suite (`tsx --test`): 126/126 pass.

### DIFF
--- a/apps/viewer/src/components/viewer/tools/MeasurePanel.tsx
+++ b/apps/viewer/src/components/viewer/tools/MeasurePanel.tsx
@@ -16,6 +16,11 @@ import { useDraggablePanel } from '@/hooks/useDraggablePanel';
 import { useAnchorGeoreference, type AnchorGeoreference } from '@/lib/geo/useAnchorGeoreference';
 import { viewerPointToProjected } from '@/lib/geo/pick-to-geo';
 import { mapUnitsToMeters } from '@/lib/geo/cesium-placement';
+import {
+  reprojectPointToLatLon,
+  reprojectionInputKey,
+  type LatLon,
+} from '@/lib/geo/reproject';
 
 interface Vec3Like { x: number; y: number; z: number }
 interface Enh { e: string; n: string; h: string }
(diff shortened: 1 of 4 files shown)


louistrue:
@coderabbitai review

PR #1680, opened 2026-07-10
fix: recover from worker-script 404s after a redeploy; skip wasm scan over the 4GB ceiling

louistrue:
## Problem (observed live on ifclite.com)

A user's open tab failed to load a model with `Pre-pass worker failed: undefined` plus
`Loading Worker ... geometry.worker-<hash>.js was blocked because of a disallowed MIME type ('text/plain')`.

Today's merges rotated asset hashes under the open tab. The lazily-spawned pre-pass worker script 404'd - Vercel serves the 404 body as `text/plain` (with `cache-control: immutable`!), the browser blocks the worker, and `onerror` fires with an EMPTY message (the MIME detail goes only to the console). The existing recovery paths never trigger:

- `vite:preloadError` covers lazy CHUNK imports, not worker spawns.
- The wasm-MIME matcher (#1363) needs a message signature; there is no message.

The dead pre-pass then starved the parser of its shared entity index; after the 60s timeout it attempted a byte-level WASM scan of the 3.9GB source - a guaranteed wasm32 OOM trap (`unreachable executed`) - before finally reaching the JS tokeniser.

## Fix

1. **`notifyIfWorkerScriptUnavailable(err, receivedAnyMessage)`** (geometry pkg): an `onerror` with NO message from a worker that NEVER delivered a message is classified as the spawn-failure signature and dispatches the existing `ifclite:wasm-asset-unavailable` event - the viewer's subscriber reloads once (sessionStorage-bounded) onto the current deployment. Guard rails: message-bearing errors keep the strict wasm matcher (an in-worker crash never causes a reload), and an empty-message crash AFTER the worker spoke stays local. Message extraction is strict string-only - `String(err)` would stringify an ErrorEvent with an undefined message to `[object Object]` and defeat the classifier.
2. Pre-pass + pool worker `onerror` handlers track a `spoke` flag and synthesize actionable error text instead of the literal `undefined`.
3. **`wasmBytesScanAllowed`** (parser pkg): skip `scanEntitiesFastBytes` above 2.5GB - the buffer copy plus entity index cannot fit in wasm32's 4GB address space (mirrors the geometry prepass heuristic from #1630) - and fall straight to the JS tokeniser with a clear one-line warn instead of a wasm panic.

## Tests

- 5 new cases for the spawn classifier (empty-message/never-spoke fires; empty-message/spoke does not; real crash message does not; wasm-MIME message still fires; null error counts as empty) + existing matcher regressions: geometry 140/140.
- `wasmBytesScanAllowed` boundary tests: parser 244/244.
- tsc clean on both packages.

Changeset included (patch: @ifc-lite/geometry, @ifc-lite/parser).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when worker scripts fail to load after a deployment update, automatically reloading the current version when safe.
  * Prevented repeated reload loops when browser storage is unavailable.
  * Large files over 2.5 billion bytes now avoid unsafe WASM scanning and use the tokenizer fallback instead.
  * Improved handling and classification of worker and WASM loading failures.

### DIFF
--- a/.changeset/worker-script-skew-recovery.md
+++ b/.changeset/worker-script-skew-recovery.md
@@ -1,0 +1,6 @@
+---
+"@ifc-lite/geometry": patch
+"@ifc-lite/parser": patch
+---
+
+Harden huge-file loads against stale deployments and the wasm32 ceiling. (1) A geometry or pre-pass worker whose SCRIPT fails to load (a redeploy rotated the hashed asset; the 404 is served as text/plain and the browser blocks the worker with an empty-message onerror) now dispatches the existing version-skew recovery event so the viewer reloads once onto the current deployment, instead of dying with "Pre-pass worker failed: undefined". (2) The parser skips the byte-level WASM entity scan for sources over 2.5GB: the buffer copy plus entity index cannot fit in wasm32's 4GB address space, so the scan always trapped with `unreachable executed` before the JS tokeniser fallback ran anyway.
--- a/apps/viewer/src/lib/wasm-version-skew.test.ts
+++ b/apps/viewer/src/lib/wasm-version-skew.test.ts
@@ -6,6 +6,7 @@ import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
 import {
   recoverFromWasmVersionSkew,
+  recoverFromWorkerScriptSkew,
   installWasmVersionSkewRecovery,
   __resetWasmVersionSkewForTests,
   type VersionSkewDeps,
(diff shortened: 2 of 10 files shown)


louistrue:
Adversarial testing (5-lens: reload loops, false positives, spoke-flag races, boundary consistency, ungated callers) found the initial commit was an end-to-end NO-OP: the viewer subscriber re-validated the dispatched message with the strict wasm matcher, which rejects the synthetic worker-script text by design. Fixed by adding a `kind` discriminator to the event and routing `worker-script` through a trusted-but-bounded reload path. Two more findings hardened in the same commit: the reload policy now refuses to fire when the attempt cannot be recorded (storage-blocked embeds would otherwise loop under permanent CSP/proxy conditions - pre-existing hole, same policy as the vite:preloadError handler), and the parser gate boundary now agrees with the geometry huge-file heuristic at exactly 2.5e9 bytes. The adversarial suites are committed as regression tests. Refuted by the pass (with proofs): per-worker spoke-flag isolation, ungated scanEntitiesFastBytes callers, GiB/GB unit drift.

louistrue:
@coderabbitai review

Issue #1682, opened 2026-07-10
Viewer memory: 9.8GB avg / 16GB peak loading an 883MB CATIA model (BIMVision: 2.6/4GB)


louistrue:
Split out of #1661 (the rendering defect there is fixed by #1673; the reporter's follow-up memory numbers deserve their own workstream).

## Report

Loading CatiaA1.ifc (883MB, IFC2X3, 19.1M entities, 16653 faceted breps): the viewer averages ~9.8GB and peaks ~16GB. BIMVision on the same model: ~2.6GB average, ~4GB peak.

## Known context

- The streaming prepass copies the whole file into wasm linear memory alongside the entity index (#1630).
- Geometry worker pool sizing is memory-bound (workerCount log: bound=memory) but each worker holds its own wasm instance.
- SharedEntityIndex (#1445) already removed ~600MB of duplicated index.
- Source is retained in a SharedArrayBuffer during streaming.

A profiling pass on the actual model is in progress to attribute the 16GB peak (source SAB, prepass wasm, per-worker wasm, JS mesh retention, cache serialization, renderer staging) and rank fix levers.

louistrue:
Profiled the actual model (native probe + code-verified worker mechanics). Key numbers:

- The model is pure tessellation: 19.1M entities, 59k meshes, 46.7M verts, 23.5M tris, ZERO CSG. Geometry mass is only ~1.31GB per full CPU copy (positions+normals+indices).
- The 16GB peak is worker redundancy, not geometry: each geometry worker is a private MVP-wasm realm that (a) receives a full 883MB copy of the source per batch call (wasm-bindgen copies the SAB view into linear memory; high-water stays), and (b) rebuilds its own ~436MB entity-index FxHashMap. Per-worker heap ~1.32GB - matching the codebase's own fileSize x 1.5 heuristic.
- Smoking gun: computeWorkerCount's cores>=16 tier is the ONLY tier without the fileSizeMB>512 clamp (10-core caps at 3, 12-core at 4), so a 16-core desktop spawns 8 workers = ~10.6GB in geometry-worker wasm alone. The reporter's 9.8GB 'average' matches the 3-4 worker tiers; the 16GB peak matches the unclamped tier.
- Also resident: source SAB 0.86GB, CPU mesh retention 1.31GB (releaseGeometryAfterFinalize is hardcoded false), GPU copy 1.31GB, parser+prepass workers ~2.8GB during overlap, +1.1GB finalize transient.
[...]

Issue #1661, opened 2026-07-08
IFCWall cannot parse and render correctly on catia exported model


xyzbety:
### Version

v1.32.5

### Which surface

Viewer (browser)

### Reproduction

open file CatiaA1.ifc

### Expected vs actual

In the attached model（CatiaA1），IFCWall(Guid: 38a6l$RjzHpgFQVBmGl8CK、0oBPfgMJ9M8P6a_A7yNtSN) failed to parse and render correctly. Referring to the following comparison images.

The model file is quite large. I have sent it to you as an email attachment with the subject line "Model CatiaA1", and it is available for public testing.

<img width="2879" height="1738" alt="Image" src="https://github.com/user-attachments/assets/aac63536-237a-4008-9677-fff45fcf011e" />

<img width="2879" height="1686" alt="Image" src="https://github.com/user-attachments/assets/6e6754de-da82-4167-a9a2-433c81521eb8" />

### Environment

Windows 11, Edge 150.0, NVIDIA GTX 3060

xyzbety:
Thank you very much for all your work. I have an additional question: the IFCLite Viewer consumes a large amount of memory when parsing this model, with an average usage of approximately 9.8 GB and a peak of around 16 GB. In contrast, BIMVision averages roughly 2.6 GB with a peak memory usage of 4 GB. @louistrue

PR #1684, opened 2026-07-10
perf(geometry): clamp the 16-core worker tier for huge files (#1682)

louistrue:
## Problem

Issue #1682 (reporter follow-up on #1661): loading the 883MB CatiaA1 model peaks at ~16GB (avg 9.8GB) vs BIMVision's 4GB peak.

Profiling the actual model attributed the swing to the worker multiplier, not the geometry (which is only ~1.31GB per full copy: 46.7M verts / 23.5M tris / zero CSG):

- Each geometry worker is a private MVP-wasm realm that receives a **full 883MB source copy** per batch call (wasm-bindgen copies the SAB view into linear memory; the high-water sticks) and rebuilds its **own ~436MB entity-index hashmap** - ~1.3GB per worker, matching the codebase's own `fileSize x 1.5` heuristic.
- `computeWorkerCount`'s **`cores >= 16` tier was the only tier without the `>512MB` huge-file clamp** (10-core caps at 3, 12-core at 4), so a 16-core desktop spawned `cores/2 = 8` workers: **~10.6GB of redundant worker heap** - the 16GB peak. The 3-4 worker tiers produce the reported 9.8GB 'average'.

## Fix

Apply the same `>512MB` bandwidth ceiling as the 12-core tier: cap at 4. Rationale: huge-file geometry wall-time is memory-bandwidth bound, not core bound - the 12-core `?geomWorkers` A/B sweep showed a 5th/6th worker gives no speedup and starves the co-running parser; this model additionally has zero CSG (pure tessellation, parse-bound at 62%). `?geomWorkers=N` still overrides per host (and stays memory-budget-clamped).

Expected effect on the reporter's model: peak drops by ~5GB (8 -> 4 workers x ~1.3GB) on high-core machines, with no wall-time cost.

## Not in this PR (tracked in #1682, ranked)

- Per-worker index hashmap -> binary search over the already-shared sorted columns (~0.4GB x N)
- Threaded/shared-memory wasm (removes source duplication entirely; #1429 prior art)
- Gate `releaseGeometryAfterFinalize` on the 500MB threshold (~1.3GB CPU retention)
- Oct-encoded normals (~0.9GB)

## Tests

29/29 worker-count tests: new case pinning 16-core/883MB -> 4 workers ('cores'), the 2GB desktop case updated (clamp now binds before the memory budget), 400MB desktop case unchanged (still 8 - the clamp only bites >512MB). Full geometry suite 123/123, tsc clean.

Closes nothing (top lever of #1682; the issue stays open for the deeper levers).

### DIFF
--- a/.changeset/worker-count-huge-file-clamp.md
+++ b/.changeset/worker-count-huge-file-clamp.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Clamp the 16-core worker tier for huge files (issue #1682). The `cores >= 16` tier was the only one without the `>512MB` file-size clamp, so a 16-core desktop spawned `cores/2 = 8` geometry workers on an 883MB model - each a private wasm realm holding a full source copy plus its own entity index (~1.3GB per worker, ~10.6GB total), producing 16GB memory peaks. Huge-file geometry is memory-bandwidth bound (a 5th/6th worker gives no measured speedup), so the tier now caps at 4 workers above 512MB, matching the 12-core tier. `?geomWorkers=N` still overrides per host.
--- a/packages/geometry/src/worker-count.test.ts
+++ b/packages/geometry/src/worker-count.test.ts
@@ -223,15 +223,27 @@ describe('computeWorkerCount', () => {
     expect(r.count).toBe(8);
   });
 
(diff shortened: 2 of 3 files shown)


louistrue:
@coderabbitai review

PR #1681, opened 2026-07-10
docs: guides for MCP, clash detection, model diff, scripting SDK, collab, lens, lists, solar

louistrue:
Part 2 of the docs program (after #1676). Eight shipped feature areas had published packages but no guide on the docs site. This adds them, nav included, with every claim verified against source.

## Pages

- **MCP for AI Agents** (flagship): stdio/HTTP quickstart with the empty-registry gotcha, scope model, full tool catalogue with the four UNSUPPORTED_OPERATION stubs listed as planned, wiring for Claude Code / Claude Desktop / raw Streamable HTTP, programmatic embedding
- **Clash Detection**: engine + adapters, discipline matrix, CLI, review workflow with durable keys and max-interop BCF mapping
- **Model Diff**: fingerprint semantics (quantities in, AABB-centre out), type exclusions, and an explicit note that the CLI diff command is a separate lighter comparison, not the engine
- **Scripting SDK**: bim.* namespaces, eval/run (explicitly unsandboxed) vs the QuickJS sandbox (64MB/30s defaults, read-only perms), schema dump for LLM tooling
- **Real-Time Collaboration**: session API, sync server env vars and endpoints, branching/federation/E2E at mention level, the two runnable demos
- **Lens Visualization**: rule grammar (8 criteria types, first-match-wins), all 7 built-in lenses, renderer wiring via Scene.setColorOverrides
- **Lists & Schedules**: 7 column sources, all 6 presets, CSV formula-injection guard
- **Solar Analysis**: sunPosition/sunTimes/dayPath/analemma APIs with the year-not-Date gotcha, viewer Sun & Sky integration

## Verification

mkdocs build --strict clean; all internal anchors resolve; no em dashes or fancy unicode; docs-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added guides for clash detection, model comparison, real-time collaboration, lens visualization, schedules and lists, solar analysis, scripting, and MCP integration.
  * Documented key workflows, configuration options, APIs, CLI usage, integrations, and practical examples.
  * Updated documentation navigation so the new guides are easy to find.

PR #1676, opened 2026-07-10
docs: refresh README and all user-facing docs to current codebase state

louistrue:
## What

Full audit of every user-facing doc against the current source tree: root README, the mkdocs site (guides, tutorials, API reference, architecture pages, contributing docs), all published package READMEs, and example READMEs. Every claim kept or added was verified against the code; fabricated or drifted content was removed.

## Highlights

**Root README**
- Package/crate counts corrected: 36 npm packages, 6 crates on crates.io, plus the `ifclite-geom` PyPI wheel (was "14 TypeScript packages, 3 Rust crates")
- The "~260 KB gzipped" claim replaced: the WASM core alone is ~1.2 MB gzipped today
- "5x faster" headline qualified with the benchmarked median (~2.2x)
- New CLI + MCP section (the 30-command `ifc-lite` toolkit was absent entirely); package picker table expanded; all four examples listed

**API reference**
- `api/typescript.md` rewritten around all 36 published packages; stale API dumps corrected (`ParseOptions`, `ParquetExporter`, `StepExporter`, `Drawing2DGenerator`, `IfcCreator`, ...)
- `api/rust.md` now covers all 7 crates with corrected `Token`/`IfcType`/`EntityIndex` shapes; `api/wasm.md` gains the zero-recopy batching APIs; `api/python.md` verified accurate

**Guides and tutorials**
- Code samples fixed against actual exports (WorkerParser SAB requirement, IFC5 columnar API, query SQL column names, exact-CSG and unified mesh pipeline descriptions, GPU instancing section, WASM cache poison-recovery post-#1563)
- Privacy guide now accurately documents the hosted viewer's scrubbed PostHog telemetry (npm packages have zero analytics code) and that models are never uploaded
- Contributing docs: pinned nightly toolchain, manifest-based fixture fetch (Git LFS references removed), real test-suite matrix, OIDC trusted publishing (RELEASE.md NPM_TOKEN section fixed too)

**Package READMEs**
- Real API bugs fixed in samples: `validateIDS` signature, BCF `topics` Map access, cache GLB loader, codegen schema paths
- Stale "planned" markers removed from mcp/collab/collab-server (features shipped)
- READMEs added for the ten published packages that shipped without one: cli, create, sdk, sandbox, lens, lists, embed-sdk, embed-protocol, encoding, viewer-core

## Release

One changeset patch-bumps the 29 published packages whose README changed so the fixes reach npmjs.com.

## Verification

[...]

PR #1687, opened 2026-07-10
feat(docs): typecheck doc samples, generate drifting tables, guard READMEs

louistrue:
Part 3 of the docs program (#1676, #1681). The refresh fixed the state; this PR fixes the process, so the docs cannot silently rot again.

## What

**1. Doc samples compile now (`pnpm docs:check-samples`)**
Every `ts`/`typescript` fence in the README, all 32 guides, and all tutorials - 245 snippets across 38 files - is extracted and typechecked against the real workspace API (tsconfig paths into each package's src, no build needed). Conventional free variables (`store`, `canvas`, `meshes`...) come from a shared ambient file with REAL @ifc-lite types, so API names and signatures are genuinely validated. 5 deliberately illustrative fences carry an explicit skip marker.

Proof it bites: the checker found live drift in guides merged this week (`discoverDataSources` takes a categories object, not a string array; `extractUserLayer` takes a `captureBaseline()` state vector, not the IfcxFile) plus a batch of older drift (extractor signatures, section-plane normal is a tuple, the server-client vs geometry `MeshData` field mismatch now shown with an honest conversion). All fixed here, source-verified.

**2. Hand-maintained tables are now generated (`pnpm docs:generate` / `docs:check-generated`)**
Three marker-delimited regions: the 36-package index in `api/typescript.md` (from package.json descriptions), the CLI command table in `guide/cli.md` (parsed from the CLI's own HELP text, fails if <20 commands parse), and a perf-numbers block in `guide/performance.md` stamped from the committed CI benchmark baseline - the only machine-readable perf data in the repo; nothing un-backed is stamped.

**3. Published packages must ship a README (`pnpm docs:check-readmes`)**
Ten packages shipped to npm for months with none; now CI fails on the eleventh.

## CI wiring

The three checks run inside the existing `node-tests` job (no new matrix). Because that job is path-filtered to frontend/rust changes, a lightweight `docs-checks` twin (install + three node scripts, free runner) covers docs-only PRs - exactly the PRs most likely to break samples - and is skipped whenever `node-tests` already runs. The aggregate `Build + WASM + Rust + Node` gate includes it, so no branch-protection changes are needed.

AGENTS.md gains a short documentation-upkeep section (API/CLI change = guide update in the same PR; numbers live only in the generated region).

## Verification

All three checks green on this branch; `docs:generate` idempotent; `mkdocs build --strict` clean; workflow YAML validated; negative test confirmed (a fabricated wrong argument fails with the exact doc line).


louistrue:
Codex finding confirmed and fixed in the follow-up commit: the parser was silently dropping diagnose-geometry, extract-entities, generate-spaces (single space after long names), plus mcp (single-space description gap) and schema (no args column). The generated table now carries all 27 commands, and docs:check-generated re-verified.

PR #1691, opened 2026-07-10
feat(docs): serve the documentation at ifclite.dev/docs

louistrue:
Final part of the docs program (#1676, #1681, #1687). The docs move onto the main domain: **https://ifclite.dev/docs/** (landing stays at the root, per the agreed layout).

## How it works

- The existing `ifclite.dev` Vercel project (static, root `apps/landing`) gains a build step: `build-docs.sh` assembles `dist/` = landing static files + the mkdocs site at `dist/docs`. Pure vercel.json config - no dashboard changes needed (the project already includes files outside the root directory).
- `ignoreCommand` skips the build for pushes that touch neither docs nor landing, keeping Vercel build minutes flat.
- `site_url` becomes `https://ifclite.dev/docs/` (apex is canonical; www 308s to it). The GitHub Pages workflow keeps publishing unchanged as a mirror - its canonicals now point at the new home, so search consolidates without breaking old links.
- Repo-wide sweep: all 76 files referencing `ltplus-ag.github.io/ifc-lite` (READMEs, docs cross-links, landing page, package.json `homepage` fields) now point at `ifclite.dev/docs`. One changeset patch-bumps all 36 published packages so the new links reach npm.
- Landing nav's Docs entry links straight to `/docs/`.

## Verified before opening this PR

- Cloud-built preview deployment (real `buildCommand` in Vercel's builder): landing `/`, `/docs/`, guide pages, assets, and the search index all 200; `/samples/*` files unaffected. The PEP 668 externally-managed-Python wrinkle in the build image is handled with a venv.
- Canonical URLs on built pages point at `https://ifclite.dev/docs/...`.
- `docs:check-samples`, `docs:check-generated`, `docs:check-readmes` (from #1687) all green after the sweep; `mkdocs build --strict` clean.
- Failure mode is safe: if a production build ever fails, the domain keeps serving the last good deployment.

## After merge

Production deploys automatically on the merge commit; I will verify `ifclite.dev/docs/` live.

### DIFF
--- a/.changeset/docs-move-to-ifclite-dev.md
+++ b/.changeset/docs-move-to-ifclite-dev.md
@@ -1,0 +1,40 @@
+---
+'@ifc-lite/bcf': patch
+'@ifc-lite/cache': patch
+'@ifc-lite/clash': patch
+'@ifc-lite/cli': patch
+'@ifc-lite/codegen': patch
+'@ifc-lite/collab': patch
+'@ifc-lite/collab-server': patch
+'@ifc-lite/create': patch
+'create-ifc-lite': patch
+'@ifc-lite/data': patch
+'@ifc-lite/diff': patch
+'@ifc-lite/drawing-2d': patch
+'@ifc-lite/embed-protocol': patch
+'@ifc-lite/embed-sdk': patch
+'@ifc-lite/encoding': patch
+'@ifc-lite/export': patch
+'@ifc-lite/extensions': patch
+'@ifc-lite/geometry': patch
+'@ifc-lite/ids': patch
+'@ifc-lite/ifcx': patch
+'@ifc-lite/lens': patch
+'@ifc-lite/lists': patch
+'@ifc-lite/mcp': patch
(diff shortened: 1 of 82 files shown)


louistrue:
Both findings addressed in the follow-up commit:

- **ignoreCommand range**: valid concern in general, but main advances only by squash merges (verified: every recent main commit is a single-PR squash), so HEAD^..HEAD is exactly the push delta. The logic now lives in a commented vercel-ignore.sh stating that assumption and its fail-open behavior (any git error means build, never skip), so a future merge-policy change has a marked landmine.
- **rustdoc**: no doc page links into the GH Pages rustdoc subtree, and the published crates have authoritative rustdoc on docs.rs. api/rust.md now links docs.rs per crate (all six verified live); the GH Pages mirror keeps hosting its rustdoc snapshot unchanged. Building cargo doc inside the Vercel static build would cost a Rust toolchain per deploy for a copy docs.rs already provides.

PR #1694, opened 2026-07-10
fix(viewer): measure controls collision + geo toggle discoverability + hierarchy tab fit + toolbar a11y

louistrue:
Four UI/UX defects found by adversarial in-browser testing of the viewer on production. All verified in a real browser (localhost, chrome-devtools) with `getBoundingClientRect`, plus `tsc --noEmit`, the MeasurePanel/hierarchy/sidebar test suites (94 pass), and `vite build`.

## Defect 1 (HIGH) - Measure toggles collided with the Presentation pill
The Measure tool's Snap and Geo XYZ toggle pills rendered at `bottom-4 left-1/2 -translate-x-1/2 z-30` - the exact anchor of the persistent Presentation pill (`BasketPresentationDock`, same classes). Measured on prod:

| element | x range | y |
| --- | --- | --- |
| Presentation | 592-716 | bottom row |
| Snap | 567-633 | same row |
| Geo XYZ | 641-742 | same row |

The toggles were visually buried behind the pill and partly unclickable - which is why the controls "could not be seen".

**Fix:** move Snap and Geo XYZ INTO the draggable Measure panel as an always-visible controls row (present whether the panel is collapsed or expanded), top-anchored and well clear of the bottom strip. The instruction hint and live XYZ readout step their bottom offset up while the Presentation dock is visible (mirrors the storey-name pill's `bottom-4` to `bottom-28` shift), and the offset contract is documented in a comment next to the constants.

After the fix the toggles measure at y 122-149 (inside the top panel); the hint sits above the `Presentation` pill with no overlap.

## Defect 2 (HIGH) - Geo XYZ toggle was undiscoverable without georef
The toggle was hidden entirely when `hasUsableMapGeoref` was false, so users could not learn the feature existed or why it was missing.

**Fix:** always render the toggle while the Measure tool is active; disable it with the tooltip `Requires map georeferencing (IfcMapConversion) in the model` when unusable, unchanged behaviour when a georef exists. Verified: on a model without map conversion the toggle renders, is disabled, and shows that exact tooltip; on a georeferenced model it stays enabled and functional.

## Defect 3 (MEDIUM) - Hierarchy grouping tabs overflowed at default width
The `SPATIAL / CLASS / TYPE / MATERIAL / GROUPS` row truncated and overlapped ("MATERIAGROUPS") because five full uppercase labels need more room than the default sidebar pane provides.

**Fix:** a scoped container query (`.hierarchy-grouping-tabs`) collapses the row to icon-only below 384px - each button keeps its `title` tooltip so meaning stays discoverable - and restores clean labels above. Verified icon-only with no row overflow and no label clipping down to the minimum sidebar width, and clean labels with no clipping above the threshold. Scoped so the shared 200px panel-compact behaviour elsewhere is untouched.

## Defect 4 (LOW, a11y) - Icon-only buttons had no accessible name
Added `aria-label` (reusing existing tooltip text) to the toolbar tool/action buttons (`ToolButton`, `ActionButton`), Open / Add model / Export / Edit properties / Basket presentation / Info, the viewport Home / Zoom in / Zoom out controls, and the hierarchy expand / eye / remove buttons (with `aria-pressed` / `aria-expanded` where stateful). Verified zero icon-only buttons remain without an accessible name.

## Verification
- `npx tsc --noEmit` clean (rebuilt stale @ifc-lite/data + @ifc-lite/clash dists per the known worktree gotcha).
- MeasurePanel / hierarchy / sidebar test suites: 94 pass, 0 fail.
- `vite build` succeeds.
- All four fixes exercised in a live browser with measured rects.

No SpaceMouse files touched.

### DIFF
--- a/apps/viewer/src/components/viewer/HierarchyPanel.tsx
+++ b/apps/viewer/src/components/viewer/HierarchyPanel.tsx
@@ -804,7 +804,7 @@ export function HierarchyPanel() {
   // Multi-model layout with resizable split
   // Grouping mode toggle component (shared by both layouts)
   const groupingToggle = (
-    <div className="flex gap-1 mt-2">
+    <div className="hierarchy-grouping-tabs flex gap-1 mt-2">
       <Button
         variant={groupingMode === 'spatial' ? 'default' : 'outline'}
         size="sm"
--- a/apps/viewer/src/components/viewer/MainToolbar.tsx
+++ b/apps/viewer/src/components/viewer/MainToolbar.tsx
@@ -142,6 +142,8 @@ function ToolButton({
         <Button
           variant={isActive ? 'default' : 'ghost'}
           size="icon-sm"
+          aria-label={label}
+          aria-pressed={isActive}
           onClick={(e) => {
             // Blur button to close tooltip after click
             (e.currentTarget as HTMLButtonElement).blur();
@@ -274,6 +276,7 @@ function ActionButton({ icon: Icon, label, onClick, shortcut, disabled }: Action
         <Button
           variant="ghost"
           size="icon-sm"
+          aria-label={label}
           onClick={(e) => {
             // Blur button to close tooltip after click
             (e.currentTarget as HTMLButtonElement).blur();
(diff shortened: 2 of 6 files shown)


louistrue:
@coderabbitai review

PR #1700, opened 2026-07-10
feat(lists): edit columns in place + immediate-container column (#1591 follow-up)

louistrue:
Follow-up to the latest feedback on #1591 (does not close it).

## 1. Edit a list column in place

- Property/quantity columns in the Columns list get a pencil affordance that reopens the column editor pre-filled with the column's set + property (incl. `/regex/` patterns), following the existing add-column UX with the same live match preview and invalid-pattern guard.
- Saving updates the definition **in the same slot with the same id**, so column order, per-id table widths and index-based sort all survive; the list re-runs from the existing Run action.
- `CustomColumnEntry` is now a thin wrapper over the shared `ColumnEditorPanel` (add mode keeps the panel open and clears only the property; edit mode is a one-shot apply), so the add and edit UIs can never drift.
- Built-in attribute / material / classification / spatial / model columns carry no free-text formula, so they stay chip-toggled only.

## 2. `Container` built-in column

- New quick-add column (and spatial filter level) resolving the element's **immediate** spatial container via `IfcRelContainedInSpatialStructure`: the storey for buildings, or the `IfcBridgePart` / `IfcRoadPart` / `IfcSpatialZone` it actually sits in for infrastructure.
- Shows the container's real IFC `Name`, falling back to the container's IFC class when unnamed, and blank when the element is uncontained. Aggregated parts (not directly contained) fall back to their storey.
- Wired through a new optional `ListDataProvider.getContainerName` resolved per model provider, so federated multi-model lists resolve each row against its own model's spatial tree.
- Changeset included (patch `@ifc-lite/lists`); providers without the new method keep returning blank cells.

## Tests

- `packages/lists` engine: Container column values (storey, non-storey, blank when uncontained) + Container condition filtering.
- `spatialHierarchy`: `containerOf` resolves the immediate non-storey container (IfcBridgePart), class fallback when unnamed, '' when unplaced, storey fallback for aggregated parts.
- New `column-edit` suite: editability rules, draft round-trip preserving the id, in-place update preserving order and object identity of untouched columns, and an executeList re-run proving the edited slot yields the new values.

## Verification

- `npx tsc --noEmit` clean (viewer)
- `vitest` (packages/lists): 57 passed
- viewer lists + spatialHierarchy suites: all passing
- `vite build` succeeds

### DIFF
--- a/.changeset/container-adversarial-fixes.md
+++ b/.changeset/container-adversarial-fixes.md
@@ -1,0 +1,10 @@
+---
+"@ifc-lite/data": patch
+"@ifc-lite/parser": patch
+"@ifc-lite/lists": patch
+---
+
+Harden the immediate-Container spatial level (#1591 follow-up):
+
+- The spatial hierarchy now records an aggregated-descendant containment walk for ANY spatial container node, not just storeys, via a new optional `SpatialHierarchy.elementToContainer` map (also carried across data-store transport). A part nested through an IfcElementAssembly under an IfcBridgePart / IfcRoadPart / IfcSpatialZone now resolves that container instead of a blank cell. Storey-only `elementToStorey` semantics are unchanged.
+- The list engine matches the spatial level string case-insensitively, so a hand-edited / imported list carrying `container` resolves the Container level rather than silently falling back to the storey name. An empty or unrecognised level still defaults to Storey.
--- a/.changeset/lists-container-spatial-level.md
+++ b/.changeset/lists-container-spatial-level.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/lists": patch
+---
+
+Add a `Container` spatial level: a `spatial` column or condition with `propertyName: 'Container'` resolves the element's IMMEDIATE spatial container (its direct IfcRelContainedInSpatialStructure parent - the storey, or for infrastructure the IfcBridgePart / IfcRoadPart / IfcSpatialZone it sits in) via the new optional `ListDataProvider.getContainerName`. Providers without the method keep returning blank cells; existing levels (`Storey` default, `Building`, `Site`, `Project`) are unchanged.
--- a/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
+++ b/apps/viewer/src/components/viewer/lists/ListBuilder.tsx
@@ -42,6 +42,8 @@ import {
   isEditableColumn,
   draftFromColumn,
   columnFromDraft,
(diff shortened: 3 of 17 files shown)


louistrue:
## Adversarial findings addressed

Adopted three adversarial suites as regression tests (CONFIRMED cases flipped to assert the fixed behavior) and fixed the confirmed findings.

**1. Duplicate guard in edit mode + disarmed add-guard.** The edit-mode `ColumnEditorPanel` now gets an `isDuplicate` check (excluding the edited slot itself). Crucially, duplicate detection now keys on a content-derived `columnDefKey` (`source + set + property`) rather than the column id. The column id is kept stable across edits (sort/width binding relies on it), so before, a definition-changing edit drifted the definition away from its id and the add-guard stopped recognizing it. Keying on content means an edit into an exact duplicate is caught, and a re-added identical custom column is blocked, regardless of id.

**2. Label override destroyed on save.** `columnFromDraft` now takes the previous column and preserves a deliberate label override (a label that differs from the auto-label its own definition would generate, e.g. an imported `.list.json` display name). Only an auto-generated label is regenerated. A zero-change save is now a perfect no-op.

**3. Aggregated parts blank under non-storey containers.** The aggregated-descendant containment walk previously ran only for `IfcBuildingStorey` nodes, so a beam nested through an `IfcElementAssembly` contained in an `IfcBridgePart` got a blank Container. The builder now records that walk for ANY spatial container node into a new optional `SpatialHierarchy.elementToContainer` map (also carried across data-store transport). `elementToStorey` semantics are unchanged, so the Storey column stays byte-identical.

**4. Cheap wins.** Container filter value suggestions now offer actual container names (collected from every node that directly contains elements), not the storey fallback. The engine level parser matches the level string case-insensitively, so a hand-edited/imported list carrying `container` resolves the Container level; the level-less legacy default (-> Storey) is untouched.

### Gates
- `@ifc-lite/lists` vitest: 69 passed (incl. adopted engine adversarial suite)
- viewer lists/utils/hierarchy/search/schedule suites: 315 passed (incl. both adopted viewer adversarial suites)
- `@ifc-lite/parser` vitest: 247 passed (elementToStorey unchanged)
- `tsc --noEmit` clean in `apps/viewer` and `packages/lists`; `vite build` succeeds
- Changeset added for `@ifc-lite/data` / `@ifc-lite/parser` / `@ifc-lite/lists` (patch)

louistrue:
@coderabbitai review

PR #1704, opened 2026-07-10
perf(renderer): opt-in spatial chunk bucketing (#1682 phase 2)

louistrue:
Phase 2 of the chunked-residency plan for #1682. **Stacked on #1703** (uses its frame-stats observability to prove itself); rebase/retarget to main once #1703 merges.

## What

Opt-in **spatial chunk bucketing**: `Scene.setSpatialChunking({ cellSize })` partitions colour buckets by world grid cell (`cell~colour` keys), so batches become spatially compact and the existing per-batch frustum + contribution culls fire at chunk granularity. This is the structural prerequisite for the residency/eviction phases (a chunk-batch is an independently evictable GPU unit with a tight AABB).

Design points:

- **Pure reorganization.** Batches keep the ONE shared frame origin, so vertex relativization, the draw path, and highlight bit-coincidence are untouched. Verified pixel-identical (max channel delta 0 across a 1280x800 canvas) on FZK-Haus with the worst-case 16 m cells.
- **A mesh never splits across cells.** Cell membership comes from `origin + first vertex` (a true world point of the mesh, O(1), no placement transform needed - `localBounds` is pre-placement and would be wrong under rotation).
- **Every bucket-key derivation funnels through one helper** (`bucketBaseKey`): streaming append, fragment grouping (fragments inherit the parent mesh's key), finalize re-grouping, recolour routing (per-mesh now, since cells differ between an entity's pieces), partial-batch piece filtering, and move/rotate re-bucketing. The new `rebucketMovedMesh` migrates a mesh that crossed a cell boundary after an in-place translate/rotate, mirroring the recolour move mechanics - without it the partial-batch piece filter would silently drop moved elements under hide/isolate.
- **Off by default.** Session knob `globalThis.__IFC_LITE_CHUNKS` (`1` = on with the 32 m default cell, a number = cell size in metres, or `{cellSize}`), read once at renderer init; benchmark A/B env `VIEWER_BENCHMARK_CHUNKS`. Also adds a read-only e2e hook `globalThis.__ifc_lite_render_stats__` (live frame stats + resident GPU bytes, cleared on viewport teardown).

## Measured (headed Chrome, real WebGPU, FM_ARC_DigitalHub 14 MB)

Default framing (whole model visible - the draw-call cost side):

| | draws | resident GPU | meshes |
|---|---|---|---|
| chunks off | 21 | 20.0 MB | 4058 |
| 32 m cells | 85 | 20.0 MB | 4058 |
| 16 m cells | 193 | 20.1 MB | 4058 |

Zoomed into one corner (the culling-granularity payoff side):

| | batches culled | drawn |
|---|---|---|
| chunks off | 2 / 19 | 17 |
| 16 m cells | 43 / 191 | 148 |

Fully inside the model both configurations cull everything off-screen, but only the chunked scene does it at useful granularity. Load time, mesh counts and resident bytes are unchanged in all configurations; the ~0.1 MB delta is the per-batch uniform buffers.

## Verification

- 224 renderer + 1659 viewer app tests green (new: chunk-grid cell assignment, negative-coordinate flooring, determinism, empty/NaN anchors, key composition + `#N` overflow-suffix round-trip).
- Pixel-identity harness (chunks off vs 16 m cells, post-settle canvas screenshot diff): 0 differing pixels.
- Benchmark A/B via `VIEWER_BENCHMARK_CHUNKS` on FZK + DigitalHub: identical mesh counts and KPIs, drawCalls metric captures the multiplication.

Changeset included (`@ifc-lite/renderer` minor).


louistrue:
CI note: the Test/Benchmark workflows only trigger on PRs targeting main, so this stacked PR only runs the parity + Vercel checks until it is retargeted after #1703 merges. Full validation was run locally on this head: 224 renderer + 1659 viewer app tests, app typecheck + production build, api-surface snapshot updated, CodeRabbit CLI review clean (0 findings), pixel-identity harness (0 differing pixels, 16 m cells) and the headed-Chrome draw-call / cull A-B numbers in the description.

Issue #1717, opened 2026-07-11
Viewer integration of Layer PRs (stacks, drafts, merge/review UI)


louistrue:
Integrate the Layer PRs system (#1027) into the web viewer. The viewer already has the primitives: multi-layer IFCX federation (`useIfcFederation` / `addIfcxOverlays`, `composeFederated` with per-node source ids), the path <-> expressId bridge (`registerEntityMaps`), a component-level compare presentation (`ChangeDetailView`, overlay ghosting, `BcfFromChange`), and the registry REST API on collab-server. Spec: docs/architecture/layer-prs/08-review.md + 12-roadmap.md (L4).

### V1 — Read: layer stacks in the viewer
- [ ] `layerStackSlice`: ordered stack (`IfcxFile[]` + blake3 ids), active ref, source (local files / registry). NOT named anything with `mergeLayers` (that is the multilayer-WALL geometry feature).
- [ ] Layers panel: id `layers` appended to `WORKSPACE_PANELS` (group `review`; never reorder the first seven), body in `renderPanelBody.tsx`, feature-gated. Stack list, provenance badges (author kind, intent), per-layer info.
- [ ] Load a stack: multi-.ifcx open via the federation path; keep source-layer attribution.
- [ ] Layer diff: `diffLayerStacks` main-thread; path->expressId translation; reuse `ChangeDetailView` + `lib/compare/overlay.ts` ghosting.

### V2 — Write: draft layers from viewer edits
- [ ] `ChangeSet` -> `changeSetToOps` -> in-browser `publishLayer` (+ `deriveScopeOps`), GlobalId identity resolver backed by the data store.
- [ ] IndexedDB `LayerRefStore` mirroring the CLI `.ifc-lite/` layout (layers by blake3 id, refs record, draft descriptor); merge engine is storage-agnostic, so preview/merge/revert/rebase work offline.
- [ ] Publish to a local ref; follow-up: collab Y.Doc `publishLayer(doc)` for live sessions.

### V3 — Merge + review UI (L4)
- [ ] Registry client over `collabServerUrl()` HTTP base (same bearer scheme as the blob route).
- [ ] Conflict queue per 08-review.md; `subtree`-grouped delete decisions; merge gated on empty queue + green checks.
- [ ] Checks panel (IDS deep links into 3D selection, waivers with reasons), provenance panel, author-kind lens.
- [ ] Review comments as BCF topics bound to (review, entity, componentKey).

### V4 — Perf + polish
[...]

PR #1726, opened 2026-07-12
feat(collab-server): durable disk-backed layer registry, enabled in the deployed binary

louistrue:
PR 1 of the #1717 completion pass: the registry stops being memory-only. Until now the deployed collab-server binary never mounted the layer registry at all, and the only store lost every layer, ref policy, and recorded approval on restart.

## What it does

- **\`FsLayerRegistry\`** (\`src/layer-registry-fs.ts\`): a \`LayerRegistryStore\` on the data-dir volume. Layout under \`<dataDir>/layer-registry/\`: content-addressed \`layers/<blake3-hex>.json\` (read on demand, never held in RAM — the same hosting-cost trade the blob route made), \`refs.json\` (whole-record atomic rewrite, so arbitrary ref names never become filenames), \`reviews/<uuid>.json\`. Refs and reviews hydrate at boot and write through on change.
- **Durable before ack**: writes are synchronous atomic temp+rename. Registry ops are rare and bounded by the route's \`maxBytes\`; an accepted push that evaporates on restart would silently break refs.
- **Fail-closed hydration**: a corrupt refs or review file throws at boot instead of starting empty — an empty start would drop protected-ref policies and recorded approvals, the exact gates this store exists to persist.
- **Same integrity gates as the memory store**: the push gate (id verification + provenance validation) is extracted as shared \`assertPushableLayer\`; first-write-wins byte-identity and the layers/refs/reviews caps behave identically, and persisted layers count against the caps across restarts.
- **Deploy enablement**: \`COLLAB_LAYER_REGISTRY=1\` in \`bin.ts\` mounts the registry backed by \`FsLayerRegistry\` on \`COLLAB_DATA_DIR\` — the same mounted volume Railway already provides for rooms and blobs. Off by default; the startup log now reports \`registry: fs|off\`.

## Verification

- 8 new tests (suite 97 green): disk round-trip with isolation, id-mismatch + first-write-wins on stored bytes, restart rehydration of layers/refs-with-policy/reviews, caps counting persisted state, filename-unsafe review ids refused, fail-closed on corrupt state, leftover temp files ignored, and an HTTP e2e that pushes + fast-forward merges + opens a review, bounces the server onto a fresh store instance over the same dir, and reads everything back.
- Binary smoke: \`COLLAB_LAYER_REGISTRY=1 node dist/bin.js\` mounts \`/api/v1\`, serves layers/refs, and writes the on-disk layout.
- \`pnpm api-surface:update\` (three new exports), typecheck clean, changeset (minor) included.

## Notes

- Ops follow-up to actually turn it on: set \`COLLAB_LAYER_REGISTRY=1\` on the Railway service (volume already mounted).
- Next PRs from the #1717 completion plan: IDS evidence blobs + checks deep links, \`edited\`/bulk conflict resolutions, BCF review comments, Y.Doc drafts, webhooks + auto-merge policies.

## Summary by CodeRabbit

* **New Features**
  * Added an optional disk-backed layer registry that persists layers, refs, and reviews across server restarts.
  * Enable with `COLLAB_LAYER_REGISTRY` (startup logs indicate active mode).
  * Added registry-side room-token authorization support and registry access rules for registry routes.
* **Documentation**
  * Updated environment variable docs for the new persistence/registry settings.
  * Refined the layer registry roadmap status for ongoing phases.
* **Tests**
  * Added coverage for disk persistence, restart recovery, integrity/capacity enforcement, corrupted-state handling, and HTTP integration.


louistrue:
Review findings addressed in 1a140462: (1) room-token deployments locked the registry out (the default authorizer asks for a pseudo-room no room-bound token can match) — new `createRoomTokenRegistryAuthorizer` verifies signature/expiry/revocation without the room binding, wired via a new `authorizeRegistry` option and used by `bin.ts`; regression-tested against the default adapter. (2) `writeAtomic` now fsyncs before rename so the documented durable-before-ack guarantee holds through power loss.

PR #1727, opened 2026-07-12
feat(layers): fetchable IDS check evidence — reports route, layer push, checks deep links, waiver authoring

louistrue:
PR 2 of the #1717 completion pass (08-review.md §8.4). Stacked on #1726 — merge that first, then retarget this to main. Until now a manifest check's `report` digest was a dead-end hash: nothing stored the evidence bytes, nothing could fetch them, and waivers existed in the engine but could not be authored anywhere.

## Registry (collab-server)

- **`PUT/GET /api/v1/reports/<digest>`**: check-evidence blobs, content-addressed under the digest the manifest already names. blake3-verified on write (`assertReportDigest` accepts the raw-bytes digest or the NFC-normalized-text digest the CLI produces), first-write idempotent, capped (`maxReports`), durable on `FsLayerRegistry` (`layer-registry/reports/`), 501 on caller-supplied stores that predate evidence.

## CLI

- **`ifc-lite layer publish --check`** now also keeps the spec/report bytes in the local store (`.ifc-lite/evidence/<hex>`) — previously only the digests survived, so the bytes were unrecoverable.
- **New `ifc-lite layer push <ref|layer-id> --registry <url> [--token] [--set-ref]`**: uploads a ref's stack (or one layer) plus every evidence blob its manifests reference. Missing local evidence warns instead of failing (digests stay verifiable). Protected-ref refusal maps to exit 3 like `layer merge`. This is the first CLI-side registry client; until now only the viewer could push.

## Viewer

- **Checks deep links**: a check row's digest chip in the provenance detail expands into fetched evidence — spec pass/fail summary, entity-level failures with failure reasons, each clicking through to 3D selection (report `globalId` = composition path via `layerStackPathToId`; the report's own expressIds are never trusted), plus raw download. Offline/missing/foreign evidence degrade to plain messages.
- **Waiver authoring** (§8.4): after preview, a Required-checks block scores the target ref's `requiredChecks` against the candidate manifest; failing ones take a waive-with-reason input. Merge stays disabled until every conflict is decided AND every failing required check is waived (§8.3 gating). Waivers ride both backends (`MergeInit.waivers` locally, `waivers` on the registry merge route) and land durably in the merge manifest's `waived_checks` with the resolver.

## Verification

- New tests: 5 collab-server (digest-verified stores incl. fs restart, route PUT/GET/tamper/501), 4 CLI (push with evidence + bearer against an http stub, missing-evidence warning, protected-ref exit-3 path, base normalization, evidence-store round-trip), 1 viewer (requiredCheckStatus + policy-failure refusal + waived merge with durable waived_checks). Suites: collab-server 102, CLI 182, viewer 1686, all green; typecheck clean; api-surface + docs gates green.
- Localhost end-to-end (vite + Chrome DevTools MCP against a real registry): CLI published a candidate with a failing fire-safety.ids check, `layer push` uploaded layer + both evidence blobs; in the viewer the check's digest chip fetched the report ("1/1 specs failing, 1 entity failure"), clicking the failing wall selected it in 3D; a registry-ref merge preview showed the failing required check, merge stayed disabled until "theirs" + a waiver reason were entered, and the server-side merge layer recorded the resolution and the waiver verbatim.

## Notes

- `manifest.checks` evidence remains publish-time attached (in-process IDS execution over composed IFCX state is the L3 accessor-bridge item, unchanged).
- Remaining #1717 scope after this: `edited`/bulk conflict resolutions, BCF review comments, Y.Doc drafts, geometry-edit publishing, webhooks + auto-merge.


louistrue:
Review findings addressed in 7df71dd8: (1) `loadEvidence` validates the exact blake3 digest shape before any path join — a crafted manifest digest like `../../.env` can no longer read files outside the evidence store (regression test included). (2) the evidence download routes through the shared `downloadBlob` helper so the file-download event fires.

PR #1728, opened 2026-07-12
feat(layers): conflict queue completion — edited resolutions, edit-in-place editor, bulk actions

louistrue:
PR 3 of the #1717 completion pass, finishing the 08-review.md §8.3 conflict queue. Stacked on #1727 (which stacks on #1726) — merge the chain in order, retargeting as each lands.

## What was missing

The merge engine has supported `choice: 'edited'` (reviewer-typed replacement attributes) since #1027, but it was structurally unreachable: the viewer offered only ours/theirs, the orchestration explicitly filtered `edited` out of registry calls, and the registry route rejected it. Bulk resolution did not exist at all.

## Registry (collab-server)

- The merge route accepts `{ path, component_key, choice: "edited", attributes }` with strict shape validation (missing/array attributes are 400s). The engine's edited-target rules (componentKey-scoped only, no relation pseudo-components) surface as 400s instead of 500s.

## Viewer

- **Edit-in-place**: componentKey-scoped, non-relation conflicts gain an "edit" choice that opens a JSON attribute editor seeded with the current winner. Invalid JSON marks the row and disables the merge button until fixed. The parsed attributes ride `ResolutionInput.attributes` through both backends (the registry filter is gone).
- **Bulk actions**: "all ours" / "all theirs" over the whole queue, plus per-componentKey group actions ("Pset_FireSafety ×2: ours | theirs") whenever several conflicts share a key — the spec's bulk-by-selector, scoped to what the queue actually shows. Per-row choices still override.

## Verification

- New tests: collab-server e2e (conflicts 409 → malformed edited 400s → merged 200 with a reviewer-typed value composing over both sides), viewer orchestration (edited resolution merges locally, custom value wins composition, merge manifest records `choice: 'edited'`). Suites: collab-server 103, viewer 1687, typecheck + api-surface green.
- Localhost (vite + Chrome DevTools MCP): seeded a 3-conflict divergence (two walls on Pset_FireSafety, one on Pset_Acoustic). The bulk row and the ×2 group selector rendered; group-"theirs" resolved both fire conflicts; "edit" on the acoustic conflict seeded the editor with the current value; broken JSON disabled merge and showed the hint; after fixing, the merge produced a 3-layer ref whose composed state had REI120 on both walls and the custom Rw 47 (neither ours 45 nor theirs 50).

## Notes

- Remaining #1717 scope after this: BCF review comments, Y.Doc drafts, geometry-edit publishing, webhooks + auto-merge.

### DIFF
--- a/.changeset/edited-resolutions-registry.md
+++ b/.changeset/edited-resolutions-registry.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/collab-server": patch
+---
+
+The registry merge route accepts edit-in-place resolutions (`{ path, component_key, choice: "edited", attributes }`), strictly validated, with the engine's edited-target rules surfaced as 400s. Completes the conflict-queue spec (08-review.md §8.3) alongside the viewer's new edit choice and bulk actions.
--- a/apps/viewer/src/components/viewer/layers/LayerMergeSection.tsx
+++ b/apps/viewer/src/components/viewer/layers/LayerMergeSection.tsx
@@ -29,6 +29,7 @@ import { getBrowserLayerStore, DEFAULT_LOCAL_REF } from '@/lib/layers/browser-st
 import { LayerRegistryClient } from '@/lib/layers/registry-client';
 import {
   candidateLabel,
+  editedWithRemovals,
   executeMergeInto,
   previewMergeInto,
   refStackFiles,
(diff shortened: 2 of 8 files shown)


louistrue:
Review finding addressed in 85675d7a: keys the reviewer removes from an edited resolution now become explicit null tombstones (`editedWithRemovals`, filled against the union of both sides' keys) — per-attribute LWW no longer resurrects them. End-to-end regression test composes the merged stack and asserts the removed member is gone.

PR #1729, opened 2026-07-12
feat(layers): review comments as BCF topics — bound to (review, entity, componentKey), viewpoints, .bcf export

louistrue:
PR 4 of the #1717 completion pass, implementing 08-review.md §8.6: review comments are standard BCF topics bound to (review, entity, componentKey?), stored with the PR object, exportable as plain BCF, and structurally readable by agents. Stacked on #1728 → #1727 → #1726; merge the chain in order.

## Registry (collab-server)

- **`GET/POST /api/v1/reviews/:id/topics`**: topics carry title, optional description, the entity's composition path (the IFC GUID — layer data stays path-keyed, expressIds are never persisted), optional componentKey, and an optional BCF viewpoint payload (camera + snapshot data URL). Authors are server-derived from the credential, never caller-asserted. Writes follow the same named-reviewers gate as decisions; strict shape validation. Topics ride the review record, so `FsLayerRegistry` persists them across restarts for free.

## Viewer

- **Review section** in the Layers panel (registry target + candidate selected): finds or opens the PR object for the pair, lists the comment thread (clicking a topic selects the commented entity in 3D through the path bridge), and posts comments bound to the current 3D selection with an optional captured viewpoint (`createViewpointFromState`: camera + PNG snapshot; the non-serializable `snapshotData` is stripped, the data-URL snapshot rides the wire).
- **Export as plain BCF**: the thread downloads as a `.bcfzip` built with `@ifc-lite/bcf` (`createBCFTopic` + `addViewpointToTopic` + `writeBCF`), preserving guids, authors, creation dates, and viewpoints — foreign BCF tools can consume the review.
- Registry client gains the reviews surface it was missing entirely (`listReviews`, `openReview`, `getReview`, `listTopics`, `postTopic`).

## MCP (agent loop, 06 §6.3)

- New **`add_review_topic`** tool; **`get_review_feedback`** now returns `topics` alongside decisions, closing the reviewer-comments → agent-reads → agent-responds loop with one structure shared between the local workspace and the registry.

## Verification

- New tests: 4 collab-server (topic round-trip with viewpoint, malformed shapes 400, named-reviewer gate with per-principal bearer identities incl. author recording, fs-store restart persistence), MCP round-trip extended with `add_review_topic` + topics in `get_review_feedback`. Suites: collab-server 107, MCP 78, viewer 1687, typecheck + api-surface + docs gates green.
[...]

PR #1730, opened 2026-07-12
feat(layers): live-session Y.Doc draft publishing

louistrue:
PR 5 of the #1717 completion pass: the Layers panel's Draft section can freeze a LIVE collab session's edits into a published layer — the "collab Y.Doc drafts" leftover. Stacked on #1729 → #1728 → #1727 → #1726; merge the chain in order.

## What was missing

`@ifc-lite/collab` has had `publishLayer(doc, ...)` (CRDT session → frozen, provenance-stamped layer) since #1027, but the viewer never called it: viewer publishing read the LOCAL undo stacks only, so peers' edits in a live session were unpublishable, and there was no fork point to diff the doc against.

## @ifc-lite/collab

- **`CollabSession.captureDocState()`**: full-state fork point (`Y.encodeStateAsUpdate`). The existing `captureBaseline()` returns a state VECTOR (for the per-user `extractUserLayer` filter) which `publishLayer`'s minimal-layer diff cannot replay — this was the missing seam.

## Viewer

- **`collabDraftBaseline`** on the collab slice: captured when the session commits (after sync + owner seeding, so joining a populated room does not read its history as "edits"), cleared on stop, and re-captured after each publish so every session layer is delta-only.
- **`publishCollabDraft`** (`lib/layers/publish.ts`): the Y.Doc-source counterpart to `publishViewerDraft` — same local-ref append and recompose flow. Identity is shared by construction: collab docs address entities by GUID path, the same paths the composition uses. `author.kind` becomes `hybrid` when peers are present (their edits are in the doc too), `human` when solo. An empty delta refuses with "No session edits since the last publish" instead of minting an empty layer. `@ifc-lite/collab` loads dynamically, so Yjs stays out of the Layers panel chunk when no session exists.
- **Draft section**: in a session it grows a "Publish session edits" action (with a peers/hybrid hint) alongside the local-edits publish; local pending mutations clear on session publish since the doc already carries them (the mutation bridge mirrors every local edit).

## Verification

- New viewer test: memory-provider session end to end — hybrid-authored delta layer on the local ref with only post-baseline edits, empty-delta refusal, then a second delta-only publish (`human`, stack base) composing to the final value. Suites: collab 170, viewer 1688, typecheck + api-surface green.
- Localhost (vite, `VITE_COLLAB_ENABLED=true`, local-only IndexedDB session + Chrome DevTools MCP): loaded a 2-layer stack, started a session (baseline auto-captured), applied session edits directly on the live Y.Doc, published from the panel — stack grew to 3 with the session layer strongest and intent + identity-derived author on its manifest, composed FireRating took the session value, the baseline advanced, and an immediate re-publish with no new edits was refused (no empty layer on the ref).

## Notes

- Publishing only one peer's edits (`extractUserLayer`) stays a follow-up; the whole-doc + hybrid stamp matches how sessions are actually co-edited.
- Remaining #1717 scope after this: geometry-edit publishing, webhooks + auto-merge policies.

### DIFF
--- a/.changeset/collab-capture-doc-state.md
+++ b/.changeset/collab-capture-doc-state.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/collab": minor
+---
+
+`CollabSession.captureDocState()`: full-state fork point (`Y.encodeStateAsUpdate`) for whole-doc layer publishing via `publishLayer`, distinct from `captureBaseline()`'s state vector for the per-user `extractUserLayer` path. Backs the viewer's live-session draft publishing (#1717).
--- a/apps/viewer/src/components/viewer/layers/LayerDraftSection.tsx
+++ b/apps/viewer/src/components/viewer/layers/LayerDraftSection.tsx
@@ -146,8 +146,9 @@ export function LayerDraftSection() {
         stackFiles: state.layerStack.map((e) => e.file),
         intent: trimmedIntent,
         authorPrincipal: trimmedAuthor,
-        // Peers in the room means the doc may carry their edits too.
-        hybrid: state.collabPeers.length > 0,
+        // Anyone present since the baseline may have edits in the doc —
+        // including peers who already left.
+        hybrid: state.collabPeers.length > 0 || state.collabPeersSinceBaseline,
         refName: DEFAULT_LOCAL_REF,
       });
       // The published layer absorbed everything since the fork point;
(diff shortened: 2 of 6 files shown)


louistrue:
Review finding addressed in 42908549: hybrid authorship now derives from `collabPeersSinceBaseline` (peers present at any point since the draft fork, reset on baseline capture) rather than live presence at publish time — a peer who edited and left still stamps the layer hybrid.

PR #1731, opened 2026-07-12
feat(layers): geometry-affecting edit publishing — session placement layers, retype opinions, no silent drops

louistrue:
PR 6 of the #1717 completion pass: the "geometry-affecting edit publishing" leftover. Stacked on #1730 → #1729 → #1728 → #1727 → #1726; merge the chain in order.

## What the audit actually found

Three distinct gaps hid under "geometry edits are silently skipped":

1. **Placement edits in IFCX compositions are collab-native** (`usd::xformop` on the Y.Doc) — and the session publish path from #1730 already freezes them, but nothing pinned that behavior.
2. **`UPDATE_ENTITY_TYPE` (retype) was the one mutation type that genuinely fell through `changeSetToOps` silently** — a retyped entity published a layer that did not carry the retype.
3. **Nothing reported unrepresentable mutations**: a future/unknown type would vanish without a trace. (Local STEP placement moves, which write positional attributes on the placement chain's `IfcCartesianPoint`, already surface honestly through the existing `unresolved` reporting — sub-entities have no GUID identity by design.)

## Changes

- **@ifc-lite/mutations**: `UPDATE_ENTITY_TYPE` serializes as a class opinion (`attr:class` → `bsi::ifc::class { code }`), with `PredefinedType` riding the core-attribute channel (explicit null clears). New `skipped: Mutation[]` on the result reports anything with no component representation — the switch keeps a defensive default so runtime data from newer builds degrades loudly, never silently.
- **Viewer**: the wire layer maps the class opinion; `publishViewerDraft` returns `skippedCount` and the Draft section toasts when edits stayed out of a published layer.
- **Session placement publishing pinned by test**: a `setEntityPlacement` move on the live doc freezes into the layer delta as `usd::xformop` (row-vector 4×4, translation verified), composing onto the stack like any other opinion.

## Verification

- New tests: mutations (retype → class + PredefinedType ops, unknown type → `skipped`), viewer publish (retype wire form `bsi::ifc::class {code}` + `prop::PredefinedType`, mixed edits report `skippedCount` while publishing the representable rest), collab publish (placement move → `usd::xformop` in the delta with the exact translation).
- Suites: mutations 73, viewer 1690, typecheck + api-surface green.

## Notes

- Mesh-level edit publishing (beyond placement) stays out: there is no IFCX opinion form for mesh deltas yet; the collab geometry channel (content-addressed blobs) is the substrate a future spike would build on.
- Remaining #1717 scope after this: webhooks + auto-merge policies (PR 7, last one).

### DIFF
--- a/.changeset/retype-and-skipped-ops.md
+++ b/.changeset/retype-and-skipped-ops.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/mutations": minor
+---
+
+`changeSetToOps` serializes entity retypes (`UPDATE_ENTITY_TYPE` → a `bsi::ifc::class` opinion, with `PredefinedType` on the core-attribute channel) instead of silently dropping them, and reports unrepresentable mutation types in a new `skipped` result field so callers can warn instead of under-publishing.
--- a/apps/viewer/src/components/viewer/layers/LayerDraftSection.tsx
+++ b/apps/viewer/src/components/viewer/layers/LayerDraftSection.tsx
@@ -99,18 +99,22 @@ export function LayerDraftSection() {
         authorPrincipal: trimmedAuthor,
         refName: DEFAULT_LOCAL_REF,
       });
-      if (result.skippedCount > 0) {
-        toast.info(
-          `${result.skippedCount} edit${result.skippedCount === 1 ? '' : 's'} had no layer representation and stayed out of the published layer.`,
-        );
-      }
-      if (result.unresolved.length > 0) {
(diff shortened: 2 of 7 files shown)


louistrue:
Review finding addressed in 0e07c174: a publish with skipped (unrepresentable) edits now takes the same keep-pending path as unresolved identities — no `clearAllMutations`, no recompose, combined toast — so those edits are never dropped from the undo stacks.

PR #1749, opened 2026-07-15
feat(viewer): add plain OSM slippy-map base for the 3D world context

### LINKED ISSUE #1744: option for OSM map
### The problem

Hi there, it would be awesome if it was possible to integrate OSM as alternative, simple, background.
Sometimes Google Earth can be overwhelming. All you need is a simple map
A small 'switch to OSM feature  may be useful.

### Proposed solution

An OpenStreetMap (OSM) XYZ tile service allows you to stream global map tiles (slippy maps) into GIS software or web applications. The standard, free-to-use base map URL for OpenStreetMap is:https://tile.openstreetmap.org/{z}/{x}/{y}.png 
It would be possible to implement this URL as mapping applications 

### Alternatives considered

_No response_

louistrue:
Closes #1744.

## What
Adds a third **"OSM Map"** base map to the 3D world context, alongside Photorealistic (Google 3D Tiles) and OSM Buildings. It drapes plain OpenStreetMap tiles over the globe — a simple, uncluttered flat map for when the photorealistic globe is overwhelming (exactly what the issue asked for).

The base-map picker is also lifted out of the sun study into a **"Base map" dropdown** shown whenever the world context is enabled, so users can switch backgrounds without starting a sun study.

## How
- **`cesiumSlice`**: `CesiumDataSource` gains `'osm-map'`; the value is persisted/validated like the others.
- **`CesiumOverlay`**: for `osm-map`, show the globe with a `OpenStreetMapImageryProvider` layer (`tile.openstreetmap.org`, capped at `maximumLevel: 19` per the OSM tile usage policy, attribution credit carried automatically), `RECEIVE_ONLY` shadows like osm-buildings, and no 3D tileset.
- **`SunSkyPanel`**: the three-way source picker becomes a labeled `Base map` `<select>` in the world-context block.

## Incidental fix (dev-only)
While verifying, the OSM map (and the existing OSM-Buildings satellite base) rendered **black in dev**. Root cause: the `/cesium/*` dev middleware writes its own responses and never mirrored the app's COOP/COEP headers, so under cross-origin isolation Chrome blocked every Cesium worker (`ERR_BLOCKED_BY_RESPONSE`), killing the globe's terrain + imagery. The middleware now emits the same `Cross-Origin-Opener-Policy` / `Cross-Origin-Embedder-Policy` that `server.headers` (dev) and `vercel.json`'s `/(.*)` rule (prod) already apply. **Prod was never affected** — this only ever broke `pnpm dev`.

## Verification
Loaded a georeferenced sample (`building-architecture.ifc`, UTM 60S) in a real browser and drove the UI:

| Base map | Result |
|---|---|
| OSM Map | Flat OSM slippy map renders (land/sea/coastline/vegetation), model placed correctly |
| OSM Buildings | Satellite base renders (no longer black) |
| Photorealistic | Google 3D tiles render |

No console errors across all three; switching is durable. Typecheck clean on touched files; existing `cesium-*` unit tests pass (19/19).

<sub>Note: `apps/viewer` is `private: true` (not npm-published), so no changeset.</sub>

### DIFF
--- a/apps/viewer/src/components/viewer/CesiumOverlay.tsx
+++ b/apps/viewer/src/components/viewer/CesiumOverlay.tsx
@@ -261,7 +261,21 @@ export function CesiumOverlay({
         scene.globe.showGroundAtmosphere = false;
         scene.backgroundColor = Cesium.Color.TRANSPARENT;
         scene.globe.baseColor = Cesium.Color.TRANSPARENT;
-        if (dataSource === 'osm-buildings') {
+        if (dataSource === 'osm-map') {
+          // Plain OpenStreetMap slippy map: a simple, uncluttered flat base
+          // map (no satellite imagery, no 3D massing) for users who find the
+          // photorealistic globe overwhelming (#1744). The globe drapes the
+          // OSM tiles (over terrain when enabled) and receives cast shadows.
+          scene.globe.show = true;
+          scene.globe.shadows = Cesium.ShadowMode.RECEIVE_ONLY;
+          try {
+            // maximumLevel 19 = OSM's deepest tile zoom; conforms to the
+            // OpenStreetMap tile usage policy and avoids 404s past z19.
+            // The provider carries OSM's attribution credit automatically.
+            const osm = new Cesium.OpenStreetMapImageryProvider({ maximumLevel: 19 });
+            if (!cancelled) viewer.imageryLayers.addImageryProvider(osm);
+          } catch (e) { console.warn('[CesiumOverlay] OSM base map unavailable:', e); }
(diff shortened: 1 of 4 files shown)


louistrue:
@coderabbitai review

louistrue:
Good catch — fixed in the follow-up commit. The provider now passes an explicit `credit` of `© <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors` instead of Cesium's default MapQuest/CC-BY-SA string. Verified in-browser: the on-canvas attribution renders "© OpenStreetMap contributors" linking to the copyright page, matching the OSMF tile policy.

Issue #1766, opened 2026-07-16
Server-parse path: IDS values[] candidate arrays for enumerated/bounded/table properties not on the wire


louistrue:
Follow-up from #1759.

The WASM `parsePropertyValue` surfaces a `values[]` candidate-array for `IFCPROPERTYENUMERATEDVALUE` / `IFCPROPERTYBOUNDEDVALUE` / `IFCPROPERTYLISTVALUE` / `IFCPROPERTYTABLEVALUE` (each bound / member / cell), which IDS facet checks iterate ("any matching value passes", per upstream ifctester). #1759 made the **display value** identical across parse paths for Lists, but the data-model wire (`Property` = `property_value` string + `property_type` + `data_type`) has no `values[]` column, so **IDS validation on server-parsed models** can't do candidate matching for these property kinds.

Out of scope for Lists (which only reads the display value); relevant to IDS parity.

### To close
- Add a nullable repeated/`values` column to the server `Property` DTO + parquet properties table (additive; bump datamodel cache version) populated the same way as the WASM `values[]`.
- Carry it through `data-model-decoder.ts` → `serverDataModel.ts` onto the property entry's `values`.
- Cover with an IDS-on-server-path test.

PR #1795, opened 2026-07-17
fix(ids): resolve type-inherited property sets on server-parsed stores (#1787)

### LINKED ISSUE #1787: IDS on server-parsed models can't see type-inherited property sets
Found during the adversarial review of #1766.

The IDS bridge resolves type-inherited psets via `extractTypePropertiesOnDemand` (`packages/parser/src/on-demand-extractors.ts` ~415), which bails on empty `source`. `convertServerDataModel` sets `source: new Uint8Array(0)`, so on **server-parsed** stores IDS never sees a type's own property sets — only instance-attached properties. (Lists already got a server-path type fallback in #1759 via the `TYPEHASPROPERTYSETS` merge into `entityToPsets[typeId]`; the IDS bridge needs the equivalent.)

Consequence: an IDS facet checking a property that lives only on the element's IfcTypeProduct passes on the WASM path but is invisible on the server path. #1766 fixed candidate `values[]` for **instance** multi-valued properties; this is the remaining IDS/server gap.

### To close
Give the IDS bridge (`packages/ids/src/bridge/properties.ts` `appendInheritedPropertySets`) the same server-store fallback Lists uses — resolve the element's type via `getRelated(DefinesByType, 'inverse')` and read `store.properties.getForEntity(typeId)` when `source` is empty. Cover with a bridge test over a server-shaped store carrying a type pset.

louistrue:
Closes #1787. Final IDS/server parity gap after #1766.

## Problem
`packages/ids/src/bridge/properties.ts` `appendInheritedPropertySets` resolved type-inherited psets only via `extractTypePropertiesOnDemand`, which returns null when `store.source` is empty — i.e. on **server-parsed** stores. So an IDS facet targeting a property that lives on the element's `IfcTypeProduct` (not the instance) passed on the in-browser path but was invisible on the server path.

## Fix
Fall back to `store.properties.getForEntity(typeId)` — resolving the type through `IfcRelDefinesByType` — exactly like the Lists adapter's server-path type fallback (#1759). The server materialises type sets under the type's own id (the `TYPEHASPROPERTYSETS` merge), so the table lookup surfaces them with their candidate `values[]` (#1766) intact. **Guarded on empty `source`**, so the WASM path never enters this branch and is provably unchanged. No wire/cache/Rust change — pure TS bridge fix.

## Verification
New bridge test over a server-shaped store where the wall's `Pset_WallCommon` lives on its `IfcWallType`:
- a facet on a type-only property (`FireRating = REI 90`) **passes**
- a type-only candidate reachable **only via `values[]`** (`AcousticRating` requiring `R2`) **passes**
- a non-existent value still **fails**

`@ifc-lite/ids` 295/295, build + api-surface + viewer typecheck clean. Changeset: `@ifc-lite/ids` (patch).

## Deploy
No server change — but the property extraction it consumes ships with #1766 (already live). This is client-only, so it goes out with the next Vercel deploy of the viewer; no `railway up` needed.

### DIFF
--- a/.changeset/ids-server-type-psets.md
+++ b/.changeset/ids-server-type-psets.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/ids": patch
+---
+
+IDS validation on server-parsed models now sees type-inherited property sets (#1787). The bridge's `appendInheritedPropertySets` resolved type psets only via `extractTypePropertiesOnDemand`, which bails on the empty `source` buffer of a server-parsed store — so a facet checking a property that lives on the element's `IfcTypeProduct` (rather than the instance) passed on the in-browser path but was invisible on the server path. It now falls back to the prebuilt property table keyed by the type id (resolved through `IfcRelDefinesByType`), mirroring the Lists server-path type fallback. No wire or cache change; the WASM path is unaffected (guarded on empty `source`).
--- a/packages/ids/src/bridge/properties.ts
+++ b/packages/ids/src/bridge/properties.ts
@@ -165,12 +165,17 @@ function appendInheritedPropertySets(
   scale: number | undefined,
   out: PropertySetInfo[]
 ): void {
-  const inherited = extractTypePropertiesOnDemand(store, expressId);
-  if (!inherited || !inherited.properties || inherited.properties.length === 0)
-    return;
+  // Source-backed extraction (WASM/columnar parse) first; it bails on stores
+  // with no `source` buffer — i.e. server-parsed stores — so fall back to the
+  // prebuilt property table keyed by the element's IfcTypeProduct id (issue
+  // #1787), mirroring the Lists adapter's server-path type fallback.
+  const inheritedPsets =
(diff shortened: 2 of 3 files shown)


louistrue:
### Review verification
- **CodeRabbit hosted 'pass' was a rate-limited false pass** (fair-usage marker) — re-verified via the local CLI, which raised 1 Minor: the 'does not fabricate' test checked a *wrong value* rather than an *absent* property. Strengthened in 58f66256: split into a value-mismatch case and a genuine missing-property case asserting `PROPERTY_MISSING` (directly proves the type fallback doesn't fabricate psets).
- **Fable adversarial review: MERGE-READY** — empirically proved the new test FAILS on main, confirmed the WASM path is a strict no-op (guarded on non-empty `source`; `extractTypePropertiesOnDemand` never returns non-null-with-empty-properties), server shape/precedence/null-safety all correct. Also caught + I reverted an unrelated `packages/wasm/pkg/README.md` local-build drift hunk and a doc typo.

Issue #858, opened 2026-05-28
IfcTriangulatedFaceSet ignores IfcIndexedColourMap resulting in single color rendering


xyzbety:
In the attached model（tessellation-with-individual-colors）, the IfcTriangulatedFaceSet geometry in the model includes an IfcIndexedColourMap definition, mapping red, green, and yellow colors to specific triangular faces. However, the renderer ignores this color mapping and displays the entire object in a single color (red), failing to correctly render the multi-color definition specified in the file.Referring to the following comparison images.

[tessellation-with-individual-colors.zip](https://github.com/user-attachments/files/28334575/tessellation-with-individual-colors.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/4ccca205-c0c6-4fe7-8f41-ebd7e668e43d" />

<img width="2880" height="1719" alt="Image" src="https://github.com/user-attachments/assets/d5892079-0131-41d6-9277-b7ef5e9aa744" />

xyzbety:
In the latest version, this issue has reappeared. @louistrue 

<img width="2870" height="1635" alt="Image" src="https://github.com/user-attachments/assets/ffcc8415-d5a8-4185-b097-4e0790cb8033" />

xyzbety:
It feels like it's still a little off, the color of the top surface should be yellow. @louistrue 

<img width="2870" height="1635" alt="Image" src="https://github.com/user-attachments/assets/091daff9-f8a7-4384-97b5-99380cd5ca32" />

PR #1808, opened 2026-07-18
perf(geometry): don't-bake instancing keeps uniform indexed-colour sources (#1807)

### LINKED ISSUE #1807: perf(geometry): don't-bake instancing excludes ALL indexed-colour sources, even uniform ones (#858 over-broad)
## Summary

The don't-bake instanced-output path (#1623) produces **0 instances** on models
that are structurally ideal for instancing but store colour as
`IfcIndexedColourMap`, because the #858 exclusion drops *every* indexed-colour
source — including single-colour (uniform) sources that would never be split.

## Repro / symptom

A metering-station IFC (explicit `IfcTriangulatedFaceSet` + `IfcCartesianPointList3D`,
29852 elements, **22283 `IfcMappedItem`**) is exactly what instancing targets:
~2794 unique sources reused across 25287 occurrences. With `enable_instancing`:

- the don't-bake plan arms (2794 sources, `coord_space = raw_ifc`), eligibility
  passes for all 25287 items;
- **instances = 0**, ~40 s / ~50M tris / **27218 meshes** — no change vs
  instancing-off;
- every source is dropped by the indexed-colour exclusion: 100% carry
  `IfcIndexedColourMap` (`indexed_colour_full = 4725`).

## Root cause

`rust/processing/src/processor/mod.rs` (~L1211-1221) builds the exclusion set
from **all** `indexed_colour_full` keys:

```rust
let indexed_colour_split_ids = (instancing_plan.is_some() && !indexed_colour_full.is_empty())
    .then(|| Arc::new(indexed_colour_full.keys().copied().collect::<FxHashSet<u32>>()));
```

This forces any mapped source with an indexed-colour map onto the flat path so
#858's per-triangle split isn't lost to a single-colour instance placeholder.

But it is **over-broad**: the split itself only fires for genuinely multi-colour
maps. `split_mesh_by_indexed_colour` returns `None` when `distinct_used() < 2`
(`rust/processing/src/style/indexed_colour.rs` L134-136). A **uniform**
indexed-colour source (`distinct_used() == 1`) never splits — the flat path
already collapses it to a single `dominant()`-coloured mesh. Excluding it from
instancing buys nothing and blocks the entire model.

`meshes ≈ items` on this model (27218 ≈ 22283) confirms the colours are ~uniform
per source, not varying per triangle.

## Proposed fix

Only exclude sources whose map is genuinely multi-colour:

```rust
let indexed_colour_split_ids = (instancing_plan.is_some() && !indexed_colour_full.is_empty())
    .then(|| Arc::new(
        indexed_colour_full
            .iter()
            .filter(|(_, m)| m.distinct_used() >= 2) // only maps that actually split must bake
            .map(|(&id, _)| id)
            .collect::<FxHashSet<u32>>()
    ));
```

`FullIndexedColourMap::distinct_used()` is already `pub(crate)` and in the same
crate, so no visibility change is needed.

### Colour correctness

Uniform sources carry the right colour for free: `dominant()` is already resolved
into the element style index (`prepass::merge_indexed_colours`), and the flat path
already renders single-colour indexed maps with exactly that colour. The instanced
template inherits the same resolved style colour → pixels identical, just
deduplicated into a template + per-occurrence transforms. Multi-colour sources
(`distinct_used() >= 2`) still bake/split byte-identically to today.

## Tests

- `instancing_dont_bake`: add a mapped source with a **uniform** `IfcIndexedColourMap`
  → assert it instances (`instances >= 1`, not flat) and the instance colour equals
  `dominant()`.
- Keep a **multi-colour** mapped source asserting it still bakes/splits (regression
  guard for #858).
- Metering-station class: expect `instances ≈ 22283`, meshes `27218 → ~2794`.

## Impact

Cold first-open of explicit-tessellation + uniform-indexed-colour models (metering
stations, CATIA / 3DEXPERIENCE exports). Repeat-open is already instant via the
consumer's on-disk shard cache, so this is a first-open lever only.

Blogbotana:
Closes #1807.

## Problem

The don't-bake instanced-output path (#1623) produced **0 instances** on models
that are structurally ideal for instancing but store colour as
`IfcIndexedColourMap` — the #858 exclusion routed *every* indexed-colour source to
flat, including single-colour (uniform) sources that would never be split.

Root cause: `processor/mod.rs` built `indexed_colour_split_ids` from **all**
`indexed_colour_full` keys. But `split_mesh_by_indexed_colour` only fires for
genuinely multi-colour maps — it returns `None` when `distinct_used() < 2`
(`style/indexed_colour.rs`). A uniform source never splits; the flat path already
collapses it to one `dominant()`-coloured mesh, which an instance carries
losslessly. Excluding uniform sources bought nothing and disabled don't-bake
wholesale on metering-station / CATIA-class models.

## Fix

Filter the exclusion set to `distinct_used() >= 2`:

```rust
indexed_colour_full
    .iter()
    .filter(|(_, m)| m.distinct_used() >= 2) // only maps that actually split must bake
    .map(|(&id, _)| id)
    .collect::<FxHashSet<u32>>()
```

`FullIndexedColourMap::distinct_used()` is already `pub(crate)` and in the same
crate. Multi-colour sources still bake byte-identically (the #858 palette split is
preserved); uniform sources now instance carrying their dominant colour, which is
already resolved into the element style index by `prepass::merge_indexed_colours`.

## Validation

New test `uniform_indexed_colour_source_instances_keeping_dominant_colour`
(fixture `mapped_instances_indexed_colour_uniform.ifc`): the uniform source now
instances, recomposes to the flat world triangles within 1um, and keeps its single
dominant colour in both runs. The existing multi-colour guard
(`indexed_colour_source_palette_survives_instancing`) still asserts the palette
stays flat and ON==OFF byte-identical.

Measured on a 338 MB metering-station model (100% explicit `IfcTriangulatedFaceSet`
+ per-triangle `IfcIndexedColourMap`, 22283 `IfcMappedItem`), `enable_instancing`:

| | meshes | instances | baked tris | cold open |
|---|---|---|---|---|
| off (flat) | 27218 | 0 | 50.0M | 37.8 s |
| on (this PR) | 4725 | 22493 | 8.7M | **23.1 s** |

## Scope

Internal to `rust/processing`; no wasm/JS public-API change, so no changeset /
api-surface / `.d.ts` update. The change is gated behind `enable_instancing`
(off by default), so the default-path output and the determinism/parity manifests
are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved instancing behavior for mapped geometry that uses indexed colors, correctly identifying when the source is truly single-color.
  * Ensured retained instanced meshes keep the correct dominant color for uniform indexed-color inputs.
* **Tests**
  * Added a new fixture covering uniform indexed-color mapped instances.
  * Added assertions that validate instancing routing and exact retained color (bit-exact).


Blogbotana:
Addressed all review feedback in a9e8531:

- **CodeRabbit (multi-colour check perf):** replaced `distinct_used()` (clone + sort + dedup) with a zero-alloc early-exit `has_multiple_colours()`. Both call sites only need the `>= 2 distinct` threshold — the #1807 instancing filter and the existing #858 split guard (`style/indexed_colour.rs`) — so I switched both and removed `distinct_used()` (no other caller).
- **Greptile (empty guard on all-uniform models):** the split guard now stays unarmed (`None`) when no multi-colour source is collected. This is the common case for the #1807 target (metering stations are 100% uniform), so it avoids arming every per-job router with an empty set + the `Arc` allocation.
- **Greptile (fixture indexing):** clarified the 1-based `ColourIndex` → 0-based `triangle_palette` mapping in the test comment.

All `instancing_dont_bake` + indexed-colour tests still green.

louistrue (approved):

PR #1813, opened 2026-07-18
perf(geometry): drop parametric-profile fillets to sharp corners at Low/Lowest (#1809)

### LINKED ISSUE #1809: perf(geometry): drop parametric-profile fillets to sharp corners at Low/Lowest detail (steel LOD)
## Summary

At `Low`/`Lowest` tessellation quality, collapse the **fillets / edge radii of
parametric steel profiles** (`IfcIShapeProfileDef`, `IfcCShapeProfileDef`,
`IfcTShapeProfileDef`, `IfcLShapeProfileDef`, `IfcUShapeProfileDef`, `IfcZShape…`)
to **sharp corners** instead of merely coarsening the arcs. This is a preview/LOD
triangle-budget win on large steel models (Tekla-class), and it stacks with
instancing.

## Why this is cheap here (infra already exists)

- Parametric profiles are built **analytically** in `rust/geometry/src/profiles/`
  (`outline.rs`, `steel_shapes.rs`) — the section params (root fillet, flange-edge /
  web-edge radii, toe radii) are known, not baked into a mesh.
- `round_corner()` **already emits a sharp corner** when `radius < 1 µm`
  (`profiles/outline.rs:158`). The sharp path is written and tested.
- The LOD system already threads quality into the profile arcs:
  `TessellationQuality::profile_arc_segments(base, min)` (`tessellation.rs:115`)
  coarsens fillet arcs below `Medium` (Low = 50%, Lowest = 25% of `base`).

The only thing keeping fillets alive at Low is the **`.max(min)` floor** in
`profile_arc_segments`: they get coarser but never disappear. The proposal is to
let steel-section fillets go to **zero segments → sharp corner** at `Low`/`Lowest`.

## Proposed change

At `Low`/`Lowest`, route the steel-section corner builders to the sharp path — e.g.
pass `radius = 0.0` to `round_corner` (or have the fillet segment count resolve to 0
and treat 0 as "sharp") for the parametric I/C/T/L/U/Z profile outlines. `Medium`
and above are unchanged.

`Medium` stays the byte-identity case, so the mesh-determinism manifests
(`mesh_determinism.json`, pinned at Medium) are untouched; only the Low/Lowest
outputs change.

## Why it pays off (and honest bounds)

- **Cross-section triangles:** an I-section with root + toe radii is ~30–50 perimeter
  vertices vs **12** for a sharp I. Extrusion side walls are one quad per profile
  edge, so perimeter −60% ⇒ proportionally fewer side triangles + a cheaper cap
  earcut. Roughly **−40…60% of a slender member's triangles**.
- **Compounds with instancing (#1623):** identical sections already dedup to one
  template; fillets don't block that. Dropping fillets makes each template lighter,
  so the instancing win (fewer templates) and this win (lighter template) multiply on
  a model that is thousands of repeats of a handful of sections.
- **Open speed:** fewer arc points ⇒ less earcut / weld / transform work per member,
  stacked across 100k+ members.
- **Bounds (not oversold):** long flat extrusion faces dominate the render, and Low
  already halves fillet segments, so the marginal Low→sharp step is smaller than
  Medium→sharp. This is a **complementary** LOD lever, secondary to instancing, not a
  standalone silver bullet. Clash at Low becomes marginally conservative (a sharp
  corner is slightly outside the filleted one) — acceptable for a preview level, and
  Medium+ keeps the true section.

## Scope note

Only helps **parametric** profiles (`IfcExtrudedAreaSolid` + `IfcXShapeProfileDef`),
which is how Tekla exports steel (no faceted BREP). Faceted/tessellated exports bake
the fillets into triangles and are out of scope here (that is a mesh-decimation
problem, not profile LOD).

## Tests

- I-profile (`IfcIShapeProfileDef` with non-zero FilletRadius): at `Low` the profile
  outline is the 12-vertex sharp section (no arc vertices); at `Medium` the fillet
  arcs are present and byte-identical to today.
- One L- or C-section too (asymmetric toe/edge radii) to cover the concave root vs
  convex toe corner paths in `round_corner`.

Blogbotana:
Closes #1809.

## What

Steel sections authored as parametric profiles carry their root fillets and toe edge radii as attributes, and the outline builders tessellate each one into an arc. At preview tessellation levels that fillet is a sub-pixel detail which still costs a full arc's worth of cross-section triangles on every extruded member.

`TessellationQuality::profile_fillet_radius` returns `0.0` at `Low`/`Lowest` and the declared radius unchanged at `Medium` and above. `round_corner` / `push_arc` already emit a sharp corner when the radius falls below 1 µm, so zeroing the radius at the source needs no new branching in the builders — it sits alongside the existing `profile_arc_segments` / `circle_profile_segments` helpers and keeps the same "Medium is the identity" invariant.

Applied to the four profiles that model radii today:

| builder | radii collapsed |
|---|---|
| `process_i_shape` | FilletRadius |
| `process_l_shape` | FilletRadius + EdgeRadius |
| `process_u_shape` | FilletRadius + EdgeRadius |
| `process_t_shape` | FilletRadius + Flange/WebEdgeRadius |

The issue also lists C and Z. Those need no change: `process_c_shape` and `process_z_shape` already build plain sharp point lists and never read their radius attributes, so they are at the target output at every level already. Left alone rather than expanding scope into modelling fillets that do not exist yet.

## Measurements

A/B of the same build with and without the change, on a real Tekla steel model (`130_KM.ifc` — 682 `IfcExtrudedAreaSolid`, 25 parametric profiles carrying radii, IFC2X3):

| | triangles |
|---|---|
| medium (reference) | 2 933 223 |
| low **without** the change | 1 301 370 |
| low **with** the change | 1 092 875 |

−208 495 triangles, a further −16% on top of what `low` already did. `medium` matched to the triangle across both builds — the no-regression-above-Medium guarantee, confirmed on real data rather than only in unit tests.

On a synthetic file of pure parametric members (HEA300 / L150 / U200 / T200, all with fillets): 528 → 120 triangles, −77%. That is the 40–60% cross-section figure from the issue, seen without the surrounding plate/cut geometry diluting it.

## Tests

- Filleted W180 I-beam → exactly 12 outline vertices at `Low` and `Lowest`, area matching the closed-form sharp section (`180·171 − 174·152`), bbox unchanged (fillets are interior).
- `Medium` keeps its arcs; `High`/`Highest` asserted **coordinate-identical** to `Medium`.
- Asymmetric radii (L 12/8, U 12/6, T 12/6/4) collapse together to 6/8/8 sharp points, each paired with an assertion that `Medium` still rounds it.
- Unit test on the helper itself.

`cargo test -p ifc-lite-geometry` green (469 lib + all integration suites). Clippy clean on the touched files.

## Scope

Only parametric profiles swept by `IfcExtrudedAreaSolid` benefit. Faceted / tessellated exports have their fillets baked into the mesh and are unaffected — verified on a second Tekla model exported as `SurfaceGeometry` (2036 `IfcFacetedBrep`, zero profile definitions), where this change is correctly a no-op.

Worth flagging for review: dropping a concave root fillet removes material, so the cross-section area moves (~4.5% on a W180 with r=15). That is fine for the on-screen preview levels this targets, but it is a reason not to route `Low`/`Lowest` into anything that derives geometry-based numbers.

## Summary by CodeRabbit

* **New Features**
  * Added quality-aware profile geometry processing.
  * Low and Lowest quality now use sharp corners by removing profile fillets.
  * Medium, High, and Highest quality preserve rounded profile geometry.

* **Bug Fixes**
  * Improved consistency across I-, L-, U-, and T-shaped steel profiles.
  * Preserved profile dimensions while simplifying low-quality tessellation.

* **Tests**
  * Added coverage confirming geometry, vertex counts, and areas across tessellation quality levels.

### DIFF
--- a/rust/geometry/src/profiles/shapes.rs
+++ b/rust/geometry/src/profiles/shapes.rs
@@ -118,12 +118,16 @@ impl ProfileProcessor {
         // FilletRadius (attr 7) rounds the four web↔flange junctions (concave,
         // adds the root-fillet material). FlangeEdgeRadius (8) and FlangeSlope
         // (9) are not yet modelled (rare; absent in the ara3d set).
-        let fillet = profile
-            .get_float(7)
-            .unwrap_or(0.0)
-            .clamp(0.0, ((overall_depth - 2.0 * flange_thickness) * 0.5)
-                .min((overall_width - web_thickness) * 0.5)
-                .max(0.0));
+        // Below Medium the radius collapses to 0 and the section stays at its
+        // 12 sharp corners (issue #1809).
+        let fillet = self.quality().profile_fillet_radius(
+            profile
+                .get_float(7)
+                .unwrap_or(0.0)
+                .clamp(0.0, ((overall_depth - 2.0 * flange_thickness) * 0.5)
+                    .min((overall_width - web_thickness) * 0.5)
+                    .max(0.0)),
+        );
 
         let half_width = overall_width / 2.0;
         let half_depth = overall_depth / 2.0;
--- a/rust/geometry/src/profiles/steel_shapes.rs
+++ b/rust/geometry/src/profiles/steel_shapes.rs
@@ -51,13 +51,19 @@ impl ProfileProcessor {
         // FilletRadius rounds the inner re-entrant corner (concave, adds
         // material); EdgeRadius rounds the two leg toes (convex, removes the
         // sharp tips). Both optional; clamp so the arcs stay inside the legs.
-        let rf = profile
-            .get_float(6)
-            .unwrap_or(0.0)
-            .clamp(0.0, (width - t).min(depth - t).max(0.0));
-        let re = profile.get_float(7).unwrap_or(0.0).clamp(0.0, (t * 0.999).max(0.0));
+        // Below Medium both collapse to 0 — sharp corners only (issue #1809).
+        let q = self.quality();
+        let rf = q.profile_fillet_radius(
+            profile
+                .get_float(6)
+                .unwrap_or(0.0)
+                .clamp(0.0, (width - t).min(depth - t).max(0.0)),
+        );
+        let re = q.profile_fillet_radius(
+            profile.get_float(7).unwrap_or(0.0).clamp(0.0, (t * 0.999).max(0.0)),
+        );
         // Root-fillet segments per corner: 6 at Medium+, coarser below.
-        let seg = self.quality().profile_arc_segments(6, 2);
+        let seg = q.profile_arc_segments(6, 2);
         let half_pi = std::f64::consts::FRAC_PI_2;
         let pi = std::f64::consts::PI;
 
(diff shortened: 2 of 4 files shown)


louistrue:
thx, @Blogbotana 🚀

PR #1806, opened 2026-07-18
perf(geometry): beat web-ifc 0.0.77 on the full corpus — 2D + analytic-prism opening subtraction

louistrue:
## Summary

Wires IfcOpenShell's `boolean-attempt-2d` technique into the void-cutting pipeline. For an extruded host whose openings penetrate **straight through the extrusion depth** (parallel to the host axis), the expensive exact 3D mesh-boolean is replaced by a cheap **2D polygon difference** on the host profile (`bool2d::subtract_multiple_2d`, i_overlay) followed by a **re-extrude** of the holed profile.

The pre-existing `rect_fast` parametric path only handled clean rectangular box hosts; this generalises the idea to **arbitrary extruded profiles** and runs after it (rect+rect hosts keep their proven output byte-identical).

### Per-opening hybrid
A host's openings are split at capture:
- **Eligible** — parallel through-cuts → subtracted in 2D and re-extruded.
- **Residual** — perpendicular sleeves, partial-depth recesses, non-extruded voids → cut by the exact kernel on the re-extruded host.

So a single ineligible opening no longer forfeits its host's cheap ones.

### Correctness (any miss defers to the exact kernel with the FULL opening set)
- Host = one `IfcExtrudedAreaSolid` swept along local ±Z (arbitrary profile; mapped items unwrapped; clipped / multi-item bodies defer).
- Each footprint is gated **strictly interior** to the host profile (adjacent ones may legitimately merge; boundary-breaching ones go to the residual).
- The 2D difference must yield **one connected shape** (a profile-splitting void is rejected — the largest-shape keep would drop geometry) with at least one hole.
- The no-hole re-extrude **reconciles by bounds + volume** against the real host mesh (catches clipped / mis-recovered hosts); the holed re-extrude uses **watertight CDT caps** (earcut hole-bridge slivers would crack a many-hole cap) + an outward-orientation pass, and self-checks watertight.
- Emitted in the host mesh's **own per-element frame** (origin preserved) so the wasm local-frame / georeferenced precision path (#1297) is unaffected.

New helpers: `extrude_profile_watertight`, `cdt::triangulate_constrained` (unrefined, slivers-free, no Ruppert Steiner cost), `bool2d::subtract_multiple_2d_counted` (split detection).

Gate: `IFC_LITE_VOID_2D=0` forces the exact kernel (A/B / bisection). Default **ON**, native and wasm.

## Performance (native, 10 threads)

| model | before (t_geom) | after (t_geom) | 2D fires |
[...]

Issue #1788, opened 2026-07-17
Correctness pass residuals (T1-T6 vs IfcOpenShell, 2026-07-17): 12 mesh divergences + 1 uncut opening across 4 models


louistrue:
Ran the profiling repo's T1-T6 geometry-correctness framework (ifc-lite native @ 33a952f7 vs IfcOpenShell 0.8.2) on duplex, ISSUE_129, ISSUE_098, 170_KM after fixing two harness frame bugs (per-mesh local-frame origin not folded in the dumper; T6 opening meshes not rebased by the georef anchor — both fixed in the profiling repo, commits e4fc22b + 89a62be). Headline: coverage is exact everywhere (every element matched by GUID) and pass rates are 99.7-100%, but the residuals below are real engine divergences worth adjudicating.

## duplex (215 elements)
Clean: 210 pass, 5 warn:hausdorff, 0 fails. T6: 86 pass / 2 skips.

## ISSUE_129 (959 elements) — 2 fails, T6 361/361 pass
- `fail:volume` IfcColumn `Poteau béton - Rectangulaire:IGC_POT_1.8` — tris 24 vs 18, AABB volume ratio 1.20, voxel IoU 0.968 (borderline)
- `fail:shape` IfcWallStandardCase `Mur de base:IGC_MUR_EGC-TUN-VOILE-300:28` — tris 30 vs 12, vol ratio 0.83, voxel IoU 0.777

## ISSUE_098 (11,123 elements, georef) — 9 fails + 1 T6 fail; T6 5850 pass / 5 warn
- 3x `fail:shape` IfcSlab `Floor:1AR_PAV_CS008 - Esterno_...` (08knHh8Yz0EvVY$BMfRgcG, ...RrJd, ...RrSb) — ifc-lite 60-64 tris vs IOS 392, identical AABB volume, voxel IoU 0.64-0.71. Pattern: we tessellate far less and lose shape — likely an arbitrary-profile slab (holes?) meshed as its outer hull.
- 5x `fail:volume` + 1x `fail:shape` IfcWallStandardCase `1AR_MUR_CV003`/`PV101+102` family — ifc-lite 500-870 tris vs IOS 180-316, AABB identical, voxel IoU 0.63-0.97. We emit ~3x the triangles of IOS on these Poroton walls (opening-cut fragmentation?), hull volume drifts ~5-20%.
[...]

PR #1816, opened 2026-07-18
perf(geometry): fuzzy-boolean f64 subtract fast-tier with exact fallback

louistrue:
## Design

A tolerance-based **f64 boolean SUBTRACT** (ported from web-ifc's `fuzzybools`), added as an opt-in **fast tier** in front of the exact mesh-arrangement kernel, hooked into the batched disjoint-cutter path `mesh_bridge::subtract_many`. New module `rust/geometry/src/kernel/fuzzy/`:

1. **Shared-position snap** — both operands are quantised onto the kernel's `SNAP_GRID` (a power of two → exact, deterministic), so a point two triangle pairs compute independently (an opening corner shared by two faces) lands on the same cell. This is the coincidence that lets re-triangulated seams close watertight without a shared symbolic interner.
2. **Intersect** — BVH-cull (reusing `kernel/broadphase`) host×cutter triangle pairs, compute the intersection segment in plain f64 (Möller interval form), and push the **same** snapped endpoints to both faces.
3. **Re-triangulate** only the intersected faces via the in-tree deterministic CDT (`crate::cdt::triangulate_pslg`); **untouched faces pass through verbatim** — the speedup vs re-arranging everything. Per-face **collinear-merge** (a box side-face splits into two triangles → overlapping collinear segments the CDT rejects) and **cross-face T-junction resolution** (inject global endpoints that lie on a face's edges) keep the seams conforming.
4. **Classify** each sub-triangle inside/outside by a ray-cast from a **non-barycentric centroid** `(a+1.02b+1.03c)/3.05`, voted across 3 fixed directions. SUBTRACT keeps host tris OUTSIDE the cutters + cutter tris INSIDE the host (winding-flipped) as reveal caps.

All f64, FMA-free (no `mul_add`), fixed ray directions, deterministic ordering + hashers, IEEE `sqrt` only → **native==wasm bit-identical** (the shared CDT is already manifest-pinned via the prism-cut path).

## Safety: strict self-check + exact fallback

This is **not a replacement**. Every fuzzy result is strict self-checked in `subtract_many`:
- **(a) watertight** — every directed edge matched by its exact reverse (f64-bit keyed), and
- **(b) removed-volume == the disjoint oracle Σ\|host∩cutterᵢ\|** (the exact-kernel intersection oracle already used by `subtract_many`) within a tight 1% relative tolerance.

If **both** pass → use the fuzzy result. If **either** fails → discard it and run the existing exact path unchanged. A bad fuzzy result can therefore never regress correctness — the self-check catches it and the exact kernel takes over.

## Default: OFF

Env gate `IFC_LITE_FUZZY_BOOL` (default **OFF**). A default build is byte-identical to the exact kernel, so `cargo test -p ifc-lite-processing` `mesh_determinism` manifest and every pinned boolean manifest are unperturbed. **Rationale for shipping OFF:** (1) the A/B below is performance-neutral, so there is no user-visible upside to flipping it yet; (2) turning it ON would change tessellation on the determinism battery's void hosts, forcing a manifest re-pin (a real geometry change) — not warranted for a neutral first increment.

## i129 / i098 before→after (10 threads, best-of / warm interleaved, isolated `t_geom`)

| model | OFF | ON | Δ |
|---|---|---|---|
| **ISSUE_129** | min 1.153 / median 1.215 s | min 1.139 / median 1.232 s | **neutral (±~1%, in noise)** |
| **ISSUE_098** | ~1.45 s | ~1.45–1.49 s | neutral (no regression) |
| duplex | 0.035 s | 0.035 s | unchanged (no batched voids) |

Does **not** beat or meaningfully approach web-ifc's 0.31 s on i129 — see honesty note.

## Fuzzy hit-rate (per `subtract_many` call, `IFC_LITE_FUZZY_STATS=1`)

- **ISSUE_129**: 2 claimed / 1 fell back (3 batched groups total).
- **ISSUE_098**: 4 claimed / 6 fell back (10 batched groups total; all fallbacks = `not-watertight`, caught by the self-check).

## Honest magnitude

This is a **first increment of a large lever**, and the numbers say so plainly:
- The tier only hooks the **batched** `subtract_many` path. i129 invokes that only **~3 times** — the bulk of i129's opening cost is in the **binary** `subtract_mesh` path and the #1806 analytic/2D paths, none of which this tier touches.
- On the calls it *does* claim, the **strict exact-volume oracle** (Σ\|host∩cutterᵢ\|, computed via the exact kernel on every accepted call) costs about as much as the exact difference it's trying to avoid, so the fast f64 boolean's savings are cancelled by the correctness insurance. Net: performance-neutral.

The value delivered here is the **correct, deterministic, self-checking infrastructure** (guaranteed zero-regression), not a speedup yet. Clear follow-ups: (1) hook the binary `subtract_mesh` path, (2) a **cheaper independent volume oracle** so the self-check doesn't re-pay exact-kernel cost.

## Correctness harness — zero verdict regression (fuzzy ON)

IfcOpenShell (pip 0.8.2) harness, `correctness/run.sh`, ON vs OFF:

| model | semantic (T6 opening-cut) | T1–T5 vs IOS |
|---|---|---|
| **duplex** | 86 pass / 2 skip (both) | 215/215 matched, 0 fail (both) |
| **ISSUE_129** | 361/361 pass (both) | 959/959 matched, 1 pre-existing fail (both) |
| **ISSUE_098** | 5855 pass / 1 warn (both) | — |

Every verdict identical ON vs OFF. The self-check + fallback guarantee this by construction.

## Validation

- `cargo test -p ifc-lite-geometry` green (fuzzy OFF **and** `IFC_LITE_FUZZY_BOOL=1`); `cargo test -p ifc-lite-processing` green incl. `module_size_ratchet` + `mesh_determinism`.
- `cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings` clean.
- 8 new fuzzy unit tests: box-from-slab (watertight + volume), two disjoint openings, corner-overlap cube, determinism (byte-identical), disjoint-cutter keeps host, watertight-rejects-open-surface, flush-bottom notch **fails self-check → fallback**, default-off guard.
- Module-size ratchet respected (mesh_bridge kept under its 904 budget by homing the integration helpers in `fuzzy/`; each new file <400 lines; tests in `*_tests`-style `tests.rs`).

Refs #1788.

## Summary by CodeRabbit

* **New Features**
  * Added an optional fuzzy fast path for mesh subtraction (single and batched).
  * Improved performance for suitable geometry boolean operations using tolerance-based processing.
  * Added automatic validation to ensure subtraction results are watertight and preserve expected removed volumes.

* **Bug Fixes**
  * Invalid or uncertain fast-path results now safely fall back to the existing exact processing.

* **Configuration**
  * The performance optimization is disabled by default and can be enabled with `IFC_LITE_FUZZY_BOOL`.

### DIFF
--- a/.changeset/perf-fuzzy-boolean-tier.md
+++ b/.changeset/perf-fuzzy-boolean-tier.md
@@ -1,0 +1,7 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+perf(geometry): fuzzy-boolean f64 subtract fast-tier with exact fallback
+
+Add a tolerance-based f64 boolean SUBTRACT (ported from web-ifc's `fuzzybools`) as an opt-in FAST TIER in front of the exact mesh-arrangement kernel, for the batched disjoint-cutter path (`mesh_bridge::subtract_many`). It shared-position snaps both operands onto the kernel grid, intersects BVH-culled host×cutter triangle pairs into per-face constraint segments, re-triangulates only the intersected faces with the in-tree deterministic CDT (untouched faces pass through verbatim — the speedup vs re-arranging everything), and classifies each sub-triangle inside/outside by a ray-cast from a non-barycentric centroid voted across three fixed directions. All f64, FMA-free, IEEE-sqrt only, so native==wasm bit-identical. Every fuzzy result is STRICT self-checked — watertight AND removed-volume == the disjoint oracle Σ|host∩cutterᵢ| within 1% — and discarded (exact kernel runs unchanged) on any failure, so correctness can never regress. Gated OFF by default via `IFC_LITE_FUZZY_BOOL`; a default build is byte-identical to the exact kernel (mesh-determinism manifest unperturbed). First increment: on ISSUE_129 the tier is performance-neutral (the strict exact-volume oracle costs about as much as the exact difference it replaces, and only the few batched many-opening hosts are hooked), with zero verdict regression on the IfcOpenShell correctness harness (ISSUE_129 361/361, duplex 86/86, ISSUE_098 unchanged). Refs #1788.
--- a/rust/geometry/src/kernel/fuzzy/classify.rs
+++ b/rust/geometry/src/kernel/fuzzy/classify.rs
@@ -1,0 +1,103 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Point-in-mesh classification by f64 ray-cast crossing parity, BVH-culled and
+//! voted across three fixed directions for robustness.
+//!
+//! A closed oriented mesh contains `p` iff a ray from `p` crosses its surface an
+//! odd number of times. We use Möller–Trumbore ray–triangle tests (no back-face
+//! cull — a solid's far side must count), prune candidates with the shared BVH,
+//! and take the MAJORITY verdict over three fixed generic directions so a single
+//! grazing/edge-hit degeneracy on one ray cannot flip the result. Any residual
+//! misclassification only makes the fuzzy mesh fail the caller's watertight /
+//! volume self-check → exact fallback, never a silent wrong answer.
+//!
+//! DETERMINISM: FMA-free f64, fixed directions, IEEE ops only; the crossing
+//! COUNT (hence parity) is independent of BVH candidate order ⇒ byte-identical
+//! native==wasm.
+
+use super::super::arrangement::Tri;
+use super::super::broadphase::Bvh;
+use super::{cross, dot, sub};
+
+/// Three fixed generic ray directions: none axis-aligned, no two components
+/// equal, and no pairwise ratio near a simple architectural slope — so a ray is
+/// unlikely to lie in any operand face plane. (The first mirrors the exact
+/// kernel's parity direction.)
+const DIRS: [[f64; 3]; 3] = [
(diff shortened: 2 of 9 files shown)


louistrue:
Closing as a **documented negative result**.

The fuzzy-boolean f64 subtract tier is correct (self-check + exact fallback verified, zero production regression — shipped OFF) but does **not** deliver a speedup, proven empirically: even with a *free* self-check oracle it is ~+2.4% slower on ISSUE_129 (real A/B: i129 +1.4%, i098 +1.6% slower when ON).

Why web-ifc's fuzzy advantage does not transfer to ifc-lite:
1. ifc-lite's exact kernel is already in the fast well-conditioned regime on these openings — the BigRational escalation rarely fires on simple openings, and the #1806 analytic-prism path already skims the easy rectangular openings *before* the mesh boolean, so there is no slow exact path for a fuzzy boolean to undercut (web-ifc's 4x came from float booleans replacing *its* slow exact path).
2. 80-90% of the remaining cuts have flush/coplanar opening interfaces the transversal-only f64 intersection cannot make watertight, so they fall back — and each wasted intersect+retri+classify attempt costs more than the 10-20% that succeed save.
3. Through-openings straddle the host boundary, so the cheap containment oracle almost never fires and defers to the exact oracle anyway.

The corpus goal is already achieved and shipped (#1806): ifc-lite is faster than web-ifc 0.0.77 across the full corpus. ISSUE_129's remaining gap is not economically closable via a fuzzy boolean tier. Kept for the record; not pursuing.

PR #1830, opened 2026-07-21
perf(viewer,geometry): prewarm the engine wasm on idle instead of on file open

louistrue:
## Why

On a cold cache the ~3.9 MB engine binary (~1.3 MB brotli) is fetched lazily by the load path — that is, on the click that opens a model. Nothing overlaps the seconds the user spends in the file picker, so the whole download sits in front of first geometry.

Measured on production (ifclite.com), cold browser profile, ~4 Mbit link, 225 KB sample model:

| | |
|---|---|
| wasm fetch duration | **2535 ms** (1 266 173 B brotli / 3 927 682 B raw) |
| wasm fetch start | `10270 ms` — the file was handed to the loader at `10198 ms` |
| parser.worker.js | serialised after it, `start=12983 ms`, +169 ms |

~2.7 s of pure network in front of geometry, for a quarter-megabyte file. Warm cache is already fast (a 24.6 MB steel model reaches first visible geometry in 463 ms), so this is purely a cold-start problem.

Field telemetry agrees this is a tail problem, not a median one: p50 first-visible-geometry has been improving week over week (1683 ms → 880 ms over five weeks) while p90 sits at 3.7–11 s.

## What

1. **Prewarm.** `@ifc-lite/geometry` exports `prewarmSharedWasmModule(wasmUrl?)`; the viewer calls it from the first idle callback after mount (`apps/viewer/src/lib/wasm-prewarm.ts`). The compile was already memoised per resolved binary URL (#1818), so the later load awaits the in-flight promise rather than starting its own fetch. A prewarm that fails is swallowed and the load path degrades to exactly today's behaviour.

   **Skipped on metered / 2g-class links.** A visitor who never opens a model should not spend 1.3 MB of a data plan on an engine they did not ask for. Browsers without `navigator.connection` (Safari, Firefox) prewarm — the common broadband case.

2. **One memo for both init paths** (`packages/geometry/src/wasm-shared-module.ts`). Adversarial review caught that `processAdaptive` routes files under the 2 MB sync threshold — and the bridge/export APIs — through `IfcLiteBridge.init()`, which let wasm-bindgen do its *own* fetch and compile. That path never consulted the worker pool's memo, so it was deduped against a concurrent prewarm only by the HTTP cache. Since browsers are not required to coalesce in-flight requests for the same URL, a small file opened while the prewarm was still streaming could mean **two concurrent 1.3 MB downloads** — on exactly the slow links and file sizes this PR targets.

   The memo now lives in one module both paths import, and `IfcLiteBridge.init()` reuses an already-started compile via `getStartedSharedWasmModule` — which joins an in-flight compile but never starts one, so a consumer with no prewarm behaves exactly as before. Node still hands over bytes read off disk and is untouched. This also removes a pre-existing redundant compile (a small file opened after a large one used to recompile the binary the pool had already built).

## Verification

Chrome, against the production build served by `vite preview`:

- Prewarm fetch starts at **467 ms**, right after `loadEventEnd`, with no file picked.
- A **220 KB** file (the sync path) opened afterwards produces **exactly one** wasm fetch for the whole page lifetime — it was two before item 2.
- The **24.6 MB** file (worker-pool path) still yields an identical **2569 meshes / 258k verts / 22 batches**, zero console errors.
- With the `saveData` gate forced on via an init script: **zero** wasm fetches before the file is picked.
- `pnpm typecheck` green; viewer suite **1765 tests / 0 failures**; `@ifc-lite/geometry` 146/146; new `shouldSkipPrewarm` unit tests cover the affordability gate.
- CodeRabbit CLI: converged to **no findings** (it raised a single-flight eviction race and a gate-ordering bug on the second commit; both fixed in 6f326f31).

## Known, unchanged

A prewarm whose fetch stalls is now awaited by the sync path too, where it would previously have started its own. That exposure already existed for the worker path (`processParallel` has awaited the shared compile since #1818) and is bounded by the browser's socket timeout, after which the failed URL is evicted and the next attempt refetches.

## Deliberately not in this PR

Surfaced while measuring, both pre-existing, neither a regression:

- Vercel serves the binary at **1 266 173** bytes where `brotli -q 11` reaches **947 530** — ~25% of every cold download is compression-level waste. (`wasm-opt` is *not* the answer: it cuts raw size to 3.63–3.71 MB but *increases* brotli to ~960 KB, which is why it is correctly left disabled.)
- The parser worker still calls its own `init()` and compiles the binary independently of the shared module.


louistrue:
@coderabbitai review

PR #1849, opened 2026-07-23
perf(harness): interleaved base-vs-branch ab.sh + .manifest-ignore redistribution guard

louistrue:
**PR 2 of the perf program** (stacked on #1845). Two authoring-loop tools that turn perf/corpus rules — until now only prose in the ledger and session memory — into machinery. No behavior change to the product; scripts only.

## `scripts/perf/ab.sh` + `ab-report.mjs` — trustworthy native A/B
The #1 perf mistake here is measuring a "win" that's machine drift or a contaminated artifact (the whole failure ledger). This makes a *trustworthy* base-vs-branch delta one command:
- builds `perf_probe` from **both** a base ref (throwaway worktree, isolated `CARGO_TARGET_DIR` so the two sides can't share a binary) and the working tree;
- runs them **interleaved** (base, branch, base, …) so a mid-run slowdown drags both sides equally;
- reports per-phase **median + the base side's own spread as the noise floor**, and **refuses to call anything a win inside that floor**;
- **fingerprints mesh/vertex/triangle counts** and flags loudly when a "faster" branch changed the output.

Self-tested on identical code (this branch adds only scripts): every phase correctly reported *within noise, byte-identical* — i.e. no false delta. This is the tool the next step (profiling the parse/decode path) runs on.

```
scripts/perf/ab.sh tests/models/ara3d/AC20-FZK-Haus.ifc --iters 7
scripts/perf/ab.sh /abs/model.ifc --base origin/main --iters 5 --json /tmp/ab.json
```

## `tests/models/.manifest-ignore` — redistribution guard
Prerequisite for any corpus-touching harness work. `fixtures:manifest` walks `tests/models/` and would **silently add** any on-disk `.ifc` to the manifest, which `fixtures:upload` then pushes to the **public** release bucket:
- fixtures matching a glob in `.manifest-ignore` are **never** written to the manifest (client models, un-licensed samples, anything whose name must not enter git);
- regeneration prints a **NEW-fixtures warning** for any on-disk fixture not previously manifested — a private drop can't be published silently.

Validated: a `**/*_private.ifc` glob excluded a synthetic client model from the manifest; new public fixtures triggered the warning. `local/` (already skipped) stays the first choice for private files; `.manifest-ignore` covers fixtures a test references at a fixed non-local path.

## Risk
Scripts + one doc file. `manifest.json` intentionally unchanged. No published-package source → no changeset.


louistrue:
Superseded by #1852 (GitHub would not reopen this PR after the base-retarget + rebase onto main). Same branch/content; continuing there.

PR #1855, opened 2026-07-23
fix(viewer): make error tracking group, classify and redact correctly

louistrue:
Triaged the **full PostHog corpus** — 35 issues across the retention window — and found five defects in the exception pipeline itself. None of these are cosmetic: they were losing properties, fragmenting one bug into eleven issues, and leaking client model names.

## 1. `captureException` properties were nested one level too deep

posthog-js spreads its second argument onto the event ([`captureException(error, additionalProperties?: Properties)`](https://posthog.com/docs/error-tracking/capture)). All five call sites passed:

```ts
posthog.captureException(err, { additional_properties: { context, error_kind: kind } });
```

So the event shipped one opaque blob property literally named `additional_properties`. **Verified against live data** — real `$exception` rows carry `additional_properties = {"context":"geometry_processing","error_kind":"geometry_stream_stalled"}` with the top-level `error_kind` empty. Neither field was ever filterable, and `analytics-scrub`'s "don't clobber a kind set at the capture site" guard could never fire, because it never saw one. Flattened at all five sites.

## 2. One bug arrived as eleven issues

PostHog groups by exception type + message unless the client sets `$exception_fingerprint`, which [takes priority over every server-side rule](https://posthog.com/docs/error-tracking/grouping-issues#how-posthog-prioritizes-issue-grouping-logic). Our recognised families embed volatile numbers:

```
Geometry stream stalled after 40000ms. Last rendered meshes: 120070.
Geometry stream stalled after 38374ms. Last rendered meshes: 0.
```

Every distinct mesh count minted its own issue — and its own auto-filed GitHub issue. **11 stall issues + 4 worker-crash issues + 3 OOM issues in one window, for 3 underlying bugs.**

Each classified family now gets one stable fingerprint (`ifc-lite:geometry_stream_stalled`, …). Scoped strictly to the families `classifyLoadError` recognises — unrecognised exceptions keep PostHog's default per-message grouping, so we don't fall into the [over-grouping trap](https://posthog.com/docs/error-tracking/grouping-issues#issue-grouping-best-practices) their docs warn about. The volatile numbers stay in the message and remain queryable inside the issue.

## 3. The privacy guard only walked top-level properties

`scrubProperties` iterated `Object.keys(props)` and only string-tested values, so a nested payload — `{ meta: { file_name: 'Confidential-Tower.ifc' } }` — was **completely invisible** to it. That matters because this guard exists as a net for call sites that don't exist yet.

Now recurses, bounded depth + cycle-safe. **Critically, it does NOT walk into `$exception_list`**: a stack frame carries `resolved_name` / `mangled_name`, which match the sensitive-key rule (`name` as a `_`-delimited word) and would have been **deleted — silently destroying every stack trace we have**. There's a test pinning that.

## 4. Exception messages were never redacted — and four real model names leaked through them

Before the watchdog message was fixed at source, these reached error tracking verbatim:

```
… while loading SV3822-UIH-CN-001-M3D-EST-003-PA.ifc.
… while loading 16201598 Østraadt Havn - Hovedfil BT1 Fabian 2.ifc.
… while loading bwk_rv bn15.ifc.
… while loading Luxembourg_EMEA RTC_SST_EXE_4000.1_TOUS_Maquette HYD.ifc.
```

Messages are now redacted for model-file tokens, keeping the surrounding text so the issue stays triageable:

```
Geometry stream stalled after 40580ms while loading [file]. Last rendered meshes: 0.
```

Getting this right was the fiddly part. Names contain spaces and non-ASCII, so the match can't stop at whitespace — but a naive space-permitting regex eats the entire sentence. The matcher extends leftwards **only across name-ish words** (containing a digit / `_` / `-`, or starting upper-case), so ordinary message words (`while`, `loading`, `after`) end the run. Case-insensitivity is applied per-character rather than with the `i` flag **on purpose**: `/\p{Lu}/iu` case-folds and would match lower-case letters too, turning every word into a name word and swallowing the message. Quantifiers are bounded throughout, with a test asserting no catastrophic backtracking.

## 5. Two production failures were unclassified

| Error | Was | Now |
|---|---|---|
| `NotReadableError: The requested file could not be read…` | `unknown` — user told the model was too big | `file_unreadable` → "the file is no longer available to the browser… please select the file again" |
| `createBuffer failed, size (193836) is too large for the implementation` | `unknown`, untagged | `out_of_memory` |

The WebGPU one is worth calling out: Chromium's wording is misleading. 193 KB is not "too large" for any device advertising hundreds of MB — what actually failed is mapping host memory. Same user guidance as OOM, so same family.

## Plus: two unactionable third-party exceptions dropped client-side

Alongside the existing Cesium filter:
- **Outlook SafeLinks' injected crawler** — `Object Not Found Matching Id:2, MethodName:update, ParamCount:4`. At 18 occurrences this was the **single highest-volume issue in the project**, and it's someone else's crawler.
- **`Script error.`** — the browser's opaque cross-origin report: no file, no line, no stack, information-free by construction (and [known to fragment into hundreds of orphaned PostHog issues](https://posthog.com/questions/critical-bug-error-tracking-issue-fragmentation-single-fingerprint-creating-163-orphaned-issues)). Dropped **only when it carries no stack frames** — if a real one ever arrives with a stack, it's ours and it reports.

## Testing

27 new cases, adversarial rather than happy-path:
- the four real leaked model names, **verbatim from the PostHog corpus**
- over-redaction guards (`Could not parse the .ifc header` must survive intact)
- ReDoS probe on a hostile message
- stack-frame keys must survive the recursion
- self-referencing property object
- noise regexes must not be greedy (`Object Not Found in scene graph` still reports)
- a `Script error.` **with** frames must not be dropped
- each family gets its own fingerprint (no over-grouping); unknown gets none; capture-site fingerprint never overridden

Full viewer suite: **1786 passed, 0 failed**. `tsc --noEmit` clean.

> Note: stack traces in these issues are all unsymbolicated (`Could not find sourcemap for source url`). Source-map upload is a separate PR — I decoded the frames for this triage by pulling the exact deployed bundle from its immutable Vercel deployment using the `app_build_sha` on the events.

### DIFF
--- a/apps/viewer/src/hooks/useClash.ts
+++ b/apps/viewer/src/hooks/useClash.ts
@@ -220,7 +220,7 @@ export function useClash() {
       } catch (err) {
         console.error('[clash] detection run failed', err);
         state.setClashError(err instanceof Error ? err.message : String(err));
-        posthog.captureException(err, { additional_properties: { context: 'clash_detection' } });
+        posthog.captureException(err, { context: 'clash_detection' });
       } finally {
         state.setClashRunning(false);
         state.setClashProgress(null);
@@ -302,7 +302,7 @@ export function useClash() {
     } catch (err) {
       console.error('[clash] duplicate scan failed', err);
       state.setClashError(err instanceof Error ? err.message : String(err));
-      posthog.captureException(err, { additional_properties: { context: 'clash_duplicates' } });
+      posthog.captureException(err, { context: 'clash_duplicates' });
     } finally {
       state.setClashRunning(false);
       state.setClashProgress(null);
--- a/apps/viewer/src/hooks/useIDS.ts
+++ b/apps/viewer/src/hooks/useIDS.ts
@@ -444,7 +444,7 @@ export function useIDS(options: UseIDSOptions = {}): UseIDSResult {
     } catch (err) {
       const message = err instanceof Error ? err.message : 'Validation failed';
       setIdsError(message);
-      posthog.captureException(err, { additional_properties: { context: 'ids_validation' } });
+      posthog.captureException(err, { context: 'ids_validation' });
       console.error('[IDS] Validation error:', err);
       return null;
     } finally {
--- a/apps/viewer/src/hooks/useIfcLoader.ts
+++ b/apps/viewer/src/hooks/useIfcLoader.ts
@@ -1546,9 +1546,9 @@ export function useIfcLoader() {
         // and tag the captured exception so it is filterable in error tracking.
         const kind = classifyLoadError(err);
         setError(formatLoadError(err, file.name));
-        posthog.captureException(err, {
-          additional_properties: { context: 'geometry_processing', error_kind: kind },
-        });
+        // Flat properties: posthog-js spreads this object onto the event, so a
+        // wrapper key would bury `error_kind` in an unfilterable nested blob.
+        posthog.captureException(err, { context: 'geometry_processing', error_kind: kind });
         setLoading(false);
         setGeometryStreamingActive(false);
         return;
@@ -1640,9 +1640,7 @@ export function useIfcLoader() {
         loadError: friendly,
       });
       setError(friendly);
-      posthog.captureException(err, {
-        additional_properties: { context: 'ifc_model_load', error_kind: kind },
-      });
+      posthog.captureException(err, { context: 'ifc_model_load', error_kind: kind });
       setLoading(false);
       setGeometryStreamingActive(false);
     }
--- a/apps/viewer/src/lib/analytics-scrub.ts
+++ b/apps/viewer/src/lib/analytics-scrub.ts
@@ -41,14 +41,16 @@ const PATHISH =
 //
 // Real leaked names contain spaces ("16201598 Østraadt Havn - Hovedfil BT1
 // Fabian 2.ifc"), so the match cannot stop at whitespace — but it must not run
-// away and eat the whole sentence either. It therefore extends leftwards only
-// across NAME-ish words: a word containing a digit, `_` or `-`, or starting
-// with an upper-case letter. Ordinary message words ("while", "loading",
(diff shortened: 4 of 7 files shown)


louistrue:
### Codex review — both findings resolved

**1. `Check DOMException.name for unreadable files`** — valid, fixed in 8bb3e0f.
`classifyLoadError` only ever saw `messageOf(err)`, and a DOMException's `.message` is engine-specific prose that need not repeat the name. `.name` is the stable identity and is now checked first; the message match stays for the analytics path, where all we have is the already-stringified value. Test added with a `NotReadableError` whose message says only "The operation failed".

**2. `Redact lowercase filename prefixes too`** — valid, and it was a real leak. Fixed in the same commit.
The run was keyed on "looks name-ish" (contains a digit/`_`/`-`, or starts upper-case), so `while loading acme tower.ifc` scrubbed to `while loading acme [file]` and still shipped the client name. Inverted to a **stop-word rule**: the run crosses any word except the short list of English connectives that realistically precede a file name. That takes the whole name and removes the need for a case test entirely.

Fixing it surfaced two further defects worth recording:

- A heredoc mangled the regex escapes on the way in (`\b` became a literal backspace), so redaction silently matched **nothing**. Caught by re-probing the real production corpus rather than trusting the test suite.
- Without a leading `\b` the run could start in the **middle** of a stop-word (`loading` → `oading`), sailing past the lookahead and leaking the prefix regardless. Regression test added for exactly that.

Re-probed against the full corpus: all four real leaked names plus the lower-case cases redact to `[file]`, non-file messages are untouched (`Could not parse the .ifc header` survives intact), ReDoS probe 4 ms.

Verification: viewer **1790 passed**, `tsc` clean. Local CodeRabbit CLI re-run on the final tree: **`findings: 0`** (the hosted check was rate-limited, so it isn't a real pass).

PR #1856, opened 2026-07-23
fix(viewer): a failed GPU upload no longer blanks the canvas

louistrue:
Production reported an **unhandled** exception that kills the viewport:

```
RangeError: Failed to execute 'createBuffer' on 'GPUDevice': createBuffer failed,
size (193836) is too large for the implementation when mappedAtCreation == true
```

Chromium's wording is misleading. **193 KB is nowhere near any device limit** — `device.ts` raises `requiredLimits` to the adapter maximum. What actually failed is mapping *host* memory for the new buffer: the device was lost, or could no longer service allocations. The same session also reported `OperationError: Instance dropped in popErrorScope`, which `index.ts` already documents as the signature of a dropped device.

## How I pinned it

Stack traces are unsymbolicated. I pulled the exact deployed bundle for the reporting build (`app_build_sha=9a70a18d`) from its immutable Vercel deployment and read the frames directly:

```
store-CcfUOECO.js:4914  e.appendToBatches        → this.rebuildPendingBatches(n, r);
store-CcfUOECO.js:4925  e.rebuildPendingBatches  → r.batchedMesh = this.createBatchedMesh(...);
store-CcfUOECO.js:5410  e.createBatchedMesh      → l ? (c = n.createBuffer({ ... mappedAtCreation: !0 })
```

and the frames **above** `appendToBatches`:

```
index-B42fho_u.js:5449/5457/5491   React commitPassiveMountEffects recursion + Fc(9, t)
index-B42fho_u.js:59072            t && (s ? (u.queueMeshes(g), I()) : (u.appendToBatches(g, i, t, !1), e.clearCaches()))
```

That last line is verbatim `useGeometryStreaming.ts:375`. So the throw escaped through **React's commit phase** — with no error boundary around the viewport, React unmounts the tree and the `<canvas>` goes with it.

### 1. The renderer kept handing out a dead device

A lost device is never torn down — `WebGPUDevice.destroy()` is the only thing that nulls the handle, and it is **not** called for an involuntary loss (Windows TDR / GPU-process crash / driver reset). So `isInitialized()` stayed `true` and `getGPUDevice()` returned the zombie `GPUDevice`. Callers kept calling `createBuffer()` on it, bypassing both `render()`'s own `deviceLost` guard **and** every `onDeviceLost` listener.

It now returns `null` when the device is lost, which routes into the `if (!device) return` check every call site already has. (Also worth knowing: nothing in `apps/viewer` currently subscribes to `renderer.onDeviceLost(...)` at all — the recovery API exists but is unwired. Out of scope here; this PR makes the unwired case degrade safely instead of crashing.)

### 2. Seven GPU-upload call sites were unguarded

Where the throw lands is what makes it fatal:

| Site | Landing | Consequence |
|---|---|---|
| `useGeometryStreaming` `appendToBatches` | React passive effect | tree unmounts → **blank canvas** (the reported crash) |
| `useAnimationLoop` `flushPending` | rAF callback | skips the tail-position `requestAnimationFrame(animate)` → **rendering stops permanently** |

Both failure modes were **already understood in this file**. The residency-restore block directly below `flushPending` carries a `try/catch` and the comment *"an uncaught throw here (e.g. buffer creation on a lost device) would kill the rAF loop and blank the canvas"*; the instanced-shard effect carries an equivalent one. The remaining sites simply never got the same treatment.

## Change

Adds `runGpuUpload(label, fn)`, which contains the throw, keeps a console breadcrumb **de-duplicated per call site**, and reports **once per session** to error tracking — these failures repeat every frame and would otherwise flood both the console and the exception quota. A batch that fails to upload now costs that batch, not the session.

Applied to all seven sites: `flushPending` (rAF + queue pump), `appendToBatches`, `rebuildPendingBatches` (removals / translations / rotations), `finalizeStreaming` (translations / rotations).

## Testing

8 new cases, aimed at the ways this could go wrong rather than the happy path:
- **the falsy-return trap** — `flushPending` returns a boolean the caller acts on, so `false` must stay `false` and must not read as "it threw"
- containment returns `undefined` rather than rethrowing
- the guard keeps working after a failure (one bad batch ≠ dead session)
- warn-once-per-site across 50 consecutive failures, but separately per site
- non-`Error` throwables (`throw 'string'`, `throw null`)

Viewer suite **1772 passed**, renderer suite **300 passed**, `tsc --noEmit` clean. Changeset included for `@ifc-lite/renderer` (patch).

> Pairs with #1855, which classifies this error into the `out_of_memory` family so it groups and gets the right user guidance. The two are independent and can land in either order.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented rendering errors after unexpected GPU device loss by safely stopping GPU uploads.
  - Kept the viewer and animation loop running when GPU uploads fail.
  - Added user notifications and diagnostics for GPU upload failures.
  - Preserved pending geometry changes so failed uploads can be retried.
  - Prevented streaming finalization failures from leaving the scene blank or partially rebuilt.

### DIFF
--- a/.changeset/renderer-lost-device-returns-null.md
+++ b/.changeset/renderer-lost-device-returns-null.md
@@ -24,3 +24,6 @@ empty — array. Callers that contain the throw to keep the canvas alive would
 therefore have turned a crash into a silently blank model. On failure the
 previous drawables are restored (their GPU resources are still live, since the
 destroy step never runs) and the error is rethrown for the caller to handle.
+Batches the failed attempt had already created are freed, while anything that
+predates it — including cold shells aliased into both the old and the new array
+— is left alone.
--- a/.changeset/renderer-lost-device-returns-null.md
+++ b/.changeset/renderer-lost-device-returns-null.md
@@ -1,0 +1,18 @@
+---
+'@ifc-lite/renderer': patch
+---
+
+`Renderer.getGPUDevice()` now returns `null` once the GPU device has been lost,
+instead of handing back a zombie device.
+
+A lost device is not torn down: `WebGPUDevice.destroy()` is the only thing that
+nulls the handle, and it is never called for an involuntary loss (Windows TDR,
+GPU-process crash, driver reset). `isInitialized()` therefore stayed `true` and
+the accessor kept returning the dead `GPUDevice`, so callers went on calling
+`createBuffer()` on it — throwing a `RangeError`, and bypassing both
+`render()`'s own `deviceLost` guard and every `onDeviceLost` listener.
+
+Returning `null` routes into the `if (!device) return` check that call sites
+already have, so a lost device degrades to "stop uploading" rather than an
+uncaught throw. `isDeviceLost()` / `onDeviceLost()` are unchanged and remain the
+recovery contract.
--- a/.changeset/renderer-lost-device-returns-null.md
+++ b/.changeset/renderer-lost-device-returns-null.md
@@ -16,3 +16,11 @@ Returning `null` routes into the `if (!device) return` check that call sites
 already have, so a lost device degrades to "stop uploading" rather than an
 uncaught throw. `isDeviceLost()` / `onDeviceLost()` are unchanged and remain the
 recovery contract.
+
+`finalizeStreaming()` is now rollback-safe. It detaches the old fragments and
+batches before the replacement GPU buffers exist, so a `createBuffer` failure
+part-way through the rebuild left the scene rendering a half-built — often
+empty — array. Callers that contain the throw to keep the canvas alive would
+therefore have turned a crash into a silently blank model. On failure the
+previous drawables are restored (their GPU resources are still live, since the
+destroy step never runs) and the error is rethrown for the caller to handle.
--- a/apps/viewer/src/components/viewer/gpu-upload-guard.test.ts
+++ b/apps/viewer/src/components/viewer/gpu-upload-guard.test.ts
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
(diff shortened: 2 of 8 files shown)


louistrue:
### Codex review — finding resolved

**`Keep finalization failures from blanking the scene`** — valid, and the sharpest catch on this PR.

Confirmed against the code: `finalizeStreamingInner` detaches the old drawables (`streamingFragments = []`, then `batchedMeshes = []`) **before** `rebuildPendingBatches` creates the replacement GPU buffers. On its own that was a crash; combined with the containment in this PR it became worse *in kind* — the caller swallows the throw to keep the canvas alive, so the user would sit in front of a silently blank model instead of a hard failure.

Steps 1-4 now run under a rollback: on failure the previous fragments and batches are put back and the error is rethrown. That's safe because the destroy step never ran, so every old GPU resource is still live and restoring the arrays restores exactly what was on screen.

CodeRabbit then pushed back on my first attempt, which accepted "a bounded leak" of the partially-rebuilt batches, and it was right that the two goals aren't actually in tension. The hazard is only **aliasing**: carried cold shells appear in both the old and the new array, so destroying the partial array wholesale would leave the restored arrays pointing at dead buffers. Excluding anything already live before the rebuild frees exactly the batches this attempt created and nothing else — no leak, no dangling reference. That matters when the allocation failure is transient rather than a lost device, where the leak would otherwise persist for the session.

Scoped deliberately to the **synchronous** path, which is the one this PR contains. `finalizeStreamingAsync` has the same shape but no guarded caller, so I left its behaviour unchanged rather than altering it as a side effect — flagging it as a known follow-up rather than silently widening this PR.

4 regression tests, exercising a realistic *partial* rebuild: the failure case installs a carried cold shell plus a newly built batch before throwing and asserts only the new one is freed; the success case asserts the replacement survives while both old drawables are freed; plus in-progress-flag cleanup.

Verification: renderer **304 passed**, viewer **1772 passed**, `tsc` clean. Local CodeRabbit CLI re-run on the final tree: **`findings: 0`** (the hosted check was rate-limited, so it isn't a real pass).

PR #1858, opened 2026-07-24
fix(viewer): stop auto-recovered wasm version-skew from polluting error tracking

louistrue:
## What

The two active WASM-load error-tracking issues (15 occurrences, 8 users) are **noise for a failure the user never actually experiences.**

When a tab open across a deploy hits the rotated `ifc-lite_bg-<hash>.wasm` (404 → served as `text/plain`), the skew handlers already **auto-reload** onto fresh assets and the engine loads fine. But the exception is captured — by PostHog autocapture, or our own `captureException` in the load catch — **before** the reload navigates away. So a 100%-recovered condition still files an `$exception`, and every rotated hash mints more.

Both the finder and verifier in the triage flagged this: *"the error is captured before the reload runs, so this cluster keeps appearing even on 100% successful auto-recovery — its occurrence count cannot be used to judge whether users are stuck."*

## How it distinguishes recovered noise from a genuine block

The one thing I must **not** do is hide a genuinely-stuck deploy (new assets also unreachable — a CSP-blocked worker, a proxy rewriting every wasm to `text/plain`). The distinguisher is **recurrence**:

- **Recovered skew** → fires once or twice around the reload, then gone.
- **Genuine block** → keeps firing, because the reload never fixes it.

So `before_send` suppresses only the first few skew exceptions within a short window, then lets the rest surface. A decaying counter (5-min window) means a genuinely-new skew later in a long session (a second deploy) gets a **fresh** budget rather than being mistaken for a persistent block.

Keyed on the existing `isWasmAssetUnavailableError` signature — which already matches exactly the two active messages and **excludes** genuine `CompileError` / corruption (those still surface).

## Safe defaults (the part I care most about)

- A non-skew or non-`$exception` event is **never** dropped.
- If the counter can't be persisted (blocked storage: private mode / sandboxed embed), we do **not** suppress — verified by reading the write back. A storage-blocked tab surfaces the error for a manual reload (it never auto-reloads either, per the existing reload veto), so its failures must not be silently swallowed. My first draft got this wrong — it would have suppressed forever on blocked storage — and the test I wrote for that case caught it before commit.

Wired in `analytics.ts` (composed **before** the scrub) so `analytics-scrub.ts` stays dependency-free (no `@ifc-lite/geometry` import) and independently unit-testable.

## Testing

8 new cases: both active signatures, the recovered-vs-persistent split, the decay window (fresh budget after 5 min; no decay within it), blocked-storage → surfaces, `$exception_values` fallback, and non-skew passthrough. Full viewer suite **1826 passed**, `tsc` clean.

Resolves the noise in the consolidated `wasm_engine_load` issue while preserving genuine-block visibility.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced telemetry noise by suppressing repeated, auto-recovered WebAssembly version-skew `$exception` events, while allowing other error types and non-`$exception` events through.
  * Improved the event processing flow so suppression is applied before existing privacy scrubbing.
  * Added durability safeguards so suppression won’t get “stuck” when browser storage isn’t writable or readable.
* **Tests**
  * Added automated coverage for suppression thresholds, decay timing, `$exception` payload variations, and browser storage edge cases.

### DIFF
--- a/apps/viewer/src/lib/analytics.ts
+++ b/apps/viewer/src/lib/analytics.ts
@@ -4,6 +4,20 @@
 
 import posthogClient from 'posthog-js';
 import { scrubEvent } from './analytics-scrub.js';
+import { shouldSuppressWasmSkewNoise } from './wasm-version-skew.js';
+
+// `before_send` gate: drop the noise from an auto-recovered wasm version-skew
+// (the tab reloads onto fresh assets, so the captured exception describes a
+// failure the user never actually hits — see shouldSuppressWasmSkewNoise),
+// then run the privacy/tagging scrub on everything that remains. Kept here
+// rather than inside scrubEvent so analytics-scrub.ts stays dependency-free
+// (no @ifc-lite/geometry import) and independently unit-testable.
+const beforeSend = <
+  T extends { event?: string; properties?: Record<string, unknown> } | null,
+>(event: T): T | null => {
+  if (shouldSuppressWasmSkewNoise(event)) return null;
+  return scrubEvent(event);
+};
 
 // `import.meta.env` is undefined under the Node test runner (no Vite define
 // plugin), and this module is loaded transitively by most viewer tests. The
@@ -39,7 +53,7 @@ if (enabled) {
       // (current + future) before it leaves the browser, drop unactionable
       // third-party noise, and tag the geometry error family. See scrubEvent
       // in ./analytics-scrub.ts.
-      before_send: scrubEvent,
+      before_send: beforeSend,
     });
     // Register build attribution as super-properties so every event (incl.
     // ifc_model_loaded) carries the deploy it was served from — this is what
--- a/apps/viewer/src/lib/wasm-skew-noise.test.ts
+++ b/apps/viewer/src/lib/wasm-skew-noise.test.ts
@@ -15,16 +15,20 @@ function harness(startMs = 1_000_000): {
   deps: SkewNoiseDeps;
   advance: (ms: number) => void;
   store: Map<string, string>;
+  setReloaded: (v: boolean) => void;
 } {
   let t = startMs;
+  let reloaded = false;
   const store = new Map<string, string>();
   return {
     store,
     advance: (ms) => { t += ms; },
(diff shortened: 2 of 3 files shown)


louistrue:
### Review feedback resolved — both findings were valid

**1. CodeRabbit (major): `Verify timestamp persistence before suppressing`**

The read-back guard only confirmed the *count* write. Partial persistence is the subtler runaway: with no durable `lastTs`, every later call sees the episode as decayed, resets the count to `1`, and would suppress forever — reaching the same failure through the decay branch instead of the counter. Both read-backs now confirm, with a regression test for the timestamp-only failure alongside the fully-blocked-storage case.

**2. Codex (P2): `Surface the post-reload wasm failure`** — this one broke the guarantee the whole gate is built on.

A genuinely broken deploy produces exactly two events: the stale tab's original skew (suppressed, we reload) and the post-reload failure when the new assets are *also* unreachable (CSP-blocked worker, a proxy rewriting every wasm). Both fell inside the count budget, so **both were suppressed** — and a stuck user typically hits this once and gives up rather than retrying enough times to exhaust the budget. Error tracking would have missed exactly the stuck-deploy case this gate exists to catch.

The decisive signal was already in the module: `recentlyReloaded`. If a recovery reload has run and we're *still* seeing the error, the reload didn't fix it → **surface immediately**, rather than spending budget on it. The count cap now only does what it's genuinely needed for: capping a pre-reload flood when several workers throw the same skew error in the tick before the reload navigates away.

The test is rewritten around the real scenario (pre-reload suppressed → post-reload surfaces), with the flood cap covered separately.

Net contract now:
| Case | Behaviour |
|---|---|
| Recovered skew (pre-reload) | suppressed — no noise |
| **Stuck deploy (post-reload)** | **surfaces on the first recurrence** |
| Pre-reload flood | capped after 3 |
| Blocked / partial storage | never suppressed |
| Non-skew, incl. genuine `CompileError` | never suppressed |

Local CodeRabbit CLI on the final tree: **`findings: 0`**. Full viewer suite **1828 passed**, `tsc` clean.

PR #1859, opened 2026-07-24
feat(viewer): retry a resource-limit load once at lower detail before giving up

louistrue:
## What

The three biggest error-tracking families — `geometry_stream_stalled`, `geometry_worker_crash`, `out_of_memory` (**34 occurrences, ~25 users**) — are resource-exhaustion on models too heavy for the device's memory. Today they dead-end with a "close tabs / load something smaller" message and **the user gets nothing.**

But the engine can tessellate the *same* model far more cheaply. The `lowest` tier drops parametric fillets to sharp corners and coarsens curves, cutting triangle count — and peak memory — dramatically. A model that OOM'd at the default density has a real chance to load at `lowest`.

So on one of these failures we now **retry the same in-memory file once at `lowest`** before surfacing the error. *Something at reduced detail beats nothing.*

## Why this design (and not the obvious alternatives)

- **Not a reload.** A reload would lose the freshly-picked `File` (browsers can't persist a File handle across navigation). We re-invoke `loadFile` in-memory, keeping the file.
- **Not new UI.** It's the same bounded-once auto-recovery pattern the WASM-skew reload already uses — an informative toast, no button, no error-panel surface to build.
- **Straight to `lowest`, not stepping down.** One retry at the maximal memory reduction, rather than grinding the user through medium → low → lowest across several failed full loads of a huge model.

## Guard rails (the pure, tested part)

The decision is a pure function (`resource-retry.ts`); the `loadFile` catch is a thin call site. It retries **only** when all of these hold:

| Guard | Why |
|---|---|
| kind ∈ {stall, worker-crash, OOM} | a parse error / cancel / engine-load failure isn't helped by less detail |
| the WASM tessellation path actually ran (`attemptedTessellationTier` set) | a GLB / point-cloud / server / cache load isn't re-tessellated by our engine, so a lower IFC tier can't help it — don't repeat the work |
| primary load | a federated add must never silently downgrade the whole federation |
| not already a retry (`isResourceRetry`) | a second failure at `lowest` surfaces normally instead of looping |
| not already at `lowest` | nowhere lower to go |

## Safety / no-regression reasoning

- The retry bumps the load session and re-meshes at the lower density — the tier is folded into the cache key, so it **never** serves the failed attempt's cached bytes.
- The failed attempt's geometry and workers are torn down as the catch unwinds, freeing the peak the retry needs.
- **Worst case** (`lowest` also fails): surfaces exactly as before — no regression, no loop.
- The original failure is still captured (with a `resource_retry` property) so the memory wall stays visible in analytics; the retry just gives the user a shot at a result first.

## Testing

8 pure-decision tests covering each guard and their composition (every disqualifier wins over a qualifying kind). Full viewer suite **1826 passed**, `tsc` clean.

Improves the user outcome for the consolidated stall / worker-crash / OOM issues. Pairs with #1858 (WASM-skew noise); independent.

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery for a subset of model-loading failures by retrying tessellation at a lower quality level.
  * Shows a user notification when a resource-based retry is triggered.
  * Ensures retries don’t leave the app stuck in an active streaming state and maintains consistent geometry across quality levels.
* **Bug Fixes**
  * Prevents unintended repeated retries by applying strict eligibility rules for when an auto-retry can occur.
* **Tests**
  * Added tests covering retry eligibility, supported failure types, quality-tier boundaries, and key guard conditions.

### DIFF
--- a/apps/viewer/src/hooks/useIfcLoader.ts
+++ b/apps/viewer/src/hooks/useIfcLoader.ts
@@ -29,7 +29,9 @@ import {
   type MeshData,
   type CoordinateInfo,
   type GeometryResult,
+  type TessellationQuality,
 } from '@ifc-lite/geometry';
+import { resolveResourceRetryTier } from '../lib/resource-retry.js';
 import { acquireFileBuffer, type AcquiredBuffer } from '../utils/acquireFileBuffer.js';
 import { buildSpatialIndexGuarded, buildSpatialIndexForModel } from '../utils/loadingUtils.js';
 import { buildGeometryCacheKey } from './geometryCacheKey.js';
@@ -58,7 +60,7 @@ import { extractModelGeoref, alignGeometryToReference, findReferenceGeorefModel 
 import { toast } from '../components/ui/toast.js';
 import { posthog } from '../lib/analytics.js';
 import { reportRenderStats } from '../utils/renderStatsReport.js';
-import { classifyLoadError, formatLoadError } from '../lib/load-errors.js';
+import { classifyLoadError, formatLoadError, type LoadErrorKind } from '../lib/load-errors.js';
 
 /**
  * The skip-tiny-cuts flag is no longer a hard constant: it is derived per-load
@@ -182,7 +184,18 @@ export function useIfcLoader() {
   // Latest `loadFile`, so the background revalidation can reload without being a
   // dependency of `loadFile` itself (avoids a definition cycle). Kept current by
   // the effect below.
-  const loadFileRef = useRef<((file: File, target?: LoadTarget) => Promise<void>) | null>(null);
+  const loadFileRef = useRef<
+    | ((
+        file: File,
+        target?: LoadTarget,
+        options?: {
+          sourceHandle?: FileSystemFileHandle;
+          tierOverride?: TessellationQuality;
+          isResourceRetry?: boolean;
+        },
+      ) => Promise<void>)
+    | null
+  >(null);
 
   /**
    * Background revalidation for a SERVED source-decoupled (mesh-only) cache hit:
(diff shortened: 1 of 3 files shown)


louistrue:
### Review feedback resolved

Both bots independently flagged the same **major** issue, and they were right — it would have made this PR dead code for the two families it exists for.

**`Put resource retries in the geometry error path` / `Run resource recovery from the geometry-processing catch as well`**

`geometry_stream_stalled` and `geometry_worker_crash` are thrown *inside* the geometry-streaming block, which has its own inner catch (it must close the WASM iterator and swallow the orphaned parser promise). That catch `return`s rather than rethrowing, so those failures never reached the outer catch where the retry originally lived. Only failures raised outside the stream would ever have retried.

Fixed by extracting the recovery into a single `tryResourceRetry` helper called from **both** catches, so the policy lives in one place and the streaming failures actually reach it. The capture context is threaded through, so analytics still attribute the failure to where it happened (`geometry_processing` vs `ifc_model_load`).

Also from the review: the retry is now **awaited** rather than fire-and-forget. Callers await `loadFile` to know the load finished, so the original promise must stay pending until the replacement settles (`loadFile` never rethrows — both its catches return — so this can't throw back into the caller). Hardened the toast's dynamic import with a `.catch` so a failed chunk load can't become an unhandled rejection that masks the retry.

Local CodeRabbit CLI on the final tree: **`findings: 0`**. Full viewer suite **1826 passed**, `tsc` clean.

> Housekeeping: an earlier commit accidentally swept three stray `.pr1761-*.png` screenshots (~860 KB) into the branch via `git add -A`. They're gone — the branch is now exactly 3 files. They were never on `main` or in any merged PR.

PR #1877, opened 2026-07-25
fix(clash): report mesh-level penetration depth for contained contact pairs

### LINKED ISSUE #1866: Clash engine reports AABB signed gap as penetration depth for contained contact pairs
Found while adversarially validating a generated model whose openings/fills are designed face contacts (inset 2 mm), on branch `worktree-moonshot-tech-2026-07-24` (scripts/moonshot/diff-spike/optimize.mjs).

## Symptom

For a contact pair where one element's AABB is fully contained in the other's, the clash finding's `distance`/depth field reports the AABB signed gap rather than the actual mesh penetration depth. This overstates depth: pairs whose true interpenetration is at the 1e-5..1e-6 m level (inside `TOUCHING_EPSILON = 1e-4` and correctly classified as touching/contact) carry a reported depth that reflects how deep the small AABB sits inside the big one, not how far the meshes actually overlap.

## Observed

21 designed face contacts with true worst depth 7.39e-6 m; the reported depth for contained pairs tracked the AABB containment offset instead. Classification (hard/touching) was still correct in this corpus because the contact-band gate uses the right quantity; only the reported magnitude is misleading, which matters for depth-sorted triage UIs and BCF export.

## Repro pointer

On the moonshot branch: `node scripts/moonshot/diff-spike/optimize.mjs` generates the optimum IFC and runs `ifc-lite clash --mode hard`; inspect the 21 contact findings' depth vs the known 2 mm inset geometry.

## Suggested fix

For pairs flagged by the narrowphase, derive depth from the mesh-level penetration (max point-to-surface among intersecting triangles or the exact CSG overlap extent) instead of the AABB signed gap when one box contains the other.

louistrue:
Fixes #1866

## Problem

For a contact pair where one element's AABB is fully contained in the other's, the hard-clash finding's `distance` reported the AABB signed gap - how deep the small **box** sits inside the big one, which for a contained box is its own smallest-axis extent. Designed face contacts (opening fills inset 2 mm in their host, true interpenetration 1e-5..1e-6 m) carried centimetre-to-decimetre depths, corrupting depth-sorted triage UIs and BCF export magnitudes. Classification (hard/touch via the contact band) was already correct; only the reported magnitude was wrong.

## Fix

In both kernels (TS `engine-ts/narrow.ts` and the bit-identical Rust/WASM port `rust/clash/src/narrow.rs`):

- detect the contained-AABB case up front and collect the triangle pairs that genuinely cross,
- report depth as the max point-to-surface distance over crossing-triangle vertices that lie inside the other solid (deduped by vertex, iterated in index order so TS and Rust stay bit-identical; `distanceToSurface` is an exact point-to-triangle minimum over the mesh),
- fall back to the previous AABB estimate when no crossing vertex lies inside (e.g. a thin member piercing straight through, whose crossing-triangle vertices all sit outside the other solid).

Non-contained pairs are untouched, as are classification, contact bounds and points.

## Tests

- `packages/clash/src/engine-ts/engine.test.ts`: new `#1866` suite with a concave L-prism + contained box. Macro case: expected depth `1 - xMin` (~0.05 m) asserted to 1e-9, where the old AABB gap was 0.7 m (14x overstated). Micro case at the issue-corpus scale (~1e-6 m, inside `TOUCHING_EPSILON`) asserted to 1e-12.
- `packages/clash/src/differential.test.ts`: contained-pair case pinning TS/WASM parity through the new mesh-depth path.
- `rust/clash/src/tests.rs`: `contained_pair_reports_mesh_depth_not_aabb_gap`, same fixture, depth asserted to 1e-9.

## Verification

- `pnpm --filter @ifc-lite/clash test` (vitest): 17 files, 140 tests passed, including the differential suite against a freshly built WASM artifact.
- `cargo test --workspace --no-fail-fast`: green. (One pre-existing infra hiccup: `wall_opening_cut_regression` needs the ara3d `advanced_model.ifc` fixture, whose download timed out in a fresh worktree; after staging the fixture the target passes 7/7. Unrelated to this change.)
- Typecheck: `pnpm --filter "@ifc-lite/clash..." build` (tsc) clean.
- Changeset: patch for `@ifc-lite/clash` + `@ifc-lite/wasm`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved clash penetration-depth reporting for contained or partially overlapping mesh contacts.
  - Depth now reflects actual mesh geometry instead of potentially overstating overlap based on bounding boxes.
  - Improved accuracy for face-like contacts, openings, concave surfaces, and very small penetrations.
  - Added regression coverage across TypeScript and WebAssembly clash detection paths.


louistrue:
@coderabbitai review

Issue #595, opened 2026-04-28
3D World Context (Cesium) shows many artifacts


semhustej:
When trying to view 3D World Context with specified georeferenced file, I get weird artifacts in 3D view. The result is that it is not usable for viewing.
The georeferencing data is displayed correctly in the 2D Map bellow georeferencing information.

Sample file (remove the .txt extension):

[test-cesium.ifc.txt](https://github.com/user-attachments/files/27160023/test-cesium.ifc.txt)

IFClite version: v1.17.6 · Apr 25, 2026

See the behavior in the video:

https://github.com/user-attachments/assets/00a994b1-27ad-4e0d-88f9-f669ba9d1a49

louistrue:
wow, haven't seen this one before 😅 thx, @semhustej. Will look into it!

semhustej:
Hi @louistrue I think I found the culprit of the issue. It seems to me IfcLite does not take the `Scale` attribute of the IfcMapConversion.

The OP file uses `mm` as default project units and the Porjected CRS uses `m`. So the `Scale` is set to `0.001`. This results in the issue described in the original post.

When I change the `Scale` value to nothing, `1` us used and the model context is loaded correctly in IfcLite. I tested this on another file too.

Note that I am using the `Scale` attribute as recommended by Bonsai, I did not check directly in IFC schema. But to me, Bonsai is usually as reliable as IFC schema documentation :)

louistrue:
Thx for finding the root problem, @semhustej 👍 fixing now:

<img width="808" height="545" alt="Image" src="https://github.com/user-attachments/assets/2cce60b1-8196-4e17-9f44-7cef7641bad0" />

louistrue:
done 👍

louistrue:
thx, @semhustej

semhustej:
Thanks, @louistrue . My testing shows now the scale is taken into account correctly. Having warning that when the scale seems to be not setup correctly is nice too.

louistrue:
thanks for the feedback, @semhustej glad at least _someone_ noticed 😉

PR #1878, opened 2026-07-25
feat(pointcloud,renderer,viewer): align georeferenced point clouds to IfcMapConversion (#1804)

### LINKED ISSUE #1804: Pointcloud feature (alignment with ifcmapconversion)
### The problem

Pointcloud are geospatial layers. To make them more useful it should respect the IfcMapConversion values used to position the IFC model.   When a scan is displayed using absolute XYZ coordinates — meaning without applying (contra) translation and rotation defined in IfcMapConversion — the pointcloud will not align with the IFC model.

Having support for this transformation would be highly beneficial.

Image

### Proposed solution

By default, the viewer should apply the translation and rotation defined by the IFC import (IfcMapConversion).
A separate toggle should be available to turn this transformation on or off when needed.

### Alternatives considered

_No response_

louistrue:
Closes #1804

## Root cause

`.las`/`.laz` scans carry **absolute projected-CRS coordinates** (eastings / northings / orthogonal height, ~1e6–1e7 m). A georeferenced IFC model places its local engineering frame in that same CRS via `IfcMapConversion`, but the viewer rendered scans at their raw coordinates — so scan and model never overlapped. Two hard constraints shape the fix:

1. **f64 before f32.** The decode pipeline narrows positions to `Float32Array` for the GPU. At 2.6e6 m (Swiss LV95) the f32 ulp is ~0.25 m — any subtraction done *after* narrowing inherits that quantisation. The map-conversion offset must be subtracted **inside the decoder, in f64**.
2. **The GPU uniform matrix is f32 too.** A ~1e6-magnitude translation in the per-asset matrix quantises just the same, so the matrix must never carry map-magnitude translations.

## What this PR does

**Coordinate contract (plain terms):** at scan load time we compute, from the federation anchor's `IfcMapConversion`, (a) `decodeOriginOffset` — the map-space position (in the CRS's **native unit**) of the point that renders at the viewer origin, subtracted from every LAS/LAZ coordinate in f64 inside the decode worker before f32 narrowing; and (b) an f32 GPU matrix holding only rotation + scale (`k = mapUnitScale / effectiveScale`, inverse `XAxisAbscissa/XAxisOrdinate` rotation, Z-up→Y-up) with **exactly zero translation** — the entire viewer shift (`originShift` + `wasmRtcOffset`, the same `totalYupOffset` federation alignment uses) is folded into (a), where it lives in f64. A second "unaligned" matrix adds the decode offset back, bug-compatibly reproducing the raw pre-#1804 placement, so the panel toggle ("Align to model georeference", ON by default per the issue) flips between the two without re-streaming.

The math reuses the repo's canonical helpers (`getEffectiveHorizontalScale` / `resolveMapUnitToMetreScale` from `lib/geo/geo-scale.ts` — issue #595 double-scale semantics included — and `totalYupOffset` from `lib/geo/ifc-origin.ts`); the spec-level forward/inverse map conversion mirrors `rust/core/src/georef.rs` `local_to_map`/`map_to_local` and is unit-tested against worked examples.

## Defects found during adversarial review (all fixed here)

The branch was inherited mid-flight; review against the codebase's known failure modes found five real defects:

1. **f32 translation collapse** (`pointCloudAlignment.ts`): the aligned matrix carried `-totalYupOffset` as an f32 translation. For files authored at map magnitude (RTC offset ~2.6e6 m) that quantises to ~0.25 m — defeating the feature for exactly the Swiss-LV95-class files that need it. Fixed by the zero-translation fold; a test asserts the translation column is **exactly** 0 with a 2.6e6 m RTC offset.
2. **Unit mixing on non-metre CRSs** (`pointCloudAlignment.ts`): `decodeOriginOffset` was premultiplied to metres but subtracted from LAS coordinates stored in native CRS units (wrong for an explicit foot-based `MapUnit`). Now native-unit offset + `k = mapUnitScale/effectiveScale` in the matrix; foot-CRS test added.
3. **Wrong-frame deviation compute** (`packages/renderer/src/index.ts`, `deviation-*.ts`): the BIM↔scan deviation pass queried the world-space triangle BVH with untransformed chunk-local points — garbage distances for aligned scans (the headline use-case). The compute shader now applies the same per-asset matrix as the splat shader; params packing unit-tested.
4. **Stale bounds** (`point-cloud-renderer.ts` `getBounds()`): height-ramp min/max and scene framing used untransformed extents. Now folds each node's matrix (8-corner AABB transform, unit-tested); `setPointCloudTransform` re-derives scene bounds so toggling updates framing.
5. **Misaligned-by-default non-LAS formats** (`pointCloudIngest.ts`): the aligned matrix was applied to PLY/PCD/E57/PTS/XYZ streams whose decoders never consumed the decode offset — rotating *absolute* coordinates, strictly worse than raw, and ON by default. Alignment is now gated to LAS/LAZ with a user-facing notice.

Plus toggle-state hygiene (availability derived from the live registry, so no dead toggle after removing the last aligned scan) and an ordering toast: loading an IFC **after** a scan explains that alignment is baked at scan load time and the scan should be reloaded.

## Adversarially tested

- No `IfcMapConversion` / synthesised site-location georef → no alignment, raw behaviour, no toggle (`findReferenceGeorefModel` + `hasStandardGeoreferencing` gate).
- `Scale` absent (→1 via effective-scale heuristics) and `Scale = 0` / degenerate `XAxisAbscissa=XAxisOrdinate=0` → alignment refused (`null`), scan loads raw (tests).
- Non-unit axis direction vectors → normalised (test).
- Scan **before** model (ordering) → raw + explanatory toast on model load; scan after model → aligned. Multiple scans share the registry; per-handle unregister on remove/reload/error paths.
- Revert check executed: removing the decoder's f64 `originOffset` subtraction makes `las.test.ts` fail (`expected 500012.34 to be close to 12.345`); the zero-translation and behaviour-oracle tests likewise pin the fold and the full decode→swap→matrix chain.
- LAS header scale/offset factors: covered by the f64-precision decode test (UTM-scale header offset, mm-scale point residuals preserved to µm).

## Gate results (actual)

- `pnpm typecheck` — **33/33 tasks successful**
- `pnpm --filter @ifc-lite/pointcloud test` — **14 files, 119 passed, 2 skipped**
- `pnpm --filter @ifc-lite/renderer test` — **314 pass, 0 fail**
- `pnpm --filter @ifc-lite/viewer test` — **1848 pass, 0 fail, 2 skipped**
- `node scripts/check-test-wiring.mjs` — ✅
- `pnpm check:api-surface` — ✅ (37 packages, 3945 exports, snapshot unchanged)
- Changeset: `minor` for `@ifc-lite/pointcloud` + `@ifc-lite/renderer` (additive optional fields/methods only).

## Known limitations

- **LAS/LAZ only.** Other formats' decoders don't consume the f64 decode offset yet; they load raw with a notice. Threading the offset through PLY/E57/etc. is a straightforward follow-up.
- **Alignment is baked at scan load time** against the then-current federation anchor. Loading the scan first, or swapping the anchor model afterwards, requires reloading the scan (surfaced via toast; retro-applying a matrix to already-quantised f32 positions would give ~0.5 m error, which we refuse to ship silently).
- Height uses the same uniform scale as the spec (`Scale` applies to x, y and z, matching `rust/core/src/georef.rs`); no vertical datum transform, same as the rest of the viewer.

https://claude.ai/code/session_01M76mvdz8Hz7WHEoxwn694V


louistrue:
Hosted CodeRabbit was rate-limited (its "pass" is a false pass), so I ran the CodeRabbit CLI locally against origin/main (23 files reviewed, 2 findings):

- **major** — `Renderer.setPointCloudTransform` recomputed model bounds but didn't push them via `camera.setSceneBounds()` like every other bounds-mutating point-cloud method → **fixed in 7688d5be**.
- **minor** — suggested per-handle try/catch inside `applyPointCloudAlignmentToggle`'s loop. Skipped: `setAssetTransform` is a guarded `Map` lookup + field assignment with no throwing path, and a speculative logged catch per iteration is defensive clutter.

louistrue:
@coderabbitai review

louistrue:
@coderabbitai review

louistrue:
Local CodeRabbit CLI review of the rebased head (first genuine review this PR has had)

The hosted CodeRabbit check on this PR has been green throughout, but there has never been an actual hosted review — every attempt hit the Fair Usage limit, so the green was a false pass. I ran `coderabbit review --committed --base main --agent` locally against the rebased head: **2 findings, both real, both fixed** in `bb30b48a`.

**1. `packages/pointcloud/src/streaming/las-source.ts` — bbox reported in a different frame from the points.** `toInfo()` returned `header.bbox` verbatim, but every emitted point has `originOffset` subtracted in f64 first. Bounds and points were therefore off by the full offset. At map magnitudes (LV95 easting ~2.6e6) that puts framing, culling and the height-ramp colour range nowhere near the geometry. Flagged 'minor', but it is precisely the frame-consistency class this issue exists to eliminate. The bbox is now translated by the same offset on the same axes, and the new test cross-checks the invariant directly — every emitted point must fall inside the reported box.

**2. `point-cloud-node.ts` / `point-cloud-uniforms.ts` — two independent model-matrix guards, neither finite-checked.** `transformAabb()` and `writePointCloudUniforms()` each carried their own `length === 16` test. Duplicated guards drift, and a matrix one accepted while the other rejected would put the reported bounds in a different frame from the rendered points — the same failure mode as finding 1, arrived at from the other direction. Both now share `isUsableModelMatrix()`, which additionally requires finite entries so a single NaN falls back to identity rather than propagating to all eight box corners and every point on the GPU. (No public API change — the helper is internal; api-surface snapshot unchanged.)

Both mutation-verified rather than assumed: reverting the bbox translation fails the new frame test; reverting the finite check fails `falls back to identity on a non-finite matrix, in BOTH consumers` (362 pass / 1 fail); both restore to full green.

Gates on `bb30b48a`: typecheck 33/33 · renderer 363/363 · pointcloud 120 pass / 2 skipped · test-wiring clean · API surface matches snapshot (3957 exports).

Rebase note: resolved against #1875 (point-cloud measure) in `point-cloud-node.ts` keeping both sides — `spatialIndex` and `model?` are both present and wired, as are `getRayQuerySources` and `setPointCloudTransform`.

PR #1879, opened 2026-07-25
feat(create,encoding): opt-in determinism hooks for reproducible IFC generation

louistrue:
## Problem

Byte-deterministic IFC generation through `IfcCreator` was impossible without monkey-patching globals: `packages/create/src/ifc-creator.ts` reads the wall clock (`Date.now()` for `IfcOwnerHistory`, `new Date().toISOString()` for the STEP `FILE_NAME` header and the work-schedule `CreationDate` default), and every GlobalId comes from `@ifc-lite/encoding`'s `generateUuid()`, which calls `crypto.randomUUID()` with no seed hook. The world-gym harness (moonshot branch, `tools/world-gym/lib/deterministic-runtime.mjs`) had to temporarily replace `globalThis.Date` and `crypto.randomUUID` to get reproducible output - that shim can be deleted once this lands (world-gym itself is not on main and is deliberately not touched here).

## Change (additive, defaults unchanged)

- `@ifc-lite/encoding`: `generateUuid(random?)` / `generateIfcGuid(random?)` accept an optional `RandomSource` (`() => number` in `[0, 1)`, exported type). With a source, UUID bytes are drawn from it (v4 version/variant bits still forced); without one, the platform CSPRNG path is byte-for-byte the old code.
- `@ifc-lite/create`: `ProjectParams.Timestamp` (`number | Date`) pins the creation instant used by the header, owner history and work-schedule defaults; `ProjectParams.GuidSource` (`() => string`) supplies GlobalIds. GuidSource output is validated (`isValidIfcGuid`) and the uniqueness loop is bounded, so a repeating source raises a clear error instead of spinning forever.
- Docs: `docs/api/typescript.md` `ProjectParams` listing updated plus a self-contained reproducible-output snippet (passes `docs:check-samples`).
- `scripts/api-surface.json`: +1 entry (`RandomSource: type` on `@ifc-lite/encoding`).

## Tests

- `packages/encoding/src/guid.test.ts`: same seed => same UUID/IFC GUID, different seed diverges, version/variant bits preserved, `() => 0` pins the exact `00000000-0000-4000-8000-000000000000` (proves the CSPRNG is bypassed), default path still random.
[...]

PR #1882, opened 2026-07-25
fix(create,encoding): reject out-of-range Timestamps and correct the RNG docs

louistrue:
Follow-up to #1879, which merged from a stale PR head — these three fixes were pushed to the branch but GitHub's PR object had not caught up, so the squash landed without them.

Found by a local CodeRabbit CLI pass on #1879 that the hosted review did not surface (the hosted review was clean).

## Fixes

- **`IfcCreator` accepted a finite `Timestamp` outside the Date range.** Epoch-ms past ±8.64e15 cleared `Number.isFinite` and then failed much later as a `RangeError` from `toISOString()` while writing the header. Now rejected in the constructor, where the message can still name the parameter.
- **`RandomSource` doc was wrong.** It promised "the default platform CSPRNG", but the unseeded path falls back to `Math.random` when Web Crypto is absent.
- **The default-path GUID test asserted luck, not contract.** It compared two draws for inequality; it now pins well-formedness (v4 version/variant, valid IFC GUID) and distinctness across 100 draws.

## Verification

- `@ifc-lite/encoding` 51 pass, `@ifc-lite/create` 103 pass.
- Adversarial probe against the seeded path (hostile `RandomSource` returning NaN/1/-1/1e9, a repeating `GuidSource`, Date-vs-epoch equivalence, v4 bit preservation, 2000-draw collision run): all green.
- Patch changeset included for both published packages.

### DIFF
--- a/.changeset/determinism-timestamp-range.md
+++ b/.changeset/determinism-timestamp-range.md
@@ -1,0 +1,6 @@
+---
+"@ifc-lite/create": patch
+"@ifc-lite/encoding": patch
+---
+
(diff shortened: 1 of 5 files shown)


louistrue:
@coderabbitai review

PR #1900, opened 2026-07-27
feat(provenance): B4.2 - spatially coupled merge semantics (bets AGAINST B2.1; B2.1 stays OPEN)

louistrue:
Bet B4.2 of the finishing plan, as it stands after the G4 review and re-attestation. **Stacked on #1897** - merge that first.

## What this closes

G2 red-team finding B2.1: the v0 op model applied ops purely per-node, so node-disjoint ops **always** commuted and the spatial half of the conflict predicate could never produce a true conflict - measured, 35 false and provably **zero** true.

## The conclusion, restated to what the artifact supports

**Replay-cost avoidance on 28 schedules at the price of 19 false refusals**, with the derived-cut sensitivity below. The spatial rule does **not** buy certificate soundness: `createCommutationCertificate` replays both orders regardless of the predicate, so **zero unsound certificates are emitted with or without it**, verified across all four semantics.

## The measured 2x2 (schedule-matched, so every cell attributes the same 9 events)

| cuts \\ containment | enforced | off |
|---|---|---|
| **lazy** (this bet's default) | **9** (3 apply-failed, 6 diverged) | **9** (0, 9) |
| **derived** (IFC, and this repo) | **3** (3, 0) | **0** |

Regenerating each stream under its own semantics: derived/enforced 3, lazy/off 13, derived/off 1.

- **9 -> 3 confirmed exactly**, and the 3 survivors are the same schedule indices the baseline recorded as apply-failures.
- **The G4 review's mechanism split was wrong**, and this is the substantive correction: containment-off *alone* still reads 9, because the three rejections simply become three divergences of the stale cut. **The stored lazy cut is the dominant mechanism**; containment is load-bearing only once cuts are derived.
- The residual case under regenerated derived/off is **not** a survivor of the 9 (schedule-matched that cell is 0) - it is a schedule the baseline never drew. It is worth naming because it is the only conflict in the grid with an **IFC basis**: an opening added into an element another client removes, i.e. `IfcRelVoidsElement` referential integrity, which only the spatial half catches.

## Ratios, with their denominators named

| | value | denominator |
|---|---|---|
| headline false-conflict rate | 8.78% (84/957) | ground-truth-commuting schedules - **this is the plan's <20% bar's denominator; it PASSES** |
| flagged precision-complement | 66.14% (84/127) | flagged schedules |
| spatial-fired precision-complement | **40.82%** (20/49), Wilson 95% [28.22%, 54.75%] | flagged schedules where the spatial rule fired |

40.82% was reported as a FAIL against the <20% bar. It stands on its own terms - the Wilson interval excludes 20% - but it is a **different ratio** from that bar, and its like-for-like comparator is 66.14%, against which the spatial rule over-approximates **less** than the predicate as a whole.

`10.63% -> 8.78%` is a **distribution change, not an improvement**: this bet added hosted openings to the base model, so the two figures are measured on different generative models.

## Withdrawn

The claim that lazy cuts are "exactly as in every BIM authoring kernel" is **dropped with repo evidence, not defended**: this codebase *derives* cuts (`rust/geometry/src/void_index.rs` builds host->openings from `IfcRelVoidsElement`; `router/voids/` subtracts at load time from the uncut Body), and its kernel handles overhanging cutters (`drop_faces_outside_host`), so IFC imposes no containment either. The couplings remain the default - a model where nothing couples is the v0 model G2 already refuted - but they are switchable and the sensitivity is published rather than argued.

## Freeze safety and one v1 question

`node-hash.ts` untouched, no kind's encoding moved, the 55 golden vectors unaffected, pinned by a test - **safe against the open freeze PR #1886**. Flagged, not acted on: `hostId` lives in `ModelState`, not `ElementPayload`, so two states differing only in hosting hash to the same root. Unreachable under this op model; a genuine v1 spec question.

## Verification

Provenance **176 tests** green. `node scripts/moonshot/g2-merge-soundness.mjs` exits 0 with the grid in the verdict block. Act-4 output unchanged, B3.5 golden byte-identical, no re-bless. Head `7ddb4de7`.

*Body rewritten 7ddb4de7 after the G4 re-attestation found it still carrying the pre-restatement story, including a 42-seed/36,000-schedule pooling row that exists in no committed artifact.*


louistrue:
## G4 adversarial review: two corrections to this PR's claims

The standing adversarial review (`docs/vision/reviews/g4-red-team-2026-07-29.md`) failed gate G4 and made two findings that land on this PR specifically. Both confirmed.

**1. The 42-seed pooling study does not exist in the repository.** This PR's body cites "pooled over 42 more seeds / 36,000 schedules ~31%, 399 more spatial-only true conflicts" as the strongest row in its evidence table. `grep` over `scripts/`, `packages/provenance/` and `docs/` finds no script, no scorecard, no artifact. It is an uncommitted measurement quoted in a PR body, and per the program's own standing-evidence instrument it does not count. **Treat the committed result as n=49 only** — where the FAIL survives on its own terms anyway: Wilson 95% CI on 20/49 is [28.2%, 54.7%], excluding the 20% bar.

**2. The 40.82% and the <20% bar are different ratios.** The headline rate is `falseConflicts / groundTruthConvergent` — a false-positive rate over commuting schedules. The restricted rate is `spatialFiredFalse / spatialFiredFlagged` — a precision-complement over flagged schedules. Under the headline's own convention the analogous unrestricted number is 84/127 = **66%**, not 8.78%. The comparison is not like-for-like in either direction. Reporting a failing number was the right instinct; the failure is against a bar the number does not address.

**Related, and not this PR's fault:** `10.63% -> 8.78%` is a **distribution change, not an improvement** — the base model now plants a hosted opening in every second grid element. Flagged conflicts went *down* while a coupling that creates conflicts was added, which is the tell.

**What the review credits, unprompted:** the golden re-bless is auditably scoped (48 lines, all under `acts/act4/data/battery`; acts 1/2/3/5 byte-identical including every hash), and declining to make `conflictPredicate` host-aware to fix the rate is called out as the behaviour the program says it wants.

**Required before this can merge** (review items 5 and 6): restate the two comparisons or say plainly they are different ratios; mark the 10.63%→8.78% arrow as a distribution change; and add the ablation — coupled semantics with the spatial rule **disabled**, counting unsound auto-merges. The soundness argument now *depends* on the spatial rule, so if the ablation is non-zero, KEEP is justified by soundness rather than by 9 conflicts out of 1,000. That is a categorically stronger result than the one this PR currently claims, and it is a one-line experiment.

louistrue:
## CodeRabbit CLI findings on this branch (hosted CodeRabbit never ran)

Surfacing this from #1933, which is stacked on this PR. Both PRs show:

```
CodeRabbit  pass  0  Review skipped: reviews are disabled for this base branch
```

That is a **false pass**, not a review — and `test.yml` and `moonshot.yml` don't run either, since both are scoped `pull_request: branches: [main]`. So this branch has had no real review and no real CI.

I ran the CodeRabbit **CLI** (0.7.1) over the combined stack (`--base origin/docs/moonshot-finishing-plan`, i.e. this PR + #1933). Findings in **this PR's** code, all MINOR, none addressed:

**1. `merge-battery.ts` ~1175-1184 — `wilsonInterval` doesn't validate `z`.** `successes`/`trials` are validated but `z` is not; a non-finite or negative `z` yields NaN or an inverted interval instead of a descriptive throw. This one has teeth given the interval is published evidence.

**2. `merge-battery.test.ts` ~342-351 — Wilson bracket assertions lack float tolerance.** "brackets the point estimate" compares `low`/`high` to `p` exactly; a boundary case can fail on representation rather than on maths.

**3. `merge-model.ts` ~448-492 — `assertOpeningsContained` is inconsistent with `assertInsideHost`.** The latter early-returns for geometry-less openings; the former does not, so a geometry-less opening takes a different path depending on which entry point runs.

**4. `g2-merge-soundness.mjs` ~360-371 — singular narrative under a plural count.** The IFC-basis explanation is phrased "one case" but is emitted for any non-zero `derivedNoContainment.unsoundAutoMerges`.

**5. `g2-merge-soundness.mjs` ~145-160 — equal rates fall through to a wrong conclusion.** When `flaggedRate` and the spatial-restricted rate are equal (both zero included), the output still reads "is NOT lower" / "over-approximates MORE".

I have **not** fixed any of these — they are this PR's code, not mine, and the ratio/narrative ones are claims-about-evidence where the author's intent matters. Flagging rather than patching.

Also worth knowing before this merges up: **E9 (`check-report-numerals.mjs`) has never run against this branch's prose.** It arrived in `9039d402`, which is not an ancestor of `feat/b42-spatial-merge` (branched at `5c1ed287`). The first time the numeral gate sees B4.2's write-up will be when this lands on `docs/moonshot-finishing-plan`.

louistrue:
## Round 2: I fixed the first 5, and the re-review found 6 more — including two MAJOR in the footprint contract

Louis asked me to clear this PR's findings so the stack could progress. I fixed the original five (patch below, uncommitted in a worktree — say the word and I'll push them). **All of B4.2's published numbers reproduce byte-identically after them**: 873 auto-merged, 0 unsound, 8.78%, spatial-ONLY true 9, ablated unsound 9, and 176/176 tests green. They are behaviour-preserving by construction.

But the re-review surfaced **six new findings**, and two are MAJOR in the write-set contract that B4.2's soundness argument rests on. I am not fixing these — they are semantic decisions about your claim, not tidy-ups.

### MAJOR 1 — `deriveAllCuts` performs an UNDECLARED write

Under `cuts: 'derived'`, `applyOp`'s `entity-add` calls `deriveAllCuts(next)`, which reaches `recutHost` and **mutates the host's mesh payload**. But `computeMergeOpFootprint`'s `entity-add` declares only the op's own fresh ids:

```ts
const writtenNodes = new Set<string>([op.entity.entityNodeId]);
for (const p of op.entity.psets) writtenNodes.add(p.psetNodeId);
for (const m of op.entity.meshes) writtenNodes.add(m.meshNodeId);
```

The host's mesh node is never named. So under derived semantics, adding an opening writes a node the footprint says it does not write — two ops can be judged structurally disjoint while both writing the same host mesh. The spatial rule likely catches most of these in practice (the regions overlap by construction), but "the spatial rule happens to cover it" is not the same as a sound write set, and the ablation explicitly measures what happens when the spatial rule is off.

Either add the affected host mesh nodes to `writtenNodes`, or redefine `writtenNodes` as *logical targets* in the `CutSemantics` docs and align every consumer. The second is defensible — but it has to be stated, because `footprint.ts` currently documents it as a write set.

### MAJOR 2 — host-region widening is asymmetric

`entity-add` widens its region through `withHostRegion`; `entity-remove` and `geometry-replace` do not. With `containment: 'off'` that asymmetry is load-bearing, and the `clipAabb` precondition comment claims containment is always enforced, which is not true once the knob exists.

### The other four (MINOR)

- `footprint.ts` ~214-230 — docs conflate replay convergence with predicate approval around `createCommutationCertificate`/`runSpatialAblation`.
- `merge-model.ts` ~280-289 — `baseVertexFloatCount` mishandles an empty `indices` array.
- `merge-battery.test.ts` ~67-71 — `spatialFiredFlagged` reconciliation assertion is weaker than it reads.
- `merge-model-hosting.test.ts` ~219-220 — disjointness loop passes vacuously if `fpA.writtenNodes` is empty; needs a non-empty assertion.

### Where that leaves the stack

I was asked to merge this if it was ready. **It is not, and I'm not merging it.** Two review passes have each found more, the second reaching the write-set contract itself. Hosted CodeRabbit has still never run on this branch, and E9 has never seen its prose.

#1933 (stacked on this) is independently green: 183 tests, CLI 0 findings, and the full moonshot lane dispatched and passing on real CI ([run 30689501751](https://github.com/LTplus-AG/ifc-lite/actions/runs/30689501751)). It is blocked only by this PR.

Issue #780, opened 2026-05-24
CSG boolean operations fail on bath example from buildingSMART IFC 4.3


kamyarba:
I tried to generate geometry for the standard buildingSMART CSG example (bath) from IFC 4.3 Annex E: [bath-csg-solid.html](https://standards.buildingsmart.org/IFC/RELEASE/IFC4_3/HTML/annex_e/advanced-geometric-shape/bath-csg-solid.html).

[csg-solid.ifc.txt](https://github.com/user-attachments/files/28190486/csg-solid.ifc.txt)

Expected:
- ifc-lite should generate the solid geometry as shown in the buildingSMART example when loading the associated IFC file.

<img width="500" height="375" alt="Image" src="https://github.com/user-attachments/assets/1e98b42a-578d-47e2-b470-edc30fbe38d6" />

Actual:
- ifc-lite does **not** generate the geometry for this example. The CSG/boolean operations appear unsupported or fail to produce the expected shape.

**Note:** CSG/boolean geometry support is patchy across the BIM ecosystem. 
However, supporting these IFC 4.3 reference examples could make IFC-Lite the open-source champion of geometry handling! 🏆

louistrue:
<img width="881" height="903" alt="Image" src="https://github.com/user-attachments/assets/da8e9529-e635-421f-a3c8-701f56cad02a" /> 🚀

kamyarba:
Fantastic! 🥇

PR #1916, opened 2026-07-30
fix(geometry): conform seams across plane buckets — fixes void-cut hairline cracks

louistrue:
## What

Void-cut geometry left T-junctions that render as hairline cracks. `consolidate_coplanar`
re-triangulates each coplanar plane bucket independently, and its collinear simplify could
drop a boundary vertex the abutting bucket keeps: the shared edge is then one long edge on
one side against two short ones on the other.

Minimal case `dental_clinic.ifc` #217 — a 12-triangle wall, watertight with no voids, comes
out of one opening subtract with **6 unmatched half-edges**. The end face splits at
mid-height while both corner verticals keep their full-height edge.

## The discrimination this rests on

A blanket "protect shared vertices" rule does not work: on the #098 reveal wall it drives
unpaired edges from 42 to **324**, because a faceted reveal shares almost every vertex and
blanket protection defeats the merge the simplify exists to perform. Narrowing to
non-parallel neighbours only reaches 298.

What actually separates the two cases is **the simplify's own judgment, read across
buckets**:

- a genuine seam vertex is a hard corner in the abutting bucket, so it survives *that*
  bucket's simplify and gets recorded;
- an `i_overlay` phantom is near-collinear in every bucket that touches it, is dropped
  everywhere, and therefore never becomes an insertion source.

No geometric proxy for "is this a real corner" is needed.

## How

Three phases over the existing bucket map:

- **A** unions / welds / simplifies as today, and records each surviving ring vertex at
  0.1 mm with the count of distinct buckets that kept it.
- **B** inserts, per bucket, the vertices some *other* bucket kept that lie on this
  bucket's plane and strictly interior to one of its ring edges. The plane test is the
  strong filter, so the scan stays cheap.
- **C** triangulates and emits in today's order.

Today's output is emitted **first**, and the conformed mesh is taken only when it is fully
watertight. So an already-watertight host pays nothing, and **no host can come out worse
than before** — the change is monotone by construction.

Two decisions that are measured, not cautious:

- **All-or-nothing accept.** Accepting partial improvements sends a half-conformed
  intermediate into the next sequential cut and compounds: it lowered the per-cut count on
  all seven of #098's cuts yet ended at 282 unpaired edges. At a zero bar the wall improves
  to 22.
- **0.1 mm decision grid**, not the 1 mm default. At 1 mm a chorded seam merges away and
  reads as closed — which is exactly what accepted the 282.

Also unifies the analytic prism cut's new crossing vertices at ulp scale: `prism_cut`
derives a crossing once per (host face, cutter plane) pair, so two faces sharing an edge
computed the same point one float step apart (`x = -20.289999008` vs `-20.290000916`,
exactly 2^-19).

## Measured, 116 fixtures / 1355 void-hosting elements

| metric | before | after |
|---|---|---|
| total unmatched boundary edges | 18252 | **17792** |
| elements with any tear | 238 | **199** |
| closed-solid body, not watertight | 80 | **41** |
| of which boolean-caused | 52 | **28** |
| of which arrived torn | 28 | **13** |
| triangulator-sensitive | 139 | **136** |
| snap-collapsed | 55 | **51** |
| #098 reveal wall unpaired edges | 42 | **22** |
| dental_clinic #217 | 6 | **0** (watertight) |

Every metric strictly better.

## New gate

`tests/triangulation_invariance.rs` (behind the test-only `triangulation-alt` feature,
which links a second independent ear-clipper as a differential oracle). It auto-discovers
every `.ifc` under `tests/models` up to 50 MB and asserts five pinned baselines, all
re-pinned to the improved numbers here.

It gates **total unmatched edges**, not just the count of torn elements, because element
counts are severity-blind: an earlier attempt took torn elements from 76 to 62 while
driving one wall from 42 unpaired edges to 324, and an element-count gate reads both as
"one torn element".

The harness also segments by `IfcShapeRepresentation`, since `Curve2D`/`Curve3D` axis
representations and `SurfaceModel` cannot be watertight, and excludes triangles collapsed
by the snap (georeferenced models below the RTC offset sit at UTM coordinates where the f32
step is ~3 cm).

## Module split

The new code would have blown three `module_size_ratchet` budgets. Rather than raise them,
the affected files were split so all three **shrink**:

- `consolidate.rs` 990 -> 825 (budget 956)
- `prism_cut.rs` 3251 -> 2958 (budget 3160)
- `triangulation.rs` 798 -> 763 (budget 782)

New files: `consolidate/{conform,ring_ops}.rs`,
`prism_cut/{closure_checks,vertex_dedup}.rs`, `triangulation/{alt_oracle,ring_geom}.rs` —
all under 400 lines, none needing an allowlist row.

## Verification

- 82 `ifc-lite-geometry` targets + 35 `ifc-lite-processing` targets green
- `clippy --all-targets` clean; `cargo check --workspace --all-targets` clean
- `mesh_determinism` manifest **unchanged** (native + the wasm pairing assertion)
- `module_size_ratchet` passes
- arm64 and wasm32 cross-compile clean
- corpus census re-run after the module split: byte-identical numbers
- one snapshot re-pinned, `annex_e__advanced__bath-csg-solid_780`: 128 -> 152 triangles,
  384 -> 216 vertices, with `total_surface_area`, `bbox_min`, `bbox_max` and
  `spike_triangles` **identical**

## Known trade-off

Conforming adds vertices, so hosts that get accepted grow: #217 goes 14 -> 26 triangles and
the #780 bath 128 -> 152. `bath_csg_solid_test`'s `tris < 180` ceiling still passes but its
headroom drops from 52 to 28.

Remaining: 28 boolean-caused closed-solid defects are untouched by construction, since the
all-or-nothing bar makes the pass a no-op on any host it cannot fully close.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved void-cut coplanar seam conformance to better preserve true seam vertices while avoiding near-collinear phantom inserts, reducing hairline cracks and boundary mismatches.
  * Enhanced analytic prism-cut seams by deduplicating cut vertices against the host mesh, improving numerical consistency, and tightening when conformed output is emitted (watertight completeness required).

* **Performance / Geometry Quality**
  * Improved constrained Delaunay triangulation determinism and refinement efficiency via bounded, more robust refinement and constraint handling.

* **Tests / CI**
  * Updated triangulation invariance coverage to a manifest-driven corpus and strengthened CI watertightness/invariance sweeps.

### DIFF
--- a/.changeset/geometry-conform-plane-bucket-seams.md
+++ b/.changeset/geometry-conform-plane-bucket-seams.md
@@ -1,0 +1,22 @@
+---
+'@ifc-lite/wasm': patch
+---
+
+Fix hairline cracks in void-cut geometry. `consolidate_coplanar` re-triangulates each
+coplanar plane bucket independently, and its collinear simplify could drop a boundary
+vertex that the abutting bucket keeps — leaving the shared edge spanned by one long
+edge on one side and two short ones on the other. That T-junction renders as a
+hairline crack under DoubleSide.
+
+The pass now conforms seams across buckets. What separates a genuine seam vertex from
+an `i_overlay` phantom is the simplify's own judgment read across buckets: a real seam
+vertex is a hard corner in the abutting bucket, so it survives that bucket's simplify;
+a phantom is near-collinear in every bucket that touches it and is dropped everywhere.
+Today's output is emitted first and the conformed mesh is taken only when it is fully
+watertight, so a host can never come out worse than before.
+
+Measured over 116 fixtures and 1355 void-hosting elements: total unmatched boundary
+edges 18252 to 17792, elements with any tear 238 to 199, and elements whose body is a
+closed solid yet not watertight 80 to 41. Also unifies the analytic prism cut's new
+crossing vertices at ulp scale, which fixes the case where two host faces sharing an
+edge computed the same point one float step apart.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,6 +83,9 @@ jobs:
               - 'apps/server/Cargo.toml'
               - 'Cargo.lock'
               - '.github/workflows/test.yml'
+              # The watertightness census sweeps the fixture set named here, so a
+              # corpus change moves its pinned counts even with no code change.
+              - 'tests/models/manifest.json'
             plato:
               - 'tools/plato/**'
               - 'rust/clash/src/generated/**'
@@ -460,10 +463,14 @@ jobs:
       - name: Rust tests
         run: cargo test --workspace --exclude ifc-lite-wasm
       # The geometry watertightness/invariance census is feature-gated because it
-      # links a second ear-clipper as a differential oracle and sweeps every fixture
-      # (~6 min), so `--workspace` above skips it. Without this step the whole suite
-      # early-returns and its pinned baselines gate nothing.
+      # links a second ear-clipper as a differential oracle and sweeps the whole
+      # fixture manifest, so `--workspace` above skips it. Without this step the
+      # suite early-returns and its pinned baselines gate nothing.
+      #
+      # Narrowed to `geometry` rather than `rust`: the sweep costs ~20 min, and only
+      # rust/geometry, rust/processing or the fixture manifest can move its numbers.
       - name: Geometry watertightness census (differential oracle)
+        if: needs.changes.outputs.geometry == 'true'
         run: cargo test -p ifc-lite-geometry --features triangulation-alt --test triangulation_invariance
 
   # Plato clash-math single-source freshness gate. The clash narrow-phase math
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -462,16 +462,48 @@ jobs:
         run: cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings
       - name: Rust tests
         run: cargo test --workspace --exclude ifc-lite-wasm
-      # The geometry watertightness/invariance census is feature-gated because it
-      # links a second ear-clipper as a differential oracle and sweeps the whole
-      # fixture manifest, so `--workspace` above skips it. Without this step the
-      # suite early-returns and its pinned baselines gate nothing.
-      #
-      # Narrowed to `geometry` rather than `rust`: the sweep costs ~20 min, and only
-      # rust/geometry, rust/processing or the fixture manifest can move its numbers.
-      - name: Geometry watertightness census (differential oracle)
-        if: needs.changes.outputs.geometry == 'true'
-        run: cargo test -p ifc-lite-geometry --features triangulation-alt --test triangulation_invariance
+
+  # Geometry watertightness / triangulation-invariance census.
+  #
+  # Its own job, not a step in `rust-tests`: the sweep runs the whole fixture
+  # manifest and takes ~20 min, which pushed that lane past its 28-minute timeout
+  # and cancelled it. Here it runs concurrently instead of serially, so neither
+  # lane's wall-clock depends on the other.
+  #
+  # Feature-gated because it links a second ear-clipper as a differential oracle,
+  # so `cargo test --workspace` does not run it — without this job the suite
+  # early-returns and its pinned baselines gate nothing.
+  #
+  # `geometry`, not `rust`: only rust/geometry, rust/processing or the fixture
+  # manifest can move its numbers.
+  geometry-census:
+    name: Geometry watertightness census
+    needs: changes
+    if: needs.changes.outputs.geometry == 'true'
+    runs-on: depot-ubuntu-24.04-4
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          lfs: false
+          persist-credentials: false
+      - name: Setup Rust (pinned by rust-toolchain.toml)
+        run: rustup show
+      - name: Cargo cache
+        uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+        with:
+          prefix-key: ci-rust-census
+      - name: Cache test fixtures
+        id: fixtures-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: tests/models
+          key: ci-fixtures-${{ runner.os }}-${{ hashFiles('tests/models/manifest.json') }}
+      - name: Fetch fixtures
+        if: steps.fixtures-cache.outputs.cache-hit != 'true'
+        run: node scripts/fixtures/fetch-fixtures.mjs
+      - name: Census
+        run: cargo test -p ifc-lite-geometry --features triangulation-alt --test triangulation_invariance -- --nocapture
 
   # Plato clash-math single-source freshness gate. The clash narrow-phase math
   # is written once in Plato and transpiled to Rust + TypeScript; the committed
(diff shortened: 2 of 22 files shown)


louistrue:
@coderabbitai full review

louistrue:
@greptileai review

louistrue:
## Perf verdict (AGENTS.md §Performance) — now **faster than main**

The first version of this PR cost **+56% geometry time** on CSG-heavy models. That is
fixed, and the branch is now materially faster than `main` *with* the conform running.

`scripts/perf/probe.sh <fixture> --iters 5 --json`, interleaved on an idle machine:

| fixture | main | this branch |
|---|---|---|
| `ISSUE_129_N1540_...` (11.5 MB, 1402 meshes) | **991 ms** `[1038,1026,1008,1032,1021]` | **651 ms** `[689,668,692,681,722]` |
| `AC20-FZK-Haus.ifc` | 5 ms | 5 ms |

**−34% geometry on the CSG-heavy fixture.** No monotonic climb in either array, so
neither run is thermally contaminated. Two independent measurement sessions agreed on
both baselines.

### Why the first attempt was slow, and what actually fixed it

Six optimisations of the conform pass all measured flat or worse (lazy seam map,
x-range pruning of the seam scan, caching the pass-1 CDT by clone and by move).
Instrumenting per phase explained why:

| phase | cpu-ms |
|---|---|
| `build_seam_map` | 5 |
| `conform_plans` | 15 |
| open-edge scan | 5 |
| **CDT** | **~1400** |

The conform was never the cost — it only *triggers* more CDT work, and the CDT was
carrying O(T)-per-item bookkeeping that the added second pass exposed. Every earlier
lever was aimed at the 20 ms rather than the 1400 ms. Removing those scans speeds up
the existing path too, which is why the branch beats `main` rather than merely
breaking even.

The lesson is recorded in the `scripts/perf/README.md` ledger.

### A lever that is NOT available

Gating the conform on a tear-size cap looked like an easy win (989 ms, flat) but made
`issue_098_reveal_wall` **fail** rather than revert to baseline: that host is cut by
seven sequential kernel operations, and a partially-conformed intermediate feeds the
next one and compounds. The all-or-nothing property has to hold across the whole
chain, so any lever that conforms some cuts and not others is unsafe. Noted here so it
is not re-attempted.

### Identity

Correctness is unchanged by the perf work — every census metric is identical:
torn void hosts 199, TOTAL unmatched edges 17792, closed-solid defects 41,
non-invariant 133, snap-collapsed 51. `mesh_determinism` manifest unchanged.
82 `ifc-lite-geometry` + 35 `ifc-lite-processing` targets green, clippy clean.

### Review fixes in this push

All nine findings from the previous round are addressed. Two worth calling out:

- **The census never ran in CI.** `cargo test --workspace` does not enable
  `triangulation-alt`, so the whole suite early-returned and its pinned baselines
  gated nothing. There is now an explicit step in the Rust lane.
- **The differential oracle short-circuited before `safe_earcut`'s sanitisation**, so
  the two triangulators saw different input and part of the invariance count was
  measuring duplicate-stripping rather than diagonal choice. Moved to `ear_clip`, the
  single seam both now go through — non-invariant dropped 136 → 133, which is exactly
  the artefact CodeRabbit predicted.

Module-size ratchet satisfied by splitting, not by raising budgets: `cdt.rs`
1866 → 1607, `conform.rs` 465 → 380, all new files under 400 lines.

louistrue:
@coderabbitai full review

louistrue:
@greptileai review

louistrue:
Both latest findings triaged.

**Greptile P1, "failed regions are silently dropped" — valid, fixed in `da262210`.**

Correct and reachable. `emit_plans` skipped a region whose CDT failed, and the accept
bar only tested watertightness — so a dropped region that happened to be its own
closed component leaves the remainder edge-balanced, and a candidate missing a whole
surface component would have been accepted. It is reachable specifically on the
conformed pass, because inserting seam vertices is what can make a previously-fine
ring fail the CDT.

`emit_plans` now returns whether every region landed, and completeness is part of the
accept bar rather than something to reason about case by case. Corpus is unchanged
(torn 199, unmatched edges 17792, defects 41, non-invariant 133), so nothing in the
fixture set trips it today — this is defence in depth, not a behaviour change.

**CodeRabbit, "finish splitting `consolidate.rs`" — acknowledged, not actioned here.**

`consolidate.rs` is an allowlisted file: the ratchet's rule for existing debt is that
it "can only stay flat or shrink", with a recorded budget of **956**. This PR takes it
**990 → 825**, well inside that, and `module_size_ratchet` passes. The ~400-line
guideline applies to *new* files, and every file this PR adds is under it:

| new file | lines |
|---|---|
| `consolidate/conform.rs` | 380 |
| `consolidate/conform/emit.rs` | 142 |
| `consolidate/ring_ops.rs` | 145 |
| `prism_cut/closure_checks.rs` | 218 |
| `prism_cut/vertex_dedup.rs` | 101 |
| `cdt/predicates.rs` | ~100 |
| `cdt/scratch.rs`, `triangulation/{alt_oracle,ring_geom}.rs` | 27–36 |

`cdt.rs` also went 1866 → 1607 and `conform.rs` 465 → 380, both by splitting rather
than raising a budget.

Taking `consolidate.rs` from 825 to under 400 means restructuring the plane-bucket
planner, which is pre-existing code this PR does not otherwise touch — worth doing,
but as its own change where it can be reviewed on its merits rather than buried in a
correctness-plus-perf PR. Happy to open a follow-up.

louistrue:
Round 4 triaged — all three valid, fixed in `bccd6bda`.

**"Do not discard the baseline completeness result"** — correct, and it was a real gap
I introduced. `emit_plans(.., false)` reports `complete`, and I was throwing it away
with `let (mut output, _) = ...`. A region that fails to triangulate is skipped, so an
incomplete baseline can be missing a whole closed component while still reading as
watertight — the same defect Greptile found on the conformed path, left live on the
baseline one. Consolidation's standing contract is "never worse than the raw kernel
output", so an incomplete baseline now returns the raw mesh rather than a known-lossy
consolidation.

**"Enforce `MAX_FIXTURE_BYTES` against the actual fixture"** — correct. The filter
trusted the manifest's recorded `size` and then only checked `is_file()`. Size is now
read from disk, so a stale manifest or a replaced fetch cannot silently change the
swept population.

**"Comment states the wrong threshold"** — correct, and I only half-fixed this last
round: the doc block was reworded but `// (pre-broken, csg-broke) with |coord| < 1e4`
at the bucket split was still stale. Gone.

Corpus unchanged — 111 models / 1165 void hosts, torn 199, unmatched edges 17792,
closed-solid defects 41, non-invariant 133. Nothing in the fixture set trips the new
baseline guard, so it is defence in depth rather than a behaviour change.
ISSUE_129 geometry 664 ms against main's 991.

---

For the record, the earlier threads that GitHub still shows open are addressed; their
anchors moved so they did not auto-resolve:

| finding | where |
|---|---|
| Codex: run the differential test in CI | `443d819b` — explicit step in the Rust lane, now narrowed to `geometry` in `c7d76acb` |
| Codex: record the perf verdict | posted above; 991 → 664 ms, ledger entry in `scripts/perf/README.md` |
| CodeRabbit: gate passes vacuously when fixtures are missing | `443d819b` floor, then `cd41accb` made the corpus manifest-driven so CI and local sweep the identical 111 models |
| Greptile: failed regions silently dropped | `da262210` — completeness is part of the accept bar |

louistrue:
Latest round: two were already fixed in `bccd6bda` (one CodeRabbit has self-marked
addressed). The third I am declining, with the repo's own rule as the reason.

**"Split the overlarge test module" (`triangulation_invariance.rs`, ~592 lines)** —
not actioned.

`module_size_ratchet` is the executable form of the ~400-line guideline, and it
**exempts test files explicitly**:

```rust
fn is_exempt(rel: &str) -> bool {
    let base = rel.rsplit('/').next().unwrap_or(rel);
    rel.contains("/generated/")
        || rel.contains("/tests/")      // <- this file
        || rel.contains("/examples/")
        ...
```

with the module doc stating "Generated code and test/example/bench/fuzz files are
exempt." `rust/geometry/tests/triangulation_invariance.rs` is under `/tests/`, and the
ratchet passes.

That exemption looks deliberate rather than incidental: an integration test that
sweeps a corpus is read as one unit, and splitting discovery, metric collection and
assertions across submodules would make the invariant it pins harder to follow, not
easier. Every *production* file this PR adds is under 400 lines, and the three
allowlisted files it touches all shrank (`cdt.rs` 1866 → 1607, `consolidate.rs`
990 → 825, `conform.rs` 465 → 380) rather than having budgets raised.

Happy to revisit if you would rather the exemption did not apply here, but that is a
change to the ratchet's policy rather than to this PR.

Issue #1904, opened 2026-07-28
Ctrl + click +drag rectangle-select returns no selection, `pickRect` returns empty on large models


ruornil:
### Version

ifc-lite v1.33.0

### Which surface

Viewer (browser)

### Reproduction

Bug found by me, details and a patch proposed by Claude

## Steps to reproduce 

1. Load a large model (large enough to trip the >500 pick-mesh threshold — see cause 2).
2. Hold Ctrl and drag a box over several elements.
3. Release.

**Expected:** every element under the box is selected.
**Actual:** nothing is selected (or exactly one element — whatever was under the cursor
at release).

`pickRect` returns empty on large models

`PickingManager.pick()` switches to CPU raycasting once a model needs more than
`MAX_PICK_MESH_CREATION = 500` pick meshes (`picking-manager.ts:154-165`), precisely
because "GPU picking requires individual mesh buffers; for 60K+ elements this is too
slow." As a result the per-element GPU pick meshes are **never hydrated**.

`pickRect` has no equivalent fallback. This is already documented at
`picking-manager.ts:236-240`:

> *Limitations: skips the CPU-raycast and dynamic-mesh-creation fallbacks that `pick()`
> uses for very large batched models, so rect pick may miss entities whose individual
> meshes haven't been hydrated. Acceptable for an MVP — rect select is a power-user tool
> and the user can fall back to single-click pick.*

In practice this isn't "may miss some entities" — on any model past the 500-mesh
threshold the GPU pick pass renders nothing at all and rect select misses **everything**.
That's also why Ctrl+click still works (CPU path) while Ctrl+drag doesn't (GPU path).

**Measured on the same drag:** `pickRect` → `count: 0`; a CPU AABB fallback → **108
entities**.

### Suggested fix

Give `pickRect` a CPU fallback mirroring the one `pick()` already has. What worked for me:
when the GPU pass returns empty, project each entity's cached world AABB through the
camera `viewProj` matrix and keep the entities whose projected screen rect intersects the
drag rect. `Scene.getAllMeshDataExpressIds()` + `Scene.getEntityBoundingBox()` are
sufficient — both work without GPU hydration and even after geometry release.

Sketch:

```js
const ids = scene.getAllMeshDataExpressIds();
const m = camera.getViewProjMatrix().m; // column-major: element(row,col) = m[col*4+row]

for (const id of ids) {
  if (hiddenIds?.has(id)) continue;
[...]

PR #1917, opened 2026-07-30
fix(renderer): stop the pick path aborting on a dead GPU device (#1901)

### LINKED ISSUE #1901: DOMException
AbortError: Failed to execute 'mapAsync' on 'GPUBuffer': A valid external Instance reference no longer exists.

[View in PostHog](https://eu.posthog.com/project/199147/error_tracking/019fa28d-35d9-7763-9f20-c0132fb5af3a)

louistrue:
Closes #1901.

## Root cause

`Renderer.pick()` / `pickRect()` were the one caller sitting outside the renderer's own device-liveness contract. `render()` and `getGPUDevice()` early-return once `deviceLost` is set or the device is uninitialised, but the pick path drove a full GPU round trip ending in `mapAsync` on a dead device. That readback rejects with `AbortError: … A valid external Instance reference no longer exists.`, and because the DOM listeners that reach it are `async` and nobody awaits them, it escaped as an unhandled rejection — once per click.

That matches the production stack exactly (canvas handler -> `selectionHandlers` -> `store.pick` -> `store.pick` -> `Picker.pick`, `mechanism.handled === false`).

## The fix, in two halves

**Entry guard** (3416b421) — `pick()` returns `null` and `pickRect()` returns an empty `Set` when the device is gone, using the same `!deviceLost && device.isInitialized()` predicate `render()` already uses. Also fixes a real leak found on the way: `Renderer.destroy()` nulled only `this.picker`, never the picker the `PickingManager` holds, so `PickingManager`'s own `if (!this.picker) return null` bail could never fire. `Picker` now also honours its own `destroyed` flag, which was previously write-only.

**In-flight guard** (9a1429af) — added during adversarial review, and this half is the one that matters most in production. An entry guard cannot pre-empt a pick that is *already parked* on `mapAsync` when `device.destroy()` or a driver reset lands. Worse: on an involuntary loss, `deviceLost` is only set when the `device.lost` promise resolves, so **every click between a GPU-process crash and that resolution sailed straight past the entry guard**.

## Adversarial verification

The verifier re-ran the fail-then-pass proof independently rather than trusting it. Reverting just the two source files and re-running `pnpm --filter @ifc-lite/renderer test` gives 5 failures with the verbatim production message; restoring gives 378/378.

It then attacked the entry-guard-only commit with a throwaway probe using a deferrable `mapAsync` so teardown could land mid-readback — **4 of 6 probes still reproduced the exact production string** (mid-flight destroy, mid-flight involuntary loss, `pickRect` mid-flight destroy, and both of two overlapping picks). That is what produced the second commit. Every dead-device assertion is on `stats.mapAsync === 0`, so a try/catch symptom patch could not have made these tests pass.

Also independently refuted: overlapping picks are *not* the cause on their own (`renderPickPass` -> `copyTextureToBuffer` -> `queue.submit` is synchronous, and both readback buffers are per-call locals), and the `picker.ts:359` comment named in the issue is a red herring — it is about coordinate clamping, and the clamp is present.

- **Behaviour change worth a look:** `PickingManager.pick()` has a CPU-raycast fallback for large batched models that needs no GPU. Because the guard sits at the `Renderer` level, a pick during device loss now returns `null` instead of taking that pure-CPU path. On a lost device the viewport is frozen anyway and this matches how `render()` degrades, so it was judged correct — but it is a real semantic change.
- **Not reproduced in a real browser.** Both halves are verified against a stub GPU that rejects with a real `DOMException(<verbatim Dawn message>, 'AbortError')`. Device loss cannot be forced locally.
- **This stops the exception storm, not the underlying UX failure.** Nothing in `apps/viewer/src` subscribes to `onDeviceLost` (grep: zero hits), so after a loss the canvas still freezes on its last frame with no user-facing signal. Deserves its own issue.
- **#1904 is NOT fixed by this and is a separate root cause** — confirmed independently by both the implementer and the verifier. `PickingManager.pickRect` reads only `Scene.getMeshes()` (hydrated meshes); batched geometry lives in `batchedMeshes` and never reaches the pick pass, and `pickRect` has neither the CPU-raycast fallback nor the hydration path `pick()` has. Decisive discriminator: `pickRect`'s only call site logs `'[useMouseControls] Rectangle selection failed:'` on a rejection, so an abort would have been *logged*, not silent. Repro #1904 with the console open — a warning means same cause, silence means the missing batched path.
- `packages/drawing-2d/src/gpu-section-cutter.ts:358,380` has the identical unguarded `mapAsync` shape. Left alone to avoid widening scope, but it will produce the same unhandled rejection if a section-cut readback races teardown.
- Separately confirmed: the `ifc-lite:cancelled` fingerprint on these events was **not** a deliberate classification. `analytics-scrub.ts:258-265` runs `classifyLoadError` over every `$exception` and `load-errors.ts:159` matches `/aborterror/i`, so any `AbortError` anywhere in the app is bucketed as a model-load cancellation. These events stop once this ships, but the mis-classification remains for future `AbortError`s.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `pick()` / `pickRect()` behavior when the graphics device is lost, uninitialized, or destroyed—picking now safely degrades (no result / empty set) instead of attempting GPU work.
  * Prevented mid-readback aborts from surfacing as unhandled failures; genuine picking errors still propagate.
  * Enhanced teardown and readback cleanup during destruction and interrupted readbacks to avoid stale picker usage and potential resource leaks.
* **Tests**
  * Added a regression test to ensure readback buffers are freed even when readback fails with a non-abort error.

### DIFF
--- a/.changeset/renderer-pick-guards-dead-device.md
+++ b/.changeset/renderer-pick-guards-dead-device.md
@@ -1,0 +1,27 @@
+---
+'@ifc-lite/renderer': patch
+---
+
+`Renderer.pick()` / `pickRect()` no longer drive a destroyed or lost GPU device.
+
+`render()` and `getGPUDevice()` both early-return once the device is gone; the
+pick path skipped that contract entirely. A pick is a full GPU round trip that
+ends in a `mapAsync` readback, so on a dead device the readback rejects with
+`AbortError: Failed to execute 'mapAsync' on 'GPUBuffer': A valid external
+Instance reference no longer exists.` Nothing on the pick path is in a position
+to handle it — the DOM click/contextmenu listeners that reach it are `async`
+functions whose promise nobody awaits — so it escaped as an unhandled
+rejection. And because the picker stayed dead, this was not a one-shot teardown
+race: it fired again on every subsequent click at a frozen viewport.
+
+`pick()` now resolves `null` and `pickRect()` an empty set once
+`isDeviceLost()` is true or the device is uninitialised, mirroring how
+`render()` degrades to skipping the frame. The GPU call is never issued, so no
+error is being hidden — consumers that want to react to the loss still
+subscribe to `onDeviceLost()`.
+
+Two supporting leaks are closed: `Renderer.destroy()` now also clears the
+picker reference held by the internal picking manager (it previously nulled
+only its own, leaving the manager pointing at a destroyed picker), and `Picker`
+— a public export usable standalone — now honours its own `destroyed` flag in
+`pick()` / `pickRect()` instead of only in `destroy()`.
--- a/.changeset/renderer-pick-guards-dead-device.md
+++ b/.changeset/renderer-pick-guards-dead-device.md
@@ -20,6 +20,14 @@ race: it fired again on every subsequent click at a frozen viewport.
 error is being hidden — consumers that want to react to the loss still
 subscribe to `onDeviceLost()`.
 
+The entry guard cannot close the window between `queue.submit()` and the map
+settling: a pick that was legal when it started is still aborted if the canvas
+unmounts, the model reloads (`Renderer.destroy()` ends in `device.destroy()`)
+or the driver resets while the readback is in flight — same `AbortError`, same
+escaping rejection. `Picker` now treats an `AbortError` from its own readback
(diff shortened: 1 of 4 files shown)


louistrue:
@coderabbitai review

PR #1918, opened 2026-07-30
fix(sandbox): own every QuickJS handle across throws so dispose cannot abort the module (#1905)

louistrue:
Refs #1905. **Deliberately does not close the issue** — see "Why this stays open" below.

## Root cause (one of two)

A handle created through a `QuickJSContext` is an **unmanaged** Lifetime: `QuickJSContext.dispose()` does not free it (verified — `ContextMemory.heapValueHandle` returns a bare `new Lifetime(...)`, unlike the scope-managed `staticHeapValueHandle`). Every container handle the bridge built was held in a bare local and freed on the success path only, so any throw between `newObject()`/`newArray()` and the matching `.dispose()` orphaned a JSObject. The orphan survives `JS_FreeContext`, and `JS_FreeRuntime` then trips `assert(list_empty(&rt->gc_obj_list))` — aborting the whole emscripten module, which is why the sandbox is unusable until a page reload.

The reachable trigger is a host-side exception raised while a `bim.*` result is being marshalled: a throwing own getter, a revoked `Proxy`, anything that makes `Object.entries` or the recursion fail partway through a structure.

## The fix

All in `packages/sandbox/src` — take ownership of every handle across throws:

- `bridge-schema.ts` `marshalValueWithGuard`: each container is disposed if the loop throws, each child in a `finally` so a throwing `setProp` cannot leak it either. Every level of the recursion is freed, not just the innermost.
- `bridge-schema.ts` `buildNamespace`: `nsHandle` and the in-flight `fn` move to `finally`.
- `bridge.ts` `buildBridge` / `buildConsole`: same for `bimHandle`, `consoleHandle`, each `fn`.
- `sandbox.ts` `eval`: the `executePendingJobs()` result was dropped entirely and its failure branch owns a live error handle — now disposed; the eval-result handle moves into a `finally`.
- `sandbox.ts` (review follow-up, 8b0cb59d): the `evalStartTime = 0` reset that stands the CPU interrupt down sat inside the `try`, so a throw from job draining left the interrupt armed against a stale start time and the next VM code in that realm — including teardown — would be cut short as a timeout. A copy now lives in the `finally`.

Regression test `dispose-leak.test.ts` drives the real `createSandbox` -> real bridge -> real `dispose()`. The shipped wasm's assertion is itself the "no live handles remain" oracle. On unfixed source 4 of its 8 cases fail with the exact production message; with the fix all 8 pass. Verified independently by the reviewer.

## Why this stays open

Adversarial review **demonstrated a second, independent trigger of the identical assertion that is still live with this fix applied**: heap exhaustion inside a queued QuickJS job.

Reproduced end-to-end through the real `createSandbox` at production defaults (64 MB / 30 s — exactly what `useSandbox.ts:167-170` configures), on this branch:

```js
async function run() { await 0; const a = []; for (;;) { a.push({ k: "v" }); } } run(); "started"
```

`eval()` resolves with `"started"`, then `sandbox.dispose()` throws the verbatim production message. It reaches OOM in ~1 s, well inside the timeout. This is squarely in plausible-production territory — `sandbox.ts`'s own comment notes that entry points are commonly wrapped as `async function run()`, whose post-await body runs as a job.

It is an **upstream defect, not a handle we own.** A minimal fresh-process repro with no ifc-lite code at all (only `getQuickJS` + `evalCode` + `executePendingJobs`, every handle disposed) aborts at every memory limit tested, 256 KB through 64 MB. Bisected: OOM inside a *drained job* aborts; OOM in the main eval body is clean; a job that merely throws is clean; the same script without draining is clean. It aborts whether the rejection is unhandled, `.catch()`ed, or `try`/`catch`ed inside the job.

No in-repo mitigation exists: quickjs-emscripten 0.32.0 exposes no GC trigger. The only remaining lever would be to deliberately skip `runtime.dispose()` on a heuristic OOM signal — trading a document-wide module death for an unbounded WASM runtime leak. That was **not** shipped; it is a maintainer call and wants an upstream issue against quickjs-emscripten first.

**Also honest about attribution:** the script text is not in telemetry, and both triggers fit the production event equally well (marshal-throw makes `eval()` reject then abort in the `finally`; OOM makes `eval()` resolve then abort in the `finally` — both surface as `mechanism.handled = false` from the same `dispose()` frame). So the claim that this commit fixes *the reported occurrence* is unproven either way. It removes a real root cause; it does not let us assert the issue is done.

## Other gaps

- `buildNamespace` / `buildBridge` / `buildConsole` exception-safety is correct by construction but has no test — neither agent could construct a realistic throw during registration through the public surface.
- `pnpm check:api-surface` was not run (needs a full workspace build). Verified by substitute: the diff contains no changed export lines and `disposeOrphan` is module-private, so both `scripts/api-surface.json` entries are unchanged. `patch` changeset is correct.
- The published export `marshalValue` changes behaviour on the throw path (no longer leaves an orphan). No callers outside `bridge-schema.ts` in-repo, but it is a >=1.0 surface.

### DIFF
--- a/.changeset/sandbox-dispose-handle-leak.md
+++ b/.changeset/sandbox-dispose-handle-leak.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Fix `sandbox.dispose()` aborting the whole QuickJS WASM module when a `bim.*` result could not be marshalled. Handles created through a QuickJS context are unmanaged lifetimes — `context.dispose()` does not free them — so a container handle orphaned by a mid-marshal exception (a throwing getter, a revoked `Proxy`, any host error raised while a result was being converted) kept a JSObject on the runtime's GC list and made `JS_FreeRuntime` assert `list_empty(&rt->gc_obj_list)`. Emscripten then `abort()`ed, leaving the sandbox unusable until a page reload. The bridge now owns every handle it creates across throws (`marshalValue`, namespace registration, the `bim` and `console` globals), and the disposable result of `executePendingJobs()` is freed instead of dropped. Script errors surface unchanged; only the abort is gone.
--- a/packages/sandbox/src/bridge-schema.ts
+++ b/packages/sandbox/src/bridge-schema.ts
@@ -218,32 +218,41 @@ function buildNamespace(
   schema: NamespaceSchema,
   context: BridgeCallContext,
 ): void {
+  // try/finally, not a bare sequence: a throw anywhere in the registration
+  // loop would otherwise orphan `nsHandle` (and the in-flight `fn`), and an
+  // orphaned handle makes the runtime's own dispose() abort the WASM module
+  // — see disposeOrphan() below for the full mechanism (#1905).
   const nsHandle = vm.newObject();
-
-  for (const method of schema.methods) {
-    const fn = vm.newFunction(method.name, (...handles: QuickJSHandle[]) => {
-      // Host-side errors (capability denials, SDK exceptions, type
-      // errors) MUST be re-thrown as a plain `Error` with a string
-      // message. Throwing a custom Error subclass — or any non-plain
-      // object — across the QuickJS native-callback boundary leaves
-      // the realm in a corrupt state: a subsequent handle access
-      // throws "Lifetime not alive". Normalising to a plain Error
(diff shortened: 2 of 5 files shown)


louistrue:
@coderabbitai review

PR #1919, opened 2026-07-30
fix(viewer): degrade the location minimap when WebGL is unavailable (#1914)

### LINKED ISSUE #1914: Error
{"requestedAttributes":{"antialias":false,"preserveDrawingBuffer":false,"powerPreference":"high-performance","failIfMajorPerformanceCaveat":false,"desynchronized":false,"alpha":true,"depth":true,"stencil":true,"premultipliedAlpha":true},"statusMessage":"C

[View in PostHog](https://eu.posthog.com/project/199147/error_tracking/019fae1d-b463-7721-ae17-7f810d5dd4e4)

louistrue:
Closes #1914.

## Root cause

`new maplibregl.Map(...)` was constructed inside an un-`catch`ed `.then()` callback in `LocationMap`. maplibre's constructor calls `_setupContainer()` then `_setupPainter()`, which throws **synchronously** when the browser refuses a WebGL context — so the throw became an unhandled rejection that PostHog autocaptured with `mechanism.handled === false`.

The two production samples had *different* status messages, which is why this is not one narrow GPU quirk:

- `OES_packed_depth_stencil support is required.`
- `Could not create a WebGL context, VENDOR = 0x1002 … ANGLE (AMD Radeon, D3D11) … ErrorMessage = BindToCurrentSequence failed:`

And it fired twice in one session because every collapse/expand of the Radix Collapsible around the Georeferencing panel remounted the component and re-threw.

## The fix

Removes the cause rather than only catching the throw:

- A cheap WebGL capability **probe** before construction, plus a **session latch** so an unsupported device does not re-attempt on every remount.
- Graceful fallback UI — the rest of the properties panel keeps working.
- The condition is reported as *handled* with a `reason` (`map_construction_failed` / `map_load_failed`) and a `webgl_status`, not as an unhandled error, and a one-slot report cap stops remount churn from spamming.
- A `.catch()` backstop around the async load path.

New `apps/viewer/src/lib/geo/map-webgl-support.ts` (+ 223 lines of unit tests) and an `analytics-scrub.ts` rule.

## Adversarial verification

The reviewer reproduced the fail-then-pass proof, then simulated the refusal by patching `HTMLCanvasElement.prototype.getContext` to dispatch a `webglcontextcreationerror` carrying `statusMessage` and return `null` — **the resulting Error payload matched the production one in shape exactly.** Verified live in Chrome. Found no regressions.

- **Dead export:** `isWebglContextCreationError` is exported with 6 unit tests but is never used in production code (knip would flag it; knip is in `package.json` but not wired into any workflow). Consequence: the `catch` around `new maplibregl.Map` does not discriminate, so *any* synchronous ctor throw latches a device-level "no WebGL" verdict for the session. Practical risk is low — the options are fixed and a refused context is the only realistic sync throw — and it stays distinguishable in PostHog via `reason: map_construction_failed` with no `webgl_status`. Left deliberately: every alternative traded label accuracy for remount churn.
- `takeMapWebglReportSlot()` is **one global slot shared across all reasons**, so a transient `map_load_failed` consumed first means a genuine device failure later in the same session is never reported.
- The `.catch()` backstop labels everything `map_load_failed`. An error thrown *after* successful construction (`addControl`, `new Marker`) would get the wrong copy and leave the map undisposed, because `mapRef.current` was never assigned. Narrow, and it converts a hypothetical app bug from uncaught to handled.
- The maplibre chunk (1,027 kB / 273 kB gzip) is still downloaded before the probe runs, so the **first** mount on an unsupported device pays the download; subsequent mounts short-circuit.
- The container purge in the catch does not undo two side effects maplibre performs before throwing (the `scroll` listener on our container, and the global `ImageRequest.addThrottleControl` handle). One dead Map object retained per session. Cosmetic.
- **Not verified on the reported hardware** (Chrome 150 / Windows 10 / AMD Radeon on ANGLE-D3D11).
- The repo has no jsdom/testing-library and zero `*.test.tsx`, so the component-level behaviour was verified live in Chrome but is **not** captured as an automated regression test.

## Merge-order note

This branch and `fix/posthog-1903-safari-load-failed` (#1903) both edit `apps/viewer/src/lib/analytics-scrub.ts`. Whichever merges second will need a rebase there.

### DIFF
--- a/apps/viewer/src/components/viewer/properties/LocationMap.tsx
+++ b/apps/viewer/src/components/viewer/properties/LocationMap.tsx
@@ -25,6 +25,11 @@ import type { CoordinateInfo, GeometryResult, MeshData } from '@ifc-lite/geometr
 import { downloadBlob } from '@/lib/export/download';
 import { reprojectToLatLon, reprojectFromLatLon, queryTerrainElevation, computeFootprintGeoJSON, type LatLon } from '@/lib/geo/reproject';
 import { buildKmz } from '@/lib/geo/kmz-exporter';
+import {
+  probeMapWebglSupport, markMapWebglUnsupported, takeMapWebglReportSlot,
+  getMapWebglVerdict, describeMapInitFailure, type MapWebglFailureReason,
+} from '@/lib/geo/map-webgl-support';
+import { posthog } from '@/lib/analytics';
 
 // Lazy-load maplibre-gl to avoid bloating the initial bundle
 let maplibrePromise: Promise<typeof import('maplibre-gl')> | null = null;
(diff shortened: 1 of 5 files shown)


louistrue:
@coderabbitai review

PR #1920, opened 2026-07-30
fix(geometry): a wasm trap must not brick every geometry consumer in the document (#1898)

### LINKED ISSUE #1898: Error
IFC-Lite WASM cannot recover from a fatal runtime error within the same document lifetime. Reload the page or recreate the worker process before calling init() again.

[View in PostHog](https://eu.posthog.com/project/199147/error_tracking/019fa277-447c-7943-ab17-2d06a55e6d58)

louistrue:
Closes #1898.

## Root cause

`IfcLiteBridge` stored one `Error` in a **module-scoped global** (`fatalWasmRuntimeError`) the first time *any* operation took a WebAssembly trap, then threw that same stale object from every subsequent `init()` for the life of the realm.

So an ordinary, per-operation memory-exhaustion abort during a GLB export **permanently disabled every other main-thread geometry consumer** — the next model load, the grid and drawing meshers, every other exporter. Mis-classifying a recoverable, per-operation failure as realm-fatal is the actual bug, and the guard text in the issue is just that mis-classification talking.

## The fix

- An **operation** trap now drops and frees only the `IfcAPI` handle that took it, and propagates unchanged. The next `init()` builds a fresh handle.
- Only an **init-time** trap is reported as unrecoverable — and advisory, never latched.
- New `packages/geometry/src/wasm-runtime-trap.ts` centralises trap detection/recording, with `context` / `error_kind` / `from_source_bytes` telemetry so the *next* occurrence tells us why the wasm trapped.
- `GLBExportDialog` gets a humanised message instead of raw engine text.

## Adversarial verification

The reviewer confirmed the root-cause claim against the code, re-ran the fail-then-pass proof, and drove the **real 4.1.4 wasm binary** in a Node harness. Two things it fixed itself (commit 158b4ce6):

- `GLBExportDialog` called `posthog.captureException` unconditionally in its catch, **including for the typed, expected `NO_RENDER_GEOMETRY` case** (user filtered every element out of the export). That classifies as `unknown`, so `stampFingerprint` leaves it on PostHog's default per-message grouping and every filter mistake mints its own PostHog issue — noise that would bury the wasm-trap signal this capture exists to surface. Now gated; the `export_failed` product event still counts it.
- The double-fault branch of `recordWasmRuntimeTrap` (the `try`/`catch` around `dispose()`, for when `free()` itself traps because the allocator is what aborted) had no test. Added one, verified red when the guard is removed.

It also probed the one new hazard it identified: switching the trap path from `reset()` to `dispose()` means we now `free()` a handle other code may still hold (`useIfcLoader.ts:1130` passes `geometryProcessor.getApi()` to the main-thread parser and holds it across awaits). Probed free-then-use against real wasm — wasm-bindgen guards it with a clean `Error: null pointer passed to rust` (**not** a `RuntimeError`, so no cascade), and a fresh bridge in the same realm still exports correctly afterwards. Degrades safely, but it is newer surface than main's leak-but-keep-alive behaviour.

- **Why the wasm trapped originally is still unknown.** Circumstantial evidence (the reporting user's 409/440 MB GLBs, the from-bytes path re-meshing into wasm linear memory) points at memory exhaustion, and the reviewer's own repro is that class of abort — but a Rust panic and an allocator abort are indistinguishable from JS under `panic='abort'` (`rust/Cargo.toml:39`). The new telemetry is what will answer it on the next occurrence. Post-fix this costs one export instead of the document either way.
- `rust/export/src/gltf.rs` still uses `assert!` for the glTF 4 GiB container limit, turning a user-actionable size limit into an `unreachable` trap. Worth a separate Rust issue converting it to a typed `ExportError::TooLarge`.
- **`WASM_RUNTIME_UNRECOVERABLE_EVENT` has zero subscribers anywhere in the repo** (contrast `WASM_ASSET_UNAVAILABLE_EVENT`, which `wasm-version-skew.ts:128` listens for). The "host offers a reload" promise is delivered only via `formatLoadError` text in the paths that call it. An init-time trap raised from `wasm-prewarm.ts`, `useGridLines3D`, `useSymbolicAnnotations` or `useDrawingGeneration` dispatches into the void.
- Only `GLBExportDialog` got the humanised message + telemetry. `HbjsonExportDialog.tsx:93`, `KmzExportDialog`, and the csv/obj paths still surface raw `err.message` and capture nothing.
- `wasm_runtime_crashed` is deliberately not in `RESOURCE_LIMIT_KINDS`, so a main-thread trap gets no automatic tessellation-tier downgrade even though `geometry_worker_crash` — the same failure class inside a worker — does. No change versus main, but an asymmetry worth a follow-up given the repro shows main-thread traps are memory-driven.
- Concurrent `init()` on one bridge still leaks an `IfcAPI` (`if (this.initialized) return` is not set until after `new IfcAPI()`). Pre-existing, but more reachable now that re-init after a trap is routine rather than refused.
- **No browser verification** on the reported Chrome 150 / Windows 10; verification is unit-level plus the Node + real-wasm harness.

## Merge-order note

This branch and `fix/posthog-1903-safari-load-failed` (#1903) both edit `packages/geometry/src/ifc-lite-bridge.ts` and `apps/viewer/src/lib/load-errors.ts`. Whichever merges second will need a rebase.

### DIFF
--- a/.changeset/wasm-runtime-trap-recovery.md
+++ b/.changeset/wasm-runtime-trap-recovery.md
@@ -1,0 +1,14 @@
+---
+"@ifc-lite/geometry": minor
+---
+
+A WebAssembly runtime trap no longer bricks every geometry consumer in the document.
+
+`IfcLiteBridge` used to store one `Error` in a module-level global the first time any operation trapped, and then throw that same object from every subsequent `init()` for the rest of the page's life. Because the global is per-realm, a trap in one consumer (a GLB export, say) disabled every other main-thread consumer — the next model load, the grid and drawing meshers, all the other exporters — even though the trap could only ever have damaged the engine handle that took it. The stored object also carried the stack of the call that first trapped, so the error reported later, from an unrelated call, was undiagnosable.
+
+Now:
+
+- A trap taken by an *operation* drops (and `free()`s — it used to leak) only the `IfcAPI` handle that took it, and propagates unchanged to the caller. Any later `init()`, on that bridge or another, builds a fresh handle and works.
+- A trap taken while *initializing* is the one unrecoverable case: this realm has no working engine, and neither half is retryable in practice (a trap in `new IfcAPI()` leaves the module singleton already built, a trap in instantiation is deterministic). It is reported as a freshly constructed error whose message carries the stable `WASM_RUNTIME_UNRECOVERABLE` marker and the underlying trap verbatim, with the trap as `cause`, and it dispatches `WASM_RUNTIME_UNRECOVERABLE_EVENT` on `globalThis` so the host can offer the user a reload. The library never reloads the page itself.
+
+New exports: `isWasmRuntimeTrap`, `isWasmRuntimeUnrecoverableError`, `WASM_RUNTIME_UNRECOVERABLE_CODE`, `WASM_RUNTIME_UNRECOVERABLE_EVENT`.
--- a/apps/viewer/src/components/viewer/GLBExportDialog.tsx
+++ b/apps/viewer/src/components/viewer/GLBExportDialog.tsx
@@ -44,6 +44,7 @@ import { buildHiddenIfcTypes } from '@/store/typeVisibilityFilter';
 import { posthog } from '@/lib/analytics';
 import { toast } from '@/components/ui/toast';
 import { GeometryProcessor, isNoRenderGeometryError, type MeshData } from '@ifc-lite/geometry';
+import { classifyLoadError, formatLoadError } from '@/lib/load-errors';
 import { exportGlbFromGeometry } from '@/lib/export/glb';
 import { downloadBlob, sanitizeFilename } from '@/lib/export/download';
 import { withInstancedMeshes } from '../../utils/instancedExport.js';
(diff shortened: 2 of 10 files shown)


louistrue:
@coderabbitai review

louistrue:
**CodeRabbit CLI round** (the hosted check was a false pass — rate-limited, and the `@coderabbitai review` re-trigger was a no-op because it treats the non-review as already done). 3 findings on this PR, none acted on. Reasoning, so you can overrule:

1. *"Scope this changeset to the changes in this layer"* — premised on this being one layer of a stack where `ifc-lite-bridge.ts` lands separately. It isn't; both the contracts and the bridge behaviour are in this PR, so the changeset describes what actually ships. **Dismissed as a wrong premise.**

2. *"Split GLBExportDialog.tsx, 517 lines > ~400 threshold"* (flagged major) — the house rule is real, but the file was **already 484 lines on `main`** before this PR; the fix adds 33. Splitting a 517-line dialog is a refactor well outside a telemetry-and-recovery fix, and the workflow brief was explicit about not widening scope. **Left for a follow-up** — worth doing, just not here.

3. *"Encode `from_source_bytes` as an enum — telemetry must contain only registry IDs, enums and numbers"* — CodeRabbit over-generalised the rule. AGENTS.md:92 scopes it to **tour** telemetry (`lib/tours/telemetry.ts`), not all capture sites. A boolean is filterable in PostHog and carries no file or model name, so it is already scrub-safe. **Dismissed as a mis-cited rule.**

The other four PRs did get code changes out of this round — including a genuinely loose drop-matcher on #1919 and two vacuous-test risks.

PR #1921, opened 2026-07-30
fix(viewer,geometry): retry + attribute the engine-binary fetch (#1903)

### LINKED ISSUE #1903: TypeError
Load failed

[View in PostHog](https://eu.posthog.com/project/199147/error_tracking/019fa699-9538-7062-85fc-9dba9f431b00)

louistrue:
Closes #1903.

## Root cause — two, both real, both fixed

**1. A missing retry on the one self-fetching wasm init.** `IfcLiteBridge.init()` — reached on the main thread via `GeometryProcessor.init()` in `useIfcLoader.loadFile` — was the **only** wasm-bindgen `init()` in the repo without `initWasmWithRetry`. One blip on the cold-cache ~1.3 MB `ifc-lite_bg.wasm` download killed the whole load.

**2. That is also why the PostHog event had no stack.** wasm-bindgen awaits its *own* `fetch()`, so WebKit's rejection propagated raw as `TypeError: Load failed` with an empty stack — the rejection originates inside `fetch`, not in any of our frames. That is exactly the "no `stacktrace` key" the production event showed, and it is the observability half of the issue: the capture sites forwarded the bare throwable with only `context: 'ifc_model_load'` — a 1400-line function — so nothing discriminated *which* fetch failed.

## The fix

- `initWasmWithRetry` on the geometry bridge init (and the parser's), so a single blip retries instead of failing the load.
- Attribution at the capture sites: which operation, which phase, retry count — spread **flat** onto the event (not nested under `additional_properties`, which is the mistake that made props unfilterable last time).
- New `network_unavailable` / `cancelled` classification so a genuine user-side offline blip or a user-aborted fetch is a warning with a stable fingerprint rather than an error, while real failures stay at error level.
- Scrubbing untouched on `$exception_list` (frame keys `resolved_name` / `mangled_name` look sensitive and would destroy stack traces).

## Adversarial verification

The reviewer found and **fixed a real regression this branch introduced** (commit 5769bdd4):

The new `network_unavailable` bucket matched `"failed to fetch"` as a **substring**, so Chromium's stale-deploy wording `Failed to fetch dynamically imported module: /assets/Foo-<hash>.js` landed in it. Measured with a probe: pre-branch `classify(...)` = `unknown` -> level `error`, default grouping; on the branch = `network_unavailable` -> level `warning`, fingerprint `ifc-lite:network_unavailable`.

A chunk 404 that survives `main.tsx`'s one-shot `vite:preloadError` reload budget is **our** breakage, and it was being downgraded to warning and merged in with genuine offline blips — precisely the "silence the category" outcome the issue forbids. Fixed by excluding module-import phrasings, mirroring the `MODULE_LOAD_FAILURE` guard `wasm-init-retry.ts` already applies for the same reason (`.wasm` module imports unaffected because `isWasmEngineLoadError` runs first). Pinned by a test verified to fail without the guard.

- **Behaviour change worth knowing:** a worker-side engine-load blip now classifies as `wasm_engine_load` instead of `geometry_worker_crash` (the worker's message is now attributed, and `isWasmEngineLoadError` outranks `isGeometryWorkerCrashError`). Strictly more specific, but it will re-bucket existing PostHog issues once.
- **No real-browser verification on Safari 26.4.** Neither agent reproduced a throttled/failing wasm fetch against a production build. The load-bearing inference — that a WebKit fetch rejection carries an empty stack — is corroborated by the production event itself having no `stacktrace` key, and by the repo's pre-existing assertion that `Load failed` is the Safari engine-load string.
- **Probe finding (pre-existing, widened):** calling `bridge.dispose()` while the retry is sleeping still lets `init()` resolve and set `initialized = true` on a disposed bridge, with `free()` never called. There is no in-flight/disposed guard; the retry widens the window from one fetch to fetch + 300 ms + fetch.
- **Probe finding (pre-existing, widened):** two concurrent `bridge.init()` calls have no in-flight promise guard and produced 3 wasm `init()` calls. Both resolved and `dispose()` freed exactly once (no double-free), so wasteful rather than broken. In practice `loadFile` awaits `init()` serially and the shared-module memo covers the worker pool.
- `isOfflineNetworkFailure` drops the event with no `$exception_level` guard, so an explicit `fatal` would also be dropped when `online === false`. Theoretical today — nothing in the codebase sets `fatal`.
- **Post-deploy action:** any PostHog alert or suppression filtering on `$exception_level = error` will stop surfacing `network_unavailable` / `cancelled`, and adding `error_kind` at the `useClash` / `useIDS` sites moves recognised families onto `ifc-lite:<kind>` fingerprints, so some existing issues will split or merge once.
- **Pre-existing, not introduced here, but it violates a hard project rule:** `apps/viewer/src/lib/analytics-scrub.ts` (~line 44, already on `main`) carries a real client model file name in a code comment as an example. Confirmed absent from this branch's added lines. Should be scrubbed in a separate change.

## Merge-order note

This branch conflicts with two others: `fix/posthog-1914-maplibre-webgl-context` (#1914) on `analytics-scrub.ts`, and `fix/posthog-1898-wasm-fatal-recovery` (#1898) on `ifc-lite-bridge.ts` and `load-errors.ts`. Whichever merges later needs a rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - WASM engine downloads now retry once after transient failures, improving model loading reliability.
  - Persistent engine-loading errors now identify the affected binary and loading context.
  - Added clearer guidance for connection-related load failures, including retry and network checks.
  - Improved error reporting for model loading, validation, clash detection, and GPU upload failures.
  - Benign offline and transient network failures are now handled with reduced alert severity or omitted from error tracking.


louistrue:
@coderabbitai review

louistrue:
**Rebased onto the post-merge `main`** (#1917/#1918/#1919/#1920 all landed first). Two things in this rebase were judgement calls, flagged for the record:

**1. The `ifc-lite-bridge.test.ts` conflict was semantic, not textual.** This branch carried a test `blocks in-process reinitialization after a fatal wasm runtime error`, which asserts the module-global latch that #1898 **deleted as the root bug**. Confirmed against `origin/main`: #1898 replaced it with `the bridge that trapped recovers on the next init()`. I dropped the obsolete test and kept #1898's; a naive resolution would have re-asserted the fixed behaviour.

**2. CodeRabbit CLI raised a `major` on the rebased state asking to reinstate that same latch** — "store the generated fatal error and rethrow it immediately at the start of subsequent `init()` calls; do not clear this marker from `reset()`", citing a coding guideline. The guideline is real and was stale: `packages/geometry/AGENTS.md:9` on `main` still said *"After a fatal WASM runtime error, `init()` re-throws forever ... not in-process."* #1920 changed the behaviour and never updated the doc. **Rejected the finding, fixed the doc instead** (e7593412) — following it would have re-introduced the exact defect #1898 fixed.

Local suites post-rebase: geometry 166/166, parser 317/317, viewer 2076 pass / 0 fail / 2 skipped (pre-existing).

PR #1925, opened 2026-07-31
chore(wasm): bump wasm-bindgen family to 0.2.126 — clears wide-arithmetic gate 1

louistrue:
## What

Bumps the pinned wasm-bindgen family so the **WASM wide-arithmetic tripwire's gate 1**
can pass. `wasm-bindgen 0.2.106`'s bundled `walrus` cannot parse a wide-arithmetic
code section, so `pkg-wide` failed to build at all:

```
error: failed getting Wasm module for .../ifc_lite_wasm.wasm
  1: failed to parse code section
  2: wide arithmetic support is not enabled (at offset 0x3a8e1)
Error: Running the wasm-bindgen CLI
```

Upstream fix: `walrus` 0.26.0 (wasm-bindgen/walrus#306, 2026-03-25); wasm-bindgen
>= 0.2.115 is the floor that depends on it, and 0.2.126 is current.

| pin | from | to |
|---|---|---|
| `wasm-bindgen` | =0.2.106 | =0.2.126 |
| `js-sys` | =0.3.83 | =0.3.103 |
| `web-sys` | =0.3.83 | =0.3.103 |
| `wasm-bindgen-futures` | =0.4.56 | =0.4.76 |
| `wasm-bindgen-test` | =0.3.56 | =0.3.76 |

They move together because js-sys/web-sys pin wasm-bindgen exactly — bumping only
wasm-bindgen does not resolve. `serde-wasm-bindgen 0.6.5` and `wasm-bindgen-rayon
1.3.0` are unchanged and still build.

## This clears gate 1 ONLY

`BUILD_WIDE=1 bash scripts/build-wasm.sh` now succeeds and the script's own
`wasm-tools` litmus prints `wide-arithmetic bundle emits wide ops`.

**Gate 2 is untouched.** V8 still has wide-arithmetic behind the default-off
`--experimental-wasm-wide-arithmetic`, so the determinism step can still go red on
the runner's Node — by design. Nothing here makes wide-arithmetic shippable. What it
buys is a tripwire that isolates one variable: gate 1 can no longer fail, so a future
red run means the V8 side moved, which is the signal the workflow exists to give.

## Verification

| check | result |
|---|---|
| `BUILD_WIDE=1` build | ok — litmus passes, **gate 1 clears** |
| default bundle | ok, 3.8M |
| `BUILD_THREADED=1` | ok — litmus `threaded bundle imports shared memory` |
| wasm32 mesh-determinism | pass — pinned `mesh_determinism.wasm32.json` still matches, **emitted mesh bytes unchanged** |
| `cargo test --workspace` | 133 targets, exit 0 |
| `rust/csg-thread-bench` (own `[workspace]`) | builds for wasm32 |
| `pnpm typecheck` | 33/33 |
| exported `.d.ts` API | **557 normalised signature units, identical** |

The `ifc-lite.d.ts` diff is large (~3.2k lines) but carries **no API change**: it is
indentation (2→4 spaces), member reordering, and newly emitted Rust doc comments. It
is committed because `test.yml` gates it; `pkg/README.md` and `pkg/package.json` churn
on every `wasm-pack` run and that gate excludes them by design, so they are reverted
(see `5254b458` — they slipped into the first commit when a later verification build
re-dirtied them). They are stale on `main` independently of this change, verified by
rebuilding `pkg` on a clean `main`: those two move there too, `ifc-lite.d.ts` does not.

## Perf verdict

Perf-sensitive path (`rust/wasm-bindings`, `scripts/build-wasm.sh`), so per AGENTS.md:
no runtime regression is expected or observed — this changes the JS glue and codegen,
not the Rust hot path, and mesh output is byte-identical.

Bundle size: published `@ifc-lite/wasm` 4.1.4 is 3,937,900 B; this build is
4,004,113 B (+66 KB / +1.7%). **That delta is not attributable to the bump alone** —
it spans ~30 files of Rust changes since the 4.1.4 tag, and js-sys 0.3.103 newly links
`futures-util`. Flagging it rather than claiming it is free. The script's "1100 KB
budget" warning has been dead since well before this PR (4.1.4 already exceeded it).

## Two things reviewers should know

**Cost window.** `scripts/ci-wasm-prebuilt-eligible.sh` and `scripts/vercel-install.sh`
diff `rust/ Cargo.lock …` against the `@ifc-lite/wasm@4.1.4` tag. Merging this means
every subsequent PR compiles wasm32 from source on paid Depot runners, and Vercel
deploys compile Rust, until the next `@ifc-lite/wasm` release. Three pending changesets
already touch wasm, so sequencing a release soon closes the window.

**Next nightly bump will need a flag.** Since wasm-bindgen 0.2.122, threaded builds on
nightlies dated 2026-05-06+ also require `-C link-arg=--export=__heap_base`. The pinned
`nightly-2025-11-15` predates that, so it is deliberately absent and the threaded build
works today. A breadcrumb is now in `scripts/build-wasm.sh` so the next toolchain bump
does not hit an unexplained link failure.

## Docs

`docs/architecture/wasm-wide-arithmetic.md` carried a standing verdict — *"do NOT pay a
wasm-bindgen major-pin bump for a bundle no browser can run"* — and stated the pin as
`=0.2.106` in three places. That verdict assumed the bump had a real cost; measured, it
does not (see the table above). The doc now records the bump, why the verdict changed,
and that blocker 2 still gates shipping.

The workflow also hard-coded "still pins 0.2.106" in its header and in the gate-1
failure diagnosis. Left alone, a future gate-1 failure would print a known-blocked
explanation for what would now be a genuine regression. Both are corrected.

## Summary by CodeRabbit

* **Improvements**
  * Updated WebAssembly tooling for improved build and parsing compatibility.
  * Preserved existing APIs, TypeScript signatures, mesh data, and runtime behavior.

* **Documentation**
  * Clarified wide-arithmetic support status, validation results, and shipping requirements.
  * Documented threaded WebAssembly build requirements for newer toolchain versions.
  * Noted current browser support limitations for wide-arithmetic operations.

* **Tests**
  * Enhanced compatibility checks and diagnostics across supported and unsupported WebAssembly engines.
  * Added bundle validation and deterministic-output checks where runtime support is available.

---

## Update: the bump exposed a broken gate 2, fixed here too

Dispatching the tripwire against this branch confirmed gate 1 clears — `pkg-wide` builds on 0.2.126. That let the lane reach **gate 2 for the first time**, which failed with:

```
node: bad option: --experimental-wasm-wide-arithmetic
```

That flag has never existed in any Node. Node exited during argv parsing, before loading any wasm, so **gate 2 had never measured engine support at all**. The signal was inverted too: a shipped V8 feature drops its experimental flag, so on the day wide-arithmetic lands the hard-coded flag would still be `bad option` and this lane would stay red exactly when it should turn actionable.

**Blocker 2 measured properly** (47-byte module whose only body is `i64.add128`): `WebAssembly.validate` → `false`, compile throws `invalid numeric opcode: 0xfc13`, and `node --v8-options` lists no wide-arithmetic flag under any name — reproduced on Node 22, 24 and 26, and independently on the CI runner. So blocker 2 is real, just not for the reason the doc recorded. Corrected.

Changes: gate 2 now **probes** the engine instead of assuming a flag; the determinism check runs only when the engine can execute the module and is skipped rather than failed otherwise; blocked-on-engine is **green** with a step-summary note (a permanently red lane is one nobody reads — which is how the bogus flag survived); and the probe fails **closed**, so a failed probe or missing node errors out instead of quietly reporting "unsupported".

Verified green end to end on runs 30618694830 and 30619120410.

**Scope note:** this is more than a dependency bump. I judged it in scope because the PR's stated benefit is "the tripwire now isolates one variable", which was false while the gate could not run, and the bug is only observable because of this bump. Can be split out if preferred.

### DIFF
--- a/.changeset/wasm-bindgen-0-2-126.md
+++ b/.changeset/wasm-bindgen-0-2-126.md
@@ -1,0 +1,17 @@
+---
+'@ifc-lite/wasm': patch
+---
+
+Bump the pinned wasm-bindgen family to 0.2.126 (js-sys/web-sys 0.3.103,
+wasm-bindgen-futures 0.4.76, wasm-bindgen-test 0.3.76). They move together because
+js-sys and web-sys pin wasm-bindgen exactly.
+
+This unblocks gate 1 of the WASM wide-arithmetic tripwire: wasm-bindgen 0.2.106's
+bundled `walrus` parser could not read a wide-arithmetic code section, so the
+`pkg-wide` bundle failed to build at all. It builds now.
+
+No API or behaviour change for consumers. The generated TypeScript surface is
+unchanged — 557 normalised signature units, identical before and after; the large
+`ifc-lite.d.ts` diff is indentation, member ordering and newly emitted doc comments.
+The pinned wasm32 mesh-determinism manifest still matches, so emitted mesh bytes are
+unchanged. Default, threaded and wide bundles all build and pass their litmus checks.
--- a/.github/workflows/wide-arithmetic.yml
+++ b/.github/workflows/wide-arithmetic.yml
@@ -24,22 +24,26 @@
 #      2026-03-25) fixed it; wasm-bindgen >=0.2.115 is the floor that depends on
 #      it. The bump was taken in its own PR, as this workflow always intended —
 #      a green run here is proof the pin moved, never a workaround for it.
-#   2. Engine-support gate: even once (1) clears, no stable browser runs the
-#      module — V8 has it prototyped behind the (default-off) flag
-#      `--experimental-wasm-wide-arithmetic`. Node ships the same V8, so we
-#      pass that flag to the Node runner `wasm-pack test --node` uses via
-#      `NODE_ARGS` (a direct node argv, since V8 flags are
-#      rejected inside NODE_OPTIONS), the earliest in-repo signal that V8
-#      accepts the module — well before it is on-by-default in any browser.
+#   2. Engine-support gate: STILL BLOCKING. No stable engine runs the module.
+#      Measured 2026-07-31 on Node 22 (same V8 a browser ships): a module whose
+#      only body is `i64.add128` is rejected outright —
+#      "invalid numeric opcode: 0xfc13" — and `--v8-options` lists no
+#      wide-arithmetic flag whatsoever. An earlier revision of this file assumed
+#      a `--experimental-wasm-wide-arithmetic` flag and passed it unconditionally
+#      via NODE_ARGS; that flag has never existed, so node exited with
+#      "bad option" before loading any wasm and this gate measured nothing. It is
+#      now probed at runtime (see the probe step) instead of assumed.
 #
-# Until BOTH clear, the "Build pkg-wide" and/or "wasm32 mesh-determinism"
-# steps below are EXPECTED to fail — that is the tripwire working, not a bug
-# in this workflow. What must never happen quietly is a determinism
-# divergence: if both gates clear (build succeeds, Node/V8 runs the module)
-# but the pinned wasm32 manifest no longer matches, wide-arithmetic changed
-# the kernel's exact-predicate output and it must not ship. The three step
-# names below are deliberately distinct so a failure notification alone says
-# which of the three happened without opening the log.
+# Gate 2 being blocked is the KNOWN state, so it is green — a permanently red
+# lane is one nobody reads, which is exactly how the bogus flag survived. Red
+# means the state moved and someone should look:
+#
+#   * build step red      -> toolchain regression (gate 1 came back)
+#   * tripwire step red   -> the engine now ACCEPTS wide-arithmetic. Good news,
+#                            and the signal this workflow exists to deliver.
+#   * determinism red     -> wide-arithmetic changed the kernel's exact-predicate
+#                            output. Must not ship. Only reachable once the
+#                            engine runs the module at all.
 #
 # Cost: weekly + on-demand only, same profile as determinism.yml, and
 # deliberately NOT a `pull_request` trigger — this must never become a
@@ -131,30 +135,67 @@ jobs:
           echo "If the error instead mentions a missing wasm-tools/wasm-pack binary, that is an"
           echo "environment problem with this job, not a toolchain-support gap."
 
-      # --- Gate 2 + the real check: engine support + byte-determinism ------
+      # --- Gate 2: does this engine actually accept wide-arithmetic? -------
+      # Probe FIRST, then decide how to run the determinism test.
+      #
+      # The previous shape hard-coded `NODE_ARGS: --experimental-wasm-wide-arithmetic`
+      # and could never work: no Node exposes that flag (V8 has no wide-arithmetic
+      # entry in `--v8-options` at all), so node exited with "bad option" during
+      # argv parsing, before loading any wasm. Gate 2 was never measured. The
+      # signal was also inverted: a shipped V8 feature DROPS its experimental
+      # flag, so on the day wide-arithmetic lands, a hard-coded flag would still
+      # be "bad option" and this lane would stay red exactly when it should turn
+      # actionable.
+      #
+      # So ask the engine directly, with a 47-byte module whose only body is
+      # `i64.add128` (opcode 0xfc13). Order matters: try no flags first (the
+      # shipped case), then only flags this V8 actually lists.
+      - name: Probe engine support for wide-arithmetic (gate 2)
+        id: engine_probe
+        if: always() && !cancelled()
+        run: |
+          # (module (func (export "f") (param i64 i64 i64 i64) (result i64 i64)
+          #   local.get 0 local.get 1 local.get 2 local.get 3 i64.add128))
+          mod='AGFzbQEAAAABCgFgBH5+fn4Cfn4DAgEABwUBAWYAAAoOAQwAIAAgASACIAP8Ews='
+          probe='process.exit(WebAssembly.validate(Buffer.from(process.argv[1],"base64")) ? 0 : 1)'
+
+          if node -e "$probe" "$mod" 2>/dev/null; then
+            echo "support=default" >> "$GITHUB_OUTPUT"
+            echo "node_args=" >> "$GITHUB_OUTPUT"
+            echo "engine accepts i64.add128 with no flags -> wide-arithmetic is shipped here"
+            exit 0
+          fi
+
+          for f in $(node --v8-options 2>/dev/null \
+                     | grep -oE -- '--experimental-wasm-[a-z0-9-]+' | sort -u); do
+            case "$f" in *wide*|*arith*|*128*) ;; *) continue ;; esac
+            if node "$f" -e "$probe" "$mod" 2>/dev/null; then
+              echo "support=flag" >> "$GITHUB_OUTPUT"
+              echo "node_args=$f" >> "$GITHUB_OUTPUT"
+              echo "engine accepts i64.add128 under $f"
+              exit 0
+            fi
+          done
+
+          echo "support=unsupported" >> "$GITHUB_OUTPUT"
+          echo "node_args=" >> "$GITHUB_OUTPUT"
+          echo "engine rejects i64.add128 and exposes no flag that enables it"
+
       # Reuses the SAME wasm32 determinism manifest check determinism.yml runs
       # (rust/wasm-bindings/tests/mesh_determinism.rs against the pinned
       # rust/processing/tests/manifests/mesh_determinism.wasm32.json), just
-      # compiled with +wide-arithmetic and run under a Node started with V8's
-      # experimental flag for it. `if: always()` so this gate's own signal
-      # never gets swallowed by gate 1 failing above — a future world where
-      # wasm-bindgen is bumped but Node still rejects the module (or vice
-      # versa) should still show up as exactly one red step, not zero.
+      # compiled with +wide-arithmetic. Only runs once the engine can actually
+      # execute the module — otherwise it measures nothing about determinism.
       - name: wasm32 mesh-determinism vs pinned manifest (wide-arithmetic bundle)
         id: determinism_check
-        if: always() && !cancelled()
+        if: always() && !cancelled() && steps.engine_probe.outputs.support != 'unsupported'
         env:
           RUSTFLAGS: ${{ env.WIDE_ARITH_RUSTFLAGS }}
-          # V8's implementation is staged behind this flag (default off) — see
-          # docs/architecture/wasm-wide-arithmetic.md "Delivery status". Node
-          # ships the same V8 build a browser would, so passing it here is the
-          # earliest in-repo signal for when the engine actually accepts the
-          # module, independent of when any browser turns the flag on by
-          # default. NODE_ARGS (honored by wasm-bindgen-test-runner) passes the
-          # flag as a direct node argv: V8 flags are NOT allowed inside
-          # NODE_OPTIONS (Node exits with "not allowed in NODE_OPTIONS"), so
-          # routing it through NODE_OPTIONS would keep this lane red forever.
-          NODE_ARGS: '--experimental-wasm-wide-arithmetic'
+          # NODE_ARGS (honored by wasm-bindgen-test-runner) passes V8 flags as
+          # direct node argv: they are NOT allowed inside NODE_OPTIONS (node
+          # exits with "not allowed in NODE_OPTIONS"). Empty when the feature
+          # is on by default.
+          NODE_ARGS: ${{ steps.engine_probe.outputs.node_args }}
         run: wasm-pack test --node rust/wasm-bindings --test mesh_determinism
 
       - name: Diagnose determinism-check failure
@@ -166,16 +207,45 @@ jobs:
           echo "enabled\" -> gate 1 (build-toolchain): same cause as the pkg-wide build step above,"
           echo "this compile of the TEST crate hit the same wasm-bindgen/walrus limitation."
           echo ""
-          echo "Failure mentions WebAssembly.validate / instantiate / \"invalid opcode\" / RuntimeError"
-          echo "from Node -> gate 2 (engine support): this runner's Node/V8 build still rejects"
-          echo "wide-arithmetic even with --experimental-wasm-wide-arithmetic set. Expected until V8"
-          echo "stages the feature on by default (docs/architecture/wasm-wide-arithmetic.md)."
-          echo ""
           echo "Failure is an assertion/panic naming mesh_determinism.wasm32.json (a manifest diff"
           echo "report, e.g. positions_hash/indices_origin_hash/normals_hash mismatch) -> REAL"
-          echo "regression: wide-arithmetic changed the kernel's emitted bytes. Both gates above must"
-          echo "have cleared for this to even be reachable, so treat it as high-priority regardless of"
-          echo "the weekly/non-blocking schedule -> do NOT flip BUILD_WIDE on by default until fixed."
+          echo "regression: wide-arithmetic changed the kernel's emitted bytes. The engine probe and"
+          echo "the build must both have cleared for this to even be reachable, so treat it as"
+          echo "high-priority regardless of the weekly/non-blocking schedule -> do NOT flip"
+          echo "BUILD_WIDE on by default until fixed."
+          echo ""
+          echo "A WebAssembly.validate / instantiate / \"invalid opcode\" error should NOT appear here:"
+          echo "the probe step gates this on the engine accepting i64.add128. If one does, the probe"
+          echo "and the real module disagree -> suspect the probe, not the kernel."
+
+      # The actual tripwire. Blocked-on-V8 is the known state, so it is GREEN:
+      # a lane that is permanently red teaches everyone to ignore it, which is
+      # how the bogus-flag bug above survived. Red here means the state MOVED
+      # and someone should look.
+      - name: Tripwire — engine support changed
+        if: always() && !cancelled() && steps.engine_probe.outputs.support != 'unsupported'
+        run: |
+          echo "::error::wide-arithmetic is now ENGINE-SUPPORTED on this runner (${{ steps.engine_probe.outputs.support }}) — this lane is red on purpose."
+          echo ""
+          echo "This is the signal the tripwire exists to give, not a regression."
+          echo "Next steps: confirm the determinism step above passed, then re-evaluate"
+          echo "flipping BUILD_WIDE and update docs/architecture/wasm-wide-arithmetic.md"
+          echo "(blocker 2) plus the expected state encoded in this workflow."
+          exit 1
+
+      - name: Record blocked-on-engine state (expected)
+        if: always() && !cancelled() && steps.engine_probe.outputs.support == 'unsupported'
+        run: |
+          {
+            echo "### Gate 2: engine support — still blocked (expected)"
+            echo ""
+            echo "This runner's V8 rejects \`i64.add128\` (opcode \`0xfc13\`) and exposes no flag"
+            echo "that enables it, so the wasm32 determinism check was skipped: it would have"
+            echo "measured engine support, not determinism."
+            echo ""
+            echo "This is the known state (blocker 2 in docs/architecture/wasm-wide-arithmetic.md)"
+            echo "and is intentionally **not** a failure. The lane turns red when it changes."
+          } >> "$GITHUB_STEP_SUMMARY"
 
       # --- Informational only: NOT a timing benchmark ----------------------
       # There is no in-repo wasm timing harness for this — the predicate and
--- a/.github/workflows/wide-arithmetic.yml
+++ b/.github/workflows/wide-arithmetic.yml
@@ -265,9 +265,15 @@ jobs:
           # historically exposed a proposal through it before giving it a
           # dedicated flag. Without it the early-warning lane would report
           # "unsupported" through exactly the rollout phase it exists to catch.
-          candidates=$(printf '%s\n' "$v8opts" \
-                       | grep -oE -- '--experimental-wasm-[a-z0-9-]+' | sort -u || true)
-          if printf '%s\n' "$v8opts" | grep -qE -- '--wasm-staging(\s|$|\))'; then
+          # Here-strings, not pipes. `--v8-options` is ~108 KB, and `grep -q`
+          # exits on first match, so a pipe leaves the producer writing into a
+          # closed pipe: harmless under plain `bash -e` (the pipeline takes
+          # grep's status) but SIGPIPE=141 under `-o pipefail`, which is what
+          # GitHub's `shell: bash` means. That would silently stop probing
+          # staging — blind in exactly the rollout this lane watches for — from
+          # nothing more than someone adding a `defaults: run: shell: bash`.
+          candidates=$(grep -oE -- '--experimental-wasm-[a-z0-9-]+' <<<"$v8opts" | sort -u || true)
+          if grep -qE -- '--wasm-staging(\s|$|\))' <<<"$v8opts"; then
             candidates=$(printf '%s\n%s' '--wasm-staging' "$candidates")
           fi
 
--- a/.github/workflows/wide-arithmetic.yml
+++ b/.github/workflows/wide-arithmetic.yml
@@ -219,8 +219,16 @@ jobs:
           # mangled-constant case specifically: Buffer.from tolerates bad base64
           # silently, and a truncated module fails validation for the wrong
           # reason.
+          # Digest, not length: substituting a base64 character (rather than
+          # dropping one) still decodes to 102 bytes, so a length check passes
+          # while the module is malformed — validate would say no, the lane would
+          # record "unsupported", and it would stay green forever even after the
+          # engine shipped support. The digest pins the exact bytes.
+          #   printf %s "$mod" | base64 -d | sha256sum
           probe='const b=Buffer.from(process.argv[1],"base64");
-                 if(b.length!==102){console.log("PROBE_ERR:module is "+b.length+" bytes, expected 102");process.exit(9)}
+                 const d=require("crypto").createHash("sha256").update(b).digest("hex");
+                 if(d!=="d83c3e467c96f8a8bc62e05f63efa8184f52bed0029011180705cad77c083503"){
+                   console.log("PROBE_ERR:module digest "+d+" ("+b.length+" bytes) is not the pinned probe module");process.exit(9)}
                  console.log(WebAssembly.validate(b)?"PROBE_YES":"PROBE_NO")'
 
           classify() {  # $@ = node argv prefix; echoes yes|no, or returns 1 on a broken probe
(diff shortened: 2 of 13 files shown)


louistrue:
@coderabbitai full review

louistrue:
@greptileai review

louistrue:
All three findings triaged; both distinct ones are fixed.

**"Revert generated WASM package metadata and README churn"** (CodeRabbit + Codex P1)
— correct, fixed in `5254b458`.

I had reverted those two before committing, then a later `BUILD_THREADED=1`
verification run re-dirtied them and `git add -A` swept them back in. The PR body
claimed they were left alone, which was also wrong; corrected. Only
`pkg/ifc-lite.d.ts` remains, which `test.yml`'s type-sync gate requires.

For the record on why they don't belong here: they are stale on `main` independently
of this change. Rebuilding `pkg` on a clean `main` moves `README.md` and
`package.json` but leaves `ifc-lite.d.ts` untouched — so the `.d.ts` diff is genuinely
caused by 0.2.126's codegen, and the other two are pre-existing drift for whatever
regenerates them next.

**"Mark the cleared build gate consistently"** (Codex P2) — correct, fixed in
`e31052aa`.

I appended the bump note without updating the status section above it, so the document
simultaneously said `pkg-wide` "fails in the bindgen step" and that all three bundles
build, and left blocker 1 unticked in the shipping checklist. The heading, blocker 1
and the checklist item now say cleared, with the original failure mode kept in past
tense so the history stays readable. Blocker 2 is untouched and still gates shipping —
that distinction is the one thing in this document worth getting right.

louistrue:
## The bump surfaced a broken gate — scope grew, deliberately

Dispatching this workflow against the branch (the only way to exercise it; it is `schedule` + `workflow_dispatch`, never `pull_request`) showed gate 1 clearing exactly as claimed — `pkg-wide` builds on 0.2.126. That let the lane reach **gate 2 for the first time**, which is where it got interesting.

Gate 2 failed with:

```
node: bad option: --experimental-wasm-wide-arithmetic
Error: Node failed with exit_code 9
```

That is none of the three failure modes this workflow's own diagnostics enumerate. **That flag has never existed in any Node.** Node exited during argv parsing, before loading any wasm — so gate 2 had never measured engine support at all, it just looked blocked for an unrelated reason.

The signal was also inverted. A shipped V8 feature *drops* its experimental flag, so on the day wide-arithmetic actually lands, the hard-coded flag would still be `bad option` and this lane would stay red **exactly when it should turn actionable**.

### What is actually true about blocker 2

Measured rather than assumed, with a 47-byte module whose only body is `i64.add128`:

| check | result |
|---|---|
| `WebAssembly.validate` (Node 22, no flags) | `false` |
| compile | throws `invalid numeric opcode: 0xfc13` |
| `node --v8-options`, any wide/arith flag | none, at any name |
| CI runner's own V8 (this workflow, run 30618694830) | `engine rejects i64.add128 and exposes no flag that enables it` |

Blocker 2 is real and still blocks. Just not for the reason the doc recorded — V8 does not implement the opcodes here, staged or otherwise. The old claim that it sat behind a default-off flag was wrong, and I have corrected it rather than leaving it to mislead the next reader.

Verified independently that the module isolates the right variable: `wasm-tools validate` fails with **only** "wide arithmetic support is not enabled", and passes under `--features wide-arithmetic`. So `validate === false` is about wide-arithmetic support, not a malformed module.

### Changes

- gate 2 now **probes** the engine (no flags first — the shipped case — then only flags this V8 actually lists) instead of assuming a flag name
- the determinism check runs when the engine can execute the module, and is **skipped, not failed**, when it cannot
- blocked-on-engine is now **green** with a step-summary note. A permanently red lane is one nobody reads, which is how the bogus flag survived. Red now means the state moved
- the probe fails **closed**: a failed probe or a missing node errors out rather than quietly reporting "unsupported" and going green

The lane is now green end to end for the first time, with every step meaning what its name says.

### Caveat on scope

This is more than a dependency bump. I judged it in scope because the PR's stated benefit is "the tripwire now isolates one variable" — which was false while the gate could not run, and the bug is only observable *because* of this bump. Happy to split the workflow fix into its own PR if you would rather keep the bump minimal.

louistrue:
@coderabbitai review

The workflow rewrite (87919302, 214ad873) landed after your last pass, so it has not been reviewed. Please look specifically at `.github/workflows/wide-arithmetic.yml` — the probe step's shell logic and the two `if:` expressions gating on `steps.engine_probe.outputs.support`.

louistrue:
@greptileai review

Your last reviewed commit is `2a516e28`, which is 4 commits behind — it predates both the generated-artifact revert (`5254b458`) and the whole `.github/workflows/wide-arithmetic.yml` rewrite (`87919302`, `214ad873`). Please re-review the branch head.

louistrue:
## Local CodeRabbit CLI review of the branch head — 4 findings, 3 taken

The hosted bots were both behind: Greptile's last reviewed commit was `2a516e28` (4 behind, predating the whole workflow rewrite), and CodeRabbit's last pass was 08:28Z against commits that landed before it. Ran the CLI locally against `main` so the workflow rewrite actually got read. Greptile has since re-reviewed the head and raised nothing.

**Taken:**

**1. (major) The failure-reading guide still called a walrus/parse error "status quo, not a regression."** Correct at the `=0.2.106` pin, wrong the moment 0.2.126 landed — and it contradicted this same document's "build red = a TOOLCHAIN regression" two sections away. Now classified as a build regression that keeps `BUILD_WIDE` blocked and needs fixing rather than waiting out.

This is the third time in this PR I've left a status line behind a status change. The pattern is mine, not the reviewer's: I keep updating the section I'm editing and missing the one that restates it elsewhere.

**2. (minor) Blocker 2's scope was overstated.** I measured Node 22, which is V8 — that covers Chrome/Edge and says nothing about SpiderMonkey or JavaScriptCore. Both the doc and the workflow header generalised it to "no stable engine runs the module." Now scoped to V8, with Firefox/Safari explicitly unverified and left to the per-engine checklist.

**3. (minor) Tripwire step interpolated `steps.engine_probe.outputs.support` directly into the script body.** No untrusted input reaches it (the probe writes only `default`/`flag`/`unsupported`), but it now goes through `env:` like `NODE_ARGS` does — consistent and lint-clean.

**Skipped:**

**4. (minor) `holesOffsets` range documentation in `packages/wasm/pkg/ifc-lite.d.ts`.** Pre-existing content — this PR's only change on that line is indentation, confirmed against the diff. The real fix is the Rust doc comment plus a regeneration, unrelated to a wasm-bindgen bump. Better as its own change than smuggled into this one.

louistrue:
Both taken — thanks, the first one is a genuine hole.

**1. (major) `node --v8-options` failure swallowed by `2>/dev/null`.** Correct. I'd added a `node --version` preflight so a *missing* node fails closed, but not for enumeration breaking on a node that otherwise works. In that case an empty flag list is indistinguishable from "no such flag", so the probe would report `unsupported` on incomplete information and the job would go green — the same silent-green class as the bug this whole change exists to remove. It now errors rather than guessing.

**2. (minor) Line 79 still said "no stable browser runs the module."** Correct, and slightly embarrassing: I scoped exactly this claim 120 lines further down in the previous commit and missed the earlier statement of it. Now reads "V8 rejects the module (Firefox and Safari unverified)".

That is the fourth time in this PR I've updated one statement of a fact and left another copy stale. Worth naming as a pattern rather than four separate slips.

All five probe paths re-verified against stubs after the change — engine ships the feature, engine has it behind a flag, engine rejects it, node missing, `--v8-options` broken — plus a live dispatch of the workflow.

louistrue:
Taken — and worth flagging that this one arrived as an **"outside diff range"** comment, so the review rendered as a clean pass with zero inline comments. Anyone reading the check status rather than the review body would have missed it.

**`determinism_check` could run after `build_wide` failed.** Correct. It compiles the test crate with the same toolchain, so in that state it can only fail for the same reason — two red steps for one cause, against this file's own claim of one red per cause. The `always()` there was protecting gate 2's independence, and the probe step now provides that without depending on the build at all, so the gating costs nothing.

Also realigned the diagnostic underneath it, which still led with the walrus/parse case that this gating makes unreachable. It now leads with the manifest diff (the only genuinely reachable cause) and keeps parse and invalid-opcode as "this should not appear, suspect the harness" notes.

---

@codex review

Your last pass was against `2a516e28`, now 6 commits behind. Everything since — the generated-artifact revert and the entire `.github/workflows/wide-arithmetic.yml` rewrite — is unreviewed by you. The workflow's probe step and its `if:` conditions are the part worth your attention.

louistrue:
Both taken, and the ledger one is the most valuable finding in this PR.

**`scripts/perf/README.md` — corrected.** You're right that this is the file that matters: it's the canonical lever ledger per AGENTS.md, so it's what a future optimization pass consults. It still said V8 has the feature behind `--experimental-wasm-wide-arithmetic`, that `WebAssembly.validate` fails on every shipping engine, and that maintainers should wait for the flag to be staged. All three are wrong — the flag never existed, only V8 was measured, and there is nothing to wait for. Leaving it would have sent the next investigation down exactly the dead path this PR just walked. Now carries the measured verdict with Firefox/Safari explicitly unverified.

**Architecture doc CI-lane section — corrected.** It still described determinism as gated only by the probe with `always()` keeping it independent of step 1, which is the opposite of the build-success gate. Now states that determinism depends on both gates while gate 2 stays independently legible through the probe.

**One more, found by sweeping.** Your second finding was the sixth time in this PR I corrected one statement of a fact and left another copy stale, so rather than wait for a seventh I grepped the tracked tree for every wide-arithmetic claim. It found one: this workflow's own opening line still called the blocker "a V8 flag no stable browser ships yet". Fixed. Every surviving mention now states explicitly that the flag never existed.

louistrue:
Final round: the last two commits (`c094e730` perf-ledger correction, `a9baab11` repo-wide sweep) landed after every bot's last pass, so none of you have seen them.

- CodeRabbit last reviewed `70c9761`
- Codex last reviewed `76946b1`
- Greptile last reviewed `214ad873`

@coderabbitai review
@codex review
@greptileai review

louistrue:
Two taken, one rejected.

**Taken — roadmap docs still carried the false claim.** You're right, and my earlier "repo-wide sweep" was not one: I grepped for the exact phrases I had already written rather than for the claim, so every restatement in different wording survived it.

- `docs/vision/moonshots-execution-plan.md:65` — M6a row said "blocked on V8 flag" → now "blocked on V8 opcode support"
- `docs/vision/moonshots-tech.md` — "blocked only on browser flags", and "ships the moment browser engines flip the flag (**V8 has it behind one today**)". That last one is the most explicit statement of the false claim anywhere in the repo, and it's the doc a future planning pass reads.

**Taken — cross-engine claims in the architecture doc.** Lines 71-73 asserted "no shipping browser engine supports it yet" and inferred Safari; line ~150 said the probe returns false "on every engine today". Both contradicted the V8-only measurement stated 40 lines away. Now scoped to V8, with Firefox/Safari explicitly unverified and left to the per-engine checklist.

**Rejected — `rust/csg-thread-bench/Cargo.lock`.** The premise doesn't hold: that file is **gitignored** (`.gitignore:75`) and has never been tracked.

```
$ git ls-files rust/csg-thread-bench/
rust/csg-thread-bench/Cargo.toml
rust/csg-thread-bench/build.sh
rust/csg-thread-bench/src/lib.rs
rust/csg-thread-bench/web/index.html
rust/csg-thread-bench/web/serve.mjs

$ git check-ignore -v rust/csg-thread-bench/Cargo.lock
.gitignore:75:rust/csg-thread-bench/Cargo.lock  rust/csg-thread-bench/Cargo.lock
```

So there is no tracked lockfile for this PR to refresh, and none is "still resolving 0.2.106" in the repo. `cargo check --locked --offline --manifest-path rust/csg-thread-bench/Cargo.toml` also does not fail here — it resolves and compiles against the local (ignored) lockfile, which already carries 0.2.126. On a clean checkout `--locked` would fail for the unrelated reason that no lockfile exists at all, which is pre-existing and deliberate for this detached bench.

Happy to be shown wrong if you're reading a tracked copy I'm not seeing.

louistrue:
`95163f69` corrects the roadmap docs (`docs/vision/moonshots-tech.md`, `moonshots-execution-plan.md`) and the two remaining cross-engine claims in the architecture doc. Doc-only, no code or workflow changes since your last pass.

@codex review
@greptileai review

louistrue:
`7833fad6` addresses all four findings from your `95163f69` pass:

- **Probe engine now explicit and recorded.** Implemented differently from the suggestion, deliberately: pinning to the repo's Node 22 would make the lane permanently blind, since that V8 line is frozen and will never gain these opcodes. It installs the newest Node instead and records node/V8 in the output, summary and tripwire message, so a state change is attributable to V8 rather than to a runner image bump. Live run: `probing node v26.5.1 / V8 14.6.202.34-node.24` — still unsupported, which is a materially stronger result than the Node 22 measurement.
- **Roadmap**: `browser flags` and `V8 flag timing ... shipping is a flag flip` corrected.
- **Timeline**: no longer calls V8's state "flag/experimental at best".
- **Red-tripwire alignment**: the CI section said blockers flip "red to green" and the checklist was introduced with "once this workflow is green end to end". Both inverted by this PR. Now state that green is the blocked state and red is the trigger.

@coderabbitai review
@codex review
@greptileai review

CodeRabbit — your last pass was `70c97613`; you skipped `a9baab11` and `95163f69`.

louistrue:
Taken — and it's a hole in the attribution I added in that same commit, so thanks for catching it immediately.

`Install newest Node` had no `if:`, so it inherited the implicit `success()` and was skipped whenever the pkg-wide build failed. The probe below it has `always()` and would have run anyway — on the image's preinstalled Node, while recording it as "newest". Gate 1 being red is exactly the moment gate 2's verdict needs to stay trustworthy, so the setup step now carries the same `always() && !cancelled()`.

I audited every step's condition in the file while there; the remaining implicit `success()` steps are setup and build steps that should skip on failure.

louistrue:
`47d727ae` takes the three findings from a local CodeRabbit CLI run:

- **The probe now exercises every wide op the bundle emits** — `i64.add128`, `i64.sub128`, `i64.mul_wide_s`, `i64.mul_wide_u` — instead of `add128` alone. This was the substantive one: an engine with a partial implementation would have read as "supported" and been handed a determinism run it could not execute. The 102-byte module is still rejected by `wasm-tools validate` without `--features wide-arithmetic` and accepted with it, so a false verdict still isolates to feature support rather than to a malformed module. Same bar added to the blocker-2 checklist.
- Checklist now states explicitly that the shipping release must accept the probe with **no experimental flags enabled** — implied before, worth saying outright given this PR began with a flag that never existed.
- Timeline heading said "researched 2026-06" above a 2026-07-31 measurement.

Live run on the new module: `probing node v26.5.1 / V8 14.6.202.34-node.24` → `engine rejects the wide-arithmetic opcodes and exposes no flag that enables them`. Green.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Two taken as written, one corrected in the opposite direction.

**Taken (major) — the delivery-plan `PROBE` snippet validated `i64.add128` alone** while the checklist two sections down now requires all four opcodes. You're right that this matters more than it looks: a future implementation would have copied that snippet and shipped exactly the partial-probe bug the CLI caught in the CI probe an hour ago. It now carries the same 102-byte module, and I verified the TS byte array decodes byte-identical to the one `wide-arithmetic.yml` probes with, so CI and runtime cannot drift apart.

**Taken (minor) — bundle gating described in the present tense** for behaviour that is not wired. Reworded, with an explicit "none of this is wired yet".

**Corrected — `wasm-features.ts` does not exist.** You read `scripts/build-wasm.sh` as evidence that it already uses `WebAssembly.validate()` to select the bundle, and asked me to stop describing it as "not yet created" in the perf ledger. But:

```
$ ls packages/geometry/src/wasm-features.ts
ls: no such file
$ git ls-files packages/geometry/src/wasm-features.ts
(nothing)
```

So the perf ledger is right and `scripts/build-wasm.sh:190` is the file implying otherwise — it names the path without saying it is still planned. I fixed the build script rather than the ledger. The inconsistency you spotted is real; the direction was inverted.

louistrue:
@codex — you and CodeRabbit independently landed on the same finding (the delivery-plan `PROBE` snippet embedding a single-opcode module), which is a good sign it mattered. It's fixed in `55d872ca`, pushed just after your snapshot of `47d727ae`: the snippet now carries the same 102-byte all-opcode module, and I verified the TS byte array decodes byte-identical to the one `wide-arithmetic.yml` probes with, so the shipping path and CI cannot diverge.

`55d872ca` is doc-only plus a one-line comment in `scripts/build-wasm.sh`. No workflow logic changed since `47d727ae`.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Local CLI run against `47d727ae` — all three taken, and the sweep they prompted found three more of the same kind.

**(major) The measurement paragraph inferred Chrome/Edge from Node 22.** Fair hit. "Node 22 — i.e. V8, so this speaks for Chrome/Edge" is an inference, not a measurement; Chrome ships its own V8 build. It now states both versions actually tested — Node 22 / V8 12.4 locally and Node 26.5.1 / V8 14.6 in CI — and says outright that Chrome and Edge were not themselves measured.

**(major) Runtime probe flow still referenced the single-opcode probe.** Same class as the `PROBE` snippet fix; the CI-lane section still said the determinism step runs when the engine "can execute `i64.add128`".

**(minor) Ambiguous gate description.** "It is additionally gated on the pkg-wide build succeeding" read as if "it" were the checklist. Now names the determinism step and says it is skipped, not failed, when gate 1 is red.

**Three more from sweeping afterwards** — the probe was still described as 47-byte/40-byte single-opcode in `docs/architecture/wasm-wide-arithmetic.md:252`, `.github/workflows/wide-arithmetic.yml:32` and `scripts/perf/README.md:231`. Same doc-lag pattern as the rest of this PR, caught by grep this time rather than by a reviewer.

`7217bfed`. Doc and comment changes only; no workflow logic touched since `47d727ae`.

louistrue:
Head is now `7217bfed`. Since your `47d727ae` pass there are two doc-only commits — `55d872ca` (all-opcode `PROBE` snippet, which you and CodeRabbit both flagged) and `7217bfed` (measurement scoped to the V8 versions actually tested, plus three stale probe descriptions). No workflow logic has changed since `47d727ae`.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Taken, and it is a contradiction I introduced in `7217bfed` itself. Narrowing the measurement paragraph left the summary above it still concluding "so Chrome and Edge cannot run it" — exactly the inference I had disclaimed 45 lines below.

Now reads: the V8 builds we tested (12.4 and 14.6) reject the module; no browser release was measured directly, because Chrome and Edge ship their own V8 builds and Firefox/Safari are different engines entirely. Every browser is unverified, deferred to the per-engine checklist.

That is the ninth time in this PR I have corrected one statement of a fact and left another copy of it stale. I swept for browser conclusions drawn from the Node probes afterwards and none remain, but the pattern is worth naming plainly rather than treating each instance as a one-off.

`5030cea5` — doc-only.

louistrue:
Head is `5030cea5` and CI has settled fully green — 20 pass, 5 skip, 0 pending, 0 failures. Greptile has reviewed this SHA with no findings.

Still outstanding for the two of you: CodeRabbit last reviewed `47d727ae`, Codex last reviewed `7217bfed`. Since those, the only change is `5030cea5`, which scopes the blocker summary to the V8 builds actually tested (it still concluded "Chrome and Edge cannot run it" — the inference Codex correctly flagged). Doc-only; no workflow logic has changed since `47d727ae`.

@coderabbitai review
@codex review

louistrue:
Head is `06520cfa`. Both of your findings on `5030cea5` are taken:

- **`WebAssembly.validate()` validates, it does not execute.** Reworded: the probe is the validation check, the determinism step is the execution check.
- **`NODE_OPTIONS` allowlist.** Now says `NODE_OPTIONS` honours an allowlisted subset and Node refuses to start when given an option outside it, whereas direct argv bypasses that allowlist.

The same commit also closes the last fail-open path, found by an adversarial pass: `node -e` returning non-zero was read as "engine says no", so a crash, an OOM, or a mangled base64 constant all reported `unsupported` and went quietly green — a mangled constant would have made the lane permanently blind. That is the same failure class this lane was rewritten to remove.

The verdict is now a printed sentinel (`PROBE_YES`/`PROBE_NO`/`PROBE_ERR`), never an exit code, with a length assertion for the mangled-constant case. Verified by execution:

| case | before | after |
|---|---|---|
| probe crashes (exit 139) | `support=unsupported`, rc 0 | rc 1, no verdict written |
| base64 constant truncated | permanently blind green lane | length guard fires, rc 1 |
| newline in engine string | forged `support=default` in `$GITHUB_OUTPUT` | flattened, 0 forged lines |

Also wrapped probe invocations in a 30s timeout so a hanging flag cannot burn the job timeout. Live lane run green on this content.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Head `5d786779`. Adversarial re-pass returned APPROVE on `06520cfa`, confirming by execution on both bash 3.2 and 5.3 that the fail-open is closed (crash stub and truncated module both exit 1 with no verdict; a timeout kill mid-write cannot flip a verdict, since a truncated `PROBE_NO` can never spell `PROBE_YES`). I have taken its two remaining non-blocking nits rather than ship them:

- **Attribution failure was silent.** The `tr` pipeline added for output-flattening also swallowed a failing `node -p`, so a runner unable to report its V8 version would have recorded a verdict attributed to "V8 " and carried on. Now captured before the pipeline and red on failure — a verdict with nothing to attribute it to is exactly what this lane exists to avoid.
- **`.github/workflows/wide-arithmetic.yml:271`** still said the determinism step runs "once the engine can actually **execute** the module". The probe validates; this step is the execution check. Tenth stale sibling of a claim already corrected elsewhere in this PR.

Probe matrix re-verified after the change: `node -p` failure exits 1 with no verdict written, crash stub still red, shipped and flag paths unchanged.

No probe *logic* changed — the verdict path, sentinel handling and `if:` conditions are identical to the ones just approved.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Both taken — these are the two best findings on this PR, because each one silently defeats the lane's entire purpose rather than just failing loudly.

**1. The length guard only caught truncation.** You're right that substituting a base64 character still decodes to 102 bytes. Demonstrated:

```
orig bytes: 102   substituted bytes: 102  -> length guard passes: true
validate(orig): false   validate(substituted): false
orig sha256: d83c3e46...083503
bad  sha256: d3fcf8f4...752c29
```

So the guard passes, validate says no, the lane records `unsupported` — and stays green forever even after the engine ships support. The module is now pinned by **sha256 of its decoded bytes**, which catches any corruption, not just length changes. Verified: the substituted module now yields `PROBE_ERR:module digest … is not the pinned probe module`, exit 1, no verdict written.

**2. `--wasm-staging`.** Also correct, and confirmed on a real engine — `node --v8-options` lists `--wasm-staging (enable staged wasm features)`, and my extraction accepted only `--experimental-wasm-*` names containing wide/arith/128, so it would never have been tried. Through exactly the rollout phase this early-warning lane exists to catch, it would have answered "unsupported". Staging is now a candidate and bypasses the proposal-name filter, since it is a catch-all rather than a proposal flag.

Verified against a stub engine that accepts the opcodes **only** under `--wasm-staging`: `support=flag`, `node_args=--wasm-staging`. Real Node 22/26 still correctly report unsupported, with staging enumerated and tried.

Head `84b96cba`. Also confirming the earlier `Validate Server Binary (linux-x64)` red was infrastructure — rustup timed out fetching `channel-rust-nightly.toml` (`tcp connect error: Connection timed out`) before compiling anything, `linux-x64-musl` passed on identical code in the same run, and a re-run passed. CI is now 20 pass / 5 skip / 0 fail.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Taken — with one correction to the premise, which does not change the conclusion.

**The finding is valid, the stated mechanism is not live today.** Plain `run:` steps on this runner use `bash -e`, not `-eo pipefail` — confirmed from the run log for this exact step:

```
Probe engine support for wide-arithmetic (gate 2)   shell: /usr/bin/bash -e {0}
```

Under `bash -e` the pipeline takes grep's status, so the check has genuinely been working. But the fragility is real and I reproduced your mechanism exactly:

```
v8-options bytes: 108185
WITH pipefail:    staging MISSED (rc=141)
WITHOUT pipefail: staging detected
```

So it is one `defaults: run: shell: bash` away from switching on — and GitHub's `shell: bash` *is* `-eo pipefail`. A silent gap in the flag search is precisely the failure mode this lane keeps getting bitten by, so it should not depend on which bash flags a future edit happens to pick. Both reads of `$v8opts` now use here-strings, no pipe, no SIGPIPE.

Verified under **both** `bash -e` and `bash -eo pipefail`: real node classifies unsupported, and a stub engine that accepts the opcodes only under `--wasm-staging` is still detected as `support=flag`, `node_args=--wasm-staging`.

Head `813f2182`.

@coderabbitai review
@codex review
@greptileai review

louistrue:
Both taken. Head `00e2c81f`.

**1. (major) One unusable flag ended the whole search.** Correct, and the consequence is worse than it looks: on an engine where `--wasm-staging` crashes but a dedicated wide-arithmetic flag works, the step went red without ever trying the flag that answers the question. A candidate that cannot be evaluated is now skipped with a warning and the search continues.

But an incomplete search must not end in a confident `unsupported` either, so unevaluated candidates are counted and turned into an error rather than a verdict. `classify()` now reports the diagnosis and the caller decides severity — fatal for the flagless probe, a warning for one candidate.

**2. (major) Hyphen-only flag matching.** V8 accepts `--foo-bar` and `--foo_bar` interchangeably (verified: `node --experimental_wasm_jspi` and `--experimental-wasm-jspi` both accepted). Today `--v8-options` prints hyphens, so this was latent — but we parse what it prints, and a printer that ever switched would yield zero candidates and a silent `unsupported`. Both spellings now match, and staging's spelling is taken from the listing rather than hardcoded.

Verified under `bash -eo pipefail`:

| case | result |
|---|---|
| staging crashes, dedicated flag works | warns, continues, **finds the dedicated flag** |
| every candidate crashes | exit 1, **no verdict written** |
| V8 prints underscore form | detected, `node_args=--wasm_staging` |
| real node | still `unsupported` |

Lane green on this head: `probing node v26.5.1 / V8 14.6.202.34-node.24` → rejects, no enabling flag.

@coderabbitai review
@codex review
@greptileai review

louistrue:
@codex review

You last reviewed `84b96cba` and have skipped `813f2182` and `00e2c81f`. Both deltas since are small and worth your eyes specifically because one of them is **your own finding**:

- `813f2182` — your SIGPIPE finding. Both reads of `$v8opts` moved from pipes to here-strings. Worth confirming I implemented it the way you meant, and that I did not introduce a new pipe.
- `00e2c81f` — a candidate flag that cannot be evaluated is now skipped with a warning instead of aborting the search, with unevaluated candidates counted and turned into an error rather than a confident `unsupported`; flag matching broadened to V8's underscore spelling.

Everything else on this head is clean: CI 20 pass / 5 skip / 0 fail, Greptile clean, a local CodeRabbit CLI run returning 0 findings across 11 files, and an adversarial pass that executed the full stub matrix on bash 3.2 and 5.3 under both `-e` and `-eo pipefail`.

PR #1930, opened 2026-08-01
fix(viewer): consume the exportMenu slot so extension exporters are reachable (#1907)

### LINKED ISSUE #1907: Extension `exporters` contributions load but are never rendered — viewer does not consume the `exportMenu` slot
**Type:** bug / gap
**Area:** extensions / viewer
**Context:** raised while answering an external request about calling external OpenAPI
endpoints from IFC Lite

## Summary

The extension model declares an `exporters` contribution point. A third-party extension
declaring one validates, loads, and registers — but nothing in the viewer ever displays
it, so it is unreachable for users.

## Detail

Declared and wired on the extension side:

- `ExporterContribution` — `packages/extensions/src/types.ts:278-284`
  (`{ id, name, mimeType, extension, handler }`)
- listed in `ManifestContributions` — `types.ts:187-197`
- activation event `onExporter:<id>` — `types.ts:147`, `manifest/validate.ts:214`
- capability `export.create:<format>` — `capability/catalogue.ts:48`
- registered to slot `exportMenu` — `packages/extensions/src/host/loader.ts:285-287`
- documented — `docs/guide/extension-authoring.md:115`

Not wired on the viewer side. `useSlotContributions` is only consumed for
`commandPalette`, dock, toolbar and contextMenu:

- `apps/viewer/src/components/extensions/*`
- `CommandPalette.tsx:269`
- `EntityContextMenu.tsx:433`

Nothing reads `exportMenu`.

## Impact

The documented extension authoring path for custom exporters produces an extension that
appears to install correctly and then does nothing. This is the natural place to host
third-party integrations — e.g. posting the active model to an external OpenAPI endpoint
and loading the returned IFC back into the same view — so the gap blocks a whole class
of community contributions.

## Proposed

1. Consume the `exportMenu` slot in the viewer's export UI, same pattern as the existing
   toolbar and context-menu slot consumers.
2. Add an end-to-end test that installs a fixture extension declaring an exporter and
   asserts it appears and runs.
3. While there: consider whether a symmetric `importer` / round-trip contribution is
   wanted, so an extension can push the model out and bring a result back into the view.

BIMvoice:
## What and why

Closes #1907 — implements steps 1 and 2 of your proposal. See the open question on step 3 below.

An extension declaring an `exporters` contribution validated, loaded, and registered to `exportMenu`, and then nothing displayed it and nothing could run it. Two halves were missing, so this adds both:

- **`host-exporters.ts`** — the sibling of `host-commands.ts`. Resolves the owning extension, activates, fires `onExporter:<id>`, wraps the handler via `wrapEntrySource`, runs it. `ExtensionHostService.runExporter()` delegates to it, mirroring `runCommand()` including the dead-sandbox retry and the "never pre-emptively dispose" note.
- **`ExtensionExportSlot`** — the slot consumer, rendered by `ExportDialog`. Output goes through `downloadFile` from `lib/export/download.ts`, per `AGENTS.md`, so tour progression still observes the save.

## The one thing I checked rather than assumed

An exporter's `handler` is **not** in `manifest.entry`. `ManifestEntry` (`types.ts:180`) declares only `activate` / `deactivate` / `commands` / `triggers` — there is no exporters map. The path lives on the contribution and `validate/cross-ref.ts:84` already resolves it against the bundle file set, exactly like `lenses[].evaluator`.

This mattered for scope: had `handler` been a command id, the fix would have needed a manifest schema change and a spec conversation instead of a viewer-only patch. There is a test pinning it (`resolves the handler from contributes.exporters[].handler, not manifest.entry`), and swapping the lookup to the command-style one fails 6 of the 13 new tests.

## Two deliberate behaviours worth a look

**A handler returning nothing raises** instead of writing a 0-byte file. A user who clicks Export and silently receives nothing has no way to tell a broken extension from a broken viewer.

**Output coercion accepts numeric-keyed objects.** Typed arrays flatten across a sandbox boundary, so a handler returning `Uint8Array` can arrive as `{0: 12, 1: 34, …}`. Accepting only `Uint8Array` would reject valid extensions for a reason the author cannot see or fix. Strings, `ArrayBuffer`, views, and number arrays are handled too; anything else raises with what it actually got.

## How it was verified

- `npx tsx --test src/services/extensions/host-exporters.test.ts` — **13 passed**, driven against a fixture extension (manifest + bundle file + storage record), evaluating the real `wrapEntrySource` output rather than a stub, so these break if the wrapper contract changes.
- **Mutation-checked** the central assumption: resolving the handler via `manifest.entry` fails 6 of 13; restoring returns to green.
- `pnpm --filter @ifc-lite/viewer test` — **2089 passed, 0 failed** (2078 before).
- `pnpm test` 74/74 · `pnpm typecheck` 33/33 · `check-test-wiring` clean.
- No changeset: every change is under `apps/viewer`, which is not published.

## Where I stopped, and why

**The "appears" half of your requested end-to-end test is not here.** The "runs" half is covered end-to-end above, but asserting the button renders needs React rendering, and `apps/viewer` has **zero** `.test.tsx` files — the suite is `tsx --test` with no DOM. Standing up a rendering harness, or a Playwright fixture that installs a signed extension bundle, is a much larger call about test infrastructure than a bug fix should make unilaterally. Tell me which you'd prefer and I'll add it, here or as a follow-up.

**Step 3 (a symmetric `importer` / round-trip contribution) is left for you.** It is a new contribution point and an API design decision, not a gap in an existing one. Worth its own issue.

## Checklist

- [x] A test asserts the new behavior through a fixture or a stated invariant — fixture extension, mutation-checked
- [x] Published `packages/*` touched: none — `apps/viewer` only, so no changeset and no `api-surface:update`
- [x] No geometry output changed
- [x] House rules: no `as any` / `@ts-ignore`, no silent `catch {}`, new UI in its own component rather than growing the 840-line `ExportDialog.tsx`, no repo-wide `cargo fmt`

## Summary by CodeRabbit

- **New Features**
  - Added support for extension-provided exporters in the export dialog.
  - Export results can be downloaded with appropriate filenames and file types.
  - Export actions display loading, success, and error states.
  - Exporters support text and binary output formats.

- **Bug Fixes**
  - Improved filename sanitization, extension handling, and length limits.
  - Prevented invalid or empty exporter results from being downloaded.
  - Prevented concurrent export actions from running simultaneously.
  - Improved handling of exporter activation and execution failures.


BIMvoice:
Pushed 9f6d946a addressing all three review findings.

**Ownership discarded** (@greptile-apps P1) — the owning extension id now threads through from `SlotContribution.extensionId` into `runExtensionExporter`, and the not-found error names the extension when the run is scoped to one. `ExtensionHostService.runExporter` takes it as an optional second argument.

**Sparse numeric keys** (@greptile-apps P2) — took your suggested dense/in-order check. `{0: 65, 100: 66}` now returns null and the caller raises, instead of allocating two bytes and silently dropping the out-of-range write.

**Filename truncation** (@coderabbitai) — confirmed by reading `sanitizeFilename`: default `maxLength: 60` with the slice applied after the extension is appended. Stem and extension are now sanitized separately with the stem budgeted around the extension.

Two notes on how I applied that last one, since both differ from the suggested patch:

The logic moved to a new `apps/viewer/src/lib/export/extension-filename.ts` rather than staying inline in the component. `apps/viewer` has zero `.test.tsx` files and no DOM in `tsx --test`, so a pure module is the only way to test this without introducing a rendering harness — which felt like a maintainer's call, not mine.

The extension's own sanitize budget is capped independently at 16 rather than derived from `maxLength`. Deriving it lets a pathological extension consume the whole budget and push the result *past* the limit the function documents: `buildExportFilename('model', 'x'.repeat(80), 60)` returned a 62-character name. There is a test pinning that now.

Verification: 23 targeted tests pass, viewer suite 2099 pass / 0 fail, `pnpm test` 74/74, `pnpm typecheck` 33/33. Each fix has a regression test verified to fail when that fix alone is reverted — 17→15, 17→15, and 6→4 respectively.

louistrue:
Reviewed 9f6d946a. Nothing pushed to this branch, it is yours to finish. I had a parallel fix in progress when you pushed, and I threw it away rather than clobber you, because your approach to the filename split is better than mine was.

**Your three fixes are correct and I verified them rather than taking your word.** The confused-deputy fix is real: I reverted the `extensionId` filter at `host-exporters.ts:116` and got exactly the 15/17 you claimed, failing precisely the two ownership tests. I also confirmed the identity invariant it rests on actually holds, so a rendered button's strict match cannot fail: `loader.ts:173` rejects `manifest.id !== record.id`, `:215` registers under `record.id`, `:221` caches under `record.id`, and `slot-registry.ts:43` hard-throws on mismatch. 17/17, 6/6, viewer suite 2099 passing, typecheck 33/33.

Your tests evaluate real `wrapEntrySource` output rather than stubs. That is the right standard and it is why the mutation check was able to prove anything.

Four things still open, most important first.

**1. Zero-length output still downloads a 0-byte file with a success toast.** This is the one that matters, because it makes the docstring at `host-exporters.ts:105-107` false: "Never returns a silent empty result. A user who clicks Export and gets nothing has no way to tell a broken extension from a broken viewer." A handler returning `''`, `[]`, or `new Uint8Array(0)` coerces to zero length, so the `data === null` check at `:174` never fires. I probed all three. Fix is `if (data === null || data.length === 0)` plus an `describeReturn` case for each, and a test per shape.

**2. Byte values are still unvalidated.** You fixed sparse and colliding keys, and I confirmed both are rejected now. But the value half is untouched in both branches (`:68-69` and `:83`): `[300, -5, 1.7]` gives `Uint8Array[44, 251, 1]` and `{0:300, 1:-5}` gives `[44, 251]`, silently wrapping mod-256 and truncating floats. Your commit message says "rejects anything else, so a malformed return raises instead of writing a corrupt file", which is broader than what the code does. Want `Number.isInteger(v) && v >= 0 && v <= 255` in both branches.

**3. `host-exporters.ts:192` still emits "returned a object".** User-visible in the toast. `typeof value === 'object' ? 'an object' : \`a ${typeof value}\``.

**4. The `as never` casts grew rather than shrank.** `host-exporters.test.ts:97` remains and `:206` adds a second. AGENTS.md line 17 bans that escape-hatch family because it erases type-checking of the whole fixture against `RunExporterDeps`, so a signature drift compiles silently. There is a clean way out here: narrow `RunExporterDeps` at `:36-42` to structural picks, `storage: Pick<IdbExtensionStorage, 'listExtensions'>`, `loader: Pick<ExtensionLoader, 'getBundle'>`, `runtime: Pick<ExtensionRuntime, 'activate' | 'deactivate'>`, `dispatcher: Pick<ActivationDispatcher, 'fire'>`. `Pick` strips `ExtensionLoader`'s `private bundleCache`, which is the thing actually forcing the cast, and then both `as never` can just go.

Two smaller ones, your call:

**`extensionId` is still optional**, so the old first-enabled-match scan survives for any caller that omits it. The slot is the only caller and it always passes it, so exposure is zero today, but making it required closes the door properly and deletes the unscoped error branch.

**Placement of `extension-filename.ts`.** Important: you did *not* duplicate the sanitiser, you import and compose it, so this is not the #1299 drift bug and I want to be clear I am not accusing you of that. But the stated reason for a separate module ("testable without a React rendering harness") argues against the component, not against `download.ts`, which is itself a plain module and already has `download.test.ts` in the same harness. `download.ts` declares itself the canonical home for exactly this ("one filename sanitiser and one download trigger"), and about 15 call sites hand-append extensions today and would benefit from finding a budget-aware joiner sitting next to the sanitiser. Also "extension-filename" puns file-extension against viewer-extension. Moving the two functions verbatim and merging the test file is zero-risk.

One thing that is *not* a defect: `buildExportFilename('model','csv',0)` returns `"m.c"`, which exceeds a `maxLength` of 0. No caller passes `maxLength` at all, so it is unreachable. Mentioning it only so it does not surprise you if a bot flags it.

Happy to take any of this myself if you would rather, but it is all well-defined and it is your PR.

BIMvoice:
Pushed 7aadc748 — all six items.

**1. Zero-length output.** You were right that it made the docstring false. Only `null` was rejected, so `''`, `[]` and `Uint8Array(0)` all coerced to zero length and downloaded a 0-byte file with a success toast. Now rejected with `describeEmptyReturn` naming which shape arrived, one test per shape. Reverting the check alone fails 3 of 22.

This also killed a message the previous round had added telling authors *"for an empty result, return `''`"* — i.e. steering them straight into the bug. Good catch that this needed to be a rejection rather than better advice.

**2. Byte validation.** Both branches now require `Number.isInteger(v) && v >= 0 && v <= 255`. I verified your examples directly: `[300, -5, 1.7]`, `[300]`, `[-1]`, `[1.7]` and `{0:300, 1:-5}` all rejected; `[1,2,3]` and `{0:1,1:2}` still pass. Reverting fails 2 of 22. You were also right that my commit message overclaimed — it said malformed returns already raised, which was broader than the code did.

**3.** Already correct from the prior round; verified, unchanged.

**4. `as never` gone**, via your `Pick<>` diagnosis — `ExtensionLoader`'s private field was indeed what forced it. Worth reporting: with the casts removed, four fixtures turned out never to have type-checked at all and had to be filled out to their real shapes. It was hiding drift, not just untidy.

One step further than you asked. Two `as unknown as` casts survived on the `sdk` field, since `BimContext` is a class no plain object satisfies. But `ExtensionContextV1.bim` is declared `unknown` (`runtime.ts:152`) and this module only forwards the value, never calls into it — so `BimContext` overstated the contract and the casts existed only to satisfy the overstatement. `sdk` is now `ExtensionContextV1['bim']` and both casts are deleted. `rg 'as never|as any|@ts-ignore'` over both directories returns only prose hits.

**5. `extensionId` required**, unscoped branch deleted. No runtime test can discriminate this since the slot always passed it, so I verified the compile-time regression instead: omitting the argument gives `TS2554`.

**6. Moved into `download.ts`** (153 lines), `extension-filename.ts` deleted, tests merged into `download.test.ts`. Agreed on the reasoning, including the pun.

Left `buildExportFilename('model','csv',0)` alone as you flagged.

Verification: 38/38 targeted, viewer 2104 passing, `pnpm test` 74/74, typecheck 33/33.

**Two open items I'd rather you decide than guess at.** CodeRabbit wants a rendering regression test for the dialog — that is the same "appears and runs" half I flagged earlier, and `apps/viewer` has zero `.test.tsx` files and no DOM in `tsx --test`. Standing up a harness, or a Playwright fixture that installs a bundle, is a test-infrastructure decision. It also flags `ExportDialog.tsx` at 851 lines against the ~400 rule — fair, and I cited that rule myself, but the extraction is unrelated to this PR and I did not want to bundle it. Happy to do either.

louistrue:
@coderabbitai review

louistrue:
@coderabbitai review

PR #1933, opened 2026-08-01
fix(provenance): bind storey containment and opening hosting into the DAG (F1/F2/F3)

louistrue:
**Stacked on #1900 — merge that first.** Supersedes the standalone `fix/dag-producer-bindings` (`9b4edba9`), which should be **abandoned rather than merged**, or the duplicate `hostId` conflict comes back.

Closes three producer bugs found by an adversarial attack on `node-hash-v0` run before the freeze. None is a format bug: the format already carries what was missing, the spec already names it, and golden vector `rel-voids` already pins it.

## Reproduced before, fixed after

| | before | after |
|---|---|---|
| element moved between storeys | root **identical** | root changes |
| two elements swap storeys | root **identical** | root changes |
| opening re-hosted A→B | root **identical** | roots differ |
| certificate verified against a different base | `ok: true` | `base-root-mismatch` |

## The collision with #1900 dissolved on reading it

B4.2 had already built a **better hosting op model** than mine — containment as an invariant, derived void re-cuts, `SpatialRejectionError`, remove-cascades-to-openings, host coupling expressed spatially via `withHostRegion`. Its `canonicalStateBytes` line for `hostId` is character-for-character identical to the one I wrote independently.

So this drops my **entire F2 op-model half** — the `hostId` field, the add/remove validation, and critically the `writtenNodes` host coupling — and keeps b42's.

**That is what dissolves the ablation problem.** My structural coupling would have refused host/opening pairs *structurally*, stopping B4.2's ablation from isolating the spatial rule for exactly the pairs B4.2 is about. It is gone, and b42's measurements reproduce unchanged: **873 auto-merged, 0 unsound, 8.78% false-conflict, 9 spatial-only true conflicts**, and the ablation still finding **9 unsound clears** with the rule disabled.

What survives is only what b42 declined to build: F1's `RelatingStructure` plus the storey's own `element` node, F2's `IfcRelVoidsElement` node, F3 falling out of both, and four script producers ported verbatim.

## The adjudication, and the test it overturned

b42 had committed a test **asserting the bug**:

```
it('hosting is NOT part of any node payload, so it cannot perturb a node hash')
  expect(await hashModelState(withHost)).toBe(await hashModelState(withoutHost));
```

with a comment calling it *"a v1 spec question, not a change this package may make unilaterally."* Inverted to `not.toBe` with a `rel-voids` node assertion, and both the comment and module docstring rewritten to say why this is a **producer obligation, not a format change** — the measurable argument being that `node-hash.ts` is untouched and **0 of the 55 frozen vectors move**.

Added the reachable two-coincident-walls construction, geometrically valid against b42's own containment check: under lazy cuts the host meshes are byte-identical, so the `IfcRelVoidsElement` node is the only thing separating the two states.

## Verification

**182 tests** (b42's 176 + 7 new − 1 duplicate). Negative control: reverting only the DAG binding fails **7 of 7**. `g0` / `g1` / `g2-footprint` / `g2-merge-soundness` all exit 0. B3.5 golden re-blessed on the same 7 paths with the same before/after as the standalone branch; no act1/3/5 movement; b42's act4 battery figures unchanged.

## One thing to know before the merge up

**E9, the numeral gate, is not an ancestor of `feat/b42-spatial-merge`** — b42 branched from `5c1ed287`, before `9039d402` added it. So E9 has never run against b42's prose and will fire for the first time when b42 merges into the docs branch. Worth knowing before that merge rather than after.

### DIFF
--- a/packages/provenance/src/merge-model.ts
+++ b/packages/provenance/src/merge-model.ts
@@ -100,13 +100,23 @@
  * B4.2 headline to switching them is measured and published rather than
  * argued. See `merge-battery.ts`'s `runDerivedCutSensitivity`.
  *
- * Wire format: nothing here changes node-hash-v0 serialization. Hosting lives
- * in the model state and in {@link canonicalStateBytes}; the cut is recorded
- * inside the existing `GeometryMeshPayload` vertex arrays, so the frozen
- * `geometry-mesh` encoding hashes it unchanged (a different cut is a
- * different mesh, which is correct). A v1 spec would model hosting as its own
- * `IfcRelVoidsElement` `relationship` node; that is a spec question, not a
- * change this module is allowed to make.
+ * Wire format: nothing here changes node-hash-v0 serialization. The cut is
+ * recorded inside the existing `GeometryMeshPayload` vertex arrays, so the
+ * frozen `geometry-mesh` encoding hashes it unchanged (a different cut is a
+ * different mesh, which is correct).
+ *
+ * Hosting lives in the model state, in {@link canonicalStateBytes}, AND in
+ * the DAG: {@link buildStateDag} emits one `IfcRelVoidsElement`
+ * `relationship` node per hosted element, carrying both of that
+ * relationship's singular EXPRESS roles. An earlier draft of this module
+ * deferred that to "a v1 spec question"; it is not one. node-hash-v0 already
+ * has a `relationship` kind, the frozen golden vector `rel-voids` already
+ * pins this exact two-role shape, and the spec names both roles -- so
+ * omitting the node was a PRODUCER bug, not a format limitation. Leaving it
+ * out meant `hostId` moved `canonicalStateBytes` but not the Merkle root, so
+ * two genuinely different buildings (same opening, different host) committed
+ * to the same root and a commutation certificate issued over one verified
+ * against the other.
  */
 
 import { ProvenanceDag, type AnyNodeSpec } from './dag-engine.js';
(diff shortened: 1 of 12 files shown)


louistrue:
## CI evidence: the automatic checks on this PR are vacuous

Recording this because #1933's green ticks do **not** mean what they look like, and the same is true of #1900.

```
CodeRabbit  pass  0  Review skipped: reviews are disabled for this base branch
```

That is a **false pass**, not a review. And neither `test.yml` nor `moonshot.yml` ran at all — both are scoped `pull_request: branches: [main]`, and this PR targets `feat/b42-spatial-merge`. The only substantive automatic check that executed is `parity (in-tree fixtures)` (8s); the Vercel builds were "Canceled by Ignored Build Step".

This is **structural to the stack, not to this PR** — #1900 (base `docs/moonshot-finishing-plan`) shows the identical vacuous set. Real CI first runs at #1897 → `main`, i.e. after three levels of change land together.

### So the lane was dispatched manually against this branch

[Run 30689501751](https://github.com/LTplus-AG/ifc-lite/actions/runs/30689501751) — `workflow_dispatch`, `holter=true` (full fixture scope), **conclusion: success**.

| step | result |
|---|---|
| E1 provenance unit suite | success |
| E2a G0 certificate demo (169 MB Holter, verify < 500 ms) | success |
| E2b G1 memoized recompute (mesh + 100k-node scale) | success |
| E2c G2 footprint tightness (kernel-meshed) | success |
| E2d G2 merge soundness (1000-schedule battery) | success |
| E3a B3.5 five-act demo | success |
| **E3b B3.5 deterministic subtree vs committed golden** | **success** |
| E4 certified optimization (v1 + v2 chains) | success |
| E5a/E5b chain verification (full + spot) | success |
| E6a/E6b tamper batteries (5 v1 + 18 v2) | success |
| E7 gradient battery (1000 points) | success |
| E8 world-gym determinism (20/20 byte-identical) | success |

E3b passing on a clean Linux runner at pinned Node 22.14.0 is the load-bearing one: it confirms the re-blessed golden reproduces byte-for-byte off this laptop.

E3c and E9 do not exist on this branch — `assert-b35-golden.mjs --self-test` and `check-report-numerals.mjs` arrived in `9039d402`, which is not an ancestor of `feat/b42-spatial-merge` (b42 branched at `5c1ed287`). **E9 will run against b42's prose for the first time when #1900 merges up**, not before.

### Review status

CodeRabbit **CLI** 0.7.1, `--base origin/feat/b42-spatial-merge`: 3 findings, all resolved, re-review **0 findings**. Hosted CodeRabbit has **not** reviewed this PR and will not while the base is a feature branch — so the house "both hosted and CLI" rule is only half satisfied. Cleanest fix is to retarget to `docs/moonshot-finishing-plan` once #1900 merges.

PR #1968, opened 2026-08-02
fix(ids,lens): inherit type properties per property, not per property set (#1913)

### LINKED ISSUE #1913: IDS validation is not seeing Pset_CoveringCommon.SurfaceSpreadOfFlame
When chekcing for a value in Pset_CoveringCommon.SurfaceSpreadOfFlame I get a failure:

Image

While Bonsai isn't detecting the failure:

Image

because the Property is present and correct:

Image

This is just one Property I came across, I don't know if there are more.

Thanks!

louistrue:
Closes #1913.

## The bug

IDS reports `Pset_CoveringCommon.SurfaceSpreadOfFlame` as missing on elements that carry it, while other tools resolve it:

```
Property "SurfaceSpreadOfFlame" not found in "Pset_CoveringCommon".
Available: Pset_CoveringCommon.IsExternal, Pset_CoveringCommon.Reference
```

## Root cause

IFC inherits type properties **per property, not per property set**. An occurrence and its `IfcTypeProduct` routinely both carry a set of the *same name* holding *different* properties. In the reported model the `IfcCovering` and its `IfcCoveringType` split `Pset_CoveringCommon` exactly that way:

| | property set | properties |
|---|---|---|
| occurrence, via `IfcRelDefinesByProperties` | `Pset_CoveringCommon` | `IsExternal`, `Reference` |
| type, via `IfcRelDefinesByType` → `HasPropertySets` | `Pset_CoveringCommon` | `Combustible`, **`SurfaceSpreadOfFlame`**, `ThermalTransmittance` |

`appendInheritedPropertySets` deduped by set **name**:

```ts
const seen = new Set(out.map((p) => p.name));
for (const pset of inheritedPsets) {
  if (seen.has(pset.name)) continue;   // drops the ENTIRE inherited set
```

so the whole type-side set was discarded and its three type-only properties became invisible. The "Available:" list in the error is exactly the occurrence-side pair, which is the tell.

The same defect existed independently in the viewer's Lens adapter (`lib/lens/adapter.ts`), where it silently removed those properties from Lens grouping and filtering. `getPropertyValue` in that same file was already correct — it falls through to the type sets per property — so the adapter disagreed with itself.

## The fix

`@ifc-lite/parser` gains `mergeInheritedPropertySets(ownSets, inheritedSets)`: union the two sides per property, occurrence wins on a property-name collision (the more specific definition). This matches `IfcRelDefinesByType` semantics and what `ifcopenshell.util.element.get_psets` does. Both call sites now use it, so the rule lives in one place next to `extractTypePropertiesOnDemand` rather than as two divergent copies. Neither input is mutated — both consumers hand it cached extractor results.

Only the collision case changes: a type set whose name the occurrence does not use was already appended and still is; a property defined on both sides still resolves to the occurrence's value; a property on neither side is still absent.

## Verified against the reported model

The reporter's model + IDS were supplied privately and are **not** committed. Running the CLI IDS validator on them, base vs branch:

| failing requirement (×27 entities) | base | branch |
|---|---|---|
| `Pset_CoveringCommon.SurfaceSpreadOfFlame` must equal "B" — *not found* | FAIL | **PASS** |
| `Pset_ManufacturerTypeInformation.ModelReference` must equal "79 x 18" | FAIL | FAIL |
| `Pset_CoveringCommon.Finish` must exist | FAIL | FAIL |

The two remaining failures are genuine content, not engine defects, and I checked both in the file rather than assuming:

- `ModelReference` really is `IFCLABEL('d x h x l')` on that type — a real mismatch against the IDS literal `79 x 18`. It is red in the reporter's own screenshot too.
- `Finish` occurs **0 times** in the entire file. It is in the schema's `Pset_CoveringCommon` but absent from this model, so "must exist" correctly fails.

The merged set is now reported as `Pset_CoveringCommon.IsExternal, Reference, Combustible, SurfaceSpreadOfFlame, ThermalTransmittance` and `Pset_CoveringCommon.SurfaceSpreadOfFlame = B` passes.

## Tests

Fixtures are synthetic reproductions of the shape — no reporter data.

- `packages/parser/test/merge-inherited-property-sets.test.ts` — the helper: union on collision, occurrence value wins, non-colliding set appended, neither input mutated, empty-inherited passthrough.
- `packages/ids/src/bridge/type-inherited-merge.test.ts` — parses a real IFC4 fixture with the split `Pset_CoveringCommon` and drives `checkPropertyFacet`: the type-only property resolves, the occurrence-only ones still do, one merged set carries all five, and a property on neither side is still absent.
- `apps/viewer/src/lib/lens/adapter.test.ts` (new file — the adapter had no tests) — the merged set, plus `getPropertyValue` agreeing with `getPropertySets` on both sides.

**Mutation-checked per suite:** reverting only `packages/ids/src/bridge/properties.ts` fails 2 of the 4 IDS tests (the 2 controls stay green); reverting only `lib/lens/adapter.ts` fails 1 of the 2 lens tests, and the one that stays green is the `getPropertyValue` control — which documents that the two accessors disagreed before this change.

## Verification

- `packages/parser`: 322 passed
- `packages/ids`: 300 passed
- `apps/viewer`: 2172 tests, 0 failed
- `pnpm typecheck`: 33/33
- `scripts/api-surface.json` updated for the one new export; changeset included (`parser` minor, `ids` patch)


louistrue:
Thanks — all three findings were real and are fixed in 43c5670.

**Duplicate same-named occurrence sets** (Greptile P1 / Codex P2). Confirmed reachable: `extractPsetsFromIds` emits one entry per `IfcRelDefinesByProperties` and never dedupes by name, so an element can legitimately carry two `Pset_CoveringCommon` sets. Codex's consequence framing is the accurate one — `checkPropertyFacet` requires *every* set matching the facet's `propertySet` constraint to satisfy it, so augmenting only the first would leave its twin bare and reproduce the exact false "not found" this PR fixes, just in a narrower case. The merge now tracks every index per name. New test `augments EVERY same-named occurrence set, not just the first` pins it, including that each set keeps its own occurrence-specific properties.

(Greptile's stated mechanism — IDS preferring the inherited value over a later occurrence value — isn't what happens: `singlePropertyPasses` evaluates candidates within a set rather than resolving one winner across sets. The defect is real, the failure mode is the one above.)

**Module placement** (Codex P1). Agreed — `on-demand-extractors.ts` was already 1167 lines, well past the ~400-line rule, and the merge rule is independent of how either side was extracted. It now lives in `packages/parser/src/property-set-merge.ts`.

**Documentation** (Codex P1). `docs/guide/parsing.md` gains a "Type-Inherited Properties" section covering the per-property precedence, the duplicate-set behaviour and the immutability contract. The snippet is typechecked by `pnpm docs:check-samples` (251 snippets clean).

Re-verified: parser 323 passed, ids 300 passed, viewer 2172 tests / 0 failed, `pnpm typecheck` 33/33, API surface unchanged (the export just moved modules).

PR #918, opened 2026-06-04
feat: cloud storage connection (Dropbox, Google Drive, OneDrive/SharePoint)

louistrue:
## Summary

Adds cloud storage integration to ifclite, allowing users to browse and load IFC files directly from Dropbox and Google Drive without downloading to disk first. Files stream directly from the provider to the browser—they never pass through ifclite servers.

## Key Changes

- **Cloud provider abstraction** (`apps/viewer/src/services/cloud/types.ts`): Provider-agnostic interface for cloud storage operations (list folders, download files)

- **OAuth 2.0 flow** (`server/cloud-oauth/oauth-core.ts`, `server/cloud-oauth/oauth-handlers.ts`): Generic, provider-agnostic OAuth implementation handling authorization code flow with refresh tokens. App secret and refresh token stay server-side in httpOnly cookies; browser only holds short-lived access tokens in memory.

- **Provider implementations**:
  - Dropbox (`server/dropbox/dropbox.ts`, `apps/viewer/src/services/cloud/dropbox.ts`)
  - Google Drive (`server/google/google.ts`, `apps/viewer/src/services/cloud/google-drive.ts`)

- **Browser-side OAuth base** (`apps/viewer/src/services/cloud/oauth-provider-base.ts`): Shared logic for popup-based consent flow, token management, and connection lifecycle. Uses localStorage to signal consent completion (works around `Cross-Origin-Opener-Policy: same-origin` header).

- **Importer UI** (`apps/viewer/src/components/viewer/cloud/CloudImportDialog.tsx`): Dialog for browsing cloud folders, downloading files, and integrating with existing `addModel()`/`loadFile()` pipeline. Shows progress during download.

- **Vercel edge endpoints** (`api/{dropbox,google}/{auth-start,auth-callback,token,disconnect}.ts`): Four routes per provider handling OAuth flow and token refresh.

- **Toolbar integration** (`apps/viewer/src/components/viewer/MainToolbar.tsx`): Added cloud import button to main toolbar.

- **Documentation** (`docs/guide/cloud-import.md`): Architecture overview and provider addition guide.

- **Tests** (`tests/api/cloud-oauth.test.ts`): Unit tests for OAuth core, provider specs, and token exchange.

## Notable Implementation Details

- **Privacy-first design**: IFC bytes stream directly from provider to browser; only OAuth token exchange touches ifclite servers
- **Secure credential handling**: Refresh tokens stored in httpOnly, Secure, SameSite cookies; JavaScript never accesses them
[...]

Issue #1663, opened 2026-07-08
Dalux box integration


bruadam:
### The problem

Manually download and upload ifc files from Dalux Box. 

### Proposed solution

Enter API key, files to download from Dalux Box and ifc-lite fetches and recreate the federated model directly from Dalux Box.

I suggest using Dalux Build API and storing the projectId, fileAreaId, fileIds as the anthropic api key at the moment. We can add a Dalux Build Api key in a setting modal and select the files.

EloyRD:
I see your point. How to handle identifying which IFC files are linkted to with Dalux Site? I haven't found a way in Dalux Build API to query that.

bruadam:
Dalux Build is a bit tricky...

There is files endpoints and folders endpoints. 

Files are related to folder using a folderId but you can't get only the files in a specific folder. You need to query all files (paginated endpoint) and then map them to folders. 

The steps are the following:

- Query /projects
- Query all the files with bookmarks.
- Query all the folders 
- Map folders hierarchy and map files to folders.
- Download specific files using their downloadLinks.

So the user should:
- Choose a project in project list 
- Choose folders and files to download

The nice to have function:
- save metadata for the chosen files and add a sync function to pull from Dalux any updated version of the file. 

I hope that answers your question.

bruadam:
@louistrue I am available for a call if you want to finally implement that. I have started it yesterday but I would prefer to know how you would see the integrations in ifc-lite. 

I think I would like to look into Trimble Connect, SharePoint and Autodesk Forma also. Maybe also Revizto. 

The goal is to make an integration  folder similar to plugins in Raycast - so it is scalable to add more integrations.

louistrue:
@bruadam sorry currently moving and difficult to meet next few days...  Love it! 🙌 go for it, see also https://github.com/LTplus-AG/ifc-lite/pull/918 (from a while back, never got around finishing it...). It would be lovely if community could easily contribute integrations... 👍

louistrue:
@bruadam picking this up on our side now. You mentioned you already started implementing the Dalux Build API client - could you push your branch (or open a draft PR), so we build on your work instead of duplicating it? Interface-wise the current plan is a provider-style integration (fetch selected file IDs from a Dalux Box file area into the federated viewer, API key stored locally like the Anthropic key). If your WIP already covers the API client, we would wire the viewer side around it and credit you as co-author.

bruadam:
@louistrue no problem for me. 

[...]

PR #1971, opened 2026-08-02
feat(cloud): import IFC from Dropbox, Google Drive and OneDrive (continues #918)

BIMvoice:
## What and why

Continues and supersedes #918 (@louistrue's draft). His five commits are preserved as ancestors — merge commit, no rebase, no squash, no force-push. Closes #918 when merged.

The draft was 731 commits behind `main`. It merged with **3 conflict hunks, all in `MainToolbar.tsx`**, which was refactored 1870 → 1238 lines in the meantime.

## Google Drive: Picker + `drive.file`, no server

Per review, Drive is now browser-only and the server-side flow for it is **removed**: `server/google/google.ts`, `api/google/*` (5 routes), `services/cloud/google-drive.ts`, and the Google cases in `tests/api/cloud-oauth.test.ts`. `server/cloud-oauth/*` stays — Dropbox and OneDrive still need the server-side code exchange, unchanged.

**Why this matters:** the draft requested `drive.readonly`, one of Google's **restricted** scopes, requiring restricted-scope verification *and* a CASA security assessment before publication — a recurring cost, precisely because the design held a refresh token server-side. `drive.file` is **non-sensitive** and requires neither. Combined with the Picker, the user chooses which files the app may touch, and the app can read exactly those; the Picker *is* the browse UI, so nothing is lost for "open an IFC from my Drive".

The cost, stated plainly: no refresh token, so each session starts disconnected and there's no server-side folder browsing.

## The rest

- **`ingestExternalFile()`** (new, in `toolbar/useFileCommands.tsx`) — cloud imports now run the *same* primitives as the drop handler: `splitDxfFiles`, `isSupportedModelFile`/`describeUnsupportedFormat`, `recordRecentFiles`/`cacheFileBlobs`, then `addSupportedFiles` vs `loadFile`. So DXF underlays, unsupported-format messaging and federation behave identically whether a file is dropped or imported. Replaces the draft's ad-hoc `addModel`/`recordRecentFiles`, neither of which survives on current main.
- **OneDrive on a personal account.** `fetchFollowedSites()` returned `[]` for *every* failure, so a consumer Microsoft account — which has no SharePoint tenant, so Graph 400/404s — was indistinguishable from an outage, and the UI showed nothing at all. Now 400/404 is separated from 5xx/network, and the OneDrive root shows an inert row ("SharePoint sites need a work or school account"). The SharePoint code is untouched and still works for tenant accounts. Four mocked-fetch tests.
- **CI gap.** `tests/api/*` passed but **no workflow ever invoked `pnpm test:api`** — true of the pre-existing `chat.test.ts` too. Wired into `test.yml`'s `node-tests` job.

## Two consequences worth knowing before merge

**Build-time credentials.** `VITE_*` values are baked into the bundle at build time. They are not secrets — this flow has no client secret — but the hosted deployment must set `VITE_GOOGLE_CLIENT_ID` / `VITE_GOOGLE_API_KEY` / `VITE_GOOGLE_APP_ID` in its build environment, and a self-hosted deployment must rebuild with its own. Without them Drive stays visible but every action rejects with a message naming the two required vars — chosen over hiding the entry, since a silently missing menu item is indistinguishable from a bug, and it matches the existing `501 <provider>_not_configured` behaviour.

**Not exercised against a live Google session.** Everything downstream of a valid token is unit-tested — download, 401, other failures, zero-byte, cancel, unsupported-file rejection — behind fake auth/picker clients. The actual OAuth popup and Picker rendering need a real account and a configured Cloud project. Being tested manually now; I'll report back. Flagging rather than implying end-to-end coverage.

## One rough edge left deliberately

The picked file still appears as a single row in `CloudImportDialog` which the user clicks to download. Routing it straight to `ingestExternalFile` would mean either duplicating the dialog's connect/loading/error handling in a Google-only bypass, or widening the shared `CloudProvider` interface with a `quickPick` capability — real surface area and test burden to save one click. Reasoning recorded next to the code; happy to do it if you'd rather.

## Verification

- Viewer suite **2191 tests / 2189 passed / 0 failed / 2 skipped**
- `pnpm typecheck` 33/33 · `docs:check-samples` clean · `check-test-wiring` clean
- `pnpm test:api` **33/33** (was 42; 9 Google-specific tests removed with the flow)
- `grep -rn "drive.readonly"` — no scope usage anywhere; the guide names it only to explain why it is avoided
- Credential scan across the diff — clean; `.env.example` carries placeholders and `.env.local` is gitignored

## Checklist

- [x] Maintainer's commits preserved (no rebase/squash/force-push)
- [x] Tests assert real behaviour
- [x] No `as any` / `@ts-ignore` / `as unknown as` — the `window.google`/`gapi` surface is typed in a `.d.ts`
- [x] No credentials committed

## Summary by CodeRabbit

* **New Features**
  * Added cloud file imports from Dropbox, OneDrive/SharePoint, and Google Drive.
  * Browse folders, select supported IFC files, download with progress, and import them into the viewer.
  * Added cloud-provider connection, disconnection, authentication, and token handling.
  * Added helpful handling for unavailable configurations and provider errors.

* **Documentation**
  * Added setup and usage guidance for cloud imports.

* **Tests**
  * Expanded coverage for cloud providers, OAuth flows, downloads, and API routes.


BIMvoice:
All three Greptile findings and CodeRabbit's tier-1 set are fixed in `e6bcd2a5`. Rather than list 25 items, the four that mattered:

**1. Transient refresh failures were disconnecting accounts.** The catch cleared the refresh cookie on *any* throw — a provider 5xx, a 429, a network blip all destroyed a valid token. The information to tell them apart was already there and thrown away: `postToken` throws `OAuthTokenError(res.status)`, and a network failure throws something with no status at all. Now only 400/401 clears it; everything else returns 503 and keeps the cookie, and the client says "try again in a moment" rather than implying you have been signed out.

**2. `token` and `disconnect` accepted any method.** Greptile caught the disconnect route; CodeRabbit caught the worse sibling. With `SameSite=Lax`, a top-level GET navigation sends the refresh cookie, so `/api/<id>/token` would **mint and return an access token** into that navigation. Both now require POST and reject a mismatched `Origin`.

Being precise about the risk, because it is easy to overstate: this is forced logout and a token minted into a navigation response — **not** exfiltration, since the attacker page cannot read a cross-origin response body. And the *state* cookie deliberately stays `SameSite=Lax`: tightening that one to `Strict` looks like hardening and silently breaks the provider redirect.

**3. OneDrive refresh-token rotation was dropped** — CodeRabbit's find, and the one I'd rank highest of the whole batch. Microsoft returns a **new** refresh token on every refresh; the handler read `tokens.refresh_token` and never stored it, so the cookie kept the original indefinitely. Now written back when it differs, and a natural no-op for Dropbox, which does not return one on refresh.

**4. Success was reported before ingestion.** Ours, and it needed a real change underneath rather than an `await`. `loadFile`/`addModel` never reject on a parse failure *by design* — a batch load must survive one bad file — so awaiting alone would still have reported success on a corrupt IFC. `ingestExternalFile` now checks the shared `error` field the loader sets, which is only sound because `loadFile` clears it at entry (`useIfcLoader.ts:377`); I verified that before relying on it, since reading a stale error would have failed every subsequent import.

Also fixed: script-loader promises cached a *rejection* forever, so one ad blocker permanently broke Drive for the page session — and the failed `<script>` tag was left in the DOM, so a retry falsely "succeeded" via the existing-tag fast path without loading anything. `goUp` called `load()` inside a `setTrail` updater, which StrictMode double-invoked into **opening the Picker twice**. Plus the stale-listing race, the 401 not clearing the cached token, an unhandled `disconnect` rejection, a 10s token timeout, `parseDriveRef` prefix validation, and clearing the OAuth popup's result key first (an identical value does not fire `storage`, so the promise never settled).

**Skipped, with reasons:** a null-prototype cookie map (only known keys are ever read), the duplicate edge-runtime exports in the Dropbox routes (your convention, not ours to churn mid-review), and extracting the shared folders-first comparator — real duplication, but cosmetic churn across two providers in an already-large PR. Happy to take that last one as a follow-up.

One testing note worth passing on: the `goUp` StrictMode bug **only** fails inside `<StrictMode>`. A plain render passes with the bug live, because the double-invoke is what surfaces it. Same shape as the `act()` batching trap from #1958 — the fixture has to be hostile to the property, or the test cannot see it.

Verification: viewer suite **2202 tests / 2200 passed / 0 failed / 2 skipped**, `pnpm test:api` **51/51** (was 33), `pnpm typecheck` 33/33, `docs:check-samples` clean, `check-test-wiring` clean. Mutation-tested: retryable-vs-fatal refresh, method/origin on both routes, script-loader promise clearing, and the StrictMode double-invoke.

Still not exercised against a live Google session — that is being tested manually and I will report back either way.

louistrue:
@BIMvoice — closing this one, and I owe you a full explanation rather than a one-liner, because the work is good and the reason is architectural timing rather than quality.

**First, the part that matters most: your OAuth engineering came out of an adversarial security review clean.** I sent a reviewer at it specifically to break token storage, `state` handling, the popup bridge, SSRF and the callback XSS. It failed on all five:

- No token ever reaches `localStorage`/`sessionStorage` or a URL — access tokens live in a private class field, refresh tokens stay server-side in an httpOnly, path-scoped cookie
- `state` is CSPRNG, host-only httpOnly cookie, compared on callback, and cleared on **every** branch including the failure ones. I could not construct a fixation or a replay
- No server route fetches a user-supplied URL. This repo had exactly that bug class before ("follow an upstream redirect off-host with the key still attached") and you did not repeat it
- `sanitizeMessage` whitelists before escaping, so `</script>` and ` ` breakouts both reduce to inert text
- No client secret in the bundle; `/api/*` deliberately has no `Access-Control-Allow-Origin`, so the missing method check is hardening rather than a live token-theft path

That is a better result than most of this repo's own API surface would get.

**What happened.** While this PR was open, an unmerged 127-file branch of mine carried a *source plugin architecture* — manifest-declared providers, semver-gated against a host `PLUGIN_API_VERSION`, a `permissions.network` allowlist, sandboxed `host-fetch`/`host-storage`, and manifest-declared relays. That branch was invisible to you, which is on me, not you. It has now landed as **#1976** (contract + host + UI), with `packages/source-dalux` and `packages/source-msgraph` to follow.

The contract question was decided by a use case neither of us is arguing about: **Dalux (#1663) serves no CORS headers.** It needs a same-origin relay, declared and validated. A flat `connect/list/download` interface cannot express that, and every provider needing a relay ends up with bespoke serverless code anyway — the same path, without the guardrails.

**Three findings, so this isn't only about architecture:**

1. **It ships dark.** `CloudImportDialog` is referenced only from `MainToolbar.tsx`, the *classic* toolbar. `constants.ts:338` — "the ribbon is the default toolbar… every new browser lands on the ribbon." Nothing in `ribbon/` or the command palette reaches it. On ifclite.com today, a clean browser would never see this feature.
2. **Google Drive is very likely dead under our own headers.** We send `COOP: same-origin` + `COEP: credentialless` app-wide. GIS `initTokenClient` hands its token back over `window.opener`, which COOP severs — the exact reason your Dropbox/OneDrive paths use the localStorage bridge — and the Picker is a cross-origin iframe that COEP blocks. Every Google test injects fakes, so the suite passes with the real flow broken.
3. **OneDrive is settled, and I verified it rather than assuming.** MSAL v5 routes through a same-origin bridge page using `BroadcastChannel`, not `window.opener` — I grepped `@azure/msal-browser@5.17.1`'s `redirect_bridge`: zero references to `opener`, and it posts on `new BroadcastChannel(id)`. I also brought the dev server up and confirmed `/` serves `COOP: same-origin` while `/msal-redirect.html` serves `unsafe-none`. So MSAL survives where GIS does not, by design. Your server-relay OneDrive isn't needed — that's a mechanism result, not a preference.

**What I'd like, if you're up for it.** Two concrete things, either or both:

- **`packages/source-dropbox`, as the first external plugin on the v2 contract.** Your OAuth core and `api/dropbox/*` routes carry over close to unchanged — a v2 `auth: 'interactive'` provider with a server-side token custodian is exactly the shape you already built. Two things worth bringing along regardless of contract: the method/Origin/size/rate-limit floor that `api/chat.ts` already sets on main, and tests for `oauth-provider-base` and the sanitizers (right now, deleting all three sanitizers leaves every test passing — a live XSS the suite is blind to, guarding a defence that currently works).
- **The open threads on #1976.** It's green but has ~16 unresolved review items — two "split the oversized module" (`SourceBrowser.tsx`, `useSourceCatalogSync.ts`), a couple of behavioural questions on revision-watch semantics, and some missing tests around `syncSourceModel`. If you'd rather work on the host than a provider, that's yours for the taking.

No obligation on either. But you've merged six PRs here in two days and your last three reviews found things I'd missed, so I'd rather have you shaping this than not.

Sorry this one cost you 2,971 lines to land somewhere other than where you aimed. The transport layer was the hard part and it survived.

BIMvoice:
No apology needed for the close — you're right on the architecture, and the Dalux/CORS argument settles it: a flat `connect/list/download` interface can't express "this provider needs a validated same-origin relay", and pretending otherwise just means every such provider grows bespoke serverless code without the guardrails. Better to find that out at 2,971 lines than at 10,000.

**Your finding #2 is the one I most needed, and I verified it rather than taking it.** `vercel.json:34-39` and `apps/viewer/vite.config.ts:285-289` both set `COOP: same-origin` + `COEP: credentialless`. GIS `initTokenClient` returns its token over `window.opener`, which COOP severs; the Picker is a cross-origin iframe, which COEP blocks. So the browser-only Drive path cannot work under our own headers — and I had *just* written setup instructions telling someone to go try it.

The part that stings is the second half of your sentence: *"Every Google test injects fakes, so the suite passes with the real flow broken."* I wrote "not exercised against a live Google session" as a caveat in the PR body and thought that discharged the obligation. It doesn't. A caveat says *I haven't checked*; it doesn't say *this cannot work*, and I had everything needed to know the difference — the headers are in the repo, and the reason my Dropbox/OneDrive paths use a localStorage bridge instead of `window.opener` is the same COOP behaviour, sitting in code I'd read. I documented the gap instead of closing it.

Worse, my setup doc says to register `http://localhost:5173` as the authorized origin. The dev server is on **port 3000** (`vite.config.ts:284`). So the instructions were wrong twice over, and the second one I could have caught with a single `grep`.

**Finding #1 is a straight miss on my part.** `CloudImportDialog` reachable only from `MainToolbar.tsx` while `constants.ts:338` says the ribbon is the default — a clean browser never sees it. I traced the ingest path carefully enough to route cloud files through the same `splitDxfFiles`/`isSupportedModelFile` primitives as drag-and-drop, and never asked the simpler question of whether anyone could click the thing. Entry-point reachability isn't something I currently check at all; it is now.

**Finding #3** — thank you for grepping `@azure/msal-browser@5.17.1` for `opener` and bringing the dev server up rather than reasoning about it. "MSAL survives where GIS does not, by mechanism" is exactly the kind of claim that's worth the ten minutes to establish, and it means the OneDrive work isn't wasted.

**On the sanitizers:** you're right and it's the sharpest of the lot — deleting all three leaves every test green, guarding a defence that currently works. That's the same shape as the untested `popupRef.contains` guard you caught on #1958, which is now twice I've shipped a working defence with nothing pinning it. I've started running line-level mutation coverage over diffs for exactly this (delete each changed line, see whether any test notices); it found a genuinely unpinned two-digit-year pivot in the schedule importer. It would have found these sanitizers too.

**I'll take the #1976 threads.** The module splits and the `syncSourceModel` test gaps are concrete, and getting the host landed helps more right now than a provider built against a contract that's still moving. Once it's in, `packages/source-dropbox` on the v2 contract is the natural follow-up — and I'll bring the method/Origin/size/rate-limit floor from `api/chat.ts` plus real tests for `oauth-provider-base` and the sanitizers, since those carry over regardless of which contract wins.

One question before I start on #1976: for the two oversized modules, do you want a mechanical split along existing seams, or is there a decomposition you already have in mind? I'd rather match your intent than impose a structure you'll want to redo.

Issue #961, opened 2026-06-06
Render IfcBlobTexture / IfcSurfaceStyleWithTextures on tessellated geometry (follow-up to #957)


louistrue:
Follow-up to #957. The visibility fix (PR for #957) makes type-only tessellated geometry render, but **untextured** — the boiler from the buildingSMART annex-E `tessellation-with-blob-texture.ifc` shows flat white instead of its PNG texture.

Full texture rendering was deliberately deferred because it requires building a **new textured render path in the WebGPU renderer**, which has no texture infrastructure today (no UV vertex lane, no sampler/texture bindings in the WGSL; colour is a per-batch uniform). It also can't be browser-verified in a headless agent environment.

### Scope (staged plan)
1. **Decode blob** — `IfcBlobTexture.RasterCode` is a STEP binary literal (`"…hex…"`, leading char = bit padding). Decode hex → PNG bytes; keep `RasterFormat` ('PNG') as the MIME hint. (`IfcImageTexture` = URL, `IfcPixelTexture` = raw pixels — handle or explicitly skip.)
2. **UVs in the kernel** — `IfcIndexedTriangleTextureMap` (`TexCoordIndex`, parallel to `CoordIndex`) + `IfcTextureVertexList` → per-vertex UVs; apply `IfcCartesianTransformationOperator2D`. Add a `uvs` channel to `Mesh`.
3. **Transport** — add `uvs` + `texture {imageBytes, format, transform2d}` to processing `MeshData`, `MeshDataJs` (zero-copy getters; note the Y/Z flip must not touch UVs), and the TS `MeshData` (all three declarations).
4. **Renderer** — a non-batched textured path: textured meshes skip colour-bucketing and become individual draws with a UV `BufferAttribute` + a `DataTexture`/`CanvasTexture` decoded from the PNG; new WGSL pipeline with a `texture_2d` + `sampler` binding. Integrate with picking/selection/section planes.
5. **Viewer ingestion** — route textured meshes to that path.

### Minimal viable subset
`IfcBlobTexture` (PNG) + `IfcIndexedTriangleTextureMap` + `IfcTextureVertexList` + the 2D scale/offset from `IfcCartesianTransformationOperator2D`. Defer `IfcPixelTexture`, `IfcImageTexture` (URL), `IfcTextureCoordinateGenerator`, `IfcIndexedPolygonalTextureMap`.

[...]

PR #1974, opened 2026-08-02
fix(renderer): apply the per-element origin to textured meshes (#1973)

### LINKED ISSUE #1973: Textured face sets render collapsed toward the world origin (per-element `origin` not applied)
### Version

v1.33.1

### Which surface

Viewer (browser)

### Reproduction

1. Open `tests/models/textures/ifc-hbim-convento-textures.ifczip` in the WebGPU viewer.
2. IFC4 SketchUp/BIM-Tools export, no `IfcBuildingStorey` — elements sit directly in `IfcSite`/`IfcBuilding`, placed via a normal `IfcLocalPlacement → IfcAxis2Placement3D` chain carrying the real elevation (roof id#506 ≈ 10.5 m, window id#248 ≈ 6.8 m).
3. Try to select a textured surface. Clicking the visible texture usually selects nothing — pick/raycast geometry is at the authored position, not where the texture draws (offset toward the origin, often below ground). Click where the element *should* be and it selects, offset from the visible texture.

### Expected vs actual

- **Expected:** textured surface renders at its authored elevation, coincident with its pick geometry.
- **Actual:** each textured face set is drawn offset by `−origin`, collapsed toward the world origin ("sunk to storey 0", often below ground). Picking uses the correctly placed geometry, so clicking the visible texture misses. Only textured meshes are affected; untextured geometry is fine.

## Root cause

Placement math is fine — `get_placement_transform_with_depth` resolves the chain correctly (replaying it by hand gives a ~14.35 m model). The issue is renderer-side: the textured layer drops the per-element `origin`.

For precision, `transform_mesh_world_framed` (wasm, `local_frame_enabled()`) stores vertices relative to a per-element `origin` (carrying the world placement incl. Z) and defers the world magnitude into `MeshData.origin`. Contract downstream: `world = origin + position`.

The batch path honours it — `mergeGeometry` folds `origin` into a shared frame (`stored = (origin − sharedFrameOrigin) + position`, `scene-geometry.ts:126`) and draws `model = translate(sharedFrameOrigin)`.

The textured path drops it:
- `interleaveTexturedVertices` (`scene.ts:3449`) writes raw `positions`, never adds `origin`.
- `TexturedMesh` (`scene.ts:63`) has no `origin` field.
- The textured sub-pass zeros the model translation (`index.ts:2410-2413`):
  ```js
  // Textured meshes carry absolute (origin-0) positions...
  tpl[28] = 0; tpl[29] = 0; tpl[30] = 0;
  ```
  → `world = 0 + position`, i.e. off by `−origin`.

The `origin-0` comment was right for the #961 orphan type-geometry path (`transform_mesh_local` → absolute positions, `origin == 0`). #1793 added the occurrence path (direct `Body` textured `IfcTriangulatedFaceSet` — what real exporters write) via `apply_submesh_placement` → `transform_mesh_world(relativize=true)`, which produces local-frame positions and a non-zero `origin`. The textured pass wasn't updated for it. `origin` crosses the wasm boundary fine (`zero_copy/mesh.rs:395`); it's only dropped renderer-side.

## Note
Textured meshes run their own pipeline (separate buffers/uniform/draw loop, own visibility/override/translate handling — `index.ts:2407`, `scene.ts:1467`/`1660`/`3464`), so batch-path invariants get restated there; `origin` folding is the one that drifted. Reusing the batch frame would keep `world = origin + position` in one place.

## Fix options
- **Small:** carry `origin` on `TexturedMesh` (from `meshData.origin`, already Y-up) and set `tpl[28..30] = tm.origin` instead of `0` at `index.ts:2413`. Covers both #1793 (`origin != 0`) and #961 (`origin == 0`, no-op).
- **Bigger:** put textured meshes on the batch frame — relativize in `interleaveTexturedVertices` as `position + (origin − sharedFrameOrigin)`, draw with `translate(sharedFrameOrigin)`. Keeps textured/opaque bit-coincident so the `depthCompare:'equal'` overlay trick holds. Not needed to fix the bug.

## Regression test
Assert the textured path reconstructs `world = origin + position` — e.g. a textured face set and its untextured twin land at the same world AABB. `packages/renderer` scene test like `scene-local-bounds.test.ts`.

### Environment

Not GPU/OS-specific — it's a CPU-side transform.

### Privacy

- [x] My attached files and text contain no confidential or client-identifying data.

louistrue:
Closes #1973.

The report's analysis is correct in full — I verified each claim against the code rather than taking it as given, and every one holds. Credit to @janbrouwer: the root cause, the two candidate fixes and the suggested regression test were all accurate.

## The bug

Local-frame meshes store vertices relative to a per-element `MeshData.origin` so the world magnitude stays out of f32 (#1114). The contract is `world = origin + position`.

| path | stored | model translation | world |
|---|---|---|---|
| batch (`mergeGeometry`) | `(origin − sharedOrigin) + position` | `translate(sharedOrigin)` | ✅ `origin + position` |
| textured (before) | `position` | `translate(0)` | ❌ `position` |

Off by exactly `−origin`. CPU picking uses the correctly placed geometry, so clicking a visible texture selected nothing.

The zeroing was right when written: the #961 orphan type-geometry path runs `transform_mesh_local`, leaving positions absolute with `origin == 0`. #1793 then added the occurrence path — a `Body` textured `IfcTriangulatedFaceSet`, which is what real exporters write — via `apply_submesh_placement` → `transform_mesh_world` → `transform_mesh_world_framed(relativize = local_frame_enabled())`, and `local_frame_enabled()` defaults **on for wasm**. So the viewer path produces a non-zero origin and this sub-pass was never updated.

This is the third instance of the same class (the void fast-path #1806/#1815 was the last): a world-space consumer that doesn't fold `MeshData.origin`.

## The fix

`TexturedMesh` carries `origin`; the draw sets the model translation from it per mesh. The loop already wrote `tpl` per mesh (per-mesh colour at `tpl[32..35]`), so this is a hoisted-constant becoming a per-mesh value — no restructuring.

**Interleaved positions deliberately stay local.** Folding the world magnitude back into the f32 vertex data is the obvious-looking alternative and it is wrong: it defeats the local frame this origin exists to provide and reintroduces the coordinate collapse on exactly the georeferenced models that need it. The test asserts the positions are still local.

I took the report's "small" option. The "bigger" one (put textured meshes on the batch frame) is a genuine improvement for keeping textured/opaque bit-coincident under `depthCompare:'equal'`, but it is not needed to fix this and is a larger change to the id/frame space.

Also checked, not assumed: the translate (`scene.ts:1489`) and rotate (`scene.ts:1682`) re-upload paths re-interleave from `meshData` and mutate only `positions` — rotation reads `meshData.origin` to fold the pivot but never writes it — so the cached `TexturedMesh.origin` stays valid across gizmo moves with no extra work. And leaving `tpl[28..30]` set no longer leaks: `renderBatch` always overwrites it from `batch.origin`, and the transparent passes overwrite the whole model matrix via `tpl.set(mesh.transform.m, 16)`.

## Verification on the reported model

`tests/models/textures/ifc-hbim-convento-textures.ifczip`, loaded in the WebGPU viewer. Read from the live store:

```
totalMeshes: 109,  textured: 107,  texturedWithNonZeroOrigin: 107
sample origins: [-19.03, 9.45, -12.79], [-19.22, 5.33, -14.84], [-19.13, 5.03, -14.11], …
```

Every textured mesh on this file was mis-drawn, by up to ~19 m.

Live A/B in the browser (the viewer aliases `@ifc-lite/renderer` to source, so reverting just the draw line gives a clean HMR comparison):

- **Before:** the model renders as a crushed heap of overlapping fragments collapsed around the origin.
- **After:** a coherent building — roof, courtyard and walls in their authored positions.

Screenshots are in the issue thread below.

## Tests

`packages/renderer/src/scene-textured-origin.test.ts` — the invariant the issue asks for, GPU-free via a capturing device stub:

1. `MeshData.origin` is recorded on the `TexturedMesh`.
2. A textured face set lands at the **same world AABB** as its untextured twin through `mergeGeometry` — and the interleaved positions are asserted still local, so a "bake the origin in" regression fails here too.
3. `origin` stays `[0,0,0]` for the absolute-position orphan path.

**Mutation-checked:** reverting `scene.ts` fails all three.

**Scope note, stated plainly:** these cover the scene-side bookkeeping and the reconstruction arithmetic. The one-line uniform write in `index.ts` is not unit-covered — the existing render-path harness stubs `writeBuffer` as a no-op and does not reach the textured sub-pass, and building a frame-level harness for one line was not worth it. That line is what the browser A/B above exercises directly. I also could not drive picking through synthetic pointer events in this harness, so I am *not* claiming to have verified the click-to-select half interactively; test 2 covers it by construction, since pick geometry is the batch-path world position the textured mesh now matches.

## Verification

- `packages/renderer`: 411 passed, 0 failed
- `pnpm typecheck`: 33/33
- Changeset included (`@ifc-lite/renderer` patch)

### DIFF
--- a/.changeset/textured-mesh-origin.md
+++ b/.changeset/textured-mesh-origin.md
@@ -1,0 +1,13 @@
+---
+'@ifc-lite/renderer': patch
+---
+
+Fix textured face sets rendering collapsed toward the world origin (#1973).
+
+`transform_mesh_world_framed` (on by default for wasm) stores each element's vertices relative to a per-element `MeshData.origin`, keeping the world magnitude out of f32 so building-scale coordinates can't collapse adjacent vertices into degenerate fans. The contract downstream is `world = origin + position`.
+
+The batch path honours it — `mergeGeometry` folds `origin` into the shared frame and draws with `translate(sharedFrameOrigin)`. The textured sub-pass hard-zeroed its model translation, so every textured mesh drew offset by `-origin`. On a typical textured export that is metres: on the reported model, 107 of 109 meshes are textured and every one of them has a non-zero origin, up to ~19 m. The whole model rendered as a crushed heap around the origin. CPU picking uses the correctly placed geometry, so clicking a visible texture selected nothing.
+
+The zeroing was correct when written: the #961 orphan type-geometry path runs `transform_mesh_local`, which leaves positions absolute and `origin` at zero. #1793 then added the occurrence path (a `Body` textured `IfcTriangulatedFaceSet`, which is what real exporters write) via `apply_submesh_placement` → `transform_mesh_world`, producing local-frame positions and a non-zero origin — and the textured pass was not updated for it.
+
+`TexturedMesh` now carries `origin`, applied as the draw's model translation. Interleaved vertex positions stay local: folding the world magnitude back into f32 vertex data would defeat the local frame this origin exists to provide. Both cases are covered — `origin == 0` for the orphan path is a no-op, `origin != 0` for occurrences is the fix.
--- a/packages/renderer/src/index.ts
+++ b/packages/renderer/src/index.ts
@@ -2407,10 +2407,6 @@ export class Renderer {
                 const texturedMeshes = this.scene.getTexturedMeshes();
                 if (texturedMeshes.length > 0) {
                     pass.setPipeline(this.pipeline.getTexturedPipeline());
-                    // Textured meshes carry absolute (origin-0) positions, so the
-                    // model translation must be identity here — reset the column
-                    // that renderBatch left set to the last opaque batch's origin.
-                    tpl[28] = 0; tpl[29] = 0; tpl[30] = 0;
                     for (const tm of texturedMeshes) {
                         // Honour hide/isolate — textured meshes bypass the batch
                         // visibility filtering above, so apply it per-mesh here or
@@ -2435,6 +2431,16 @@ export class Renderer {
                         // batch overlay paint pass doesn't iterate textured meshes,
                         // so applying the override here is what recolours them.
                         const txOverride = colorOverrides?.get(tm.expressId);
+                        // `world = origin + position`: the vertex buffer stores
+                        // positions in this mesh's per-element local frame, so the
+                        // model translation carries the world magnitude (#1973).
+                        // Per mesh, which also overwrites the column renderBatch
+                        // left set to the last opaque batch's origin. This used to
+                        // be hoisted out of the loop and hard-zeroed — right only
+                        // for the orphan type-geometry path (origin == 0), and it
+                        // drew every textured occurrence collapsed toward the
+                        // world origin.
+                        tpl[28] = tm.origin[0]; tpl[29] = tm.origin[1]; tpl[30] = tm.origin[2];
                         tpl[32] = txOverride ? txOverride[0] : tm.color[0];
                         tpl[33] = txOverride ? txOverride[1] : tm.color[1];
                         tpl[34] = txOverride ? txOverride[2] : tm.color[2];
--- a/packages/renderer/src/renderer-render-paths.test.ts
+++ b/packages/renderer/src/renderer-render-paths.test.ts
@@ -56,6 +56,8 @@ interface Harness {
         createdBuffers: FakeBuffer[];
         /** how many times a readback buffer was actually mapped */
         mapAsync: number;
+        /** every `queue.writeBuffer` payload, copied at call time */
+        writes: { buffer: unknown; floats: Float32Array }[];
     };
     knobs: {
         /** 'texture' = getCurrentTexture succeeds; 'null' = returns null */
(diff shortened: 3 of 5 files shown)


louistrue:
Real CodeRabbit run (the CI check is a rate-limited false pass) raised two; one taken, one skipped with reason.

**Taken — no render-path coverage for the uniform write.** Correct, and it was the gap I flagged in the PR body myself. Closed in ebc300f2 rather than left as a caveat: `renderer-render-paths.test.ts` already drives the real `render()` loop against a stub GPU, so it now captures `writeBuffer` payloads and asserts three things — the origin reaches `tpl[28..30]`, a mesh with no origin still gets zero, and **two textured meshes each get their own origin**. That last one is the property the old hoisted single write structurally could not have, so it is the case that pins the bug class rather than just the symptom.

The capture copies each payload eagerly; the renderer reuses one scratch array for every per-mesh uniform, so holding the reference would read back only the frame's last write and the multi-mesh assertion would pass vacuously.

Mutation-checked: reverting the one-line fix fails cases 1 and 3. Case 2 stays green, correctly — `origin == 0` is the no-op path where old and new behaviour agree.

**Skipped — "remove supplied-model metrics from the changeset".** The numbers are not from a supplied or client model. `tests/models/textures/ifc-hbim-convento-textures.ifczip` is a committed public fixture (`tests/models/manifest.json:772`, fetched by `pnpm fixtures`), so the mesh counts and origin magnitudes are reproducible by anyone with a checkout. They are also the load-bearing evidence for severity: "107 of 109 meshes textured, every one with a non-zero origin up to ~19 m" is what distinguishes this from a cosmetic sub-metre offset. Generic wording would drop that.

Renderer suite now 414 passed / 0 failed; `pnpm typecheck` 33/33.

PR #1987, opened 2026-08-02
fix(diff): widen stableHash to FNV-1a-64 (#1962)

### LINKED ISSUE #1962: Widen `stableHash` beyond 32-bit FNV-1a (it is used as an identity key)
## What

`stableHash` (`packages/diff/src/fingerprint.ts:47-55`) is a textbook 32-bit FNV-1a. It produces `dataHash` via `buildDataFingerprint` and every sub-hash in `buildComponentFingerprints`. 32 bits is too narrow for any use that treats hash equality as *identity*, and #1961 added exactly such a use.

This issue tracks widening it. Raised by Greptile on #1961 and deliberately kept off that PR: the change alters every published fingerprint and the hash is described in the frozen `node-hash-v0` spec (§1c), so it wants its own review and a stated migration path.

## Collisions are reachable, not theoretical

Found by enumerating `buildDataFingerprint`; all three are now fixtures in `packages/diff/src/content-match.test.ts`:

| entity A | entity B | shared `dataHash` |
| --- | --- | --- |
| `IfcWall` / `W-168484` | `IfcDoor` / `D-35800` | `b4b48be` |
| `IfcWall` / `Wall`, `Pset_WallCommon.Reference = R-1129599` | same, `R-1732382` | `36c92241` |
| `IfcWall` / `W-129599` | `IfcWall` / `W-732382` | `bc5b589a` |

One caveat on interpreting these: my enumeration hit its first collision well past the uniform birthday bound (~77k distinct values), so FNV-1a over these highly structured strings is evidently not behaving uniformly. **I have not measured a collision rate for real IFC content and no one should quote one from this issue.** The claim here is only that collisions are reachable and that exposure grows with the square of the number of distinct fingerprints compared.

## Why component sub-hashes are not a substitute

This is the part that surprised me, and it constrains the design of any partial mitigation.

FNV-1a's per-character update is `h = (h ^ c) * prime`. Multiplication by an odd prime mod 2^32 is a bijection, and so is XOR, so the whole update is a bijection on the state. Two entities differing only in `Name` hash payloads of the form `prefix + name + identical suffix`, and appending an equal suffix preserves state equality. Therefore, for a difference confined to `attr:core`, a `dataHash` collision **implies** an `attr:core` collision.

Confirmed on the third pair above: both sides produce `attr:core = 871c2784`.

So no amount of sub-hash cross-checking closes the gap for that class. Only a wider hash does. #1961 ships the guards that *are* sound (bucket by `ifcType` + `dataHash`; require `components` agreement when both sides carry them) and pins the residue in a test named `KNOWN LIMIT: a collision confined to attr:core is not detectable and still pairs`.

## Options

1. **FNV-1a-64.** Smallest diff — same construction, 64-bit offset basis/prime, `BigInt` in JS. `rust/processing/src/determinism.rs` already uses FNV-1a64, so the repo has precedent and constants to match.
2. **xxHash / a 128-bit construction.** Stronger and faster, but adds a dependency or a hand-rolled implementation to a package that currently has neither.
3. **blake3**, matching `packages/ifcx/src/canonical.ts`, which already depends on `@noble/hashes/blake3` for layer ids. Most consistent with the newest code in the repo; heaviest per call.

`node-hash-v0` §3.1 already picked FNV-1a64 for its `geometry-mesh` leaf and §6 discusses why FNV is acceptable for non-adversarial drift-catching but not for verifier-facing claims, which is worth re-reading before choosing.

## The migration question I could not answer

**Is any `stableHash` output persisted across sessions, or is every one recomputed in-process?** If everything is recomputed, widening is nearly free at runtime — base and head are always fingerprinted by the same build, so a width change cannot produce a spurious diff. If any are persisted, they all invalidate on upgrade and need a version tag.

What I traced: `apps/viewer/src/lib/compare/buildFingerprints.ts` computes them per compare run; `packages/merge/src/three-way.ts` caches them in a `WeakMap` within a single call. What I did **not** verify: whether any conflict record carrying `ComponentSnapshot.hash` (`packages/merge/src/types.ts:71-72`) is serialized into CRDT state, a BCF payload, or anything else that outlives the process. That is the load-bearing question for the migration path and someone who knows the collab persistence model should answer it before this starts.

## Scope note

Not every current use is a defect. `snapshotOf` in `packages/merge/src/component-state.ts:334` compares two already-identified components of the same entity — a single comparison at 2^-32, not an all-pairs birthday problem. The identity-style use introduced in #1961 is the one that motivated this.

louistrue:
## What and why

`stableHash` (`packages/diff/src/fingerprint.ts`) was a textbook 32-bit FNV-1a. It produces `dataHash` and every sub-hash in `buildComponentFingerprints`. #1961 introduced a use that treats hash equality as **identity**: the content-matching pass buckets unpaired entities by `(ifcType, dataHash)` and its 1:1 branch retires a real `added` and a real `deleted`. 32 bits is too narrow for that, and exposure grows with the square of the number of fingerprints compared, with a from-scratch re-export leaving the whole model unpaired.

Now FNV-1a-64, matching `rust/processing/src/determinism.rs`'s offset basis and prime exactly. Same construction, same signature, 16 hex chars out.

## The migration question #1962 could not answer, answered

**Nothing is persisted.** `packages/merge/src/component-state.ts:334` is the only importer outside this package; every `ComponentSnapshot` is a per-call computation, the only route out of the process is a transient 409 body (`layer-registry-route.ts:598`), and the durable registry records carry no hashes. Layer content-addressing is blake3, not this. So no version tag and no invalidation are needed: base and head are always fingerprinted by the same build.

Two unrelated private 32-bit FNV copies were deliberately left alone, notably `packages/bcf/src/writer.ts:95` — that one **is** persisted inside BCF zips, which is exactly why it must not move.

## The frozen spec

`docs/vision/spec/node-hash-v0.md` §1c described `stableHash` as 32-bit. That is a **descriptive survey of pre-existing systems**, not the frozen wire format: the freeze covers §3, which uses only fnv1a64 and SHA-256 and never `stableHash`, and §0 line 75 explicitly scopes changing these systems as out of scope for v0. Added a dated post-freeze correction note and refreshed the stale line references. No §3 byte encoding is touched.

## Tests

The three real 32-bit collisions this package pinned no longer collide. The **guards** keep their coverage: those tests now set a deliberately shared `dataHash` on hand-built fingerprints, which is what they were always really about. `KNOWN LIMIT: a collision confined to attr:core is not detectable and still pairs` is kept, because widening makes it less likely, not gone: FNV-1a's per-character update is a bijection at any width. Deleted only `the fixtures really do collide`, whose sole job was guarding fixtures against a hash change.

## Verification

`packages/diff` 58/58 · `pnpm typecheck` 33/33 · `pnpm test` 74/74 · `docs:check-samples` clean · `check:api-surface` unchanged (no export moved).

Nine mutations run, each confirmed to red the intended tests then restored — including reverting the hash body to 32-bit (kills 4), and perturbing the prime and the offset basis independently.

Closes #1962.

## Summary by CodeRabbit

* **Improvements**
  * Upgraded fingerprint hashing from 32-bit to 64-bit, reducing collision likelihood.
  * Standardized hashes as fixed-width, lowercase hexadecimal values.
  * Type fingerprints now remain stable when assigned type `GlobalId` values change, while still detecting type name or IFC class changes.
  * Existing same-build comparisons and merges remain unaffected.
  * Persisted fingerprints must be recomputed after upgrading.
  * Duplicate type names with the same IFC class may produce matching fingerprints.

* **Documentation**
  * Updated fingerprint and model-diff guidance, including remaining collision limitations.

* **Tests**
  * Added coverage for hash formatting, known values, re-export matching, and collision handling.


louistrue:
Ran the CodeRabbit CLI against this branch (the hosted check has been reporting `success` with the description **"Review rate limited"** all day, which is not a review). Two MAJOR findings. **One was a real bug and is fixed; the other I am declining.**

## Fixed: `localeCompare` made `dataHash` depend on the machine

This is a genuine correctness bug and a good catch. All four canonical sorts used `localeCompare` with no explicit locale, so they used the *runtime's* default collation. Collations really do disagree:

```
en-US  a,A,ä,B
sv-SE  a,A,B,ä      <-- ä after B
code-unit  A,B,a,ä
```

Every one of those sorts feeds a `JSON.stringify` that is then hashed. So the same model fingerprinted on a Swedish-locale host and a US-locale host produced **different `dataHash` values for identical content** — in a hash whose entire purpose is cross-machine identity. Base fingerprinted by the CLI on one machine, head by the viewer on another, and nothing pairs.

All four sorts now use a `compareCodeUnits` helper. Two tests added, and the pin is mutation-checked against reverting the helper to `localeCompare`:

```
--- MUTATION: revert to localeCompare ---
  × orders property sets by UTF-16 code unit, not by any collation
  Tests  1 failed | 66 passed (67)
```

The second test deliberately asserts the two orderings are *distinguishable*, so it cannot pass vacuously.

Pre-existing since #939, not introduced here — but this PR is precisely about hash correctness, and it already moves every fingerprint value, so it is the right place to fix it rather than a follow-up that would move them all again.

## Declined: renaming `globalId` to `GlobalId`

`TypeAssignmentInput.globalId` is camelCase because that is the convention across this codebase (`store.entities.getGlobalId`, `EntityFingerprint`, every adapter) and across TypeScript generally. Renaming it to match the IFC attribute's own casing would be a breaking change to a published package's public type, for every consumer, with no behavioural gain — and it would make this one field inconsistent with its neighbours rather than consistent with anything.

Marked as skipped rather than silently ignored.

67 tests in `@ifc-lite/diff` (was 65) · `@ifc-lite/merge` 65 pass · typecheck 33/33 · docs-samples 251 clean · api-surface unchanged.

louistrue:
Good catch, and it is a flaw in the test I added for your *previous* finding — so worth spelling out.

**My locale-independence test was itself locale-dependent.** It used `'B'` versus the a-umlaut, which distinguishes code-unit order from collation order under `en-US` but **not** under `sv-SE`, where collation happens to agree with code units for that pair. On a Swedish-locale runner a regression to `localeCompare` would have sailed through green.

Fixed two ways, because the second is the one that cannot rot:

1. **Better characters.** `'a'` versus `'B'`. Code units put `'B'` (0x42) first; every collation puts `'a'` first. They disagree everywhere, so there is no locale with a hole.
2. **The direct assertion you suggested.** A new test patches `String.prototype.localeCompare` to count calls, runs `buildDataFingerprint` over property sets, quantity sets and type assignments, restores it in a `finally`, and asserts the count is **zero**. That does not depend on characters, collations, or the runner's locale at all — it asserts the collation API is never reached.

Both kill a revert of `compareCodeUnits` to `localeCompare`:

```
--- MUTATION: revert to localeCompare ---
  × orders property sets by UTF-16 code unit, not by any collation
  × never calls localeCompare at all, whatever the runner locale is
  Tests  2 failed | 14 passed (16)
```

The reasoning is written into the test comment, since "the test for locale independence was locale dependent" is exactly the kind of thing that gets fixed once and then reintroduced by the next person who picks convenient characters.

68 tests in `@ifc-lite/diff` (was 67).

Issue #1891, opened 2026-07-26
Comparison of models


bruadam:
### The problem

Related to https://github.com/IfcOpenShell/IfcOpenShell/issues/8412.

GlobalIds are not always reliable for comparison between two models. It does not capture correctly the replace geometry (deleted and added at the same place with different GlobalIds), nor splitter or merged elements. 

In a construction/design phase, the models are updated constantly and GlobalIds fall short in comparison tasks. Sometimes, a model is re-exported from scratch with totally new GlobalIds but do not mean the project has been entirely changed.

### Proposed solution

A hash-based method that represents an element’s geometry, position, and properties. The hashes are then compared to generate a diff with relevant change types:

- Added
- Deleted
- Moved
- Property changed
- Split
- Merged
- Duplicated
- Deduplicated

This should be additional to the default method from IfcOpenShell IfcDiff but not replace it.

xyzbety:
In my opinion, comparison via hashing is not feasible. Once a GlobalId changes, the corresponding hash value will also change accordingly. Furthermore, hashes cannot trace the same component before and after modification. This scenario resembles the Ship of Theseus paradox, which means we have to rely solely on GlobalId for identification, and that is precisely what GlobalID is designed for.

bruadam:
Thank for your answer.

To be clear, I'm not suggesting hashing replace GlobalId for identification; GlobalIds exist for exactly that and remain the primary key. This is meant as a complementary method for a specific gap: models get re-exported during design/construction and come back with entirely new GlobalIds on elements that didn't substantively change. Pure GlobalId comparison reports that as delete+add, which is noise in practice.

On the hash claim specifically: it won't change "accordingly" if GlobalId isn't part of the hash input in the first place. A hash computed over ifc class, type, name, property sets (with ids stripped), materials, and spatial placement, explicitly excluding GlobalId and OwnerHistory, stays stable across a re-export, and that's the point: it lets you re-associate the same component across a GlobalId change instead of losing it.

[...]

PR #1988, opened 2026-08-02
feat(geometry): expose a per-entity world AABB from the geometry hash pass (#1891)

louistrue:
## What and why

The content matcher (#1961) reports `moved` whenever two content-identical entities have different geometry hashes. That is a lie in several of its causes: the hash (`rust/geometry/src/geom_hash.rs`) is a world-space quantized triangle multiset, so "hash differs" conflates moved, reshaped and re-tessellated. A truthful `moved` needs a placement, and split/merge detection will need extent.

This exposes a per-entity **world AABB** from the same hashing pass. `corner()` already reconstructs the full f64 world coordinate of every triangle corner, so the box is close to free: no second pass, no extra traversal.

Surfaced as `MeshCollection.geometryAabbValues` (`Float64Array`, 6 per id, in `geometryHashIds` order), behind the same `setComputeGeometryHashes` gate. Nothing is computed when the diff feature is off. **The consuming diff-engine change is #1989; this PR ships only the data.**

Two details are mutation-pinned because both are easy to get subtly wrong: bounds take the **unquantized** f64 world coordinate, and they include corners of triangles the hash skips as post-quantization degenerate (a degenerate triangle still contributes real extent).

## Volume: measured, and deliberately not shipped

The obvious companion is a divergence-theorem volume. I built a harness running the **real** production path at the hasher's own granularity over the fixture corpus (17,919 elements / 55,427 segments) before deciding:

```
segments closed + consistently wound : 47277 (85.3%)
elements where ALL segments closed   : 15400 (85.9%)
elements with negative summed volume                : 62
elements whose volume exceeds its own AABB volume   : 73

reference-point invariance of the divergence sum
  CLOSED segments  : mean rel change 1.9e-9, max 7.3e-6
  OPEN/non-manifold: p50 7.6e-9   p90 7.5e-1   p99 2.6e8
```

The reference-point column is decisive. For an open surface the divergence sum is not inaccurate, it is **arbitrary** — the boundary flux scales with distance to the reference point. Closed segments are invariant to 2e-9 as theory demands; open ones move by 75% at p90.

This is not a corpus quirk. `rust/geometry/src/mesh_orient.rs:244-246` deliberately leaves open and non-orientable components as authored (`// open / non-orientable — leave winding as authored`), and `rust/processing/src/element.rs:311` records that since #1311 material-layer wall slices are **OPEN bands** whose union is the watertight skin — so their per-slice sum is not the wall's volume.

The reason to refuse rather than ship-with-caveats: only 135 of 17,919 elements produce a *detectably* wrong number (negative, or exceeding their own AABB). The rest look like ordinary plausible volumes. A split/merge detector fed those would emit confident false claims, which is worse than not having the feature.

**To do it properly** needs (a) a per-element closedness verdict carried out of the mesher — `orient_mesh_outward` already computes `closed` per component and discards it, so this is cheap — and (b) a defined answer for the multi-segment sum, since 3,205 elements are multi-segment and summing overlapping submeshes over-counts. Both belong in their own PR. The rationale is written into `GeometryHasher::world_aabb`'s doc comment so a future reader does not re-derive it.

I did reuse rather than reinvent: `rust/geometry/src/kernel/signed_volume.rs` already exists. Note that `tests/cdt_volume_watertight_verify.rs`'s `roof_1112` case computes a "volume" of 179.98 for a surface it documents as having 37 legitimate open edges, used purely as a self-consistency baseline. That is the trap in miniature.

## No geometry output changed

The geometry hash is byte-identical before and after, pinned by a test whose literals were taken from the **pre-change** implementation (checked out `HEAD:geom_hash.rs`, printed the values, restored, confirmed the new code reproduces them).

## Module-size ratchet

`geom_hash.rs` sat at exactly its recorded 410 budget, so it was **split, not allowlisted**: tests moved to `geom_hash_tests.rs` (ratchet-exempt) and the file is now 314 lines, so its allowlist row is **deleted**. Three rows grew, each with a written justification.

## Verification

`cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings` clean · `cargo test --workspace --no-fail-fast` exit 0, 138 test binaries, zero failures · `module_size_ratchet` 4/4.

`packages/wasm/pkg/ifc-lite.d.ts` regenerated via `scripts/build-wasm.sh` and committed (additive: `readonly geometryAabbValues: Float64Array`); the `pkg/README.md` / `pkg/package.json` churn was reverted.

15 mutations run, each confirmed to red the intended test then restored — including storing quantized indices instead of world coordinates, accumulating after the degenerate `continue`, and subtracting RTC back out.

Toward #1891.

### DIFF
--- a/.changeset/geometry-world-aabb-wasm-surface.md
+++ b/.changeset/geometry-world-aabb-wasm-surface.md
@@ -1,0 +1,15 @@
+---
+'@ifc-lite/wasm': minor
+---
+
+`MeshCollection.geometryAabbValues`: per-entity world bounding boxes from the geometry-hash pass.
+
+A new read-only member on the committed type surface (`packages/wasm/pkg/ifc-lite.d.ts`), so this is additive public API. Nothing was removed or renamed.
+
+Six `f64` per entry, `[minX, minY, minZ, maxX, maxY, maxZ]`, in `geometryHashIds` order — entry `i` occupies `[6*i, 6*i+6)`, so the array is always exactly `6 * geometryHashCount` long. An entity with a hash but no box reserves its six slots as `NaN` rather than shortening the array, which would mis-attribute every later entry. Populated only when `IfcAPI.setComputeGeometryHashes()` is on, the same switch that gates the hashes; empty otherwise, so nothing is computed when the diff feature is off.
+
+The box is in the WebGL Y-up frame, like every other box, position, origin and placement crossing this boundary. It carries absolute world coordinates (the file's RTC folded back in) while `positions` are RTC-relative, so a consumer comparing the two folds `rtcOffset*` in. See `docs/api/wasm.md`.
+
+Why: a changed geometry hash conflates *moved*, *reshaped* and *re-tessellated* into one bit, which is what makes the diff engine's `moved` match kind a guess. The box separates them — same extent at a new centre is a move, a different extent is a reshape, an identical box with a different hash is retriangulation.
+
+No companion volume is exposed. A divergence-theorem volume needs a closed, consistently wound surface, and 14.7% of the mesh segments reaching this pass are open or non-manifold, where that sum is arbitrary rather than approximate — and looks like an ordinary positive number from the outside.
--- a/docs/api/wasm.md
+++ b/docs/api/wasm.md
@@ -268,8 +268,41 @@ class MeshCollection {
   readonly geometryHashIds: Uint32Array;
   readonly geometryHashValues: BigUint64Array;
   readonly geometryHashCount: number;
+  readonly geometryAabbValues: Float64Array;  // 6 per id, see below
 }
 ```
+
+#### geometryAabbValues
+
+Per-entity world bounding boxes from the same pass as the hashes, gated by the
+same `setComputeGeometryHashes()` switch — empty when geometry hashing is off.
+
+Six `f64` per entry, `[minX, minY, minZ, maxX, maxY, maxZ]`, in
+`geometryHashIds` order: entry `i` occupies `[6*i, 6*i+6)`, so
+`geometryAabbValues.length === 6 * geometryHashCount` always. An entity that
+produced a hash but no box reserves its six slots as `NaN` rather than
+shortening the array, which would mis-attribute every later entry.
+
+The box is in the **WebGL Y-up frame**, like `positions`, `origin` and
+`localBounds` — not the IFC Z-up frame the hasher accumulates in. It holds
+absolute world coordinates while `positions` are RTC-relative, so fold the
+collection's RTC offset in (itself Y-up-swapped) before comparing the two:
+
+```js
+const world = [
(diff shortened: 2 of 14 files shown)


louistrue:
All four findings fixed in `9beca3fb`. Thanks — the first one was a real bug and would have surfaced much later as "the viewer highlights are offset".

**1. Coordinate frame.** Correct, and the fix goes slightly further than the report: rather than adding a second conversion, `swap_zup_to_yup_aabb` is now generic over the scalar type, so the `f32` `local_bounds` and the `f64` world box share **one** copy of the min/max reversal and cannot drift apart. No downcast, so the f64 box keeps full precision at world coordinates. Both helpers moved to a new `frame_swap.rs` so the Z-up→Y-up conversion lives in one place.

Audited every other surfacing point: the getter is the only place this data reaches JS (both the flat and partitioned paths funnel through `push_geometry_hash` into the same `MeshCollection`), and all four constructors plus `Clone` carry the field. Nothing else was missing the swap.

**2. Changeset.** Added, `@ifc-lite/wasm: minor`. `scripts/api-surface.json` needs no update — it snapshots top-level exports, not class members.

**3. Docs.** `docs/api/wasm.md` now documents `geometryAabbValues`: six-value layout, the `6 * geometryHashCount` invariant, NaN-reservation, ordering against `geometryHashIds`, gating, the Y-up frame, and how to fold in `rtcOffset*` (which is reported Z-up).

**4. Boundary test.** This is the one that mattered, so it was verified rather than assumed. `scripts/test-wasm-contract.mjs` now asserts through the real `buildPrePassOnce`/`processGeometryBatch` path that the boxes are in the viewer frame and tightly enclose the indexed vertices of their own element's meshes. Confirmed it actually catches finding 1 by reverting the fix, rebuilding wasm, and re-running:

```
  ✅ geometryAabbValues is a Float64Array of exactly 6 values per hashed id
  ❌ geometryAabbValues is in the viewer frame and encloses its own meshes
     id 71 axis 1: exposed min 7.2136 must not cut into the mesh min 1.2192
  ✅ geometryAabbValues is empty when setComputeGeometryHashes is off
```

Exactly one test reds, and it is the frame test.

**Also, forced by the ratchet:** `mesh.rs` grew past its budget, so it was **split, not allowlisted** — now 770 lines with its allowlist row *lowered* 792 → 770.

Gates: clippy `-D warnings` exit 0 · `cargo test --workspace --no-fail-fast` exit 0, 138 binaries, zero failures · `build-wasm.sh` + `test-wasm-contract.mjs` exit 0, 39/39 · `docs:check-samples` exit 0.

One correction to note for anyone reading the agent trail on this branch: an intermediate report claimed `docs:check-samples` is not wired into CI. That is wrong — `.github/workflows/test.yml:357` and `:583` invoke `scripts/docs/check-doc-samples.mjs` directly, so a grep for the pnpm script name misses it. The transient failure seen there was a worktree with incomplete per-package deps; after a clean install the gate passes 251/251.

Issue #1923, opened 2026-07-31
Lens rule: Classification System is not filtering


MennoMekes:
I choose to make a lense to filter the system 'NL-SfB tabel 1'

<img width="495" height="442" alt="Image" src="https://github.com/user-attachments/assets/d9aabf70-446a-4e83-be83-5d6b8fa6f365" />

It shows:
(1) NL-SfB tabel 1
(2) CCI constrcution
(3) NL-SfB Tabel 1

<img width="428" height="290" alt="Image" src="https://github.com/user-attachments/assets/244a7668-6854-4574-b175-f8300fd29cf5" />

The other lenses where I cna choose things work. (Attribute and Property)
Thanks!

Issue #1197, opened 2026-06-18
Wall reported as moved when its position never changed


BIMvoice:
When comparing two revisions of the same model, one wall is reported as moved by 1.09 m. That wall was never touched between the two versions.

Steps to reproduce:

Open two revisions of the same IFC model.
Open Panels > Compare Models and run the comparison.
Check the geometry changes list.

Expected: The wall is not listed, because it was not moved.
Actual: The wall is listed as moved 1.09 m. Verified by hiding each version separately, the wall sits in the exact same position in both. Only this one wall is affected, so it looks like a false positive in the geometry change detection.

Models compared uploaded.

[Archive 2.zip](https://github.com/user-attachments/files/29094203/Archive.2.zip)

PR #1989, opened 2026-08-02
feat(diff): tiered content matching so repeated components pair (#1891)

louistrue:
> **Stacked on #1987** (`feat/1962-widen-stablehash`). Review that one first; this branch contains its commit. Base moves to `main` once #1987 lands.

## The defect, verified

#1961 shipped content-keyed matching, but `contentBucketKey` is `ifcType + NUL + dataHash` — geometry is consulted *after* pairing and never used to separate candidates. So:

**Three data-identical doors at three different unmoved positions, all re-GUIDed, reported one `ambiguous` group and paired nothing** (`counts.added` stayed 3, `deleted` 3). One door on its own paired fine. Since a real model is mostly repeated components, the feature paired only the unique minority — the headline behaviour, not an edge case.

## The fix: refine inside the bucket, do not change the bucket key

Putting `geometryHash` into the outer key would **regress** the case that works today: a unique element that genuinely moved would land in a different bucket from its own base and never meet it, reverting every real move to add+delete noise. So the outer key is unchanged (it is also the load-bearing collision defence) and the tiers are local to each bucket:

- **Tier 1** — sub-bucket by geometry hash. `undefined` hashes go to the residue (undefined matching undefined is vacuous, not evidence). Equal non-zero counts on both sides retire as one `renamed` match. This alone fixes the three doors.
- **Tier 2** — 1:1 residue. `reshaped` if per-axis AABB size differs, `moved` if size agrees and the centre moved, `reshaped` at distance 0 if both agree but the hash differs (the re-tessellation case; the AABB genuinely cannot separate that from an interior reshape, and the comment says so). No AABB → today's bare `moved`.
- **Tier 3** — N:M residue via **iterated mutual-nearest-neighbour** under a distance cap. Not greedy (order-dependent, commits to bad chains) and not Hungarian (minimising *total* distance forces pairings at the cap edge; a global optimum is not evidence of identity). Mutual-NN is deterministic, order-independent, and **abstains on ties** — a symmetric layout of identical doors that all moved genuinely is ambiguous.

`componentsAgree` guards every retiring path. Tolerances default to `moveTolerance` 2e-3 / `reshapeTolerance` 1e-3, lifted deliberately from `describeChange.ts`'s `MOVE_EPS`/`RESHAPE_EPS` so the #1197 lesson (a phantom "moved 1.09 m" on a wall that never moved) is shared rather than re-learned; `maxMoveDistance` 10 m.

## A pre-existing bug fixed here

`geometryEqual` treats one-side-`undefined` as "geometry differs", so a capability mismatch (one revision fingerprinted with hashing on, the other off) fabricated `moved` for an entire model. There is now an explicit per-side abstention: if one side carries no geometry hashes at all, geometry classifies nothing. The same exposure exists on the **key-based** pass, but fixing it there would change `DiffState`/`DiffCounts` for every existing caller — out of scope here, worth its own issue.

## Three judgement calls for your review

1. **Uneven (N≠M) tier-1 sub-buckets fall to the residue** rather than reporting per-sub-bucket. Reporting them would be worse: base `{g1,g1}` / head `{g1,g2}` (one door stayed, one moved) would read "deduplicated", whereas the residue lets tier 3 pair them correctly.
2. **A tier-3 leftover of exactly 1+1 reports `ambiguous`, not `duplicated`** — it only exists because the cap or a tie refused it, and "we could not tell" is `ambiguous`.
3. **`renamed` no longer implies `base.length === 1`**, because of the N:N group case. Nothing consumes `contentMatches` yet, so no consumer breaks, but it is a real contract change and is called out in the changeset and docs.

`DiffState` and `DiffEntry` remain byte-identical, so the exhaustive `Record<DiffState, number>` in the viewer's `exportReport.ts` still compiles untouched.

## Module split

`diff.ts` 365 → 173. New: `content-match.ts` (233), `content-tiers.ts` (249), `geometry-compare.ts` (165), `mutual-nearest.ts` (127). All under the ~400 house limit.

## Verification

88 tests pass (was 58) · `pnpm typecheck` 33/33 · `docs:check-samples` 252 snippets clean · `check:api-surface` matches after `api-surface:update` (one additive line, `EntityAabb`) · `check-test-wiring` clean.

**One pre-existing test legitimately changed meaning** and is not quietly edited: the #1923 many-to-many group test used a 2:2 group sharing both `dataHash` *and* `geometryHash`, which is now exactly the N:N retire case. Its #1923 intent is preserved by removing the geometry hashes (2:2 with no geometry evidence still abstains and retires nothing); the N:N retirement is added as a separate, explicitly argued test.

18 mutations run. One of them came back green and turned out to be an incomplete mutation on my side rather than a missing test; re-run correctly, it reds two tests. All 18 confirmed, source restored.

Addresses the core of #1891 (@bruadam).

### DIFF
--- a/.changeset/tiered-content-match.md
+++ b/.changeset/tiered-content-match.md
@@ -1,0 +1,26 @@
+---
+'@ifc-lite/diff': minor
+---
+
+**diff**: make `matchUnpairedByContent` work on models full of repeated components.
+
+Unpaired entities were bucketed by (`ifcType`, `dataHash`) and paired only when a bucket held exactly one entity per side. A real model is mostly repeated components, so that paired only the unique minority: three data-identical doors at three different unmoved positions, all re-GUIDed by a re-export, landed in one bucket, reported a single `ambiguous` group, and pairs nothing at all.
+
+Each bucket is now refined hierarchically from the inside. Geometry stays out of the *outer* bucket key on purpose — with it there, an element that genuinely moved would never meet its own previous revision and every real move would revert to add+delete noise.
+
+- entities carrying a `geometryHash` are sub-bucketed by it. One per side, or the same count `N` on both sides, retires as `renamed`. `undefined` hashes are excluded: `undefined` agreeing with `undefined` is vacuous, not evidence.
+- a 1:1 leftover pairs as `renamed`, `moved`, or the new `reshaped` kind.
+- an N:M leftover is paired by iterated mutual nearest neighbour under a distance cap — a base and a head pair only when each is the other's *unique* nearest — which abstains by construction on a symmetric layout instead of guessing. The collision guard is part of that pairing test, so a pair it rejects stays in the candidate pool and the later rounds still see it. Groups above 128 per side report as `ambiguous`.
+
+New API:
+
+- `EntityFingerprint.aabb?: { min, max }` — optional world-space bounding box, both revisions in the same frame and units. It separates `moved` from `reshaped`, carries `ContentMatch.distance`, and enables the positional pairing above. Without it the pass degrades to the previous behaviour: `moved` for a 1:1 leftover, `ambiguous` for a group.
+- `ContentMatchKind` gains `'reshaped'` — the bounding boxes changed size, or agreed entirely while the geometry hash changed (a re-tessellation). An axis-aligned box cannot separate a re-tessellation from a reshape confined to the interior, and this does not pretend it can.
+- `ContentMatch.distance?: number` — centre displacement in the caller's units, clamped to `0` below `moveTolerance`.
+- `DiffOptions.moveTolerance` (default `2e-3`), `reshapeTolerance` (default `1e-3`), `maxMoveDistance` (default `10`). The two tolerances are lifted from `MOVE_EPS`/`RESHAPE_EPS` in the viewer's `describeChange.ts`, which encode issue #1197 (a phantom "moved 1.09 m" on a wall that never moved).
+
+Also fixed, in **both** matching passes: when one revision was fingerprinted by a build that produces geometry hashes and the other by a build that does not, every one-sided `undefined` read as "the geometry differs". The content pass reported the whole model as `moved`; the key-based pass reported every key-matched entity as `modified` with `changeKinds: ['geometry']`, so two revisions that may be identical read as a wholly changed model. Neither pass now uses geometry to classify anything in that case — a capability difference between two fingerprinting runs is not a model change. Both derive the decision from one shared helper so they cannot drift apart. The abstention needs a *whole side* to carry no hashes: with both sides hashing, a single entity gaining or losing geometry is still a real change and is still reported, and `excludeTypes` is applied before the scan so a dropped entity is not evidence that its side hashes.
+
+**`DiffCounts` changes for that previously broken case, and only for it.** A mixed-capability comparison that used to return every matched entity as `modified`/`['geometry']` now returns them as `unchanged` (or `modified` on data alone under `scope: 'both'`/`'data'`): `counts.modified` falls and `counts.unchanged` rises by the same number. This is a false positive being replaced with the truth, not a lost signal, and it applies whether or not `matchUnpairedByContent` is set. The cost is the mirror case: a base revision that genuinely carries no geometry at all, compared against a head that added geometry to everything, is indistinguishable from a capability difference and now reports `unchanged`. `DiffState` and `DiffEntry` are unchanged in shape.
+
+Behaviour change for existing callers of `matchUnpairedByContent`: a bucket with `N` entities per side that agree on both the data hash and the world geometry hash now retires as one `renamed` match carrying all `N` per side, where it previously reported an `ambiguous` group and retired nothing. `renamed` therefore no longer implies exactly one entity per side; `moved`/`reshaped` still do. `DiffState`, `DiffEntry`, and `DiffCounts` are unchanged, so exhaustive switches over `DiffState` keep compiling.
--- a/docs/guide/model-diff.md
+++ b/docs/guide/model-diff.md
@@ -72,43 +72,101 @@ Property sets, quantity sets, their members, and type assignments are all sorted
     but never deduplicated, so an occurrence bound to two types still hashes
     differently from one bound to a single type.
 
-**Geometry hash** — an opaque fingerprint of the entity's mesh, supplied separately (a `bigint` from the WASM mesh pass, `MeshCollection.geometryHashValues`, or a string for callers that fingerprint geometry another way). Two entities are geometry-equal when both hashes are absent, or both are present and their normalized values match; one side missing means geometry was added or removed.
+**Geometry hash** — an opaque fingerprint of the entity's mesh, supplied separately (a `bigint` from the WASM mesh pass, `MeshCollection.geometryHashValues`, or a string for callers that fingerprint geometry another way). Two entities are geometry-equal when both hashes are absent, or both are present and their normalized values match; one side missing means geometry was added or removed - unless one whole revision carries no hashes while the other does, which is a difference between two fingerprinting runs rather than a model change and is handled by [capability abstention](#capability-abstention).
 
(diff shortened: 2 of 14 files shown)


louistrue:
**CodeRabbit's hosted check reports `pass` on this PR, but its own status line says "Review rate limited" — it did not actually review.** Recording that here so the green check is not mistaken for a real one.

Ran the CodeRabbit CLI locally against the branch base instead (`--committed --base feat/1962-widen-stablehash`). It reviewed all 13 changed files and returned **one minor finding**, on `geometry-compare.ts`: `geometryEqual` compares `String(a) === String(b)`, so a hex-rendered string hash such as `'0x2a'` does not equal the `bigint` `42n`, and the two land in different content buckets.

**I did not apply the proposed fix, deliberately.** Canonicalizing numeric-looking strings would take an opaque caller token and reinterpret it as a number, colliding it with an unrelated `bigint`. `GeometryHash = bigint | string` documents strings as belonging to "callers that fingerprint geometry another way", so their internal structure is not ours to parse.

What I did instead: documented the contract on the type (both revisions must use the same representation) and **pinned it with two tests**. The finding does identify something this PR makes sharper, and that is now stated explicitly — before, a mixed-representation caller got a spurious "changed"; now that tier 1 *buckets* on the hash, they get no pairing at all.

The pin is mutation-checked against the exact change CodeRabbit proposed:

```
--- MUTATION: canonicalize hex (CodeRabbit's proposal) ---
  × treats a differently-rendered hash as a different fingerprint, not an equal one
  Tests  1 failed | 89 passed (90)
--- RESTORED ---
  Tests  90 passed (90)
```

Now 90 tests (was 88). Rebased onto the updated #1987.

louistrue:
On the `content-match-tiers.test.ts` finding ("this fixture cannot reach the code path the comment describes"): **that one is incorrect, and I checked before acting on it.**

The claim is that with no geometry on either side the capability abstention sets `useGeometry` to `false`, so `refineByGeometryHash` never runs. But the abstention is

```ts
const useGeometry = considerGeometry && baseHasGeometry === headHasGeometry;
```

Both sides carrying no geometry is `false === false` — **parity, not a mismatch**. `useGeometry` stays `true`, tier 1 does run, and it declines the `undefined` pair on its own merits, exactly as the comment says. The abstention only fires when one side has hashes and the other does not.

Demonstrated rather than argued. Mutating tier 1 so `undefined` hashes are no longer routed to the residue:

```
--- MUTATION: tier 1 buckets undefined hashes together ---
  × does not treat undefined geometry hashes as agreement
```

The original fixture reds. It was never inert.

**Kept the suggestion anyway, as an addition rather than a replacement**, because the underlying instinct is worth something even though the reasoning was wrong: a model where *nothing* has geometry is not what a real comparison looks like. There is now a second test, `still declines an undefined pair inside a model that does carry geometry`, where an anchor pair keeps `useGeometry` unambiguously true. Both fixtures kill the mutation, for different reasons.

91 tests (was 90).

louistrue:
Checked this one properly rather than taking it at face value, and **I am declining it — but the behaviour is now pinned by tests, because it is a real change from #1961 that a reader could trip over.**

**Why declined.** A `ContentMatch` whose `base` is empty asserts that some head entities are ambiguous against *nothing*. `groupKind(0, n)` would label that `deduplicated` or `ambiguous`, which is a strictly less honest description of an entity that is simply **added** — and it is already reported as `added` in `entries`, carrying its own `dataHash`, so a caller wanting the grouping can still reconstruct it. Nothing is lost that was not recoverable.

What replaces it is better information, not less. For base `{b1@gA}` / head `{h1@gA, h2@gB}`:

- **before:** one `duplicated` group of 1 base + 2 head, retiring nothing — "here are three entities, you work it out"
- **now:** `renamed b1 -> h1` (settled by tier 1), and `h2` reported as the plain addition it is

Two tests added: the case above, and a companion proving a group *is* still emitted when both sides genuinely have leftovers.

**One correction to the report, and one to my own verification.** The `continue` at `content-match.ts:212` is not the line that fires here — the equivalent guard right after tier 1 (line 175) catches it first.

And I nearly recorded a bad mutation check. Mutating line 212 alone left all 102 tests green, which looked like my test was vacuous. It is not: the two guards are **redundant**, so mutating either one alone leaves the other catching it. Mutating both together reds the test:

```
--- both guards mutated ---
  × reports the exact pair as renamed and leaves the odd one out as a plain add
  Tests  1 failed | 101 passed (102)
```

That nuance is written into the test comment, because "I mutated a line and nothing broke" is exactly the kind of result that gets quietly rounded up to "verified".

102 tests (was 100).

louistrue:
All four addressed. **The MAJOR was real and I was wrong to defer it.**

## 1. The key-based pass had the same mixed-capability bug [MAJOR] — fixed

I argued in this PR's description that fixing `geometryEqual`'s one-sided-`undefined` behaviour on the *key-based* pass was out of scope because it changes `DiffCounts`. That argument was weak, and there is a corroborating fact I had missed: **the viewer already sets `geometryUnavailable` when either side lacks hashes** (`apps/viewer/src/hooks/useCompare.ts:89`). So the engine was contradicting the viewer's own warning banner — telling the user "geometry comparison unavailable" while simultaneously reporting every entity as geometry-modified.

Fixed by resolving the abstention *once*, before classification, and routing both passes through it (`resolveUseGeometry`). The per-side scan applies the exact participation rule the classification walk uses — skipping an entity whose own class is excluded, and one whose keyed counterpart's class is excluded, since the either-side rule drops the pair. The content pass no longer re-derives its own scan; the two covered identical entity sets, so its behaviour is byte-identical.

**Counts change, stated plainly:** in a mixed-capability comparison only, entities formerly `modified`/`['geometry']` become `unchanged` (or data-only `modified`); under `scope: 'geometry'` such a pair now yields `modified: 0`. This reaches callers who never set `matchUnpairedByContent`. The commit's old "DiffCounts are untouched" line was corrected rather than left standing.

**And it cost a real test, which was not weakened to fit.** `geometry appearing or disappearing counts as a change` used a *one-entity* fixture where head gained a hash — indistinguishable from a capability difference. The fixture gained an anchor entity that hashes in both revisions, so both sides demonstrably hash and gained/lost geometry still reports `modified`. The genuinely degenerate case (base entirely geometry-free, head geometric) now reads `unchanged`, and that price is documented in the guide, the changeset and the commit message. The content pass already made this exact trade.

## 2-3. `maxMoveDistance` and the `aabb`-omitted sentence [MAJOR + MINOR, docs] — fixed

`maxMoveDistance` reaches only `pairByPosition` → `mutualNearestPairs`; the tier-2 1:1 path goes through `classifyGeometryChange`, which has no cap. The guide now scopes it to the N:M stage and says plainly that a 1:1 leftover is `moved` however far it travelled — the cap bounds *positional guessing among repeated components*, not how far the engine will believe an element moved.

## 4. My own doc overclaimed [MINOR] — corrected

I had written that mixing hash representations "stops the entities pairing at all". Wrong: `refineByGeometryHash` puts `42n` and `'0x2a'` in two 1-vs-0 sub-buckets, neither pairable, so both fall to the residue — where the 1:1 path pairs them and `classifyPair` returns `moved`. They pair fine. The real cost is "a missed `renamed` and a spurious geometry difference, not an unpairable entity", and that is what it now says. Both tests I added alongside it still assert true things; I re-derived them against the code rather than trusting the green.

## Mutation checks — 8 run, one of them instructive

M4 (`sideHasGeometryHash` drops the entity's own exclusion check) **survived first time**. Not a vacuous test and not a redundant guard: my fixture was wrong. The excluded hash-carrying entity had a counterpart in the other revision, so the *counterpart* check caught it and masked the own-entity check. They are genuinely independent — an excluded entity with no counterpart is only caught by the own-entity check. Narrowed the fixture so the entity exists in base only; M4 now kills exactly that test and M5 exactly the other, one guard each. The reason is recorded in a comment on the test.

## Verification

109 tests (was 102) · typecheck 33/33 · docs-samples 252 clean · api-surface matches · test-wiring, readmes, generated regions all clean · viewer compare suite 29/29. All modules under the 400-line limit.

PR #1992, opened 2026-08-02
feat(diff,cli): identity map so re-GUIDed elements stop reappearing (#1891)

louistrue:
> **Stacked on #1989**. Review that first; this branch contains its commit. Base moves to `main` once #1989 lands.

## What and why

This is the tractable half of @bruadam's "bringing old GlobalIds back" from #1891. The benefit he wants is that downstream references survive a re-export. The mechanism he suggested, rewriting GlobalIds into the file, is a one-way door: an identity claim that can be wrong, indistinguishable from truth once written, propagating to every copy. **This PR delivers the benefit without the mechanism.**

The repo already had the vocabulary: `IdentityMapEntry { base, here, reason }` (`packages/ifcx/src/provenance.ts:42`, specced in `layer-prs/03-provenance.md` §3.1). This makes content matching a **producer** of those entries and `diffModels` a **consumer**, so accepted matches from one comparison feed the next and re-GUIDed elements stop reappearing every time.

**No in-file GlobalId rewriting.** `04-identity.md` already states human-in-the-loop identity beats wrong automatic identity.

## Three pieces

**Producer** — `identityMapFromContentMatches`. Claims come only from `renamed | moved | reshaped` **and** only from 1:1 matches. `reason` is `content-match:<kind>`, deliberately not the bare `"derived"` that `04-identity.md` §4.1(3) reserves for the content-derived fallback, so the origin of every claim stays filterable.

**Consumer** — `DiffOptions.keyAliases`, applied before the key-based pass indexes anything. Four rejection rules, each degrading to today's behaviour rather than throwing: self-alias, stale target, target is a live head entity's own key, and target claimed by more than one head entity. The last two encode "the alias loses, both stay unaliased" — a collision means the map is wrong, and the map does not get to silently drop a real entity. `ModelDiff.appliedKeyAliases` echoes what took effect, mirroring `excludedTypes`.

**Sidecar + CLI** — `{ format, version, base: {hash}, head: {hash}, entries[] }`, pinned to the content hashes of *both* files, because a floating rename list is not a trustworthy claim. `--by-content` / `--identity-out` / `--identity-in` are an opt-in branch on the existing `diff` command; the default type-count path is untouched, so no existing user sees a behaviour change.

## Why N:N groups emit nothing

The tempting argument is that an N:N group is harmless to pair arbitrarily, since both hashes agreed N times and the members are indistinguishable. That is true *in this revision only*. The map is persisted and replayed. The pairing first matters in the later revision where two members diverge, or where one carries a BCF topic, a review comment, or a cost line — at which point a coin flip silently swaps two elements' histories, and the artifact records nothing about it having been a coin flip. The group stays in `contentMatches` for a UI to offer a human.

## Two findings the tests pinned

`IfcRel*` objects carry GlobalIds but live in the relationship graph, not the entity index, so a re-GUIDed relationship object correctly does not read as churn.

`--identity-in x --identity-out x` carries applied claims forward with their original `reason` and drops those that no longer hold. Without that, consume+emit erodes the map on every run.

Also fixed a latent bug this PR would otherwise have introduced: the command's `args.filter(a => !a.startsWith('-'))` would have swallowed `--identity-out map.json`'s value as a third file path.

## Verification

`@ifc-lite/diff` 138/138 · `@ifc-lite/cli` identity suite 12/12 · `pnpm typecheck` 33/33 exit 0 · `pnpm check:api-surface` exit 0 (37 packages, 3981 exports; +14 added, 0 removed → minor) · `docs:check-samples` 255 snippets clean · `docs:check-generated` 4/4 up to date · `check-test-wiring` clean.

**21 mutations, all kill.** One is worth calling out: M1 (`ambiguous` made claimable) initially **survived**, because the test used a 2:2 group that the 1:1 arity guard rejected anyway — so the *kind* exclusion was untested. `groupKind(1, 1)` returns `ambiguous` and is genuinely reachable when the position pass resolves part of an N:M group and leaves one per side, which is precisely the dangerous case. Added 1:1 `ambiguous` / `duplicated` / `deduplicated` tests; M1, M20 and M21 then all kill.

One honest limitation: "aliasing happens before content matching" has no single-line mutation. It is covered by M8 plus the assertion shape — the aliased pair reports `changeKinds: ['geometry']`, a key-pass classification a post-hoc alias could not produce, and `contentMatches` is empty where the un-aliased run has one match.

Toward #1891.

## Summary by CodeRabbit

* **New Features**
  * Added content-based model comparison for re-GUIDed IFC files.
  * Added geometry-aware detection of renamed, moved, and reshaped entities.
  * Added identity-map import/export with validated sidecar files for repeatable comparisons.
  * Improved collision and ambiguity handling for repeated entities.
  * Upgraded fingerprint hashing to improve distinction between similar data.

* **Documentation**
  * Added CLI usage, identity-map workflows, validation details, and content-matching guidance.

* **Tests**
  * Expanded coverage for content matching, geometry classification, hashing, and sidecar replay.


louistrue:
All four findings addressed. **Two fixed, one skipped with measurements, one confirmed-but-not-a-live-bug.** Plus one that was already fixed upstream.

## 1. Conflicting sidecar aliases [Greptile P1] — confirmed, fixed

Real, and worse than it looked: the consumer's collision validation **could not** have caught it. `resolveKeyAliases` rule 4 keys on the *target* (base) side, while the map is keyed by `here`, so a second claim for the same `here` was never visible to it. The arbitrary winner also propagated — `mergeIdentityClaims` builds its `reasons` map first-wins by `here` and re-emits.

Rejected at the **document** level (`parse` refuses, `create` throws), on an asymmetry the code did not previously distinguish:

- **Two `base`s for one `here`** — both entries are about the same head entity. It cannot be two base entities, and **no pair of files can break the tie**, so deferring only moves the guess downstream. Rejected outright, same stance as an unknown version.
- **Two `here`s for one `base`** — left alone. That document *is* coherent against some file pairs: one of the two head entities may have been deleted since, leaving one live claim that is simply true. Only the consumer knows, so rule 4 still adjudicates.

Conflict detection runs *before* self-claims are dropped, so `{base:'x',here:'x'}` plus `{base:'b',here:'x'}` cannot let the other through as winner.

## 2. Non-reproducible sidecar [Codex P2] — confirmed, fixed

`created` now preserves an incoming stamp and is otherwise omitted, so identical inputs give byte-identical output. No flag added: it would be CLI surface plus a generated-docs region for a field the format already documents as optional, and `git log` dates a committed artifact better than a self-report.

## 3. Repeated `EntityExtractor` construction [CodeRabbit Major] — measured, and skipped

**The flagged mechanism is free.** `EntityExtractor`'s constructor is one field assignment — **2.7 ns**; the ~11.5k constructions on the largest model total ~30 µs.

The real cost is re-parsing property-set records *shared between entities*. So rather than argue, I measured the ceiling: a perfect per-record memo, interleaved A/B, median of 9, six real models.

| model | no cache | cached | delta |
|---|---|---|---|
| duplex.ifc (2.4 MB) | 35 ms | 27 ms | −23% |
| dental_clinic.ifc (13.0 MB) | 373 ms | 213 ms | −43% |
| C20-Institute-Var-2.ifc (10.8 MB) | 212 ms | 249 ms | **+17% slower** |
| FM_ARC_DigitalHub.ifc (14.1 MB) | 44 ms | 40 ms | −8% |
| advanced_model.ifc (35.3 MB) | 258 ms | 260 ms | −1% |
| ISSUE_068 (56 MB) | 544 ms | 508 ms | −7% |

One model wins meaningfully, one **regresses**, the rest are noise. The cache also trades bounded transient allocation for retaining every parsed record for the run — the wrong trade in a CLI that must open a 177 MB file. No code change.

## 4. Export rows use the wrong key [CodeRabbit Minor] — correct, but not a live bug

Grepped every `keyAliases` call site: only the CLI passes it, so **the viewer cannot currently produce an aliased entry**. Fixed as future-proofing via `reportKey(entry)`, and the comment says so rather than implying a shipped defect was repaired.

One honesty note: `entry.base.key` for deleted rows is correct by symmetry but **provably equivalent today** — aliasing renames head keys onto base keys, so a deleted entry's `key` *is* its base key by construction. I wrote a regression test for it, found via mutation that it **could not fail**, and deleted it rather than ship a check that cannot fail. The model assertion moved into the killable test instead.

## Also: the type-GlobalId finding is fixed upstream

`sortedTypeAssignments` now projects `{ name, type }` only, and it is the sole reader of `input.typeAssignments` on both hash paths. The adapter still passing `globalId` is inert — the field is declared and read by nothing. Fixed in `e30e155f`, confirmed an ancestor of this branch.

## Verification

`@ifc-lite/diff` 156 · `@ifc-lite/cli` 215 (16/16 files, 8 pre-existing conditional skips) · typecheck 33/33 · docs-samples 255 clean · generated regions up to date · api-surface matches · viewer `exportReport` 12/12. **11 mutations, every surviving test killed** — except M10, which is the one I deleted.

louistrue:
All six addressed. **One of them turned out to be the smaller half of a worse bug.**

## 1. `--identity-out` could overwrite an input IFC [MAJOR, data loss] — fixed

Guarded in `contentDiffCommand` *before* any read, compare or write, by **two independent rules**, each isolated by its own test:

- **resolved-path equality** — catches `./v1.ifc` vs `v1.ifc`, and `${dir}/./sub/../v1.ifc` where `sub` does not exist, so no `stat` could resolve it
- **`dev`+`ino` identity** — catches symlinks, hard links, and case-insensitive filesystems

The inode rule is exhaustive on its own for the destructive case (a path that does not resolve to an existing file cannot be overwriting an input, since inputs must exist to be read); the lexical rule is what gives a clean error before touching the filesystem. Every test asserts the models are **byte-identical afterwards**, not merely that the command exited non-zero. `--identity-in x --identity-out x` still works.

## 3. Category-skipped entities [MAJOR] — verified, and the reported half was the smaller one

Probed before changing anything: `IFCTASK`/`IFCACTOR` are in `entityIndex.byType`, but `getGlobalId` returns `''` and `getTypeName` returns `'Unknown'`, because `CAT_SKIP` keeps non-`IfcProduct` `IfcObject`s out of the `EntityTable`. Silent drop-out confirmed.

**The unreported half is worse.** The parser fills the table's GlobalId column *positionally*, and slot 0 of a resource entity is a **Name**. Measured across the four bundled sample models, the old fingerprint set contained **7-9 colliding keys per file** (a material and a surface style sharing a name landing on one key) and 23/25/16/11 entries that are not `IfcRoot` at all — `IfcMaterial`, `IfcSurfaceStyle`, `IfcClassification`, `IfcProjectedCRS`, `IfcGeometricRepresentationSubContext`.

Membership now comes from the schema registry inheritance chain: every `IfcObjectDefinition` in, `IfcRelationship` and `IfcPropertyDefinition` out by rule, classified once per *type* so geometry buckets are dismissed without touching a row. `ifcType` comes from the registry spelling, never `'Unknown'`. On real models the new key set loses nothing that is an `IfcRoot`, gains zero spurious entries, and duplicate keys go **7-9 → 0**.

I also had to correct two claims in my own earlier module comment: `IfcRel*` do **not** "live in the relationship graph rather than the entity index" (`byType` holds every entity; they were absent from the `EntityTable`, which is a different thing), and this was never "the same set `--by-entity` compares".

## 2, 4, 5, 6 — fixed

`readModel` helper with the same `fatal` convention; exact-two positional validation; the README import; and 10 command-level tests with `contentDiffCommand` mocked.

One correction to finding 5's premise: `check-doc-samples.mjs` covers the root `README.md`, `docs/guide` and `docs/tutorials` only, so `packages/diff/README.md` is **not** CI-typechecked. The fix is still right for a reader, but it was not repairing a gate failure.

## The mutation that survived

`unknown` → `independent` initially passed. Investigated rather than reported done: it exposed a real inconsistency — an unregistered `IFCRELVENDORLINK` enters the parser's table via its name-prefix rule and would have been the single relationship still reaching the comparison. Added the `IFCREL` name rule plus three vendor-class fixtures; the branch is now killable in both directions.

## Known gap, deliberate and pinned

A vendor `IfcRoot` subtype outside the codegen pin stays uncompared unless its name ends in `TYPE`. Treating unrecognised classes as possible `IfcObject`s would cost one STEP extraction per row of every unrecognised bucket on every file. Pinned by a test with the reasoning in a comment.

Separately: `--by-entity` (pre-existing, different flag) still reads the `EntityTable` and so still misses tasks and actors. Left alone as out of scope, and the module comment no longer claims parity with it.

## Verification

`@ifc-lite/diff` 163 · `@ifc-lite/cli` 233 (18 files, 8 pre-existing skips) · typecheck 33/33 · docs-samples 255 clean · generated regions up to date · api-surface matches. 17 mutations, all killed. `diff-content.test.ts` had reached 565 lines and was split into three files under the house limit, with all mutation kills re-verified after the split.

PR #1993, opened 2026-08-02
feat(geometry): carry the mesher's closedness verdict out, and ship volume only where it holds (#1891)

louistrue:
> **Stacked on #1988.** Review that first; this branch contains its commit. Base moves to `main` once #1988 lands.

## What and why

#1988 deliberately shipped no per-element volume, because 85% of mesh segments are closed and for an *open* surface the divergence-theorem sum is not inaccurate but **arbitrary** — the boundary flux scales with distance to the reference point. Worse, only ~1 bad case in 130 is detectable as bad; the rest look like ordinary plausible volumes.

The missing piece was already being computed and thrown away: `orient_mesh_outward` decides per connected component whether it is closed and orientable, then discarded the verdict (`mesh_orient.rs:244-249`). This carries it out, and ships volume **only where it holds**.

`orient_mesh_outward` is now a one-line wrapper over `orient_mesh_outward_verdict`, so **no caller signature changed**. New at the wasm boundary: `geometryVolumeValues` (Float64Array, `NaN` = absent, matching #1988's convention) and `geometryClosureFlags` (Uint8Array bitfield), behind the same `setComputeGeometryHashes` gate.

## The gate, and why it is this strict

Volume is emitted only when `segments == 1 && all_closed && all_orientable && all_single_component`. Coverage: **24,073 / 33,701 elements (71.4%)**. The rest report nothing rather than something plausible.

**The multi-segment sum is refused, on evidence.** Of 4,472 all-closed multi-segment elements, 2,971 (66%) have a segment pair whose world boxes overlap by >1% of the smaller — usually total containment. Summing them yields a volume **larger than the element's own AABB on 987 elements** (22%), p90 2.99x, max 3.00x. Restricted to one segment: **zero** such elements, p50 fill 0.96, max exactly 1.00.

A code-level argument agrees without needing the measurement: when the sub-mesh path fails, `router/processing.rs` merges every item into one mesh, which then has 2 components and is already refused by `all_single_component`. Accepting the sum on the sub-mesh path would make the same element report a volume or not depending on which router entry point happened to succeed.

**`all_single_component` is load-bearing independently of overlap.** The orienter flips each closed component to *positive* volume, so a solid plus its own cavity shell sums to `outer + cavity`, not `outer - cavity`.

## Two things found on the way

One element in 24,073 arrives inward-wound: `orient_mesh_outward`'s flip decision sums about the mesh's *local frame origin*, which on a far-offset frame is the same reference-point sign hazard documented on `signed_volume6`. That is a pre-existing normals defect, left untouched and out of scope; `volume()` takes the magnitude and says so rather than pretending.

`signed_volume6`'s tetrahedron determinant was extracted to `tetra_volume6` and both callers use it. The existing expression is unchanged.

## No geometry output changed

`reporting_the_verdict_does_not_change_a_single_index` runs both entry points over five mesh shapes asserting byte-identical positions, indices and flip bool. `hash_values_are_byte_identical_to_the_pre_aabb_implementation` still passes.

## Verification

`cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings` exit 0 · `cargo test --workspace --no-fail-fast` exit 0, 1248 passed / 0 failed · `module_size_ratchet` 4/4 · `build-wasm.sh` + `test-wasm-contract.mjs` exit 0, 39/39 · `docs:check-samples` exit 0.

Module sizes: three splits instead of new rows (`geom_closure.rs` 189, `mesh_orient_tests.rs`, `mesh_fingerprint.rs` 161). `zero_copy/mesh.rs` goes **down** 792 to 724. Two rows up with written justifications.

**15 mutations, 14 red, 1 untestable — and two of those deserve your attention rather than the summary:**

- **One mutation passed when it should not have.** The far-origin volume test used round offsets (400000, -250000) at which every product stays exactly representable in f64, so it stayed green even with the reference point wrong. Switched to non-dyadic offsets; it now reds, and the reason is written into the test.
- **One mutation is still green and could not be fixed.** Instrumenting both hash sites in `produce_inner` across 26 fixtures shows the single-mesh fallback and element-level palette-split paths fire **zero times** with hashing enabled. They are unreachable by any available fixture, so that site is pinned only at unit level. The failure mode is volumes disappearing, not becoming wrong. Worth knowing #1988's AABB never exercised those paths either.

Unblocks the quantity-conservation split/merge detector for #1891 — at 71.4% coverage, with the remaining 29% declining to make a claim rather than making a bad one.


louistrue:
Both P1s addressed. **The closure finding was real and is reproduced by a purpose-built fixture; the perf verdict is now recorded, including a number that argues against me.**

## 1. Closure metadata precedes mesh cleanup [Greptile P1] — confirmed

Verified the ordering rather than assuming it: `orient_mesh_outward_verdict` runs on the assembled body and its verdict goes straight to `hasher.add_oriented_mesh`; `build_mesh_data` runs **later** and calls `drop_degenerate_triangles`. The hasher consumes the mesh before cleanup.

Does dropping a degenerate triangle actually open a closed shell? **Yes — for the aspect-ratio branch**, not the bit-identical one (the orienter's 10 µm weld already treats those as non-manifold). A triangle whose shortest edge sits *above* that weld grid keeps three distinct welded vertices, so removing it takes each of its three edges from 2 incidences to 1.

The fixture is a needle tetrahedron: A–B edge 0.1 mm, body ~25 m, two faces at aspect ~2.2e5 (dropped) and two at ~1.1 (kept). 0.1 mm is deliberately above the weld grid so the pre-cleanup shell genuinely is one closed orientable solid. Against unfixed code:

```
panicked at tests/geometry_closure_after_cleanup.rs:251:
the verdict still calls this a single closed solid (0b1111) after cleanup opened it
```

The premise asserts (`dropped == 2`, returned mesh re-derives as open) pass first, so that failure is the defect and not a broken fixture.

**Fixed by retracting, not re-deriving**, and the reason matters: re-running the orienter on a copy would give the correct post-cleanup topology, but `volume6` was accumulated over the *pre-cleanup* triangle set, and a dropped needle's tetrahedron is not zero — it scales with the lever arm to the reference corner. A re-derived verdict would license a stale number. So any drop clears the three topology bits and the volume with them. The gate is untouched, and not one vertex or index moves.

Now documented in three places, because it is the kind of thing a future reader will get wrong: **a clear flag bit means "not proved", never "proved false"**.

A third mutation is worth noting: raising `MAX_ASPECT` from 1e5 to 1e7 so the backstop cannot fire makes the fixture's *premise* assert fail. That proves the test genuinely depends on the mechanism it claims to test, rather than passing for some incidental reason.

## 2. Performance verdict [Codex P1] — recorded in `scripts/perf/README.md`

Output byte-identical throughout: mesh/vertex/triangle counts unchanged, and an FNV-1a over every `geometryHashValues` entry is *equal* base-vs-branch on all three fixtures.

**Hashing OFF** (the path every normal user hits), house probe, interleaved A/B: AC20 0%, ISSUE_129 +1.4% inside a ±10% band, Holter −1.0%.

**Hashing ON** — and the house probe **cannot reach it**: `process_geometry` hardcodes `MeshProductionOptions::default()`, so `geometry_hash` is always `None` natively and the pass only exists behind `setComputeGeometryHashes`. Drove the real wasm entry point instead: AC20 +0.8%, ISSUE_129 +0.5%, Holter +1.6%.

Turning the *switch* on is the real cost, and it is nearly identical on both sides — off→on is +2.9/+0.6/+6.9% on branch versus +2.2/+0.3/+6.6% on base. This branch adds roughly 0.3–0.7 pp to a surcharge only the diff feature pays.

**Reported against interest:** hashing-off on Holter reads **+1.2%** at the wasm boundary while the native probe on the same fixture reads **−1.0%**. Nothing in that path changed structurally (the hasher is `None`, every new accumulator is dead) and it sits inside the base's own 3528..3584 ms spread, so I read it as a binary-size/code-layout shift from 3.7 KB — but it is in the ledger rather than rounded away, along with the harness gap that made the ON measurement need a different driver.

**Verdict: viable.** Default path unaffected, feature cost sub-1% of an already-small surcharge.

## Module sizes

Split, not allowlisted: the degenerate backstop moved to a child module `element_degenerate.rs` (72 lines) — the drop tally stopped being a diagnostic the moment it gated the verdict. `element.rs` 932 → **910**, its allowlist row refreshed **downward**.

## Gates

clippy `-D warnings` exit 0 · `cargo test --workspace --no-fail-fast` exit 0, 140 binaries, 0 failed · module-size ratchet 4/4 · `build-wasm.sh` + contract test exit 0, 39/39 · `hash_values_are_byte_identical_to_the_pre_aabb_implementation` and `reporting_the_verdict_does_not_change_a_single_index` both still pass.

PR #1995, opened 2026-08-02
fix(changeset): version the packages that re-expose the breaking IfcCreator (#1986)

louistrue:
Fixes the three Codex P1 threads on #1986. The fix belongs here rather than on the release PR — #1986 is regenerated from `.changeset/*.md`, so edits to its `package.json`/CHANGELOG would be overwritten on the next run.

## The problem

`uniform-storey-placement.md` correctly declares `@ifc-lite/create: major`, but three packages re-expose `IfcCreator` and were left on compatible bumps:

| package | current | #1986 proposes | correct |
|---|---|---|---|
| `@ifc-lite/create` | 1.17.0 | **2.0.0** ✅ | 2.0.0 |
| `@ifc-lite/sdk` | 1.21.4 | 1.21.5 (patch) ❌ | **2.0.0** |
| `@ifc-lite/sandbox` | 1.16.4 | 1.17.0 (minor) ❌ | **2.0.0** |
| `@ifc-lite/cli` | 0.21.1 | 0.21.2 (patch) ❌ | **0.22.0** |

A consumer on `@ifc-lite/sdk@^1.21` accepts 1.21.5 and receives storey-relative placement with nothing in the range to warn them. On a non-zero-elevation storey their elements move by one elevation, silently — which is the exact failure the `create` major exists to announce.

## Verified, not inferred from the dependency graph

Being depended-on isn't enough to justify a major; I checked that each package actually re-exposes the changed surface:

- **sdk** — `packages/sdk/src/index.ts:222`: `export { IfcCreator } from '@ifc-lite/create';` — the class itself, directly.
- **sandbox** — `packages/sandbox/src/bridge-create.ts:13` imports `IfcCreator` and `buildCreateMethods()` auto-discovers `IfcCreator.prototype`, so every affected constructor is reachable from a sandbox script. Breaking for the script author even though the sandbox's own export surface is unchanged — which is why minor was wrong.
- **cli** — `packages/cli/src/commands/create.ts:21` constructs it and `:50` reads `--elevation`, passing it straight through.

`@ifc-lite/wasm`, `@ifc-lite/mcp` and the viewer apps neither construct nor re-export it, so they keep their dependency-only bumps.

**cli takes a minor, not a major**, because it is pre-1.0 — AGENTS.md maps a breaking change to `minor` for a 0.x package. That matches Codex's own recommendation of a 0.22 release.

## Verified output

Ran `changeset version` on this branch and reverted the result, so only the changeset is committed here:

```
@ifc-lite/create   2.0.0
@ifc-lite/sdk      2.0.0
@ifc-lite/sandbox  2.0.0
@ifc-lite/cli      0.22.0
@ifc-lite/mcp      0.9.3   (unchanged bump)
@ifc-lite/wasm     4.2.3   (unchanged bump)
```

## What happens on merge

The changesets action regenerates #1986 with the corrected versions and the added changelog section, which explains to each downstream consumer why *their* package went major. Nothing needs doing to #1986 by hand.

One file changed, +13/-1.

## Summary by CodeRabbit

* **Breaking Changes**
  * Storey-relative placement behavior has been updated across the SDK, creation tools, and sandbox.
  * Applications may need migration updates to align with the new placement model.
  * Unknown storeys now produce clear errors instead of being silently accepted.

* **Bug Fixes**
  * Corrected placement metadata in the sandbox.

* **Documentation**
  * Added placement metadata guidance for 13 constructors.
  * Documented migration steps and package release impact.

### DIFF
--- a/.changeset/uniform-storey-placement.md
+++ b/.changeset/uniform-storey-placement.md
@@ -52,7 +52,7 @@ Thirteen constructors that carried no `llmSemantics` at all — `addIfcRamp`, `a
 
 ## Downstream packages carrying the break
 
-The behaviour change is not confined to `@ifc-lite/create`: three packages re-expose `IfcCreator` and therefore ship it to their own consumers. Each is versioned to say so, rather than letting a caller pick the change up through a range they believed was compatible.
+The behaviour change is not confined to `@ifc-lite/create`: four packages re-expose `IfcCreator` and therefore ship it to their own consumers. Each is versioned to say so, rather than letting a caller pick the change up through a range they believed was compatible.
 
 - **`@ifc-lite/sdk` (major)** — re-exports the class directly (`packages/sdk/src/index.ts`: `export { IfcCreator } from '@ifc-lite/create'`). Without a major, a consumer on `^1.21` accepts the release and gets storey-relative placement with no signal.
 - **`@ifc-lite/sandbox` (major, was minor)** — `buildCreateMethods()` auto-discovers `IfcCreator.prototype` and dispatches to it, so every affected constructor is reachable from sandbox scripts. A script passing absolute coordinates against a non-zero-elevation storey now emits geometry one elevation off. That is breaking for the script author even though the sandbox's own surface is unchanged.
--- a/.changeset/uniform-storey-placement.md
+++ b/.changeset/uniform-storey-placement.md
@@ -1,6 +1,8 @@
 ---
 '@ifc-lite/create': major
-'@ifc-lite/sandbox': minor
+'@ifc-lite/sdk': major
+'@ifc-lite/sandbox': major
+'@ifc-lite/cli': minor
 ---
 
 **BREAKING:** every `IfcCreator` element constructor now places its product relative to the storey it is added to. Element coordinates are storey-relative across the whole API.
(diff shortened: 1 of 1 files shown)


louistrue:
Both CLI findings triaged against the code.

**Fixed (minor):** the section said "three packages" while listing four — I added `@ifc-lite/mcp` without updating the count. Corrected in `52979ef`.

**Dismissed (major), with reason:** the suggestion to exclude `addElement` from the storey-relative migration is not right. `ifc-creator.ts:2409`:

```ts
addElement(storeyId: number, params: GenericElementParams): number {
  const placementId = this.addLocalPlacement(this.getStoreyPlacement(storeyId), params.Placement);
```

It chains the caller's `Placement` to `getStoreyPlacement(storeyId)`, so `Placement.Location` is storey-relative like everything else. A caller who previously passed an absolute `[0, 0, 3.2]` must now pass `[0, 0, 0]` — the migration applies to it exactly as written.

The apparent contradiction is a terminology collision, not an error: the sandbox section's `addElement` **explicit-placement** tag describes how coordinates are *supplied* (a `Placement` object rather than named `Position`/`Start`/`End` keys), not which datum they sit in. Both statements are true on different axes. Leaving the migration list as-is.

Also flagging for the record: **the `@ifc-lite/mcp` P1 was correct and I was wrong.** I'd checked whether mcp constructs or re-exports `IfcCreator`, found only test files, and wrote "unaffected". The right question was whether mcp hands consumers something that constructs it — and `loadIfcModel()` → `LoadedModel.bim` → `bim.create.building({ StoreyElevation })` does exactly that. Same transitive reasoning I'd applied to sandbox and then failed to apply one package over. mcp is now `minor` → 0.10.0.

Final verified output of `changeset version` on this branch:

```
create 2.0.0 · sdk 2.0.0 · sandbox 2.0.0 · cli 0.22.0 · mcp 0.10.0 · wasm 4.2.3
```

Issue #1957, opened 2026-08-01
Undone attribute edits are resurrected on export — StepExporter replays mutation history, undo writes with skipHistory


BIMvoice:
### Summary

An attribute edit that has been undone is resurrected on export. `StepExporter` reconstructs attribute values by replaying the append-only mutation history rather than reading the current overlay, and undo deliberately does not write to that history — so the original pre-undo value survives and is baked into the output.

This affects STEP export today. Surfaced during review of #1956 (@greptile-apps), which gives HBJSON export the same `StepExporter` path and therefore the same exposure.

### The chain

1. `packages/mutations/src/mutable-property-view.ts:1181` — `getMutations()` returns `[...this.mutationHistory]`, the append-only history, not the current overlay. Its sibling `hasPendingChanges()` documents the distinction explicitly.
2. `packages/export/src/step-exporter.ts:267` — `const mutations = this.mutationView.getMutations();`
3. `packages/export/src/step-exporter.ts:280` — the attribute map is populated from the record itself: `mutation.newValue == null ? '' : String(mutation.newValue)`.
4. Undo applies its reverse edit with `skipHistory: true` (see the `skipHistory` parameter documented at `mutable-property-view.ts:321` and `:422`), so the original `UPDATE_ATTRIBUTE` record remains in history carrying its stale `newValue`.

Net effect: the overlay is correct and the editor shows the reverted value, but the exporter replays the superseded one.

### Reproduction

1. Edit an attribute on any entity.
2. Make a second, unrelated edit so the overlay still has pending changes.
3. Undo the first edit.
4. Export to IFC/STEP.

Expected: the exported attribute matches what the editor shows. Actual: the pre-undo value is written.

Only the attribute path was traced. Whether the pset/quantity paths share the shape was not checked, and is worth confirming before fixing.

### Why this is not fixed in #1956

Two candidate fixes, both with reach well beyond that PR:

1. **Read the overlay instead of the history inside `StepExporter`.** This is probably the correct model — the overlay is what `hasPendingChanges()` calls "the same set the exporter reads", which is currently not true. But it changes STEP export output for every existing user and wants its own regression corpus.
2. **Make undo rewrite history.** A semantics change to the mutation model, with knock-on effects for anything else replaying `getMutations()`.

#1956 is a one-line input fix for HBJSON; it neither causes this nor is the right place to resolve it.

### Note on severity

[...]

PR #1996, opened 2026-08-02
fix(export): stop deleting an edited entity from emitting a dangling relation (#1978)

### LINKED ISSUE #1978: Deleting an edited entity emits a dangling IfcRelDefinesByProperties reference (invalid IFC)
## Summary

Editing a property or quantity on an entity and then **deleting that entity** produces a STEP file containing an `IFCRELDEFINESBYPROPERTIES` that references an expressId with no corresponding entity line — a dangling reference, i.e. invalid IFC rather than merely a dropped edit.

Found while tracing a review finding on #1967. Deliberately not fixed there: it is the same category as #1957 (the exporter reading mutation state that the overlay has superseded) and has a wider blast radius than that PR.

## Reproduction

1. Load a model.
2. Edit a property or quantity on some entity `#N`.
3. Delete `#N`.
4. Export to IFC/STEP.

The output contains no line for `#N`, but does contain `#R=IFCRELDEFINESBYPROPERTIES('…',…,(#N),#P);`.

## The chain

- `MutablePropertyView.getMutations()` (`packages/mutations/src/mutable-property-view.ts:1196`) returns the **unfiltered** `mutationHistory`. It does not consult tombstones.
- `packages/export/src/step-exporter.ts:304-389` groups property/quantity mutations by `mutation.entityId` from that unfiltered history into `newPropertySets` / `newQuantitySets`.
- The entity loop at `step-exporter.ts:586-590` **correctly** skips the deleted entity, so no line is emitted for it:
  ```ts
  if (overlayActive && typeof this.mutationView!.isDeleted === 'function' && this.mutationView!.isDeleted(expressId)) {
    continue;
  }
  ```
- But the property-set emit loop at `step-exporter.ts:689-712` is **unconditional** — it never checks `isDeleted` — and reaches `generatePropertySetEntities`, whose relation template at `step-exporter.ts:967` writes the entity reference directly:
  ```ts
  const relLine = `#${relId}=IFCRELDEFINESBYPROPERTIES('${relGlobalId}',${...},$,$,(#${entityId}),#${psetId});`;
  ```
- `allowedEntityIds` is not a guard here — it is the isolation/subset filter and is `null` when not isolating.

So attribute, positional and type mutations for a tombstoned entity are correctly discarded, and property/quantity **set** mutations are not.

## Why this is worse than a dropped edit

A dropped edit is a wrong model. This is a **structurally invalid** one: consumers that resolve references will fail or silently drop the relation, and it violates the reference-closure property the exporter otherwise maintains (`subset_export_is_reference_closed` covers the isolation path, not this one).

## Candidate fixes

1. **Filter at collection.** Skip mutations whose `entityId` is tombstoned when building `newPropertySets`/`newQuantitySets`. Narrowest change, keeps the fix next to the cause.
2. **Filter at emit.** Guard the `689-712` loop with the same `isDeleted` check the entity loop uses. Symmetrical with the existing skip, but leaves the collection pass doing wasted work.
3. **Filter at source.** Have `getMutations()` (or a new accessor) exclude tombstoned entities. Broadest — it would fix every consumer at once, but it changes published behaviour and would need checking against the other callers.

I'd lean to 1 or 2. 3 is tempting but `getMutations()` is public API and #1957 shows other consumers already depend on its exact semantics.

## Related

- #1957 — `StepExporter` replays mutation history while undo writes with `skipHistory`, so undone attribute edits are resurrected on export. Same root shape: the exporter trusting history over the effective overlay.
- #1967 fixes the *review* side of this (`getEffectiveChanges()` now reports only `entity-deleted` for a tombstoned entity, so the UI no longer claims the superseded edits will be exported). It does not touch the exporter.

Not fixed in a PR because the right layer is a call worth making deliberately, and the exporter has its own regression corpus.

BIMvoice:
Closes #1978.

## The bug

Editing a property or quantity on an entity and then deleting that entity produced a STEP file with a live `IFCRELDEFINESBYPROPERTIES` whose `RelatedObjects` referenced an expressId that has no defining line. That is **structurally invalid IFC**, not merely a dropped edit — strict readers reject the file.

`step-exporter.ts` had exactly one `isDeleted` call in the whole file, guarding the entity-emission loop. `newPropertySets`, `newQuantitySets` and `typeOwnedPsetNamesByEntity` are built from `getMutations()`, which returns unfiltered history and does not consult tombstones, so their consumers still emitted for the deleted entity.

## The fix

Option 2 from the issue — filter at emit. All three consumers now check `isDeleted` before emitting anything, using the same optional-method probe the entity loop already used (the overlay may not implement it).

Not option 3: `getMutations()` is public API and other consumers depend on its current semantics, as #1957 shows. You flagged that yourself and I've stayed off it.

**One thing the issue didn't list:** the type-owned pset rewrite is the same defect on a separate path. An entity with no `newPropertySets` entry never reaches the first loop's guard, so it needed its own. Fixing only the two loops named in the issue would have left that case broken.

The four guard sites are now a single `isTombstoned` closure rather than the four-clause condition repeated verbatim.

## Tests

Structural, not string-matching an id: export, then require every `#N` referenced anywhere in the output to have a defining `#N=` line. That pins the whole dangling-reference class rather than this one instance, and reuses `findDanglingRefs`, already in the file.

- RED first: both tests failed with `expected [ 1 ] to deeply equal []` — a real dangling `#1`, not a setup error
- Mutation: removing the three guards fails exactly those two tests, **248 pass / 2 fail**; restored, **250 pass / 28 skipped**
- `pnpm typecheck` **33/33**

## Found but deliberately not fixed

`ifc5-exporter.ts` has **no `isDeleted` check anywhere** — IFC5/IFCX export appears not to honour overlay deletions for *any* entity, not just ones with pset edits. That is a broader gap than this issue's scope and a different shape from the bug reported here, so I left it alone. Happy to file it separately if you agree it's real.

Draft until you've had a look.


louistrue:
Heads-up from #2011, which touches the same file: **there is a case this fix does not close, and it is worth knowing before this merges.**

While auditing `StepExporter` for the same source-buffer-vs-overlay shape, I probed `entity_delete` on an **overlay-created** entity that had property edits. It still emits a dangling relation:

```
#17=IFCRELDEFINESBYPROPERTIES(...,(#14),#16);      <- no #14 anywhere
```

The reason your four `isTombstoned` guards do not catch it: `deleteEntity` **forgets** a new entity rather than tombstoning it — it is removed from `newEntities` instead of being marked deleted — so `isDeleted()` returns `false` and every guard passes.

So the two id sets are disjoint by construction: your fix covers source-buffer entities that were tombstoned, and this case is a created entity that was never tombstoned at all. Same symptom, different path in.

**Not a criticism of the fix** — it closes exactly what #1978 described, and the type-owned pset path you found on top of the issue's list was a good catch. This is an adjacent hole that the issue did not name either.

Your call whether to fold it in here or leave it. If you leave it, say so in the PR body so the next person does not assume "dangling relations from deleted entities" is fully closed — that assumption is what made this one survive.

## No conflict with #2011

Different regions of the file: yours at ~264 plus four guards in 700-770; mine at ~250, one line in the modified-count loop, the new-entities pass at ~818, and a new method at ~1110. Different test files. Whichever lands second rebases cleanly.

Broader observation from that audit, for whatever it is worth: this is one defect with five faces, all of them `StepExporter` reaching for `this.dataStore` where the overlay is the authority — your dangling relation, #2011's dropped attribute edits, #2000's empty-`store.properties` base, `visibleOnly: true` silently dropping every created entity, and `isTypeEntity()` misclassifying an overlay-created type. A single "effective model" view answering *exists / class / attributes / deleted* from the overlay first would close all of them, but that is a refactor rather than a patch, and none of us should attempt it inside a fix PR.

BIMvoice:
Thanks — folding it in, not leaving it.

Your reasoning is the argument for doing so: if the PR title says "dangling relation from a deleted entity" and only half the paths are closed, the next person assumes it's done. That assumption is what let this one survive, and I'd rather not rebuild the trap one layer up.

I verified the mechanism before acting on it rather than taking it on faith — `mutable-property-view.ts:1039`:

```ts
if (this.newEntities.has(expressId)) {
  this.newEntities.delete(expressId);   // forgotten, not tombstoned
  ...
  return true;
}
...
this.tombstones.add(expressId);
```

with `isDeleted()` reading only `this.tombstones`. So the id sets are disjoint exactly as you describe, and no `isTombstoned` guard could ever have caught it — it isn't a gap in the guards' placement, it's a gap in what the predicate can see.

Work in progress now, with three constraints taken from your comment:

- **Not touching `deleteEntity`.** Making it tombstone new entities would change overlay semantics well past this PR, and `restoreNewEntity` pairs with the current forget-behaviour for undo. Exporter-side fix.
- **Staying in our regions** (~264 and the guards in 700-770), clear of #2011's ~250 / ~818 / ~1110.
- **No "effective model" refactor here**, per your explicit call. Patch only.

Aiming at the general invariant — don't emit a relation referencing an entity that won't appear in the output — rather than special-casing `createEntity`, since that closes both faces at once. One thing I'm having verified rather than assumed: line ~618 suggests overlay entities may also sit in `completeIndex` with zero-length refs, so a naive `completeIndex.has(id)` might not exclude a forgotten entity at all. If that turns out to be true the predicate has to be built differently, and I'll say so instead of quietly shipping something that passes for the wrong reason.

On the five faces — agreed, and agreed it doesn't belong in a fix PR. Once these land I'm happy to write it up as its own issue with the five instances enumerated, so the refactor has somewhere to be argued on its merits rather than accreting piecemeal.

BIMvoice:
Folded in — `89c48cae`.

Your repro is closed, and establishing the right predicate turned up a fact worth recording, because it invalidated my first plan.

I expected to filter on `completeIndex.has(id)`, on the theory (from the `entityRef.byteLength === 0 || entityRef.byteOffset < 0` skip at the entity loop) that overlay entities sit in the index with placeholder byte ranges. They don't. **An overlay-created entity is never in `getCompleteEntityIndex` in either state** — `StoreEditor.addEntity` builds an `EntityRef` with `byteOffset: -1 / byteLength: 0`, but nothing ever writes it into `entityIndex.byId`. Probed both states directly: alive and created-then-deleted are both simply absent, index size unchanged.

So a `completeIndex` membership test could never have separated overlay-alive from overlay-forgotten, and that skip at the entity loop appears to be defensive/legacy for this exporter — in the current path it never fires for overlay entities, because they aren't there to skip.

The predicate that does work resolves the two id spaces through their own authorities:

```ts
const willBeEmitted = (entityId: number): boolean => {
  if (isTombstoned(entityId)) return false;
  const ref = completeIndex.get(entityId);
  if (ref && ref.byteLength > 0 && ref.byteOffset >= 0) return true;
  if (overlayActive && typeof this.mutationView!.getNewEntity === 'function') {
    return this.mutationView!.getNewEntity(entityId) !== null;
  }
  return false;
};
```

Source ids through `completeIndex`, overlay ids through `getNewEntity` — which returns null precisely once `deleteEntity` forgets one. General "will this have a defining line" invariant, not a `createEntity` special case, so both faces close together.

**Evidence they're genuinely distinct paths:** reverting only the predicate and restoring the plain tombstone guards fails exactly the two new tests while the four original #1978 tests stay green — 250 pass / 2 fail. Restored: **252 pass / 28 skipped**. Typecheck 33/33.

Over-filtering, checked rather than assumed:
- **Type-owned psets** — `entityId` may be a type entity, but types are ordinary `completeIndex` members and resolve identically.
- **`deltaOnly` / `exportPropertiesOnly`** — no index-less dataStore variant exists in the repo; the same populated `entityIndex.byId` is used either way, so real entities resolve exactly as before. Only genuinely-nonexistent ids are newly filtered.
- **No overlay** — `newPropertySets`/`newQuantitySets` are only populated inside the `mutationView && applyMutations !== false` branch, so the `return false` tail is unreachable when there's no overlay.

Left alone per your call: `deleteEntity`'s forget-vs-tombstone semantics (`restoreNewEntity` pairs with them), and the unified overlay-authority refactor. Changes stay in ~576 and 742-793, clear of #2011's regions.

Happy to write up the five faces as its own issue once these land, so that refactor gets argued on its merits instead of accreting.

BIMvoice:
Pushed `fd05dba9`. A self-review of this PR against its own claim turned up a **third** route to the same dangling reference, and it was live.

`willBeEmitted` is named for *"will this id have a defining line in the output"*, but it only answered "is it tombstoned, and does it resolve as a source or overlay entity." It never consulted `allowedEntityIds`.

Visibility filtering is a third way the line gets dropped, alongside the tombstone and your forgotten-create: the source-entity loop skips anything outside the visible closure, and the new-entities pass does the same. So editing a property on an entity and then hiding it emitted a relation pointing at a line the filter had already removed.

Confirmed before fixing rather than argued from the code — hidden, property-edited door exported with `visibleOnly`:

```
HAS_DOOR_LINE: false
HAS_REL:       true
DANGLING:      [3]
```

**No test in `step-exporter.test.ts` combined `visibleOnly` with a property mutation** — zero occurrences of `visibleOnly` in the whole file. That is why three passes over this code missed it: #1978 named the tombstone, you named the forgotten create, and nothing named the filter. There is a test now; removing the check fails it and only it (252 pass / 1 fail), restored **253 pass / 28 skipped**, typecheck 33/33.

Worth being precise about what this does and doesn't do: it makes the export internally **consistent**, not more complete. An overlay-created entity under `visibleOnly` is still dropped — that's your #2012 instance 4, and it stays yours.

This is now the third face of the same defect found on this one PR, which I think supports your read that the real fix is the unified "effective model" view rather than another predicate. Each round has added a condition to a function whose *name* already promised the general answer; that gap between the name and the implementation is what let each new route through. Happy to write that up as its own issue with all the instances enumerated once these land, if you want it argued somewhere other than a fix PR.

louistrue:
Reviewed. **This is the cleanest PR of the four you have open** — 3 files, +74/−3, changeset in, zero unresolved threads, CI green.

Three things I'd single out:

**You found a path the issue didn't list.** I wrote #1978 naming the two loops built from `getMutations()`; the type-owned pset rewrite is the same defect on a third path, and an entity with no `newPropertySets` entry never reaches the first guard. Fixing only what the issue said would have left it broken — worth noting because that's the failure mode of working from a bug report rather than the code.

**The test pins the class, not the instance.** Asserting that every `#N` referenced in the output has a defining `#N=` line catches any future dangling-reference regression from any path, not just this one. Reusing `findDanglingRefs` rather than string-matching an id is the right call.

**Staying off `getMutations()` was right.** I floated changing its semantics on #1978 and you were correct to decline — it is public API and #1957 is the evidence that consumers depend on the current behaviour. Filtering at the emit sites keeps the blast radius to this file.

## Your open question: confirmed, and worth filing

> `ifc5-exporter.ts` has no `isDeleted` check anywhere … Happy to file it separately if you agree it's real.

It's real. I checked `origin/main`: **zero** occurrences of `isDeleted` or `tombstone` in that file, while it holds a `mutationView` (`:157`, `:177`) and applies mutations at `:481`. So IFC5/IFCX export honours no overlay deletion at all — not just for entities with pset edits, but for any deleted entity. Broader than #1978 and a different shape, so keeping it out of this PR was the right scoping.

Please do file it. Worth checking the other exporters in the same sweep — if `ifc5-exporter` never learned about tombstones, the parquet and glTF paths are worth the same grep before assuming they did.

Nothing blocking from me. Flip it out of draft when you're ready and I'll run the full gate on it.

PR #1998, opened 2026-08-02
fix(examples): stop the Three.js compare demo teaching two fixed fingerprint bugs

louistrue:
> **Merge after #1999.** This PR mirrors the total-order sort fix; #1999 is the package-side original. Landing this first would make the example correct against a package that is not yet.

## What and why

`examples/threejs-viewer/src/compare.ts` carried its own copy of the data fingerprint, and that copy contained **both** bugs fixed in `@ifc-lite/diff` today plus a third nobody had reported:

1. **The assigned type's GlobalId was hashed.** `IfcTypeObject` is an `IfcRoot`, so a from-scratch re-export regenerates it too, moving the fingerprint of every *typed* element on exactly the re-export a comparison is meant to survive.
2. **`localeCompare` in seven places.** Runtime-default collation: `a, A, ä, B` under `en-US`, `a, A, B, ä` under `sv-SE`. Same model, different machine, different hash.
3. **`normalizeValue` used raw `JSON.stringify`** rather than the key-sorted form — a third latent nondeterminism for object-valued properties, found while porting.

An example that demonstrates a bug teaches it.

## Why a copy, not a dependency on `@ifc-lite/diff`

This was checked rather than assumed, and the answer is a hard no *for today*:

- **The example resolves `@ifc-lite/*` from the registry, not the workspace.** Its `package.json` declares `"@ifc-lite/data": "^2.0.0"` and the lockfile records a real tarball, not `link:../../packages/data`. The collab examples use `workspace:*`; these viewer examples deliberately do not, because the README documents a standalone `npm install && npm run dev` that `workspace:` breaks outside the monorepo.
- **Published `@ifc-lite/diff@0.5.0` still contains both bugs.** Unpacked: `dist/fingerprint.js` has `let hash = 2166136261` (32-bit), five `localeCompare` sorts, and `globalId` in the hashed type projection. Today's fixes are unreleased — they sit in `.changeset/wide-hashes-count.md`. Adding the dependency would have re-imported the exact fossil this PR removes.

So the hashing moved to a sibling module `data-fingerprint.ts` whose header names `packages/diff/src/fingerprint.ts` as the real implementation, shows the one-line import a real app should write, and says to swap the copy for it once a fixed release ships.

## Equivalence is verified, not assumed

Identical inputs hashed through `packages/diff/src/fingerprint.ts` and the example's copy: `4e028c81d998b835 == 4e028c81d998b835`, `5ef5939632317e81 == 5ef5939632317e81`, both 16 hex chars. Mutation-checked — swapping `compareCodeUnits` back to `localeCompare` flips the second case to `ff29bbdff4c59b45` and the check exits 1.

## A CI gap this exposed, and what the fix does and does not buy

`examples/threejs-viewer` had **only `dev` and `preview` scripts** — no `typecheck`, no `build`, no `test`. So `pnpm typecheck` never covered it (turbo had no task; every run was a cache hit) and the drift was invisible to CI for as long as it existed. This PR adds `"typecheck": "tsc --noEmit"`, taking the repo from 33 to 34 typecheck tasks.

**Being precise about what that catches:** API and type drift, so the example breaks loudly when a package it copies from changes shape. It would **not** have caught the three bugs above — `localeCompare` is perfectly type-valid. The real guard against semantic copy-drift is the equivalence check above, which is currently a manual script rather than a test, because the example has no test infrastructure. Worth a follow-up if these copies persist.

Deliberately not added: a `build` script, which would make `turbo build` bundle a 2.3 MB example on every CI run.

## Scope notes

- `grep localeCompare examples/` now returns only prose mentions. `babylonjs-viewer`, `threejs-collab` and `collab-demo` carry no fingerprint code.
- The remaining 32-bit FNV in `compare.ts` is `buildGeometryFingerprint`, a sampled geometry-drift hash compared only *within* an already GlobalId-matched pair. Left as is, with a comment stating why it is not identity and must not be reused as such.
- `apps/viewer/src/lib/compare/buildFingerprints.ts` already delegates to the package and is not a copy of the bug.

## Verification

`vite build` exit 0 · example `tsc --noEmit` exit 0 · `pnpm typecheck` 34/34 exit 0 · `docs:check-samples` 255 clean.

Toward #1891.


louistrue:
CodeRabbit's hosted check is rate-limited on this PR, so I ran the CLI locally. **Two MAJOR findings, both valid, both fixed — and the second one turned out to be a bug in the shipped package, not just in this copy.**

## 1. TODO marker on the duplicated implementation

Added, with the three things asked for and a checkable condition rather than a vague one:

```
TODO(remove-by: the first @ifc-lite/diff release that contains the FNV-1a-64
stableHash, {name, type} type projection and code-unit sorts; owner: the #1891
workstream; tracking: issue #1891).
```

plus the mechanical test for "is it time yet": `npm view @ifc-lite/diff version` past 0.5.0 and a `dist/fingerprint.js` free of `2166136261` and `localeCompare`.

## 2. Name-only sorts are not a total order — and the package had it too

The finding is right: `Array.prototype.sort` is stable, so two records sharing a name keep their **input** order, and the sorted result is serialized and hashed. The adapter's walk order reaches the fingerprint.

Rather than fix only this copy, I checked the shipped package. It has the same hole:

```
A 2e607403b1ebbb7e   B 3bc705d1858f727c   ORDER-DEPENDENT
```

(one `IfcWall`, two property sets both named `P`, in the two possible orders — through `@ifc-lite/diff` on `main`)

That is reachable rather than theoretical: a type pset and an occurrence pset of one name are an ordinary IFC arrangement (#1913). The package fix is **#1999**; this PR carries the mirror so the copy stays faithful.

## Equivalence re-verified after the change

The whole point of this file is that it matches the package byte for byte, so that claim was re-checked rather than assumed — both implementations, including the new tiebreak and the duplicate-name case:

```
pkg A 44786f4fd09c9db7 | ex A 44786f4fd09c9db7 MATCH
pkg B 44786f4fd09c9db7 | ex B 44786f4fd09c9db7 MATCH
order-independent: YES
```

Example `tsc --noEmit` exit 0, `vite build` exit 0.

louistrue:
Cursor's "fingerprint sort diverges from package" is **correct, and it catches an imprecision in my own verification** — worth stating plainly rather than resolving quietly.

I claimed this file "matches the package byte for byte" and posted matching hashes. Those hashes were produced against the **#1999 branch**, not against `main`. `#1999` is the package-side fix for the same name-only-sort defect and is **not merged yet**. So as of `main` today, the example's `byNameThenContent` really does diverge from the package's name-only sort — exactly as the finding says.

Two consequences I am accepting rather than arguing away:

1. **This PR should merge after #1999**, not before. Merged in the wrong order, the example would be "correct" against a package that is not yet, which is precisely the drift this PR exists to end. I would rather state the dependency than pretend the ordering is free.
2. **The equivalence claim needs its qualifier.** The accurate statement is: identical output to `@ifc-lite/diff` *once #1999 lands*, verified against that branch —

```
pkg A 44786f4fd09c9db7 | ex A 44786f4fd09c9db7 MATCH
pkg B 44786f4fd09c9db7 | ex B 44786f4fd09c9db7 MATCH
order-independent: YES
```

— and **not** identical to the currently published `0.5.0`, which still carries the 32-bit hash, the `localeCompare` sorts and the GlobalId-in-type-projection. That was always true and is the reason this file is a copy at all; the new tiebreak just adds a fourth difference to the same list.

The alternative — drop the tiebreak here so the copy matches today's `main` — would mean knowingly shipping an example with an order-dependent hash, having just proved the defect is real. Not worth the tidier diff.

Marking #1999 as a merge-order prerequisite.

louistrue:
**#1999 is merged, so the divergence this thread reported no longer exists** — and I re-verified rather than assuming, because my earlier equivalence numbers were measured against #1999's branch, not `main`.

Rebased onto merged `main` and re-ran the check against the **shipped** package, including a case with type assignments and quantity sets:

```
A pkg 44786f4fd09c9db7 | ex 44786f4fd09c9db7 MATCH
B pkg 44786f4fd09c9db7 | ex 44786f4fd09c9db7 MATCH
C pkg 3b141dca2f49aa65 | ex 3b141dca2f49aa65 MATCH
order-independent in both: YES
```

A and B are the same content with two same-named property sets supplied in opposite orders — the defect that started this. The example and `main`'s `@ifc-lite/diff` now agree byte for byte, and both are order-independent.

The merge-order prerequisite is discharged.

PR #2000, opened 2026-08-02
feat(mcp): put model_diff on the real diff engine, opt-in (#1891)

louistrue:
## What and why

`packages/mcp/src/tools/diff.ts` was a naive GlobalId **set intersection** that predated `@ifc-lite/diff` and never imported it. So for an agent, a model re-exported from scratch read as *the entire model deleted and added* — the exact pathology #1891 exists to fix, on the surface least able to notice a wrong answer and most likely to act on it unsupervised.

## The numbers, end-to-end on real models

Every `IfcRoot` GlobalId rewritten, then diffed through the real MCP transport:

| model | `by_content` OFF | `by_content` ON |
|---|---|---|
| **AC20-FZK-Haus** (1322 re-GUIDed) | `+1304 / -1304`, common 21 | `{added: 29, deleted: 29}` · `{renamed: 117, ambiguous: 4}` |
| **Building-Architecture** (119 re-GUIDed) | `+118 / -118`, common 14 | `{added: 0, deleted: 0}` · `{renamed: 40}` |

The 4 ambiguous groups are honest, not a shortfall: 14 identical `IfcAnnotation`s, 3 `IfcDoorType`s, 2 `IfcWindowType`s — data-identical with no geometry hash to separate them, since Node has no mesh pipeline. Their members are exactly the 29 that stay added/deleted. 146 ms for the content pass on 2.5 MB.

## API

`by_content: boolean` (default **false**) on `model_diff`, not a sibling tool: a sibling would duplicate two-model resolution and the allowlist guard, and force an LLM to choose between near-identically-named tools. `typeDiffs`/`entityDiff` are byte-for-byte unchanged when it is off.

Adds `contentDiff: { scope, counts, contentMatchCounts, contentMatches[], truncatedMatches }`. Groups stay groups — `contentMatches` entries carry `base: [keys]` and `head: [keys]` rather than flattening, because collapsing "we could not tell" into a scalar is the dishonest option in a headless contract.

`max_matches` (default 200) exists because an MCP result lands in a context window. Two deliberate details: unresolved kinds sort **first**, and `contentMatchCounts` is computed **before** the cap — so truncation can never be what makes a model look cleanly matched.

Runs at `scope: 'data'` and echoes the scope back. Supplies `components`, so the collision guard is live.

## Mutation checks: 12 run, 2 survived, both were bad tests

- **Removing the unresolved-first sort survived.** The engine emits in bucket order and the fixture happened to write the twins first, so `ambiguous` came first naturally — the assertion pinned nothing. Fixture rewritten so both orderings agree that ambiguous is *naturally last*; the mutant now fails.
- **Dropping the `dependent`-role skip survived.** Its observable effect on this store is that `IfcMaterial` enters keyed on its Name (`brick`); the test only checked content *matches*, and `brick` matched by key so it produced none. Added a direct assertion on the exact key set; the mutant now fails.
- A third weakness surfaced the same way — dropping `components` survived — so the `attr:core` assertion the CLI suite has was added too.

## One judgement call for your review

The hand-seeded `apps/viewer/src/generated/mcp-catalog.json` would otherwise advertise `by_content` to the browser playground, whose dispatcher is a *separate* implementation that ignores it — silently returning a GlobalId intersection labelled as a content diff. `playground-dispatcher.ts` now throws `UNSUPPORTED_OPERATION` instead. That is a 9-line touch in `apps/viewer`; if you would rather this PR stayed out of the viewer entirely, reverting that hunk plus the catalog description line is clean and independent.

## Code path

`packages/mcp/src/tools/diff-fingerprints.ts` is a second copy of the CLI's `diff-engine.ts`, so it carries the chain-checked extractor (`IfcTask`/`IfcActor` in via the STEP record; `IfcMaterial`, psets and rels out). Duplicated rather than moved, justified in the file header: `@ifc-lite/cli` already depends on `@ifc-lite/mcp` (cycle), `@ifc-lite/diff` is deliberately dependency-free, `@ifc-lite/parser` must not learn about diffing, and a concurrent PR in this program is live in `packages/cli`. Both copies pin the shared contract by test.

## Verification

`@ifc-lite/mcp` 87 tests · typecheck 33/33 · docs-samples 255 clean · generated regions up to date · api-surface matches · readmes clean. Docs updated in-PR: `docs/guide/mcp.md`, `docs/guide/model-diff.md` (new MCP section), `packages/mcp/README.md`, the skill's tool list. Changeset: `@ifc-lite/mcp` minor.

Toward #1891.


louistrue:
Both findings confirmed and fixed. **The P1 was real, wider than reported, and sitting next to a data-loss bug.**

## 1. Content diff ignores pending mutations [Greptile P1] — confirmed

Reproduced through the real tools before touching anything: same file as base and head, then on head rename a wall, set a property, create a wall, delete a storey, then `model_diff by_content`:

```
entityDiff: {added: [], removed: [], common: 4}
contentDiff.counts: {added: 0, modified: 0, deleted: 0, unchanged: 4}
```

Four queued edits, "nothing changed".

Mechanism: `MutablePropertyView` is an overlay; the parsed store's buffer and index are never mutated, and the overlay only materialises in `StepExporter`. **All three passes were blind, not just the new one** — so this PR widened the blast radius rather than introducing it, which is exactly what the finding said.

**Scope: fixed all three passes of `model_diff`, filed the rest.** Fixing only `by_content` would emit a self-contradicting payload (`contentDiff.added: 1` beside `entityDiff.added: []`). Rewriting the whole MCP read surface is a server-wide change. `model_diff` is the one tool where a wrong "nothing changed" is unrecoverable, so it folds the overlay and reports `contentDiff.pendingMutations`, making its divergence from `entity_get` explicit rather than silent.

## The adjacent bug: editing one property dropped its siblings on save

While wiring the overlay I found `packages/mcp/src/headless-backend.ts` built the overlay with `store.properties` as its base — which **the columnar parser leaves empty** (verified: `store.properties.getForEntity(72)` returns `[]` while `extractPropertiesOnDemand` returns both properties).

`StepExporter` re-emits `getForEntity(id)` for every entity carrying a property mutation and skips the originals. So the view's only source was the overlay itself, and **editing one property silently dropped every sibling property in that pset on save.** Data loss in the shipped MCP server, unrelated to diffing.

Fixed by wiring the parser's on-demand extractors, mirroring what `apps/viewer/src/utils/configureMutationView.ts` already does. Pinned by a test.

## 2. `max_matches` bounded groups, not members [Codex P2] — confirmed

`max_group_members` (default **20**) caps each side; every match carries `baseCount`/`headCount` and `baseTruncated`/`headTruncated`, computed **before** either cap, so truncation can never make a model look cleanly matched. 20 because a group is the engine *declining to guess* — its members are a hand-off, and 20 a side is already past what anyone resolves by reading rather than by querying the models. The common 1:1 `renamed` pair is untouched.

## End-to-end still holds

`+1304/-1304` by GlobalId → `{added: 29, deleted: 29}` with `{renamed: 117, ambiguous: 4}`.

One honest note: my first re-run gave 111/111 and 35 renamed. That was **my re-GUID script**, which was rewriting every 22-character literal including tags; anchoring it to the first STEP slot reproduced the documented numbers. Not a regression, but worth recording since I nearly reported it as one.

With mutations queued: `pendingMutations: {base: 0, head: 2}` and a summary suffix. With `max_group_members: 5`, the 14-member ambiguous group reports `listed=5, baseCount=14, baseTruncated=true`, counts unchanged.

## Verification

13 mutations, every one killed a test. M13 exists **because** the 1:1 "not truncated" assertion survived M10 and M11 — a test that was passing for the wrong reason until a targeted mutant exposed it.

`@ifc-lite/mcp` 96 tests · `pnpm typecheck --force` 33/33 (uncached) · docs-samples 255 clean · generated regions clean · `pnpm build --force` 40/40 · api-surface matches.

## Follow-ups, deliberately not in this PR

1. The rest of the MCP read surface (`entity_get`, `entity_query`, `model_info`) is still blind to the overlay.
2. `apps/viewer/src/components/mcp/playground-dispatcher.ts` is a separate browser reimplementation of `model_diff` with the same staleness against the viewer's own overlay.

louistrue:
Fixed in `9966e2e2`. **The finding was right, the trap was worse than described, and the most useful thing it exposed was a false claim in my own comment.**

## The fix

`classifyType` now reads `getInheritanceChainAcrossSchemas` (IFC2X3 + IFC4 + IFC4X3) and picks the leaf **by name**, matching #2001.

**The order trap is worse than "opposite order", and I measured it rather than trusting the warning:** `getInheritanceChain` *falls back* to the pin for vendor classes, so **one function returns both orders depending on its input**. Under positional indexing, **717 of the 776 pinned classes** would get the wrong leaf. The comment in the file now says this, because "they're in opposite order" would have been a misleading thing to leave behind.

## IFC4 equivalence

All **776** classes in `SCHEMA_REGISTRY`: **0 role differences, 0 leaf-name differences**. No IFC4 file changes behaviour. The gap the pin leaves, confirmed independently: **23 IFC2X3** and **77 IFC4X3** `IfcObjectDefinition` classes, plus 4 post-ADD2 IFC4 ones (`IfcAlignment`, `IfcReferent`, `IfcPositioningElement`, `IfcLinearPositioningElement`).

## The real finding: my header comment was false

The file's header claimed "a drift in either copy fails a build". **It did not and could not.** The two copies had *parallel* suites asserting the same behaviour separately, so fixing one and not the other passes both — which is **exactly how this shipped**. A comment promising a safety net that does not exist is worse than no comment, because it stops the next person looking.

Deleted, and replaced with something that can actually fail: a parity test that runs **both copies over the same store** across three fixtures and requires identical key sets and identical `ifcType` per key. Mutation-verified to fire in **both** directions — injecting drift into the *CLI* copy (`IFCACTOR` → empty chain) fails it too, not just changes on the MCP side.

## Duplication: kept, but now detectable

I did not end it. The cycle argument still holds (`cli → mcp`; `@ifc-lite/diff` is dependency-free; the parser must not host diff policy). Two cheap homes rejected explicitly: hosting the walk in `@ifc-lite/mcp` and importing it from the CLI is legal by dependency direction but puts core diff scope in a server package as published API; a `@ifc-lite/diff/scope` subpath gives that package a runtime parser dependency. The only honest home is a fourth published package — a release decision, not a review fix.

## One thing for your judgment

The parity test imports the CLI copy by **relative path** (`../../../cli/src/commands/diff-scope.js`). There is no precedent for cross-package relative imports under `packages/*/src`. It is confined to a test file (`packages/mcp/tsconfig.json` excludes `**/*.test.ts`, so it is neither typechecked nor built), and it is the only way to compare the copies without adding a dependency edge.

If you would rather not have that reach at all, the alternative is the fourth package. I chose detectable-drift over no-drift-detection, but it is your call and I would rather you saw it than found it.

## Verification

`@ifc-lite/mcp` 102 tests (14 files) · typecheck 34/34 · docs + generated regions clean · build 40/40 · api-surface matches · `check-test-wiring` clean · `turbo test --filter=@ifc-lite/cli` 244 passed (CLI untouched) · `turbo test --filter=@ifc-lite/mcp` via the CI invocation path, and `turbo`'s `test` task is `cache: false` so the cross-package test cannot be skipped by a stale cache.

End-to-end unchanged, as the IFC4 equivalence predicts: `+1304/-1304` → `{added: 29, deleted: 29}` with `{renamed: 117, ambiguous: 4}` through the real stdio transport.

louistrue:
Fixed in `c200e6fe`. Reproduced first through the real tools: create a wall, `entity_set_attribute` on it, then `by_content` — the new values never reached the hash.

## What the field audit found

The finding named `Name` and `Description`; checking the full set `buildDataInput` reads turned up one worse case:

- **Name, Description** — stale. Now `override(edited.x, payload.x)`, the same helper the parsed path uses, so a clear (`value: ''`) actually clears rather than falling back to the creation payload.
- **ObjectType** — **absent entirely**, not merely stale. It has no readable payload slot (slot 4 is `ApplicableOccurrence` on an `IfcTypeObject`), so it exists *only* as an override — meaning an ObjectType set after creation could never reach the hash under any circumstances. Now included.
- **Tag** — the fourth attribute `entity_set_attribute` accepts; no fingerprint field carries it, correctly ignored.
- **PredefinedType, typeAssignments** — stay absent. Both are store reads and neither is reachable through `entity_set_attribute` (retype is not on the MCP surface). Documented, and the failure direction is the safe one: reports as added rather than mis-matched.
- **Property sets and quantities** — **no asymmetry**. Both already went through `overlay.propertySets()` / `overlay.quantitySets()`, so `entity_set_property` on a created entity already landed in its hash. Verified by reading the path rather than assuming it matched.

## A second data-loss bug on save, found and NOT fixed here

`export_ifc` writes an overlay-created entity from its authored payload alone: `StepExporter` applies `getAttributeMutationsByEntity()` only to source-buffer records (`packages/export/src/step-exporter.ts` ~768-806). **So a create-then-rename is dropped on save.** Probed directly — after a `Name → 'Wall C'` edit the exported line was:

```
#73=IFCWALL('2aaa…',$,'untitled',…)
```

Pre-existing, different package, needs its own changeset. Filed rather than folded in.

That is the **second** data-loss-on-save bug this PR has surfaced without looking for either: the first was the overlay built on an empty `store.properties` base, which dropped every sibling property in an edited pset.

## The other Cursor comment is stale

`classifyType` was already switched to `getInheritanceChainAcrossSchemas` in `9966e2e2`, with the leaf picked by name and a cross-copy parity suite running both implementations over the CLI's own fixtures. `getInheritanceChainForEntity` appears nowhere in the file except in prose explaining why it is not used. Nothing changed for it.

## Verification

Both new tests confirmed **red on the pre-fix code** before the fix landed (`expected {} to deeply equal { renamed: 1 }`). Four mutations, no survivors, including a re-run of the primary mutation against the final committed file.

`@ifc-lite/mcp` 104 tests · typecheck 34/34 · docs + generated regions clean · build 40/40 · api-surface matches. End-to-end unchanged: `+1304/-1304` → `{added: 29, deleted: 29}` with `{renamed: 117, ambiguous: 4}`.

One note: `diff-fingerprints.ts` is now 403 lines against the ~400 guideline, roughly 130 of which are header rationale. Splitting it would separate the rationale from the code it explains; say the word if you would rather it were split anyway.

PR #2001, opened 2026-08-02
fix(cli): compare object identities in --by-entity, not positional Name slots (#1891)

louistrue:
## What and why

`--by-entity` compared GlobalIds read from the `EntityTable`. The parser fills that table's GlobalId column **positionally**, and slot 0 of a resource entity is a **Name** — so the flag was silently comparing materials and surface styles under their names, and colliding them with each other.

Measured before changing anything, replaying the old loop against the chain-checked walk:

| model | old keys | dup keys | non-`IfcRoot` | new keys | dup | non-`IfcRoot` |
|---|---|---|---|---|---|---|
| `hello-wall.ifc` | 39 | 4 | 18 | 12 | 0 | 0 |
| `building-architecture.ifc` | 132 | 8 | 23 | 40 | 0 | 0 |
| `building-architecture-rev-b.ifc` | 133 | 9 | 25 | 40 | 0 | 0 |
| `infra-bridge.ifc` | 232 | 7 | 16 | 96 | 0 | 0 |
| `ara3d/dental_clinic.ifc` (209k) | 41,100 | 12 | 44 | 3,780 | 0 | 0 |

The non-`IfcRoot` entries were `IfcMaterial`, `IfcSurfaceStyle`, `IfcClassification`, `IfcProjectedCRS` and geometric subcontexts.

## Two corrections to the premise I was given

**The "silent drop-out" claim does not hold for this flag.** It is true of the pre-fix `--by-content` path (which read `store.entities.getGlobalId` only), but `--by-entity` went through `EntityNode.globalId`, which falls back to `extractRootAttributesFromEntity` on the STEP record. Measured on `856_wall_decode_failed.ifc`: `IFCTASK` count 8, table GlobalIds 0, `EntityNode` GlobalIds **8**. Tasks and work schedules were already compared; "gained keys" was 0 on every model probed. The fix keeps them, and the test says explicitly that this is not a regression pin on the old flag.

**A third defect, larger than either, surfaced from the probe: scope.** Relationships and property sets dominated the old key set (14,701 `IfcRelDefinesByProperties` + 14,432 `IfcPropertySet` on the 209k model). On a real re-export pair:

```
BEFORE                          AFTER
Added:   183                    Added:   0
Removed: 179                    Removed: 0
```

Every one of those 362 was a re-GUIDed pset or relationship. **Not one object identity had changed.** Anyone reading that output as "183 things were added" was reading noise.

## The fix

Membership now comes from the schema registry inheritance chain, shared with `--by-content`: every `IfcObjectDefinition` in, `IfcRelationship` and `IfcPropertyDefinition` out by rule, classified once per *type* so geometry buckets are dismissed without touching a row.

`classifyType`, the per-type dismissal, the source read and the `ifcType` resolution moved verbatim out of `diff-engine.ts` into a shared `diff-scope.ts` (174 lines) with no logic change — `diff-engine.ts` drops 263 → 145 lines and its 8 existing tests pass untouched.

The vendor gap is inherited, not widened: `IFCVENDORTASK` stays uncompared, `IFCVENDORWALLTYPE` stays compared, pinned with reasoning — including the note that this is *narrower* than the old flag, which saw vendor roots via the positional slot-0 fallback at the cost of reading a Name for every vendor non-root.

## Verification

`@ifc-lite/cli` 241 tests (19 files) · typecheck 33/33 · docs-samples 255 clean · generated regions up to date · api-surface matches · `tsc --noEmit` on the package directly.

That last one matters: **`pnpm typecheck` does not cover `packages/cli`** — it has no `typecheck` script, so the package is only typechecked by its build. Same class of gap as #1998 found in the examples. Flagging rather than fixing here, since it belongs in its own change.

8 mutations, all killed, every one of the 8 new tests killed by at least one — including reverting `--by-entity` to the old `EntityNode` walk.

Changeset (`@ifc-lite/cli` minor) states plainly that **the reported numbers change on every model**, with before/after counts, and points anyone who read added/removed as a "pset appeared" proxy at the type-difference table and `--by-content`.

Toward #1891.

### DIFF
--- a/.changeset/by-entity-shared-scope.md
+++ b/.changeset/by-entity-shared-scope.md
@@ -1,0 +1,18 @@
+---
+'@ifc-lite/cli': minor
+---
+
+**diff**: `ifc-lite diff --by-entity` now compares the same entities as `--by-content` — every `IfcObjectDefinition`, decided from the schema inheritance chain — instead of asking every row of the entity index for a GlobalId (issue [#1891](https://github.com/LTplus-AG/ifc-lite/issues/1891)).
+
+**The reported numbers change, on every model.** `Common`, `Added` and `Removed` are now counts of objects, so they get much smaller: on the bundled sample models the key set goes 132 → 40 (`building-architecture.ifc`), 133 → 40 (`-rev-b`), 232 → 96 (`infra-bridge.ifc`) and 39 → 12 (`hello-wall.ifc`); on a 209k-entity model it goes 41,100 → 3,780. What left the count is entities whose identity was never their own:
+
+- **Relationships and property sets.** An `IfcRelDefinesByProperties` is identified by its endpoints, and a property set's contents already travel with the element that owns it, so counting them reported every edited property twice and turned a re-GUIDed relationship into churn. They are the large majority of the old key set — 14,701 `IfcRelDefinesByProperties` and 14,432 `IfcPropertySet` on the 209k model alone — and the churn is not hypothetical: on a real re-export pair the flag reported 183 added and 179 removed entities where **every single one** was a property set or a relationship the exporter had re-GUIDed. It now reports 118 common, 0 added, 0 removed, which is what happened.
+- **Entities keyed by their Name.** The columnar parser fills its GlobalId column positionally, and slot 0 of an `IfcMaterial`, `IfcSurfaceStyle`, `IfcClassification` or `IfcProjectedCRS` is a *Name*. Those entities were compared under that name — and two of them sharing a name collided into one key, so they were compared as a single entity. Every sample model had collisions: 8 (`building-architecture.ifc`), 9 (`-rev-b`), 7 (`infra-bridge.ifc`), 4 (`hello-wall.ifc`), and 12 on the 209k model.
+
+The chain is read across every bundled schema (IFC2X3 + IFC4 + IFC4X3), not from the parser's IFC4 codegen pin, which matters most on IFC2X3. IFC4 dropped 23 `IfcObjectDefinition` classes that IFC2X3 files still carry — `IfcMove`, `IfcOrderAction`, `IfcScheduleTimeControl`, `IfcSpaceProgram`, `IfcServiceLife`, `IfcTimeSeriesSchedule`, … — and the pin alone has nothing to say about any of them: the ones the parser's entity table does not hold would have gone uncompared even though their STEP records carry a GlobalId, and the IFC2X3-only *resource* classes it does hold (an `IfcSymbolStyle`, taken in because the name ends in `STYLE`) would still have been keyed on the Name in slot 0. The bundled sample models are unaffected — 132 → 40, 133 → 40, 232 → 96 and 39 → 12 as above, and 118 common / 0 added / 0 removed on the re-export pair.
+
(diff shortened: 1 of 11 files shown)


louistrue:
**The finding held, and it was worse than reported.** Fixed in `9f49e290`.

## Evidence

Two lookups exist, and only one knows about IFC2X3:

- `getInheritanceChainForEntity` — generated from the **IFC4_ADD2_TC1 pin only**, returns `[]` for anything else
- `getInheritanceChain` — walks the union of **IFC2X3 + IFC4 + IFC4X3**, and was **not exported** from `@ifc-lite/parser`

I enumerated the gap rather than trusting the example: the pin misses **23 IFC2X3 `IfcObjectDefinition` classes** (`IfcMove`, `IfcOrderAction`, `IfcScheduleTimeControl`, `IfcSpaceProgram`, `IfcServiceLife`, …) and **77 IFC4X3 ones**. `IfcMove` was a correct example.

Probe on a constructed IFC2X3 file, old walk vs this branch:

```
DROPPED: 0MOVE…, 0RELC…, 0SPGM…, 0STCL…
```

Three real IFC2X3 objects the old code compared, gone.

**And the second half nobody named:** the branch still carried `hatch` — the *Name* of an `IfcSymbolStyle`. That IFC2X3-only resource class ends in `STYLE`, so the parser's name-based branch puts it in the `EntityTable`, and with no chain to disprove it the `unknown` verdict let it through. **The exact defect this PR exists to fix, still live on IFC2X3.**

Real CLI on a re-GUIDed IFC2X3 pair:

| | added | removed | common |
|---|---|---|---|
| before | `[]` | `[]` | 5 (incl. `hatch`) |
| after | `["NEWM…"]` | `["OLDM…"]` | 5 (no `hatch`) |

## A near-miss worth recording

The two chain functions return their chains in **opposite order** (root→leaf vs leaf→root). A naive swap would have made `name` = `'IfcRoot'` for every source-read entity and quietly poisoned the content fingerprint. The leaf is now picked **by name**, and the equivalence was verified across all **776 pinned classes: 0 role differences, 0 leaf-name differences** — so no IFC4 file changes behaviour.

## No PR-body number moved

| model | old walk | branch HEAD | with fix |
|---|---|---|---|
| building-architecture.ifc | 132 | 40 | **40** |
| infra-bridge.ifc | 232 | 96 | **96** |
| re-export pair | 309 / 313 | 118 | **118** |

Stated honestly: no downloaded fixture contains any of these classes, so the IFC2X3 evidence is a constructed file rather than a corpus model.

## Also

`getInheritanceChainAcrossSchemas` is now exported from `@ifc-lite/parser` (changeset added, api-surface 3981 → 3982), the counterpart of the existing `getAttributeNamesAcrossSchemas`. The docs paragraphs that already *claimed* the chain spans all three schemas are now true.

4 mutations, none survived. Each test carries a precondition assertion on the parser's own state (`getGlobalId(IFCMOVE) === ''`, `getGlobalId(IFCSYMBOLSTYLE) === 'hatch'`) so it fails loudly rather than passing vacuously if the parser changes.

`@ifc-lite/cli` 244 tests · `@ifc-lite/parser` 324 · typecheck 33/33 · CLI `tsc --noEmit` clean · docs + generated regions clean · api-surface matches.

Issue #1470, opened 2026-06-30
Opening reported when Compare models


BIMvoice:
When a window is removed in the latest version of a model, I do not think the related openings should be reported as changes.

The opening is only the connection between the wall and the window. It is not a meaningful change for the end user by itself.

In this case, the relevant change is that the window was removed.

sample models attached

[Project01_PC01_edit_v2.ifc.txt](https://github.com/user-attachments/files/29515236/Project01_PC01_edit_v2.ifc.txt)
[Project01_PC01_edit_v1.ifc.txt](https://github.com/user-attachments/files/29515235/Project01_PC01_edit_v1.ifc.txt)

<img width="1432" height="722" alt="Image" src="https://github.com/user-attachments/assets/972e67cb-d406-458e-ba86-d7fe042391dc" />

louistrue:
Thanks Petru, appreciate it. I can't just decide for the users what is relevant and what not. Some might care about openings. What we could add is a sort of blacklist where you add the classes you don't want considered. Thoughts?

BIMvoice:
That's definitely not a bad idea actually.

PR #2002, opened 2026-08-02
feat(viewer): render content matches in Compare, on by default (#1891)

louistrue:
## What and why

Before this PR, `grep -rn "matchUnpairedByContent\|contentMatches" apps/` returned **zero hits**. The tiered content matcher merged in #1989 was exercised only by its own unit tests — no shipped surface reached it. This makes it reachable, and is therefore also its **first field test on real geometry**.

Default **ON**, with a visible toggle (*"Match re-exported elements by content"*) persisted like the ifcType blacklist. Only an explicit opt-out is remembered, so the default can move later without stale preferences pinning users to the old behaviour.

## Real-model verification, executed rather than asserted

`AC20-FZK-Haus.ifc` as A; B = the same file with every IfcRoot GlobalId regenerated (1322 rewrites), one wall translated +1 m, one property edited. Loaded as a federation in the running viewer, clicked **Run comparison**, drove the toggle, clicked a section header and a row, downloaded the CSV through the UI.

| | matching ON | matching OFF |
|---|---|---|
| added / deleted | **7 / 7** | 119 / 119 |
| content matches | 111 `renamed` + 1 `moved` | none |
| meshes hidden | 112 | 0 |
| meshes coloured | 126 (7 red, 7 green, **112 blue**) | 238 |
| **stranded** | **0** | **0** |
| CSV rows | 126 | 238 |

**The stranded-mesh risk was measured, not assumed.** With the overlay's content-match loop disabled, the same run leaves **224 of 238 meshed elements stranded at full material colour** in a ghosted scene. Retiring a pair removes its entries, and `overlay.ts` derived everything from `diff.entries` — so this is the exact failure `excludedHiddenIds` exists to prevent for #1470, and the counterfactual proves the new channel is load-bearing rather than decorative.

**An accidental control that held.** The 7 unmatched pairs are all `IfcSpace`, and they are *correct*: my fixture generator rewrote every 22-character GUID-shaped first string attribute, and `'SpaceTemperatureSummer'` / `'SpaceTemperatureWinter'` are exactly 22 characters in the IFC base64 alphabet — so those property **names** genuinely changed. The pass refused to pair 7 elements whose content really differs while pairing 112 that did not.

The single `moved` row is the wall I actually translated, with `MovedDistance_m` **blank** — the documented no-`aabb` fallback, which PR 2 fills in.

## The four surfaces

- **Overlay** — new `matched` channel: retiring kinds hide every A copy and colour B blue. Review kinds (ambiguous/duplicated/deduplicated) retire nothing, so they keep their add/delete colours and get badged instead.
- **Results list** — new **Matched** and **Needs review** groups from `diff.contentMatches`, with `match:<n>` row keys that cannot collide with a `DiffEntry` key.
- **Report** — one row per head entity of a retiring match, with `MatchedGlobalId` when 1:1 and **blank for a group**, because the engine has no bijection to claim there. Unresolved groups annotate their existing add/delete rows via a new `Match` column rather than duplicating a GlobalId. `counts` gains `matched` + `needsReview`; new CSV columns are appended.
- **Telemetry** — `model_compare_run` gains `content_matching` and per-kind counts. This is the default-on rollout's evidence.

`components` is now supplied, moving the viewer out of the weakest row of the collision table (previously `dataHash` only, so the guard protecting destructive retirement was inert).

## Mutation checks: 21 applied, 21 killed

**One survived first and was a masking fixture, not a redundant guard.** M17 (default OFF) passed because the test ran with no `window`, so `loadPersistedMatchByContent` early-returned at its `typeof window === 'undefined'` guard and the default branch was never reached — a test proving nothing about the headline default. Rewrote the suite with a fake `window.localStorage` so the real branch is under test, and added three more covering load/persist/error paths.

## Verification

viewer suite 2285 pass / 0 fail · typecheck 33/33 · docs-samples 255 clean · viewer build exit 0 · test-wiring clean. 0 console errors across every run.

`pnpm knip` exits 1 on unmodified `main` too; this branch reduces unused exports 363 → 362 and adds none.

## Known gaps, deliberate

A content-match row has no "what changed" detail and no BCF pre-fill — it is not a `DiffEntry`, and its key is namespaced so it can never resolve against `diff.byKey`. `useCompare.ts` has no unit test in this repo and still doesn't; the toggle wiring and re-diff guard are covered by the real-model drive rather than a killable unit test. A hook test is the follow-up if you want it locked down.

Toward #1891.

## Summary by CodeRabbit

* **New Features**
  * Added content-based matching to model comparisons, enabled by default with a persistent opt-out setting.
  * Added Matched and Needs review result sections with focus actions, match details, counts, and overflow indicators.
  * Added matched-item visualization with dedicated coloring.
  * Comparison reports and CSV exports now include match statuses, matched IDs, and review counts.

* **Documentation**
  * Updated Compare mode guidance with matching behavior, visualization, and export details.

### DIFF
--- a/apps/viewer/src/components/viewer/ComparePanel.tsx
+++ b/apps/viewer/src/components/viewer/ComparePanel.tsx
@@ -10,7 +10,7 @@
  */
 
 import { useEffect, useMemo } from 'react';
-import { GitCompareArrows, Loader2, Play, X, Trash2, Download, ChevronLeft } from 'lucide-react';
+import { GitCompareArrows, X, Trash2, Download, ChevronLeft } from 'lucide-react';
 import { Button } from '@/components/ui/button';
 import { cn } from '@/lib/utils';
 import { tourAnchor, TOUR_ANCHORS } from '@/lib/tours/anchors';
@@ -26,22 +26,14 @@ import { ChangeDetailView } from './compare/ChangeDetailView';
 import { BcfFromChange } from './compare/BcfFromChange';
 import { useBcfFromChange } from './compare/useBcfFromChange';
 import { CompareResultsList, CountBadge, LISTED_STATES, type CompareBucket } from './compare/CompareResultsList';
-import { CompareBlacklist } from './compare/CompareBlacklist';
-import { changedTypeCounts, type CompareRow } from './compare/changeRow';
-import type { DiffScope, DiffState, DiffEntry } from '@ifc-lite/diff';
+import { CompareRunControls } from './compare/CompareRunControls';
+import { changedTypeCounts, contentMatchRows, hasReportableChanges, MAX_ROWS_PER_GROUP, type CompareMatchRow, type CompareRow } from './compare/changeRow';
+import { contentMatchCounts, contentMatchingRan } from '@/lib/compare/contentMatches';
+import type { DiffState, DiffEntry } from '@ifc-lite/diff';
 
 interface ComparePanelProps {
   onClose?: () => void;
 }
-
-const SCOPES: { id: DiffScope; label: string }[] = [
-  { id: 'both', label: 'Both' },
-  { id: 'data', label: 'Data' },
-  { id: 'geometry', label: 'Geometry' },
-];
-
-/** Cap rows rendered per group so a huge diff can't stall the DOM. */
-const MAX_ROWS_PER_GROUP = 1000;
 
 /** The side actually drawn for an entry: base for deletions, head otherwise. */
 function renderRef(entry: DiffEntry<CompareRef>): CompareRef | undefined {
@@ -56,12 +48,14 @@ export function ComparePanel({ onClose }: ComparePanelProps) {
   const headModelId = useViewerStore((s) => s.compareHeadModelId);
   const scope = useViewerStore((s) => s.compareScope);
   const showUnchanged = useViewerStore((s) => s.compareShowUnchanged);
+  const matchByContent = useViewerStore((s) => s.compareMatchByContent);
   const excludedTypes = useViewerStore((s) => s.compareExcludedTypes);
   const selectedKey = useViewerStore((s) => s.compareSelectedKey);
   const setBaseModelId = useViewerStore((s) => s.setCompareBaseModelId);
   const setHeadModelId = useViewerStore((s) => s.setCompareHeadModelId);
   const setScope = useViewerStore((s) => s.setCompareScope);
   const setShowUnchanged = useViewerStore((s) => s.setCompareShowUnchanged);
+  const setMatchByContent = useViewerStore((s) => s.setCompareMatchByContent);
   const addExcludedType = useViewerStore((s) => s.addCompareExcludedType);
   const removeExcludedType = useViewerStore((s) => s.removeCompareExcludedType);
   const clearExcludedTypes = useViewerStore((s) => s.clearCompareExcludedTypes);
@@ -123,6 +117,18 @@ export function ComparePanel({ onClose }: ComparePanelProps) {
     return out;
   }, [result, models]);
 
+  // Content-match rows (#1891). They live on `diff.contentMatches`, NOT in
+  // `diff.entries` - a retiring match removed its entries outright - so this is
+  // separate plumbing rather than another bucket of `groups`.
+  const matchRows = useMemo<CompareMatchRow[]>(
+    () =>
+      contentMatchRows(result?.diff.contentMatches, (ref) =>
+        models.get(ref.modelId)?.ifcDataStore?.entities.getName(ref.localId) || '',
+      ),
+    [result, models],
+  );
+  const matchCounts = useMemo(() => contentMatchCounts(result?.diff.contentMatches), [result]);
+
   const counts = result?.diff.counts;
   const canRun = !!baseModelId && !!headModelId && baseModelId !== headModelId && !running;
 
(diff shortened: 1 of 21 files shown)


louistrue:
All three fixed in `ee227864`. **Both P1s confirmed on a real model, not just reasoned about.**

## Findings 1 + 3 are the same bug — verified

`runComparison` read the options before its awaits (the frame yield, plus `buildEntityFingerprints` which itself yields every 1500 entities), then published from those captured values. The reconciliation effect cannot rescue it: the option change already fired the effect against a null-or-older result, and the stale result landing afterwards re-runs nothing.

**Confirmed by mutating the live dev server**, not by argument: unchecking the box while `compareRunning === true` published **7/7 with 112 matches**, a Matched section in the DOM, 112 elements hidden by the match channel and a 126-row CSV — the entire panel disagreeing with its own checkbox.

**Fix: read the options at publish time**, after every await. Chosen because it is the only option with no wrong intermediate state:

- adding `result` to the deps publishes the stale result *then* patches it — a second full diff, one visibly wrong frame, and it arms the exact `setCompareResult` → re-render → re-diff cycle the missing dep exists to prevent
- disabling the controls during a run trades a correctness bug for a UI restriction

The publish is atomic (no `await` between the `getState()` read and `setCompareResult`, so nothing can interleave), and because the result always lands agreeing with the store, the effect stays a no-op — which is what keeps the `result` omission safe. That reasoning is in the code, not just here.

**Scope and blacklist share the hole**, latent since #924 — those controls are not disabled during a run either. The content checkbox is simply the first one users reach for mid-run. The publish-time read closes all three; each has its own regression test.

## Finding 2 — verified

Checked the established pattern first: Compare uses a **1000-row cap**, not virtualization (virtualization exists in HierarchyPanel/ClashPanel/SearchModal but never here). Applied the same cap, display-only — header count and select-all still cover the full set, overflow uses the existing `+N more not shown` wording. `MAX_ROWS_PER_GROUP` moved to `changeRow.ts` so both consumers share one constant.

Verified in the real UI with the cap temporarily lowered to 50: 50 rows rendered plus `+62 more not shown`, header still `Matched (112)`, clicking the header still selected **112** entities in 3D.

## Two verification notes worth reading

**A mutation survived, and it was correct that it did.** M2 (adding `result` to the deps) left every test green. Investigated rather than assumed: once the publish is correct, M2 is genuinely behaviour-preserving — it is the *rejected alternative*, not a defect. M3 (M1+M2 together, i.e. actually shipping that alternative) fails test 5. The test comment now names which alternative it rejects.

**A flaky test was caught and fixed before it shipped.** The first version used `await act(async …)` to start the run, which sometimes let the `setTimeout(0)` fire and complete the whole run before the toggle — 2 of 6 runs failed. Switched to synchronous `act` with no await between start and mutate: 6/6 deterministic, plus an in-flight assertion that would catch a regression of the harness itself.

## Real-model re-verification

Revision B: all **1322 IfcRoots re-GUIDed, zero GlobalId overlap**, one wall translated +2.5 m.

| | matching OFF | matching ON |
|---|---|---|
| counts | 119 / 119 | **7 / 7** |
| matches | – | **112** (111 Renamed + 1 Moved) |
| overlay | 238 coloured, 0 stranded | 126 coloured + 112 hidden, **0 stranded** |
| CSV | 238 rows | 126 rows |

Post-fix mid-run toggle in both directions: everything agrees with the checkbox, 0 stranded either way.

## Verification

viewer suite **2298 pass / 0 fail** · typecheck 33/33 (re-run uncached) · docs-samples clean · viewer build exit 0. Module sizes all under 400 (ComparePanel 383, useCompare 293). Single amended commit so the stacked `aabb-plumbing` rebase stays clean.

louistrue:
Both addressed in `0c9c1320`.

## 1. Publish helper — extracted, and the invariant is now *structural* rather than documented

`publishCompareResult(built)` is the only place a `CompareResult` is produced; both call sites go through it.

The part that makes this worth doing rather than just tidier: **the helper performs the `getState()` read itself**, so a caller cannot hoist it back before the awaits. The previous shape had that read as four loose `const`s in `runComparison` — one line-move away from reintroducing the P1. It is synchronous, so no `await` can sit between the read and `setCompareResult`. The doc comment names both properties, says they are why the effect may omit `result` from its deps, and states that adding an `await` reintroduces the bug.

Knock-on: `BuiltPair` now carries `baseModelId`/`headModelId` instead of a space-joined key, which deletes `pairKey` and an aliasing hazard where two different id pairs could join to the same string.

## 2. CodeRabbit's literal claim is wrong — and the real duplication was worse

`ComparePanel` does not gate the match sections at all; it passes `matchRows` down and `CompareMatchGroups` hides empty ones. So the stated finding does not hold.

But one layer over, there was a genuine and nastier duplication: **the panel's "Download report" bar and the results list's "the models match" empty state are exact negations, derived independently and from *different inputs*.** The bar counted match **records**; the empty state counted match **rows**. Those differ — `contentMatchRows` drops a record carrying no entities, and there is already a test for that. Drift meant **offering a report above a panel claiming the models matched.** Both now call one `hasReportableChanges(counts, matchRows)`, counting rows, because rows are what the list renders and what the CSV exports.

To your question about "present but empty": **it does need distinguishing, and the code already relied on it** — just spelled two ways (`!== undefined` in the hook, bare truthiness in the panel). `undefined` = never asked; `[]` = looked and paired nothing. It is also the effect's gate, so collapsing it to `length > 0` would make a zero-match result deny the flag it ran with and **re-diff forever**. Now one named `contentMatchingRan(matches)` with that reasoning attached.

## Mutation checks

**The prior P1 mutations were re-run against the refactored code and both still kill**, identically to the previous round:

| Mutation | Result |
|---|---|
| Pre-fix: options captured before the awaits, handed to the publisher | tests 1-4 fail |
| The rejected alternative: publish stale + add `result` to deps | **test 5 fails** (`compareRunSeq` reads 2) |

The second is the one that mattered to check — test 5's reach is exactly what a publish-path refactor could silently break, and it still discriminates.

Both new predicates needed **their own** tests, because each mutation left the entire `useCompare` suite green:

- `contentMatchingRan` → `!== undefined && length > 0` is killed only by `separates "ran and found nothing" from "never ran"`
- `hasReportableChanges` dropping the `+ matchRows.length` term is killed only by `counts a comparison whose entries are all zero but has matches`

I also drafted and then **dropped** a fourth case — it was assertion-identical to another and therefore not independently killable.

## Gap, stated

There is no `ComparePanel`/`CompareResultsList` render test, so the *wiring* of `hasReportableChanges` into the export bar and the empty state is covered by typecheck only. The predicates themselves are tested. Unchanged by this work, but worth knowing.

## Verification

viewer suite **2303 pass / 0 fail** · typecheck 33/33 · docs-samples clean · viewer build exit 0. `ComparePanel.tsx` back to exactly 383 lines. Single amended commit so the stacked #2005 rebase stays clean.

PR #2005, opened 2026-08-02
feat(geometry,viewer): carry the world AABB out of wasm, so Compare can say how far (#1891)

louistrue:
> **Stacked on #2002.** Review that first; this branch contains its commit. Base moves to `main` once it lands.

## What and why

#1988 added `MeshCollection.geometryAabbValues` and #1993 added volume and closure flags — and **none of them survived the FFI boundary**. `packages/geometry` extracted only the hashes, so two merged PRs had zero consumers and `EntityFingerprint.aabb` was never populated by anything. The diff engine's truthful `moved` vs `reshaped` distinction and its whole tier-3 mutual-nearest pass were dark in the viewer.

#2002's real-model run showed the symptom precisely: a wall that had genuinely been translated reported **`Moved` with a blank distance** — the documented no-`aabb` fallback.

## The acceptance test

`Building-Architecture.ifc`, every IfcRoot re-GUIDed, one wall given a fresh placement chain translated **+2500 mm** in X:

- **After: `Moved · 2.500 m`** against 2.5 m applied. Every other element `Renamed`, 0 added / 0 deleted, clean console.
- **Before**, same fixture with only the `aabb` hand-off reverted and a full reload: bare `Moved`, blank distance — reproducing #2002 exactly.

## Frame, confirmed three ways

The box is **absolute world, WebGL Y-up**; the Z-up→Y-up swap and the RTC fold happen in Rust, and nothing is folded in on the JS side. This is the part that would be silently wrong rather than obviously broken, so it was checked three independent ways:

1. `test-wasm-contract.mjs` reconstructs `origin + position + S(rtc)` and asserts the exposed box equals it tightly — passes against the wasm built here.
2. In the live viewer, a sampled mesh's `geometryAabb` centre `[5.8, -0.125, -5.9]` was **exactly** its `origin`. `origin` is the per-element AABB centre, so equality is only possible if the box is absolute; an origin-relative box would centre near zero.
3. The measured distance came out at the applied value — a Y/Z mix-up or a double-counted origin would not.

## Scope

`geometryVolumeValues` / `geometryClosureFlags` are **deliberately not plumbed**. Volume is groundwork for the split/merge detector; an unread array on every batch is dead weight. Stated in the commit and the changeset so the omission reads as a decision.

The two near-identical hash extractors in `geometry-coordinate.ts` and `geometry.worker.ts` collapsed into one shared reader (`geometry-fingerprints.ts`), and the session map now holds `{hash, aabb}` as a unit so the two cannot misalign. The aabb array never travels without its id array, and is allocated lazily so a boxless batch ships nothing.

## Mutation checks: 21 run, and 3 survivors were real defects in my own code

- **M3** — the array-length guard in `geometryAabbAt` was masked by the finite check, since out-of-range typed-array reads are `undefined`, which isn't finite. Collapsed to one guard, documented why a separate length check is unreachable.
- **M8** — an inherited `Math.min(ids.length, values.length)` guard had no test. Without it a short array stores `undefined` as an entity's `hash: bigint`. Kept and pinned with a malformed-source test.
- **M10** — the `geometryHashCount === 0` early return is a perf fast-path, not a correctness guard: it exists so a load with the diff feature off never touches the three copy-to-JS getters. Made observable with a getter-spy test.
- **V2** — a `!has` first-wins guard turned out to be **genuinely unfalsifiable** (the box is per *entity*, so every submesh gets the identical object). Removed rather than writing a test that pins an arbitrary tie-break.

## Two things stated rather than glossed

**The instanced-only side channel is not exercised on a real model.** GPU instancing did not fire on any local fixture (`pendingInstancedShards` null even on `dental_clinic.ifc` with 1656 mapped items), and `instancedGeometryHashes` was likewise null — so that route was *already* unexercised before this change. It is covered by unit tests at both ends.

**A fixture bug nearly read as a feature failure.** The first generator remapped every quoted 22-character token, silently renaming `Qto_WallBaseQuantities` (exactly 22 characters) and breaking the data hashes, so nothing matched at all. A fixture that shows nothing matched is a fixture bug until proven otherwise — this is the second time today that exact 22-character trap has bitten.

## `boundsIndex` recompute: follow-up, not here

`exportReport.boundsIndex` and `describeChange.meshBounds` share a defect — both fold raw `positions` and ignore `origin`, so with per-element local frames a moved element measures as stationary. Fixing one without the other splits the report from the detail panel, on the key-matched path this PR doesn't touch. Both must also keep serving cache-restored models, where the GLB cache round-trips `origin` but not fingerprints, so there is no box to substitute.

## Verification

`@ifc-lite/geometry` 178 tests · viewer compare 62 · typecheck 33/33 · docs-samples clean · build 40/40 · api-surface updated (`EntityWorldAabb`, `geometryAabbAt`) · `test-wasm-contract.mjs` 39/39 after `pnpm fixtures` (it skips clean on a fresh worktree with no fixtures — a green there proves nothing unless you fetch them first).

Changeset: `@ifc-lite/geometry` minor.


louistrue:
**The finding held.** Fixed in `26b04cd8`, and reproduced twice before touching anything.

## Reproduction

**Real models, real wasm, real pipeline:** `Building-Architecture.ifc` aligned to `Infra-Bridge.ifc` — same `EPSG:32760`, but MapConversions 2123 m / 32077 m apart with **60° between the two grid norths**, so a genuinely non-identity, rotating alignment.

- Pre-alignment, the wasm box equals its mesh bounds to **0.000000 m** across 14 entity boxes.
- Unfixed, the box ends up **36.45 m from its own mesh**.
- Feeding those to the real `classifyGeometryChange`: a **stationary** element reports `distance: 34.80 m`. Fixed: `distance: 0`.

A unit suite against the unfixed code failed 6 of 7 assertions, on **both** the same-CRS and cross-CRS paths.

## Why re-measure rather than corner-transform

The affine is a rotation about the vertical plus scale and translation, so `min`/`max` cannot be mapped component-wise — corner-transforming is the textbook answer. I did not use it for meshed entities, and the reason is the interesting part:

**Re-boxing a rotated box inflated it by up to 0.87 m on that real pair, against a 1 mm `reshapeTolerance`.** That inflation would flip `moved` to `reshaped` against an unaligned side that kept its tight box — trading a wrong distance for a wrong *classification*.

So meshed entities are **re-measured from the vertices the alignment just wrote**. I verified the wasm box is exactly the bounds of the entity's vertices (indexed and all-positions bounds agree to 0.000000 m), so re-measuring reproduces the same construction in the new frame. Post-fix box-vs-mesh agreement: **0.000000 m**. It is also idempotent, and the box literally cannot drift from its mesh again.

**The instanced channel** has no vertices on that path, so it takes all **eight** corners with the AABB re-derived — conservative under rotation, and nothing tighter exists without vertices.

Both need the frame hop the vertices get for free (positions are viewer-local, the box is absolute): strip the source offset, align, re-apply the reference offset. The cross-CRS per-vertex pipeline is now one shared closure so proj4 cannot reproject vertices and corners along two different chains.

**Also fixed:** `realignFederation` restored positions and normals but not boxes, so a second re-align would have compounded the error.

## Two mutation survivals worth reading

1. **A 90° rotation fixture failed to catch the 2-corner mutation** — a quarter turn maps a box onto a box, so two corners suffice. Replaced with 30°. A fixture that cannot discriminate is not a test.
2. After the re-measure change, two mutations survived because the corner path's offset arithmetic was only reachable through the instanced channel, which no test drove with non-zero RTC offsets. Added that test.

11 mutations, 11 killed.

## What the evidence does not cover, stated plainly

**A discriminating real-model federated compare pair cannot be built from `tests/models/`, and I did not fake one.** It needs two revisions of *one* building with *different* georeferences; the fixture set ships different-discipline models sharing no GlobalIds. The real-model evidence above is the invariant that underwrites any distance — the box sits on its own mesh, measured on real files through the real pipeline — plus the real diff engine's verdict on real boxes. The end-to-end `Moved · 2.500 m` analogue is a unit test.

**Blast radius:** a re-export whose georeference changed, or moved to another CRS, between revisions. Two revisions sharing a georef get an identity transform and were never affected.

## Verification

`@ifc-lite/geometry` 178 · viewer **2317 tests, 0 fail** (new suite 11/11) · typecheck 33/33 · docs-samples clean · build 40/40 · api-surface unchanged · viewer build exit 0.

No changeset change needed: all edits are in `apps/viewer`; `@ifc-lite/geometry` is untouched. The new logic went into its own 275-line module rather than growing `federationAlign.ts`, which was already 508 lines.

louistrue:
**The failure shape is real. The state is not reachable.** I skipped it as a defect, deleted the branch that created the coupling anyway, and did not write a test for it — reasoning below.

## The shape is genuine

A probe drove the exact described sequence — align, restore positions but not the instanced snapshot, re-align to a different anchor:

```
after first align : mesh [2500,0,0]  instanced [2500,0,0]
after re-align    : mesh [700,0,0]   instanced [3200,0,0]   <- double-transformed
```

If the state existed, instanced-only compare distances really would be wrong.

## But it cannot exist — four independent reasons

1. **Only two writers** of the five `preAlignment*` fields, and both assign all five as one statement group inside one `if`. No `updateModel` call patches a subset.
2. **`preAlignmentGeometryAabbs` is never absent while positions are present** — it is `meshes.map(m => m.geometryAabb)`, unconditionally an array wherever positions are set. So the mesh-box half of the finding has no path at all, and it is *doubly* harmless: with no box snapshot restored the mesh box still lands correctly, because last round's change re-measures meshed entities from their aligned vertices.
3. **`preAlignmentInstancedGeometryAabbs` is `undefined` exactly when the channel is** — confirmed by probe: alignment leaves an absent channel absent and an empty one empty, and the only other writers construct a fresh `GeometryResult` *before* capture. Absent at capture implies absent forever, so the skipped restore restores nothing that exists.
4. **"Aligned before this field existed" needs two code versions coexisting.** Models are not persisted, the geometry cache round-trips neither `preAlignment*` nor `instancedGeometryAabbs`, and `updateModel` is a plain spread so an explicit `undefined` overwrites rather than silently preserving a stale value.

All four alternative paths I was asked to check are covered by 2-4.

## What I changed, and what I deliberately did not

One deletion, not a guard: the restore is now unconditional. The `if (instancedAabbSnapshot)` made the capture and the restore answer *different questions* — "was there a channel?" versus "was there a snapshot?" — and removing it makes them agree by construction, collapsing reason 3 from a cross-module invariant into a local one.

**No test for it.** A test would have to construct the impossible state, which is precisely the unfalsifiable-guard trap this workstream already deleted one of.

I did add a test for the invariant reason 3 *used to* rest on, because that one is real, reachable and exercised on every alignment: **alignment must not conjure an instanced channel a model never had.** Absent and empty are distinguishable downstream — `buildFingerprints` branches on presence.

## Verification

13 mutations, 13 killed. Two new ones against the new test (conjuring the channel from absent; seeding it from meshed boxes), plus all 11 from the previous round re-run against the updated suite.

geometry 178 · viewer **2323 tests, 0 fail** · typecheck 34/34 · viewer build exit 0.

## Unrelated pre-existing bug found while tracing

`realignFederation` `continue`s the **anchor** model without ever restoring it. Switch the anchor to model A and then back to X, and X keeps geometry baked into A's frame while defining the federation frame. Predates this PR; filed separately rather than folded in.

louistrue:
Both held. Fixed in `f87438ba`.

## A — fingerprint cache key: held, fixed

Same class as the P1 this branch already fixed: **a cache key that doesn't cover a dimension the data varies on.** `builtRef` was keyed on the A/B model ids, but `realignFederation` mutates meshes **in place** — same ids, same `geometryResult` object — so a second Run reused pre-alignment boxes against post-alignment meshes.

The signal already existed: `geometryContentVersion`, whose only caller is `realignFederation` and only when something actually moved. Nothing new added to the store.

Two details that took the thinking:

- The version is read **before** the extraction awaits. A realign landing mid-extraction then stamps the version the meshes had when reading *started*, so the next Run rebuilds. Stamping the newer one would mark a half-stale extraction as current.
- A bump retires the **published result**, not just the cache. Nulling `builtRef` alone would make the re-diff effect hit `if (!result || !built) return;` on the next option change, and the panel would disagree with its own controls — exactly the staleness `publishCompareResult`'s invariant exists to prevent. The invalidation effect is declared *before* the re-diff effect so a commit changing both clears first, and it is seeded with the mounted value so a remount is not mistaken for a bump.

## B — partially-reprojected entity: held, but the stated mechanism is wrong

**The box does not mix old and reprojected coordinates.** `continue` on a failed vertex skips `extendEntityBounds` too. Probed on a 9-vertex mesh with one outlier:

```
[ifc-lite] Cross-CRS alignment: 1/9 vertex transforms failed ... left at their original positions.
geometryAabb: {"min":[-271623.06,0,-17540.25],"max":[-271621.97,1,-17539.18]}
outlier vertex now: 300000000 0 0
```

The real defect is the adjacent one: **the box covers only the reprojected subset while the mesh spans two coordinate frames**, and it is published as authoritative.

Reachability measured rather than assumed — proj4 returns **`Infinity`, it does not throw**, and it fails **per point**:

```
normal UTM point   -> [28377.03, 5017540.25]   finite
outlier 1e7        -> [9836943.90, 8625749.60] finite
outlier 1e8        -> [Infinity, Infinity]     NOT finite
```

The domain edge for EPSG:32632→32633 sits between 1e7 and 1e8, so one outlying vertex is left behind while its neighbours move. This codebase already has explicit outlier filtering for exactly such vertices.

**Chosen: drop the box for that entity.** Measuring covers only the half that moved; corner-transforming asserts a position for geometry that is not there. Neither is honest, and the engine has a documented absent-box fallback — no answer beats a confident wrong one. The load-bearing half: **a meshed entity no longer falls back to the corner transform at all.** That path is now exclusively the instanced-only channel, which has no vertices on this side.

I left the per-vertex best-effort geometry behaviour alone — it predates the box, it already warns, and re-litigating it is not this PR's job.

## A survivor I am not claiming as pinned

M8 (fabricate a box for a boxless mesh) **now survives**, because removing the corner-transform fallback closed the route it exercised. Three diagnostic mutations rather than a shrug:

| mutation | result |
|---|---|
| drop the boxless guard only | **survives** — inert on its own today |
| accumulate bounds for every mesh, guard kept | **survives** — the guard absorbs it |
| both together | **killed** — the test catches the fabricated box |

So the guard is redundant with an upstream conditional accumulator, and demonstrably load-bearing the moment that accumulator becomes unconditional — a one-line, plausible edit. Kept as defence-in-depth **with a demonstrated kill**, and I am explicitly not claiming it is individually pinned. Making it individually falsifiable would need a mixed-submesh entity (one submesh boxed, one not), which the wasm path cannot produce, so that test would pin an unreachable state.

M9 was **removed as superseded** — it mutated a fallback that no longer exists.

## Verification

geometry 178 · viewer **2328 tests, 0 fail** · typecheck 34/34 · viewer build exit 0 · test-wiring, docs-samples, api-surface all clean. 16 mutations, 15 killed, the survivor accounted for above.

Note: typecheck failed on first attempt because the rebase brought #2002's new workspace dependency and this worktree's `node_modules` had no symlink for it — the known fresh-worktree staging gap, not a code problem. `pnpm install --frozen-lockfile` fixed it, lockfile unchanged.

The anchor-restore bug found while tracing this is filed as #2007.

louistrue:
Held, and closed as a **shape** rather than a third patch. `75c5c736`.

## The bug

`runComparison` stamped the version before its awaits; the invalidation effect cleared `builtRef` and called `clearCompare()`; the in-flight run then reassigned `builtRef.current` and published — **putting back exactly what the invalidation had just removed.** Reachable through the unconditional frame yield alone, and through the extraction's `setTimeout(0)` every 1500 entities on any real model. The worse variant: extraction yields *between* models, so side A can be post-align while side B is pre-align.

## Unified — one of the two prior guards is subsumed, the other deliberately is not

A new module asks one question — *do these fingerprints still describe the meshes?* — at the two moments it matters: before reusing a cached pair, and after the extraction's awaits. One predicate, three call sites, so they cannot drift. I verified the two windows are genuinely different (distinct mutations kill distinct tests), so both checks are load-bearing — but they are now one definition asked twice instead of two hand-written guards.

**The options guard stays separate, on purpose.** Scope/blacklist/matching are a different axis with a strictly *better* remedy: `publishCompareResult` reads them synchronously at publish time, so a stale captured value never exists at all. A version check cannot express "there is nothing to compare because nothing was captured", and collapsing them would trade a stronger guarantee for a tidier diagram.

## "Abandon" does not abandon

It **re-extracts against the world that replaced it** and completes. Silently doing nothing after the user pressed Run is its own bug, and the re-align had already cleared the panel, so bailing would leave it blank after a deliberate action. Bounded at 3 attempts — a re-align is discrete; the bound exists only so a pathological stream cannot spin. Exhaustion sets a real error rather than publishing something known stale:

> The model geometry kept changing while comparing. Run the comparison again.

That message is load-bearing: `setCompareRunning(false)` in the `finally` runs on the early return too, so without it the panel would go from "Comparing…" to blank with no explanation.

## The mutation sweep earned its keep

- **Q5 survived, so I deleted the code rather than pinning it.** Clearing the cache after a failed attempt was provably redundant — versions move forward only, so the reuse check rejects the same stale cache next pass for the same reason. The function is simpler now.
- **Q3 initially survived** because a wrong reuse decision was silently *rescued by the retry*. Fixed by giving that test `maxAttempts: 1`, so the up-front decision has to be right on its own.
- **Q8 initially survived** because the hook wiring had no coverage at all. Closed by extracting the version read as a named live reader with its own test.
- **Q7 hung the runner instead of failing.** A hanging test reports nothing, which is worse than a red one — so the fixture now throws past a bound, and the harness treats a timeout as a kill.

8 new mutations, all killed. Earlier sweeps (M1-M7, M10, M11, P1-P6) re-run and still killed. **M8 still survives, unchanged and already accounted for** in the previous comment: inert alone, absorbs an unconditional accumulator, caught in combination — defence-in-depth with a demonstrated kill, not claimed as individually pinned.

## Verification

geometry 178 · viewer **2335 tests, 0 fail** · typecheck 34/34 · viewer build exit 0 · test-wiring, docs-samples, api-surface clean. Lockfile unchanged.

## One residual, on the record rather than bolted on

`runComparison` captures the two models once, so a model **removed** mid-run would still be published. Same shape, different dimension — the version counter covers mesh content, not model membership. It predates this branch and its reachability is unestablished, so I would rather name it than add a fourth dimension to the guard without evidence. Say the word and I will file it.

Also still open and unrelated: `realignFederation` never restores the *anchor* model (#2007).

PR #2016, opened 2026-08-03
docs(agents): carve out the post-model-swap case from the FederationRegistry rule

louistrue:
`AGENTS.md:25` says resolve IDs through `FederationRegistry`, never ad-hoc math. Main's own code says the opposite for one specific window — `modelSlice.ts:81`:

> **BULLETPROOF:** Resolve globalId using model store data instead of singleton registry. This is more reliable because it uses Zustand state which is always in sync with React.

Both are right. The registry is fine in general and genuinely unreliable immediately after a model swap, because it is a singleton that is not guaranteed synced with Zustand at that instant — which is exactly when `purgeStaleEntityState` and `syncSourceModel` run.

## Why now

This surfaced on #1976. A reviewer flagged `syncSourceModel.ts:254` for resolving global ids without the registry; @BIMvoice correctly declined to "fix" it and escalated instead, because following the rule there would have reintroduced the staleness the existing pattern exists to prevent. That is the right instinct, and it is also the signal that the rule was under-specified rather than that the code was wrong.

Leaving it as-is means the next person either violates a documented rule or introduces a real bug, with nothing in the repo telling them which.

## What changed

One sub-bullet under the existing rule. It names `resolveGlobalIdFromModels` as the helper for that window, so the guidance is "call this" rather than "hand-roll the offset arithmetic again" — the duplication was the actual defect at that call site, not the choice of data source.

Docs only; no code, no changeset.


louistrue:
Both findings were right, and both were mine. Verified each against the code rather than taking them on description.

**@greptile-apps — absent symbols.** Confirmed: `purgeStaleEntityState` and `syncSourceModel` have **zero occurrences** on main. They exist only on the #1976 branch, which is where I hit this originally, and I cited them without checking they had landed. A contributor reading the rule could not have found the boundary I was describing.

**@chatgpt-codex-connector — the scope was wrong, and wrong in the harmful direction.** This is the more important of the two. I framed the exception as a narrow post-model-swap window and closed with "outside that window the registry rule above stands". That sentence would have caused the bug it was meant to prevent:

- `modelSlice.ts:319-340` runs a second pass over `mutationViews`, with the comment *"the federation resolver knows nothing about overlay state"*
- `resolveEntityRef.ts:22` calls `resolveGlobalIdFromModels` as the canonical resolution path

So an entity created through `StoreEditor` fails to resolve through the registry **at any time**, not just after a swap. Directing future changes back to the registry outside that window would break resolution for newly created entities.

Rewritten to lead with overlay awareness (always applies) and treat staleness-immunity as the secondary reason (applies after a swap), citing only symbols that exist on main.

Worth naming the mistake rather than just fixing it: I wrote a general rule from the single case I had just debugged, and that case turned out to be the *lesser* of the two reasons the exception exists. The reviewer who had no context on #1976 saw the broader reason precisely because they read the code instead of my framing.

Issue #1959, opened 2026-08-01
12 GeometryProcessor call sites never dispose their WASM handle, including the canonical load path


BIMvoice:
### Summary

`AGENTS.md` requires WASM handles to be freed deterministically in `try/finally`. An audit of all 39 `new GeometryProcessor` sites found **12 that never call `dispose()` on any path**, including the viewer's canonical model-load path.

This came out of #1956, where one such site (`packages/cli/src/hbjson-export.ts`) was flagged in review and fixed. I said at the time that my list of other suspects was "a grep, not a careful read" — this is the careful read.

**Most sites are fine.** 11 of 22 non-test call sites are correct, and the example/vanilla-template singletons are page-lifetime by design and not leaks. The problem is concentrated in the highest-frequency paths.

### Confirmed, ranked by blast radius

| Site | Why it matters |
|---|---|
| `apps/viewer/src/hooks/useIfcLoader.ts:1047` | The canonical load path. Fires on **every** IFC load, primary and federated, success or failure. No `dispose` anywhere in the file |
| `apps/viewer/src/components/mcp/PlaygroundViewer.tsx:163` | Re-fires on every model swap; also escapes via the `cancelled` early return |
| `apps/viewer/src/components/viewer/HbjsonExportDialog.tsx:93` | Every "Export HBJSON" click. The `finally` at `:111` only resets `isExporting` |
| `apps/viewer/src/components/mcp/playground-dispatcher.ts:336` | One per model clash-checked |
| `packages/mcp/src/tools/clash.ts:60` | Long-lived MCP server process — accumulates across tool calls |
| `packages/cli/src/headless-backend.ts:645` | Server path, leaks per request |
| `packages/export/src/lod1-generator.ts:134,172` | Both the primary and the fallback processor, including on the `catch` path |
| `packages/create-ifc-lite/src/templates/react.ts:251` | Mount effect's cleanup disposes `viewerRef` but not the processor. Scaffolding, so it teaches the pattern to every generated React project |
| `diagnose-geometry.ts:65`, `extract-entities.ts:355`, `cli/commands/clash.ts:49`, `gym.ts:217` | Real violations, but one-shot CLI processes that exit immediately. Low urgency; `gym.ts` is the exception if its REPL is ever long-lived |

I verified the top two myself; the rest come from the audit trace and are worth a second look before anyone acts on them.

### Correct, for the record

[...]

PR #2022, opened 2026-08-03
fix(viewer): free the geometry WASM handle on every exit path (#1959)

BIMvoice:
Partial fix for #1959 — the two P0 sites that have no design question attached.

## What leaked

Both sites construct a `GeometryProcessor`, run it, and never dispose it. Both re-fire per model: `PlaygroundViewer`'s effect on every model swap, `meshForClash` once per clash-checked model.

Neither leaked only on the happy path, which is the part worth noting — a `dispose()` appended after the last statement would have fixed neither:

- `PlaygroundViewer` escapes early via the `cancelled` check immediately after `process()`, and again on the empty-mesh branch.
- `meshForClash` **throws** `UNSUPPORTED_OPERATION` when a model yields no meshes.

So both are `try/finally`, and the tests assert disposal on *every* exit rather than just success.

## Verified before freeing

Neither call site calls `processor.getApi()` — that is internal to `process()` — and the returned `MeshData` is already copied out into plain JS typed arrays, so nothing downstream (mesh cache, clash engine) holds a reference into WASM memory. Freeing at these two points is safe.

## One structural change, and why

`PlaygroundViewer`'s loader is extracted into `loadPlaygroundGeometry`. Mounting the component runs a sibling effect that constructs a real `THREE.WebGLRenderer`; happy-dom returns `null` from `canvas.getContext('webgl')`, and the renderer can't be stubbed either — `mock.method` on the frozen ES-module namespace throws `Cannot redefine property`. No existing pattern in the repo fakes a WebGL context.

Extracting the loader was the proportionate way to get the paths under test. Building a WebGL-context Proxy to verify a `dispose()` call was not. Happy to revert the extraction if you'd rather keep the component intact and take the fix untested there.

## Evidence

Removing both `dispose()` calls fails all five new tests (**2333 pass / 5 fail**); restored, **2338 pass / 0 fail**. Typecheck 34/34.

## Deliberately not fixed

`useIfcLoader.ts` — the canonical load path, and the biggest of the leaks. Its processor's handle is handed to `IfcParser.parseColumnar` via `getApi()`, so disposal has to wait for both the parse promise and `processAdaptive` to settle. You flagged that as needing a design decision and I don't think it should be made inside this PR. The P1/P2 sites from the issue are also untouched.

Draft until you've looked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model geometry loading and clash checking reliability.
  * Ensured processing resources are consistently released after successful, cancelled, empty, or failed operations.
  * Prevented callbacks from running when processing is cancelled or produces no usable geometry.
  * Preserved mesh caching and successful mesh generation behavior.

### DIFF
--- a/apps/viewer/src/components/mcp/PlaygroundViewer.test.ts
+++ b/apps/viewer/src/components/mcp/PlaygroundViewer.test.ts
@@ -1,0 +1,136 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `loadPlaygroundGeometry` (#1959 P0 leak) creates a GeometryProcessor per
+ * call and, before the fix, never disposed it — including the `cancelled`
+ * bail-out that fires on every model swap (the effect's cleanup flips
+ * `cancelled` and a fresh IIFE starts for the new model, so a busy playground
+ * session leaked one WASM handle per swap).
+ *
+ * This exercises the extracted loader function directly rather than
+ * mounting `<PlaygroundViewer>`: mounting pulls in the "mount Three.js once"
+ * effect, which constructs a real `THREE.WebGLRenderer` — happy-dom's
+ * `canvas.getContext('webgl')` returns `null`, so that effect throws before
+ * the geometry-loading effect under test ever gets to run, and there is no
+ * existing pattern in this repo for stubbing a WebGL context convincingly
+ * enough for `WebGLRenderer`'s capability-detection calls (`getParameter`,
+ * `getExtension`, ...). Testing the disposal contract through the standalone
+ * function is the proportionate alternative to building that stub.
+ */
+
+import { describe, it, mock } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { GeometryProcessor, type GeometryResult, type MeshData } from '@ifc-lite/geometry';
+import { loadPlaygroundGeometry } from './PlaygroundViewer.js';
+import { parsePlaygroundModel, type LoadedPlaygroundModel } from './playground-dispatcher.js';
+
+function ifc4(body: string): string {
+  return [
+    'ISO-10303-21;', 'HEADER;', "FILE_DESCRIPTION((''),'2;1');",
+    "FILE_NAME('','',(''),(''),'','','');", "FILE_SCHEMA(('IFC4'));", 'ENDSEC;',
+    'DATA;', body, 'ENDSEC;', 'END-ISO-10303-21;', '',
+  ].join('\n');
+}
+
+async function schemaOnlyModel(globalId: string): Promise<LoadedPlaygroundModel> {
+  const bytes = new TextEncoder().encode(
+    ifc4(`#1=IFCWALL('${globalId}',$,'Wall A',$,$,$,$,$,.STANDARD.);`),
+  );
+  return parsePlaygroundModel(bytes.buffer as ArrayBuffer, `${globalId}.ifc`);
+}
+
+/** A promise plus its resolver, so the test controls exactly when
+ *  `processor.process()` settles relative to flipping `cancelled`. */
+function deferred<T>(): { promise: Promise<T>; resolve: (v: T) => void } {
+  let resolve!: (v: T) => void;
+  const promise = new Promise<T>((res) => { resolve = res; });
+  return { promise, resolve };
+}
+
+describe('loadPlaygroundGeometry WASM disposal (#1959 P0 leak)', () => {
+  it('disposes the GeometryProcessor handle when the caller cancels mid-flight', async () => {
+    const gate = deferred<GeometryResult>();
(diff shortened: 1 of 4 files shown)


louistrue:
Quick process note in case you're holding these back waiting on us: **you don't need our sign-off to mark a PR ready.** Undraft whenever *you* think it's done — that's the signal we review against, and we'd rather review a ready PR than have it sit.

Both of these read as complete work to me, so I've started adversarial review passes on them now rather than waiting. I'll post findings either way.

Two specifics:

**#2022** — the partial scope is the right call, and I checked the reasoning holds. Fixing the two sites with no design question attached, and leaving `useIfcLoader.ts:1047` (the canonical load path, #1959's highest blast radius) for a separate decision, is better than bundling a lifetime question into a leak fix. If the remaining ten turn out to be mechanical rather than design-blocked, say so and we'll take them as a follow-up.

**#1967** — the `MutablePropertyView.hasChanges()` semantics change is the part I'm scrutinising hardest, since `@ifc-lite/mutations` is published at 1.21.x and that method is public API. Your reasoning (undo doesn't pop `mutationHistory`, so the old reading over-reported after an undo, and `getModifiedEntityCount()` feeds a user-visible count) is right on the merits — I verified separately that nothing in-repo consumes the view's `hasChanges()` in production, and that the similarly-named `mutationSlice.hasChanges(modelId)` is unrelated. What I want the review to establish is `getEffectiveChanges()`'s `previousValue` derivation under undo→redo and repeated edits to the same field.

Unrelated heads-up on the FederationRegistry question you escalated on #1976: answered and merged as `53bb4aa2` — `AGENTS.md:25` now carves out globalId→model resolution, and the reason turned out broader than the one raised. The registry cannot see overlay-allocated ids at all, so a `StoreEditor`-created entity fails to resolve through it at any time, not just after a model swap. You were right not to apply that finding.

louistrue:
Reviewed this adversarially. **It's clean** — I went looking for reasons not to merge and didn't find one.

What I checked rather than assumed:

**Both try/finally blocks are actually complete.** In `loadPlaygroundGeometry` (`PlaygroundViewer.tsx:114-155`) construction sits inside the outer try with the inner try opening on the very next statement and no `await` between, so the `cancelled` return, the empty-mesh return, and any `init()`/`process()` rejection all pass through the `finally`. In `meshForClash` (`playground-dispatcher.ts:340-364`) the cache-hit return happens *before* construction — nothing to free — and the `UNSUPPORTED_OPERATION` throw is inside the try. Your comment that construction can't throw synchronously is true: the constructor only builds `IfcLiteBridge` and caches flags, and with `preferNative: false` the bridge always exists, so `dispose()` is never a null surprise.

**No double-free.** `dispose()` nulls the handle after freeing and is a no-op before `init()`. Each invocation owns a fresh processor, so a model swap or a StrictMode double-invoke yields two processors each disposed once — never the same handle twice.

**It fixes the leak rather than moving it.** This was my main target, since AGENTS.md lists `MeshCollection`, `MeshDataJs` and the pre-pass cache separately. Traced it: `process()` returns copy-out typed arrays, each `MeshDataJs` is freed in a per-mesh finally and the `MeshCollection` in an outer one (`geometry-coordinate.ts:52-160`), and `clearPrePassCache()` runs in `collectMeshesViaPrePass`'s own finally. So the clash cache and `loadMeshes` hold plain JS data, and disposing the processor releases the only remaining handle. Your prose matches the geometry package, not just the diff.

**The partial scope is justified, not deferral in disguise.** I checked `useIfcLoader` specifically: its processor's raw wasm handle is handed to `IfcParser.parseColumnar` (`:1171`) *and* drives `processAdaptive` (`:1317`), so disposal needs two consumers coordinated plus the own-and-reuse redesign #1959 flags. That is a real design question. And `HbjsonExportDialog` was already fixed in #1956, so the P0 list genuinely reduces to these two.

**The tests pin the fix.** I judged each of the five against "would this fail if the `finally` were removed" — all five would. The cancelled test uses a deferred gate to flip `cancelled` mid-`process()` then asserts dispose count 1 *and* `onMeshes` count 0; the empty-mesh test pins the phase sequence plus dispose; the meshForClash test asserts `UNSUPPORTED_OPERATION` plus dispose. A naive appended `dispose()` passes only the two success tests and fails the other three — which is exactly the discrimination that matters here.

## Two nits, neither blocking

1. **No rejection-path test.** Nothing covers `init()`/`process()` *rejecting*. The same finally covers it structurally, but an `init()` rejection exits before `result` exists, which is a slightly different path from the empty-mesh throw. One `process: () => Promise.reject(...)` case each would close it.
2. **Pre-existing, repo-wide:** a bare `finally { processor.dispose() }` can mask the original error if `free()` traps after a wasm runtime trap — and traps in the raw `getApi()` path bypass `recordWasmRuntimeTrap`, leaving the handle live. But this is the blessed pattern at every existing dispose site (`useGridLines3D.ts:54`, `GLBExportDialog.tsx:268`, `HbjsonExportDialog.tsx:148`), so it's not a regression here. Worth a follow-up across all sites if it bothers anyone, not a reason to hold this.

Ready from my side. Undraft whenever you like — I'm running a CodeRabbit pass now since the hosted one skips drafts, and I'll post if it surfaces anything.

PR #1761, opened 2026-07-16
feat(source): add cloud source plugin architecture with Dalux Build provider

bruadam:
## What and why

Add Source and Dalux Box Integration

Closes #1663

## How it was verified

Manual testing and debugging on personal Dalux Box project.

Additionally, before opening this PR:
- Ran `pnpm test` for every touched package/app: `@ifc-lite/plugin-api` (5/5), `@ifc-lite/source-dalux` (14/14), `@ifc-lite/viewer` (1699/1701, 2 pre-existing skips) — all passing.
- Ran `pnpm typecheck` for the viewer — clean.
- Fixed a real bug found while running the suite: `source-dalux`'s test mocks didn't include a `headers` property on the mocked fetch `Response`, but the client logs `response.headers.get(...)` before checking `response.ok` on every request — 11/14 tests were failing until the mocks were corrected.
- Synced `pnpm-lock.yaml`, which was missing the `dalux-build-api` dependency entry despite it being in `package.json`.

- [x] `cargo test --workspace` (Rust) and/or `pnpm test` (TS) pass locally
- [x] Geometry/WASM change: ran `scripts/build-wasm.sh` then `pnpm test:wasm-contract` — N/A, no geometry/WASM change in this PR

## Checklist

- [x] A test asserts the new behavior through a fixture or a stated invariant
- [x] Published `packages/*` touched: added a changeset (`pnpm changeset`) and, if
      the export surface changed, ran `pnpm api-surface:update`
- [x] Geometry-output change: re-pinned **both** `mesh_determinism.json` and
      `mesh_determinism.wasm32.json` — N/A, no geometry output changed
- [x] House rules: no `as any` / `@ts-ignore`, no silent `catch {}`, no repo-wide `cargo fmt`.
      Module size: split `HierarchyNode.tsx` and `provider.ts` back under ~400 lines;
      `SourceBrowser.tsx` (760 lines) is a known, deliberate exception — a single
      cohesive step-wizard component with no render-test coverage to safely verify a split.
- [x] No client or project identifiers in code, tests, commit messages, or this PR
      ("Dalux" is the named third-party integration this feature ships, not a client/engagement name)

## Summary by CodeRabbit

* **New Features**
  * Added a new **Cloud sources** sidebar panel to browse projects, containers, and IFC files with a step-by-step flow.
  * Enabled **download**, **load as federated models**, and **sync** from the latest available provider revisions.
[...]

PR #2023, opened 2026-08-03
feat(source-dalux): Dalux Build provider on the v2 plugin contract (2/3)

### LINKED ISSUE #1663: Dalux box integration

louistrue:
**Closes #1663.** Second of three PRs landing cloud/CDE source integration — **stacks on #1976** (contract + host + UI). `@ifc-lite/source-msgraph` follows in PR C.

> Base this on #1976. The diff shown against `main` will include A's changes until A merges; the 27-file diff against A's branch is the actual content here.

## Credit

The provider and the architecture are @bruadam's, from #1761. He asked that one person carry the PRs rather than splitting authorship across branches — *"easier to have only one person on it, no bottlenecks"* — so this carries `Co-authored-by: Bruno Adam <bruno.adam@pm.me>`.

What changed since #1761: migrated to the v2 contract (`Page<T>` listing, `SourceFileRef` download, `watchRevisions` replacing `checkRevisions`, declared `capabilities`/`auth`/`permissions.relay`), dropped the third-party `dalux-build-api` client, hardened the relay, and fixed the pagination/validation defects listed in the changeset.

## The manifest fix that unblocked it

#1761 declared `manifest.api: '^0.1.0'` against a host at `PLUGIN_API_VERSION = 2.0.0`, so the provider could **never register** — the app white-screened instead of failing loudly. It now declares `^2.0.0`.

## The guard PR A left behind did exactly its job

#1976 shipped with `source-host.test.ts` asserting `createRegisteredProviders()` was **empty** — deliberately inverted, because with no providers the per-provider loop below it is vacuous and would pass forever. It was written to fail the moment a provider appeared, with a message saying what to restore.

It failed here, as designed. Restored to `assert.ok(providers.length > 0, ...)`.

**Mutation-checked:** setting the Dalux manifest back to `^0.1.0` fails that test with the version mismatch. The exact defect that blocked #1761 is now a CI failure rather than something that ships.

## Relay

Dalux Build sends no CORS headers, so requests go through the same-origin `/api/dalux` relay — declared in the manifest, validated by the host against routes it actually has configured, and refusing upstream redirects that leave the declared host so the API key cannot be carried off-origin. The production rewrite (`vercel.json`) and the dev proxy (`vite.config.ts`) mirror each other deliberately; a drift between them is the classic "works in dev, 404s in prod" bug.

This is the use case that decided the architecture: a five-method `connect/list/download` interface cannot express a validated relay, and every provider needing one would grow bespoke serverless code anyway.

## Changeset is `minor`, not `major`

The inherited changesets said `major`. `@ifc-lite/source-dalux` is at 0.1.0 and **404 on npm** — never published, so there are no consumers to break, and this repo bumps breaking changes on `0.x` packages as `minor`. Publishing a brand-new package straight at 1.0.0 would imply a stability promise we are not making. Consolidated into one changeset that reads as a first release.

## Verification

- `pnpm typecheck` 35/35 · `pnpm test` 80/80 tasks
- `@ifc-lite/source-dalux` 77 tests · `source-host` 25
- `api-surface.json` regenerated (39 packages, 4001 exports) — new published surface, expected
- `check-test-wiring` and package READMEs clean
- `packages/wasm/pkg/*` reverted to A's base: a local build had regenerated the `.d.ts` with bindings from a newer `main` than A branched from, which is A's rebase to do, not this PR's

### DIFF
--- a/.changeset/source-dalux-initial.md
+++ b/.changeset/source-dalux-initial.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/source-dalux": minor
+---
+
+First release of `@ifc-lite/source-dalux`, a Dalux Build (Box) file-source provider on the v2 plugin contract (`manifest.api: '^2.0.0'`, declared `capabilities`/`auth`/`permissions.relay`, `Page<T>`-returning listing methods, `SourceFileRef`-based `download`, `watchRevisions`). Closes #1663.
+
+Dalux Build's API sends no CORS headers, so browser requests go through the same-origin relay at `/api/dalux` (`vercel.json` rewrite in production, vite proxy in dev), declared in the manifest and validated by the host against its configured routes. The relay refuses upstream redirects that leave the declared host, so a redirect cannot carry the API key off-origin, and it does not forward the key to any other host.
+
+Talks to the Dalux HTTP API directly rather than through the third-party `dalux-build-api` client.
+
+`minor`, not `major`, despite superseding an earlier unreleased shape: the package has never been published, so there are no consumers to break, and this repo bumps breaking changes on `0.x` packages as `minor`.
--- a/api/dalux.ts
+++ b/api/dalux.ts
@@ -1,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import {
+  createDaluxRelayHandler,
+  loadDaluxRelayConfig,
+} from '../server/sources/dalux-relay.js';
+
+export const runtime = 'edge';
+export const config = { runtime: 'edge' };
+
+const handler = createDaluxRelayHandler(loadDaluxRelayConfig(process.env), {
+  fetchImpl: fetch,
+});
+
+export default handler;
--- a/apps/viewer/package.json
+++ b/apps/viewer/package.json
@@ -50,6 +50,7 @@
     "@ifc-lite/parser": "workspace:^",
     "@ifc-lite/pointcloud": "workspace:^",
     "@ifc-lite/query": "workspace:^",
+    "@ifc-lite/source-dalux": "workspace:^",
     "@ifc-lite/renderer": "workspace:^",
     "@ifc-lite/sandbox": "workspace:^",
     "@ifc-lite/sdk": "workspace:^",
--- a/apps/viewer/src/services/sources/registered-providers.ts
+++ b/apps/viewer/src/services/sources/registered-providers.ts
@@ -3,6 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import type { FileSourceProvider } from '@ifc-lite/plugin-api';
+import { DaluxBuildProvider } from '@ifc-lite/source-dalux';
 
 /**
  * Every file-source provider the viewer actually registers, in one place.
@@ -12,13 +13,8 @@ import type { FileSourceProvider } from '@ifc-lite/plugin-api';
  * drifting apart) read from this list — so the test can never drift from
  * what the running app actually does.
  *
- * Deliberately EMPTY in this change: it lands the contract, the host and the
- * UI, and the first providers follow in their own reviewable PRs
- * (`@ifc-lite/source-dalux`, then `@ifc-lite/source-msgraph`). The Sources
- * panel renders its empty state until one is added, and the conformance kit
- * in `@ifc-lite/source-fixture` exercises the host against a fixture provider
- * meanwhile, so the host is covered without shipping a real integration.
+ * `@ifc-lite/source-msgraph` (SharePoint/OneDrive) follows in its own PR.
  */
 export function createRegisteredProviders(): FileSourceProvider[] {
-  return [];
+  return [new DaluxBuildProvider()];
 }
--- a/apps/viewer/src/services/sources/source-host.test.ts
+++ b/apps/viewer/src/services/sources/source-host.test.ts
@@ -132,23 +132,11 @@ describe('every provider the viewer actually registers satisfies PLUGIN_API_VERS
   it('registers cleanly against the real host', () => {
     const providers = createRegisteredProviders();
 
-    // `createRegisteredProviders()` is deliberately EMPTY in the change that
-    // lands the contract, the host and the UI — the first provider ships in its
-    // own PR (`@ifc-lite/source-dalux`), so this suite has nothing real to
-    // register yet.
-    //
-    // Asserting emptiness, rather than skipping or loosening to `>= 0`, is the
-    // point: the loop below is vacuous while the list is empty, so without this
-    // line the whole guard would pass silently forever. This assertion FAILS the
-    // moment a provider is added, which forces that PR back here to restore
-    // `assert.ok(providers.length > 0)` — the real anti-vacuity guard — instead
-    // of inheriting a test that cannot fail.
-    assert.equal(
-      providers.length,
-      0,
-      'a provider is now registered: replace this assertion with ' +
-      '`assert.ok(providers.length > 0, ...)` so the per-provider checks below stop being vacuous',
-    );
+    // Restored from the deliberate `=== 0` assertion the contract-only PR
+    // carried: with no providers registered the per-provider loop below is
+    // vacuous, so that PR inverted this to fail the moment one was added —
+    // which is what brought it back here. This is the real anti-vacuity guard.
+    assert.ok(providers.length > 0, 'expected at least one real provider to be registered');
 
     const host = new SourceHost();
     for (const provider of providers) {
--- a/apps/viewer/tsconfig.json
+++ b/apps/viewer/tsconfig.json
@@ -30,7 +30,8 @@
       "@ifc-lite/lists": ["../../packages/lists/src"],
       "@ifc-lite/sandbox": ["../../packages/sandbox/src"],
       "@ifc-lite/sandbox/schema": ["../../packages/sandbox/src/bridge-schema"],
-      "@ifc-lite/plugin-api": ["../../packages/plugin-api/src"]
+      "@ifc-lite/plugin-api": ["../../packages/plugin-api/src"],
+      "@ifc-lite/source-dalux": ["../../packages/source-dalux/src"]
     }
   },
   "include": ["src/**/*", "../../packages/renderer/src/webgpu-types.d.ts"],
--- a/apps/viewer/vite.config.ts
+++ b/apps/viewer/vite.config.ts
@@ -297,6 +297,14 @@ export default defineConfig({
         target: 'http://localhost:3001',
         changeOrigin: true,
       },
+      '/api/dalux': {
+        // Dalux Build's API sends no CORS headers, so browser requests must
+        // go through this same-origin relay (mirrors /api/bsdd below, and
+        // vercel.json's rewrite in production).
+        target: 'https://node1.field.dalux.com',
+        changeOrigin: true,
+        rewrite: (p) => p.replace(/^\/api\/dalux/, '/service/api'),
+      },
       '/api/bsdd': {
         target: 'https://api.bsdd.buildingsmart.org',
         changeOrigin: true,
--- a/packages/source-dalux/README.md
+++ b/packages/source-dalux/README.md
@@ -1,0 +1,35 @@
+# @ifc-lite/source-dalux
+
+Dalux Build (Box) file-source provider for ifc-lite.
+
+Implements `FileSourceProvider` from `@ifc-lite/plugin-api` to browse projects,
+file areas, folders, and files in Dalux Build, and download IFC revisions
+directly into the viewer.
+
+Has no runtime dependencies beyond `@ifc-lite/plugin-api`. Requests go through
+the host-provided `fetch`, and responses are narrowed by hand-written decoders
+in `src/dalux-types.ts` that reject wrong-typed fields rather than coercing
+them, dropping an individually invalid row from a listing instead of failing
+the whole page.
+
+Targets the Dalux **API Identities** auth model (legacy API keys expired
+2026-02-28). The API base URL is fixed at `https://node1.field.dalux.com/service/api`
+— it's not company-specific, so there's no base URL setting.
+
+## CORS
+
+The Dalux API does not send CORS headers, so direct browser fetches to it
+will fail. The viewer routes Dalux requests through a fixed same-origin path
+(`/api/dalux/*`) that the app's own server forwards upstream — a plain
+reverse-proxy rewrite (see `vercel.json` in production, and the Vite dev
+proxy in `apps/viewer/vite.config.ts`), the same pattern already used for
+bSDD/EPSG. The user's own API key is attached client-side and never leaves
+the browser except as part of the proxied request; there is no shared
+server-side secret and no separate relay service to deploy.
+
+## API key storage
+
+The API key is stored in the browser's `localStorage`, unencrypted, with no
+expiry. Anything that can execute script in the page (e.g. the viewer's own
+script panel) can read it. Users who want to revoke access should rotate the
+key in Dalux and use the "forget key" action in Source Settings.
--- a/packages/source-dalux/package.json
+++ b/packages/source-dalux/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "@ifc-lite/source-dalux",
+  "version": "0.1.0",
+  "description": "Dalux Build (Box) file-source provider for ifc-lite",
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "default": "./dist/index.js"
+    }
+  },
+  "scripts": {
+    "build": "pnpm exec tsc",
+    "dev": "pnpm exec tsc --watch",
+    "test": "vitest run"
+  },
+  "dependencies": {
+    "@ifc-lite/plugin-api": "workspace:^"
+  },
+  "devDependencies": {
+    "typescript": "^6.0.3",
+    "vitest": "^4.1.9"
+  },
+  "license": "MPL-2.0",
+  "author": "Louis True",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/LTplus-AG/ifc-lite.git",
+    "directory": "packages/source-dalux"
+  },
+  "homepage": "https://ifclite.dev/docs/",
+  "bugs": "https://github.com/LTplus-AG/ifc-lite/issues",
+  "keywords": [
+    "ifc",
+    "bim",
+    "dalux",
+    "cde",
+    "file-source"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist",
+    "README.md"
+  ]
+}
--- a/packages/source-dalux/src/dalux-types.ts
+++ b/packages/source-dalux/src/dalux-types.ts
@@ -1,0 +1,237 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Response shapes for the Dalux Build API, and hand-written decoders for them.
+ *
+ * This package previously borrowed types and zod schemas from a third-party
+ * `dalux-build-api` client by deep-importing its `src/` internals. That is
+ * gone: deep-importing another package's internals breaks on any upstream
+ * reorganisation, it made a published `@ifc-lite/*` package depend on a
+ * single-maintainer package, and it dragged zod into the viewer's eagerly
+ * loaded bundle for what amounts to a few dozen field checks. The sibling
+ * SharePoint/OneDrive provider talks to Microsoft Graph with nothing but
+ * `fetch` and hand-written narrowing; there is no reason this one needs more.
+ *
+ * The decoders below reproduce the previous schemas' semantics as closely as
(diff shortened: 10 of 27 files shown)


louistrue:
## Pickup notes — state, blockers, and what's left

Recording this so anyone (including future me) can take it over without reconstructing context.

### Merge order

```
#1976 (contract + host + UI)  ──►  #2023 (this, Dalux)  ──►  PR C (msgraph)
```

This PR's base is `feat/source-plugin-api-v2-host`, **not** `main`. When #1976 merges, GitHub retargets this to `main` automatically and the diff collapses to the 27 Dalux files. Do not rebase onto `main` before then — you would pull A's whole diff in here.

### State

Ready as far as I can take it. Typecheck 35/35, `pnpm test` 80/80 tasks, 77 source-dalux tests, 25 source-host, api-surface regenerated, test-wiring and READMEs clean.

**Not yet done on this PR:** no adversarial review pass, and the hosted CodeRabbit has been rate-limited all day (its "pass" is a false green — check the detail text, not the tick). Worth a `coderabbit review --agent --committed --base feat/source-plugin-api-v2-host` from a worktree before merging.

**Never exercised against the real Dalux API.** All 77 tests are against fixtures and a mocked HTTP client. The relay path (`vercel.json` rewrite ↔ `vite.config.ts` proxy) has not been hit end-to-end with real credentials, and #1663's reporter is the natural person to confirm it. That is the highest-value verification remaining and it needs an account, not a code review.

### If #1976 changes under this

Two things in this PR were derived from A and will need re-checking if A's versions move:

- `apps/viewer/src/services/sources/registered-providers.ts` — A ships it returning `[]`; this restores the Dalux entry
- `apps/viewer/src/services/sources/source-host.test.ts` — A asserts the registry is **empty** (deliberately inverted); this restores `assert.ok(providers.length > 0)`. If you touch that test, keep the guard: with an empty registry the per-provider loop below it is vacuous, which is why A inverted it in the first place

### PR C is blocked, and on what specifically

Not on code. It needs a **live MSAL sign-in against a real Entra app registration**, which needs a tenant, not a reviewer.

What is already verified: `@azure/msal-browser@5.17.1`'s `broadcastResponseToMainFrame` has **zero** `window.opener` references and hands the response back over `new BroadcastChannel(id)`, and the dev server really does serve `COOP/COEP: unsafe-none` on `/msal-redirect.html` while the app root stays `same-origin`/`credentialless`. So the mechanism that makes MSAL survive our isolation headers — where Google's GIS does not — is confirmed.

What is **not** verified: an actual token round-trip. `msal-client.ts` is BYO client ID, so there is no shared credential to test with. The gate for PR C is: preview deploy, real registration, confirm `loginPopup` and `acquireTokenSilent` both complete with COOP/COEP active.

Until that runs, the OneDrive decision is provisional. If the bridge fails live, the fallback is @BIMvoice's `server/onedrive` relay from #1971 behind a v2 `RelayDeclaration` — which is why #1971 was closed with that code called out as salvageable rather than dismissed.

### Related

- #1663 — closed by this PR
- #1761 — @bruadam's original, closed as superseded; he asked that one person carry the PRs, hence the `Co-authored-by` here
- #1971 — @BIMvoice's parallel cloud stack, closed; its OAuth core is the basis for a future `packages/source-dropbox` on this contract

BIMvoice:
Ran the adversarial pass you said was missing. Read-only, diffed against `feat/source-plugin-api-v2-host` so this is the 27-file Dalux content, not A's.

**The relay claim holds.** I went at it specifically to break it and could not:

- `dalux-relay.ts:114` uses `next.origin !== target.origin`. `URL.origin` is scheme+host+port as one string, so `node1.field.dalux.com.evil.com` yields a different origin and is rejected — no `startsWith`/`includes` bypass. A protocol downgrade to http fails the same comparison.
- The check is **inside** the redirect loop, so it re-runs on every hop, not just the first. Relative-path redirects resolve via `new URL(location, current)` onto the same origin and are correctly allowed.
- No path where `X-API-KEY` reaches a response body or log — the 502 and the generic upstream-failure branch interpolate only `err.message`/status.
- `resolveUpstreamPath` rejects traversal, absolute and protocol-relative paths, `\0`, backslashes and double-`/` before the version-prefix allowlist, and it decodes before checking, so no encoded-`..` bypass.

Pagination in `http-client.ts` also came out sound: the immediate-echo check at `:220` plus the whole-sweep `Set` at `:296` catches an A→B→A alternation that the echo check alone would miss, `MAX_SWEEP_PAGES` bounds an endless-fresh-bookmark server, and the `nextPage`-without-bookmark and `totalRemainingItems` mismatch cases throw rather than silently truncating. I looked for an off-by-one or a dropped last page and did not find one.

## One finding worth your decision — `provider.ts:281-297`

The file-area sweep's `catch` emits `{ deleted: true }` for every ref in the area regardless of **why** the sweep failed — a genuine 404, a transient 5xx, a `DaluxPaginationError`, or a `SyntaxError` from `response.json()` on a malformed body all take the identical path.

Your own comment reads *"Report its refs as deleted/skipped"* — so the intent was **skipped**, but the signal on the wire is **deleted**. A transient network blip during a poll therefore reports real, still-present files as deleted to whatever consumes that event.

The cache handling right below it is deliberately correct about this — it leaves the baseline untouched so a transient failure retries cleanly next poll. It's only the emitted event that conflates the two. `provider.test.ts:547-582` distinguishes these cases by log-message text only; the code has no branch on cause.

Your call whether `deleted` should be reserved for the confirmed-404 case with something else for sweep-level failure. Happy to implement whichever you pick.

## Smaller, and one thing I ruled out

- **Dev/prod drift is real but narrower than it looks.** `vite.config.ts:64-70` is a raw `http-proxy` passthrough running none of the relay's method/origin/path/header allowlists, so a bad-path request that 400s in prod before touching Dalux gets proxied to the real API with a live key in dev. That's a test-fidelity gap, not an exploitable hole. I initially suspected a redirect leak in dev and **ruled it out**: `source-host.ts:157` sets `redirect: 'error'` on every provider request in both modes, so a forwarded 3xx just makes `fetch` throw. The comment at `vite.config.ts:66` calling it a mirror of the production relay isn't accurate for the security-relevant parts.
- `MAX_REDIRECTS = 3` with `hop <= MAX_REDIRECTS` permits four attempts. Cosmetic; the test already tolerates it.
- `client.get()` calls `response.json()` unconditionally on `response.ok` with no try/catch, so a 200 carrying an HTML error page surfaces as a raw `SyntaxError` — and inside `watchRevisions` it disappears into the finding above. Whether Dalux ever does that needs live traffic to say.
- Worth a line in the PR body: `provider.ts:198` attaches `X-API-KEY` to whatever `downloadLink` Dalux returns, and that is safe **only** because #1976's `isAllowedHost` gates it first (it uses a dotted-suffix boundary, not raw `includes`, so lookalikes fail there too). Correct today, but inherited rather than owned by this PR — if that check ever regresses, this is where it bites.

## Not verifiable from here

Whether `downloadLink` is ever hosted off `*.dalux.com` (a presigned CDN URL would be blocked outright by the host allowlist, not leaked); whether Dalux actually issues the redirects the relay defends against; whether the adversarial pagination shapes occur in practice; and whether a 200-with-non-JSON body is real. All four are mocks asserting our own assumptions — which is exactly the gap you flagged, and it still needs the reporter's account rather than another review.

BIMvoice:
Following through on the #1976 finding — the contract's claim that *"a provider cannot smuggle credentials off-host"* is enforced only inside `ctx.fetch`/`ctx.fetchPublic`, and providers are otherwise plain same-realm JS with unrestricted access to global `fetch`/`XHR`/`WebSocket`/`import()`. Checked whether that gap has teeth in the first real provider. Read-only, diffed `feat/source-plugin-api-v2-host` (5efb03d6) against this branch's tip (cf2d8166), so this is the 27-file Dalux content.

**1. Every network call goes through `ctx.fetch`.** There are exactly two fetch sites in the whole package — `http-client.ts:95` and `http-client.ts:128`, both `this.ctx.fetch(...)`. Grepped `packages/source-dalux/src` and `/test` for `fetch(`, `XMLHttpRequest`, `WebSocket`, `navigator.sendBeacon`, and dynamic `import(`: zero hits outside those two call sites. `console.*` and `localStorage` are likewise absent from the whole package. So for this provider the claim holds in practice — it isn't merely unenforced-but-true, it's actually observed.

**2. Credential handling is clean inside the provider, but the storage underneath it isn't.**
- The API key is pulled once via `ctx.getPreference('apiKey')` (`provider.ts:362`) and only ever placed in the `X-API-KEY` header of the two `ctx.fetch` calls above (`http-client.ts:97`, `:130`).
- `debug()` logging (`http-client.ts:79-82` region) logs `url`, `params`, `status`, `contentType` — never headers, so the key doesn't reach `ctx.log.*`, and the package makes no `console.*` calls at all to leak it that way either.
- `ctx.storage` is used only for a revision-id cache (`provider.ts:311,322-323`), never credentials.
- But: `preferences.ts:20` (`apps/viewer/src/lib/sources/preferences.ts`, from #1976) persists the whole preferences record — API key included — as **plaintext JSON in `localStorage`**, under a predictable, non-namespaced-to-origin key (`ifc-lite-source-prefs:<providerId>`). The package's own README says this out loud (`packages/source-dalux/README.md:32-34`): *"stored in the browser's `localStorage`, unencrypted, with no expiry. Anything that can execute script in the page... can read it."*

That's the sharper version of the #1976 gap for this PR: it's not just that a provider *could* bypass `ctx.fetch` (Dalux doesn't) — any same-realm code at all, provider or not, can read the credential straight out of `localStorage` without ever touching `ctx.fetch`, `ctx.getPreference`, or any permission check. The wrapper protects the *network hop*; it does nothing for the *credential at rest*, which sits in a well-known, unencrypted, page-readable location. Worth deciding whether that's accepted as-is (documented, as it is here) or whether credential-typed preferences deserve something other than raw `localStorage.setItem`.

**3. The allowlist does real work here.** `manifest.ts:13` declares `permissions.network: ['*.dalux.com', 'dalux.com']`, and `source-host.ts:139`'s `isAllowedHost` check runs on every `wrappedFetch` call. Since the provider's only endpoint is `node1.field.dalux.com` (matches the wildcard) and all requests are relayed through the declared `/api/dalux` path, the gate isn't vacuous — a hypothetical bug that pointed `http-client.ts` at some other host would actually be caught by it, not merely by convention.

**Conclusion on the #1976 finding:** currently theoretical for this provider. Dalux goes through `ctx.fetch` for 100% of its network traffic, doesn't reach for raw `fetch`/`XHR`/`WebSocket` anywhere, and the network allowlist is doing real enforcement rather than rubber-stamping. The one place the underlying claim ("cannot smuggle credentials off-host") is currently *not* watertight is the `localStorage` plaintext-at-rest point above — that's a preferences-storage gap inherited from #1976, not something this PR introduced or could have avoided given the current `auth: 'preferences'` contract.

(Not re-raising the sweep `deleted`-vs-skipped, pagination, relay-redirect, or dev/prod-proxy findings already posted above — this is scoped to the #1976 follow-through only.)

BIMvoice:
## Triage: the four open Codex threads on `cf2d8166`

Read-only pass over `packages/source-dalux` at the branch tip (`cf2d8166f2b20e589f8f3d720d29b0683c82ec22` — still the head, nothing newer). No demo harness; the docs check was executed for real, threads 2–4 are traced through the host in `apps/viewer`.

| # | Thread | Verdict |
|---|---|---|
| 1 | `package.json:4` — regenerate the published-package index | **CONFIRMED** |
| 2 | `provider.ts:315` — compare the first poll against the supplied revision | **CONFIRMED** |
| 3 | `provider.ts:295` — don't turn transient area failures into deletions | **CONFIRMED** |
| 4 | `provider.ts:129` — preserve parent IDs across folder pages | **CONFIRMED** |

---

### 1. `packages/source-dalux/package.json:4` — CONFIRMED

`pnpm docs:check-generated` on `cf2d8166`, verbatim:

```
> ifc-lite@4.2.1 docs:check-generated
> node scripts/docs/generate-docs-sections.mjs --check

❌ Stale generated region "package-index" in docs/api/typescript.md:

   -
   + | [`@ifc-lite/source-dalux`](https://www.npmjs.com/package/@ifc-lite/source-dalux) | Dalux Build (Box) file-source provider for ifc-lite |
   +

❌ 1 generated doc region out of date.
Run `pnpm docs:generate` and commit the result.

 ELIFECYCLE  Command failed with exit code 1.
```

Exit status `1`. Same command on the base branch (`feat/source-plugin-api-v2-host`, `5efb03d6`) exits `0` — `✅ All 4 generated doc regions are up to date` — so the staleness is introduced by this PR, not inherited.

Cause: `packages/source-dalux/package.json` has no `"private": true` and carries `publishConfig.access: public`, so the generator picks it up. Fix is `pnpm docs:generate` + commit (AGENTS.md "Documentation upkeep").

---

### 2. `provider.ts:315` — first poll never compares against `ref.revisionId` — CONFIRMED

```ts
// provider.ts:310-315
const latestRevisionId = currentRevisionId(match);
const cached = await ctx.storage.get(cacheKey);
if (cached && cached !== latestRevisionId) {
  events.push({ fileId: ref.fileId, latestRevisionId, previousRevisionId: cached });
}
pendingSets.set(cacheKey, latestRevisionId);
```

`ref.revisionId` is ignored entirely — the only comparison operand is the cache. The cache is written *only* here, so before the first successful poll of a given `(project, area, file)` there is no entry.

That `ref.revisionId` is exactly the right operand, host-side:

- `apps/viewer/src/components/sources/SourcesPanel.tsx:143-149` — the tag is minted at load with `f.currentRevisionId`, i.e. `mapping.ts:toSourceFile` → `currentRevisionId(file)`.
- `apps/viewer/src/lib/sources/syncSourceModel.ts:199-205` — same on re-sync, with `latestFile.currentRevisionId`.
- `apps/viewer/src/lib/sources/revisionWatch.ts:62-67` — the ref handed to `watchRevisions` carries `revisionId: tag.revisionId`.

So `ref.revisionId` is produced by the *same* `currentRevisionId()` the poll compares against — directly comparable, no format mismatch.

Failure sequence (all host-side, no provider mocking needed):
1. User loads a Dalux model at rev A. No `rev:` key is written — the load path never touches `ctx.storage`.
2. Upstream publishes rev B.
3. First watch pass runs (`SourcesPanel.tsx:52-71`, gated to one pass per panel mount and `REVISION_WATCH_MIN_INTERVAL_MS = 5 * 60_000`, `revisionWatch.ts:19`). `cached` is `undefined` → no event, and line 315 stores **B**.
4. Every later poll compares B to B. The rev A → rev B update is suppressed permanently; the user is never told to Sync, while the loaded model is still rev A.

Storage is `localStorage`-backed (`apps/viewer/src/services/sources/host-storage.ts:20-70`) so this isn't a per-session quirk — it's a one-shot permanent loss per file, and it recurs after "Forget" (`clearAllSourceData`, `SourcesPanel.tsx:74-79`) wipes the namespace while the model stays loaded.

Why the tests miss it: every `watchRevisions` case in `packages/source-dalux/test/provider.test.ts` seeds the cache first (`await ctx.storage.set('rev:proj1:fa1:f1', 'rev-1')`, lines 481-482, 497, 518, 539). There is no empty-cache case.

Suggested shape: `const baseline = cached ?? nonEmptyString(ref.revisionId);` then compare on `baseline`, still guarding `baseline !== LATEST_REVISION` so the sentinel isn't reported as a previous revision.

---

### 3. `provider.ts:295` — a transient area failure is reported as deletion — CONFIRMED

```ts
// provider.ts:281-297
} catch (err) {
  ctx.log.warn('Dalux: file-area sweep failed during watchRevisions, skipping its refs this poll', { ... });
  for (const ref of areaRefs) {
    events.push({ fileId: ref.fileId, latestRevisionId: LATEST_REVISION, deleted: true });
  }
  continue;
}
```

The `try` covers `fetchAllPages` + decode (lines 272-280), so *any* failure lands here — a 500, a timeout, an aborted relay, a Dalux pagination error — not just the 404 the comment has in mind. The catch body's own comment says "skipping its refs this poll", but the emitted event says the opposite: `RevisionEvent.deleted` is documented at `packages/plugin-api/src/types.ts:306-307` as "*The file is gone upstream (deleted, or moved out of the user's reach)*". Existence was never checked.

User-visible: `apps/viewer/src/components/sources/SourcesPanel.tsx:60-65`

```ts
const deleted = updates.filter((u) => u.event.deleted).length;
...
if (deleted > 0) parts.push(`${deleted} source file${deleted === 1 ? ' is' : 's are'} gone upstream`);
toast.info(`${parts.join('; ')} — use Sync to update.`);
```

One transient 500 on a file area with N loaded models → a toast claiming N files are gone upstream, for files that are fine. The genuine-deletion path (line 304-307) is a *different* branch, correctly keyed on a completed sweep that didn't contain the file — so the fix is just to emit nothing (the warn already records it), or a non-`deleted` skip signal, rather than reusing the deletion event. Note the cache handling in this path is already right: the `continue` means nothing is buffered, so the baseline survives (asserted at `test/provider.test.ts:584`).

---

### 4. `provider.ts:129` — cross-page parent is never re-nested — CONFIRMED

```ts
// provider.ts:114 and 126-129
const folderIdsThisPage = new Set(folders.map((folder) => folder.folderId));
...
const resolvedParentId =
  !parentFolderId || parentFolderId === fileAreaId || folderIdsThisPage.has(parentFolderId)
    ? (parentFolderId ?? fileAreaId)
    : fileAreaId;
```

The set is built from **one** `fetchPage` (line 106-112), so a child on page 1 whose parent lands on page 2 fails `folderIdsThisPage.has(...)` and is rewritten to the file-area root at line 129 / 135-137.

The comment at lines 122-125 says it "re-nests correctly once the host has walked that later page and revisits it". The host does not revisit. `apps/viewer/src/components/sources/useSourceCatalogSync.ts:359-364`:

```ts
const page = await fetchContainers(area.projectId, key, current.cursor, signal);
if (requestGenRef.current !== gen) return;
const merged: PagedItems<SourceContainer> = {
  items: [...current.items, ...page.items],
  cursor: page.cursor,
};
```

Pure append. Page 1's `SourceContainer` objects — `parentId` already rewritten to the area root by the provider — are carried through untouched; nothing recomputes them once page 2 arrives. The only other path that fetches page 1 is `syncFileArea` (`useSourceCatalogSync.ts:188`), which resets the whole map and re-fetches page 1 in isolation — producing the identical rewrite, since page 1's contents are unchanged. `openContainer` (line 263) is a no-op for this provider: `capabilities.containerListing === 'flat-subtree'` (`manifest.ts:33`) makes `needChildren` false at line 270-271.

It also persists. `maybePersist` (line 139, called from `loadMoreFolders` at line 367) writes the merged list to `localStorage` once the last page's cursor is `undefined`, so the flattened tree is cached and re-applied by `applyCachedCatalog` (line 163-175) on the next session.

User-visible: for any file area whose folder list exceeds one Dalux bookmark page, a nested folder whose parent happens to sit on a later page appears permanently at the file-area root in the browser tree instead of under its parent. Files still resolve (`toSourceFile` keys off the file's own `folderId`), so it's a wrong tree shape rather than lost files.

Cheapest fix that keeps one-page-at-a-time paging: accumulate the seen folder ids across pages in `ctx.storage` (or resolve the parent lazily in the host from the flattened set it already holds) instead of scoping the set to a single page.

---

*Read-only triage — nothing pushed, nothing resolved. The two findings already on record for this chain (the `ctx.fetch` sandboxing gap; plaintext credentials in `localStorage` with the unused `'interactive'` auth kind) are deliberately not repeated here.*

BIMvoice:
Verified two of the four independently before relaying, and they hold.

**Thread 1** — `docs/api/typescript.md` on this branch contains **zero** occurrences of `source-dalux`, and `packages/source-dalux/package.json` has no `private: true`, so it is a published package that belongs in the generated index. The differential the triage ran is the useful part: `pnpm docs:check-generated` exits **1** on this branch and **0** on the base (`feat/source-plugin-api-v2-host`, `5efb03d6`, "✅ All 4 generated doc regions are up to date"). So the staleness originates here, not inherited from #1976 — worth stating explicitly because on a stacked PR the natural first guess is the base.

**Thread 4** — the crux is whether the host ever revisits an earlier page to re-parent. It does not: `apps/viewer/src/components/sources/useSourceCatalogSync.ts:362` and `:396` are both `items: [...current.items, ...page.items]`, a pure append. Page 1's already-rewritten `parentId` is carried through untouched, so the comment at `provider.ts:122-125` claiming the host "revisits" and re-nests is describing behaviour that does not exist. That makes this a prose-vs-code drift as well as a defect — the comment is what would stop the next reader from spotting it.

## One additional finding, adjacent to thread 2

Not one of the four, and I am raising it separately because it is a contract question rather than a provider bug.

`packages/source-fixture/src/conformance/watch.ts:33-34` asserts:

```ts
const result = await provider.watchRevisions!(ctx, refs, undefined, undefined);
expect(result.cursor, 'watchRevisions must return a cursor when capabilities.changeDetection is true').toBeDefined();
```

`packages/source-dalux/src/manifest.ts:44` declares `changeDetection: true`, and `provider.ts:327` returns `{ events }` with no cursor — deliberately, per the comment: *"No delta endpoint to resume from — every call is a fresh poll of the given refs, so there is no cursor to hand back."*

That reasoning is sound for an API with no delta endpoint. The problem is that **the conformance suite is not run against this provider**, so nothing surfaces the mismatch: `packages/source-dalux/package.json` has `test: vitest run`, no dependency on `@ifc-lite/source-fixture`, and none of its four test files import the suite. The first real provider on the contract does not execute the contract's own conformance tests.

Two ways to read it, and it is your call which:
- the assertion is too strict — polling providers legitimately have no cursor, and `changeDetection` should not imply resumability; or
- the capability flag is overloaded and Dalux should declare something narrower.

Either way the useful action is wiring the suite in, since that is the thing that would have asked the question. Happy to do that as a separate PR if you want it — I have not touched it here.

PR #2024, opened 2026-08-03
fix: free the geometry WASM handle at the remaining leak sites (#1959)

BIMvoice:
Continues #1959. **Stacks conceptually on #2022**, which took the two viewer P0s (`PlaygroundViewer`, `playground-dispatcher`) — no file overlap, so the two rebase cleanly in either order.

## Fixed — six sites

`mcp/tools/clash.ts`, both processors in `export/lod1-generator.ts` (primary **and** fallback), `cli/diagnose-geometry.ts`, `cli/extract-entities.ts` (triage), `cli/gym.ts`.

All `try/finally`, not a trailing `dispose()` — none of these leaked only on the happy path. The exits that a trailing call would have missed: an abort throw and a zero-mesh throw in the MCP clash tool, the forced-meshing-failure catch in the LOD1 fallback, the GlobalId-not-found throw under `--product`, and gym's `close` command returning straight out of the REPL loop.

## Deliberately not fixed — please sanity-check these two calls

**`cli/commands/clash.ts`** holds a module-level `sharedProcessor` reused across every run in the process. Exactly one handle ever exists, so it's bounded rather than accumulating — the issue's own "process-lifetime singleton is not a leak" carve-out. Adding a `finally` here would actively break the next run. Left untouched.

**`hooks/useIfcLoader.ts`** — still yours to call. Its handle goes to `IfcParser.parseColumnar` via `getApi()`, so disposal must wait for both the parse promise and `processAdaptive` to settle. Not attempted here.

`headless-backend.ts:645`, listed in the issue, no longer constructs a processor on current `main`.

## On the P2 CLI sites

I didn't sweep these reflexively. A one-shot command whose process exits immediately reclaims everything anyway, so a `finally` there is ceremony. `diagnose-geometry`, `extract-entities` and `gym` are fixed because each is reachable more than once per process from a test harness or the gym REPL; the singleton above is skipped for the opposite reason.

## Evidence

Removing the disposal fails **3** tests in mcp, **3** in export and **4** in cli, each against its own site; restored, **107 / 291 / 251** pass. Typecheck 34/34.

One implementation note: gym's processor is boxed in a ref rather than a plain `let`. TS's control-flow analysis doesn't track the assignment inside `getProcessor`, so with a `let` it still considers the variable `null` at the `finally` and fails with TS2339 on `never`. I compiled both forms rather than asserting that.

Changeset: patch for `@ifc-lite/mcp`, `@ifc-lite/export`, `@ifc-lite/cli`, `create-ifc-lite`.

Draft until you've looked, particularly at the two skips.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed resource leaks during geometry processing across clash detection, exports, CLI diagnostics, entity extraction, and interactive sessions.
  - Improved cleanup when processing is cancelled, interrupted, fails, or exits early.
  - Updated the generated React/WebGPU template to release resources during cancelled initialization and component cleanup.

- **Tests**
  - Added coverage confirming reliable cleanup across successful and failure scenarios.

### DIFF
--- a/.changeset/dispose-geometry-processor-remaining.md
+++ b/.changeset/dispose-geometry-processor-remaining.md
@@ -1,0 +1,15 @@
+---
+'@ifc-lite/mcp': patch
+'@ifc-lite/export': patch
+'@ifc-lite/cli': patch
+'create-ifc-lite': patch
+---
+
+Fixed the remaining `GeometryProcessor` WASM handle leaks tracked in issue #1959, beyond the viewer P0 sites fixed separately. Each site now frees its handle in a `try/finally` covering every early-return and throw path, not just the happy path:
+
+- `@ifc-lite/mcp`: `clash_check` / `clash_matrix`'s model meshing (long-lived MCP server process, one handle per never-before-clashed model).
+- `@ifc-lite/export`: `generateLod1`'s primary and fallback processors, including the forced-meshing-failure fallback path.
+- `@ifc-lite/cli`: `diagnose-geometry`, `extract-entities --detect`, and `gym`'s lazily-created clash-channel processor — all reachable more than once per process from a long-lived host (a test harness, a REPL session) even though each is a one-shot CLI command in normal use.
+- `create-ifc-lite`: the generated React + WebGPU template's mount effect now disposes its `GeometryProcessor` on both the mid-init cancellation path and on unmount, so scaffolded projects don't inherit the leak.
+
+`apps/viewer/src/hooks/useIfcLoader.ts` is intentionally untouched: its processor's WASM handle is shared with `IfcParser.parseColumnar` via `getApi()`, and disposal there needs a design decision (owned-and-reused vs. freed-per-call) that has not been made yet.
--- a/packages/cli/src/commands/diagnose-geometry.test.ts
+++ b/packages/cli/src/commands/diagnose-geometry.test.ts
@@ -6,6 +6,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
 import { readFile, unlink } from 'node:fs/promises';
+import { GeometryProcessor } from '@ifc-lite/geometry';
 import { isExpressId, filterWorstHosts, diagnoseGeometryCommand } from './diagnose-geometry.js';
 import { logger } from '../logger.js';
 
@@ -146,6 +147,32 @@ describe('diagnoseGeometryCommand', () => {
       expect(stderrSpy.mock.calls.some((c) => String(c[0]).includes('no entity found with GlobalId'))).toBe(true);
     } finally {
       exitSpy.mockRestore();
+    }
+  }, 30_000);
+
+  it('disposes the GeometryProcessor WASM handle on the success path (#1959 P2 leak)', async () => {
+    const disposeSpy = vi.spyOn(GeometryProcessor.prototype, 'dispose');
+    try {
+      await diagnoseGeometryCommand([SAMPLE_IFC, '--json']);
+      expect(disposeSpy).toHaveBeenCalledTimes(1);
+    } finally {
+      disposeSpy.mockRestore();
+    }
+  }, 30_000);
+
+  it('disposes the GeometryProcessor WASM handle even when --product resolution fails (throw path)', async () => {
+    const disposeSpy = vi.spyOn(GeometryProcessor.prototype, 'dispose');
+    const exitSpy = vi.spyOn(process, 'exit').mockImplementation(((() => {
+      throw new Error('process.exit called');
+    }) as unknown) as (code?: number) => never);
+    try {
+      await expect(
+        diagnoseGeometryCommand([SAMPLE_IFC, '--product', 'not-a-real-guid-value']),
+      ).rejects.toThrow('process.exit called');
+      expect(disposeSpy).toHaveBeenCalledTimes(1);
+    } finally {
+      exitSpy.mockRestore();
+      disposeSpy.mockRestore();
     }
   }, 30_000);
 });
--- a/packages/cli/src/commands/diagnose-geometry.ts
+++ b/packages/cli/src/commands/diagnose-geometry.ts
@@ -63,31 +63,36 @@ export async function diagnoseGeometryCommand(args: string[]): Promise<void> {
   const bytes = new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength);
 
   const gp = new GeometryProcessor();
-  await gp.init();
-  let diag = gp.diagnoseGeometry(bytes);
+  let diag: ReturnType<typeof gp.diagnoseGeometry>;
+  try {
+    await gp.init();
+    diag = gp.diagnoseGeometry(bytes);
 
-  if (diag && (productArg !== undefined || typeArg !== undefined)) {
-    let productId: number | undefined;
-    if (productArg !== undefined) {
-      if (isExpressId(productArg)) {
-        productId = parseInt(productArg, 10);
-      } else {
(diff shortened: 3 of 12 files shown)


louistrue:
#2022 is merged (`2e320ac5`) — adversarial review came back clean, and the thing that convinced me was that your five tests actually discriminate: a naive `dispose()` appended after the last statement passes the two success tests and fails the other three. That's the property most leak-fix tests don't have.

Answering the two you flagged, since you asked for a sanity-check rather than assuming.

**`cli/commands/clash.ts` — your call is right, and I verified rather than taking it.** `sharedProcessor` is a module-level `let` (`:45`) memoized behind `getProcessor()` (`:48-53`), so exactly one handle exists per process and it is created lazily on first use. A `finally { dispose() }` at any call site would free the handle the *next* call then reuses — you'd turn a bounded singleton into a use-after-free. Leave it. If it ever needs cleanup it belongs in a process-exit hook, not a call site.

**`useIfcLoader.ts` is genuinely a design question, not a deferral.** I checked this while reviewing #2022 and reached the same place you did: the processor's raw wasm handle goes to `IfcParser.parseColumnar` (`:1171`) *and* drives `processAdaptive` (`:1317`), so disposal needs both consumers coordinated — and if the answer is "the loader should own and reuse one processor," that is a lifetime redesign rather than a `try/finally`. Right to keep it out of a leak-fix PR.

@louistrue that one is yours: it is #1959's highest-blast-radius site (fires on every load, primary and federated) and the fix shape depends on whether the loader should own a processor for its lifetime or construct per load.

## On the six this PR does fix

The exits you list are the right ones to have found — the abort throw and zero-mesh throw in the MCP clash tool, the forced-meshing-failure catch in the LOD1 fallback, the GlobalId-not-found throw under `--product`, and gym's `close` returning straight out of the REPL loop. Those are exactly the paths a trailing `dispose()` misses, and catching the **fallback** processor in `lod1-generator.ts` as well as the primary is the kind of thing that gets missed.

I will run a proper adversarial pass once you undraft. Two things I will be checking, so you can pre-empt them if you like:

1. **Whether each test would fail with the `finally` removed** — per path, not per site. Six sites with one happy-path test each would be weaker than #2022's five, despite covering more ground.
2. **Whether disposing the processor is sufficient at each site**, or whether any of them also owes a `MeshCollection` / `MeshDataJs` / `clearPrePassCache()` release on the same path. AGENTS.md lists those separately, and for #2022 I traced that the geometry package already frees them internally — worth confirming the same holds for the CLI and MCP paths, which use different entry points.

Undraft when you're ready; no need to wait on us.

BIMvoice:
Thanks for verifying both skips rather than taking them — and for the `sharedProcessor` point. "You'd turn a bounded singleton into a use-after-free" is a sharper statement of it than mine, and a process-exit hook is the right home if it ever needs one.

Pre-empting your two checks.

## 1. Per-path discrimination, not per-site

Each site's disposal was mutated **individually**, not as a batch, and the failures land per path rather than one-per-site:

| site | disposes | tests that fail when removed |
|---|---|---|
| `mcp/tools/clash.ts` | 1 | 3 |
| `export/lod1-generator.ts` | 2 (primary + fallback, mutated independently) | correct subset each time — 2 and 2 |
| `cli/diagnose-geometry.ts` | 1 | 2 |
| `cli/extract-entities.ts` | 1 | 2 |
| `cli/gym.ts` | 1 | 2 |

The lod1 pair is the one I'd point at: mutating the primary and the fallback separately fails different subsets, so the two aren't covered by one shared assertion. Restored: 107 / 291 / 251 pass.

The property you named for #2022 — a trailing `dispose()` passing the success tests and failing the rest — holds here too, because the discriminating tests are the throw paths: the abort and zero-mesh throws, the forced-meshing-failure catch, the GlobalId-not-found throw, and gym's `close` returning out of the loop.

## 2. `MeshCollection` / `MeshDataJs` / `clearPrePassCache`

Checked rather than assumed, since you're right that CLI and MCP use different entry points. **None of the five sites reference those types at all** — grep returns nothing for `MeshCollection`, `MeshDataJs` or `clearPrePassCache` across all five. They all call `processor.process()`, so the releases stay inside the geometry package, and there they're already in `finally` blocks: `out.free()` at `packages/geometry/src/index.ts:1461-1463` and `api.clearPrePassCache?.()` at `:483-486`, the latter commented "Always release the pre-pass cache — even if processGeometryBatch throws."

So the same conclusion as #2022 holds, and for the same structural reason rather than by coincidence: these callers never own those handles, only the processor.

One caveat I'd rather state than have you find: `PlaygroundViewer.test.ts` covers cancelled, empty-mesh and success, but not `process()` throwing mid-flight. Structurally that path is covered — the `finally` wraps both `init()` and `process()` — but it isn't pinned by a test. Say the word if you want it added.

Undrafting now.

louistrue:
Adversarial pass done. **Clean — merging.** The disposal engineering holds at all six sites and the tests meet the bar #2022 set: they discriminate per escape path, not per site.

Verified rather than assumed:

- **All six sites**: `try` opens immediately after construction with no `await` or early return in the gap, and every escape you listed exists and is inside it — the abort and zero-mesh throws in the MCP tool, the GlobalId-not-found `fatal()` under `--product`, the triage `process` wrap, and gym's `close` returning from inside the `for await`.
- **The LOD1 double-processor is right.** The primary's `finally` runs *before* the throw reaches the catch that constructs the fallback, so the two are never live at once and a throw between them can't strand the first. `if (!glb) throw e` rethrows the original correctly, and the `__forceMeshingErrorForTest` hook throws before the primary exists, so that path holds exactly one handle.
- **Disposing the processor is sufficient at all six.** None touch `buildPrePassOnce`/`processGeometryBatch` raw — they go through `process`/`processAdaptive`/`diagnoseGeometry`/`exportGlbFromMeshes`, and both `collectMeshesViaPrePass` and the streaming generator call `clearPrePassCache()` in their own finallys while `convertMeshCollectionToBatch` frees `MeshDataJs`/`MeshCollection` after copying out. So the MCP `meshCache` outliving `dispose()`, `extract-entities` iterating `res.meshes` after the finally, and lod1 using `glb` after dispose are all safe — JS copies, not wasm views.
- **Generator abandonment** lands in the generator's own `clearPrePassCache` finally; gym's `for await` is over readline, no wasm generator.
- Swept every remaining `new GeometryProcessor` repo-wide: all others already carry dispose. The "remaining leak sites" claim holds, modulo your two documented exceptions.

## The two open threads — both real, neither blocking, my reasoning

**@greptile-apps on `react.ts:284`.** Correct that a `createViewer()` rejection leaves `processorRef.current` populated with `ready` false. But it's **bounded**: the effect's cleanup disposes it on unmount, so it's one handle held by an already-failed viewer, not accumulation. Same shape as the `sharedProcessor` carve-out you made in `cli/commands/clash.ts`, one scope down. Worth tidying since this is scaffold code people copy, not worth holding the PR.

**@coderabbitai on `diagnose-geometry.ts:96`** — dispose before `fatal()`. Technically right and **moot in production**: `fatal()` calls `process.exit(1)` (`output.ts:99-102`), which skips the `finally` entirely, so dispose never runs on that path either way and process death reclaims the handle. Worth knowing the test only exercises that path because `process.exit` is mocked to throw — the test is fine, but it pins mock behaviour rather than production behaviour. Flagging so nobody later reads it as proof the real path disposes.

## Smaller

- Commit message says "4 in cli"; I count 6 dispose-asserting CLI tests across 3 files. Undercounts, so the suite is stronger than claimed.
- gym: no dispose assertion for the EOF-exit or throw-out-of-loop paths (structurally covered by the same finally). And if `p.init()` throws inside `getProcessor`, `p` is dropped undisposed — safe only because the bridge self-resets on init failure, which is pre-existing.
- `extract-entities.ts` is 518 lines, still over the ~400 guideline (pre-existing, +9 here).

Merging. Pick up the template tidy in whatever you touch next.

PR #2025, opened 2026-08-03
fix(processing): canonicalise opening order so the world hash stops moving (#2019)

### LINKED ISSUE #2019: World geometry hash is not invariant to statement order for opening-cut elements
Found by the content-matching fixture in #2018, on its first scored run. Not a matcher bug — a geometry-hash one.

## The finding

For elements cut by openings, **the world geometry hash changes when the order of statements in the file changes**, with nothing else different.

Isolated with controls, over 750 hashed elements:

| perturbation | hashes changed |
|---|---|
| same bytes re-hashed | **0 / 750** |
| scanner round trip (same express ids, same order) | **0 / 750** |
| **statement order reversed, nothing else** | **48 / 750** |

The first two controls matter: they rule out nondeterminism in the hash itself and in the harness's rewriting, leaving statement order as the only variable.

## Scope

Only `IfcWall`, `IfcWallStandardCase` and `IfcCovering` — i.e. elements that go through opening subtraction. AABB deltas of **1e-6 to 1e-5 m**, which is far below any modelling tolerance but crosses the hash's **1 mm quantization grid** for some vertices, flipping the quantized value.

## Why it matters

The geometry hash is the identity signal for "did this element's shape or position change". A re-export that merely reorders its statements — which exporters do routinely, and which is precisely the scenario #1891 exists to survive — produces a false **"changed"** in Compare for the affected elements.

It also means the hash is not a reliable content address for opening-cut geometry across files, only within one parse of one file.

## Likely mechanism, not yet confirmed

Opening subtraction is a boolean whose result depends on the order operands are accumulated, and floating-point addition is not associative. If the cutters are gathered in statement order rather than a canonical one, reordering the file reorders the accumulation and moves the last bits of the result. That is consistent with the magnitude (1e-6 m is roughly f32 epsilon at building scale) and with the class restriction (only elements that get cut).

Worth confirming before fixing — the fix differs depending on whether it is operand order, an unordered container iterated in insertion order, or something in the CSG kernel's own accumulation.

## Suggested acceptance

Reversing statement order must change **0 of 750** hashes, matching the two controls that already pass. The harness in #2018 measures exactly this and is the natural regression test — it is already wired to a weekly lane.

## Related

The same run surfaced a second, separate limitation: the data fingerprint cannot separate type objects that differ only in `Tag` (eight `IfcFurnitureType` all named `800 mm` in Duplex). That one is arguably correct behaviour on the evidence available and is not filed as a bug, but it is the other half of why that run missed its recall floor.

BIMvoice:
Closes #2019.

## Mechanism — confirmed, since you asked for that before a fix

Your three candidates were operand order, an unordered container iterated in insertion order, or the CSG kernel's own accumulation. **It is the first two, feeding the third.**

`void_index` accumulates each host's openings in *statement* order, and aggregate propagation extends those lists in hashmap-iteration order — neither is a property of the model. The void kernel then subtracts them sequentially whenever they can't join the disjoint-cutter batch or the coaxial-union pass, and each pass snaps f64→f32. That operation is not associative, so reordering the operands lands on a numerically different, geometrically equivalent mesh — and the sub-millimetre delta crosses the 1 mm quantisation grid for some vertices. Consistent with both your observations: the magnitude (~f32 epsilon at building scale) and the restriction to classes that actually get cut.

## The fix, and where it deliberately is not

Sort each host's openings by express id, so accumulation order follows the model rather than the byte layout. Three lines, at the **input**.

Not in the kernel. Its non-associativity is a genuine property of exact sequential CSG, and changing accumulation there has a far larger blast radius than this issue warrants. A second test, `sequential_cut_order_is_not_associative`, documents that property as expected-to-pass so a future reader doesn't mistake it for the defect — it passes with or without the fix, by design.

## Tests

Both drive the real `resolve_prepass` entry point rather than a hand-built index, and force the sequential exact kernel via the env gates so the disjoint-batch fast path can't mask the ordering.

Fixture is **4KB, hand-authored** — one wall, two mutually overlapping openings — rather than the corpus model, so it runs in milliseconds with no fixture download. Removing the sort fails the invariance test and only it.

`cargo test -p ifc-lite-processing` and `-p ifc-lite-geometry`: **0 failing suites**. Processing lib 57/57.

## Two things for you to weigh

**1. This is a one-time hash change, by construction.** Canonicalising necessarily picks *an* order. Elements whose `IFCRELVOIDSELEMENT` statements were already ascending keep their hash; everything else hashes differently once and is order-stable thereafter. There's no fix that both stabilises order and preserves every current value — worth saying out loud since the hash is a content address. Called out in the changeset.

**2. The ratchet moved 772 → 781** for `prepass.rs`. It sat *exactly* at budget, so any fix required a bump; I trimmed the comment twice to keep it to nine lines and recorded the justification in the allowlist. Say the word if you'd rather see the file split instead.

## Not verified here

Your acceptance criterion is 0 of 750 hashes changed against the #2018 harness on the real corpus. I've proven order-invariance on a minimal fixture and identified the mechanism, but I haven't run that harness — it needs the corpus fixtures. That run is the real confirmation and it's still worth doing before you close this.

## Summary by CodeRabbit

- **Bug Fixes**
  - Made world geometry hashes consistent when opening relationships appear in different statement orders.
  - Ensured deterministic processing of multiple openings on the same host element.
  - Existing models may receive a one-time hash update when their openings were previously ordered differently.

- **Tests**
  - Added regression coverage for overlapping openings and reordered void relationships.


louistrue:
Verified this independently rather than reading the description. It holds up.

**The fix bites.** Removing the `sort_unstable()` from `prepass.rs` fails `world_geometry_hash_is_invariant_to_reordered_void_rel_statements` with the diagnostic naming the exact cause — *"resolve_prepass produced a DIFFERENT opening order for the wall from a file that only reordered its IFCRELVOIDSELEMENT statements"*. Both tests pass with the fix in place.

**The second test is the part I want to call out as good practice.** `sequential_cut_order_is_not_associative` passes with *and* without the fix, and you said so in the PR rather than letting two green tests imply two units of coverage. Pinning a property of the kernel — so a later reader doesn't mistake non-associativity for the bug — is genuinely useful, and labelling it honestly is the difference between documentation and padding. I confirmed it passes in both states.

**The diagnosis matches the code.** `void_index` accumulates in statement order and aggregate propagation extends in hashmap-iteration order, neither being a property of the model; sequential void cuts snap f64→f32 per pass and are therefore not associative, so a reorder lands on a numerically different, geometrically equivalent mesh whose sub-millimetre delta crosses the 1 mm quantisation grid. That explains both observations in #2019 — the magnitude sitting at f32 epsilon for building-scale coordinates, and the restriction to classes that actually get cut.

**Fixing at the input rather than in the kernel is the right call.** Non-associativity is a real property of exact sequential CSG, and canonicalising accumulation inside the kernel would have a blast radius wildly out of proportion to this issue. Nine lines at the boundary, where the order is genuinely arbitrary, is the correct level.

One thing worth checking before undraft, and it's the same question that bit #1970: `resolve_prepass` is one entry point, but is it the *only* place `void_index` is populated or extended? If the sharded/streaming discovery path builds its own, it needs the same canonicalisation, and a test that drives only `resolve_prepass` wouldn't notice. If you've already confirmed there's a single producer, say so and I'll stop asking.

Undraft when ready — no need to wait on us.

louistrue:
Full adversarial pass done, plus a CodeRabbit CLI run (**0 findings**, 5 files). The nine production lines are correct, minimal, at the right site, and covered by a genuinely non-tautological test. Collateral surfaces all check out — I'll list them because "nothing else broke" is worth as much as the fix here:

- **Cache needs no `FORMAT_VERSION` bump.** `grep geometryHash packages/cache/src` → zero hits: the cache persists meshes, never the hash. A stale cached model can only serve a pre-fix *mesh* differing by sub-mm jitter, never a pre-fix *hash*. 13 stays correct.
- **Determinism manifests stay valid** — the pinned fixture's two voided hosts have one opening each (`determinism.rs:87,141`), so the sort is a no-op there.
- **B3.5 golden untouched** — `world-gym/generator.mjs` emits no `IFCRELVOIDSELEMENT` at all.
- **All four pipeline entry points route through the one sort site** (native `processor/mod.rs:845`, wasm `gpu_meshes/prepass.rs:566`, `prepass_sharded.rs:273`, `styling/prepass.rs:87`), and the hash is per-element so cross-element iteration order is irrelevant. The legacy `VoidIndex` in `geometry/void_index.rs` is unsorted but has no production callers.
- **No double-cut risk** — propagation dedups at `void_index.rs:66`, and a corpus survey found zero hosts with duplicate openings. (`openings.dedup()` after the sort would be free hardening.)
- **Blast radius measured:** 283 of 5,577 voided hosts across the committed corpus (~5%) have non-ascending opening order — TUN32 162, NUS skolebygg 45, Office_A 25, DigitalHub 10. Their hashes change once. Your changeset discloses this honestly.
- Clippy at CI scope green; the new test binary 2/2; ratchet allowlist entry matches `wc -l` exactly.

Two things before this merges.

## 1. The changeset overclaims, and I'd rather fix the sentence than have it in the changelog

> order becomes a property of the model rather than of the byte layout

Express ids **are** byte layout. What the fix actually buys is a hash stable under *id-preserving* reorders and monotone renumbers (merge offsets) — which is exactly what #2019 filed, so `Closes #2019` is correct. Its isolation table is explicit that the failing perturbation was "statement order reversed, nothing else", with the scanner round-trip control confirming ids were preserved.

But #2018, where the finding came from, describes its real head revision as "every IfcRoot re-GUIDed **and every express id permuted**". An id permutation that flips two openings' relative order defeats this sort — and your own `sequential_cut_order_is_not_associative` proves it, since it shows the same physical openings in a different sequence produce a different hash.

So the scope is right and the limitation is real. Please reword to what it does, and I'd open a follow-up for an id-independent canonical key (opening GlobalId, or a geometric key) — that's the version that survives a cross-tool re-export.

I'm asking because I held #1970 this week over a changeset claiming something the fixture contradicted. Same standard, smaller gap.

## 2. #2019's own acceptance criterion hasn't been run

The issue states the reversal perturbation must go 0/750. That corpus run hasn't happened — and it *can't* from this branch: the harness commit `d42fbf1c` is on main but not an ancestor here, so main has to come in first. You flag this honestly in the PR; I'm restating it because it's the issue's stated bar, not a nice-to-have.

## Two nits

- **`force_sequential_exact_kernel`'s SAFETY comment is factually wrong.** It claims "single-threaded within this test binary" — the two `#[test]` fns run on concurrent threads by default, so two threads race `setenv`, which isn't thread-safe in libc. Both write identical values first thing so the practical risk is nil, but the justification is untrue; a `std::sync::Once` would make it true.
- **`sequential_cut_order_is_not_associative` is a canary, not a regression test** — it asserts *inequality* for two specific permutations, so a future kernel change that legitimately makes cuts order-invariant fails it. You document this in the assert message, which is the right call; just flagging that a future maintainer needs to read that message before "fixing" it.

## Unrelated footgun worth knowing

Running clippy from `rust/` picks up that directory's own `rust-toolchain.toml`, which floats on bare `nightly` → 1.99.0, and reports pre-existing lints in untouched files. From the repo root it resolves the pinned `nightly-2025-11-15` and is green. Anyone debugging a phantom clippy failure will hit this.

Reword the changeset and this is good to go from my side — the fix itself I'd merge as-is.

BIMvoice:
Pushed `e79d66db`. You're right on the changeset — that was an overclaim and I'd rather it not reach the changelog either.

**Reworded to what the fix delivers.** Express ids *are* byte layout, so the sort narrows the dependency rather than removing it. It now says: stable under any reordering that preserves express ids (the perturbation #2019 measured) and under a monotone renumber such as a merge offset, but **not** under an arbitrary id permutation — if a re-export flips two openings' relative order, the sorted sequence changes and the hash moves with it. Also carries your 283-of-5,577 (~5%) measurement, which is a better disclosure than my vaguer version.

Your own point that `sequential_cut_order_is_not_associative` demonstrates the limitation is worth keeping visible — the same physical openings in a different sequence hash differently, which is exactly the id-permutation case. Agreed on the follow-up for an id-independent canonical key; that's the version that survives a cross-tool re-export, and it's a different fix, not a bigger version of this one.

**The SAFETY comment was flatly wrong and is fixed.** You're right that the two `#[test]` fns run concurrently and both raced `setenv`. They wrote identical values so the practical risk was nil, but the justification was false rather than merely optimistic. Now serialised through a `std::sync::Once`, which makes the comment true instead of lucky.

**On the single-producer question** — your second comment answered it more thoroughly than I could have: all four entry points route through the one sort site, the legacy `VoidIndex` has no production callers, and the hash is per-element so cross-element iteration order is irrelevant. Consider it confirmed; I won't make you ask again.

**Not taking `openings.dedup()`**, and I want to give the reason rather than silently skip it. Propagation already dedups at `void_index.rs:66`, your corpus survey found no host with duplicate openings, and adding the line pushes `prepass.rs` to 782 against the 781 budget this PR just set — so it would cost a second ratchet bump for a measured no-op. Happy to add it if you'd rather have the belt-and-braces.

**The acceptance criterion still hasn't run**, and I'm not going to imply otherwise: the #2018 harness commit `d42fbf1c` isn't an ancestor of this branch, so 0/750 can't be measured from here until main merges in. That remains the issue's stated bar.

Thanks for the clippy footgun — `rust/`'s own `rust-toolchain.toml` floating on bare `nightly` versus the pinned root one would have cost me an hour eventually.

Undrafting now, per your note that you'd rather review ready PRs than have them sit.

louistrue:
The reworded changeset is exactly what I was asking for — it now says what the fix buys and where it stops:

> stable under any reordering that preserves express ids … **not** stable under an arbitrary id permutation … Express ids are themselves a property of the byte layout, so this narrows the dependency rather than removing it.

That, plus the 283/5,577 disclosure, means a consumer treating the hash as a content address can tell exactly what changed and what guarantee they now have. Merging.

The geometry-relevant gates all came back green on this head, which is the part that mattered for a mesh-affecting change: **Rust tests**, **Node tests**, **Geometry watertightness census**, and the **Moonshot exams** (provenance, gates, B3.5 golden, determinism). That last one independently confirms no committed golden encoded a pre-fix hash — which was one of the collateral surfaces I checked by hand and is better to have confirmed by CI.

Two things carried forward, neither blocking:

**The id-independent canonical key deserves its own issue.** This closes #2019 as filed, but #2018's real scenario permutes express ids, and your own `sequential_cut_order_is_not_associative` demonstrates why that defeats a sort by id. Opening GlobalId or a geometric key is the version that survives a cross-tool re-export. Worth filing while the reasoning is fresh.

**#2019's 0/750 corpus acceptance run still hasn't happened**, and couldn't from this branch — the harness landed on main after your branch point. Now that this is on main the run is possible; worth doing before closing the issue rather than treating the merge as the acceptance.

Two nits for whenever you next touch the file: `force_sequential_exact_kernel`'s SAFETY comment claims "single-threaded within this test binary", which isn't true — the two `#[test]` fns run concurrently and race `setenv`. Both write identical values first thing so the practical risk is nil, but the justification is wrong; a `std::sync::Once` would make it true. And `sequential_cut_order_is_not_associative` is a canary asserting *inequality*, so a future kernel change that legitimately makes cuts order-invariant will fail it — you document that in the assert message, which is the right call, but it's worth a maintainer knowing before they "fix" it.

Issue #2012, opened 2026-08-03
StepExporter asks the source buffer questions the mutation overlay owns (3 open instances)


louistrue:
Found while fixing #2006. This is not a new bug report so much as the pattern behind several — filed so the next instance is recognised as one rather than patched in isolation.

## The shape

`StepExporter` repeatedly asks `this.dataStore` a question the **mutation overlay** is the authority on. The buffer answers for the file as parsed; the overlay knows what the session has since created, edited, retyped and deleted. Every pass that forgets to ask the overlay produces a file that disagrees with what the user did.

Five instances, three already fixed:

| | instance | status |
|---|---|---|
| 1 | overlay built on `store.properties`, which the columnar parser leaves empty — editing one property dropped every sibling in that pset on save | fixed in #2000 |
| 2 | `newPropertySets` / `newQuantitySets` / type-owned psets built from unfiltered `getMutations()` — a deleted-but-edited entity emitted a dangling `IFCRELDEFINESBYPROPERTIES` | fixed in #1996 |
| 3 | attribute mutations applied only to source-buffer records — every edit after `entity_create` dropped on save | fixed in #2011 |
| 4 | `visibleOnly: true` drops every overlay-created entity | **open** |
| 5 | `isTypeEntity()` reads only `dataStore.entityIndex.byId` | **open** |
| 6 | `export_ifc`'s `global_ids` allowlist queries the parsed model only | **open** |

Plus a seventh that escapes #1996's fix specifically: `entity_delete` on an **overlay-created** entity still emits a dangling relation, because `deleteEntity` *forgets* a new entity rather than tombstoning it — so `isDeleted()` is false and the `isTombstoned` guards pass. Flagged on that PR.

## The three still open

**4 — `visibleOnly: true` silently drops overlay-created entities.** `allowedEntityIds` is a reference closure computed over the source buffer from source spatial roots. Nothing in the source references a new entity, and it is not a root, so it is never in the set — and the new-entities pass filters against that set. Probed: created wall absent from the export, no error, no warning. **This is data loss on save**, same class as #2006, and it is reachable from the viewer's "export visible only".

**5 — `isTypeEntity()` misclassifies overlay-created types.** A pset added to an overlay-created `IfcWallType` is emitted as an occurrence `IFCRELDEFINESBYPROPERTIES` instead of being written into the type's `HasPropertySets`, which stays `$`. Structurally wrong IFC rather than data loss.

[...]

PR #2030, opened 2026-08-03
fix(export): stop deleted and hidden entities emitting dangling relations (#1978)

BIMvoice:
Reopening #1978's fix — **PR #1996 was closed without merging**, two seconds after #2011 landed (`06:04:18` merge, `06:04:20` close). Issue #1978 is still open, and I confirmed the work is not in `main`: `step-exporter.ts` there has a single `isDeleted` call, the original entity-loop guard. I assume the close was collateral from the same-file merge rather than a rejection — if you did mean to drop it, say so and I'll close this again.

Rebased onto current `main` (with #2011 in it). **Zero conflicts** — #2011 works at ~250, the new-entities pass and a new method; this works at ~576 and the three emit loops.

## What it fixes — three routes to the same defect

An `IFCRELDEFINESBYPROPERTIES` whose `RelatedObjects` references an expressId with no defining line. Structurally invalid IFC, not a dropped edit.

1. **Tombstoned source entity** (the issue as filed) — the entity loop skipped its line; the pset/qset loops didn't.
2. **Overlay-created then deleted** (yours) — `deleteEntity` *forgets* a created entity rather than tombstoning it, so `isDeleted()` returns false and a tombstone guard can never catch it.
3. **Hidden under `visibleOnly`** (found self-reviewing this PR) — the visibility filter drops the line, and the emit loops ignored `allowedEntityIds` entirely.

All three go through one predicate answering *"will this id have a defining line at all"*, rather than three separate special cases.

Also fixes the type-owned pset rewrite, a fourth site the issue didn't list: an entity with no `newPropertySets` entry never reaches the first loop's guard.

## Evidence

Removing the three guards fails exactly five tests — one per route, plus the quantity variants — and nothing else: **288 pass / 5 fail**. Restored: **293 pass / 28 skipped**. Typecheck 34/34.

I re-ran that mutation *after* the rebase specifically, because #2011 changing the file underneath could have drained the assertions without touching the tests. It didn't; all five still bite.

## Still not fixed, deliberately

An overlay-created entity under `visibleOnly` is still dropped from the export — that's #2012's instance 4, in your file family. This PR makes the output internally consistent, not more complete.

And on the pattern: this is now the third face found on one PR, each time by adding a condition to a function whose *name* already promised the general answer. That gap is what let each new route through, and I think it supports your read that the real fix is the unified "effective model" view. Happy to write that up as its own issue with all instances enumerated.

One note for CI: the rebase pulled in `@ifc-lite/mcp`'s new dependency on `@ifc-lite/diff`, which needs a `pnpm install` to link — typecheck fails with `TS2307` until it runs. No tracked files change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed STEP exports producing dangling property-set and quantity-set references for deleted or hidden entities.
  * Improved export filtering for entities removed, created then deleted, or excluded by visibility settings.
  * Ensured exported files contain only valid relationships to emitted entities.

* **Tests**
  * Added regression coverage for deleted, removed, and visibility-filtered entities.


louistrue:
You're right on every part of that, and I verified it rather than taking your word:

- **#1996: `state=CLOSED`, `merged=NEVER`**, closed `2026-08-03T06:04:20Z` — two seconds after #2011 merged at `06:04:18`. Closed by @louistrue, so it was collateral from a same-file merge sweep, not a rejection.
- **Issue #1978 is still OPEN.**
- **The fix is not in `main`**: `grep -c isDeleted packages/export/src/step-exporter.ts` → `1`, the original entity-loop guard.

Sorry — that one is on our side, and you lost work to it. Thank you for checking the state of `main` before reopening rather than assuming, and for asking whether the close was deliberate instead of just re-pushing.

Reviewing this now.

Worth connecting to something I raised on #1967 a few hours ago, because the two are the same bug seen from opposite ends. There, `deleteEntity` on an overlay-created entity records it in `forgottenCreatedEntities` but never purges its leftover `newPsets`/`propertyMutations` — and `StepExporter` then emits `IFCPROPERTYSET` + a rel pointing at an id that no longer exists. That is #1978's failure mode reached by a different route.

So there may be one fix rather than two: purging the overlay rows in `deleteEntity` would stop the dangling relation at the source, and the exporter guard here would become defence in depth rather than the only line. Worth deciding which layer owns it before both land — otherwise we get a guard in the exporter and a hole in the overlay, and the next call site that reads the overlay directly hits the same thing again.

Your call on sequencing; I just don't want the two PRs to fix half each.

louistrue:
## These two must land in order: #2030 before #1967

@greptile-apps raised a P1 on #1967 (`effective-changes.ts:357`) that I verified and that changes the sequencing. It is the sibling of the finding I raised earlier — same bug, one case over.

**The case:** an *existing source* entity that is edited and then deleted (tombstoned), rather than an overlay-created one.

**What I confirmed in the code:**

- #1967's filter keeps only the `entity-deleted` row for a tombstoned entity: `if (c.kind !== 'entity-deleted' && snapshot.tombstones.has(c.entityId)) return false;`
- On `main`, `step-exporter.ts` contains **exactly one** `isDeleted` call — line 599, the entity-emission loop. So the entity's own line is skipped…
- …but the property-set path is unguarded: `getForEntity(entityId)` at `:321` still yields the retained pset, and `:978` emits `IFCRELDEFINESBYPROPERTIES(...,(#${entityId}),#${psetId})` pointing at the entity that was just skipped.

So on `main` + #1967 alone, the review hides those rows while the exporter still writes them — **the exact review/export divergence #1967 exists to eliminate**, reintroduced through the tombstone path rather than the forgotten-created one.

**#2030 is the fix**, and it is the right layer: it adds a shared deleted-or-hidden probe and gates **three** emit sites on `willBeEmitted(entityId)`, covering hidden entities (`allowedEntityIds`) as well as deleted ones — which the entity loop's single `isDeleted` never did.

**So:** merge **#2030 first**, then #1967. In that order #1967's filter is correct — it hides rows for an operation the exporter genuinely no longer performs. In the reverse order, or with #1967 alone, it hides an operation that still happens.

This is the "which layer owns it" question from my earlier comment on #2030, and the answer turned out to be *both, in sequence*: the overlay purge in #1967 stops the forgotten-created case at the source, and #2030's emit guards stop the tombstoned and hidden cases at the boundary. Neither subsumes the other.

@BIMvoice — nothing to change on either PR as far as I can see; it is purely an ordering constraint. Flagging it because merging #1967 first would look fine and quietly reopen the bug for a different input.

BIMvoice:
On sequencing — I checked rather than reasoned, and the answer is that these aren't two halves. They're complementary, and only one of the three routes overlaps.

First, a correction: the `deleteEntity` purge isn't merged yet. #1967 is still open (`stashAndPurgeEntityOverlay` returns 0 hits on `main`), so nothing has landed that this could be defence-in-depth *for* yet.

More importantly, the purge structurally cannot cover two of the three routes. Tested against the branch that has it:

```
tombstoned SOURCE entity, pset edited, then deleted:
  getForEntity(42) -> [{ name: "Pset_X", properties: [...] }]     <- rows still there
  getMutations()   -> 2
```

That's **correct** behaviour, not a gap. A source entity is *tombstoned*, not forgotten — its edits legitimately stay in the overlay because undo needs them and the entity genuinely exists in the file. So the purge never touches route 1, and the exporter guard is the only thing preventing the dangling relation there.

Route 3 is further still: a `visibleOnly`-hidden entity was never deleted at all. No deletion path runs, nothing is forgotten or tombstoned, and the purge has no involvement whatsoever.

So the split is:

| route | purge (#1967) | exporter guard (#2030) |
|---|---|---|
| tombstoned source entity | no — rows correctly retained | **yes** |
| overlay-created then deleted | **yes** — at source | yes (redundant) |
| hidden under `visibleOnly` | no — no deletion involved | **yes** |

Only the middle row is doubled up, and there I'd agree the purge is the better fix and the guard is belt-and-braces. The other two rows are the guard alone.

**Suggested order:** #1967 first (it fixes the overlay at source), then this. If you'd rather, I'm happy to drop this PR's created-then-deleted *test* once #1967 lands so the coverage isn't duplicated — but the predicate itself should stay, because removing it reopens routes 1 and 3.

Your call on whether that's the shape you want; I'd rather ask than assume, given you explicitly don't want the two fixing half each.

Also taking the CodeRabbit point that the changeset undersells this — it still describes only the `isDeleted` case, not `willBeEmitted`, overlay-created entities, or `visibleOnly`. I'll rewrite it to match what actually shipped.

PR #2039, opened 2026-08-03
fix(export): resolve attribute slots across schemas, not the IFC4 pin (#2032)

### LINKED ISSUE #2032: demesh-writer skips IFC4X3 elements: findAttrIndex uses the pinned attribute table
Found while sweeping for #2003 in #2031. Same bug family, different function, different package, so it is filed rather than folded in.

## The bug

`packages/export/src/demesh-writer.ts:284` resolves attribute indices with `getAllAttributesForEntity`, which is generated from the **IFC4_ADD2_TC1 pin only**. For a class outside that pin it returns an empty list, so `findAttrIndex(ref.type, 'Representation')` answers `null` and the element is skipped with `no-representation-attribute`.

Measured:

```
IfcSignal    pinnedAttrs=0  RepresentationIdx(pinned)=-1  RepresentationIdx(union)=6
IfcPavement  pinnedAttrs=0  RepresentationIdx(pinned)=-1  RepresentationIdx(union)=6
IfcCourse    pinnedAttrs=0  RepresentationIdx(pinned)=-1  RepresentationIdx(union)=6
IfcWall      pinnedAttrs=9  RepresentationIdx(pinned)=6   RepresentationIdx(union)=6
```

So a demeshed export of an IFC4X3 infrastructure model silently drops its signals, pavements and courses. The skip reason reads as a property of the model rather than of the exporter, which is what makes it quiet.

## The fix

`getAttributeNamesAcrossSchemas`, the same cross-schema walker #2001 exported and #2014 and #2031 adopted elsewhere.

## Worth doing properly rather than as a one-liner

- The failure is a **silent skip**, so the test has to assert an IFC4X3 element survives the round trip, not merely that the index resolves. Re-parse the written file and look for the element.
- Check whether `no-representation-attribute` is reported anywhere a user sees it. A skip counted as a modelling problem when it is a lookup problem will send someone debugging the wrong file.
- Sweep `packages/export` for other pinned-table lookups while there. `attribute-slot-types.ts` already has the pin-then-union path, so it is the exception rather than the rule.

## Context

This is the sixth site in the same family (#2001 `diff --by-entity`, #2003 `validate`, #2014 two MCP tools, #2031 `isKnownType` and `normalizeIfcTypeName`). The naming recommendation recorded in #2031 applies: the pinned function has the shorter name, so it keeps getting picked by default.

BIMvoice:
Closes #2032, and fixes a second instance of the same bug that the issue didn't name.

## The reported bug

`demesh-writer.ts`'s `findAttrIndex` resolved positional attribute slots through the parser's IFC4-pinned registry. For a class that exists only in IFC4X3 — `IfcSignal`, `IfcPavement`, `IfcCourse`, the infrastructure additions — that registry returns nothing, so `Representation` resolved to `null` and every element of that type was skipped with reason `no-representation-attribute`.

## The one it didn't name, and it's worse

Grepping for the shape turned up `lod0-generator.ts` carrying a **byte-identical** copy of the helper. There the null index is `ObjectPlacement`, and the walk treats that as `continue` — so the element is dropped from the LOD0 export **entirely, with no skip reason recorded anywhere**.

The de-mesh case at least leaves a trace in `report.skipped`. The LOD0 case leaves nothing: an infrastructure model can lose elements from its LOD0 output with no indication in the result that anything was omitted. Worth fixing in the same PR since it's literally the same function.

Both now use `getAttributeNamesAcrossSchemas` — the cross-schema union already used by `attribute-real-slots.ts`, `attribute-slot-types.ts` and #2033's Tag-hashing fix, so this follows an established convention rather than inventing one.

## Evidence

Each fix is pinned **separately**, so neither test stands in for the other:

| mutation | result |
|---|---|
| de-mesh writer back to the pinned lookup | its IFC4X3 test fails, and only it |
| LOD0 back to the pinned lookup | its IFC4X3 test fails, and only it |

**333 passed / 29 skipped**, up from 331 baseline — the two new tests.

## Found but deliberately not fixed

Same pinned lookup, lower severity, reported rather than swept in:

- `unit-normalize.ts:147` and `select-qualification.ts:77` — both degrade gracefully (empty plan, or left to the default serializer) rather than dropping content
- `mcp/tools/discovery.ts:242` — affects an info tool's attribute listing, not exported data

Happy to take those as a follow-up if you'd rather they were consistent, but they aren't data-integrity bugs and I'd rather not widen a fix PR on my own judgement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed IFC4X3 infrastructure elements being omitted from de-meshed and LOD0 exports.
  * IFC4X3 signal elements now retain their expected identifiers, class information, and tessellated geometry during export.

* **Tests**
  * Added regression coverage for IFC4X3 signal exports in both de-meshed and LOD0 workflows.

### DIFF
--- a/.changeset/demesh-writer-ifc4x3-attr-index.md
+++ b/.changeset/demesh-writer-ifc4x3-attr-index.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/export": patch
+---
+
+Stop silently dropping IFC4X3-only element types from de-meshed and LOD0 exports (#2032).
+
+Both `demesh-writer.ts` and `lod0-generator.ts` carried a private `findAttrIndex` that resolved positional attribute slots through the parser's IFC4-pinned registry. For a class that exists only in IFC4X3 — `IfcSignal`, `IfcPavement`, `IfcCourse` and the rest of the infrastructure additions — that registry returns nothing, so every attribute index came back null.
+
+In the de-mesh writer that meant `Representation` could not be located and the element was skipped with reason `no-representation-attribute`. In the LOD0 generator it meant `ObjectPlacement` could not be located and the element was dropped from the walk entirely, with no skip reason recorded anywhere — so an infrastructure model could lose elements from its LOD0 export with nothing in the output to say so.
+
+Both now resolve slots through the cross-schema union already used by `attribute-real-slots.ts` and `attribute-slot-types.ts`.
--- a/packages/export/src/demesh-writer.test.ts
+++ b/packages/export/src/demesh-writer.test.ts
@@ -196,6 +196,21 @@ describe('applySimplifiedGeometry', () => {
     // element #10's Representation was replaced, not skipped.
     expect(out).toMatch(/#10=IFC\w+\([^\n]*#\d+/);
     expect(danglingRefs(out)).toEqual([]);
+
+    // Write-and-reparse: `danglingRefs` only proves the text is internally
+    // consistent, not that a parser can read it back. Reparsing is what shows
+    // the element actually survives a round trip rather than merely appearing
+    // in the output. The class is not asserted here because the IFC4 target
+    // downcasts the IFC4X3-only IfcSignal on write (see above); what has to
+    // survive is #10 and its replaced representation.
+    const reparsed = await new IfcParser().parseColumnar(
+      new TextEncoder().encode(out).buffer,
+      { disableWorkerScan: true },
+    );
+    expect(reparsed.entityIndex.byId.has(10)).toBe(true);
+    const reparsedText = new TextDecoder().decode(reparsed.source);
+    expect(reparsedText).toMatch(/#10=IFC\w+\([^\n]*#\d+/);
+    expect(reparsedText).toMatch(/IFCTRIANGULATEDFACESET/);
   });
 
   it('keeps openings when stripOpenings is false', async () => {
--- a/packages/export/src/demesh-writer.test.ts
+++ b/packages/export/src/demesh-writer.test.ts
@@ -19,6 +19,34 @@ function danglingRefs(text: string): number[] {
   return [...refs].filter((id) => !defined.has(id)).sort((a, b) => a - b);
 }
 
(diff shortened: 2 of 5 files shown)


louistrue:
Verified this one thoroughly, because the risky part is not the fix but what it could quietly change underneath every existing export.

`findAttrIndex` resolves **positional** slots. Swapping the source of that list means any ordering or length difference between the pinned registry and the cross-schema union would shift every attribute read for classes that already worked. I checked that directly rather than reasoning about it, over the whole registry:

```
classes total: 776
pinned-known checked: 755   ->  INDEX DRIFT: 0
pin-empty (broken today):  21
```

Every attribute of every class the IFC4 pin already knows keeps the exact same index under `getAttributeNamesAcrossSchemas`, so this is purely additive. And 21 classes currently resolve to an empty list, which is the size of the bug you found.

Both fixes are pinned separately, as your table claims:

| mutation | fails |
|---|---|
| `demesh-writer` back to the pin | `replaces the representation on an IFC4X3-only element type` |
| `lod0-generator` back to the pin | `includes an IFC4X3-only element rather than silently dropping it` |

Finding the second copy was the valuable part. The silent-drop variant is strictly worse than the reported one precisely because it leaves nothing in `report.skipped`, and nobody would have gone looking for it from #2032 alone.

**I pushed `98b0dbc0`** for the open CodeRabbit thread, which was fair: `danglingRefs` shows the text is internally consistent, not that a parser can read it back, so the test asserted appearance rather than survival. It now reparses the output and checks `#10` is present with its replaced representation. I deliberately did not assert the class, since the IFC4 target downcasts `IfcSignal` on write and that is unrelated to this fix.

I also checked the new assertion actually discriminates rather than passing vacuously (`has(10)` true, `has(999999)` false, 12 entities reparsed) — worth doing, since a reparse assertion that silently accepts anything would be worse than no assertion at all.

`@ifc-lite/export` 333 passed. `patch` is the right bump; no export surface moves.

Issue #2044, opened 2026-08-03
importMutations replays a created entity's property mutations without the entity


BIMvoice:
### Summary

`@ifc-lite/mutations` exposes `MutablePropertyView.exportMutations()` /
`importMutations()` as a round-trip pair for a model's edit history (this is
public API — the package is published). A `CREATE_ENTITY` mutation
(`view.createEntity(...)`) is **not** reconstructed on import: `applyMutations()`
explicitly skips it, and property/attribute/quantity mutations that were
recorded against that new entity are replayed anyway — as an orphaned overlay
entry keyed to an expressId that was never created on the receiving side.

The skip is intentional in the code (there's a `console.warn` and a comment
explaining `restoreNewEntity()` is the intended companion path), but nothing
outside the file documents this — not the package README, not the doc
comments on `exportMutations`/`importMutations` themselves, and the package's
only round-trip test (`test/retype.test.ts`, "replays through
exportMutations → importMutations") only exercises a retype on a
pre-existing entity id, never a created one. A caller reading the public
surface (README, JSDoc, published `.d.ts`) has no way to learn that
`importMutations()` is not a full inverse of `exportMutations()`.

### Reproduction

`packages/mutations/src/mutable-property-view.ts`, using only the public API:

```ts
const a = new MutablePropertyView(null, 'model-a');
a.setExpressIdWatermark(1000);
const e = a.createEntity('IfcWall', ['$', '$', '$']);
a.setProperty(e.expressId, 'Pset_WallCommon', 'FireRating', 'F90', PropertyValueType.String);

const json = a.exportMutations();

const b = new MutablePropertyView(null, 'model-b');
b.setExpressIdWatermark(1000);
b.importMutations(json);
```

Printed output:

```
applyMutations: CREATE_ENTITY for #1001 requires a NewEntity payload — restore via restoreNewEntity()
getNewEntities(): []
isDeleted(1001): false
getForEntity(1001): [{"name":"Pset_WallCommon","globalId":"new_mut_...","properties":[{"name":"FireRating","type":0,"value":"F90"}]}]
```

`b` has no record that entity `#1001` was ever created — but it *does* have a
property set floating under that expressId (`getForEntity(1001)` is
non-empty), because `CREATE_PROPERTY` replay has no dependency on the entity
actually existing in `newEntities`. That is worse than a clean loss: it's an
orphaned property set attached to an id that exists in neither the source
buffer nor `newEntities`.

Verified byte-for-byte identical on both:
[...]

PR #2045, opened 2026-08-03
fix(mutations): stop importMutations replaying a skipped entity's edits (#2044)

BIMvoice:
Closes the **corruption half** of #2044. Deliberately leaves the contract question open — see below.

## What was wrong

`applyMutations` skips `CREATE_ENTITY` on purpose (warning + `skippedCreateIds`), but still replayed the property/attribute/quantity mutations recorded against that id. Through public API:

```
entity present : false
psets for id   : [{"name":"Pset_X", …}]
effective rows : [{"entityId":1,"kind":"pset-added","setName":"Pset_X"}]
```

The receiving view reports a pending change for an entity that does not exist. That is the dangling-relation shape from #1978 / #2036 — a pset surviving without its entity — but reached through **import** rather than deletion, so the emit-side guards can't help: by the time they run, the overlay already holds a normal-looking pset whose entity is merely absent.

## The fix, and the line I did not cross

The replay now skips any non-`CREATE_ENTITY` mutation whose target is in `skippedCreateIds`. The round trip becomes **lossy** — entity and edits both dropped — instead of **corrupting**.

**Not** reconstructing creates on import. That would make `importMutations` a true inverse of `exportMutations`, and #2044 poses it as an open question rather than a defect. It's a contract decision and yours to make; this closes the actual harm without foreclosing it.

## The regression risk, and how it's pinned

The guard keys strictly off `skippedCreateIds` — populated only when a `CREATE_ENTITY` for that id was actually skipped in the same batch — **not** off "absent from `newEntities`". The latter reads as equivalent and would drop every ordinary edit to a source-buffer entity, i.e. break all normal imports. There's a test for exactly that, and it stays green under the mutation below, so the two aren't coupled.

Removing the guard fails exactly the new orphan test and nothing else. Mutations **76**, export **333**, typecheck 34/34.

The `console.warn` now says dependent mutations were dropped too, so the runtime signal matches what actually happens.

## Same-shape check

- `ChangeSetManager` stores and serialises `Mutation[]` but never replays — it inherits this fix through any caller that feeds its output back into `importMutations`.
- The viewer's undo/redo reads and writes the live view's own history, never crossing a serialisation boundary.
- `step-exporter.ts` groups `mutation.entityId` from the live view's history, self-consistent by construction.

`applyMutations` is the only replay path in the repo with this shape.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented changes from being applied to entities whose creation was skipped.
  * Ensured dependent changes are skipped consistently regardless of mutation order.
  * Preserved replay of changes targeting existing or explicitly restored entities.
  * Improved warnings to clarify when dependent changes are skipped and entity data must be restored separately.

* **Tests**
  * Added coverage for skipped entity creation, mutation ordering, and existing-entity replay scenarios.


louistrue:
Verified `824c88ef`. Both directions hold, and the over-suppression risk you called out is genuinely pinned rather than just asserted.

Reproduced the corruption and the fix through the public API:

```
orphan case (created entity whose CREATE_ENTITY is skipped):
  psets for id  : []          <- was [{"name":"Pset_X", …}]
  hasPendingChanges: false    <- was a pending change for a non-existent entity

normal case (edit on a pre-existing source-buffer entity):
  value after replay: "edited"   <- unaffected
```

Two mutations, and each fails a different test, which is what makes the `skippedCreateIds` keying load-bearing rather than incidental:

| mutation | fails |
|---|---|
| guard removed | `does not orphan a property set under a created-but-unrestored entity id` |
| guard keyed off `!newEntities.has(id)` instead of `skippedCreateIds` | `still replays mutations that target a pre-existing source-buffer entity` and `replays through exportMutations → importMutations` |

That second one is the useful one. Your comment argues the guard must key off `skippedCreateIds` rather than "absent from `newEntities`", and the over-broad version does exactly what you predicted — it starts swallowing edits to ordinary source entities. A reader who "simplified" it later would be caught.

Agreed on the line you did not cross. Reconstructing creates on import would make `importMutations` a true inverse of `exportMutations`, and that is a contract decision for Louis, not something to settle inside a corruption fix. Lossy-not-corrupting is the right intermediate state, and #2044 stays open for the rest.

One note for your own local runs, not a finding: `test/mutations.test.ts` fails to load in a fresh worktree on an unbuilt `@ifc-lite/encoding`, which reads as a real failure. `pnpm --filter @ifc-lite/encoding build` first and the suite is clean.

BIMvoice:
Pushed `83465102`, hardening a hole the same self-review found here.

`skippedCreateIds` was built in a single forward pass, so a dependent mutation appearing **before** its `CREATE_ENTITY` saw an empty set and replayed — reproducing the orphaned pset this PR exists to close. Demonstrated by reversing a normal `exportMutations()` array: `getNewEntity` correctly returns null, but `getForEntity` still hands back the pset and `hasChanges` reports true.

Not reachable through anything wired up today — `exportMutations` is append-ordered by construction and nothing else calls `applyMutations`. But the method's own doc advertises "e.g., from imported change set", i.e. an arbitrary `Mutation[]`, and it goes live the moment anything merges or reorders two histories. Cheap to close now, awkward to diagnose later.

Now two-pass: collect every `CREATE_ENTITY` id first, then apply and filter against the complete set. Order-independent, one extra walk of an array already in memory.

Three mutations, all killed — reverting to single-pass, plus two pass-boundary off-by-ones. **77 passed / 10 skipped**, typecheck 34/34. The pre-existing-source-entity regression guard stays green under all three, so it isn't coupled to the new behaviour.

louistrue:
Verified `83465102`. The two-pass split is right and the ordering property is pinned in isolation, which took a slightly careful mutation to establish.

The obvious mutation — deleting pass 1 outright — fails two tests, but that is not evidence about ordering: removing pass 1 also removes the only place `skippedCreateIds` is populated, so it breaks the original orphan guard too. It proves the set matters, not that collecting it up front matters.

The mutation that isolates the actual claim is restoring the *pre-fix* shape: drop pass 1, put `skippedCreateIds.add(...)` back inside the `CREATE_ENTITY` case so population is incremental again. That fails exactly one test:

```
× skips a dependent mutation whose CREATE_ENTITY appears LATER in the array (#2044 follow-up)
```

and `does not orphan a property set under a created-but-unrestored entity id` keeps passing. So the two behaviours are separately pinned rather than one test covering both, and the order-independence is genuinely load-bearing instead of incidental.

That distinction matters here because the failure mode is invisible in normal use: `exportMutations()` emits history order, where the create always precedes its dependents, so the incremental version looks correct forever. It only breaks on an arbitrary `Mutation[]` — merged, reordered, or hand-assembled — which is exactly the case your comment calls out and exactly the case nobody exercises by accident.

`@ifc-lite/mutations` 77 passing on this head.

louistrue:
Holding this one — the Greptile P1 is real and I reproduced it. The guard over-suppresses in the flow your own warning message tells callers to use.

Restore the payload first, then replay history (which is what `restore via restoreNewEntity()` instructs):

```ts
dst.restoreNewEntity(e);                 // caller supplies the payload FIRST
dst.importMutations(exported);           // then replays history
```

Result on `83465102`:

```
ENTITY PRESENT: true      <- the entity is there
PSETS         : []        <- but its property sets are gone
PROP VALUE    : null      <- the edit was dropped
```

So the entity survives and every edit against it is silently discarded. `applyMutations` still adds the id to `skippedCreateIds` when it walks the `CREATE_ENTITY` record, even though `restoreNewEntity` has already put a live entity in `newEntities` — and the new guard then drops every dependent mutation for an id that is no longer missing.

Greptile's suggested shape looks right to me:

```ts
if (
  mutation.type !== 'CREATE_ENTITY' &&
  skippedCreateIds.has(mutation.entityId) &&
  !this.newEntities.has(mutation.entityId)
) {
  continue;
}
```

`skippedCreateIds` then means "the create was skipped **and** nothing else supplied the entity", which is the condition the orphan guard actually wants. The pre-pass that makes the skip set order-independent (`83465102`) is unaffected and still correct — I re-verified that separately.

Worth adding a test for this ordering specifically, since it is the documented recovery path and currently the one that loses data. The existing tests all import without a prior restore, so none of them can see it.

Everything else on this PR still holds: the orphan fix, the `skippedCreateIds` keying (the over-broad `!newEntities.has()` variant fails two tests), and the order-independence pre-pass.

BIMvoice:
Fixed in db112b86. You were right and so was Greptile — I verified the over-suppression myself before touching anything.

Guard is now the shape you endorsed:

```ts
if (
  mutation.type !== 'CREATE_ENTITY' &&
  skippedCreateIds.has(mutation.entityId) &&
  !this.newEntities.has(mutation.entityId)
) {
  continue;
}
```

The pre-pass is untouched.

**The test you asked for.** `keeps dependent mutations when restoreNewEntity() ran BEFORE the import` — it fails on 83465102 exactly as you described:

```
AssertionError: expected [] to match object [ { name: 'Pset_WallCommon', … } ]
- [ { "name": "Pset_WallCommon", "properties": [ { "name": "FireRating", "value": "F90" } ] } ]
+ []
```

You were right that no existing test could see it: all of them import without a prior restore. It asserts the entity, the pset, and the attribute, so a partial replay can't pass it either.

**Mutations, each with the replacement count printed** — my first attempt at this one was a silent no-op (my regex assumed the `&&` led the line when it trails the previous one), which is why I now print the count and re-read the guard before trusting a result:

| mutation | fails |
|---|---|
| drop `!this.newEntities.has(...)` | 1 failed / 77 passed — only the new test |
| drop the `skippedCreateIds` clause instead (the over-broad variant) | 2 failed / 76 — `replays through exportMutations → importMutations` and `still replays mutations that target a pre-existing source-buffer entity` |

So the two clauses are separately pinned rather than one covering both, and the coupling you flagged stays intact. Your three named tests all pass with the new clause in place.

78 passing / 10 skipped (was 77, exactly +1), typecheck 34/34.

**Two CodeRabbit points, both checked against the code and both valid.** `getForEntity` (`mutable-property-view.ts:218`) returns `PropertySet[]` and has no pending-change semantics — `hasChanges()` (`:1265`) and `hasPendingChanges()` (`:1287`) are what report it, and it's the latter the orphaned pset inflated via `newPsets.size > 0`. Changeset reworded. I also rewrote its second paragraph, which described the guard as keyed off the skip set *rather than* `newEntities` — after this fix it's both, and leaving that would have been an inaccurate changeset. `getAttributeMutationsForEntity` does exist (`:1171`), and the gap was real: neutralising the `UPDATE_ATTRIBUTE` case now fails two assertions, where before it failed none.

I also softened the `console.warn` and the comment above it — both asserted unconditionally that every dependent mutation is dropped, which stopped being true once the payload was restored first. Nothing asserts on that string.

louistrue:
Verified `db112b86`. The data loss is closed and the original orphan guard survives — three behaviours, three distinct tests, no overlap between them.

Both orderings through the public API:

```
A: restoreNewEntity() BEFORE importMutations()  -> value "v1"   (was null — the regression)
B: importMutations() with no restore            -> value null, psets []  (orphan still suppressed)
```

Mutations, each failing a different test:

| mutation | fails |
|---|---|
| drop `&& !this.newEntities.has(...)` | `keeps dependent mutations when restoreNewEntity() ran BEFORE the import` |
| remove the whole guard | `does not orphan a property set under a created-but-unrestored entity id` **and** `skips a dependent mutation whose CREATE_ENTITY appears LATER in the array` |

So the escape clause, the orphan guard and the order-independent pre-pass are each pinned separately rather than one test standing in for the group. That is what I wanted here, because the three interact: the escape clause has to not re-open the orphan hole, and the pre-pass has to keep working regardless of which of the two applies.

Reading `skippedCreateIds && !newEntities.has(...)` as "the create was skipped **and** nothing else supplied the entity" is the right meaning for the set. The earlier version conflated "we skipped the create record" with "the entity is absent", and those diverge exactly when the caller does what the warning tells them to.

`@ifc-lite/mutations` 78 passing.

Issue #2046, opened 2026-08-03
IFC5/IFCX and Parquet export do not honour overlay-deleted entities


BIMvoice:
## Summary

`Ifc5Exporter` (`packages/export/src/ifc5-exporter.ts`) and `ParquetExporter`
(`packages/export/src/parquet-exporter.ts`) iterate the raw parsed
`IfcDataStore.entities` table directly and never consult the overlay's
deletion state. An entity deleted via `MutablePropertyView.deleteEntity()`
is exported anyway. `StepExporter` does not have this problem — it resolves
deletions through `getEffectiveEntityIndex(...).isDeleted()` (added in #2036).

This reproduces the concern raised on the (closed, unmerged) PR #1996 against
current `main` (verified at commit `c65bdbe0`, "fix(sdk,parser): reject
abstract IFC classes at the authoring boundary (#2035) (#2041)"). It was not
fixed incidentally by #2036 — that work only reached `step-exporter.ts`.

## Code-path finding

`packages/export/src/ifc5-exporter.ts` imports (lines 18-27) do not include
`effective-index.js` or anything named `isDeleted`/`getEffectiveEntityIndex`.
A repo-wide grep confirms zero occurrences of `isDeleted` or `effective-index`
anywhere in the file. `Ifc5Exporter` walks `this.dataStore.entities` directly
in several places — e.g. the main node-collection loop (line 209), the
UUID/child-name builders (lines 376, 425, 442), and `buildVisibleSet()`
(lines 682-703, which only applies `hiddenEntityIds`/`isolatedEntityIds` —
UI-level show/hide, not overlay deletion). None of these consult the
overlay's tombstone state.

By contrast, `packages/export/src/step-exporter.ts` imports
`getEffectiveEntityIndex` (line 31), builds an `EffectiveEntityIndex` per
export (line 275), and gates entity emission on `effective.isDeleted(entityId)`
at lines 362, 387, and 461.

`packages/export/src/parquet-exporter.ts` is worse: its constructor
(lines 24-27) doesn't even accept a `MutablePropertyView`, so it is blind to
*all* overlay state, not just deletions — pset/quantity/attribute edits are
also silently dropped, in addition to deleted entities being included.

## Demonstration

Built two entities through the public API (`IFCWALL` #1 "Wall1", `IFCWALL`
#2 "Wall2 (deleted)"), deleted #2 via `MutablePropertyView.deleteEntity(2)`,
[...]

PR #2047, opened 2026-08-03
fix(export): honour overlay deletions in the IFC5 exporter (#2046)

BIMvoice:
Closes the **IFC5 half** of #2046. Parquet stays flagged there — see the bottom.

## The asymmetry

`Ifc5Exporter` walked `dataStore.entities` directly and never consulted the overlay. The STEP exporter has resolved deletions through `getEffectiveEntityIndex().isDeleted()` since your #2036; that seam simply never reached this file. Same `(dataStore, view)` pair, two answers:

```
STEP export contains deleted Wall2?  false
IFC5 export contains deleted Wall2?  true
```

Both now answer `false`. It uses the same seam rather than a second deletion check — a differently-shaped guard here is precisely how the two drift apart again.

## Four loops, not one

The obvious fix gates the node-collection loop and looks done. It isn't: the UUID assignment, the name index, and the two `childrenOf` grouping passes each walk the entity table independently.

- Gating only the node loop removes a deleted entity's own node **while leaving it referenced as a child** of a surviving container — `spatialHierarchy` is built once from the parsed source and still lists it. That trades one inconsistency for another.
- The grouping passes matter for a subtler reason: a tombstoned entity left in a sibling group perturbs the **collision-suffix naming** of the siblings that survive. So a deletion could silently rename an unrelated entity.

Both cases are covered by the two new tests.

`buildVisibleSet()` is untouched — `hiddenEntityIds`/`isolatedEntityIds` are UI show/hide, deliberately a separate mechanism from overlay deletion, and there's a regression test pinning that they behave as before.

## Evidence

Neutralising all five gate sites fails exactly the two new tests and nothing else (335 pass / 2 fail); restored, **337 passed / 29 skipped**. Typecheck 34/34.

## Deliberately not fixed

`parquet-exporter.ts` takes **no `MutablePropertyView` parameter at all**, so it's blind to every overlay edit rather than just deletions. That needs an API-shape decision rather than a guard, so it stays in #2046 rather than being folded in here on my own judgement.

## Summary by CodeRabbit

* **Bug Fixes**
  * IFC5 exports now consistently exclude deleted, hidden, or filtered entities.
  * Deleted entities no longer appear as standalone nodes or spatial child references.
  * Surviving children are correctly re-parented when ancestors are removed.
  * Exported names and sibling grouping remain accurate after deletions.
  * Cyclic relationships are handled safely without dangling or self-referential links.
  * Visibility filtering now preserves consistent, reachable export hierarchies.

* **Tests**
  * Added coverage for deletion, re-parenting, visibility filtering, and preserving existing behavior when overlays are disabled.


louistrue:
Verified `a7140ba2`. The fix is right and the "four loops, not one" reasoning holds — I confirmed the subtle one empirically. But **two of the five gates are not pinned by any test**, and the effect they guard is user-visible, so a later "simplification" would ship silently.

### The gates, mutated one at a time

| line | loop | removing it |
|---|---|---|
| 227 | node collection | **fails** a test |
| 401 | UUID assignment | **fails** a test |
| 451 | entity name index | passes |
| 468 | `childrenOf` grouping (by `childId`) | passes |
| 474 | root-level grouping pass | passes |

### The unpinned effect is real, and I reproduced it

Two same-named `IfcWall` siblings under one storey, second one deleted, reading the surviving sibling's child key:

```
gates 451/468/474 present:  ["Wall"]
gates 451/468/474 removed:  ["Wall_1"]
```

So exactly what your comment predicts: the tombstoned sibling stays in the name index and the sibling group, and the **survivor** picks up a collision suffix it should never have had. That changes the node's key and path in the emitted IFC5 file — a stable-identifier change for an entity nobody touched, from a deletion elsewhere in the storey.

Your reasoning was right and I would not drop those gates. The gap is that nothing fails if someone does. The existing deletion tests use distinct names (`Wall1` / `Wall2 (deleted)`), so they cannot see collision-suffix behaviour at all — the two cases are the same test until the names collide.

A test in the shape of the probe above would close it: same-named siblings, delete one, assert the survivor's child key stays `Wall` rather than becoming `Wall_1`. That is the assertion the current suite is missing, and it is the one that distinguishes gating four loops from gating one.

Everything else checks out: `@ifc-lite/export` 334 passing, the `getEffectiveEntityIndex` seam reused rather than a second deletion predicate, and the visibility-filtering regression test keeps the two mechanisms apart.

(`src/lod1-generator.test.ts` fails to load here on an unbuilt `@ifc-lite/wasm` — worktree staging, not your branch.)

BIMvoice:
Pushed `e8dbc7c7` — a self-review found a reachable bug in my own fix before you got to it.

**Gating the child but not the parent leaves an orphan.** Delete a spatial container, keep something inside it: the child survives the node loop and lands in `file.data[]`, but it was registered under the deleted parent's key — never read, since that parent is skipped — and it can't reach the root bucket either, because that only takes entities with *no* parent entry. Present in the file, unreachable from the document root. Exactly the class this PR set out to fix, arriving through parent-deletion rather than self-deletion.

Surviving children are now re-parented to their **nearest surviving ancestor**, falling back to root when every ancestor is deleted. That matches the intuition of deleting a container and keeping its contents, and keeps the tree connected rather than silently dropping a subtree.

The ancestor walk is cycle-guarded, and the ordering there is load-bearing: `visited` is seeded with the starting `childId` and checked **before** the deleted-check. Reversed, a cycle looping back to the start returns the child as its own parent, because the start node is never itself deleted. The first draft had it that way round; it was caught by tracing rather than by a test, and the mutation suite now covers it — removing the guard produces a real hang.

Seven mutations on this fix, all killed: ancestor-skip removed, cycle guard removed, reparenting bypassed, wrong deleted-check target, and the check-order bug reproduced twice. Also added the `applyMutations: false` coverage that was missing — the behaviour was already right, just unguarded.

**341 passed / 29 skipped**, typecheck 34/34.

Regression guards held: a surviving child of a surviving parent nests exactly as before, and `hiddenEntityIds`/`isolatedEntityIds` are untouched.

Parquet is still deliberately out of scope and flagged in #2046 — no `MutablePropertyView` parameter at all, so it needs an API-shape decision rather than a guard.

louistrue:
Verified `e8dbc7c7`. The orphaned-child bug is a good catch and the cycle tests are the right paranoia. But re-running the gate matrix on this head, **coverage went down, not up** — and one gate that was pinned before no longer is.

| line | loop | removing it |
|---|---|---|
| 233 | node collection | **fails 3** |
| 407 | UUID assignment | passes ← **was pinned at `a7140ba2`** |
| 453 | entity name index | passes |
| 484 | nearest-surviving-ancestor walk | **fails 3** |
| 500 | `childrenOf` grouping | passes |
| 507 | root-level grouping | passes |

Two of six pinned. At `a7140ba2` it was two of five, but the UUID gate (then line 401) **was** one of them. The re-parenting change made it redundant for the existing tests: children now reach the root through the ancestor walk, so dropping the UUID gate no longer breaks the paths those tests assert. The gate is still doing real work — it is what keeps a deleted id out of `entityUuids`, and `addChild`'s silent skip depends on that — it just no longer has a test that notices.

That is the more interesting failure mode than a plainly missing test: a fix made the suite *less* sensitive without anyone changing a test, and nothing announces it.

The collision-suffix gap from my last comment is also still open — I re-confirmed it on this head:

```
gates 453/500/507 present:  survivor child key ["Wall"]
gates 453/500/507 removed:  survivor child key ["Wall_1"]
```

Two same-named siblings, one deleted, and the survivor's stable path changes. Still nothing fails if those three are removed.

So four of six gates are load-bearing but untested, and I would not want to leave `e8dbc7c7` in that state given the file now carries two distinct deletion behaviours (skip and re-parent) that interact. Two assertions would cover it: the collision-suffix one I described before, and one that a deleted entity gets no `entityUuids` entry (or equivalently that a surviving parent's child list never contains a deleted id's uuid, asserted at the uuid level rather than through the reachable path).

`@ifc-lite/export` 338 passing on this head, and the `applyMutations: false` test correctly pins that none of this fires when mutations are off.

BIMvoice:
Pushed `18887104`. Gate 407 is pinned again, and I have an evidenced answer on the other three rather than a promise.

Your framing was the useful part: *a fix made the suite less sensitive without anyone editing a test.* That is a worse failure than a missing test, because nothing announces it — and it is the same drain-without-edit shape as the vacuous Rust tests on #1970, caused by our own change this time rather than someone else's merge.

**Matrix on this head:**

| line | loop | removing it |
|---|---|---|
| 233 | node collection | fails 3 |
| 407 | UUID assignment | **fails 1** ← newly pinned |
| 453 | entity name index | passes |
| 484 | ancestor walk | fails 4 (was 3) |
| 500 | `childrenOf` grouping | passes |
| 507 | root-level grouping | passes |

The uuid test asserts against the table directly rather than through reachability, per your point — reachability is precisely what the ancestor walk now satisfies independently, which is how the gate went dark. It reads a private field through a precise structural cast, not `as any`.

**On 453 / 500 / 507 — they cannot be pinned individually, and that is the design rather than a gap.** Verified by mutation, not assumed:

Gate 407 is a complete backstop for a deleted entity's own emission — `addChild`'s `entityUuids.get()` no-ops regardless of whether the ghost also reached `childrenOf` or `entityNameById`. So the only symptom those three can still produce is your spurious collision suffix, and that needs **453 and 500 broken together**:

- 500 alone broken → the ghost enters the sibling group but computes an `e<id>` fallback name, which cannot collide
- 453 alone broken → the ghost never enters the group, so its name is never read

Mutating 453+500 together fails the new collision test, reproducing your `["Wall"] → ["Wall_1"]` probe exactly.

I did not force individual pins. Doing so would mean engineering a ghost's numeric id so its `e<id>` fallback happens to string-match a survivor's real name — asserting an accident rather than the guard, which seemed worse than an honest "these three are jointly pinned."

Export **343 passing / 29 skipped**, typecheck 34/34. Test-only, no changeset.

Our counts differ slightly — you saw 338, I see 341 pre-change — most likely a reporter or skip-count difference rather than divergent behaviour, since `lod1-generator.test.ts` loaded here without a wasm build. Flagging in case it means something on your side.

louistrue:
Verified `18887104`. The UUID gate is now genuinely pinned, and the collision behaviour is pinned too — but the coverage is thinner than one gate per test, so here is the exact map rather than a pass/fail.

Individually removing each gate:

| line | loop | alone |
|---|---|---|
| 233 | node collection | **fails 3** |
| 407 | UUID assignment | **fails 1** ← newly pinned by your test |
| 453 | entity name index | passes |
| 484 | ancestor walk | **fails 4** |
| 500 | `childrenOf` grouping | passes |
| 507 | root-level grouping | passes |

So the regression I flagged (407 losing its detector under the re-parenting refactor) is closed.

For the remaining three, the behaviour **is** pinned, just not gate-by-gate. Removing combinations:

```
453 + 500 + 507  -> fails (survivor gets Wall_1)
453 + 500        -> fails
453 + 507        -> passes
500 + 507        -> passes
any one alone    -> passes
```

They are redundant rather than each necessary: `453 + 500` is the minimal pair that changes the observable output, and any single gate surviving is enough to keep the survivor's name bare. That is defence in depth, and I would not add gates or tests to force one-to-one coverage — the behaviour that matters is asserted, and it fails when the protection is actually gone.

Worth knowing the shape though: a refactor that drops 453 alone passes everything, and a later one dropping 500 is what breaks it. The test names the behaviour, not the mechanism, so whoever hits that failure will be looking at a change that seems unrelated to the one that removed the first gate.

`@ifc-lite/export` 342 passing.

One correction to my own method, since it nearly produced a wrong answer here: my first attempt at the pair testing used `set -- $pair` in zsh, which does not word-split, so the mutation script silently failed and every pair reported "26 passed". That reads exactly like "no pair fails". I only caught it because the Python error printed above the result — the assertion I added afterwards (`assert n==2`) is what should have been there from the start.

louistrue:
Correction to the line above: `@ifc-lite/export` is **340 passing** (340 passed | 29 skipped, 369 total), not 342. I typed the figure while the run was still finishing instead of reading it off. Nothing else in that comment changes — the gate map and the combination results were all read from actual runs.

BIMvoice:
Closing the count discrepancy I raised, and reporting an integration check across all five of our open PRs.

**The 338 vs 341 gap was neither of us being wrong — `main` was moving under both measurements.** Export-touching commits landed roughly hourly today (#2030 10:55, #2036 13:10, #2039 14:45, #2041 15:50), each adding tests to the same package. Measured from clean isolated builds:

```
main a8e58a2b (13:10)          331 passed
main c65bdbe0 / 818990b7       333
main d9abe5b4 (current)        333
#2046 branch head alone        343
#1967 + #2045 + #2047 merged   344
```

No skip-count anomaly, no missing fixture, no stale `dist` — every figure came from a fresh `pnpm install` + build in a throwaway worktree. 338 and 341 are snapshots at different points on that curve. I didn't bisect to pin exactly which commits produced each; say if you want that nailed down rather than explained.

**Integration across all five open PRs is clean**, checked properly rather than trusting the `MERGEABLE` flag — that flag already produced one false negative today, on #1967's rebase.

- Textual: all five merge into current `main` individually, all **ten pairs** merge, and all five merge together as one tree. Zero conflicts, including `scripts/api-surface.json`, which #1963 and #1967 both touch in different alphabetically-sorted sections.
- Semantic, verified by actually merging and running rather than reasoning: **#1967 × #2045** genuinely overlap in `mutable-property-view.ts` — #1967 rewrites `deleteEntity`/`restoreNewEntity` and adds the forgotten-entity stash, #2045 rewrites `applyMutations`' skip logic including the `DELETE_ENTITY` case that calls the modified `deleteEntity()`. Merged both on current main: **mutations 128 passed / 10 skipped**. They agree on what a forgotten-created entity's history looks like whether reached directly or through replay.
- One suspicion of mine was **wrong**: I flagged #2045 × #2047 as likely to interact. They don't. #2047's `options.applyMutations` is a boolean flag, and `MutablePropertyView.applyMutations()` is a method — same name, unrelated things. No branch touches `effective-index.ts`. Merged #1967+#2045+#2047: **export 344 passed**.

**Suggested landing order**, low-risk first so the mutations overlap stays simple to review: **#2051 → #1963 → #2047 → #2045 → #1967**. Verified compatible in either order for the last two; putting #1967 last means its `getEffectiveChanges()` is reviewed against an already-merged two-pass `applyMutations` rather than the reverse. Entirely your call — no rebase is needed for any of them.

louistrue:
Holding this one too. The CodeRabbit MAJOR on the document-root fallback is real, though narrower than it reads — I checked both cases.

**The common case works.** Project -> storey -> wall, delete only the storey:

```
WALL EXISTS: true | REACHABLE: true
NODES: [{kids:["Project1"]}, {n:"Project1", kids:["Wall1"]}, {n:"Wall1", kids:[]}]
```

The wall re-parents to the surviving project and stays reachable. That is the case `e8dbc7c7` was written for and it is correct.

**The degenerate case orphans the survivor.** When *no* ancestor survives (here the whole chain above the wall is deleted):

```
WALL NODE EXISTS: true
WALL REACHABLE FROM ANY PARENT: false
NODES: [{"name":"Wall1","kids":[]}]
```

The node is emitted and nothing lists it as a child, so it is unreachable from the root — the exact failure `e8dbc7c7` set out to fix, just in the branch where the ancestor walk returns `undefined`. `childrenOf.get(undefined)` is only read by the collision-naming loop, so that bucket never reaches an emission path.

This also makes two claims inaccurate rather than merely incomplete:

- `ifc5-exporter.ts:463-469` says such a child falls back to the document root
- `.changeset/ifc5-export-overlay-deletions.md:11` says the same

Neither is true today. Your own cycle test is named `falls back to the root bucket` and passes — but it passes because the child is *in the bucket*, not because anything emits it, so a cycle-affected survivor is orphaned the same way.

Either merge the `undefined` bucket into the document root node's `children`, or say plainly in both places that these children are dropped from the tree. I would take the first, since the second means a delete can silently remove an undeleted element from the exported hierarchy.

The gate map from my earlier comment is unchanged and fine: 233, 407 and 484 each fail tests alone; 453/500/507 are redundant with each other and the behaviour they protect is pinned by the collision test when all three go.

BIMvoice:
Fixed in d2c943b6, taking your first option. **One scope decision I made that differs from the literal wording — please push back if you disagree.**

You said "merge the undefined bucket into the document root node's children". I merged the *delete-orphaned subset* of that bucket, not the whole bucket. The bucket also collects entities that never had a parent at all — with `onlyTreeEntities: false` that is every non-spatial entity — and those were never in the tree, so no deletion removed them. Hoisting them under the root would invent hierarchy rather than restore it, and it breaks the existing pin `regression: a non-deleted entity still exports exactly as before` (`file.data.length === 1` for a lone parentless wall). So `rootedByDeletion` records only children where `resolveSurvivingParent` returned `undefined` after having a `parentOf` entry. If you meant the bucket literally, say so and I'll widen it.

The root node is now emitted whenever it has children, not only when a project exists — your repro had no root node at all, because the project itself was deleted.

**Your two cycle tests were passing for the wrong reason, and it was worse than "bucket membership".** `falls back to the root bucket` only asserted the names were present. The other one actively asserted **non**-reachability — "no OTHER node may reference Wall1" — so it had encoded the bug as the expected behaviour. Both now BFS `children` from the document-root path via a `reachablePaths()` helper, so membership can't satisfy them. RED on 18887104:

```
× a survivor whose entire ancestor chain is deleted is reachable from the document root
    expected undefined to be defined        (:964 — no root node exists)
× a cycle in the deleted-ancestor chain ... falls back to the root bucket
    expected false to be true               (:1095 — reachablePaths(file).has(wall))
× a cycle that loops back through the surviving child itself ...
    expected undefined to be defined        (:1161)
```

**Gate matrix re-run**, since my edit shifted the line numbers and you caught us going dark last time:

| gate | was | now | alone |
|---|---|---|---|
| node collection | 233 | 250 | **4 fail** (was 3) |
| UUID assignment | 407 | 442 | **1 fail** |
| entity name index | 453 | 488 | passes |
| ancestor walk | 484 | 521 | **4 fail** |
| `childrenOf` grouping | 500 | 538 | passes |
| root-level grouping | 507 | 551 | passes |

`488+538` → fails the collision test; `488+551` and `538+551` → pass; all three → fails. Identical redundancy triangle to your map. **No previously-pinned gate went dark and sensitivity strictly increased.** Three new gates are pinned by mutation instead: the `rootedByDeletion.add`, the root-merge loop, and the root emission condition — each kills all three reachability tests. I re-ran that mutation myself on the pushed tree (replacement count printed and grep-confirmed the line was gone): 3 failed / 341 passed, exactly the three.

Both false prose claims now match the code — the exporter comment and `.changeset` line 11.

On CodeRabbit's `as any`: it flagged one at ~698, but the identical quantities literal appeared **11 times**. All 11 now use a real `QuantityTableBuilder`. The remaining six `as any` in that file are a different, pre-existing shape (`{ meshes } as any` on the constructor) and I left them alone.

344 passed / 29 skipped, typecheck 34/34. Parquet still out of scope in #2046.

BIMvoice:
Pushed b46792cb. A review comment pointed at the hierarchy building links from pre-filtered relationships; I probed it before touching anything, and it reproduces — the same orphaning this PR is about, reached through the **visibility** filter instead of deletion.

The node loop gates on three things (tombstone, `visibleIds`, `treeIds`), but the UUID assignment, name index, ancestor walk and `childrenOf` grouping gated on the tombstone alone. So a hidden entity still got a UUID, and `addChild`'s `entityUuids.get()` guard — the thing that stops a *deleted* child dangling — could not stop a hidden one. Probe on Project -> Storey -> Wall, against d2c943b6:

```
hide Wall      DANGLING: storey-guid.children[Wall1] -> wall-1-guid   (no Wall node)
               ...identical with onlyTreeEntities on and off
hide Storey    DANGLING: project-guid.children[Storey1] -> storey-guid
               UNREACHABLE: wall-1-guid
isolate Wall   UNREACHABLE: wall-1-guid   (no document root emitted at all)
```

**One predicate, not a second guard.** `export` now derives `isOmitted` = deleted OR not-visible OR not-in-tree, and feeds it to both the node loop and `buildEntityMaps`. `resolveSurvivingParent` → `resolveEmittedParent`, `rootedByDeletion` → `rootedByFiltering`. A survivor stranded by any filter re-parents to its nearest emitted ancestor or attaches to the root, exactly as a deletion-stranded one already did. Adding a parallel visibility guard is how these two drifted apart to begin with.

**Your separation ruling is intact.** `buildVisibleSet` is untouched and no filter changes which entities it excludes — only the emitted tree's internal consistency. Your pinned regression at `:785` still passes, and it is now *also* killed by two of the mutations below, so it pins strictly more than before. No conflict between the two goals arose; I had told the agent to stop and report if one did.

**The tree disjunct was unpinned, and I did not paper over it.** Mutating it away failed nothing. Rather than leave a dark term or quietly drop it, I found where it is load-bearing: `buildTreeEntitySet` adds containment-map *values* but not their *keys*, so a container reachable only as a map key is filtered out while its contents are kept — the tree filter's own hidden-parent case. That is now a test, and re-mutating fails it. Note this is narrower than the probe's tree-filter-only case, which stays clean; I changed nothing about what the tree filter excludes.

**Gate matrix, re-run at the new line numbers:**

| gate | old | new | before | after |
|---|---|---|---|---|
| node collection | 250 | 265 | fails 4 | **fails 10** |
| UUID assignment | 442 | 459 | fails 1 | fails 1 |
| name index | 488 | 505 | passes | passes |
| ancestor walk | 521 | 538 | fails 4 | **fails 7** |
| childrenOf grouping | 538 | 554 | passes | passes |
| root bucket | 551 | 567 | passes | passes |
| name index + childrenOf | 488+538 | 505+554 | fails collision | fails collision |

No previously-pinned gate went dark; two got stronger. I re-ran the decisive one myself on the pushed tree — reverting `buildEntityMaps` to the deleted-only predicate (`REPLACEMENTS=1`, line re-read) fails all five new tests, 344 passed.

New assertions go through the existing `reachablePaths()` BFS plus a no-dangling check applied to **every** emitted node, not just the one under test. 349 passed / 29 skipped, typecheck 34/34. Changeset updated.

louistrue:
Verified `b46792cb`. The orphan case is closed, and the fix did something better than patch it — it made the existing cycle tests mean what their names say.

Entire ancestor chain deleted (storey **and** project), which was the failing case:

```
before: WALL EXISTS: true | REACHABLE: false   NODES: [{"Wall1", kids:[]}]
after : WALL EXISTS: true | REACHABLE: true    NODES: [{"(root)", kids:["Wall1"]}, {"Wall1", kids:[]}]
```

Removing `rootedByFiltering.add(childId)` fails **five** tests:

```
× a survivor whose entire ancestor chain is deleted is reachable from the document root
× a cycle in the deleted-ancestor chain does not hang the exporter — falls back to the root bucket
× a cycle that loops back through the surviving child itself does not make it its own parent
× isolating a leaf still emits a document root that reaches it
× a container outside the tree set does not strand the contained elements it parents
```

The second and third are the interesting ones. Before this commit those passed while the child sat in the `undefined` bucket unemitted — the test name said "falls back to the root bucket" and bucket membership was all it checked, so a cycle-affected survivor was orphaned exactly like the case I reported and nothing noticed. Now they fail without the rooting, so they assert reachability rather than bookkeeping.

The last two are new ground I had not raised: `visibleOnly` isolation and an out-of-tree container are two more routes to the same stranding, and they are now covered by the one gate rather than by separate special cases. Deriving the emitted tree from a single emission gate is what makes that hold — the earlier shape had the reachability decision spread across the grouping passes, which is why removing any one of them changed nothing observable.

`@ifc-lite/export` 346 passing. (`lod1-generator.test.ts` fails to load here on an unbuilt `@ifc-lite/wasm` — worktree staging, not this branch.)

That closes both items I had open on this PR: the untested UUID gate from before, and the document-root fallback.

louistrue:
Merging. A local CLI run against this head found two minors, neither touching shipped behaviour — recording them here as follow-ups rather than holding a verified correctness fix on them.

**1. `ifc5-exporter.test.ts:874` finds the root by position, not identity.**

```ts
const rootNode = file.data.find((n: any) => n.path === file.data[0].path);
```

That is a tautology — it returns `file.data[0]`. It happens to be right today because the document root is emitted first, so the walk below it is checking the tree it means to. But the file already has `DOCUMENT_ROOT_PATH` (line 88) and uses it properly at line 108, so this one test is coupled to emission order for no reason. If the root ever stops being first, the test walks a different subtree and its assertions become meaningless without failing.

**2. The changeset's ParquetExporter note says "tracked separately in #2046"**, which is the issue this PR closes. It is not wrong exactly — #2046 stays open for the Parquet half — but "tracked separately in" reads as pointing at a different issue, and this text ships in `@ifc-lite/export`'s CHANGELOG. "remains open in #2046 for the Parquet half" would say it.

Merging on the strength of the behaviour verification: removing `rootedByFiltering.add(childId)` fails five tests covering the ancestor-chain-deleted case, both cycle variants, `visibleOnly` isolation, and out-of-tree containers. The two cycle tests are the ones worth noting — before this commit they passed while only checking bucket membership, so they now assert reachability rather than bookkeeping.

Issue #2048, opened 2026-08-03
Two more one-sided overlay guards: georef CRS replacement, and IFCX ops for a tombstoned entity


BIMvoice:
Found by deliberately sweeping for the shape behind four bugs fixed today — **a guard applied to one traversal while a second traversal over related data ignores it**. Both are verified with repros against `main` (`c65bdbe0`).

Neither is the same site as #2044/#2046; both are new.

---

## 1. A replacement CRS is silently lost when the old one was deleted

`packages/export/src/step-exporter.ts:508-509`

```ts
const existingCrsIds = this.dataStore.entityIndex.byType.get('IFCPROJECTEDCRS');
const existingMcIds  = this.dataStore.entityIndex.byType.get('IFCMAPCONVERSION');
```

Every other emit path in this file routes through `willBeEmitted` / `effective.isDeleted`. This one reads the **raw** index.

So when the session has tombstoned the file's `IfcProjectedCRS`, `existingCrsIds.length` is still non-zero. The georef code takes the *modify existing* branch and queues an attribute mutation against a tombstoned id — instead of the *create new* branch. The source-iteration loop then skips that entity, and the mutation applies to nothing.

**Repro** — `infra-bridge.ifc` (has `#18=IFCPROJECTEDCRS(...)`), `deleteEntity(18)`, then export with `georefMutations.projectedCRS = { name: 'NEW:1234', … }`:

> output contains **zero** `IFCPROJECTEDCRS` lines and no `NEW:1234`

Not modified, not recreated — gone, with no error or warning. `IfcMapConversion` (`:509`, `:538-556`) has the identical shape.

Severity: silent loss of georeferencing rather than an invalid file, so it fails quietly rather than loudly.

---

## 2. A tombstoned entity's property edits survive into the published IFCX layer

`packages/mutations/src/change-set-to-ops.ts:150-165, 226-227`

`entityOps` is correctly last-write-wins — a `DELETE_ENTITY` after a `CREATE_ENTITY` collapses to `tombstone-entity`. But `components`, built by a **second pass** over the same mutation list, is never intersected with which entities ended up tombstoned.

`CREATE_PROPERTY entity=5 …Bar=42` followed by `DELETE_ENTITY entity=5` emits **both**:

```
{ op: 'tombstone-entity', entity: 'GUID-5' }
{ op: 'set-component',   entity: 'GUID-5', componentKey: 'pset:Pset_Foo', values: { Bar: 42 } }
```

[...]

PR #2050, opened 2026-08-03
fix(mutations): drop component ops for a tombstoned entity (#2048)

BIMvoice:
Closes **finding 2** of #2048. Finding 1 (the georef CRS gap in `step-exporter.ts`) is deliberately untouched — that file is in the area you have been reshaping across #2011/#2036, so it stays flagged for you.

## What was wrong

`entityOps` is last-write-wins, so `DELETE_ENTITY` after `CREATE_ENTITY` correctly collapses to `tombstone-entity`. But `components` is a **second pass** over the same mutation list, never intersected with it. An entity edited then deleted emitted both:

```
{ op: 'tombstone-entity', entity: 'GUID-5' }
{ op: 'set-component',   entity: 'GUID-5', componentKey: 'pset:Pset_Foo', values: { Bar: 42 } }
```

`publish.ts:85-131` merges both onto the same `IfcxNode` regardless of order, so a published layer carried live `bsi::ifc::prop::Bar: 42` next to `IFCLITE_ATTR.DELETED: true`.

## Where the filter runs, and why it matters

**After both passes**, against final `entityOps` state — not inside the component fold. An entity's terminal state is only known once the whole list is consumed, so filtering during the fold would miss a delete arriving *after* the property mutation. That is precisely the forward-pass hazard that produced the #2044 ordering hole, and it seemed worth not repeating it two issues later.

## Tombstone-then-recreate

Tested, and it survives correctly: `entityOps` being last-write-wins means DELETE-then-CREATE for one identity resolves to `add-entity`, the filter doesn't fire, and components from both sides are kept.

Not reachable through `StoreEditor` today — `addEntity` refuses to reuse a just-tombstoned id — but `changeSetToOps` imposes no such constraint of its own, so a future or alternate producer could emit it. Cheap to pin now.

## One test was split, not edited to fit

The pre-existing "expresses deletions" test asserted component-level deletions **and** an entity deletion on the same entity — which is the defect itself. Rather than adjust its expectation to match the new behaviour, it is now two tests: component-level deletions on a *surviving* entity, and an entity-level delete alone. Neither weakens what was covered.

Removing the filter fails exactly the new test and nothing else. Mutations **79**, viewer **2360 / 0 / 2**, typecheck 34/34.

## Left undefended on purpose

`publish.ts` still merges ops without checking for a tombstone. With this fix it never receives conflicting ops, so nothing is broken — but it isn't independently defended. I'd rather raise that than add a second guard enforcing the same invariant, since two places obscure which one is load-bearing. Say if you'd prefer belt-and-braces there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Entity deletions now correctly remove associated component changes.
  * Components are preserved when an entity is recreated or ultimately remains active.
  * Improved handling of complex create, delete, and recreate sequences.

* **Tests**
  * Added coverage for entity tombstones, component deletions, recreations, and entities without direct entity operations.

### DIFF
--- a/.changeset/tombstone-entity-drops-components.md
+++ b/.changeset/tombstone-entity-drops-components.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/mutations": patch
+---
+
+Drop component ops for entities whose final change-set state is `tombstone-entity` (#2048).
+
+`changeSetToOps` folds `entityOps` (last-write-wins per entity) and `components` (per componentKey) in the same pass over the mutation list, but previously emitted every accumulated component op regardless of whether the entity ended up deleted. A `CREATE_PROPERTY` mutation followed later by `DELETE_ENTITY` for the same entity produced both a `tombstone-entity` op and a `set-component` op carrying the now-meaningless property values — which `apps/viewer`'s `buildDeltaNodes` merged onto the same `IfcxNode`, publishing live property values alongside `IFCLITE_ATTR.DELETED: true`.
+
+Component ops are now filtered against each entity's final `entityOps` state after both passes complete (not during the fold, since an entity's terminal state — LWW — is only known once the whole mutation list has been consumed). An entity tombstoned and then recreated in the same change set keeps its components, since `entityOps` resolves to `add-entity` for that identity.
--- a/packages/mutations/src/change-set-to-ops.test.ts
+++ b/packages/mutations/src/change-set-to-ops.test.ts
@@ -29,7 +29,14 @@ function changeSet(mutations: Mutation[]): ChangeSet {
 }
 
 const resolver: EntityIdentityResolver = {
-  globalIdOf: (expressId) => (expressId === 42 ? '3fAx$GlobalId42' : undefined),
+  globalIdOf: (expressId) => {
+    if (expressId === 42) return '3fAx$GlobalId42';
+    if (expressId === 5) return '3fAx$GlobalId05';
+    if (expressId === 6) return '3fAx$GlobalId06';
+    if (expressId === 9) return '3fAx$GlobalId09';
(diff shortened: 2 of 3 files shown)


louistrue:
Verified `1a50c659`. The fix is right, the filter is pinned, and the placement argument holds — I checked it in both directions rather than only the one your test covers.

Mutation: removing `if (entityOps.get(entity)?.op === 'tombstone-entity') continue;` fails `drops component ops for an entity whose final state is tombstone-entity (#2048 finding 2)` and nothing else.

Then the three cases that decide whether filtering after both passes is actually load-bearing:

```
EDIT-then-DELETE : [tombstone-entity]                      <- your test's order
DELETE-then-EDIT : [tombstone-entity]                      <- the reverse
OTHER ENTITY     : [tombstone-entity(#5), set-component(#7)] <- unaffected
```

So the terminal-state read is genuinely order-independent, and a tombstone for one entity does not suppress a different entity's components. That last one is the over-suppression risk and it is clean.

One small gap: the reverse order (`DELETE_ENTITY` then `CREATE_PROPERTY`) is correct but only asserted by my probe, not by a test. Your comment argues placement in terms of a delete arriving *after* the property mutation, which is the direction the suite covers; the opposite order exercises the same filter through a different path and would be a cheap third case alongside the recreate test you already wrote.

Agreed on leaving finding 1 alone. The georef CRS gap sits in `step-exporter.ts`, which has moved under #2011/#2036 today, and bundling it here would put an unrelated change into a file that is already the busiest one in the tree.

`@ifc-lite/mutations` 79 passing. The `keeps components for an entity that is tombstoned and then recreated` test is the one I would have most wanted and it is already there — LWW keyed by entity means recreate wins, and that is the case where a naive "any DELETE_ENTITY suppresses" filter would have silently eaten live data.

Issue #2049, opened 2026-08-03
CI excludes ifc-lite-wasm entirely: its unit tests and clippy have never run, and two lint violations are on main now


BIMvoice:
Follow-up to the narrower version of this I raised on #1970. It is larger than I described there, and there is live proof rather than a hypothetical.

`.github/workflows/test.yml`:

```
470:  cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings
472:  cargo test  --workspace --exclude ifc-lite-wasm
```

## The proof

Running the excluded lint locally against current `main`:

```
error: manual implementation of `.is_multiple_of()`
  --> rust/wasm-bindings/src/api/bool2d.rs:70:12
error: this function has too many arguments (8/7)
  --> rust/wasm-bindings/src/zero_copy/mesh.rs:342:5
error: could not compile `ifc-lite-wasm` (lib) due to 2 previous errors
```

**Two real clippy violations are sitting on `main` right now, with the lint gate green.** Not a risk of drift — drift that has already happened and been invisible.

## The exclusion is broader than its own justification

The comment explains the exclude as wasm-bindgen-test dev-deps targeting wasm32. That is true of exactly **one** file — `tests/mesh_determinism.rs`, a `#[wasm_bindgen_test]`. Everything else in the crate is plain native `#[test]`, needing no wasm32 toolchain: `cargo test -p ifc-lite-wasm --lib` runs clean natively in well under a second.

So a justification that applies to one integration test is excluding the crate's entire unit-test suite *and* its whole clippy lint from every PR.

The wasm-pack lanes don't cover it either: `determinism.yml` and `wide-arithmetic.yml` run `wasm-pack test --node … --test mesh_determinism`, which is schedule/dispatch-only and touches only that one integration test. Nothing PR-triggered reaches the `--lib` tests.

## Why it matters beyond the two lints

This is the crate holding the pre-pass discovery paths. On #1970 two regression tests there went vacuous when #1969 merged and nothing noticed — they were green, pinning nothing, *and* they were never running in CI regardless. I added a CI-visible test in `rust/processing` as a workaround, but the underlying gap stayed.

## Suggested direction

[...]

PR #2051, opened 2026-08-03
fix(wasm): clear the two clippy violations sitting on main (#2049)

BIMvoice:
The code half of #2049. **No workflow change** — see below.

`cargo clippy -p ifc-lite-wasm --all-targets -- -D warnings` fails on current `main` with two errors. They've been invisible because CI runs `--workspace --exclude ifc-lite-wasm`, so this crate's lint has never executed on any PR.

**`bool2d.rs:70`** takes clippy's own rewrite — `!coords.len().is_multiple_of(2)`. Byte-identical semantics for an odd-length check on a `usize`.

**`mesh.rs:299`** (`MeshDataJs::set_texture`, 8 args) takes an `#[allow]` with a stated reason rather than a wrapper struct. The method sits inside a `#[wasm_bindgen]` impl, so it **is** the public JS API — every argument is a separate JS call parameter. A Rust struct wouldn't reduce the arity JS callers see; `wasm_bindgen` would still unpack a JS object field by field. That buys an indirection layer and no clarity, and the eight fields are already one cohesive operation. Happy to switch to a struct if you'd rather, but it looked like ceremony to satisfy a lint rather than a real improvement.

## Deliberately not included

Narrowing the `--exclude` in `test.yml` is your call and stays flagged in #2049. Keeping it out of this PR means that when the gate is turned on it's a clean change, rather than one that simultaneously appears to break something.

## Verification

`cargo clippy -p ifc-lite-wasm --all-targets -- -D warnings` is now **clean** — worth stating explicitly, since the crate had never been linted and fixing two could easily have uncovered more. Nothing was hiding behind them.

`cargo test -p ifc-lite-wasm --lib` **27/27**. Those have never run in CI either.

No changeset: neither change alters runtime behaviour or the wasm API surface. Say if you'd prefer one as a paper trail regardless.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for the JavaScript-facing texture attachment API, including its parameter requirements.
  * Clarified support for texture attachment calls with multiple parameters.

* **Maintenance**
  * Improved validation readability while preserving existing behavior.
  * Updated internal size tracking to reflect the documentation additions.
  * No user-facing functionality or behavior changes.

### DIFF
--- a/rust/processing/tests/module_size_allowlist.txt
+++ b/rust/processing/tests/module_size_allowlist.txt
@@ -234,7 +234,10 @@
 # +61: #1781 external image textures — textureId/textureUrl fields + getters, set_texture_ref, and the get/takeMesh/Clone field threading.
 # +17: #1891 follow-on — geometry_aabb_values field + geometryAabbValues getter (the doc carries the Y-up frame contract and the world-box rationale so a future reader does not re-derive them), the four constructor/Clone initializers, and the #[cfg(test)] include for mesh_tests.rs (the test file itself is ratchet-exempt). Net +17 and not +56, because the two Z-up→Y-up conversion helpers moved out to a new sibling frame_swap.rs (107 lines, under the house limit, no row): the box swap is now shared by the f32 local bounds and the f64 world AABB, so the min/max reversal lives in exactly one place.
 # -46: #1891 closedness added geometry_volume_values + geometry_closure_flags, and the whole per-entity fingerprint surface (the five index-parallel arrays' getters — including the Y-up swap on the way out — plus GeometryFingerprint and push_geometry_hash) moved OUT to the child module zero_copy/mesh_fingerprint.rs (new, 161 lines, under the house limit, no row) — net down despite two new arrays; budget refreshed downward.
-     724 rust/wasm-bindings/src/zero_copy/mesh.rs
+# +3: #2049 — `set_texture` was the one 8-arg fn in this crate missing the
+# `#[allow(clippy::too_many_arguments)]` its 21 peers carry. The crate's lint
+# has never run in CI (`--exclude ifc-lite-wasm`), which is how it drifted.
+     727 rust/wasm-bindings/src/zero_copy/mesh.rs
      743 rust/wasm-bindings/src/zero_copy/symbolic.rs
 # +65: #1910 follow-up — buildPrePassOnce's combined_pre_pass is a third
 # geometry-job discovery path with the same has_geometry_by_name gap as the
--- a/rust/wasm-bindings/src/api/bool2d.rs
+++ b/rust/wasm-bindings/src/api/bool2d.rs
@@ -67,7 +67,7 @@ impl Contours2D {
     /// boolean would keep rather than exposing an unsanitised soup.
     #[wasm_bindgen(constructor)]
     pub fn new(coords: &[f64], ring_lengths: &[u32]) -> Result<Contours2D, JsValue> {
-        if coords.len() % 2 != 0 {
+        if !coords.len().is_multiple_of(2) {
             return Err(js_sys::Error::new("Contours2D: coords length must be even").into());
         }
         // Checked arithmetic: on wasm32 `usize` is 32-bit, so a crafted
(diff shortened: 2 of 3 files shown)


BIMvoice:
An independent audit of this crate turned up a fact that makes the `#[allow]` choice stronger than I argued for it:

**21 other functions in `ifc-lite-wasm` already carry `#[allow(clippy::too_many_arguments)]`.** So `set_texture` wasn't a case needing a new judgement — it was the one 8-argument function in the crate *missing* the annotation its 21 peers already have. That reframes the change from "we chose an allow over a refactor" to "we restored consistency with the established convention in this file," which is a better reason than the one I gave.

Worth stating because the alternative reading — that a lint was silenced to make a PR green — is exactly what a reviewer should suspect. Here the crate had already decided this question 21 times.

The same audit read the crate as one no automated check has ever examined, and found nothing else: zero `unsafe` blocks, no `TODO`/`FIXME`/`XXX` on live paths, no `#[ignore]`d tests, no assertion-free tests, `cargo check --all-targets` clean, and `unwrap`/`expect` confined to test functions (fixture loading and a deliberate poisoned-mutex recovery test). **Unlinted but sound** — which was the honest possibility worth confirming rather than assuming a neglected crate must be rotten.

BIMvoice:
CI caught a real miss — fixed in `c25089ae`.

`no_module_grows_past_its_ratchet_budget` failed: `zero_copy/mesh.rs` sits at **exactly** its 724-line budget, and the seven-line rationale I attached to the `#[allow]` pushed it to 732.

The reason I missed it is worth stating, because it is a scoping error rather than an oversight: I checked `rust/processing/tests/module_size_allowlist.txt` for **`rust/processing` paths**. The allowlist covers other crates too, including this one. I verified against where I assumed the constraint lived rather than where it actually reaches.

The rationale is now two lines, which is enough — it does not need to argue from first principles when 21 other functions in this crate already carry the same annotation. Budget bumped 724 → 727 with the written justification the allowlist requires.

There is a small irony worth recording: the drift is itself a symptom of #2049. The crate's lint has never run in CI, which is precisely how an 8-argument function ended up as the only one of its kind here missing the annotation its peers have.

Ratchet **4/4**, `cargo clippy -p ifc-lite-wasm --all-targets -- -D warnings` clean, `cargo test -p ifc-lite-wasm --lib` **27/27**.

Also worth noting for anyone reading the earlier failure: running CI's exact `cargo test --workspace --exclude ifc-lite-wasm` locally in a fresh worktree produces a *different* and misleading failure — `rust/export`'s csv/gltf tests panic on a missing `tests/models/ara3d/duplex.ifc` because fixtures aren't fetched. That is the local-artifact trap, not the CI failure; the CI log had to be read directly to find the real cause.

louistrue:
Verified `c25089ae`. Both violations are real, both are cleared, and the allowlist edit is honest.

Ran the exact CI command on each side:

```
main:        error: manual implementation of `.is_multiple_of()`
             error: this function has too many arguments (8/7)
             error: could not compile `ifc-lite-wasm` (lib) due to 2 previous errors
this branch: Finished `dev` profile — clean
```

So this crate has been carrying two `-D warnings` failures on `main`, exactly as you said, invisible because `--workspace --exclude ifc-lite-wasm` means its lint has never executed on any PR. That is the same shape as the #1970 case earlier today, where a regression test lived in the excluded crate and a revert passed CI fully green.

Your "21 peers" claim checks out precisely — 22 occurrences of `allow(clippy::too_many_arguments)` in `rust/wasm-bindings/src/` now, one of which is the line you added, so 21 before. `set_texture` genuinely was the odd one out rather than a new exception being carved.

The allowlist edit is the part I looked at hardest, since bumping a size budget is how growth gets laundered. It is not that here: 724 -> 727 is the three lines of the annotation and its comment, the file measures exactly 727, and the delta carries a `# +3:` note in the house style. `module_size_ratchet` passes (4 passed).

Agreed on the `#[allow]` over a wrapper struct. The method is inside a `#[wasm_bindgen]` impl, so the eight arguments *are* the JS call signature; `wasm_bindgen` would unpack a struct field by field and JS callers would pass the same eight values either way. A struct would move the arity rather than reduce it, and satisfy the lint without improving anything a caller sees.

Also agreed on leaving the `--exclude` narrowing out of this PR. Turning the lane on is a CI-cost and triage decision, and it belongs with whoever owns that call — bundling it here would mean this PR either turns a new gate on for everyone or quietly does not, and neither should ride along with a two-line lint fix.

Issue #2053, opened 2026-08-03
Sharded pre-pass doc claims a delegation that no longer exists, and with it a parity guarantee


BIMvoice:
Found by sweeping for prose that asserts behaviour the code does not have — a shape that produced four real findings today. Two survived verification; this is the one that matters.

## The claim, and what the code does

`rust/wasm-bindings/src/api/gpu_meshes/prepass_sharded.rs:90`:

> Delegates to `ifc_lite_processing::scan_shard`, the exact per-chunk primitive the native `build_entity_index_parallel` fans across cores — **so the sharded merge cannot drift from the serial builder**.

Line 111 of the same function:

```rust
ifc_lite_processing::scan_shard_classified(data, start, end);
```

It calls `scan_shard_classified`, not `scan_shard`. Those are two independently-maintained loops over `EntityScanner` — `parallel_scan.rs:81` says so from the other direction, "Deliberately NOT delegating to `scan_shard_classified`".

Repo-wide, `scan_shard` has **no callers outside its own module**; only the native `build_entity_index_parallel` path uses it.

## Why this is worse than a stale comment

The sentence does not merely describe a call. It asserts a **safety property** — that the sharded merge cannot drift from the serial builder — and grounds it in a delegation that no longer happens. The two loops are logically equivalent today, so nothing is broken; the guarantee is simply no longer structural. Edit one without the other and parity breaks silently, with a doc comment still promising it cannot.

`git log -p` on that file shows an earlier version genuinely did delegate to `scan_shard` and said so. A later perf change swapped in the classified duplicate and left the prose.

This is adjacent to #1970: I added a CI-visible regression there through `scan_shard_classified` precisely because that is what the sharded path runs. Had I trusted this comment, I would have tested the wrong function.

Suggested fix: correct the sentence to name `scan_shard_classified`, and either state the parity relationship as a maintained invariant (with a test that fails if the two loops diverge) or drop the guarantee. The first is more useful — a parity test between the two loops would make the claim true again rather than merely accurate.

## Second, smaller

`packages/export/src/type-owned-psets.ts:35` says "The committed Duplex fixture has six of each". The Duplex model is **not committed** — `.gitignore` excludes `tests/models/**/*.ifc` as "Local test models (private, never committed)", `manifest.json` lists it as a checksummed download, and the corresponding test at `overlay-effective-model.test.ts:604` is `it.skipIf(!existsSync(DUPLEX))`.

[...]

Issue #1910, opened 2026-07-29
RTC offset is (0,0,0) when geometry hangs off IfcBuilding and coordinates are baked raw


louistrue:
**Type:** bug
**Area:** geometry / router / rtc
**Reported:** by email 2026-07-09

## Summary

An IFC file whose geometry is attached to `IfcBuilding` and whose vertices carry raw
projected coordinates (no `IfcMapConversion` offset, all placements at origin) loads
with correct metadata and hierarchy but renders **no visible geometry**.

The RTC offset is computed as `(0,0,0)`, the raw coordinates go into a `Float32Array`
unshifted, and the mesh collapses.

## Repro

Test file: `dgm_4x2.ifc` (1.8 MB, attached to the report).

- `FILE_SCHEMA(('IFC4X2'))`
- `#21=IFCPROJECTEDCRS('EPSG:25832', ...)` — UTM zone 32N
- `#22=IFCMAPCONVERSION(#19,#21,0.,0.,0.,1.,0.,1.)` — **eastings/northings/height all 0**
- `#29=IFCSHAPEREPRESENTATION(#19,'Horizont30','Tessellation',(#31))`
- `#31=IFCSHELLBASEDSURFACEMODEL((#32))` → `#32=IFCOPENSHELL(...)`, 5 217 `IfcFace`
- `#31335=IFCBUILDING('2RK8fpXHb9wxogL_4DBbK8',$,'Horizont30',$,$,#31336,#30,...)`
- `#31336=IFCLOCALPLACEMENT($,#15)` → `#15=IFCAXIS2PLACEMENT3D(#16,...)` → `#16=IFCCARTESIANPOINT((0.,0.,0.))`

Vertex coordinates are full UTM, e.g.
`#31326=IFCCARTESIANPOINT((496742.45807390002,5556616.4791799998,109.6418231))`.

Extent: X 496 741 … 497 123, Y 5 556 598 … 5 556 638, Z 0 … 120.2.

Expected: terrain renders.
Actual: nothing renders. Hierarchy, property sets and the EPSG code display correctly.

`IfcOpenShell`'s `IfcConvert` produces a valid OBJ from the same file.

## Root cause

1. `detect_rtc_offset_from_first_element` / `detect_rtc_offset_from_jobs`
   (`rust/geometry/src/router/rtc_offset.rs`) sample **placement translations**, and
   only for entities where `has_geometry_by_name(type_name)` is true.

[...]

PR #2054, opened 2026-08-03
fix(processing): enforce the shard-scan parity the doc only claimed (#2053)

BIMvoice:
Closes the first finding of #2053.

## The claim vs the code

`prepass_sharded.rs:90` said it *"delegates to `ifc_lite_processing::scan_shard` … so the sharded merge cannot drift from the serial builder."* Line 111 calls `scan_shard_classified`. `scan_shard` has no callers outside its own module, and the two are separately maintained loops — `parallel_scan.rs:81` says so from the other side.

The sentence asserted a **safety property** grounded in a delegation that had stopped happening. `git log -p` shows an earlier version really did delegate; a perf change swapped in the classified loop and left the prose behind.

## First question: do the loops still agree?

Checked before writing anything, because a real divergence would matter far more than a stale comment. **They agree.** Both push `(id, start, entity_end)` from the same `EntityScanner` with the same `range_end` cutoff and the same `new`/`new_at` selection; the classified one additionally computes a class byte.

## Correcting the sentence alone would have left the invariant unenforced

So the parity is now a test rather than a promise. It asserts identical records **and** identical handoff, and the boundary case sweeps **all 3694 byte offsets** of the fixture as split points, checking both shards at each — guaranteeing many splits land mid-record, which is where re-sync would diverge if it ever did.

**Where it lives is deliberate.** `rust/wasm-bindings` is excluded from CI entirely (`--exclude ifc-lite-wasm`, #2049), so a parity test placed next to the caller would never run — recreating the precise trap this issue is about. It's in `rust/processing/tests/`, which the required lane covers, and both functions are exported from `ifc_lite_processing` so it reaches them cleanly.

## Evidence

Offsetting one loop's `start` by a byte on every seventh record fails **both** tests with explicit left/right diffs. Parity **2/2**, processing lib **57/57**, ratchet **4/4**, wasm lib **27/27**.

No changeset — a doc correction plus a test, with no runtime change for any `@ifc-lite/wasm` consumer.

## Worth noting

This one had a concrete cost before it was found: on #1970 I added a CI-visible regression through `scan_shard_classified` precisely because that is what the sharded path runs. Had I trusted this comment, I'd have tested `scan_shard` — a function nothing in that path calls.

The second finding in #2053 (a comment citing "the committed Duplex fixture", which is gitignored and fetched on demand) is untouched here — happy to take it separately if you want it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency between shard scanning methods, including when entity boundaries are split across shards.
  * Ensured classified scan results and handoff behavior remain aligned with standard scanning.
  * Added broader validation across complete files and split positions to help prevent regressions.

* **Documentation**
  * Clarified shard scanning behavior, delegation, and per-record classification output.
  * Documented parity expectations between standard and classified shard scans.

### DIFF
--- a/packages/wasm/pkg/ifc-lite.d.ts
+++ b/packages/wasm/pkg/ifc-lite.d.ts
@@ -558,8 +558,11 @@ export class IfcAPI {
      * Byte offsets returned are GLOBAL (relative to file start), so shards
      * concatenate without rewriting. Returns a plain object:
      *   `{ ids: Uint32Array, starts: Uint32Array, lengths: Uint32Array,
-     *      handoff: number }`
-     * where `handoff` is the global start of the first entity at/after
+     *      classes: Uint8Array, handoff: number }`
+     * where `classes` is the parallel per-record prepass class byte
+     * (`PREPASS_CLASS_*`: named code in the low bits plus the geometry-job /
+     * type-candidate flags) the host filters on to rebuild pre-pass span
+     * lists, and `handoff` is the global start of the first entity at/after
      * `range_end` (the next shard's first real entity), or `-1` at EOF.
      */
     scanEntityIndexShard(data: Uint8Array, range_start: number, range_end: number): any;
--- a/packages/wasm/pkg/ifc-lite.d.ts
+++ b/packages/wasm/pkg/ifc-lite.d.ts
@@ -547,9 +547,13 @@ export class IfcAPI {
      * range_end)` shard; the main thread stitches the returned columns into the
      * full entity index (byte-identical to the single-threaded
      * `build_entity_index`) by binary-searching each shard for the previous
-     * shard's `handoff`. Delegates to `ifc_lite_processing::scan_shard`, the
-     * exact per-chunk primitive the native `build_entity_index_parallel` fans
-     * across cores — so the sharded merge cannot drift from the serial builder.
+     * shard's `handoff`. Delegates to `ifc_lite_processing::scan_shard_classified`
+     * — a separately-maintained loop over the same `EntityScanner` primitive as
+     * `scan_shard` (the one the native `build_entity_index_parallel` fans across
+     * cores), plus a per-record class column this sharded path also needs. The
+     * two loops' records/handoff are kept in parity by a dedicated test
+     * (`rust/processing/tests/issue_2053_shard_scan_parity.rs`), not by
+     * delegation — edit one without the other and that test catches the drift.
      *
      * Byte offsets returned are GLOBAL (relative to file start), so shards
      * concatenate without rewriting. Returns a plain object:
--- a/rust/processing/tests/issue_2053_shard_scan_parity.rs
+++ b/rust/processing/tests/issue_2053_shard_scan_parity.rs
@@ -14,10 +14,15 @@
 //! This test makes the "cannot drift" property in `prepass_sharded.rs`'s doc
 //! comment a maintained invariant instead of prose: it asserts the two
 //! functions produce IDENTICAL records (id, start, end) and handoff for the
-//! same input over the same byte range, covering both a whole-file single
-//! shard and a shard boundary landing mid-entity (the handoff/stitch case,
-//! where the two loops are most likely to diverge since it exercises the
-//! `range_end` cutoff and speculative-scan restart logic in each independently).
+//! same input over the same byte range, and additionally that
+//! `scan_shard_classified`'s extra `classes` column — the output that has no
+//! counterpart in `scan_shard`, and which the browser host indexes
+//! POSITIONALLY against the records — stays aligned with those records: same
+//! length, and each byte the class of the record at that index. Both are
+//! checked over a whole-file single shard and over a shard boundary landing
+//! mid-entity (the handoff/stitch case, where the two loops are most likely to
+//! diverge since it exercises the `range_end` cutoff and speculative-scan
+//! restart logic in each independently).
 //!
 //! `ifc-lite-processing` (this crate) is covered by the required CI lane
 //! (`cargo test --workspace --exclude ifc-lite-wasm`); `ifc-lite-wasm`, which
@@ -26,7 +31,23 @@
 //! specifically so a parity regression fails CI instead of only ever being
 //! caught locally.
 
-use ifc_lite_processing::{scan_shard, scan_shard_classified};
+use ifc_lite_processing::{classify_type_name, scan_shard, scan_shard_classified};
+
+/// Parse the raw STEP keyword at a record start (`#id=KEYWORD(...`). Mirrors
+/// `issue_1910_shard_class_ci_coverage.rs`'s copy of the same helper —
+/// duplicated rather than shared because integration tests are separate
+/// crates and the original (`prepass_discovery.rs::keyword_at`) is private to
+/// the excluded wasm crate.
+fn keyword_at(content: &[u8], start: usize, end: usize) -> &str {
+    let span = &content[start..end.min(content.len())];
+    let eq = span.iter().position(|&b| b == b'=').map(|p| p + 1).unwrap_or(0);
+    let kw_end = span[eq..]
+        .iter()
+        .position(|&b| b == b'(')
+        .map(|p| eq + p)
+        .unwrap_or(span.len());
+    std::str::from_utf8(&span[eq..kw_end]).unwrap_or("").trim()
+}
 
 /// A small synthetic DATA section with enough entities, and varied enough
 /// record lengths, that a boundary swept across the whole byte range lands
@@ -48,15 +69,37 @@ fn synthetic_ifc() -> String {
         }
     }
     content.push_str("ENDSEC;\nEND-ISO-10303-21;\n");
+
+    // The class oracle below is name-only (`classify_type_name`), so it is
+    // ground truth for this fixture only while none of its keywords can take
+    // the #1910 instance-level exception that `classify_type_name_with_content`
+    // adds on top. Asserted rather than assumed: add a keyword here that IS a
+    // spatial container and the oracle would silently start disagreeing with
+    // a correct `scan_shard_classified`.
+    for keyword in ["IFCPROJECT", "IFCWALL", "IFCDOOR"] {
+        assert!(
+            !ifc_lite_core::is_representationless_spatial_container_by_name(keyword),
+            "fixture keyword {keyword} takes the #1910 instance-level exception, so \
+             classify_type_name is no longer a valid oracle for its class byte"
+        );
+    }
     content
 }
 
 /// Compare the two scans' output for identical records (same ids, starts,
 /// lengths, i.e. (id, start, end) triples in the same order) and identical
-/// handoff over `[range_start, range_end)`.
+/// handoff over `[range_start, range_end)`, and check that the classified
+/// scan's extra `classes` column is positionally aligned with those records.
+///
+/// The alignment check does NOT compare `classes` against another
+/// `scan_shard_classified` call: the host's failure mode is this one function
+/// emitting a class column shifted against its own records, and a shift like
+/// that lands identically in every range decomposition, so a self-comparison
+/// across ranges cannot see it. The oracle is instead recomputed per record
+/// from the record's own keyword, which pins each byte to a specific record.
 fn assert_shard_scans_match(content: &[u8], range_start: usize, range_end: usize, label: &str) {
     let (plain_records, plain_handoff) = scan_shard(content, range_start, range_end);
-    let (classified_records, _classes, classified_handoff) =
+    let (classified_records, classes, classified_handoff) =
         scan_shard_classified(content, range_start, range_end);
 
     assert_eq!(
@@ -69,6 +112,23 @@ fn assert_shard_scans_match(content: &[u8], range_start: usize, range_end: usize
         "scan_shard vs scan_shard_classified handoff mismatch for {label} \
          (range [{range_start}, {range_end}))"
     );
+    assert_eq!(
+        classes.len(),
+        classified_records.len(),
+        "scan_shard_classified class column length must match its record count for \
+         {label} (range [{range_start}, {range_end})) — the host indexes classes \
+         positionally against the records"
+    );
+    for (i, &(id, start, end)) in classified_records.iter().enumerate() {
+        let keyword = keyword_at(content, start, end);
+        assert_eq!(
+            classes[i],
+            classify_type_name(keyword),
+            "class byte at index {i} does not belong to the record at that index \
+             (#{id} {keyword}) for {label} (range [{range_start}, {range_end})) — \
+             the class column is misaligned with the records"
+        );
+    }
 }
 
 /// Whole file scanned as a single shard: the simplest case, and the one the
--- a/rust/processing/tests/issue_2053_shard_scan_parity.rs
+++ b/rust/processing/tests/issue_2053_shard_scan_parity.rs
@@ -1,0 +1,108 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Parity regression for #2053: `scan_shard` (native
+//! `build_entity_index_parallel`'s per-chunk primitive) and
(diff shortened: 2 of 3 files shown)


BIMvoice:
CI caught a real one — fixed in `2aeba7b4`.

The corrected sentence sits on a `#[wasm_bindgen]` method, so wasm-bindgen emits it into `packages/wasm/pkg/ifc-lite.d.ts`. Changing the Rust doc without regenerating left the **committed types still asserting the exact claim this PR removes** — the old "delegates to `scan_shard`" text was sitting in the published `.d.ts` at line 550.

Which is a neat instance of the issue's own theme: the fix for a false claim in one place left the same false claim standing in another, generated one. Regenerated with `bash scripts/build-wasm.sh`, per the workflow's own error message; the diff is the doc comment and nothing else.

This also inverts a habit I have leaned on all session. I have been reverting `packages/wasm/pkg` churn as build noise, because a local `pnpm build` dirties it spuriously and it has caused false alarms repeatedly today. Here the churn **was** the deliverable. The distinction is simply whether the Rust source changed: if it did, the regenerated types are real and belong in the commit; if only a build ran, they are noise. I had been applying the second rule unconditionally.

Parity test still 2/2, and the rest of the checks were already green.

louistrue:
Verified `2aeba7b4`. This is the right shape for the problem, and the part I checked hardest is the one that makes or breaks it.

**The test can actually reach the code.** Both `scan_shard` and `scan_shard_classified` live in `ifc-lite-processing`, which the required lane covers, so the parity test runs in CI. The `rust/wasm-bindings` change is doc-only — filtering the diff for non-comment lines returns nothing. That matters because a fix in `ifc-lite-wasm` with its test in `processing` would be the #1970 shape in reverse, and this deliberately is not that: you put the test where CI can see it and said so in the module doc.

**It fails when the loops diverge.** Making `scan_shard_classified` record `entity_end - 1`:

```
test single_shard_whole_file_matches ... FAILED
test shard_boundary_mid_entity_handoff_matches ... FAILED
  assertion `left == right` failed: scan_shard vs scan_shard_classified record mismatch
    for single-shard-whole-file (range [0, 3695))
  ... for second-shard@1 (range [1, 3695))
```

Both cases catch it, and the mid-entity handoff one is the case worth having — that is where `range_end` cutoff and speculative-restart logic are duplicated in each loop and most likely to drift.

**Replacing a false claim with a maintained one** is the actual contribution here. The old doc said the sharded path "Delegates to `scan_shard` ... so the sharded merge cannot drift from the serial builder", and it did not delegate. A comment asserting an invariant that nothing enforced is worse than no comment, because it tells the next reader they need not check. Now the comment names the test that enforces it.

The regenerated `packages/wasm/pkg/ifc-lite.d.ts` is the same doc text and nothing else, which is correct — `.gitignore:57` documents that this file is deliberately committed so `tsc` / `turbo typecheck` resolve.

One thing worth noting for whoever lands this and #2051 together: this branch does not carry #2051's clippy fixes, so `cargo clippy -p ifc-lite-wasm -- -D warnings` still fails here for the two reasons in that PR. Not this PR's problem, and invisible to CI either way while the crate stays excluded.

louistrue:
One thing before this lands, and it is the same defect class the PR exists to close.

`scanEntityIndexShard` returns `{ ids, starts, lengths, classes, handoff }` — `classes` is the column this whole sharded path was changed to carry — but the documented return shape omits it, in both the Rust doc comment and the regenerated `packages/wasm/pkg/ifc-lite.d.ts`. So the corrected doc is still describing a return value that does not match the code.

Given the PR's own argument (a comment asserting something nothing enforces is worse than no comment, because it tells the next reader not to check), leaving a documented shape that omits a returned column seems worth the extra `bash scripts/build-wasm.sh` cycle.

Everything else verified and good:

- both `scan_shard` and `scan_shard_classified` live in `ifc-lite-processing`, so the parity test runs in the required lane — the `rust/wasm-bindings` change is doc-only, which I confirmed by filtering the diff for non-comment lines
- the test fails on real divergence: making the classified loop record `entity_end - 1` fails both `single_shard_whole_file_matches` and `shard_boundary_mid_entity_handoff_matches`
- the mid-entity handoff case is the one worth having, since `range_end` cutoff and speculative-restart logic are duplicated in each loop

BIMvoice:
Fixed in 8ee4ff02. You were right, and it is the PR's own defect class — I had corrected one false claim and left an adjacent one standing in the same comment.

`classes` is now in the documented shape, in both the Rust doc and the regenerated `.d.ts`:

```
{ ids: Uint32Array, starts: Uint32Array, lengths: Uint32Array,
  classes: Uint8Array, handoff: number }
```

I verified the type rather than taking CodeRabbit's word for it: `scan_shard_classified` (`shard_classes.rs:60`) returns the column as `Vec<u8>`, emitted at `prepass_sharded.rs:140` as `Uint8Array::from(classes.as_slice())`. `Uint8Array` is right. Since I was there anyway I named the encoding — `PREPASS_CLASS_*`, the named code in the low bits under `PREPASS_CLASS_CODE_MASK` plus the `0x80` geometry-job and `0x40` type-candidate flags — because "a class byte" would have been the same kind of under-description that got us here.

Ran your doc-only check myself:

```
$ git diff -U0 -- rust/wasm-bindings | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*(///|//)'
(no output)
```

Zero non-comment lines. And the regeneration carried no passenger churn — `.wasm`, `.js` and `package.json` all came back byte-identical, so `packages/wasm/pkg` shows only `ifc-lite.d.ts` modified, which is the same bar as 2aeba7b4.

Parity 2/2, processing lib 57/57, wasm lib 27/27. Still no changeset — doc plus generated types, no runtime change.

Noted on clippy: those two failures belong to #2051 and this branch doesn't carry its fixes.

louistrue:
Verified `8ee4ff02`. `classes: Uint8Array` is now in the documented return shape in both places, and the two agree:

- `rust/wasm-bindings/src/api/gpu_meshes/prepass_sharded.rs:101-102`
- `packages/wasm/pkg/ifc-lite.d.ts:560-566` (regenerated, same text)

The added sentence earns its place too — it says what the byte actually carries (`PREPASS_CLASS_*`, named code in the low bits plus the geometry-job / type-candidate flags) rather than just naming the type, which is the part a caller needs to filter on.

That closes the last item I had on this PR. Everything else stands from my earlier pass:

- both `scan_shard` and `scan_shard_classified` live in `ifc-lite-processing`, so the parity test runs in the required lane, and the `rust/wasm-bindings` change is doc-only
- the test fails on real divergence — `entity_end - 1` in the classified loop fails both `single_shard_whole_file_matches` and `shard_boundary_mid_entity_handoff_matches`

Resolving the thread.

louistrue:
One more before this lands — found by a local CodeRabbit run against this head (the hosted one is still on `2aeba7b4`), and I confirmed it in the source.

**The parity test discards the column the function exists for.** `issue_2053_shard_scan_parity.rs:59`:

```rust
let (classified_records, _classes, classified_handoff) =
    scan_shard_classified(content, range_start, range_end);
```

`_classes` is dropped, and the two assertions cover only records and handoff. So `scan_shard_classified`'s distinguishing output — the per-record class byte, the entire reason it does not delegate to `scan_shard` — is unvalidated. Return an empty `classes`, or one misaligned by a record, and both tests still pass.

That is narrower than the drift risk you set out to close, and the records/handoff parity is genuinely the higher-value half, so I would not call this a rewrite. But the cheap assertion is worth having:

```rust
assert_eq!(classes.len(), classified_records.len(), "...");
```

in both the whole-file and split-range cases. Misalignment is the failure that would actually reach production here — the host indexes `classes` positionally against the records to rebuild pre-pass span lists, so an off-by-one there mislabels every entity after the shift rather than erroring.

Worth noting given this PR's own thesis: the module doc now says the two loops "are kept in parity by a dedicated test ... edit one without the other and that test catches the drift." With `_classes` discarded, that sentence is true for records and handoff and not for classes. It is the same shape as the claim you replaced — an invariant asserted in prose that the test only partly enforces.

Everything else on `8ee4ff02` is verified and good, including the `classes: Uint8Array` documentation fix in both the Rust doc and the regenerated `.d.ts`.

louistrue:
Second finding from the same local run, and this one points outside the PR — worth its own issue rather than expanding this one.

**The host reads the class byte with raw literals instead of the canonical constants.** `apps/viewer/.../geometry-parallel.ts`'s span-list rebuild indexes on raw values 4–10 and does not apply `PREPASS_CLASS_CODE_MASK` before comparing. So the producer side now documents `PREPASS_CLASS_*` as "named code in the low bits **plus** the geometry-job / type-candidate flags", while the consumer compares the whole byte against bare numbers.

That is exactly the drift this PR exists to prevent, one layer further out: an added flag bit shifts every raw comparison silently, and the parity test cannot see it because it lives entirely on the Rust side.

I would not fold it in here — it is a different file, a different crate boundary, and this PR is already doing the documentation correction it set out to do. But it belongs on the issue list next to the `_classes` assertion, since the two together are the whole "is the class column actually trustworthy end to end" question.

For this PR, the `_classes` item from my previous comment is the one I would still want before merge.

BIMvoice:
Both fair. Taking the `_classes` one here; the host-side one is filed as #2065.

**On `_classes`: you're right, and the framing is the part that stings.** The module doc now says the test catches drift, which is true for records and handoff and not for classes — the same shape as the claim this PR set out to replace. I corrected a false statement about delegation and left an adjacent one standing about enforcement. Fix in progress, asserting alignment in both the whole-file and split-range cases, and I'm checking that a length assertion actually catches *misalignment* rather than only truncation — if it doesn't, the length check is not the assertion you want and I'll say so rather than ship the weaker one.

**On the host-side raw literals — filed as #2065, with one correction: it's latent today, not live.** I went looking for the live breakage and it isn't there yet:

`classify_type_name` (`shard_classes.rs:120-141`) returns the named codes early —

```rust
if named != PREPASS_CLASS_NONE {
    // The named keywords are mutually exclusive with the flag predicates …
    return named;
}
```

— so classes 4–10 never acquire a flag from that path, and the one later OR-in (`:98-103`) is gated on `is_representationless_spatial_container_by_name`, which none of the 4–10 keywords are. So the host's unmasked comparison is correct *by coincidence of the current classification rules*, not by construction.

Which I think strengthens your point rather than weakening it. The failure when it does arrive is silent and total: `counts` is a `Uint32Array(11)`, so a flagged byte like `0x80|4 = 132` is an out-of-bounds write that typed arrays **discard silently**, and `spanLists.get(132)` is `undefined` → `continue`. The entity vanishes from the span list with no error. And the trigger is someone editing Rust classification rules, who has no reason to be looking at a TypeScript span-list rebuild. Same for a new named class above 10, since `kinds` hard-codes 4–10.

Thanks for running the local CodeRabbit against the actual head — that's twice now the hosted one has been pinned behind (`2aeba7b4` here, `c29e3b9c` on #1967), and both times the real findings came from your local run.

BIMvoice:
Done in bba7008a — and your instinct to ask for it was right in a way the length check alone would not have covered.

**Both of your mutations fail now:**

```
(a) empty classes    -> left: 0,  right: 60   at :115 (length)
(b) drop first class -> left: 59, right: 60   at :115 (length)
```

**But a length assertion is not sufficient for the failure you actually described.** You said an off-by-one "mislabels every entity after the shift" — and that has a length-*preserving* form. I tried `classes.rotate_left(1)`:

```
panicked at issue_2053_shard_scan_parity.rs:124:9:
class byte at index 0 does not belong to the record at that index (#1 IFCPROJECT)
  left: 128
 right: 2
```

Note the line: **124, not 115.** The length assertion passed. So the test now also checks each byte against an oracle recomputed from that record's own keyword via `classify_type_name`. I re-ran that mutation myself on the pushed tree (`REPLACEMENTS=1`, line re-read) — both the whole-file and boundary cases fail on it.

**On your suggested stronger check — comparing the class for a record across the two range decompositions — I did something different, deliberately.** That comparison cannot see a misalignment: the shift lives inside the one function and lands identically in every decomposition. Concretely with `classes.remove(0)`, the whole-file map says record *k* → class of *k+1*, and a shard starting at *k* also gives its record *k* the class of *k+1*. They agree, and the test passes on broken code. The per-record oracle pins each byte to a specific record instead, so shift, wrap and substitution all fail.

The oracle is name-only, so the fixture builder now asserts none of its keywords (`IFCPROJECT`/`IFCWALL`/`IFCDOOR`) can take the #1910 instance-level exception — the one case where `classify_type_name` legitimately diverges from what `scan_shard_classified` actually calls. Add a spatial-container keyword to that fixture later and it fails loudly rather than silently disagreeing with correct code.

**On the module doc: I left it alone.** `prepass_sharded.rs:96` already scopes the claim to "the two loops' **records/handoff**" and never claimed classes were covered, so it was not overclaiming — it now slightly *undersells* the test. Widening it would mean editing a `#[wasm_bindgen]` doc comment and regenerating `.d.ts` for zero correctness gain, and that trap already cost this PR one CI cycle. `.d.ts` untouched, no `build-wasm.sh` run.

Parity 2/2, processing lib 57/57, wasm lib 27/27, clippy clean, ratchet 4/4 (test-only, allowlist unaffected). No changeset.

The host-side half is filed as #2065 — with a correction: it is **latent**, not live. `classify_type_name` returns the named codes early, so classes 4–10 never acquire a flag bit today, and the only later OR-in is gated on spatial containers. Which I think sharpens your point rather than blunting it: the raw comparison is correct by coincidence of the current classification rules, and when that changes the entity silently vanishes from the span list.

louistrue:
Verified `bba7008a`. You went past what I asked for, and the extra part is the part that earns its keep.

I suggested `classes.len() == classified_records.len()`. You added that *and* a per-index content assertion that re-derives the keyword from the source bytes and checks the class byte belongs to the record at that index. Two mutations, two different assertions firing:

**A — extra class byte (length diverges):**
```
class column length must match its record count for single-shard-whole-file (range [0, 3695))
  — the host indexes classes positionally against the records
```

**B — one *wrong* class byte, length still correct:**
```
class byte at index 1 does not belong to the record at that index (#2 IFCDOOR)
  for single-shard-whole-file — the class column is misaligned with the records
```

B is the case a length check alone cannot see, and it is the more likely real-world failure: a classifier change that mislabels one keyword keeps the column the right size and silently mislabels every entity of that type. The message naming `#2 IFCDOOR` means whoever hits it knows which record and which keyword without instrumenting anything.

That closes my finding on this PR. The module doc's claim — that a dedicated test catches drift between the two loops — is now true for records, handoff **and** classes, which is what I was arguing it needed to be.

Both tests pass unmutated (2 passed, 11.93s).

The second item from that CLI run stands separately and is not this PR's job: the host's span-list rebuild in `geometry-parallel.ts` indexes the class byte with raw literals 4-10 and never applies `PREPASS_CLASS_CODE_MASK`, so an added flag bit shifts every comparison silently. Worth its own issue — the Rust-side test cannot see across that boundary.

BIMvoice:
The host-side one is already filed — **#2065**, opened after your 05:39 comment. Flagging in case it didn't surface, since your note reads as though it still needs raising.

One correction in it that's worth having on the record here too: it is **latent today, not live**. I went looking for the current breakage and it isn't reachable yet. `classify_type_name` (`shard_classes.rs:120-141`) returns the named codes early —

```rust
if named != PREPASS_CLASS_NONE {
    // The named keywords are mutually exclusive with the flag predicates …
    return named;
}
```

— so classes 4–10 never acquire a flag from that path, and the one later OR-in (`:98-103`) is gated on `is_representationless_spatial_container_by_name`, which none of those keywords are. The host's unmasked comparison is therefore correct *by coincidence of the current classification rules*, not by construction.

Which sharpens your point rather than blunting it. When it does break, the failure is silent and total: `counts` is a `Uint32Array(11)`, so a flagged byte like `0x80|4 = 132` is an out-of-bounds write that typed arrays **discard silently**, and `spanLists.get(132)` is `undefined` → `continue`. The entity disappears from the span list with no error. Same for a new named class above 10, since `kinds` hard-codes 4–10. And the trigger is someone editing Rust classification rules, who has no reason to be reading a TypeScript span-list rebuild.

On the per-index assertion: your B case is exactly why I didn't stop at the length check. Worth noting I also declined the cross-decomposition comparison you'd floated — a shift lives inside the one function and lands identically in every decomposition, so whole-file vs per-shard agree on broken code. The per-record oracle was the only version that fails on it.

PR #2056, opened 2026-08-03
perf(export): speed up GLB assembly and bound large-model memory

louistrue:
## What and why

Optimizes the glTF/GLB exporter's assembly + serialization phase — flagged by a Fable code review as the dominant non-meshing cost — and bounds peak memory on large models. Output is **byte-for-byte identical** across every export path (in-memory instanced, quantized, unlit/emissive, bounded streaming, multi-buffer glTF, from-meshes); the keys/maps changed only gate dedup *equality* and never reach the bytes.

Also fixes one real correctness gap surfaced by the review: `export_glb_from_meshes` silently dropped meshes whose per-mesh counts ran past the flattened buffers (a valid GLB missing the model's tail, reported as success).

**Speed**
- `geom_color_key`: replace the double element-wise SipHash with one `xxh3_128` pass over `bytemuck::cast_slice` of positions/normals/indices (+ `u64` length frames + colour). ~20–50× less hashing; also makes the bounded path's release-mode determinism assert cheap.
- `push_mesh` / `push_mesh_quantized`: reserve the three byte runs once and write normals/indices via a single `cast_slice` memcpy (LE targets), positions via a per-vertex stack buffer, instead of per-element `to_le_bytes` pushes.
- `FxHashMap` for the dedup/material maps (insertion-order emit, unchanged bytes).

**Large-model memory**
- `with_result_views` converts Y-up **in place** on the owned `ProcessingResult` instead of allocating a full second `Vec<YUpMesh>` copy of all geometry.
- `pack_glb` writes `pos`/`norm`/`idx` straight into the container, dropping the `embedded_bin` intermediate copy on the single-buffer path.
- Streaming/bounded passes reuse one `YUpScratch` instead of 3 fresh allocations per mesh per pass; the bounded plan interns `ifc_type` as `Arc<str>`.
- `VisibilityFilter` precomputes `HashSet`s, replacing per-mesh linear `Vec::contains` scans (which also ran twice on the bounded path).

**Correctness**
- Add `try_export_glb_from_meshes` + `ExportError::MalformedMeshInput`, and make the wasm `exportGlbFromMeshes` boundary **fail closed** (throws `MALFORMED_MESH_INPUT`) instead of shipping a truncated model. The infallible core is unchanged for existing Rust callers; the TS success type stays `Uint8Array`.

**Hardening** (from adversarial review): compile-time little-endian assertion (the `cast_slice` paths assume LE), a `debug_assert` pinning the single-flush invariant the `embedded_bin` removal relies on.

## How it was verified

- `cargo test -p ifc-lite-export` — **84 passed / 0 failed**, including every existing byte-identical/determinism test and 2 new `try_from_meshes_*` tests.
- **Byte-identical golden diff** vs a pre-change baseline across all 9 export configurations on a real 2.38 MB model (`ara3d/duplex.ifc`) — re-checked after each implementation round.
- `cargo clippy -p ifc-lite-export --lib --tests -- -D warnings` — clean.
- `cargo build -p ifc-lite-wasm --release --target wasm32-unknown-unknown` — compiles (xxhash-rust builds for wasm; new `Result` boundary intact).
- 4 independent adversarial reviews (hash/ordering, byte-encoding/in-place, pack/flush, fail-closed edge cases) found **no output-affecting bug**; their hardening notes are folded in.

- [x] `cargo test --workspace` (Rust) and/or `pnpm test` (TS) pass locally — ran `cargo test -p ifc-lite-export` (84/84); full workspace via CI
- [ ] Geometry/WASM change: ran `scripts/build-wasm.sh` then `pnpm test:wasm-contract` — *in progress (browser/wasm validation)*

## Checklist

- [x] A test asserts the new behavior through a fixture or a stated invariant (byte-identical cross-path tests + new malformed-input tests)
- [x] Geometry-output change: re-pinned **both** determinism jsons **(N/A — output is byte-identical, no geometry output changed)**
- [x] House rules: no `as any` / `@ts-ignore`, no silent `catch {}`, no repo-wide `cargo fmt`

---
_Generated by [Claude Code](https://claude.ai/code/session_01GuGVRxintkLHg9PpoRmgZN)_

## Summary by CodeRabbit

* **New Features**
  * Added GLB export directly from flattened mesh data, with optional quantization.
  * Added support for shared geometry, instancing, and improved coordinate conversion.

* **Performance**
  * Improved export speed and memory efficiency for large and streaming models.
  * Reduced unnecessary allocations during geometry conversion and GLB generation.

* **Bug Fixes**
  * Malformed mesh buffers now fail safely with a `MALFORMED_MESH_INPUT` error.
  * Improved consistency for geometry transforms and exported model data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1508,6 +1508,7 @@ name = "ifc-lite-export"
 version = "4.3.0"
 dependencies = [
  "arrow",
+ "bytemuck",
  "ifc-lite-core",
  "ifc-lite-geometry",
  "ifc-lite-processing",
@@ -1515,6 +1516,7 @@ dependencies = [
  "rustc-hash",
  "serde",
  "serde_json",
+ "xxhash-rust",
  "zip",
 ]
 
--- a/packages/wasm/pkg/ifc-lite.d.ts
+++ b/packages/wasm/pkg/ifc-lite.d.ts
@@ -276,6 +276,13 @@ export class IfcAPI {
      * taken in order from the concatenated `positions`/`normals`/`indices`; `colors` is
      * RGBA per mesh, `origins` xyz per mesh, `express_ids` labels each mesh (indices are
      * per-mesh local). The caller passes exactly the meshes it wants emitted.
+     *
+     * Fails CLOSED: if the declared vertex/index counts run past the flattened
+     * `positions` / `indices`, there are fewer `index_counts` than meshes, or `normals`
+     * is empty or too short to cover every vertex, this throws an `Error` whose message
+     * starts with `MALFORMED_MESH_INPUT` — instead of silently emitting a GLB with those
+     * meshes dropped. (The viewer always passes fully-backed, normal-covered arrays, so
+     * this only fires on a caller bug.)
      */
     exportGlbFromMeshes(positions: Float32Array, normals: Float32Array, indices: Uint32Array, vertex_counts: Uint32Array, index_counts: Uint32Array, colors: Float32Array, origins: Float64Array, express_ids: Uint32Array, include_metadata: boolean, lit?: boolean | null, emissive?: boolean | null): Uint8Array;
     /**
--- a/rust/export/Cargo.toml
+++ b/rust/export/Cargo.toml
@@ -13,6 +13,10 @@ ifc-lite-core = { workspace = true }
 ifc-lite-geometry = { workspace = true }
 ifc-lite-processing = { workspace = true }
 rustc-hash = "2.1"
+# Bulk little-endian reinterpret of f32/u32 slices to bytes (glTF vertex encoding)
+# and a single-pass 128-bit content hash for mesh dedup; both pure-Rust + wasm-safe.
+bytemuck = "1.25"
+xxhash-rust = { version = "0.8", features = ["xxh3"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
--- a/rust/export/src/error.rs
+++ b/rust/export/src/error.rs
@@ -39,6 +39,16 @@ pub enum ExportError {
         /// The underlying error's `Display` output.
         detail: String,
     },
+    /// The from-meshes assembler was handed inconsistent inputs: the per-mesh
+    /// `vertex_counts` / `index_counts` sum past the end of the flattened
+    /// `positions` / `normals` / `indices` buffers. The infallible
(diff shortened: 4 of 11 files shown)


louistrue:
Thanks for the reviews — addressed in `5a1bce5`. Summary of what changed and where I diverged:

**Fail-closed validation (Greptile P1 / CodeRabbit)** — completed. `try_export_glb_from_meshes` now also rejects:
- empty **or** short `normals` (`normals.len() < vsum*3`) — empty normals would drop *every* mesh into a valid-but-empty GLB;
- fewer `index_counts` than declared meshes (a missing entry reads as 0 indices → that mesh vanishes).
…in addition to the existing vertex/index-count buffer overruns.

**Divergence — the empty-`normals` doc suggestion is now stale.** CodeRabbit suggested documenting "an empty `normals` buffer is valid and omits normals." That was true of the previous commit, but the P1 fix above makes empty normals a **rejected** input (it silently dropped the whole model), so I did *not* apply that wording — the binding doc now states empty/short normals throw `MALFORMED_MESH_INPUT`.

**Tests** — added `try_from_meshes_rejects_short_normals` (short + empty) and `try_from_meshes_rejects_index_counts_past_buffer`, so every validation branch is covered.

**Module ratchet (Codex P1 / Rust tests failure)** — split `gltf.rs` into child modules `gltf/matrix.rs` (instancing matrix math + its RTC test) and `gltf/from_meshes.rs` (the from-meshes assembler) rather than raising the budget. Pure code moves — GLB output is byte-identical (verified against a pre-change golden across all export paths) and the file is back under budget.

**wasm types** — regenerated the committed `packages/wasm/pkg/ifc-lite.d.ts` (the `exportGlbFromMeshes` TS signature is unchanged, `Uint8Array`).

**Skipped with reason:** the requested wasm-boundary test for `exportGlbFromMeshes` — the binding is a thin `try_export_glb_from_meshes(...).map(..).map_err(e.to_string())` forward, and `ifc-lite-wasm`'s tests run under `wasm-pack test` (excluded from the standard `cargo test --workspace` suite). The behavior is fully covered at the Rust layer by the five `try_from_meshes_*` tests, and the built wasm passes `pnpm test:wasm-contract` (43/0) locally. Happy to add a `wasm-bindgen-test` if you'd prefer it in-tree.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2057, opened 2026-08-03
chore: version packages

louistrue:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases

> The changelog information of each package has been omitted from this message, as the content exceeds the size limit.


BIMvoice:
## Release-notes verification for #2057

I checked every changeset entry in this release against `upstream/main` (file:line), spot-checking the highest-risk claims first (overclaiming, stale text, missing entries, wrong bump type). Summary: **accurate**. I did not find an overclaim, stale-text defect, or missing changeset in this batch. I'm flagging four entries where the changeset **itself already discloses** a behaviour change but ships on a `patch` bump — worth an explicit maintainer call, not a rule violation, since `AGENTS.md`'s changeset section ("Bump level = biggest API change... never patch when the surface shrank") only ties bump escalation to *export-surface* shrinkage, not runtime-contract changes. None of these remove/rename an export, so they're compliant with the letter of that rule; they're still the kind of change that can break an embedder that relied on the old behaviour.

### Verified accurate (spot-checked against code)

- `@ifc-lite/export` #2059/#2105/#2047/#2111 — georef effective-index, `stats.warnings`, `Ifc5Exporter`/`ParquetExporter` overlay-deletion gating. Confirmed against `packages/export/src/step-exporter.ts` (`warnings: string[]`, `MAP_CONVERSION_WITHOUT_CONTEXT_WARNING`/`_NO_CRS_WARNING`, `reportMapConversionRefused*`) and `ifc5-exporter.ts`'s single `isOmitted` predicate (L237-265, 446-554) and `parquet-exporter.ts`'s `mutationView`-optional constructor + per-table `effective.isDeleted()` filtering (L26-62,125-199,406). The #2105 "which sessions this affects" paragraph correctly scopes to the case the PR added (cross-references #2048), so the earlier over-scoped version flagged elsewhere is not present in this text.
- `@ifc-lite/mutations` #2109/#2045 — two-pass skip-set (`applyMutations`, `mutable-property-view.ts` L1734-1785) matches the "skip set built before applying anything" / "keyed on `newEntities`, not id-absence" description exactly.
- `@ifc-lite/sandbox` #2062/#2081/#2063/#2096/#2118/#2103/#2078/#2095/#2080 — all nine checked against `sandbox.ts` and `bridge.ts`. Notably: the log-budget-ceiling check (`bridge.ts` ~L159, `if (totalBytes + entryBytes > MAX_TOTAL_BYTES)`) runs pre-retention as described; `resetLogs()` zeroes both `totalBytes` and `truncated`; `ScriptError` copies `logs` at construction (`sandbox.ts` L360, `this.logs = [...logs]`); the rejected-promise path matches `sandbox.ts` L228-281 exactly, including the "interrupt checked first" ordering.
- `@ifc-lite/geometry` #2072 — `PREPASS_CLASS_CODE_MASK = 0x3F`, flag bits `0x80`/`0x40` confirmed in `rust/processing/src/shard_classes.rs` L44-54; `extractPrepassSpanLists()` exists in `packages/geometry/src/geometry-parallel.ts`.
- `@ifc-lite/wasm` #2068 — `process_mapped_item_cached` / recursion claim consistent with `rust/geometry/src/router/instancing.rs` and `processing.rs`.
- `@ifc-lite/cli`/`data`/`cache`/`create`/`geometry` #2100 — `0.0.0-unknown` fallback confirmed in `packages/cli/src/version.ts` L14-23.
- `@ifc-lite/server-client`/`extensions`/`collab-server`/`sdk` #2083 — text is shared verbatim and correctly attributed per-package.
- `@ifc-lite/pointcloud` #2113/#2086 — `memoizeAsync`/`DuckDBIntegration.init()` cross-reference confirmed (`packages/query/src/duckdb-integration.ts` L43-104, same clear-on-reject pattern).
- `@ifc-lite/geometry`/`cli`/`mcp` #2052 (usd-ascii-export, minor) — additive API, correctly minor.

### Judgment call: behaviour-visible changes on `patch`

Four entries in this release self-disclose a runtime behaviour change (bolded "Behaviour change" / "Embedder-visible") but ship as `patch`:

1. **`@ifc-lite/sandbox` #2118** (log-budget-ceiling) — a script logging one oversized payload now truncates on that call instead of the next.
2. **`@ifc-lite/sandbox` #2103** (log-budget-per-eval) — a long-lived sandbox can now retain up to 4 MB of logs *per eval* instead of 4 MB total across its lifetime — a real increase in what an embedder holds.
3. **`@ifc-lite/sandbox` #2078** (rejected-promise) — `eval()` that previously *resolved* (with the rejection folded into `value`) now throws. Code that caught nothing before will start seeing an exception propagate.
4. **`@ifc-lite/server-client` #2083** (parse-stream-truncation) — `parseStream()` that previously ended the generator cleanly on a truncated stream now throws `Stream ended without a complete event`.

None of these remove or rename an export, so none violate `AGENTS.md`'s literal rule. I could not find prior precedent in this repo's CHANGELOG history for a "Behaviour change"-labelled patch (the only prior precedent I found for a real behaviour break is `@ifc-lite/sandbox` #1979 and `@ifc-lite/export`'s GLTF/JSONLD/CSV exporter removal, both under `### Major Changes` with a bolded `BREAKING:` prefix). Given that pattern, a `minor` bump (or at least a `BREAKING`-style callout) may be more consistent with how this repo has previously signalled contract changes, even though none of these four remove API surface. Flagging for your call — the wording of all four entries is accurate, this is purely a bump-level question.

### Not applicable to this PR

`@ifc-lite/drawing-2d`'s padding-default change and `@ifc-lite/parser` changes mentioned as things to scrutinise are **not part of #2057** — there is no `.changeset/*.md` for either package on `upstream/main` yet, so they haven't queued into this release. Nothing to flag here; will need separate changesets before they show up in a future version-packages PR.

No missing changesets found: every package version bump in this PR traces either to a changeset file consumed in `.changeset/` or to changesets' own dependency-bump propagation (`Updated dependencies [...]`), consistently.

BIMvoice:
The outstanding P1 thread on `packages/geometry/package.json:3` (WASM / Rust export surface not versioned) looks **resolved by the current head** — flagging it so it can be closed rather than left blocking a release PR. Verified rather than assumed:

| | `main` | this PR's head |
|---|---|---|
| `@ifc-lite/wasm` | 4.3.0 | **4.3.1** |
| root `Cargo.toml` `[workspace.package]` | 4.3.0 | **4.3.1** |
| `@ifc-lite/geometry` | 3.6.0 | 3.7.0 |

Both halves of the finding are covered:

1. **npm side** — `@ifc-lite/wasm` now advances to 4.3.1. `@ifc-lite/geometry` depends on it via `workspace:^`, so the published range resolves to `^4.3.1` and picks up the runtime carrying `exportUsd` (`packages/wasm/pkg/ifc-lite.d.ts:389`) rather than the already-published 4.3.0.
2. **crates side** — `scripts/release-crates.mjs:48` reads the version straight out of `[workspace.package]`, and its documented rule is "version already on crates.io → skip". At 4.3.0 that skip is exactly what would have left `ifc_lite_export::export_usd` unpublished; at 4.3.1 it publishes.

One observation, not a blocker: `@ifc-lite/wasm` goes 4.3.0 → 4.3.1 as a **patch**, but relative to the published 4.3.0 tarball it gains a new public method. Strict semver would make that a minor. Nothing breaks — `workspace:^` resolves to `^4.3.1` either way — so this only matters if anyone downstream pins a narrower range.

No action requested; just closing the loop with evidence so the stale thread doesn't hold up the release.

PR #2059, opened 2026-08-04
fix(export): resolve georeferencing through the effective index (#2048)

BIMvoice:
Closes the first finding of #2048. The second is already fixed on main — see below.

## The one-sided guard

`step-exporter.ts:508-509` read `dataStore.entityIndex.byType` directly, so a deleted `IfcProjectedCRS`/`IfcMapConversion` was still found as "existing". The exporter then took the **modify** branch against a tombstoned record, which is never emitted — georeferencing silently vanished with no error.

RED, before the fix:

```
× recreates georeferencing the session deleted instead of editing the tombstone
  expected '...' to contain "IFCPROJECTEDCRS('EPSG:1234'"
  DATA;
  #1=IFCPROJECT(...); #2=IFCSIUNIT(...); #20=IFCGEOMETRICREPRESENTATIONCONTEXT(...);
  ENDSEC;
```

Now reads `effective.byType` from the existing `getEffectiveEntityIndex()` seam rather than a second deletion predicate — a differently-shaped guard here is how these drift apart again. `effective.byType` also *adds* overlay-created records, and the new-entities pass already applies `modifiedAttributes` (`:954`), so the modify branch stays correct for those too.

## The louder bug the fix exposed

Making the create branch reachable for deleted-CRS sessions surfaced a second raw-index read **in the same twenty lines**. `findPreferredGeometricRepresentationContextId()` (`:1557`) and `findLengthUnitReference()` (`:1512-1525`) also read the raw index, and the create branch uses their return values as `#id` references. Probe on unfixed code:

```
#31=IFCPROJECTEDCRS('EPSG:1234',$,$,$,$,$,#2);
#32=IFCMAPCONVERSION(#20,#31,10.,20.,30.,1.,0.,1.);
DANGLING: [ 2, 20 ]     // both deleted
```

That is an **invalid file**, and it reproduces on main independently of this change — a file with no CRS at all takes the create branch too. Fixed here because my change makes it newly reachable, with `effective` threaded through `resolveMapUnitReference` / `findLengthUnitReference` / `findUnitId` / `generatePropertySetEntities` so no half-guarded helper is left behind.

## Finding 2: already fixed, untouched

The IFCX-ops-for-a-tombstoned-entity half is closed on main by #2050 — `change-set-to-ops.ts:167-168`. No action taken.

## Evidence

| guard neutralised | replacements | result |
|---|---|---|
| `effective.byType` → raw, both georef lines | 2 | 2 failed / 335 passed — both new tests |
| context lookup → raw `byType` | 1 | 1 failed / 30 passed |
| `\|\| effective.isDeleted(unitId)` removed | 1 | 1 failed / 30 passed |

337 passed / 29 skipped, typecheck 34/34.

## Same-shape sweep — reported, not fixed

- `packages/parser/src/georef-extractor.ts:101` reads `entitiesByType.get('IfcProjectedCRS')` from the parsed store. Same shape, but the parser layer has no overlay threaded, so a deleted CRS still yields georeferencing to the viewer. Needs an effective index plumbed in — a design call rather than a guard.
- `apps/viewer/src/components/mcp/playground-dispatcher.ts:662` counts `IFCMAPCONVERSION` off the raw store.
- Broader and distinct from this shape: surviving source lines are never rewritten when they reference a deleted entity — `#30=IFCUNITASSIGNMENT((#2))` keeps `#2` after `deleteEntity(2)`.

Benign on inspection: `:1084` (`IFCOWNERHISTORY`, filtered by `willBeEmitted`), `:1677` (only ever *adds* skips), `:1797` (already cleared in the issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved georeferenced STEP exports by preventing references to deleted coordinate systems, map conversions, contexts, units, and related entities.
  * Recreates missing georeferencing records with valid identifiers when needed.
  * Ensures exported files contain no dangling entity references.
  * Preserves valid surviving georeferencing entities and synthesizes replacement units when necessary.

### DIFF
--- a/.changeset/georef-effective-index.md
+++ b/.changeset/georef-effective-index.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/export": patch
+---
+
+STEP export: keep georeferencing edits when the session deleted the file's existing georeferencing.
+
+The exporter looked up `IfcProjectedCRS` / `IfcMapConversion` in the raw entity index, so a CRS the session had deleted still counted as "existing". The edit was queued against the deleted entity, the export skipped that entity, and the replacement georeferencing vanished from the output with no error. The lookup now goes through the effective (overlay-aware) index, so a deleted CRS or map conversion is recreated instead.
+
+The same index now backs the source-CRS context and length-unit lookups the georef path uses, so newly created georeferencing can no longer reference a deleted context or unit.
--- a/packages/export/src/step-exporter.test.ts
+++ b/packages/export/src/step-exporter.test.ts
@@ -72,6 +72,22 @@ function buildMockDataStore(
   } as unknown as IfcDataStore;
 }
 
+/** A minimal already-georeferenced model: `#40` CRS, `#41` MapConversion. */
+function buildGeoreferencedMockDataStore(): IfcDataStore {
+  return buildMockDataStore([
+    [1, 'IFCPROJECT', "#1=IFCPROJECT('g',$,'Project',$,$,$,$,(#20),#30);"],
+    [2, 'IFCSIUNIT', '#2=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);'],
+    [20, 'IFCGEOMETRICREPRESENTATIONCONTEXT', "#20=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,#21,$);"],
+    [21, 'IFCAXIS2PLACEMENT3D', '#21=IFCAXIS2PLACEMENT3D(#22,#23,#24);'],
+    [22, 'IFCCARTESIANPOINT', '#22=IFCCARTESIANPOINT((0.,0.,0.));'],
+    [23, 'IFCDIRECTION', '#23=IFCDIRECTION((0.,0.,1.));'],
+    [24, 'IFCDIRECTION', '#24=IFCDIRECTION((1.,0.,0.));'],
+    [30, 'IFCUNITASSIGNMENT', '#30=IFCUNITASSIGNMENT((#2));'],
+    [40, 'IFCPROJECTEDCRS', "#40=IFCPROJECTEDCRS('EPSG:2056',$,'CH1903+',$,$,$,#2);"],
+    [41, 'IFCMAPCONVERSION', '#41=IFCMAPCONVERSION(#20,#40,1.,2.,3.,1.,0.,1.);'],
+  ]);
+}
+
 /** Count `#N` references in the output that have no `#N=` definition. */
 function findDanglingRefs(content: string): number[] {
   const defined = new Set<number>();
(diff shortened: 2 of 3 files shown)


louistrue:
Verified `a0ab7e83`. The fix itself is solid — reverting `effective.byType` to `dataStore.entityIndex.byType` (7 call sites) fails all three of your new tests and nothing else:

```
× recreates georeferencing the session deleted instead of editing the tombstone
× recreates a deleted IfcMapConversion while keeping the surviving IfcProjectedCRS
× never points new georeferencing at a deleted context or length unit
```

On the open MAJOR, I got a partial result and want to be precise about which half I actually established.

**Confirmed:** with the only source context deleted and no replacement, the requested map conversion is silently dropped while the CRS is still written:

```
#31=IFCPROJECTEDCRS('EPSG:1234',$,$,$,$,$,#2);
   (no IFCMAPCONVERSION line, no surviving IFCGEOMETRICREPRESENTATIONCONTEXT)
```

The caller asked for a map conversion and got a lone `IfcProjectedCRS` with no indication anything was refused. Not emitting a dangling SourceCRS is right; doing it silently is the part worth a second look, and it does not need an overlay-created entity to reach — deleting every context is enough.

**Not established:** the overlay-created-replacement half. I tried to probe it, and my mock store's `createEntity` allocated expressId `1`, colliding with the existing `IfcProject` at `#1`, so the run told me nothing trustworthy. I am reporting that rather than the reading it produced, because a probe with a colliding id can just as easily manufacture a pass as a failure. If you test it, use a store whose allocator starts above the fixture's max id.

So I would treat the finding as: the reachable, no-overlay part is real and narrow; the overlay part is unverified by me either way. Neither is a regression from this PR — both predate it — which is why I would be comfortable landing this and tracking the silent-drop separately rather than growing a georef fix into an overlay-resolution refactor.

`@ifc-lite/export` 334 passing.

louistrue:
On the remaining thread (overlay-created records discarded in the context and unit lookups): **you have already answered the harder half of this, on #2105**, and I verified the reasoning rather than taking it.

Your #2105 changeset says:

> Not changed here: an overlay-created replacement context is still not used as `SourceCRS` ... Selecting one would need its own decision, because `createEntity` does not require the mandatory `WorldCoordinateSystem` placement, so an overlay-created context can be a schema-invalid target.

That is accurate. `MutablePropertyView.createEntity` validates exactly one thing:

```ts
createEntity(type: string, attributes: IfcAttributeValue[]): NewEntity {
  if (!type || typeof type !== 'string') {
    throw new Error('createEntity: type is required');
  }
  ...
  attributes: attributes.slice(),
```

`attributes` is copied verbatim with no schema check, so an overlay-created `IfcGeometricRepresentationContext` really can be missing `WorldCoordinateSystem`. Wiring the lookup to accept it would then let the exporter emit an `IfcMapConversion` whose `SourceCRS` points at a schema-invalid entity — a *worse* output than the current omission, which at least refuses cleanly and (after #2105) says so in `stats.warnings`.

The same argument covers the unit half: an overlay-created `IfcSIUnit` has had no more validation than the context, so reusing it could serialize an incomplete unit rather than the duplicate-but-valid one you get today.

So this is not a lookup bug to patch — it is a question of what validation the exporter should demand of overlay-created entities before trusting them as reference targets. CodeRabbit labelled it `🏗️ Heavy lift` and that is the right read. It deserves its own issue with that framing, not a fold-in here.

I would resolve this thread on those grounds. Two things before this can merge, though:

**The Vercel statuses are red with `Authorization required to deploy`** — this branch is still fork-based. Unlike some of the others, CI itself did run here (16 check-runs, all 5 core lanes green), so the only cost is the preview deployment. Still worth migrating to an in-repo branch as you did for #2083 and #2096, since it is the last thing keeping this one visibly red.

Say the word on the overlay-validation issue and I will open it.

BIMvoice:
Agreed on all three points, with one constraint on the migration.

## The overlay-validation issue — yes please, open it

Your reading matches mine and you verified the half I could not: `createEntity` validates only that `type` is a non-empty string and copies `attributes` verbatim, so an overlay-created `IfcGeometricRepresentationContext` really can be missing its mandatory `WorldCoordinateSystem`.

The framing I would want on it is the one you landed on: **this is not a lookup bug, it is a question of what validation the exporter should demand of an overlay-created entity before trusting it as a reference target.** Wiring the lookup to accept one today trades a clean refusal — which, after #2105, is now reported in `stats.warnings` — for a schema-invalid `SourceCRS` reference. That is strictly worse output, and silently so, since the file would look complete.

Worth capturing that the unit half is the same shape and not a separate problem: an overlay-created `IfcSIUnit` has had no more validation than the context, so reusing it could serialize an incomplete unit in place of today's duplicate-but-valid one. Same root cause, same decision needed.

Please open it — you have the analysis and the `🏗️ Heavy lift` read from CodeRabbit, and I would rather it carry your framing than my paraphrase.

## Resolving the thread

Agreed, on those grounds.

## The migration — I do not think I can do it without losing this thread

GitHub fixes a PR's head **repository** at creation; it can only be re-pointed to a different branch within the same repo. So there is no way to move #2059 from `BIMvoice` to `LTplus-AG` in place — it would mean opening a fresh PR from an in-repo branch and closing this one, which discards the review history here. That history is not trivial: your partial-result note on the open MAJOR, the reasoning above, and the mutation evidence all live in this thread.

#2083 and #2096 were not migrations — they were created in-repo from the start, after your workflow request landed.

So the trade is: a cosmetic red (three `Authorization required to deploy` rows, preview deployment only) against losing the thread. CI itself ran here — 16 check-runs, all 5 core lanes green — so nothing about the verification is affected.

My inclination is to leave it and merge as-is, but it is your call and I will open the replacement PR if you would rather have the clean rollup. Everything from here on is already on in-repo branches.

louistrue:
Merging. The one open thread is dispositioned above on evidence: resolving overlay-created records as `SourceCRS` / unit targets would need `createEntity` to validate mandatory attributes first, which it does not — it checks only that `type` is a non-empty string and copies `attributes` verbatim. Wiring the lookup to accept them would let the exporter reference a schema-invalid entity, which is a worse output than the current clean refusal. CodeRabbit's own `🏗️ Heavy lift` label is the right read; it deserves its own issue, not a fold-in here.

All 5 core CI lanes green. The three red Vercel statuses are `Authorization required to deploy` — fork-branch, so preview deployments only; CI itself ran in full. Moving the branch in-repo as you did for #2083 and #2096 clears them, but it is not a correctness signal.

Resolving that thread as part of merging.

Issue #1922, opened 2026-07-30
QuickJS sandbox: OOM inside a drained job aborts the WASM module at dispose() (upstream quickjs-emscripten)


louistrue:
Split out of #1905 so the reproducer survives that issue being closed. PR #1918 fixed one root cause of the `JS_FreeRuntime` assertion (orphaned handles); **this is the other one, and it is still live.**

## Repro

On `main` with #1918 applied, through the real `createSandbox` at production defaults (64 MB / 30 s — what `useSandbox.ts` configures):

```js
async function run() { await 0; const a = []; for (;;) { a.push({ k: "v" }); } } run(); "started"
```

`eval()` **resolves** with `"started"`, then `sandbox.dispose()` throws:

> Aborted(Assertion failed: list_empty(&rt->gc_obj_list), at: ../../vendor/quickjs/quickjs.c,2036, JS_FreeRuntime)

OOM lands in ~1 s, well inside the timeout. Once it fires the emscripten module is dead for the rest of the document, so the extension sandbox is unusable until reload.

This is reachable in normal use: `sandbox.ts`'s own comment notes entry points are commonly wrapped as `async function run()`, whose post-await body runs as a job.

## It is upstream, not ours

Minimal repro in a fresh process with **zero ifc-lite code** (only `getQuickJS` + `evalCode` + `executePendingJobs`, every handle disposed) aborts at every memory limit tested, 256 KB through 64 MB. Bisected:

| scenario | result |
|---|---|
| OOM inside a drained job | **ABORT** |
| OOM in the main eval body | clean |
| job that merely throws | clean |
| same script without draining | clean |

Aborts whether the rejection is unhandled, `.catch()`ed, or `try`/`catch`ed inside the job.

## Why no fix shipped

quickjs-emscripten 0.32.0 exposes no GC trigger — the full runtime surface is constructor, alive, dispose, newContext, setModuleLoader, removeModuleLoader, hasPendingJob, setInterruptHandler, removeInterruptHandler, executePendingJobs, setMemoryLimit, computeMemoryUsage, dumpMemoryUsage, setMaxStackSize, assertOwned, setDebugMode, isDebugMode, debugLog, getSystemContext.

The only in-repo lever would be to skip `runtime.dispose()` on a heuristic OOM signal, trading a document-wide module death for an unbounded WASM runtime leak. That is a worse failure mode and was deliberately not shipped.

## Next step

File upstream against quickjs-emscripten with the minimal repro above.

[...]

PR #2062, opened 2026-08-04
fix(sandbox): stop dispose() double-freeing a runtime that aborted (#1922)

BIMvoice:
Does **not** fix the upstream abort in #1922 — that cannot be fixed from our side, and the bisect is below. It fixes a defect the investigation turned up that *is* ours, and corrects three claims in the issue.

## The abort is genuinely upstream — confirmed, not assumed

Reproduced on `upstream/main` through the real `createSandbox` at production defaults, then again with **zero ifc-lite code** (`getQuickJS` + `evalCode` + `executePendingJobs`, every handle disposed):

```
oom-job: eval RESOLVED with "started" in 722ms
oom-job: calling sandbox.dispose()...
Aborted(Assertion failed: list_empty(&rt->gc_obj_list), at: ../../vendor/quickjs/quickjs.c,2036,JS_FreeRuntime)
```

`sandbox.ts:246` is only where it surfaces; `sandbox.ts:156` (`executePendingJobs()`) is what makes it reachable. Neither is wrong. `quickjs-emscripten` 0.32.0 is the latest published version — no bump available.

## The bug that IS ours

`dispose()` set each field to null *after* freeing it, so a throwing `runtime.dispose()` left the field populated:

```
dispose #1: THREW RuntimeError: Aborted(Assertion failed: list_empty(...))
dispose #2: THREW      <- re-enters JS_FreeRuntime on a half-freed runtime
dispose #3: THREW
```

`JS_FreeRuntime` has already run most of its free sequence before hitting the assert, so every later call is a **double-free**. Reachable today: `useSandbox.ts` disposes in a `finally`, in `reset()`, and in an unmount effect — and because the `finally` threw, `activeSandboxRef.current` was never cleared, guaranteeing the second call.

Fields are now cleared before the step that frees them, and the runtime is freed from a `finally` so a throwing `vm.dispose()` no longer strands it for the page lifetime (a pre-existing leak on that path). **The throw still propagates, deliberately** — `dispose-leak.test.ts` uses "dispose() does not throw" as its leak oracle, and swallowing here would have silently made all eight of those tests vacuous.

## Three corrections to the issue

**(a) It is not OOM-specific — our own CPU timeout does it too.** The issue's table says interrupts are clean; that holds only for a *non-allocating* job:

```
main      trip=28  eval=ERR:interrupted CLEAN     # engine exception in the main body: always clean
asyncjob  trip=28  ABORT
thenjob   trip=28  ABORT
```

Through the real sandbox with `timeoutMs: 200`, `eval` **resolves with `"started"`** and then `dispose()` aborts. So the documented 30 s timeout path is itself a module-killer for any allocating async script, and it silently reports success.

**(b) A 1-line repro that defeats every sampling-based mitigation.** `async function run(){ await 0; new Array(1e9).fill(0); } run()` aborts in **90 ms with 1 interrupt**, no memory ramp. (`new Uint8Array(200MB)` and `"x".repeat(200MB)` in a job are clean — only the object-array path leaks.) Nothing to sample, and interrupting is itself a poison.

**(c) "The module is dead for the rest of the document" did not reproduce** in Node with the release-sync variant: exactly **one** abort per process, every subsequent sandbox creates/evals/disposes cleanly, WASM heap plateaus at 357.5 MB over 12 rounds. `ABORT` is only consulted in the run/exit paths, never in the FFI calls. **This is what made a mitigation worth shipping rather than cosmetic.** Worth re-checking in a browser before amending the issue text.

## Why there is no conditional mitigation

Every detector tried is dead:

- `executePendingJobs()` reports **no error** — the OOM rejection is swallowed
- a sentinel job queued after the poisoned one **still runs**
- `obj_count` after `ctx.dispose()` — massive false positives: 50 333 objects disposes clean, and a main-body OOM leaving 3 543 636 also disposes clean. Only ~166 refcount-leaked objects trip the assert
- `ffi.QTS_RecoverableLeakCheck()` returns `0` in both clean and poisoned states
- forcing GC does not reclaim them

Skipping `runtime.dispose()` unconditionally works but costs **~17 KB of WASM heap per sandbox, never returned** (measured: 200 runtimes = 16.0 → 19.3 MB) — the trade the issue rightly rejects. Sharing one runtime is also out: `extensions/src/host/runtime.ts` keeps several sandboxes alive concurrently and limits are per-runtime, so it would break per-extension isolation.

## Evidence

RED (ordering reverted, `replacements: 1`, region re-read):

```
FAIL src/dispose-abort.test.ts > does not re-enter runtime teardown after a failed dispose
AssertionError: expected [Function] to not throw an error but 'RuntimeError: Aborted(Assertion faile…' was thrown
```

The test tolerates the *first* `dispose()` throwing (upstream's behaviour) and asserts only that the second and third do not — so it stays meaningful if upstream ever fixes the abort, and cannot pass against the broken code. It lives in its own file because the abort poisons the WASM module for the rest of the vitest worker, same reason `dispose-leak.test.ts` is separate.

sandbox 38 passing (was 37), extensions 598, sdk 62, viewer 2563, typecheck 34/34.

**Known gap, flagged rather than hidden:** the viewer guard has no test — `useSandbox` has no existing harness and standing one up for a 3-line try/catch seemed out of proportion. Say the word if you want it.

## Follow-ups

An upstream report is drafted (minimal repro with no wrapper code, the full clean/abort bisect, reproduces at every memory limit from 256 KB to 64 MB) — happy to file it against `justjake/quickjs-emscripten` on your say-so.

Separately worth its own ifc-lite issue, found here and deliberately not fixed: when the CPU interrupt kills a drained job, `eval()` returns the pre-job value and reports **success** — the user gets a silently wrong answer with no timeout error. `sandbox.ts` never checks whether its own interrupt handler fired.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sandbox cleanup after runtime errors, including out-of-memory failures.
  - Prevented repeated disposal attempts from causing additional errors.
  - Ensured resources are released even when part of teardown fails.
  - Prevented cleanup errors from escaping during viewer execution, reset, and unmount operations.
  - Improved stability when recovering from failed or interrupted sandbox operations.
  - Preserved relevant teardown errors for reliable troubleshooting while completing available cleanup.

### DIFF
--- a/.changeset/sandbox-dispose-no-double-free.md
+++ b/.changeset/sandbox-dispose-no-double-free.md
@@ -6,4 +6,4 @@ Fix a double-free in `Sandbox.dispose()` when QuickJS teardown fails.
 
 An out-of-memory or CPU-timeout exception raised inside a drained promise job — an `async function run()` entry point that allocates, for example — leaves QuickJS holding objects with leaked refcounts. Upstream `JS_FreeRuntime` then trips `assert(list_empty(&rt->gc_obj_list))` and throws out of `runtime.dispose()` part-way through freeing the runtime (that abort is upstream in quickjs-emscripten and is not fixed here). `dispose()` left its `runtime` field set afterwards, so every later call — a React cleanup, an extension unload, a defensive re-dispose — re-entered `JS_FreeRuntime` on the same half-freed runtime.
 
-`dispose()` now clears each field before freeing it, and frees the runtime from a `finally` so a throwing `vm.dispose()` no longer strands it for the lifetime of the page. The failure is still reported to the caller.
+`dispose()` now clears each field before freeing it, so a step that throws is never retried. The failure is still reported to the caller.
--- a/.changeset/sandbox-dispose-no-double-free.md
+++ b/.changeset/sandbox-dispose-no-double-free.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Fix a double-free in `Sandbox.dispose()` when QuickJS teardown fails.
+
+An out-of-memory or CPU-timeout exception raised inside a drained promise job — an `async function run()` entry point that allocates, for example — leaves QuickJS holding objects with leaked refcounts. Upstream `JS_FreeRuntime` then trips `assert(list_empty(&rt->gc_obj_list))` and throws out of `runtime.dispose()` part-way through freeing the runtime (that abort is upstream in quickjs-emscripten and is not fixed here). `dispose()` left its `runtime` field set afterwards, so every later call — a React cleanup, an extension unload, a defensive re-dispose — re-entered `JS_FreeRuntime` on the same half-freed runtime.
+
+`dispose()` now clears each field before freeing it, and frees the runtime from a `finally` so a throwing `vm.dispose()` no longer strands it for the lifetime of the page. The failure is still reported to the caller.
--- a/apps/viewer/src/hooks/useSandbox.ts
+++ b/apps/viewer/src/hooks/useSandbox.ts
@@ -21,6 +21,24 @@ import {
   type RuntimeScriptDiagnostic,
 } from '../lib/llm/script-diagnostics.js';
 
+/**
+ * Tear a sandbox down without letting a teardown failure escape.
+ *
+ * An out-of-memory or CPU-timeout exception raised inside a drained promise
+ * job leaves QuickJS holding objects with leaked refcounts, and upstream
+ * `JS_FreeRuntime` aborts on it (#1922). Every call site here runs in a
+ * `finally` or a React cleanup, where a throw would discard the run's own
+ * result or escape an unmount — so report it and carry on. `Sandbox.dispose()`
+ * has already released everything it owns by the time it throws.
+ */
+function disposeSandbox(sandbox: Sandbox): void {
+  try {
+    sandbox.dispose();
+  } catch (err) {
+    console.error('[ifc-lite] sandbox teardown failed', err);
+  }
+}
+
 /** Type guard for ScriptError shape (has logs + durationMs) */
 function isScriptError(err: unknown): err is { message: string; logs: Array<{ level: string; args: unknown[]; timestamp: number }>; durationMs: number } {
   return (
@@ -199,7 +217,7 @@ export function useSandbox(config?: SandboxConfig) {
     } finally {
       // Always dispose the sandbox after execution
       if (sandbox) {
-        sandbox.dispose();
+        disposeSandbox(sandbox);
       }
       if (activeSandboxRef.current === sandbox) {
         activeSandboxRef.current = null;
@@ -210,7 +228,7 @@ export function useSandbox(config?: SandboxConfig) {
   /** Reset clears any active sandbox (no-op if none running) */
   const reset = useCallback(() => {
     if (activeSandboxRef.current) {
-      activeSandboxRef.current.dispose();
+      disposeSandbox(activeSandboxRef.current);
       activeSandboxRef.current = null;
     }
     setExecutionState('idle');
@@ -223,7 +241,7 @@ export function useSandbox(config?: SandboxConfig) {
   useEffect(() => {
     return () => {
       if (activeSandboxRef.current) {
-        activeSandboxRef.current.dispose();
+        disposeSandbox(activeSandboxRef.current);
         activeSandboxRef.current = null;
       }
     };
--- a/packages/sandbox/src/dispose-abort.test.ts
+++ b/packages/sandbox/src/dispose-abort.test.ts
@@ -1,0 +1,50 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression test for the double-free reachable from the upstream
+ * `JS_FreeRuntime` abort (#1922).
+ *
+ * An out-of-memory (or CPU-interrupt) exception raised inside a drained
+ * promise job leaves QuickJS holding objects with leaked refcounts. Upstream
+ * `JS_FreeRuntime` then trips `assert(list_empty(&rt->gc_obj_list))` and
+ * throws out of `runtime.dispose()` part-way through freeing the runtime. That
+ * abort is upstream and we cannot prevent it — but `Sandbox.dispose()` must
+ * not leave the runtime handle in place afterwards, because every later
+ * `dispose()` (a React cleanup, an extension unload, a defensive re-dispose)
+ * would then free the same half-freed structures again.
+ *
+ * This file is kept separate so vitest gives it its own worker: the abort
+ * poisons the shared WASM module for every later test in the same process.
+ */
+
+import { describe, expect, it } from 'vitest';
+import type { BimContext } from '@ifc-lite/sdk';
+import { createSandbox } from './sandbox.js';
+
+/**
+ * Exhausts the 64 MB default budget from inside a job. `eval()` resolves with
+ * `"started"` — the rejection is invisible to the host — and the damage only
+ * shows up at teardown.
+ */
+const OOM_IN_JOB = 'async function run() { await 0; new Array(1e9).fill(0); } run(); "started"';
+
+describe('sandbox dispose after an OOM inside a drained job (#1922)', () => {
+  it('does not re-enter runtime teardown after a failed dispose', async () => {
+    const sandbox = await createSandbox({} as BimContext);
+    await sandbox.eval(OOM_IN_JOB, { typescript: false });
+
+    // The first dispose() surfaces whatever upstream does. Today that is the
+    // `JS_FreeRuntime` abort; if upstream ever fixes it, this is simply clean.
+    try {
+      sandbox.dispose();
+    } catch {
+      // Reported to the caller by design — see Sandbox.dispose().
+    }
+
+    // Whatever happened, the runtime handle is gone: teardown is not retried.
+    expect(() => sandbox.dispose()).not.toThrow();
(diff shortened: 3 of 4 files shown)


louistrue:
Holding this one, and the reason is unusual: **half the diff is necessary and well-proven, and the other half should not exist.**

## The half that's right

`null`-before-free is a real fix. The mechanism is provable at the dependency's source, not just observable: in `quickjs-emscripten-core@0.32.0`, `Lifetime.dispose()` clears `_alive` **after** the disposer runs, so a throwing disposer leaves the Lifetime alive, and `QuickJSRuntime.dispose()` is `this.scope.dispose()` which re-runs any still-alive lifetime. That is why an `if (rt.alive)` guard could not have worked. Reachable on the normal script-run path (`ScriptPanel.tsx:216`, bare `await execute(...)`), and pinned — reverting the ordering alone fails `does not re-enter runtime teardown after a failed dispose`.

## The half that should go

The `try`/`finally` chain is unnecessary, unpinned, and where it would fire it makes things worse.

**Unpinned.** I ran this myself: delete the chain, keep null-before-free, and the suite is **38/38 green**. Nothing tests it.

**Precondition unreachable today.** `bridgeDispose` is `disposeSchemaNamespaceSession` -> `creatorRegistry.removeSession`, which is two `Map.delete` calls (`creator-registry.ts:53-56`) — it cannot throw. And nobody has demonstrated a throwing `vm.dispose()`; an attempt to force one by leaking a handle through the context disposed cleanly.

**If it ever did fire, it is a downgrade.** `runtime.dispose()` against a live context aborts:

```
Aborted(Assertion failed: list_empty(&rt->gc_obj_list), at: quickjs.c,2036,JS_FreeRuntime)
```

Pre-PR, a throwing `vm.dispose()` skipped `runtime.dispose()` and leaked ~17KB. Post-PR the outer `finally` runs it against a live context and kills the WASM module for the rest of the document — the exact abort `dispose-leak.test.ts`'s own header says it exists to prevent. And since "the last error thrown propagates", the abort **masks** the original error.

**It ships as a claim.** `.changeset/sandbox-dispose-no-double-free.md` states the `finally` "no longer strands it for the lifetime of the page", and the docblock at `sandbox.ts:221-223` repeats it. `@ifc-lite/sandbox` is >=1.0, so that sentence reaches consumers as fact.

Either drop the chain plus those two sentences, or demonstrate a reachable throwing `vm.dispose()`/`bridgeDispose()` and pin it.

## DRY

There are **four** existing swallow-a-teardown-throw sites, and the two nearest are in `packages/sandbox` itself — the package that should own this seam: `safeDispose` at `sandbox.ts:124`, eighty lines above the `dispose()` being rewritten, and `disposeOrphan` at `bridge-schema.ts:335`. This PR adds a fifth shape in a React hook rather than reusing one.

## Also

The branch is 4 commits behind `origin/main`, so every green result here rests on a stale merge base.

Everything above about the null-before-free half stands, and I would take that change on its own today.

BIMvoice:
You're right on all three counts. Chain dropped in 7a5a0dbc, rebased onto `d5b1fec4`.

**The downgrade argument is the one that settles it**, and I verified it rather than taking it: with the chain, a throwing `vm?.dispose()` still reaches `runtime?.dispose()` in the outer `finally`, which runs `JS_FreeRuntime` against a live context — the exact abort `dispose-leak.test.ts` exists to prevent. Trading ~17KB for killing the WASM module, and masking the original error while doing it, is strictly worse than the leak. `dispose()` is now three sequential calls.

**Unpinned**: confirmed independently — 38/38 green without the chain. **Unreachable**: `bridgeDispose` is `disposeSchemaNamespaceSession` → `creatorRegistry.removeSession`, two `Map.delete` calls (`creator-registry.ts:53-56`), which cannot throw.

**The claims are gone.** You were right that this was the sharper problem — `@ifc-lite/sandbox` is >=1.0, so the changeset sentence reached consumers as fact. Both it and the docblock now state the narrower truth, and the docblock says plainly *why* there is no `finally`, so nobody re-adds it as an obvious improvement.

**The surviving half is still pinned**: reverting the ordering fails `does not re-enter runtime teardown after a failed dispose` (1 failed / 37 passed; `REPLACEMENTS=1`, region re-read). Your `Lifetime.dispose()` mechanism — `_alive` cleared *after* the disposer runs, so `scope.dispose()` re-runs a still-alive lifetime — is a better account of the cause than mine, which was empirical. It also explains why an `if (rt.alive)` guard could not have worked.

## On DRY — I'd rather you pick than guess

The viewer guard is still needed: `dispose()` still throws, and an abort escaping a React cleanup or discarding the run's own result is the thing it prevents. But you're right that it's a fifth shape, and that `packages/sandbox` should own the seam. The two nearest — `safeDispose` (`sandbox.ts:125`) and `disposeOrphan` (`bridge-schema.ts:335`) — are both **local**, so reusing one means exporting it.

That's an API addition to a >=1.0 package, which is your call, not mine:

1. Export something like `disposeQuietly(disposable, context)` from `@ifc-lite/sandbox` and route all five sites through it — DRY, but new public surface and a wider blast radius than this PR.
2. Keep the viewer's local guard for now and track the consolidation separately.

I've left it as-is pending your answer rather than picking. If (1), say whether you want all five migrated in this PR or just the new one.

Also noted on the stale base — every earlier green here rested on it, which is why I re-ran after the rebase rather than trusting the previous run.

louistrue:
Verified `7a5a0dbc`. This is the right resolution, and the docblock is better than what I asked for.

I suggested dropping the chain and the sentences claiming it. You dropped it and **recorded why**, at `sandbox.ts:222-226`:

> The steps run in sequence rather than through a `finally` chain. Freeing the runtime after a throwing `vm.dispose()` would run `JS_FreeRuntime` against a live context, which aborts the WASM module for the rest of the document — a worse outcome than the ~17KB the skipped free costs, and it would mask the original error.

That is the version worth having. A silent removal leaves the next person free to re-add the `finally` as an obvious robustness improvement; this tells them what it costs and why the leak is the better trade. It also carries the number, so the tradeoff is checkable rather than asserted.

Re-verified on this head:

- the chain is gone — `dispose()` is now plain sequential after the three nulls
- the changeset no longer claims it (no `finally` / "strand" wording remains)
- **null-before-free is still pinned**: restoring the pre-PR ordering fails `does not re-enter runtime teardown after a failed dispose`, and only that test
- 38/38 green
- 0 commits behind `origin/main` — the stale merge base is resolved too

The remaining item from my earlier comment is the DRY one, and I do not think it should block: there are now four swallow-a-teardown-throw shapes in the tree, two of them inside `packages/sandbox` (`safeDispose` at `sandbox.ts:124`, `disposeOrphan` at `bridge-schema.ts:335`), plus the new one in the React hook. Consolidating them is a refactor with its own blast radius and does not belong in a dispose-ordering fix. Worth an issue rather than a commit here.

louistrue:
This one is legitimate and I nearly waved it through, so recording why I changed my mind.

My first read was that the bare `catch` is deliberate — the comment says so, the first `dispose()` is *expected* to surface the upstream abort, and the test's actual subject is that the second and third calls do not re-enter. Logging looked like noise on a path that fires every run today.

But it is a stated house rule. `AGENTS.md:17`:

> no silent `catch {}` (log or rethrow)

listed under the TypeScript rules that are **self-policed** — there is no eslint here, `pnpm lint` is a no-op, and `tsc --strict` does not flag `catch {}`. So nothing mechanical will ever catch this; the rule holds only if reviews hold it. That is precisely the case where letting one through because it looks reasonable is how the rule stops meaning anything.

Binding and logging costs a line and makes a change in shape visible — if upstream ever fixes the abort, or the error becomes something else, a silent catch tells you nothing while a logged one shows the transition.

Worth noting the rest of `AGENTS.md:17` since it explains why this is the kind of thing that needs a human eye: the Rust rules there are CI-enforced (clippy `-D warnings`, the module-size ratchet), and the TypeScript ones deliberately are not. The asymmetry means TS house-rule drift accumulates silently — which is the same shape as `ifc-lite-wasm` being excluded from clippy until #2051 cleared two real failures that had been sitting on main.

Everything else on `7a5a0dbc` is verified and good: the `finally` chain is gone, the changeset claim with it, null-before-free is still pinned by exactly one named test, 38/38, and the branch is current with `origin/main`.

louistrue:
Good finding, and the shape it proposes is genuinely better than what this PR removed — but I do not think it should go in, and the reason is the argument this PR already made.

**The premise is the same unreachable one.** Traced it at HEAD: `bridgeDispose` is assigned from `buildBridge`'s return (`sandbox.ts:91`), which is

```ts
dispose: () => { disposeSchemaNamespaceSession(context); }   // bridge.ts:56-58
```

and `disposeSchemaNamespaceSession` -> `creatorRegistry.removeSession` is two `Map.delete` calls. It cannot throw. That is exactly the reasoning used to delete the `finally` chain one commit ago.

**Where the suggestion is better than what was removed:** it is not harmful if it ever fires. The removed chain ran `runtime.dispose()` unconditionally, which against a live context aborts the WASM module for the document. This proposal explicitly skips `runtime.dispose()` when `vm.dispose()` failed, which is the correct handling of that hazard. So if `bridgeDispose` ever becomes throwable, **this is the shape to reach for** — retain the bridge error, still attempt `vm.dispose()` (a failed bridge teardown does not make the context live), and skip the runtime free only when the context free failed.

**Why not now:** adding it re-introduces defensive machinery for a case that cannot occur, immediately after removing defensive machinery for the same case. It would be untestable for the same reason — no test can construct a throwing `bridgeDispose` without changing `creator-registry.ts` — so it would land as another unenforced claim, which is what the last round was about.

The current gap is real but bounded: if `bridgeDispose` somehow threw, `vm` and `runtime` both leak. That is worse than the ~17KB the docblock discusses, and worth a sentence there so the next reader knows the sequential order is deliberate and what its cost would be — rather than code guarding a path nothing can take.

Resolving on that basis. If you would rather have the guard, the version in this finding is the right one and I would not argue further.

PR #2063, opened 2026-08-04
fix(sandbox): report a timeout inside a drained job instead of a stale value

BIMvoice:
Found while investigating #1922 and deliberately left out of #2062, which touches `dispose()` only. Independent of it.

## A timeout that reported success

QuickJS surfaces an interrupted **main body** as an eval error, but an interrupted **promise job** is swallowed by `executePendingJobs()`. So `result.value` still held the main body's value and `eval()` returned it as a success. Same spin loop, two answers:

```
main body spin   -> ScriptError: interrupted     (correct)
drained job spin -> RESOLVED with "started"      (silently truncated run)
```

The user gets a plausible answer from a run that was cut short — no error, nothing to notice. At the documented 30 s default this is reachable by any async script that overruns.

Nothing consulted the sandbox's own interrupt handler, which was the only remaining signal: it returned `true` and recorded nothing (`sandbox.ts:77-82`). The drain result is disposed without inspection at `:155-157`, and the only failure signal consulted was `result.error`, which reflects the main body alone (`:164-195`).

## The fix

The handler now records that it fired; the flag clears at the top of every `eval()`; it is checked after the drain — **after** the existing `result.error` branch, so the main-body path is byte-for-byte unchanged. It throws the same `ScriptError('interrupted')` QuickJS raises for a main-body timeout, so the two paths report identically rather than inventing a second error shape.

## Guarded against over-breadth

The obvious failure mode here is turning every async script into a timeout. `still returns the main-body value when the jobs complete` asserts `result.value === 'started'` **and** then evaluates `ran` in the same sandbox to prove the job actually executed — a no-op job would satisfy the value assertion alone. It passes both before and after the fix, which is the point. A fourth test pins the per-eval reset.

## Evidence

Each of the three parts is separately mutation-killed:

| mutant | replacements | outcome |
|---|---|---|
| drop `this.interrupted = true` from the handler | 1 | drained-job test fails |
| drop the post-drain `if (this.interrupted) throw` | 1 | drained-job test fails |
| drop the per-eval reset | 1 | `does not carry a timeout over to the next eval` fails |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, grep confirming the line was gone): 1 failed / 40 passed, and the negative-case test stayed green.

The test helper turns a resolve into a thrown `Error`, so a swallowed failure is not possible — the assertion fails on the *class* of what came back, not on a truthiness check.

sandbox 41 passing (was 37), extensions 598, sdk 62, typecheck 34/34.

`timeout-interrupt.test.ts` is its own file, and uses a **non-allocating** spin loop so it trips the CPU interrupt without risking the OOM abort from #1922 — that abort poisons the WASM module for the rest of the vitest worker, which is why `dispose-leak.test.ts` and `dispose-abort.test.ts` are separate too.

## Summary by CodeRabbit

- **Bug Fixes**
  - CPU timeouts during asynchronous sandbox jobs now report `ScriptError: interrupted` instead of returning an outdated result.
  - Normally completed jobs continue to return their expected results and execute side effects.
  - Subsequent evaluations remain unaffected after a timeout.

- **Tests**
  - Added regression coverage for main-script and asynchronous job timeouts, successful promise completion, and recovery after interruption.

### DIFF
--- a/.changeset/sandbox-job-timeout.md
+++ b/.changeset/sandbox-job-timeout.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Report a sandbox CPU timeout that hits an `async` script body instead of returning a stale value.
+
+QuickJS surfaces an interrupted top-level body as an eval error, so a script that spins in its main body already failed with `interrupted`. But an `async` function body runs as a promise job, and `executePendingJobs()` reports no error when the CPU deadline cuts one short — so `eval()` returned the value the main body had produced before the job ran and reported success. A script whose real work happened inside `async function run()` could therefore time out and still look like it had completed.
+
+The sandbox now records whether its interrupt handler actually fired and, after draining the job queue, raises the same `ScriptError: interrupted` the main-body path already raises. Scripts whose jobs complete normally are unaffected and still return the main-body value.
--- a/packages/sandbox/src/sandbox.ts
+++ b/packages/sandbox/src/sandbox.ts
@@ -46,6 +46,14 @@ export class Sandbox {
   private bridgeDispose: (() => void) | null = null;
   /** Mutable start time — updated by eval(), read by interrupt handler */
   private evalStartTime = 0;
+  /**
+   * Set by the interrupt handler when it actually cuts execution short.
+   * Cleared at the start of every eval(). QuickJS reports an interrupted
+   * *main body* as an eval error, but an interrupted *promise job* is
+   * swallowed by executePendingJobs(), so this flag is the only signal that
+   * the drained jobs hit the CPU deadline.
+   */
+  private interrupted = false;
   private readonly sessionId = createSandboxSessionId();
 
   constructor(
@@ -76,6 +84,7 @@ export class Sandbox {
       const timeoutMs = this.config.limits.timeoutMs ?? DEFAULT_LIMITS.timeoutMs;
       this.runtime.setInterruptHandler(() => {
         if (this.evalStartTime > 0 && Date.now() - this.evalStartTime > timeoutMs) {
+          this.interrupted = true;
           return true; // Interrupt execution
         }
         return false;
@@ -131,6 +140,7 @@ export class Sandbox {
       }
     };
 
+    this.interrupted = false;
     this.evalStartTime = Date.now();
 
     const result = this.vm.evalCode(jsCode, options?.filename ?? 'script.js');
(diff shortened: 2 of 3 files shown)


louistrue:
Verified `4c4a0324` and merging. A timeout that reports success is the sharpest version of the defect class this repo keeps turning up — the check ran, said pass, and verified nothing.

Mutation-checked both directions, which is what makes the positive path trustworthy:

| mutation | result |
|---|---|
| `if (this.interrupted)` -> `if (false)` | fails `reports a timeout inside a drained promise job the same way` |
| `this.interrupted = false` -> `= true` (never cleared) | fails **4** tests, incl. `does not carry a timeout over to the next eval` and `still returns the main-body value when the jobs complete` |

The second is the negative control and it is the reason I trust the first. If the flag were simply always set, the timeout test would pass vacuously — it does not, because four other tests catch it. 41 passing, CLI review of this head: 0 findings across its 3 files.

Two notes, neither blocking:

**1. The PR body cites `dispose-abort.test.ts`, which does not exist on this branch** — that file belongs to #2062. Easy to do when investigating two sandbox bugs at once, but it means the evidence table points at a file a reader cannot open from here.

**2. The drain's outcome is now read through two disjoint channels about thirty lines apart:**

```
sandbox.ts:166   this.runtime.executePendingJobs().dispose();   // result discarded unread
sandbox.ts:195   if (this.interrupted) { ... }                  // side-channel boolean
```

That split is exactly why the sibling bug survives: a drained job that *throws* (rather than being interrupted) still resolves as success, because the discarded result is the only thing that carries it. Worth an issue — the fix here is correct for the timeout case and I would not widen it now.

Merged 6 commits behind `origin/main`. Low risk given the change is confined to `packages/sandbox` and nothing merged since touches it, but worth noting the green CI was measured against the older base.

BIMvoice:
Thanks for merging, and both notes land.

**On the file citation — that one is mine and it is a real error.** The PR body's evidence table pointed at `dispose-abort.test.ts`, which belongs to #2062, not this branch. I was carrying two sandbox PRs at once and crossed them; the effect is that a reader following the evidence from here opens a file that does not exist. The actual test for this change is `timeout-interrupt.test.ts`, and the mutations you re-ran are the ones that matter.

**On the two-channel split: filed as #2077.** You are right that it is the reason the sibling survives. `executePendingJobs()`'s result is discarded unread at `:166` while the outcome is read from a side-channel boolean at `:195`, so a drained job that *throws* for any reason other than an interrupt still resolves as success with the stale main-body value — same defect, different trigger, and the flag added here cannot see it.

I kept it out of this PR for the reason you gave. The issue carries the suggested direction (read the drain result before disposing it and surface it the way `result.error` is surfaced, rather than adding a third channel) plus the two constraints learned here: a job that completes normally must still return the main-body value, and the result handle must still be freed on every exit including the new failure branch — a leaked handle is what makes `runtime.dispose()` abort the module.

Noted on the 6-commit-stale base too. Same issue you flagged on #2062, and I have rebased that one.

Issue #1985, opened 2026-08-02
Possible issue with geometry scaling in IFC


simple-debil:
### The problem

In general, it opens IFC files normally, but when opening certain files, I see heavily stretched pipes. I haven't looked into the code, but I suspect that the scale (in this case, uniform scale, 1000) is not taken into account somewhere in the transformation matrix of fragment of the model element.

v1.33.2

Btw, how can I see the draw call count, since multidraw indirect is not yet supported in WebGPU?

### Proposed solution

check, fix

louistrue:
Thanks for the report, and for the pointer at the mapped-item transform. I went looking there first.

## What I could not reproduce

I built fixtures matching your description exactly: geometry authored in metre numbers inside an `IfcRepresentationMap`, instantiated into a millimetre model through a uniform `IFCCARTESIANTRANSFORMATIONOPERATOR3D($,$,#origin,1000.,$)`, and compared the meshed result numerically against the identical solid authored directly in millimetres. Everything matched (same world bounds, same triangle counts):

- extruded `IfcCircleProfileDef` / `IfcCircleHollowProfileDef`
- `IfcSweptDiskSolid`
- `IfcFacetedBrep`
- `IfcBooleanClippingResult` (half space) and `IfcBooleanResult` (solid/solid, exact kernel)
- with GPU instancing off and on, and with the map owned by an `IfcPipeSegmentType`

So the uniform Scale itself is applied on all three axes. That is now pinned by a regression test.

## What I did find, and fixed

The mapped-item transform is `MappingTarget · MappingOrigin`. The mesh path never read `IfcRepresentationMap.MappingOrigin` (attribute 0) at all, so any map with a non-identity origin placed its occurrences at the wrong spot. Because the origin composes *inside* the target, a scaling target multiplies that error: with your `Scale = 1000`, a 1 mm origin offset lands 1 m away.

Three smaller defects in the same operator code, found while confirming that:

- `IfcCartesianTransformationOperator2DnonUniform` keeps `Scale2` at attribute 4 (the 2D forms have no `Axis3`), but the parser read the 3D layout, so `Scale2` was dropped and Y silently fell back to the X scale.
- `Axis2` was never read, so a mirroring frame (`Axis2` anti-parallel to `Axis3 × Axis1`) was silently un-mirrored.
[...]

PR #2064, opened 2026-08-04
test(geometry): pin sub-unit and nested mapped-item scaling (#1985)

BIMvoice:
**#1985 does not reproduce.** This PR carries no fix — it adds the coverage the investigation showed was missing. Full refutation posted on the issue.

## The reporter's mechanism is ruled out by their own file

They narrowed it to `#12=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.)`. With `LENGTHUNIT` and a null prefix, `try_extract_length_unit_scale` returns exactly `1.0` (`rust/core/src/units.rs:333-336`), and both scale sites are gated on `unit_scale != 1.0` (`rust/geometry/src/router/mod.rs:695,707`). So on that file every length-unit path is a literal no-op — the "uniform scale (1000) not taken into account" theory cannot be what they are seeing. They state they never read the code, so it was a guess.

Two further signals point at the file: **BIMcollab Zoom 9.8 renders the same wrong long pipes** (only xBIM Xplorer is correct), and the spatial tree is empty in *both* ifc-lite and xBIM. Two independent engines failing identically is a file-shape signal.

## What I could not rule out, stated plainly

Which representation type the pipes use, and whether a transform is involved at all. No file, no STEP excerpt, no screenshot, and the reporter cannot share.

## The coverage gap this found

Your #1990 suite is green on today's main and pins `Scale = 1000` in a **millimetre** model. The reported shape is the opposite: a **METRE** model with a map authored in millimetre numbers pulled down by `Scale = 0.001`. Nothing covered that direction, so I built a fixture in it — using their exact `1734.50 mm` number — plus a non-uniform 3D operator and a **nested** `IfcMappedItem`:

```
=== instancing = false ===
#74  size = [0.100000, 0.100000, 1.734500]   uniform Scale = 0.001
#81  size = [0.100000, 0.100000, 3.469000]   3DnonUniform (0.001, 0.001, 0.002)
#88  size = [0.100000, 0.100000, 1.734500]   nested map (outer 1 x inner 0.001)
#99  size = [0.100000, 0.100000, 1.734500]   authored directly in metres
=== instancing = true ===   (identical bounds on all four)
```

No stretch in either scale direction, uniform or non-uniform, flat or instanced, single-level or nested.

## The assertions discriminate

| mutation | result |
|---|---|
| drop `Scale3` (3D non-uniform Z falls back to X) | fails — axis 2 was 1.7345, expected 3.469 |
| ignore the `Scale` attribute entirely | fails all 3 — axis 0 was 100, expected 0.1 |
| apply the mapping transform only at `depth == 0` | fails **only** the nested assertion |

The second reproduces exactly the 1000x-too-long pipe #1985 describes — which is the argument for keeping this test. Had the reported defect existed in the transform path, this would name it. The third confirms the nested coverage is genuinely new rather than incidental.

## Sibling paths checked for the one-sided-scaling shape

- `parse_transformation_operator` is now the single definition of the operator layout; `profile_extractor` calls it rather than carrying its drifted copy. 2D vs 3D layouts (`Scale2` at attr 4 vs 5/6) are correct.
- Both mapped-item implementations — `process_mapped_item_cached` (`processing.rs:869`) and `collect_submeshes_from_item_inner` (`:325`) — apply `scale_mesh` at the leaf and `scale_transform` to each level's mapping transform; the nesting algebra composes to the same result as scaling once at the end.
- Caches are keyed correctly: pre-transform source coords keyed by `source_id`, and `item_dedup_key` folds `unit_scale`, `rtc_offset` and quality (`:822`). No occurrence can inherit another's scale.
- `IfcVector.Magnitude` on an `IfcLine` directrix **is** honoured (`profiles/curves_3d.rs:47,62`) — a strong candidate for a 1000x-long pipe, ruled out.

## Unrelated observation, not fixed

`process_mapped_item_cached` skips nested `IfcMappedItem`s outright (`processing.rs:958`, `continue`) as a stack-overflow guard. Reached only from `process_representation`/the void probe, not the streaming submesh path the viewer uses, and it *drops* geometry rather than stretching it. Worth a separate look; not this issue.

Test-only, no source change, so no changeset. `cargo test -p ifc-lite-processing -p ifc-lite-geometry`: 743 passed / 3 failed, identical before and after — the 3 are `wall_opening_cut_regression.rs` panicking on un-downloaded checksummed fixtures, pre-existing and untouched by this. Clippy clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of scaled and nested mapped geometry, including uniform and non-uniform transformations.
  * Ensured geometry dimensions remain accurate when converting between sub-unit and metre-based representations.
* **Tests**
  * Added regression coverage for instanced and non-instanced mapped items, nested scaling, and direct metre-authored geometry.


louistrue:
Verified `f03658d9` by running the Rust test, not by reading it. **greptile's finding is wrong** — but the probe that disproved it turned up something else.

## No panic — #81 is materialized under both settings

```
$ cargo test -p ifc-lite-processing --test issue_1985_mapped_item_transform --no-fail-fast
test uniform_scale_1000_matches_direct_millimetre_authoring ... ok
test sub_unit_scales_nonuniform_and_nested_maps_match_direct_metre_authoring ... ok
test mapping_origin_is_composed_inside_the_mapping_target ... ok
test result: ok. 3 passed; 0 failed
```

`size(81)` does not panic with instancing on. Resolving that thread.

Worth noting the Rust lane genuinely ran on this PR — `Rust tests: success`, plus `Build + WASM + Rust + Node` and the watertightness census — so this was not one of those PRs where a skipped lane makes a claim untestable.

## But the instancing loop is a duplicate for this fixture

`sub_unit_scales_nonuniform_and_nested_maps_match_direct_metre_authoring` runs `for instancing in [false, true]`, which reads as covering both configurations. It does not:

```
PROBE instancing=false meshes=4 instances=0 has81=true
PROBE instancing=true  meshes=4 instances=0 has81=true
```

Identical. `instances=0` either way — nothing in `issue_1985_metre_submm_scale.ifc` is ever instanced, so the second iteration re-runs the first. That is *why* greptile's prediction failed, and it is also why the loop is not buying anything here.

Contrast your `mapping_origin_is_composed_inside_the_mapping_target`, which handles this properly — `if !instancing` for the materialized assertions, then reaching into `res.instances` to reconstruct #47 for the instanced case, with a comment explaining that the transform "cancels here and would leave every instanced occurrence unpinned". That test really does cover both paths.

Two honest options:

1. **Drop the loop** and note that this fixture has no shared maps, so instancing cannot engage. Cheapest, and it stops the test claiming a matrix it does not have.
2. **Give the fixture a shared map** so the instanced path is real for sub-unit and nested scaling too — which is the case most likely to expose a scale composition bug, since that is what the PR is pinning.

(2) is more valuable if sub-unit scaling and instancing can interact; (1) is right if they cannot. You are better placed than I am to judge which, but as it stands the loop implies coverage the fixture cannot deliver — the same "asserts more than it tests" shape the rest of this PR is careful about.

Nothing was pushed; the probe was a temporary `eprintln!` reverted immediately.

louistrue:
Merging. Under a blast-radius-tiered review this is the one open PR touching `rust/` — and it is two files, a test and a fixture, adding coverage for sub-unit and nested mapped-item scaling. Holding kernel *coverage* over a redundancy in the test would be backwards.

Verified by running it, not reading it:

```
test uniform_scale_1000_matches_direct_millimetre_authoring ... ok
test sub_unit_scales_nonuniform_and_nested_maps_match_direct_metre_authoring ... ok
test mapping_origin_is_composed_inside_the_mapping_target ... ok
test result: ok. 3 passed; 0 failed
```

greptile's predicted panic at `size(81)` does not happen — already resolved that thread.

**The nit stands as a follow-up, not a blocker:** `for instancing in [false, true]` is a duplicate for `issue_1985_metre_submm_scale.ifc`, because nothing in that fixture is ever instanced:

```
PROBE instancing=false meshes=4 instances=0 has81=true
PROBE instancing=true  meshes=4 instances=0 has81=true
```

Identical both times. Either drop the loop with a note that the fixture has no shared maps, or give it one so the instanced path is real for sub-unit scaling too — the second is the more valuable of the two if scale composition and instancing can interact, since that is exactly what this file pins. Your `mapping_origin_is_composed_inside_the_mapping_target` already does this properly, reaching into `res.instances` for #47, so the pattern is right there.

Worth doing when you next touch this file: a loop that implies a matrix it cannot deliver is the same "asserts more than it tests" shape the rest of this PR is careful about. Not worth another round-trip on its own.

BIMvoice:
Fixed the loop-coverage gap in `sub_unit_scales_nonuniform_and_nested_maps_match_direct_metre_authoring`.

**Decision: (2), give the fixture a real shared map** — not (1) dropping the loop.

Evidence it *can* engage: the router's don't-bake eligibility (`router/processing.rs` around the `instance_solid_id` computation, backed by `router/instancing.rs::mapped_source_single_item`) fires whenever ≥2 top-level `IfcMappedItem`s share one `IfcRepresentationMap` — regardless of whether they use different `MappingTarget`s. The code even calls this out explicitly: "occurrences sharing a map but differing by target collate under one template" (processing.rs, near the `nontrivial_target` composition). Sub-unit/non-uniform scaling composing correctly under that instanced `local_transform` path is exactly the risk this PR is pinning, so a fixture that never engages it was leaving the actual sensitive path untested.

What I changed in `issue_1985_metre_submm_scale.ifc`: occurrences A (uniform `Scale=0.001`) and B (non-uniform `0.001/0.001/0.002`) now share the same `IfcRepresentationMap` (only differing by `MappingTarget`), which is precisely the router's eligibility condition. The nested-map occurrence (#88) got its own separate source instead of reusing A's — reusing it inflated the shared source's occurrence count with a *nested* `IfcMappedItem` (lower express id than either top-level occurrence), which would have gotten miscounted as the don't-bake template even though only depth-0 items ever run the template/retag path. That silently starved both real occurrences of a template and routed everything through orphan recovery — correct output, but instancing never actually engaged. Fixture comment explains this so it isn't repeated.

Probe, before → after (same fixture, `instancing=true`):
```
before: meshes=4 instances=0   (identical to instancing=false — the bug)
after:  meshes=3 instances=1
```
A materializes as the template; B rides as a real `InstanceRecord` (`template_express_id == 74`). The instanced branch reconstructs B's absolute world geometry from `record.transform` against the template, mirroring the pattern already used in `mapping_origin_is_composed_inside_the_mapping_target`.

Mutation test on the new instanced assertion: replaced the placeholder's `local_transform` composition (`geometry/src/router/processing.rs`, the non-template `let local_rm = mapping_transform.map(...)`) with `Some(IDENTITY_ROW_MAJOR)` — one exact-substring replacement, confirmed via a Python match count of 1. Result: the new instanced-reconstruction assertion failed (`axis 0 was 100.0, expected 0.1`), confirming it actually exercises the composition. Restored the router by file copy (not `git checkout --`) and confirmed `git status` is clean on that file.

Nested map (#88) stays excluded from instancing on both settings by design — `mapped_source_single_item` returns `None` for a nested-mapped source, so it's provably never don't-bake eligible; no coverage claim needed there.

Test counts: `cargo test -p ifc-lite-processing` — 156 passed, 0 failed, before and after (this is a coverage fix, not a new test). The 3-test file itself: 3 passed, 0 failed.

`cargo clippy --workspace --all-targets -D warnings` currently fails on unrelated pre-existing lints in `rust/core` (`georef.rs`, `parser/scanner.rs` — new `chunks_exact`/`question_mark` lints from a newer clippy toolchain). Confirmed these predate this branch's divergence from `main` (no commits touch those files between the merge-base and current `upstream/main`) and are untouched by this diff — not something I've fixed here, flagging in case it's news.

Diff is 2 files: the fixture and the test.

PR #2066, opened 2026-08-04
test(export): find the document root by identity, not by position (#2047)

BIMvoice:
The two minors you recorded while merging #2047. Both confirmed still on `main`.

## 1. The tautological root lookup

`ifc5-exporter.test.ts:874` was

```ts
const rootNode = file.data.find((n: any) => n.path === file.data[0].path);
```

which returns `file.data[0]` whatever it is. You were right that it happens to work today only because the document root is emitted first — and right that the file already had the correct idiom, at `:963` and in the reachability helper. It is now looked up by `DOCUMENT_ROOT_PATH` with an existence assertion.

Worth one check rather than just asserting the improvement: pointing that lookup at a bogus path now **fails** the test —

```
× re-parents a surviving child to the nearest surviving ancestor when its direct parent is deleted
  Tests  1 failed | 346 passed | 29 skipped
```

— which the previous form could not do for any input at all. So the fix is observable, not cosmetic.

## 2. The changeset wording

Reworded to "#2046 remains open for the Parquet half". Your reading is right: "tracked separately in #2046" points at the issue this PR closed, which is confusing precisely for the reader who goes looking. The changeset is **still unconsumed** in `.changeset/`, so this lands before the text reaches `@ifc-lite/export`'s published CHANGELOG rather than needing a correction after the fact.

347 passed / 29 skipped. Test and changeset only — no source change, so no new changeset.

## Summary by CodeRabbit

- **Tests**
  - Improved regression test coverage for IFC5 exports, ensuring document hierarchy handling remains reliable even when output ordering changes.

- **Documentation**
  - Clarified that the Parquet export issue remains open and is being tracked separately from the IFC5 export updates.


louistrue:
Verified `317ffcbb`. Both items are properly closed, and the fix is testable in the way that matters.

**The root lookup is now by identity at all five sites** (`878`, `968`, `1165`, `1380`, plus the traversal helper at `108`), using the `DOCUMENT_ROOT_PATH` constant the file already had.

The property that proves it: moving the root node from first to last in the emitted array (`nodes.unshift(` -> `nodes.push(`) leaves **all 32 tests passing**. The suite no longer depends on emission order, which is exactly what the tautology was silently relying on.

I tried to build the mirror-image demonstration — old test file plus the same reordering, expecting failures — and could not get a clean result out of it, so I am not claiming one. The forward property stands on its own: position-independent now, position-coupled before.

**The changeset wording** reads `#2046 remains open for the Parquet half`, which says the thing the previous phrasing implied backwards.

**The `as any` quantities stub is gone too** — replaced with `new QuantityTableBuilder(strings).build()`. That was the third finding from the CLI run on #2047 and I had not mentioned it; good catch picking it up.

`@ifc-lite/export` 347 passing. (`lod1-generator.test.ts` fails to load on an unbuilt `@ifc-lite/wasm` — worktree staging, not this branch.)

Merging once CI settles.

PR #2068, opened 2026-08-04
fix(geometry): recurse into nested IfcMappedItems instead of dropping them

BIMvoice:
Found while investigating #1985 and deliberately left out of #2064. It is the one-sided-guard shape again.

## The asymmetry

`rust/geometry/src/router/processing.rs:955-962` skipped nested mapped items outright:

```rust
// Skip nested MappedItems AND IfcBooleanClippingResult that reference MappedItems
// to prevent stack overflow from deeply nested recursive geometry
for sub_item in items {
    if sub_item.ifc_type == IfcType::IfcMappedItem {
        continue;
    }
```

The sibling `collect_submeshes_from_item_inner` (`:326`) has handled nesting all along — depth cap against `MAX_MAPPED_ITEM_DEPTH` (`:338`), a per-walk `visited` set (`:347`), recursion at `:479`. The primary viewer path for an ordinary occurrence uses the sibling, which is precisely why the two drifted without anyone noticing.

The comment also over-claims: the `IfcBooleanClippingResult` half it describes was never implemented — only the mapped-item arm exists. Corrected.

## User-visible, not latent

I went looking for whether anything real reaches it. Three live paths do:

- **Type-product geometry** — `rust/processing/src/element.rs:658` → `textured.rs:109`, flowing through `produce_type_geometry` into `MeshData`, i.e. the viewer and every geometry-fed export. Worth noting that site *already* processes a **top-level** nested mapped item deliberately, with a comment about not dropping its geometry — one level deeper was still dropped.
- **Void cutters** — `voids/synthesis.rs:95` builds the opening cutter via `process_element`. An `IfcOpeningElement` whose map contains a nested mapped item yields a truncated or empty cutter, so **the host is under-cut or not cut at all and renders solid**.
- **Void probes** — `voids/probe.rs:511,612` under-detect extent and direction.

## The fix reuses the sibling's seam

`process_mapped_item_cached` is now a thin wrapper seeding `depth = 0` and an empty `visited`, delegating to an inner function that applies the **same** cap and cycle check. Inventing a second bound here is how the two drifted in the first place. The recursive result is already unit-scaled with its own `MappingTarget` baked in, so composing this level's scaled transform over the merge is the same algebra the sub-mesh walker applies per sub-mesh.

## Evidence

| mutation | result |
|---|---|
| discard the recursive mesh | reproduces the pre-fix symptom exactly — `max X = 1 m`, expected `12 m` |
| delete depth cap **and** visited check | cyclic fixture → **real `stack overflow, SIGABRT`** |
| delete visited check only | both pass — the cap alone suffices; the two are defence in depth, matching the sibling |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, line re-read): `nested mapped item dropped: X extent is [0,5,0]..[1,6,1], expected max X = 12 m`. Two fixtures: a nested map (mm units, outer map = own 1 m cube plus a nested mapped item at 2× scale offset +10 m X, occurrence target +5 m Y) and a mutual A↔B cycle.

900 passed / 3 failed — the same three as baseline (`wall_opening_cut_regression.rs` panics on an undownloaded checksummed fixture). Clippy clean. `processing.rs` 1053 → 1112 against a 1183 budget, so no ratchet bump.

## One judgement call for your review

On a nested-item error the loop **logs and skips** rather than propagating, matching both the sibling walker and this loop's existing `if let Ok(...)` handling of failing processors. So a cyclic chain still yields the geometry it legitimately reached instead of blanking the element. Say if you would rather it propagated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed missing geometry when mapped items contain nested mappings.
  * Preserved nested transformations, including scaling and translation, across rendered geometry, exports, and fallback processing.
  * Improved handling of partially processed mapped geometry to prevent incomplete results from being reused.
  * Added safeguards for cyclic or excessively deep mappings so processing terminates reliably without disrupting valid geometry.

* **Tests**
  * Added coverage for nested mapped geometry, transformed bounds, cyclic mappings, depth limits, and caching behavior.


louistrue:
The geometry fix here is the strongest evidence of the batch — real loss on a shipped checksummed fixture (`ara3d/ISSUE_034_HouseZ.ifc`: `IfcStair #4325` and `IfcRoof #42044` return **zero** indices pre-fix vs 324/36), an end-to-end proof through the void path, and a named test that fails on full revert. I am not disputing any of that.

But I would not merge it yet, because the cache finding is real and it is **your own argument applied to a sibling case**.

## The truncated mesh gets cached

The `Err` arm at `processing.rs:1006-1018` turns a depth or cycle failure into a successful *partial* mesh — it logs and continues, so the outer walk returns whatever it accumulated. That mesh then reaches the insert at `:1039` keyed on `source_id` alone.

The existing guard there does not catch it. It skips the shared insert only when the mesh is empty or the CSG budget tripped:

```rust
if !mesh.positions.is_empty() && !crate::kernel::budget::tripped() {
```

A recursion-truncated mesh is non-empty and trips no budget, so it is cached model-wide and served to every later occurrence of that source.

## Why this is your argument, not mine

The comment immediately above that guard already makes the case:

> Caching that degraded source MODEL-WIDE would serve the wrong (uncut) mesh to a later occurrence in a fresh-budget element that would otherwise get the full exact cut.

Depth/cycle truncation is the same shape: a degraded source, cached under a key that does not encode what degraded it. The cache is keyed on `source_id` while the value now depends on walk depth and the visited set — which this PR is precisely what introduces. Your `cyclic_mapped_item_chain_terminates` test proves the truncation path is reachable.

## Second item, independently corroborated

`nested_mapped_item.rs:98` accepts an empty result: with an empty mesh `bounds` returns `NEG_INFINITY` for `max`, so both current assertions hold vacuously. A separate review pass flagged the same test for a different reason — it has 8x/25x headroom, and neither guard is individually pinned. `assert_eq!(indices, 72)` would close both.

Two reviews reaching the same test by different routes is usually a sign the test is the weak point rather than the code.

## Follow-ups worth carrying regardless

`instancing.rs:66` still holds the one-sided guard this PR deletes elsewhere, and its comment at `:43` — "Mirrors `process_mapped_item_cached`'s miss path" — is now false. That is the exact drift shape that produced this bug: a comment asserting parity with code that has since moved.

BIMvoice:
All three fixed in e93664b2. The cache one was the real find, and you were right that it is our own argument applied to a sibling case — the comment sitting directly above that guard already made it.

**Reproduced rather than reasoned about.** A 40-link chain of `IfcRepresentationMap`s: `#44` enters at link 1 and dies at the depth cap, `#49` enters at link 32 with a fresh budget and should get 9 cubes.

```
without the guard: #49 got 36 indices (1 cube), expected 324
with the guard   : 324 indices, X 0..17 m; shared cache holds exactly maps 32..40
```

Non-empty, budget-clean, model-wide — exactly your scenario. I re-ran that mutation myself on the pushed tree (`REPLACEMENTS=1`, line re-read): `depth_truncated_source_is_not_shared_cached` fails with `#49 got 36 indices`.

A per-level `truncated` flag is ORed into the caller's, so a clean sibling is still cached, and the guard is extended in place rather than duplicated. The `RefCell` arm is untouched per the `:1048-1049` comment. Neutralising just the propagation caches all 40 maps instead of 32..40, so the OR-into-caller is separately pinned.

**Negative case**: an ordinary nested source still gets cached (exactly maps 14 and 21), asserted, and it passes with *and* without the fix — so the guard did not simply switch caching off.

**On the test: 72 was right, and I measured it rather than taking it.** The cyclic fixture yields 72 indices / 144 positions, bounds max `[1.2, 1.0, 1.0]`. It now asserts that exactly plus exact bounds. Both failure modes demonstrated: an empty mesh fails `0 != 72` where `max[0] < 100.0` passed vacuously on `NEG_INFINITY`, and a truncated one fails at 36 where it passed *both* old assertions. Your read that two reviews converging on one test meant the test was the weak point was correct.

**`instancing.rs`: narrowed the claim and fixed the guard.** I did not make it recurse — its only caller (`processing.rs:444`) reaches it past `mapped_source_single_item`, which returns `Some` only for a source of exactly one non-mapped item, so a nested item cannot occur there and adding the depth/visited machinery for an unreachable case is dead weight. But the `continue` was the dangerous half: if it ever fired it would insert a short mesh into the *same* model-wide cache — item 1's bug in the other function. It now `return`s without inserting, with its own test (`return` → `continue` fails it). The doc no longer asserts parity with `process_mapped_item_cached`'s miss path; it says which guards are shared and that the recursion deliberately is not.

903 passed / 3 failed (the same pre-existing fixture-not-downloaded three), clippy clean, `processing.rs` 1112 → 1142 against a 1183 budget, no ratchet bump.

One note: `cargo fmt --check` reports diffs in `rust/geometry/` that predate this branch, and this PR's own test file already deviated at line 64. I matched surrounding style rather than reformat unrelated code — say if you want the new lines normalised.

louistrue:
Verified `e93664b2`. All three items closed, and the cache fix is pinned in both directions — which is the part I care about, because an over-broad guard here would silently disable model-wide caching.

**Truncation guard.** The `truncated` flag threads through the recursion and ORs up, so it reaches every enclosing level whose merged mesh is short. Mutations:

| mutation | result |
|---|---|
| drop `&& !level_truncated` | fails `depth_truncated_source_is_not_shared_cached` |
| replace it with `&& false` (never cache) | fails **both** `depth_truncated_source_is_not_shared_cached` **and** `fully_walked_source_is_still_shared_cached` |

The second is the negative control, and it is why the first is trustworthy: a guard that simply never caches would satisfy the truncation test on its own. `fully_walked_source_is_still_shared_cached` is what stops that, and writing it unprompted is the right instinct.

The comment carries the reasoning I was going to have to reconstruct otherwise — that `source_id` does not encode depth or the visited set, with the concrete case: a source first met at depth 31 loses everything below it, and caching that model-wide serves the short mesh to a later occurrence reached at depth 0.

**Cyclic test tightened** to `assert_eq!(mesh.indices.len(), 72)`, with the reason recorded: an upper bound alone passes on an empty mesh, since `bounds` returns `NEG_INFINITY` for max. That was flagged independently by two reviews from different directions, which usually means the test rather than the code was the weak point — and it was.

**`instancing.rs` comment corrected.** It no longer claims to mirror `process_mapped_item_cached`'s miss path. It now states which guards it shares (empty mesh, #1257 budget trip) and which it deliberately does not (the nested recursion), why that is safe today (the only caller has passed `mapped_source_single_item`, so the source is exactly one non-mapped item), and what happens if that changes. That is the drift shape that caused this bug in the first place, so closing it here matters more than it looks.

Fixtures confirmed present in my worktree before running, so the fixture-gated results are meaningful rather than dark. Geometry suite green.

Nothing outstanding from my side on this PR.

Issue #2060, opened 2026-08-04
[Bug] Visibility filter does not take effect for spatial elements in Section mode


xyzbety:
In the model (AC20-FZK-Haus.ifc), even when Spaces and Openings under Visibility are hidden, section outlines of IfcSpace and IfcOpening can still be seen in the 3D view after enabling the Section function. In the 2D Section view, fills and outlines for IfcSpace and IfcOpening remain visible.
I suspect the same issue also exists when hiding Site, Virtual Elements and Spatial Zones in Visibility.
Expected behavior: these elements should be fully hidden globally.

<img width="2551" height="1308" alt="Image" src="https://github.com/user-attachments/assets/a988e53f-a2d5-4469-9caf-01e010d30e36" />

BIMvoice:
Looks closeable — **#2070** and **#2123** are both merged.

#2070 routed the section's cut and its construction-projection profiles through the same class gate as the 3D viewport, and #2123 closed the last ungated route (the symbolic annotation overlay), which you had split out as #2121.

Verified on `main`: `Section2DPanel.tsx` reads `typeVisibility` at 3 sites, and every `typeVisibility`-mapped class is now gated in the section.

@louistrue closeable if you agree.

louistrue:
Fixed by #2070 and #2123 (both merged).

#2070 threaded `typeVisibility` into `useDrawingGeneration` so the section honours class toggles for both the cut and the construction-projection profiles. Mutation-verified: removing the profile filter fails a named test, and filtering everything fails two (the fixture carries a mesh-less-space control, so an over-broad filter is caught in the same test).

#2123 closed the last route — the symbolic annotation overlay, which was gated only on the panel's own display option and ignored `typeVisibility.ifcAnnotations` (tracked as #2121).

PR #2070, opened 2026-08-04
fix(viewer): keep type-library and hidden-class geometry out of 2D drawings (#2058, #2060)

### LINKED ISSUE #2058: [Bug] Geometry rendering corrupted after clicking Types under Visibility
In the model (AC20-FZK-Haus.ifc), the geometric rendering of the model becomes corrupted after clicking Types under Visibility.
Image

Additionally, projections of these corrupted element geometries can also be observed in standard 2D Section views.
Image

BIMvoice:
Closes #2058, and folds in the entire #2060 fix (commit `68fe1ce8`) — **class-level Visibility toggles** (Spaces, Openings, Site, Virtual Elements, Spatial Zones, Annotations) threaded through `Section2DPanel`, the `TypeVisibilityGate` type, and a regeneration trigger. Same file, same function, adjacent lines to #2058's fix; landing them separately would have been a typecheck failure, not a mergeable split (`useDrawingGeneration.test.tsx`'s harness needs the `typeVisibility` prop `68fe1ce8` adds). **Your 3D ruling stands and nothing here touches it.**

## Reproduced on the real fixture

`AC20-FZK-Haus.ifc` through the live mesh path, then fed to the real `Drawing2DGenerator` on a plan cut at y=1.561 (model spans −1.00…6.32), which is what the Section 2D panel does:

```
BEFORE: meshes=317 lines=762 polygons=235 | type-entity lines=104
        type ids drawn=[15234,17730,21177,22346,27350,31392,59692]
AFTER:  meshes=285 lines=658 polygons=197 | type-entity lines=0
```

104 of 762 cut lines and 38 of 235 cut polygons belonged to `IfcWallType`/`IfcDoorType` templates rather than to the building. **These counts come from driving the generator directly, not from looking at a rendered plan view** — no browser check has been done in either review pass, and this lands in the live 2D-plan render path. Worth one look at an actual Section 2D panel before merge.

**One correction to the report:** the bleed is **unconditional** — it does not depend on the Types toggle at all. The user simply noticed it after switching. So this was reachable in a default plan view the whole time.

## Root cause — the one-sided-guard shape

- **3D**: `ViewportContainer.tsx:829` gates every mesh through `isMeshVisibleInViewMode(...)` (`lib/type-view-visibility.ts:37`).
- **2D**: `useDrawingGeneration.ts:637-660` started from `geometryResult.meshes` and filtered only on hidden ids / isolation / storey isolation — **no `geometryClass` awareness anywhere in the file**.

Grepping `isMeshVisibleInViewMode|meshClassIsPlaced` returned exactly **one** non-test consumer. That is how the two drifted.

**A second consumer had the identical defect**, found by grepping `Drawing2DGenerator`: `useConstructionUnderlay.ts:41`, the Space Sketch plan underlay, also cut raw meshes. Fixed too.

**Category visibility (#2060) is a third mechanism**, distinct from both the hidden-id set and #2058's `geometryClass` gate: a store object of booleans resolved through `IFC_TYPE_TO_VISIBILITY_KEY`/`isTypeVisible`. The 3D mesh path applies it at `ViewportContainer.tsx:838`; the drawing path never did, so with Spaces and Openings hidden, `AC20-FZK-Haus.ifc` still put 6 `IfcSpace` and 14 `IfcOpeningElement` entities into the drawing (26+70 lines, 6+28 cut polygons) — reachable by default, since both toggles ship `false`.

`extract_profiles` (`rust/geometry/src/profile_extractor.rs:85`) gates on `has_geometry_by_name` → `is_subtype_of(IfcProduct)` (`schema_helpers.rs:102`), so today it cannot itself emit a profile keyed to an `IfcXxxType` id — verified by reading the Rust source, not assumed. **That is a property of the Rust extractor, though, not something this filter enforces from the TS side**, and a wall occurrence dropped from the cut by the class gate (`geometryClass` 1/2) still shares its express id with a real, unfiltered profile — see the #2070 review section below for why that mattered anyway.

## The fix reuses the existing predicate

`selectModelMeshes` sits beside `isMeshVisibleInViewMode` and applies it **pinned to `'model'`** — a standard 2D drawing is always the real building; Model/Types is a 3D control. Pinning rather than hard-excluding preserves the annex-E case: a pure type-library file still draws its class-1 orphans. `storeyFloorsFromMeshes` gets the filtered set too.

**On the storey-band claim — corrected per review.** The original version of this section claimed type templates "can't invent a phantom storey band." That overstates it: the guard is **inert on ordinary files**. Type-product meshes carry the *type's* own express id (`processor/mod.rs:841`), and `elementToStorey` is built from `IfcRelContainedInSpatialStructure`, whose `RelatedElements` are products — so on AC20-FZK-Haus, or any normal file, a template never reaches `elementToStorey` and contributes no storey level either way, filtered or not. The one path that *does* reach it is `spatial-hierarchy-builder.ts:300-311`, which walks `IfcRelAggregates` from every contained element and maps every descendant with no type check (`RelatedObjects` is `IfcObjectDefinition`, which admits `IfcTypeObject`) — an unusual file shape, not the fixture above. `storeyFloorsFromMeshes(modelMeshes)` is a consistency guard for that shape, pinned by a synthetic-fixture test (`useDrawingGeneration.projection.test.tsx`), not a fix for a defect observed on a real model.

## Evidence

Replacing `selectModelMeshes`' body with `return [...meshes]` fails **5 tests** across both suites — I re-ran that myself on the pushed tree (`REPLACEMENTS=1`, region re-read). Restored: 16/16.

Both negative cases asserted, because "exclude more from 2D" is the obvious wrong fix:
- **occurrence geometry still reaches the drawing** (658 lines survive on FZK-Haus) — asserted in two tests;
- **a pure type library still draws its orphans** — asserted, and it's the subtest that passed even in RED.

RED used the real hook rendered, not a mock, and the harness demonstrably renders: the passing subtest asserts geometry *is* produced while the failing one reports real cut entity ids (`got ids 100,200`).

No changeset — `@ifc-lite/viewer` is `private: true`.

## #2070 review follow-ups

**Main defect, fixed.** `projectionProfiles` (construction-projection profiles, a route into the drawing that never reads `geometryClass` at all) was filtered only on hiding/isolation/`isTypeVisible` — not on the same class gate `selectModelMeshes` applies to the cut. Added `buildModelViewIdFilter` (`lib/type-view-visibility.ts`), the same `isMeshVisibleInViewMode` predicate keyed by express id instead of by mesh, and applied it to `projectionProfiles` in `useDrawingGeneration.ts`. An id with no mesh at all (e.g. a mesh-less `IfcSpace` whose only route in is its profile) is untouched — only ids the class gate actually dropped from `meshesToProcess` are excluded from projection too.

Regression test (`useDrawingGeneration.projection.test.tsx`, "profile route mirrors the mesh-class gate"): a real `IfcWall` occurrence (real mesh, real extruded-solid profile via the wasm extractor) tagged `geometryClass: 2` reaches RED without the fix — its profile still projects (`got 38,48,99`) — and GREEN with it, while a mesh-less space's profile is asserted unaffected as a control. `showConstructionProjection: true` is required to reach the `if (projectionOn)` block at all, which — as flagged in review — both existing test files leave `false`, so this block was previously dead in the suite, not just thinly covered.

Mutation-tested both the new hook filter line and the new `buildModelViewIdFilter` predicate (`REPLACEMENTS=1` each, region re-read after edit): both drop the suite from 0 to 1 failure — the new test, specifically — confirming it is the thing catching the regression. Restored by file copy.

**Scope disclosure — fixed** (this body; title already named both issues).

**Identity-compare comment (`useDrawingGeneration.ts:~1104`) — annotated.** It was true but unpinned when first flagged. `visibilitySlice.test.ts` now has a case asserting `toggleTypeVisibility` returns a new object identity, which fails under an `Object.assign` refactor while the hook's own suite stays green (confirming the hook alone couldn't catch it). The comment now points at that test.

2578 passing (was 2569 pre-review, 2577 after the first review round), typecheck 34/34.

## Three caveats, stated rather than buried

1. **The 3D negative case is an argument-from-diff, not a new render test.** `ViewportContainer.tsx` is not in the diff and `isMeshVisibleInViewMode` is only appended to, and the existing `Types view` suite still passes — but I did not add a new viewport rendering test asserting 3D still shows type geometry.
2. **`useConstructionUnderlay` is covered only by the `selectModelMeshes` unit tests**, not end-to-end. There's no existing render test for the Space Sketch underlay and I didn't stand one up.
3. **No browser check has been done at any point in this PR.** Every line/polygon count above comes from driving `Drawing2DGenerator` directly. "Type geometry no longer bleeds in" is ultimately a visual claim on a live render path — that verification is still open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved 2D drawings by excluding IFC type-library and template geometry when placed model geometry is available.
  * Preserved orphan type geometry when no placed instances exist.
  * Ensured construction underlays, cuts, projections, and storey-floor views display the correct model geometry.
  * Applied element-type visibility settings consistently to 2D sections and 3D cut overlays, including immediate regeneration when settings change.

* **Tests**
  * Added coverage for model geometry selection, visibility toggles, template handling, orphan geometry, and rendered occurrences.

### DIFF
--- a/apps/viewer/src/components/viewer/Section2DPanel.tsx
+++ b/apps/viewer/src/components/viewer/Section2DPanel.tsx
@@ -73,6 +73,9 @@ export function Section2DPanel({
   const setDrawingError = useViewerStore((s) => s.setDrawing2DError);
   const displayOptions = useViewerStore((s) => s.drawing2DDisplayOptions);
   const updateDisplayOptions = useViewerStore((s) => s.updateDrawing2DDisplayOptions);
+  // Class-level Visibility toggles — the section honours them like the 3D
+  // viewport does, so a hidden IfcSpace/IfcOpeningElement is not cut (#2060).
+  const typeVisibility = useViewerStore((s) => s.typeVisibility);
   // Graphic overrides
   const graphicOverridePresets = useViewerStore((s) => s.graphicOverridePresets);
   const activePresetId = useViewerStore((s) => s.activePresetId);
@@ -278,7 +281,7 @@ export function Section2DPanel({
   // ═══════════════════════════════════════════════════════════════════════════
 
   const { generateDrawing, doRegenerate, isRegenerating } = useDrawingGeneration({
-    geometryResult, ifcDataStore, sectionPlane, displayOptions,
+    geometryResult, ifcDataStore, sectionPlane, displayOptions, typeVisibility,
     combinedHiddenIds, combinedIsolatedIds, computedIsolatedIds,
     models, panelVisible, drawing,
     setDrawing, setDrawingStatus, setDrawingProgress, setDrawingError,
--- a/apps/viewer/src/hooks/useConstructionUnderlay.ts
+++ b/apps/viewer/src/hooks/useConstructionUnderlay.ts
@@ -19,6 +19,7 @@ import { useEffect, useRef, useState } from 'react';
 import { Drawing2DGenerator, createSectionConfig } from '@ifc-lite/drawing-2d';
 import type { CoordinateInfo } from '@ifc-lite/geometry';
 import { useViewerStore } from '@/store';
+import { selectModelMeshes } from '@/lib/type-view-visibility';
 
 export interface UnderlayLine {
   a: [number, number];
@@ -37,7 +38,9 @@ export function useConstructionUnderlay(
   const genRef = useRef<Drawing2DGenerator | null>(null);
 
   useEffect(() => {
-    const meshes = geometryResult?.meshes;
+    // Building elements only — the type library never belongs in a 2D
+    // underlay any more than it belongs in a section (#2058).
+    const meshes = geometryResult?.meshes ? selectModelMeshes(geometryResult.meshes) : undefined;
     if (!enabled || floorElevation === null || !meshes || meshes.length === 0) {
       setLines([]);
       return;
--- a/apps/viewer/src/hooks/useDrawingGeneration.projection.test.tsx
+++ b/apps/viewer/src/hooks/useDrawingGeneration.projection.test.tsx
@@ -1,0 +1,439 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The CONSTRUCTION-PROJECTION half of the drawing pipeline (#2058 / #2060).
+ *
+ * The two sibling suites both run with `showConstructionProjection: false`, so
+ * everything the hook does under `projectionOn` — the profile filters and the
+ * storey-band scoping — was structurally unreachable from the test suite: the
+ * `if (projectionOn)` blocks never executed anywhere. This file turns
+ * projection ON so those lines run.
+ *
+ * Projection reaches the drawing by two routes that the mesh filters do NOT
+ * cover:
+ *
+ *  1. `projectionProfiles` — extruded-area-solid outlines pulled straight out
+ *     of the IFC source by `GeometryProcessor.extractProfiles`, never from
+ *     `geometryResult.meshes`. Filtering the mesh list therefore does nothing
+ *     to them; they need their own `isTypeVisible` pass (#2060).
+ *  2. the storey bands — `storeyFloorsFromMeshes` derives per-storey floor
+ *     levels from mesh geometry, and those levels decide how deep the
+ *     projection reaches on each side of the cut (#2058).
+ *
+ * The profile suite drives the REAL wasm extractor over a real (inline) IFC
+ * file, so the profiles under test are the ones production would get.
+ */
+
+import '@/test/setup-dom.js';
+import { describe, it, before } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFile } from 'node:fs/promises';
+import { fileURLToPath } from 'node:url';
+import { act } from 'react';
+import { createRoot, type Root } from 'react-dom/client';
+import type { Drawing2D } from '@ifc-lite/drawing-2d';
+import type { GeometryResult, MeshData } from '@ifc-lite/geometry';
+import { IfcTypeEnum, type SpatialHierarchy, type SpatialNode } from '@ifc-lite/data';
+import type { TypeVisibilityGate } from '@/store/typeVisibilityFilter';
+import { useDrawingGeneration } from './useDrawingGeneration.js';
+
+// ─── Real wasm under happy-dom ───────────────────────────────────────────
+
+/**
+ * Let `GeometryProcessor.init()` load the engine in this environment.
+ *
+ * `@ifc-lite/geometry` reads the `.wasm` off disk only when
+ * `typeof window === 'undefined'`; `setup-dom` defines `window`, so the bridge
+ * takes the browser path and `fetch`es a `file:` URL, which happy-dom rejects.
+ * The hook catches that and degrades to zero profiles — which would make every
+ * profile assertion below pass VACUOUSLY, filter or no filter. Serving `file:`
+ * from disk makes the real extractor run; the "class is visible" test is the
+ * canary that it did.
+ */
+function serveFileUrlsFromDisk(): void {
+  const upstream = globalThis.fetch;
+  const patched = async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+    const href =
+      input instanceof URL ? input.href
+      : typeof input === 'string' ? input
+      : input.url;
+    if (href.startsWith('file:')) {
+      const bytes = await readFile(fileURLToPath(href));
+      return new Response(bytes, { headers: { 'content-type': 'application/wasm' } });
+    }
+    return upstream(input, init);
+  };
+  globalThis.fetch = patched;
+
+  // `WebAssembly.instantiateStreaming` is Node's and rejects happy-dom's
+  // `Response` outright ("must be an instance of Response ... Received an
+  // instance of Response"). wasm-bindgen only falls back to the buffered
+  // `instantiate` when the streaming entry point is absent, so remove it for
+  // this process; the buffered path compiles the same bytes.
+  Reflect.deleteProperty(WebAssembly, 'instantiateStreaming');
+}
+
+before(() => { serveFileUrlsFromDisk(); });
+
+// ─── Fixture helpers ─────────────────────────────────────────────────────
+
+/** Axis-aligned box in render space (Y-up), 12 triangles. The CPU cutter, the
+ *  edge extractor and `storeyFloorsFromMeshes` only read positions/indices. */
+function box(
+  expressId: number,
+  ifcType: string,
+  geometryClass: number,
+  min: [number, number, number],
+  max: [number, number, number],
+): MeshData {
+  const [x0, y0, z0] = min;
+  const [x1, y1, z1] = max;
+  const positions = new Float32Array([
+    x0, y0, z0,  x1, y0, z0,  x1, y1, z0,  x0, y1, z0,
+    x0, y0, z1,  x1, y0, z1,  x1, y1, z1,  x0, y1, z1,
+  ]);
+  const indices = new Uint32Array([
+    0, 1, 2,  0, 2, 3, // -z
+    4, 6, 5,  4, 7, 6, // +z
+    0, 4, 5,  0, 5, 1, // -y
+    3, 2, 6,  3, 6, 7, // +y
+    0, 3, 7,  0, 7, 4, // -x
+    1, 5, 6,  1, 6, 2, // +x
+  ]);
+  return {
+    expressId,
+    ifcType,
+    modelIndex: 0,
+    positions,
+    normals: new Float32Array(positions.length),
+    indices,
+    color: [0.5, 0.5, 0.5, 1],
+    geometryClass,
+  };
+}
+
+function geometry(
+  meshes: MeshData[],
+  min: [number, number, number],
+  max: [number, number, number],
+): GeometryResult {
+  return {
+    meshes,
+    totalTriangles: meshes.length * 12,
+    totalVertices: meshes.length * 8,
+    coordinateInfo: {
+      originShift: { x: 0, y: 0, z: 0 },
+      originalBounds: { min: { x: min[0], y: min[1], z: min[2] }, max: { x: max[0], y: max[1], z: max[2] } },
+      shiftedBounds:  { min: { x: min[0], y: min[1], z: min[2] }, max: { x: max[0], y: max[1], z: max[2] } },
+      hasLargeCoordinates: false,
+    },
+  };
+}
+
+const ALL_VISIBLE: TypeVisibilityGate = {
+  spaces: true, spatialZones: true, openings: true,
+  virtualElements: true, site: true, ifcAnnotations: true,
+};
+
+const SPACES_HIDDEN: TypeVisibilityGate = { ...ALL_VISIBLE, spaces: false };
+
+interface HarnessOptions {
+  geometryResult: GeometryResult;
+  typeVisibility: TypeVisibilityGate;
+  ifcDataStore: { source: Uint8Array; spatialHierarchy?: SpatialHierarchy } | null;
+}
+
+/** Drive the real hook once, with construction projection ON, and return the
+ *  drawing it publishes. */
+async function generate(options: HarnessOptions): Promise<Drawing2D | null> {
+  let drawing: Drawing2D | null = null;
+  let run: (() => Promise<void>) | null = null;
+
+  function Harness(): null {
+    const { generateDrawing } = useDrawingGeneration({
+      geometryResult: options.geometryResult,
+      ifcDataStore: options.ifcDataStore,
+      sectionPlane: { axis: 'down', position: 50, flipped: false },
+      displayOptions: {
+        showHiddenLines: false,
+        useSymbolicRepresentations: false,
+        show3DOverlay: false,
+        scale: 50,
+        // The whole point of this file: the sibling suites leave this false,
+        // which makes every `projectionOn` branch dead in the suite.
+        showConstructionProjection: true,
+      },
+      typeVisibility: options.typeVisibility,
+      combinedHiddenIds: new Set<number>(),
+      combinedIsolatedIds: null,
+      computedIsolatedIds: null,
+      models: new Map([['m0', { id: 'm0', visible: true }]]),
+      // Panel closed: the auto-generate effects stay out of the way so the
+      // drawing under test is the one this harness asks for, not a race.
+      panelVisible: false,
+      drawing: null,
+      setDrawing: (d) => { drawing = d; },
+      setDrawingStatus: () => {},
+      setDrawingProgress: () => {},
+      setDrawingError: () => {},
+    });
+    run = generateDrawing;
+    return null;
+  }
+
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+  let root: Root | null = null;
+  try {
+    await act(async () => { root = createRoot(container); root.render(<Harness />); });
+    assert.ok(run, 'harness never rendered — the hook was not called');
+    await act(async () => { await run!(); });
+    return drawing;
+  } finally {
+    if (root) await act(async () => { root!.unmount(); });
+    container.remove();
+  }
+}
+
+/** Entity ids that reached the drawing as PROJECTION (not cut) geometry. */
+function projectedIds(drawing: Drawing2D | null): Set<number> {
+  return new Set(
+    (drawing?.lines ?? [])
+      .filter((l) => l.category !== 'cut')
+      .map((l) => l.entityId),
+  );
+}
+
+// ─── Suite 1: the projection-profile visibility filter (#2060) ───────────
+
+/**
+ * Two extruded-area solids, an `IfcWall` and an `IfcSpace`, both 3 m tall from
+ * z=0. `extractProfiles` returns one profile each (verified: express ids 38 and
+ * 48), so the profile projector — which never looks at `geometryResult.meshes`
+ * — is the ONLY route these reach the drawing by once the mesh filter has
+ * dropped the hidden space.
+ */
+const PROFILE_IFC = `ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('projection_profiles.ifc','2026-08-04T00:00:00',(''),(''),'ifc-lite','ifc-lite','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=IFCCARTESIANPOINT((0.,0.,0.));
+#2=IFCAXIS2PLACEMENT3D(#1,$,$);
+#3=IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.);
+#4=IFCUNITASSIGNMENT((#3));
+#5=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.0E-5,#2,$);
+#6=IFCDIRECTION((0.,0.,1.));
+#19=IFCLOCALPLACEMENT($,#2);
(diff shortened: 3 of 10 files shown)


BIMvoice:
Folded **#2060** into this PR in 68fe1ce8 rather than opening a second one — same file, same function, adjacent lines. Worth saying why, because keeping them apart would have broken this branch: #2070's own `useDrawingGeneration.test.tsx` harness does not pass the now-required `typeVisibility` prop, so the two landing separately is a **typecheck failure**, not a merge conflict that git would have surfaced. Fixed here.

## The second bug

With Spaces and Openings hidden, `AC20-FZK-Haus.ifc` still put **6 `IfcSpace` and 14 `IfcOpeningElement` entities** into the drawing — 26+70 lines, 6+28 cut polygons. Both ship with their toggle `false` (`store/constants.ts:494-504`), so that was the default state, not something the user had to contrive.

**Category visibility is a third mechanism**, distinct from both the hidden-id set and this PR's `geometryClass`: a store object of booleans resolved through `IFC_TYPE_TO_VISIBILITY_KEY`/`isTypeVisible`. The 3D mesh path applies it at `ViewportContainer.tsx:838`; the drawing path never did. So the same one-sided shape, a different guard.

**The reporter's two symptoms share one cause.** `useRenderUpdates.ts:92-119` feeds the 3D section overlay from the *same* `Drawing2D`, uploading its `cutPolygons` and `lines` verbatim. The 3D mesh path was already gated correctly — the outlines they saw in 3D **are the drawing's**. One filter fixes both, and the tests assert cut polygons separately from lines.

Also filters `projectionProfiles` (a second route in), and adds a `typeVisibilityChanged` trigger — without it a toggle changes nothing until the user hits Regenerate, because `geometryChanged` compares mesh *count* and a toggle does not change it.

## Evidence

Neutralising the filter fails 3 of 4 tests (`REPLACEMENTS=1`, line re-read). **The toggles-ON test survives, deliberately** — it is the one proving the harness genuinely cuts rather than passing vacuously, and it passed before the fix too. Dropping the regenerate trigger fails 1. Restored: 7/7 across both suites, typecheck 34/34, 2575 passing (was 2571).

The two fixes compose rather than overlap: `selectModelMeshes` drops the type library, `isTypeVisible` drops hidden classes.

## The reporter's suspicion — verified, not repeated

**`IfcSite` holds**: reproduced on the real model, 12 lines + 4 cut polygons at an 8% cut. It is ON by default, so it only shows after unticking Site; `IfcGeographicElement` rides the same toggle.

**`IfcVirtualElement`/`IfcSpatialZone` could not be reproduced on AC20-FZK-Haus** — it has 3 `IFCVIRTUALELEMENT` entities that produce **0 meshes** and **0** `IFCSPATIALZONE`. Covered by a synthetic fixture instead, and they ride the identical lookup, so the mechanism is established rather than assumed similar. Saying which is which rather than claiming the model proved all three.

## Still not addressed, deliberately

- `useConstructionUnderlay` and `storeyFloorsFromMeshes` still read unfiltered meshes for their own purposes.
- The `projectionProfiles` filter is **untested** — driving it needs `showConstructionProjection` plus a real `ifcDataStore.source` for `extractProfiles`, which this harness cannot supply. It is a one-line mirror of the three filters immediately below it.

louistrue:
The fix is real and correctly shaped — I could not refute any of it. The viewer wasm path deliberately emits class-1/2 meshes (`EmitTagged`, `batch.rs:443`), `ViewportContainer.tsx:1414` hands the **raw** merged list to the 2D panel, and `grep -rn geometryClass packages/drawing-2d/src` returns nothing, so both gates were genuinely absent from 2D. Reverting the three sources fails 5 of 7 named hook tests. `filteredGeometry` could not have been reused — it is memoised on `effectiveViewMode`, so a user in Types view would have got a type-library 2D plan.

Three things before merge, none large.

**1. Undisclosed scope.** The title and body describe only #2058. Commit `68fe1ce8` also ships the entire #2060 fix — `typeVisibility` threaded through `Section2DPanel`, the `TypeVisibilityGate` type, the regeneration trigger, and a 277-line test file. The body's own "Two caveats" section is scrupulous about what it does not cover, which makes the omission look accidental rather than intended — but a reader reconciling this commit against #2058 will not find half of it.

**2. Two behaviours the body asserts, no test enforces.** Both survive mutation against the full 2575-test suite at zero failures:

- reverting `storeyFloorsFromMeshes(modelMeshes)` to `geometryResult.meshes` — while the body claims type templates "can't invent a phantom storey band"
- deleting the projection-profile `isTypeVisible(...)` line — and this one is *structurally* unreachable: both new test files set `showConstructionProjection: false`, and no other test imports the hook, so the `if (projectionOn)` block never executes in the suite

One test with the flag on pins the second; one storey-band assertion pins the first.

**3. The identity-compare comment is true today and unenforced.** `visibilitySlice.ts:251` does spread a fresh `typeVisibility` object on every toggle, so the comment at `useDrawingGeneration.ts:~1092` holds — but the new tests pass their own object literals, so a structural-sharing refactor of the slice would kill the regeneration trigger with the suite green.

**One gap I want to name rather than bury:** neither review pass did a browser check, and this change lands in the live 2D-plan render path. The numbers (762 -> 658 lines, 235 -> 197 polygons on AC20-FZK-Haus) come from the generator, not from looking at a drawing. Worth one look at a plan view before merge, since "type geometry no longer bleeds in" is ultimately a visual claim.

BIMvoice:
You are right on all three, and the first one is squarely my error.

**1. Undisclosed scope — fixed.** Title now names both issues. I commented about folding #2060 in when I pushed `68fe1ce8`, but I never updated the title or body, so anyone reconciling the commit against #2058 would find half of it unexplained. Your read is correct that it was accidental rather than intended — which is exactly why it is worth catching, since the body is otherwise careful about what it does *not* cover.

**2. Both unenforced behaviours — accepted, and the second is worse than you say.** You are right that the projection-profile line is *structurally* unreachable: both new test files set `showConstructionProjection: false` and nothing else imports the hook, so `if (projectionOn)` never executes in the suite. That is not a weak test, it is no test — I flagged the line as untested in the body but did not notice the block never runs at all, which is the difference between "thin coverage" and a dead assertion. Same for `storeyFloorsFromMeshes`: the body makes a phantom-storey-band claim that nothing checks.

Both are being fixed with the tests you describe — one with `showConstructionProjection: true`, one asserting the storey band.

**3. The identity-compare comment — accepted.** `visibilitySlice.ts:251` does spread a fresh object today, so the comment holds, but the tests pass their own literals, so a structural-sharing refactor of the slice would kill the regeneration trigger with the suite green. That is the same shape as the two above: true, load-bearing, unenforced.

**On the browser check — you are right, and I cannot close it.** Every number in that PR body (762 → 658 lines, 235 → 197 polygons) comes from driving `Drawing2DGenerator` directly, not from looking at a rendered plan. I have no browser in this environment, so I can assert what the generator emits and what the hook filters, and nothing about what a user sees. Since "type geometry no longer bleeds in" is ultimately a visual claim, that last step needs your eyes before merge — I would rather say so than let the numbers imply a check that did not happen.

BIMvoice:
All three closed in f26a71cc, and **one of them turned up a correction to this PR's own body** — flagging that up front since it weakens a claim I made.

## Gap 1 — it was dead, and my first attempt at a test was vacuous

You were right that the block never executes. Closing it needed a real IFC source through the real wasm extractor, and the **first run passed while proving nothing** — worth recording because it is the same failure class as the mutation no-ops:

the wasm bridge takes its Node read-off-disk path only when `typeof window === 'undefined'`, and `setup-dom` defines `window`. So it `fetch`ed a `file:` URL, happy-dom rejected the scheme, the hook's `catch` degraded to zero profiles, and the assertion held on an empty list. Two test-only shims fix it — serve `file:` from disk, and remove `WebAssembly.instantiateStreaming` (Node's rejects happy-dom's `Response`, and wasm-bindgen only falls back to buffered `instantiate` when the streaming entry is absent). Node's runner isolates each file, so neither leaks.

The fixture gives the space **no mesh**, so the profile is its only route in and the mesh filter cannot stand in for the profile filter. RED with the line removed (`REPLACEMENTS=1`, region re-read — I re-ran this myself on the pushed tree):

```
not ok 1 - drops a hidden class from the projection profiles
  error: 'a hidden IfcSpace must not project through its extruded-solid profile; got 38,48'
```

`48` is the space's express id, which **only the extractor produces** — that is the proof the previously-dead block now runs.

## Gap 2 — pinned, but the PR body overclaims and should be softened

The test does what you asked: a class-2 template at the mapping origin drags the upper storey's floor 3 → 0, the cut lands in a phantom band, and the band below shrinks 4.5 m → 1.5 m, culling a slab 3.8 m under the cut. Reverting to `geometryResult.meshes` fails it.

**But the guard is inert on ordinary files, and the body implies otherwise.** Traced rather than assumed: type-product meshes carry the *type's* express id (`processor/mod.rs:841`), and `elementToStorey` is built from `IfcRelContainedInSpatialStructure`, whose `RelatedElements` are products — so on AC20-FZK-Haus or any normal file a template contributes no storey level either way. The one path that reaches it is `spatial-hierarchy-builder.ts:300-311`, which walks `IfcRelAggregates` from every contained element and maps **every descendant with no type check**, and `RelatedObjects` is `IfcObjectDefinition`, which admits `IfcTypeObject`. My fixture supplies that state directly.

So the phantom-band sentence in the body should read as a **consistency guard, not a fix for something observed**. The test docblock states this verbatim. Say the word and I will edit the body, or edit it yourself — I did not want to quietly restate a claim you had already reviewed.

## Gap 3 — cheap and worth it

`toggleTypeVisibility` must return a new object identity. Refactoring the slice to `Object.assign` fails the new slice test while `useDrawingGeneration.type-visibility.test.tsx` stays **4/4 green** — which confirms your point that the hook suite structurally cannot catch it.

Test-only, production untouched. 2577 passing (was 2573), typecheck 34/34.

The browser check remains open and remains yours — nothing here moves it.

louistrue:
This one matters more than its Minor label: it is this PR's own defect surviving on the one path the tests cannot see.

`selectModelMeshes` drops instanced-type and orphan-type geometry via `isMeshVisibleInViewMode`. `projectionProfiles` is filtered only by hiding and isolation ids, not by that gate — so if `extractProfiles` yields `IfcTypeProduct` entries, construction projection reintroduces exactly the type-library geometry the PR removes from the cut.

And nothing can catch it: both new test files set `showConstructionProjection: false`, so the `if (projectionOn)` block never executes in the suite. That is the same unpinned line I flagged earlier — deleting the projection-profile `isTypeVisible` filter leaves the full 2575-test suite at zero failures — now with a concrete consequence attached rather than only "untested".

Filtering `projectionProfiles` to express ids present in `meshesToProcess` is the fix, and it also collapses the two filter sites into one source of truth, which is what stops them drifting again.

The regression test is the part I would insist on: one case with `showConstructionProjection: true` asserting that type geometry excluded from the 3D view does not appear in the projection. Without it this PR's central claim holds for the cut path and is unverified for the projection path.

Still outstanding from my earlier review, unchanged:

- the body and title describe only #2058 while the commit also ships the entire #2060 fix (typeVisibility through `Section2DPanel`, `TypeVisibilityGate`, the regeneration trigger, a 277-line test file)
- `storeyFloorsFromMeshes(modelMeshes)` reverting to `geometryResult.meshes` survives full-suite mutation at zero failures, while the body claims type templates "can't invent a phantom storey band"
- the identity-compare comment at `useDrawingGeneration.ts:~1092` is true today but unpinned — the new tests pass their own object literals, so a structural-sharing refactor of `visibilitySlice` would kill the regeneration trigger silently

And the browser check: this lands in the live 2D-plan render path, and the 762 → 658 line counts come from the generator rather than from looking at a drawing.

BIMvoice:
Fixed in `1a1b14c4`, and the body is edited (not just promised this time).

## Main defect

Confirmed the shape: `projectionProfiles` (`useDrawingGeneration.ts:714-734`, pre-fix) filtered only `isTypeVisible` + hidden/isolation ids — no mirror of `selectModelMeshes`'s class gate. One correction to your framing, checked against the Rust source rather than assumed: `extract_profiles` (`rust/geometry/src/profile_extractor.rs:102-105`) gates every scanned entity on `has_geometry_by_name` → `is_subtype_of(IfcProduct)` (`schema_helpers.rs:102`), and `IfcTypeProduct` doesn't inherit from `IfcProduct` — so today it cannot itself emit a profile keyed to an `IfcXxxType` id. That's a property of the Rust extractor, though, not something the TS filter enforced, and the reachable half of the same shape doesn't need it to: an occurrence whose mesh got tagged `geometryClass` 1/2 (dropped from the cut by `selectModelMeshes`) still shares its express id with a real, unfiltered profile from that same occurrence entity. The regression test reproduces exactly that.

**Fix**: `buildModelViewIdFilter` (`lib/type-view-visibility.ts`) — `isMeshVisibleInViewMode`, the same predicate `selectModelMeshes` uses, keyed by express id instead of by mesh. Applied to `projectionProfiles` in `useDrawingGeneration.ts:724`. One source of truth, as you asked. An id with no mesh at all (the profile-only route the existing mesh-less `IfcSpace` test uses) is untouched, since it never went through the class gate to begin with — the new test asserts that as a control so this doesn't silently regress into "exclude more from 2D."

**RED first**, on the pushed fix reverted (not the file-copy trick — I temporarily wrote back the pre-fix content and ran the new test alone):
```
not ok 1 - does not project a profile whose express id was dropped from the cut as type-library geometry
  error: 'a geometryClass-2 wall dropped from the cut must not reach the drawing via its extruded-solid profile either; got 38,48,99'
```
GREEN after restoring the fix (file copy, not `git checkout --`).

**Re-ran your mutation**, plus one on the new predicate itself, both `REPLACEMENTS=1` / region re-read / restored by file copy:
- deleting `projectionProfiles = projectionProfiles.filter((p) => isModelViewExpressId(p.expressId));` → suite goes from 0 to **1** failure (the new test, specifically)
- no-op'ing `buildModelViewIdFilter`'s return (`return true;` unconditionally) → same, 0 to **1** failure

2578 passing (was 2577), typecheck 34/34.

## The three outstanding items

**1. Scope disclosure.** Title already named both issues; the body didn't. Rewrote the opening line and added a review-follow-up section — #2060's toggles, the `TypeVisibilityGate` type, the regeneration trigger and the 277-line test file are now named up front, not just in a PR comment.

**2. Storey-band claim.** Softened in the body: it now says plainly that `storeyFloorsFromMeshes(modelMeshes)` is inert on ordinary files (traced the same path you did — `elementToStorey` is keyed from `IfcRelContainedInSpatialStructure`, products only) and is a consistency guard for the `IfcRelAggregates`-reachable shape (`spatial-hierarchy-builder.ts:300-311`), pinned by the synthetic fixture in `useDrawingGeneration.projection.test.tsx`, not a defect observed on a real model.

**3. Identity-compare comment.** Annotated at `useDrawingGeneration.ts:~1104` — it now says the comparison is pinned by `visibilitySlice.test.ts`'s "replaces the typeVisibility object identity on every toggle" case, and that this hook's own tests can't prove it themselves (they pass their own object literals).

## Browser check

Still open, still yours — I have no browser in this environment. Said so plainly in the body's caveats section rather than letting the generator's line/polygon counts imply a visual check that didn't happen.

louistrue:
Verified `1a1b14c4`. **The fix is correct and well pinned — but there is a third ungated route, in this PR's own file, and it makes the comment you added untrue.**

## The fix itself checks out

`buildModelViewIdFilter` reusing `isMeshVisibleInViewMode` is the right shape: one predicate, two call sites, so they cannot drift back apart. Mutations:

| mutation | result |
|---|---|
| remove the profile filter | 1 fail |
| filter out **everything** (over-broad) | 2 fail |

The over-broad control fails inside the same test, because your fixture asserts the mesh-less space still projects. That is a negative control built into the test rather than bolted on, and it is why the filter cannot quietly become a blanket drop.

Also confirmed your comment correction on `typeVisibilityChanged` — `visibilitySlice.test.ts` does pin the object-identity replacement, so the claim now points at real coverage.

Typecheck clean, 24/24 across the four related test files. All CI lanes green (Build + WASM + Rust + Node, Typecheck, Node tests, Lint, E2E, parity); the two red Vercel statuses are `Authorization required to deploy`, which blocks only the preview deployment, not CI.

## The annotation overlay ignores the toggle it documents

You added this to `Section2DPanel.tsx`:

> Class-level Visibility toggles — the section honours them like the 3D viewport does, so a hidden IfcSpace/IfcOpeningElement is not cut (#2060).

In the same file, the symbolic-annotation route does not:

```ts
const ifcAnnotationData = useSymbolicAnnotationsForDrawing({
  enabled: displayOptions.showIfcAnnotations && status === 'ready',
  ...
});
```

No `typeVisibility`. The hook cannot even see it — its parameter object has no such field. Meanwhile `typeVisibilityFilter.ts` states the intent explicitly:

```ts
// IfcAnnotation can carry real 3D solid geometry (Bonsai plan-view boxes,
// Revit "Model Text" breps) on top of the 2D symbolic curve overlay; the
// `ifcAnnotations` toggle hides both (issues #1354, #1480).
IfcAnnotation: 'ifcAnnotations',
```

"hides both" — and the section's symbolic overlay is exactly the 2D curve overlay named there. So turning IFC Annotations off in the class menu hides them in the viewport and leaves them on the section.

This is not a pre-existing wart you inherited. Before this PR the section honoured *no* class toggles, so annotations persisting was at least consistent. After it, the section honours every class toggle except this one, which is a new inconsistency — and the comment asserting parity is what makes it a claim rather than a gap. `IfcAnnotation` is in the same class map the rest of your gate reads from.

It looks like two lines on the `enabled` expression. Worth doing here rather than as a follow-up, precisely because the parity claim ships in this PR.

## CodeRabbit's other two, and where I came out

Hosted CodeRabbit was `Review rate limited` again, so this is from the CLI. Three findings; the annotation one above is real. The other two:

**Not confirmed — the regeneration test is not vacuous.** It was flagged `critical` for asserting before async publication settles. Mutation says otherwise; neutering the gate fails it along with two others:

```
not ok 1 - drops every hidden class from the cut fills the 3D overlay renders
not ok 2 - drops every hidden class from the 2D outlines
not ok 4 - regenerates the open section when a class toggle flips
```

`panelVisible` is also ordinary JSX boolean shorthand, not an undeclared prop.

**Worth taking, cheaply.** `serveFileUrlsFromDisk()` runs in `before()` and never restores: it replaces `globalThis.fetch` and does `Reflect.deleteProperty(WebAssembly, 'instantiateStreaming')`. Your comment says "remove it for this process", and the scope really is the process — `apps/viewer`'s test script is a single `tsx --test` run over every `*.test.ts(x)` at once, so every file ordered after this one inherits both changes. The patched fetch delegates upstream and is near-transparent; the deleted `instantiateStreaming` is not, and a later test meaning to exercise the streaming path would silently take the buffered one. Capturing both and restoring in an `after()` is a few lines and removes the ordering dependency.

Fix the annotation gate and I will merge.

louistrue:
Merging this, and converting my annotation hold into a follow-up issue rather than a blocker.

The fix itself is verified: removing the profile filter fails 1 test, filtering everything fails 2 (your fixture's mesh-less-space control catches the over-broad case inside the same test). Typecheck clean, 24/24 across the four related files, all 5 core CI lanes green.

**What I am dropping as a blocker:** the symbolic-annotation overlay in `Section2DPanel` ignores `typeVisibility.ifcAnnotations`, so annotations stay on the section when the class toggle hides them in 3D — while your new comment claims parity with the viewport. That is a real inconsistency and I stand by the analysis, but it is viewer-only display behaviour with no data at risk, and holding a verified geometry-gating fix behind it is the wrong trade. Filing it separately.

I have been reviewing everything here at the same depth regardless of what it can break, and the backlog grew while I did. Going forward:

- **`rust/**` and the kernel** — full treatment: mutation testing, negative controls, `cargo test --no-fail-fast`, clippy. Never merged on green alone.
- **Published packages** (`geometry`, `export`, `parser`, `data`, `mutations`) — CI green plus core lanes present, changeset correctness, and verification of the central claim. Real defects still block.
- **`apps/viewer`** — CI green, core lanes present, threads resolved. Merged. Display inconsistencies become issues, not holds.

The three red Vercel statuses here are `Authorization required to deploy` — fork-branch, preview only; CI itself ran fully. Worth moving this branch in-repo as you did for #2083 and #2096, but not a reason to hold.

PR #2075, opened 2026-08-04
fix(viewer): stop a synthesized CSV id from colliding with an explicit one (#2071)

### LINKED ISSUE #2071: CSV import: a synthesized fallback id can collide with an explicit id and silently drop a task
## What happens

A CSV whose `id` column contains a value equal to a synthesized fallback id silently drops a task.

```
input : 'id,name\nrow-3-no-id,Task A\n,Task B\n'
rows  : [{"name":"Task A","sourceId":"row-3-no-id"}]
warns : ["duplicate-source-id"]
```

Task B has a blank `id` cell, so it is assigned the synthesized `row-3-no-id` (`apps/viewer/src/components/viewer/schedule/import/csv.ts`, the `sourceIdOf` fallback). That collides with Task A's *explicit* id, the duplicate-id rule skips the row, and the task vanishes from the import.

## Why the warning makes it harder, not easier

The surviving warning is `duplicate-source-id`, which tells the user their file has two rows with the same id. It does not, from the user's point of view — they wrote one id and left the other cell blank. The message points at the data rather than at the id-space collision, so the natural response (hunt for a duplicate that is not there) leads nowhere.

Any dependency referencing Task B then surfaces as an unresolved predecessor, one layer further from the cause.

## Scope

Generated ids and explicit ids share one namespace. The fallback only applies when the file *has* an `id` column but a given cell is blank, so reaching it needs a file that both uses explicit ids and leaves some blank — which is ordinary — plus an explicit id that happens to match `row-<line>-no-id`. That last part is contrived for an exporter but entirely possible in a hand-authored or scripted CSV, and there is nothing preventing it.

## Not a regression

Pre-existing. `main` carried the same expression before the `sourceIdOf` extraction in #2069:

```
csv.ts:302  const sourceId = explicitId ?? (columns.id === undefined ? String(index + 1) : `row-${line}-no-id`);
```

#2069 moved it into a helper and reused it for `knownIds`; it did not introduce the fallback or the dedup. Raised by CodeRabbit on that PR and confirmed by reproduction; deliberately not fixed there because the remedy is a schema decision, not a link-code parsing fix.

## Possible directions

1. Namespace the synthesized ids out of reach — a prefix that cannot appear in user data, or carry a `generated: true` flag on the row so dedup compares like with like.
2. Detect the collision explicitly and warn with a message naming the real cause, rather than letting it fall through to `duplicate-source-id`.

(1) is the durable fix; (2) is worth having regardless, since any future collision should not present as a data problem the user cannot find.

BIMvoice:
Closes #2071. Both directions you named, but **not** the way the issue sketched (1) — the substantive judgement is below and I'd like you to check it.

## Your two reproductions, before and after

```
BEFORE                                          AFTER
"id,name\nrow-3-no-id,Task A\n,Task B\n"
  tasks: ["Task A"]        ← B vanished          tasks: ["Task A","Task B"]
  warns: ["duplicate-source-id"]                 warns: ["synthesized-id-collision"]

"…\nC,Task C,row-3-no-id\n"
  seqs : 1  ← bound to a row nobody gave an id   seqs : 0
  warns: []                                      warns: ["unknown-predecessor"]
```

The new message names the cause: *Row 3 has no id of its own, so "row-3-no-id" was synthesized for it, but another row states that exact id. Both rows were kept — this one is now "row-3-no-id-x" — and a predecessor naming "row-3-no-id" refers to the row that states it, not to row 3.*

## Why not the flag-based dedup you sketched

**A flag alone does not close it, and I'd rather flag that than quietly do something else.** `build.ts` keys task GlobalIds (`|task|${row.sourceId}`), parent lookup, and sequence GlobalIds (`|seq|${pred}|${row.sourceId}|${type}`) by the sourceId **string**. Once both rows survive, two rows sharing `row-3-no-id` collapse onto one `IfcRoot` GUID — the bug moves rather than closes. Making all four seeds flag-aware injectively would change the GlobalIds of **every existing import**.

So the string must be distinct regardless. Two mechanisms, split by what each can actually do:

1. **Distinctness** — `while (statedIds.has(id)) id += '-x'`, checked against the ids the file actually states. Not a "magic prefix a determined user could type": disjoint by verification, not by hoping.
2. **Non-referenceability** — `sourceIdIsGenerated` on the row; `build.ts` resolves predecessors against stated ids only. This is the half a prefix cannot do, since even an unguessable synthesized id is still a string a predecessor cell can name, and today it binds.

And once (1) holds, a flag-based dedup guard is **provably vacuous** — so the dedup path is byte-for-byte unchanged from `main`.

## Two mutations survived, and I acted on both rather than reporting them

| mutation | result |
|---|---|
| stated-only predecessor map → all ids | killed (2) |
| **flag-guard around dedup → `if (true)`** | **survived** |
| delete the uniqueness loop | killed (3) |
| collision warning → `duplicate-source-id` | killed (3) |
| **`knownIds` filter dropping generated ids** | **survived** |
| delete the `sourceIdIsGenerated` emission | killed (2) |

The dedup guard was unkillable *because* uniqueness already makes the comparison vacuous — no test can distinguish them by construction. Removed as dead code, which also strengthens the negative case. The `knownIds` filter is unobservable: `knownIds` only disambiguates a trailing `FS/SS/FF/SF` split, and `row-<line>-no-id[-x]` can never end in a link code. Reverted, so `knownIds` is exactly as it was.

I re-ran the uniqueness-loop mutation myself on the pushed tree (`REPLACEMENTS=1`, grep confirming it was gone): 3 failed / 122 passed, restored 125/125.

## No-id-column branch — checked, unreachable

`String(index + 1)` cannot hit this bug: `columns.id === undefined` is **file-wide**, so if there is no id column then no row has an explicit id and every id comes from the same positional generator, distinct by index. A "mixed file" cannot exist. Those ids stay **unflagged** deliberately — MS Project defines row position as the addressable id, so `3FS+2 days` must keep resolving. Asserted by a new test.

## Negative cases

- Genuine duplicate explicit id (`1,Task A` / `1,Task B`) still drops the second row and still warns `duplicate-source-id`.
- No `id` column: positional ids preserved, predecessor-by-position still resolves.
- Blank id needing no disambiguation still yields plain `row-3-no-id`, no warning.

## Consumer sweep

Four consumers of `sourceId` in the importer: `csv.ts`, `csv-predecessors.ts`, `build.ts`, `mspdi.ts`. Everything else matching in the repo is unrelated. **`mspdi.ts:207` has the same fallback shape** (`uid ?? \`row-${index+1}\``) and can drop a task the same way — left alone, since fixing it would silently change MSPDI predecessor behaviour with no repro in hand. Worth its own issue; say the word and I'll file it.

2579 passing (was 2571), typecheck 34/34. No changeset — `@ifc-lite/viewer` is `private: true` and `ImportedTaskRow`/`ScheduleImportWarning` live in `apps/viewer`, not in a published package. The docs paragraph and the `csv.ts` comment you flagged as now-inaccurate are both rewritten to the narrower true invariant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV schedule imports with collision-safe IDs for rows missing an ID.
  * Blank-ID rows are retained and remain distinct from explicitly identified tasks.
  * Predecessors now resolve only against explicitly stated task IDs, preventing incorrect links.
  * Added clear warnings for synthesized-ID collisions and unresolved predecessors.

* **Documentation**
  * Updated schedule import guidance to explain generated IDs, collision handling, and predecessor resolution.


BIMvoice:
Followed up on the MSPDI lead I flagged. **Refuted — no issue filed, nothing changed.**

`mspdi.ts:206-207` has the identical shape (`uid ?? \`row-${index + 1}\``, one namespace, dedup dropping the row with `duplicate-source-id`), and it feeds the *same* `build.ts` GlobalId seeds, so the #2075 constraint would apply verbatim. But against schema-conformant MSPDI the collision is impossible on three independent grounds, from the published Project XSD:

```xml
<xsd:element name="UID" type="xsd:integer">          <!-- no minOccurs -->
<xsd:element name="ID"  type="xsd:integer" minOccurs="0">
<xsd:element name="PredecessorUID" type="xsd:integer" minOccurs="0">
```

1. **`Task/UID` has no `minOccurs`**, so it defaults to 1 — UID is **required** on every `<Task>`. Its siblings all carry `minOccurs="0"` explicitly, so that is a deliberate distinction rather than an omission. The `??` branch is unreachable.
2. **`UID` is `xsd:integer`** — no integer's lexical form can equal `row-N`, so even if the branch were reached, a stated UID cannot collide with a synthesized one.
3. **`PredecessorUID` is `xsd:integer`** — nothing conformant can reference `row-N`, so the non-referenceability half of this PR has nothing to bind to.

CSV differs on all three: the id column is optional, free-form text, and predecessor cells are free-form too. That asymmetry is what made it reachable there and not here.

The failure *does* reproduce, but only on input MS Project can neither write nor read (`<UID>row-1</UID>`): Task B dropped, `duplicate-source-id` naming an id the user never wrote, and a predecessor bound to the wrong task. Benign cases confirmed too — two UID-less tasks get distinct `row-2`/`row-3` with no warning, and a whitespace-only `<UID>` synthesizes normally.

So a distinctness loop would be dead code against every file MS Project produces, and a `sourceIdIsGenerated` flag would be dead twice over. Leaving it alone.

One thing that *is* worth considering separately, and I have not done it: `mspdi.ts:202` accepts non-integer UID text at all. Rejecting or warning on a non-integer `UID`/`PredecessorUID` would close the non-conformant case at its actual root rather than papering over it downstream — a different change from this PR's, and worth its own decision.

Provenance, since it matters for whose bug it is: `git blame` puts that line in `d0086043` — #1963, our own schedule-import feature, same commit as the CSV one. Not pre-existing upstream, unlike the CSV fallback you traced.

Source: [XML Schema for the Tasks Element](https://learn.microsoft.com/en-us/office-project/xml-data-interchange/xml-schema-for-the-tasks-element?view=project-client-2016)

louistrue:
Almost merge-ready — necessity, DRY, the absent changeset and all six mutations reproduce with the named tests you list. One item stops me, and it is the kind this PR is otherwise good at avoiding.

**A mutation survives while changing behaviour.** Replacing the single-derivation `sourceIds[index]!` read with a per-row `sourceIdOf(...)` call passes **all 182 tests** — and emits the collision warning **twice** instead of once. So the comment claiming the collision is reported "exactly once" is an invariant nothing enforces. A reviewer refactoring that read back to per-row computation would see a green suite and ship a duplicated warning.

One line closes it: assert the `synthesized-id-collision` warning count is exactly 1 on the collision test you already have.

**Second, the warning text is false in a reachable case.** Confirmed verbatim:

```
input: 'id,name,notes\nrow-3-no-id,A,\n,,leftover\nD,D,\n'
rows : [A, D]                      <- the blank-name row was dropped
warn : "Both rows were kept — this one is now row-3-no-id-x"
```

The row it says was kept is not in the output. It was dropped earlier for a missing name, and the collision warning does not know that. Worth fixing together with the count assertion since both concern the same warning.

Everything else holds up. The de-collision approach is right, and keeping the explicit id untouched while renaming the synthesized one is the correct direction — the author of the file wrote one of those two, and it should not be the one that moves.

BIMvoice:
Both fixed in 36ee75d3. The first one is the better catch of the two, and it is the shape this PR was otherwise built to avoid — you are right about that.

**The surviving mutation.** Replacing `sourceIds[index]!` with a per-row `sourceIdOf(...)` call passed all 182 tests *and* emitted the warning twice. The reason it slipped through is small and worth naming: the test used `find()`, which cannot see a duplicate. So "reported exactly once" was a property asserted only in prose, and a reviewer refactoring that read back would have shipped a duplicated warning against a green suite.

Now asserted as a count. I reproduced your mutation on the pushed tree (`REPLACEMENTS=1`, line re-read) and it fails the collision test where it previously passed.

**The false warning text.** Confirmed on your exact input — the blank-name row is dropped by the row loop, and the message still said it was kept. The cause is ordering: the warning fires in the id **pre-pass**, which runs before the row loop and cannot know what it will discard. So the message now describes only the id assignment, which is all the pre-pass actually knows:

> Row 3 has no id of its own, so "row-3-no-id" was synthesized for it, but another row states that exact id. Row 3 uses "row-3-no-id-x" instead, so the two do not collide, and a predecessor naming "row-3-no-id" refers to the row that states it, not to row 3.

A comment at the site records *why* it says nothing about the row's fate, so the claim does not creep back in, and a test pins that the message never asserts the row survived.

126 passing across the three import suites.

On your closing point — that the explicit id should not be the one that moves, since the author of the file wrote it — agreed, and that was the reasoning. It is now stated at the call site rather than only in the PR description, since that is the part a future reader would otherwise have to reconstruct.

louistrue:
Verified `36ee75d3`. Both items closed, and the count assertion does what it needed to.

**The surviving mutation is caught now.** Forcing the collision warning to be pushed twice fails, by name:

```
not ok 22 - warns with the real cause, not duplicate-source-id, when an explicit id
            collides with a synthesized one (issue #2071)
```

That is the class of change that previously passed 182 tests while double-emitting. `assert.strictEqual(collisions.length, 1)` is what turned the comment into something enforced — a `find()` genuinely could not have seen it, and the note you left at the assertion says exactly that.

**The false claim is gone.** Same input as before:

```
rows : ["A","D"]
warn : [synthesized-id-collision] Row 3 has no id of its own, so "row-3-no-id" was
       synthesized for it, but another row states that exact id. Row 3 uses
       "row-3-no-id-x" instead, so the two do not collide, and a predecessor naming
       "row-3-no-id" refers to the row that states it, not to row 3.
warn : [missing-name] Row has no task name, skipped.
```

No "both rows were kept". The message now describes only what it actually knows — the id assignment — and the row's later removal is reported by its own warning rather than contradicted by this one. Explaining where a predecessor token resolves is a good addition; that is the question a reader hits next.

Nothing outstanding from me on this PR. Merging once the last check lands.

Worth recording how this one was found, since it is the interesting part: the first review pass reported every line of the fix as pinned, and an adversarial second pass falsified that by constructing a mutant that changes behaviour and survives the suite. The finding came from disbelieving a clean result, not from reading the diff again.

louistrue:
One more, and it is the same claim you just fixed — surviving in a second place.

`docs/guide/schedule-import.md:43` still says:

> If your file happens to state `row-3-no-id` in an id cell, **both rows are kept**

That is conditionally true. Verified both branches:

```
'id,name\nrow-3-no-id,A\n,HasName\nD,D\n'   (blank-id row HAS a name)
  -> [A: row-3-no-id, HasName: row-3-no-id-x, D: D]      both kept ✓

'id,name\nrow-3-no-id,A\n,\nD,D\n'          (blank-id row has NO name)
  -> [A: row-3-no-id, D: D]                              one dropped ✗
```

The de-collision does its job in both cases — the id is moved out of the way either way. What varies is whether the row survives at all, and when it does not, it is for an unrelated reason (`missing-name`) that this sentence does not know about.

This is worth naming beyond the one-line fix: it is the same false claim you removed from the warning text in `36ee75d3`, still present in the guide. A claim corrected in one place and left standing in another is the shape that keeps recurring here — the changeset on #2062 said the `finally` chain prevented a strand while the code did the opposite, the `instancing.rs` comment on #2068 claimed parity with a function that had moved, and `csv.ts:252` on #2069 asserted an invariant the code does not have. Each was true when written and outlived the code it described.

Wording that holds either way: "the stated id is left exactly as written and the synthesized one is moved out of the way, so the two never collide — a row with a blank id cell is still subject to the usual rules (a row with no task name is skipped as before)."

Nothing else outstanding from me. Both earlier items are verified fixed: the double-emit is caught by name, and the warning no longer claims a kept row.

BIMvoice:
Fixed in 0c17e9af. You are right, and the pattern you named is the more useful finding.

Verified both branches before rewording:

```
'id,name\nrow-3-no-id,A\n,HasName\nD,D\n'  -> [A, HasName, D]   both kept
'id,name\nrow-3-no-id,A\n,\nD,D\n'         -> [A, D]            one dropped
```

The guide now states only what the id pre-pass actually guarantees — the stated id is untouched, the synthesized one moves, the two never collide — and says plainly that the row stays subject to the usual per-row rules, so a missing name still skips it for that reason rather than this one. I grepped both `docs/` and the schedule importer afterwards: the claim is gone everywhere, not just at `:43`.

**On the recurring shape.** Your list is right and worth stating as one thing rather than four coincidences: the #2062 changeset said the `finally` prevented a strand while the code did the opposite; the `instancing.rs` comment on #2068 claimed parity with a function that had moved; `csv.ts:252` on #2069 asserted an invariant the code lacks; and this sentence outlived the warning text it was describing. Every one was true when written.

The common mechanism is that prose is not reachable from the thing it describes. Code that drifts from code gets caught by a test; code that drifts from a sentence gets caught by a reader who happens to check. That is also why the #2053 fixture claim and the `esbuild@0.27.3` "version-pinned to match installed package" comment (#2079, a genuinely broken path) belong on the same list.

What I have taken from it, concretely: when a claim is corrected, grep the repo for the claim rather than fixing the site — which is what turned this one up as a second occurrence, and what I did before pushing.

louistrue:
Verified `0c17e9af`. The guide now says what the code does:

> the stated id is left exactly as written and the synthesized one is moved out of the way, so the two never collide … The blank-id row is then subject to the usual per-row rules like any other — a row with no task name is still skipped, for that reason rather than this one.

That last clause is the part that matters: it separates "did the de-collision work" (always yes) from "did the row survive" (depends on unrelated rules), which is exactly where the old wording overreached. Both branches I probed earlier now match the text.

CLI review of this head: 0 findings across its 7 files. 9 checks green, no unresolved threads.

Recapping what this PR ended up establishing, because the sequence is worth keeping:

1. the collision itself — a synthesized id equal to a stated one, dropping a task with a warning that named the wrong cause
2. a mutation that survived 182 tests while double-emitting the warning, closed by asserting the count rather than trusting a `find()`
3. the warning text claiming a row was kept when it had been dropped
4. the same claim still standing in the guide after the warning was fixed

Three of those four were claims outliving the code they described, in three different places, from one original fix. That is the pattern worth carrying forward more than any individual fix here.

Merging.

PR #2076, opened 2026-08-04
fix(viewer): validate MSPDI UIDs as integers and keep colliding ids apart

BIMvoice:
**Robustness for non-conformant input — not a bug reachable with files MS Project produces.** Saying that up front because the scope claim matters: the published Project XSD types `<UID>` and `<PredecessorUID>` as `xsd:integer`, so Project itself cannot emit an id that triggers this. The reachable population is hand-edited files and third-party exporters.

Follow-up from the MSPDI sweep on #2075, where the question "does the CSV id-collision bug (#2071) also affect MSPDI?" was settled as **no**. Nothing here reopens that.

## What was unguarded

`mspdi.ts:202` read `UID` as raw text with no numeric check; `:128` the same for `PredecessorUID`. `childText` returns trimmed text, and the numeric helper `childNumber` is never applied to either. So a document stating `<UID>row-3</UID>` collided with the synthesized `row-<n>` namespace:

```
ROWS  : [["row-3","Task A"],["7","Task C"]]     <- Task B silently gone
WARNS : duplicate task UID "row-3"              <- an id the user never wrote
```

After:

```
ROWS  : [["row-3","Task A"],["row-3-x","Task B"],["7","Task C"]]
WARNS : 1 task id(s) in this file are not integers ("row-3"), but the MSPDI schema
        types <UID> and <PredecessorUID> as xsd:integer. …
```

## Three judgement calls, each argued rather than defaulted

**The correctness half is the namespace, not the rejection.** Synthesized ids are suffixed until distinct from every stated id, so both tasks survive and a predecessor naming `row-3` resolves to the task that *states* it — what the file says.

**Non-integer ids are accepted with a warning, not skipped.** Skipping would turn an internally-consistent file using alphanumeric ids (`A1`, `A2`, …) into a **total** failure: every task skipped, then `rows.length === 0` throws "The file contained only the project summary row". That trades a silent wrong answer for a hard failure on a file that imports perfectly today. It also contradicts the parser's own contract at `:175-177` — per-task problems are warnings "so one malformed row cannot cost the user the whole import" — and matches the approach you endorsed on the CSV side, which keeps both rows and disambiguates the *synthesized* id rather than rejecting the stated one.

**The warning fires once per FILE, not once per task** — with a distinct-id count and up to three examples. A non-conformant id scheme is a property of the document; 500 identical warnings would bury the per-task ones (missing name, unreadable dates) the user must act on row by row. The deviation from the per-occurrence idiom is deliberate and commented. `useScheduleFileImport.ts:157-158` renders `w.message` generically and nothing switches on codes, so no UI change is needed — same as the CSV side.

## Evidence

| mutation | result |
|---|---|
| `MSPDI_UID_PATTERN` → always matches | killed, 3 fail |
| delete the uniqueness loop | killed, 1 fail |
| delete the `PredecessorUID` collection | killed, 1 fail |
| delete the `UID` validation | killed, 1 fail |
| delete the `UID === '0'` skip (sanity) | killed, 5 fail |

I re-ran the uniqueness-loop mutation myself (`REPLACEMENTS=1`, grep-confirmed gone): 1 failed / 29 passed, restored 30/30.

Negative cases: a conformant document is unaffected — asserted with a **deep-equal on an empty warnings array**, not merely the absence of the new code; the `UID 0` project-summary skip is intact and runs first; UID-less tasks with no collision still get `row-1`/`row-2` silently; and signed `<UID>-1</UID>` is accepted silently, since `xsd:integer` permits a sign.

2577 passing (was 2571), typecheck 34/34. No changeset — `@ifc-lite/viewer` is `private: true`, and only `apps/viewer` plus `docs/` changed.

## Two process notes worth recording

**A mutation that looked like a kill was a syntax error.** The first two attempts at the pattern mutation gave `REPLACEMENTS=0` and a corrupted regex, because Perl interpolates `$/` even inside `\Q…\E`. The run *looked* like a kill; only re-reading the mutated line exposed that the whole file was broken. Same class as the silent no-ops we've both hit today, with a new mechanism.

**One mutation would have survived, and the fixture was fixed rather than the survivor reported.** The first `PredecessorUID` test used a predecessor id that a task also stated as its `UID`, so the UID scan covered it and the `PredecessorUID` path was never the thing under test. Rewritten to use an id no task states, which is what turned it into a kill.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MSPDI schedule imports with safer handling of non-integer task and predecessor IDs.
  * Preserved dependencies involving non-integer IDs while preventing generated IDs from colliding with existing IDs.
  * Consolidated invalid-ID warnings to one notification per imported file.
* **Documentation**
  * Clarified MSPDI ID validation and warning behavior in the schedule import guide.


louistrue:
The correctness half is solid: silent row loss reproduced on `main` (Task B vanishes, with a `duplicate-source-id` warning naming the wrong cause), and the ~12-line `statedUids` pre-pass plus `while (statedUids.has(id)) id += '-x'` loop is load-bearing — deleting the loop alone kills the flagship regression test. Six mutations, six named kills. It also matches the house precedent set by #2071 on the CSV side, which is the right consistency argument.

Two things before I would merge it.

**1. The docs paragraph credits the wrong mechanism.** It says the check "protects the id a task with no UID is given", but the protection comes entirely from the regex-free `statedUids` suffix loop — the integer check only feeds the warning. That is an unenforced claim about which code does the work, and it will mislead whoever next tries to simplify one of the two halves.

**2. The warning half is discretionary, and it changes a currently-clean path.** A file using alphanumeric ids (`A1`, `A2`) imports **byte-identically** on `main` and on this branch — the rows, ids and links are the same. The only delta is the new warning, which flips that import from a success toast to a warning toast every time, because `useScheduleFileImport.ts:108` returns `success` only when `warnings.length === 0`.

That is a real cost paid by files that are not broken. Non-conformant against the published XSD, yes — but they import correctly today and will import correctly after. Worth deciding deliberately rather than as a side effect of the collision fix: either accept the toast change, or gate the warning behind something that indicates actual harm.

**DRY, and this one is worth acting on beyond this PR.** There are now two different solutions to the same problem in the same codebase:

- `csv.ts:263` — `row-${line}-no-id`, a fixed suffix that is only *probabilistically* collision-free, with a comment claiming it "can never equal an explicit id" (falsified by a file stating literally `row-7-no-id` — that is #2071)
- `mspdi.ts` — a `statedUids` pre-pass plus a `-x` suffix loop, which is strictly stronger and actually cannot collide

Both files also build the same "every id the file states" set up front (`knownIds` vs `statedUids`). A shared collect-stated-ids-and-uniquify helper would subsume both pre-passes and both synthesis schemes, and would let the false comment in `csv.ts` be deleted rather than corrected — enforced code instead of a claim. That belongs in its own PR, and #2071 is the natural home.

BIMvoice:
Both fixed in a2e5cba3 — and establishing the facts for item 2 **refuted the premise I was working from**, so the fix changed rather than got adjusted.

## Item 2: the three findings, one of which was the opposite of what I expected

1. `invalid-source-id` **was** the only thing flipping the toast — no other MSPDI warning keys on id format.
2. **The disambiguation path emitted no warning of its own.** `synthesizeSourceId` suffixed silently. So `invalid-source-id` was not redundant in the harmful case — it was doing *double duty*: a false alarm on clean files, and the sole, imprecise signal on broken ones. My assumption that the collision already had its own signal was simply wrong.
3. `useScheduleFileImport.ts:108` confirmed verbatim, and already test-pinned.

(c) is ruled out at source: `types.ts:54` documents `ScheduleImportWarning` as a *"Severity-free diagnostic"*, and its only consumers are the length check and a `console.warn`. There is no info channel to use.

## So (b), but not the shape I proposed

The obvious `nonInteger && displaced` would be **provably dead in its first half**: a synthesized id is always `row-N`/`row-N-x…`, which can never match `^[+-]?\d+$`, so every collision is by construction an integer-check failure. That conjunct would be unkillable by any mutation — the same unenforced-claim shape item 1 is about, reintroduced one line away from where you caught it.

So the gate is the collision alone and `MSPDI_UID_PATTERN` is deleted. The `xsd:integer` knowledge survives as prose in `synthesizeSourceId`'s doc explaining *why* a conformant file cannot collide — which is where it was always doing its real work.

The code is renamed `invalid-source-id` → `synthesized-id-collision`: it now reports an **event, not a property**, and names the substitution (`"row-3" → "row-3-x"`) because that id is visible in the imported schedule. `readDependencies` reverts to 3 args; a non-integer `PredecessorUID` is kept as written and, if unresolved, still surfaces through `build.ts:160`'s `unknown-predecessor`.

Net: a clean alphanumeric file is a clean import with a success toast; a colliding file gets a warning that says what was actually done to it.

## Item 1

Rewritten to credit the loop for the protection and to describe the **final** code rather than the old one. Every claim in it is now pinned: the loop by the regression test plus a new "keeps suffixing" case, "no warning" by the alphanumeric test, "only when a substitution happened" by the mutation below.

## Evidence

Making the warning unconditional (`size > 0` → `>= 0`) fails **4** tests — I re-ran that myself on the pushed tree (`REPLACEMENTS=1`, line re-read), including `imports an alphanumeric-id file with no warning at all — nothing went wrong`.

**One mutation initially survived**: `while` → `if` on the suffix loop, because no fixture needed more than one round. Added a file stating both `row-1` *and* `row-1-x`. Three assertions were strengthened from `!warnings.some(...)` to `deepStrictEqual(warnings, [])`, and one test was **deleted** rather than kept as a tautology — it exercised a check that no longer exists.

31 passing (was 30), viewer 2578, `tsc` clean. No stale `invalid-source-id` reference anywhere in the tree.

## Item 3 — DRY verdict, no code

Shareable, and **MSPDI's version is the one to keep**, but three specifics for whoever takes it:

- **Different shapes.** MSPDI collects only *stated* ids up front; `csv.ts:272`'s `knownIds` mixes explicit and synthesized. CSV must split into collect → synthesize → union, which is the natural helper boundary (`collectStatedIds` + `uniquify`).
- **CSV derives ids twice** — the pre-pass and the row loop — relying on `sourceIdOf` being pure. A suffix loop is order-dependent, so the helper must hand CSV a precomputed index→id map rather than a function called twice, or the two derivations silently disagree. That is exactly what `csv.ts:317-319`'s own comment warns about.
- **Visible id churn**: colliding CSV ids move from `row-7-no-id` to `row-7-no-id-x`, and those surface in the schedule and in predecessor resolution.

One asymmetry to **preserve, not unify**: CSV's no-id-column branch, where positional `"1"`, `"2"` are deliberately the real ids and must stay bare, since predecessors reference rows by position. That branch has nothing to collide with and should bypass the uniquifier entirely.

And you are right that the helper lets `csv.ts:257-258`'s comment — "so it can never equal an explicit id", falsified by a file stating literally `row-4-no-id` — be **deleted rather than reworded**. Enforced code instead of a claim, which is the whole argument.

BIMvoice:
The shared helper you asked for is built and verified, but **it stacks on this PR**, so I am holding it rather than opening a confusing diff. Details in case you want to merge in a particular order.

**Two of my own assumptions were stale, and checking them changed the plan:**

1. **`csv.ts` is already fixed.** #2075 merged as `e18356f3`, so `main` already has the `statedIds` pre-pass, the suffix loop, the `synthesized-id-collision` warning, and a truthful comment. **The false "can never equal an explicit id" claim does not exist any more** — I verified on `upstream/main`: the uniquifier is present, and that sentence greps to zero. So the thing you wanted deleted rather than reworded is already gone, and #2075 also fixed the twice-derived-id problem I was going to have to work around.
2. **`mspdi.ts` on `main` has no uniquifier at all** — it arrives with this PR. So on plain `main`, "adopt the shared helper in MSPDI" would mean re-implementing #2076.

Hence the stacking. The branch has your two commits from here as its base and the helper on top; **it should be opened after this merges**, and I will rebase then rather than now.

## What it does

```ts
collectStatedIds(items, statedIdOf): Set<string>
uniquifySynthesizedId(preferred, stated): string
```

Exactly the boundary you named. `uniquifySynthesizedId` **returns** the id and never mutates `stated` — that set means "ids the *file* states", which is what predecessor resolution is judged against, so adding an invented id would make it addressable and displace the next synthesized id for nothing. A substitution is `result !== preferred`, false for every well-formed file.

**The warning deliberately stays out of the helper.** Both sides already emit `synthesized-id-collision` so the UI groups them, but CSV attributes per row to a physical file `line` while MSPDI has none to cite — a `<Task>` is an element, and its test asserts one warning per *file* where CSV's asserts one per *row*. Folding the message in means inventing a shape that fits neither.

**The no-id-column branch bypasses the uniquifier entirely**, as you'd expect: positional `"1"`, `"2"` are the real ids and suffixing `3` → `3-x` would break `3FS+2 days`. The bypass now records why. `csv.test.ts:639` still passes.

**Considered and rejected:** `lens-editor-utils.ts:103`'s `reserveUniqueId` is the same loop shape with a different contract — it *mutates* `taken`, counts numerically, and its reserved ids **are** addressable. Unifying would change lens id output.

## Evidence

| mutation | csv.test.ts | mspdi.test.ts |
|---|---|---|
| neutralise the suffix loop | **3 fail** | **3 fail** |
| neutralise the set-build | **3 fail** | **3 fail** |

Both halves load-bearing for **both** importers — that is the proof it is one path rather than a second copy, which was the whole point.

Pure refactor, zero behaviour change: `csv.test.ts` and `mspdi.test.ts` are **byte-identical** to the base, which is itself the byte-identity proof for the no-collision cases, since both assert exact ids throughout. So the id churn I was braced for does not happen — #2075 and this PR already absorbed it.

**No RED for CSV exists and I did not fake one:** on this base its colliding case already produces a uniquified id, so a test written for it would pass against the unfixed code. Vacuous by our own standard, so the mutations above are the real evidence.

2595 passing (+8, exactly the new helper tests), `tsc` clean. `docs/guide/schedule-import.md` untouched — nothing user-visible changed, and I re-checked its four id claims are true of the final code rather than adding a fifth.

louistrue:
**Your two red checks are not yours.** Diagnosed and filed as #2094.

`Node tests` fails in `@ifc-lite/collab-server`:

```
FAIL test/access-control.test.ts > revocation retention >
     loads the legacy revoked array shape and keeps those revocations biting
AssertionError: expected 1785838912 to be 1785838913
```

Two Unix timestamps one second apart. `adminClaims()` recomputes `Date.now()` on every call, and the test calls it twice around a debounced persist and a `flush()` — if the wall clock's second boundary falls in that window, the two `exp` values differ by one. Nothing to do with MSPDI import.

`Build + WASM + Rust + Node` is the aggregate gate reporting the same failure, so one unlucky second produces both red rows.

Re-running should clear it. #2094 has the fix (capture the claims once and reuse).

---

On the PR itself: `a2e5cba3` addresses the toast concern the right way. Warning on the **collision** rather than on the id format means a file using alphanumeric ids — which imports byte-identically before and after — no longer flips from success toast to warning toast for a format that harms nothing. The warning now fires when something actually happened.

That leaves the docs item from my earlier comment: the paragraph still credits the integer check for collision protection that the `statedUids` suffix loop actually does. Smaller now that the check no longer drives the warning, but the attribution is still backwards.

And the cross-PR DRY point stands for its own PR, not this one: `csv.ts`'s `row-${line}-no-id` fixed suffix and `mspdi.ts`'s `statedUids` + `-x` loop are two solutions to one problem, with the CSV one carrying a comment #2071 already falsified.

BIMvoice:
Thanks for diagnosing #2094 — I had been about to re-run those two checks blind.

**On the docs item, I think you may be looking at the pre-`a2e5cba3` text, and I would rather ask than churn the paragraph.** Your comment landed at 09:48; `a2e5cba3` is timestamped 09:15:23Z, so it should have been in view, but the current text reads:

> The MSPDI schema types `UID` and `PredecessorUID` as `xsd:integer`, so a file MS Project wrote always has integer task ids. A hand-edited or third-party-exported file may not, and the importer takes those ids **exactly as written** — a schedule numbered `A1`, `A2`, … imports like any other, with no warning, and its dependencies resolve against the ids the file states.
>
> What is guarded is the id the importer has to invent for a task with **no** `UID`. … Before any id is synthesized, the importer collects every id the file states and extends the synthesized one (`row-3` → `row-3-x`) until it matches none of them — **that loop is what keeps both tasks**, and keeps the stated id pointing at the task that stated it.

So the integer mention is now scoped to "what MS Project writes / what we accept", and the protection is credited to the loop explicitly. `grep -nE "integer|xsd"` over the whole guide returns exactly that one line.

If there is a specific sentence still reading backwards to you, point me at it and I will fix it — but I did not want to reword a paragraph that already says what you asked for and risk making it worse. It is possible the rewrite was part of the same commit that addressed the toast item, so a review of the earlier state would show the old attribution.

louistrue:
Both my findings are closed. The docs paragraph now credits the right mechanism —

> the importer collects every id the file states and extends the synthesized one (`row-3` → `row-3-x`) until it matches none of them — **that loop is what keeps both tasks**

and says plainly that an `A1`/`A2` schedule "imports like any other, with no warning", which is what the toast change made true.

One item from a local CLI run, and it is worth taking because the behaviour is right and the test cannot show it.

**`reports every clash in one warning, not one per displaced task` has only one displaced task.**

The fixture is stated-`row-1`, UID-less Task A (displaced to `row-2-x`), stated-`row-2`. Exactly one displacement, so `clash.length === 1` holds identically whether the code aggregates or emits one warning per task. The name promises a distinction the fixture cannot make.

I checked the behaviour separately, and it does aggregate — adding a second UID-less task whose positional id is also stated:

```
WARNCOUNT 1
IDS       ["row-1","row-2-x","row-2","row-4-x","row-4"]
MSG       2 task(s) with no <UID> could not take the id their position implies, because the
          file states that id for another task ("row-2" → "row-2-x", "row-4" → …)
```

One warning, both substitutions named in it. So the fix is correct and the assertion is the thing that is thin — two extra fixture lines make the test mean its own name, and the message's `2 task(s)` plus both arrow pairs become assertable.

Worth noting the shape, since it is the third time today a test's name has promised more than its fixture could deliver (#2047's cycle test asserted bucket membership rather than reachability; #2054's parity test discarded the class column). Each time the code was right and the test was the weak point — which is the harder case to notice, because nothing fails.

BIMvoice:
Fixed in 36693ac2. You were right, and the comment was wrong in the same way the fixture was.

The test said *"Two UID-less tasks, both of whose positional ids the file states"* — and the fixture had **one**. So `clash.length === 1` held identically whether the code aggregated or emitted per-task. The name promised a distinction the fixture could not make, and the comment described a fixture that did not exist.

Now two displacements — Task A at index 1 (positional `row-2`, stated by task 3) and Task B at index 3 (positional `row-4`, stated by task 5):

```
['row-1', 'row-2-x', 'row-2', 'row-4-x', 'row-4']
```

with the single warning asserted to name **both**, since aggregating while dropping one would otherwise pass.

**Mutating the aggregation into a per-task loop now fails it.** Against the old fixture that same mutation passed — which is the whole point, and I checked both directions rather than assuming.

That is the third time today a test of mine asserted something its fixture could not distinguish, after the vacuous `StringTable.fromArray` index shift and the storage double missing `length`/`key()`. The common thread is that all three passed for a reason unrelated to the property under test, so the count or the shape looked right. Your habit of re-deriving the behaviour separately — as you did here by adding a second UID-less task and reading the actual message — is what catches that class; a green run cannot.

31 passing.

Issue #2077, opened 2026-08-04
A drained promise job that throws still resolves as success


BIMvoice:
Raised by @louistrue while merging #2063, and filed here so it does not ride along with a fix that is correct for a narrower case.

## The split

`packages/sandbox/src/sandbox.ts` now reads the drain's outcome through **two disjoint channels about thirty lines apart**:

```
:166   this.runtime.executePendingJobs().dispose();   // result discarded unread
:195   if (this.interrupted) { ... }                  // side-channel boolean
```

#2063 closed the **interrupt** case: a CPU timeout inside a drained job used to return the main body's value and report success, and the interrupt handler's own flag is now consulted because the drain reports nothing for it.

But the discarded result at `:166` is the only thing that carries a job that **throws** for any other reason. So an ordinary rejection inside a promise job still resolves as success with the stale main-body value — the same defect, a different trigger, and the flag added in #2063 does not see it.

## Why it is the same shape as the bug it sits next to

The failure is invisible in exactly the way the timeout one was: `eval()` returns a plausible value, `result.error` is null because the main body succeeded, and nothing in the return shape says a job failed. A script whose `async` entry point throws after its first `await` gets reported as a successful run.

## Suggested direction

Read `executePendingJobs()`'s result before disposing it, rather than discarding it — its failure branch owns a live error handle — and surface a job failure the same way `result.error` is surfaced for the main body, so the two paths report identically. That is the shape #2063 used for the interrupt: mirror the existing main-body error path rather than invent a second one.

Care needed on two points, both learned in #2063:
- **Don't turn every async script into a failure.** A job that completes normally must still return the main-body value and report success. That negative case is what caught an over-broad version of the interrupt fix.
- The result handle must still be freed on every exit, including the new failure branch — a leaked handle keeps a `JSObject` on the runtime's GC list and makes `runtime.dispose()` abort the whole WASM module (#1905, and the abort catalogued in #1922).

## Not urgent, but not cosmetic

`eval()` handing back a wrong answer while reporting success is the sharpest version of the defect class this repo keeps turning up — the check ran, said pass, and verified nothing.

## Related

#2063 (merged — the interrupt half), #1922 (the upstream `JS_FreeRuntime` abort), #2062 (dispose ordering).

BIMvoice:
**Correcting my own issue text.** The symptom is real, but the mechanism I proposed is wrong and the fix in the PR below does not follow it.

[...]

PR #2078, opened 2026-08-04
fix(sandbox): report a rejected entry promise as an error, not as data

BIMvoice:
Closes the returned-promise half of #2077. **The issue's suggested mechanism was wrong — and it was my suggestion, so I have corrected the issue text too.**

## What I got wrong

I proposed reading `executePendingJobs()`'s result before disposing it. `quickjs-emscripten` 0.32 documents the opposite at `index.d.ts:306`:

> executePendingJobs will not normally return errors thrown inside async functions or rejected promises.

Nine constructs probed against a raw runtime — async throw (Error and number), `.then` throw, awaited chain throw, bare `Promise.reject`, IIFE ReferenceError, interrupt inside an async body, two thenable-`then` throws. **Every one returned `drain=ok, N job(s)`** with the failure branch empty. QuickJS captures the throw in the promise; nothing reaches the drain result.

## The actual cause

One line further on, at `sandbox.ts:199`. The extension host wraps an entry as `return activate(ctx)` (`source-wrap.ts:136-138`), so for an `async` entry **the eval value IS the promise**. A throw after its first `await` settles it rejected without ever touching `result.error`, and `vm.dump` renders the rejection as **data**:

```
{"type":"rejected","error":{"name":"Error","message":"boom","stack":"..."}}
```

returned inside a *successful* `ScriptResult`. The promise's own state is the only signal, so `getPromiseState` is what the fix consults. The main-body error path's dump-and-stringify is extracted into `describeError()` and both paths report through it, rather than growing a second error shape.

## Ordering: interrupt first

The rejection check sits **after** `if (this.interrupted)`. An interrupted async entry rejects with `InternalError: interrupted`, so for that shape both channels yield the same text — but the flag is the deliberate timeout channel and does not depend on QuickJS's reason text, and for the fire-and-forget shape (#2063's actual test) only the flag ever sees it. Honest caveat: the ordering guard test asserts the entry shape reports `interrupted`, but since both channels produce that string it does not by itself discriminate the ordering.

## What this does NOT fix, stated in the changeset and at the call site

A rejection in a promise the script never hands back — `run(); 'started'`, **the literal example in #2077's body** — remains invisible. There is no handle to inspect, and `RuntimeOptions.promiseRejectionHandler` is unimplemented in 0.32 (`index.d.ts:498`, typed `TODO<...>` = `never`). Closing that needs a host rejection tracker the library does not expose. #2077 stays open for it.

## Evidence

| mutation | result |
|---|---|
| `rejectionMessage !== null` → `=== undefined` | caught, 4 failed |
| drop the promise-value dispose | **initially survived** — see below |
| drop the rejection-reason dispose | caught, `Aborted(... list_empty(&rt->gc_obj_list) ... JS_FreeRuntime)` |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, line re-read): 4 failed / 45 passed, restored 49/49.

**The survivor was a real gap in my test, not in the code.** The fulfilment value was `42` — a non-object, which never lands on the GC list, so the `dispose()` leak oracle could not see it. Both leak tests now use **object** values and rejection reasons, and the mutation fails with the #1905 abort.

Negative cases: normal async completion still succeeds, asserted twice and both proving the job **ran** via a `ran` flag read in a follow-up eval (a no-op job would satisfy a value-only assertion); a main-body throw is unchanged, with a test asserting the job path's message is **byte-identical** to the main body's for the same thrown value; and all four of #2063's interrupt tests stay green.

**Declared uncovered branch:** the new `drained.error` read is defensive — no script construct I could find reaches it (nine probed). Kept because discarding an error result unread is what the issue calls out, and it costs 8 lines.

49 passing (was 41), extensions 598, sdk 62, typecheck 34/34.

## Pre-existing, found and not fixed

`vm.dump()` **disposes** the handle it is given when that handle is a promise (`chunk-V2S4ZYJR.mjs`). So on `upstream/main` every promise-returning script already logs `[ifc-lite/sandbox] eval-result handle could not be disposed QuickJSUseAfterFree` from the `finally`. Verified it reproduces unchanged on main; this change neither adds nor removes it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Rejected promises returned by asynchronous scripts are now reported as script failures with useful error details.
  - Interrupts, timeouts, synchronous errors, and queued-job exceptions continue to be handled correctly.
  - Improved cleanup prevents resource leaks during script execution.
- **Tests**
  - Added coverage for promise rejection, fulfillment, interruptions, logging, timing, repeated evaluations, and cleanup.


louistrue:
The core fix is real and pinned — a rejected entry promise that used to surface as resolved-with-failure-as-data now becomes a `ScriptError`, and four mutations are each killed by named tests, including the genuine `JS_FreeRuntime` gc_obj_list oracle.

One thing to fix first, and it is the bar this repo keeps tripping over.

**The changeset asserts behaviour that cannot be triggered.** The sentence

> an exception that stops the queue outright is reported instead of discarded

describes the `drained.error` branch. That branch is **unpinned** — deleting it leaves the suite green — and **unreachable from script**: three separate attempts across ~20 constructs failed to produce it. `@ifc-lite/sandbox` is a published package, so that sentence ships to consumers as a statement about what the release does.

Cleanest exit is deleting the ~12-line dead branch rather than the sentence: then the changeset stays true and there is no untested code claiming to handle a case nothing reaches. If you would rather keep the branch as defence against a future quickjs-emscripten change, keep it and drop the sentence — but not both.

**Second, worth a release note rather than a change.** This turns previously-*resolving* evals into throws. That is the point of the fix, but it is a behaviour change on a patch bump: an embedder catching nothing today because `eval()` resolved will now see a `ScriptError` propagate. One sentence naming that saves someone a confusing upgrade.

Everything else checks out, including the two items the review disclosed itself (the unpinned catch wrapper and the ordering comment) — both tolerable, and disclosed rather than papered over, which is the right instinct.

BIMvoice:
Fixed in e229f64c — took your first option.

**Deleted the dead branch rather than the sentence.** You were right that keeping both was the one thing not to do: an unpinned, unreachable branch and a changeset telling consumers the release handles that case. `@ifc-lite/sandbox` is published, so that sentence would have shipped as a claim about what the package does.

The drain and its disposal stay — a leaked handle keeps a `JSObject` on the runtime's GC list and makes `runtime.dispose()` abort the whole module (#1905) — and *why* the branch is gone is now recorded at the call site, so it does not get re-added as an obvious robustness improvement. That is the same failure mode the `finally` chain on #2062 had: a plausible-looking guard nobody could argue against without the reasoning written down.

**Release note added**, in the changeset rather than only here:

> **Behaviour change for embedders.** An `eval()` that previously *resolved* — carrying the rejection as data in `value` — now throws a `ScriptError`. Code that relied on it resolving, and so caught nothing, will start seeing that error propagate. That is the point of the fix, but it lands on a patch bump.

49 passing, typecheck 34/34.

Worth noting the pattern this makes three of today: the #2062 changeset claimed the `finally` prevented a strand while the code did the opposite; the #2075 guide said both rows survive when one could be dropped; this one asserted a capability the branch could not deliver. All three were prose describing intent rather than code — which is the argument for deleting the code when the two disagree, not softening the sentence.

louistrue:
Verified `e229f64c`. You took the cleaner of the two exits and the result is better than either.

**The dead branch is gone** — no `drained.error` references remain in `packages/sandbox/src`. That is the right call over deleting the sentence: keeping an unreachable, unpinned branch to satisfy a changeset would have left untested code claiming to handle a case nothing can produce.

**The changeset now describes only what happens**, and the added paragraph is the part I would have asked for separately:

> **Behaviour change for embedders.** An `eval()` that previously *resolved* — carrying the rejection as data in `value` — now throws a `ScriptError`. Code that relied on it resolving, and so caught nothing, will start seeing that error propagate. That is the point of the fix, but it lands on a patch bump, so it is worth knowing before upgrading.

Naming a behaviour change on a patch bump, rather than letting an embedder discover it, is the difference between a fix and a surprise.

**Still pinned after the deletion** — removing the rejection throw fails four named tests:

```
× reports a rejection thrown after the first await as a ScriptError
× reports the rejection reason as the message, like a main-body throw
× carries the logs and duration a main-body error carries
× does not carry a rejection over to the next eval, and frees its handles
```

The fourth is the one worth having: it covers the handle-freeing side, so a fix that reported the rejection but leaked its handles would still fail.

49 tests pass. Nothing outstanding from me on this PR.

louistrue:
A local CLI run on this head surfaced three items. One is a real bug I reproduced, one I would decline, one is a docs follow-up.

## Real, reproduced, and pre-existing — worth its own issue

**A caught `ScriptError` loses its logs when the next `eval()` runs.** `ScriptError` stores `logs` by reference (`sandbox.ts:318`), and `eval()` clears the same array in place (`this.logs.length = 0`). So:

```
catch (e)  -> e.logs = [["from first eval"]]
await sandbox.eval('1;')
           -> e.logs = []            // retroactively emptied
```

An embedder that catches an error, keeps it, and later runs another eval finds the diagnostic context gone — and gone *after* they looked, which is the worst version. A `[...logs]` copy at construction fixes it.

**Not yours**, and I checked rather than assumed: `this.logs.length = 0` is on `origin/main` at line 120, and `ScriptError` already stored by reference there. What this PR does is take the ScriptError sites from 3 to 5 — the rejection path is now a fifth way to receive an error whose logs can be wiped. Wider exposure, same defect. Filing separately.

## The MAJOR I would decline

It asks to inspect `executePendingJobs()`'s result before disposing it — which is the branch you just deleted at my suggestion, arriving back from the other direction.

I still think deleting was right, and the reason is unchanged: three separate investigations failed to produce that error from script across ~20 constructs, and the branch was unpinned. Re-adding it would restore untestable code for a case nothing reaches. If it ever becomes reachable — a quickjs-emscripten change to how `executePendingJobs` surfaces a queue-stopping throw — that is the change that should carry both the branch and its test.

Worth noting the pattern though: this is the second time today a review has asked to reinstate defensive machinery removed as unreachable (the `finally` chain on #2062 was the first). Both times the shape proposed was reasonable; both times the premise was a path nothing can take.

## Docs follow-up

The API guide should carry the new error contract for `Sandbox.eval()` — that a previously-resolving rejection now throws. Your changeset says it well already; the guide is where an embedder will look. Non-blocking, but it belongs with this change rather than after it.

Everything from my earlier verification stands: the dead branch is gone, the rejection path is pinned by four named tests, and 49 tests pass.

BIMvoice:
Docs follow-up done in 4bec2556; the other two need nothing from me.

**The guide now carries the contract**, in `scripting-sdk.md` next to the `Sandbox` example rather than only in the changeset — you were right that the guide is where an embedder looks:

> **When `eval()` throws.** It rejects with a `ScriptError` … for a throw in the script's main body, for a CPU timeout, and — since the fix in #2078 — for a script whose entry point is `async` and whose returned promise **rejects** after its first `await`. That last case previously *resolved* … If you upgrade across that change and were relying on `eval()` resolving, you will start seeing the error propagate.

I also recorded the case still **not** reported — a rejection in a promise the script never hands back — since that is better stated than discovered.

**On the MAJOR: agreed, and thank you for holding the line on it.** Re-adding the branch would restore untestable code for a case ~20 constructs across three investigations could not produce.

Your observation about the pattern is the part worth keeping. Twice today a review asked to reinstate defensive machinery that had been removed as unreachable — the `finally` chain on #2062, and now this. Both proposals were reasonable in shape; both rested on a path nothing can take. What makes them hard to refuse is that the argument for defensive code never depends on the case being reachable, so "it costs nothing" is always available. The check that actually settles it is the one you applied both times: *can anything produce this state* — and if not, the code is a claim rather than a guard.

That is also why deleting beat softening the sentence: a comment saying "handled" with no reachable path behind it is the same defect as a changeset asserting a capability the branch cannot deliver.

**#2092 — I checked before writing, and you had already filed it**, so I have not duplicated. Confirmed the defect on `main` independently: `this.logs.length = 0` at `sandbox.ts:120` and `ScriptError` storing `logs` by reference at `:267`. Your framing that this PR widens the exposure from three sites to five without introducing the defect matches what I see. If you want it, it is a one-line `[...logs]` at construction and I can take it.

louistrue:
Verified `4bec2556`. The delta from the CLI-reviewed commit is docs-only (4 added lines in `docs/guide/scripting-sdk.md`), so the code review still stands.

The second paragraph is the part worth keeping:

> One case is still **not** reported: a rejection in a promise the script never hands back — `run(); 'started'` rather than `return run()`. There is no handle to inspect, and quickjs-emscripten exposes no host rejection tracker, so nothing observes it.

Documenting the boundary of the fix, with the mechanism for why it stops there, is the opposite of what has gone wrong repeatedly in this repo today. Seven PRs carried a comment, changeset or doc asserting more than the code delivered. This one states less than a reader might assume and says why — so someone hitting `run(); 'started'` finds the answer instead of concluding the fix is broken.

Correctly left `ScriptError.logs` aliasing alone; it is tracked in #2092 with the reproduction, and it is not this PR's to fix.

Summary of the whole PR from my side: the rejection path is pinned by four named tests including the handle-freeing one, the unreachable drain-error branch was deleted rather than papered over, the changeset names a behaviour change on a patch bump, and the guide now carries both the new contract and its limit. Merging once CI lands.

Issue #2079, opened 2026-08-04
The esbuild-wasm CDN fallback is pinned to a version that cannot initialise


BIMvoice:
Found by the house-rule sweep in the PR that follows, and left unfixed there deliberately — the fix is a behaviour change to a published package and needs a decision about how to stop it drifting again.

## The mismatch

`packages/sandbox/src/transpile.ts:60` on `upstream/main`:

```ts
wasmURL = `https://unpkg.com/esbuild-wasm@0.27.3/esbuild.wasm`;
```

with a comment claiming it is "version-pinned to match installed package". It is not:

```
packages/sandbox/package.json   "esbuild-wasm": "^0.28.1"
node_modules installed          0.28.1
```

esbuild refuses a host/binary version mismatch outright — `Cannot start service: Host version does not match binary version`. So the CDN fallback cannot initialise.

## When that path runs

It is the fallback for bundlers that do not implement Vite's `?url` asset hint. So for an embedder on a non-Vite build, TypeScript transpilation silently drops to the regex fallback rather than esbuild — a quieter, less capable transpiler — with no indication why.

## Why it went unnoticed, which is the point

The `catch` at `:65` swallowed the resulting `initialize()` failure with no log. The caller's existing warning says `esbuild unavailable, using fallback transpiler` and carries **no cause**, so the version mismatch never surfaced anywhere. That catch is one of the three the sweep was dispatched to fix, and this is what was behind it.

The literal was introduced in #215 and never touched since; the dependency range moved without it. A `^` range and a hard-coded URL cannot stay in sync by construction.

## Directions

Bumping `0.27.3` → `0.28.1` fixes it today and breaks again at the next minor bump, so it is not sufficient on its own. Two durable options:

1. **Derive the URL from the resolved package version** at build or runtime, so there is nothing to keep in sync.
2. **Exact-pin `esbuild-wasm`** in `package.json` (drop the `^`) and add a test asserting the literal matches the resolved version, so a bump that forgets the URL fails CI.

(2) is cheaper and matches the repo's existing anti-drift idiom — the `*.parity.test.ts` files that read the Rust source and assert the TS constants match. (1) removes the class entirely.

Either way the fallback warning should carry the cause, which the PR above now does.

## Not urgent for the viewer

The first-party viewer builds with Vite, so it takes the `?url` path and never reaches this. The affected population is third-party embedders of `@ifc-lite/sandbox`.

PR #2080, opened 2026-08-04
chore: close the silent catches and as-any casts the house rules forbid

BIMvoice:
Follow-through from you holding #2062 over a single bare `catch`. Your reasoning is the brief: *"nothing mechanical will ever catch this; the rule holds only if reviews hold it."*

**The sweep found a live defect, filed as #2079** — see below. That is the concrete payoff.

## Survey (2,464 tracked files; generated paths excluded and listed)

Two corrections to the raw greps first: `as any` and `@ts-ignore` appear far more often in **prose** ("avoid `as any`", "…the same as any other…") than in code. Comment-stripped counts only.

- **`@ts-ignore`: 0 in code**, repo-wide. All three hits are prose.
- **`@ts-expect-error`: 1**, in a test, **with** a justification.
- **`as any`: 130 — but only 3 in non-test source.** All three fixed here, so non-test source is now **0**. The other 127 are tests, 84 of them one pattern in `packages/query`.
- **`catch`: 1,050 blocks classified** — 405 log, 350 swallow (non-empty body), 148 comment-only, 144 rethrow, **3 literally empty**.

That last row is the survey's real finding, because **the count depends on which reading of the rule you take**:

| reading | count |
|---|---|
| strict (`catch {}` = empty or comment-only body) | **151** |
| the one you applied on #2062 (neither logs nor rethrows) | **501 of 1,050** |

The three sandbox sites you named all have *non-empty* bodies, so you were applying the broader reading. That gap decides whether a gate is landable, so it is spelled out rather than averaged over.

## What I fixed, per site

- **`transpile.ts:58`** (esbuild.wasm asset probe) — **log, not rethrow**. The CDN fallback is the intended behaviour under bundlers that ignore Vite's `?url`, so rethrowing breaks a working path. But the swallow hid that the sandbox just swapped a bundled asset for a `unpkg.com` network fetch, which an embedder needs to know.
- **`transpile.ts:65`** (esbuild.initialize) — **log, not rethrow**. Contract is documented as "returns null if unavailable" and the caller degrades. But this catch is the **only** place holding the reason; the caller's existing warning carries no cause. **This is the one that hid #2079.**
- **`transpile.ts:103`** (same file, same shape, not on your list) — already warned but dropped the error. Now passes it. No new log call.
- **`bridge.ts:91`** (JSON.stringify sizing) — **log at most once per context**. Per-occurrence logging would be a real defect: the trigger is script-supplied, so `for(;;) console.log(1n)` in an untrusted script floods the host console, the exact resource exhaustion the surrounding caps exist to prevent.

  Reachability **verified, not assumed**: a `BigInt` survives `vm.dump` but not `JSON.stringify`. A cyclic object does **not** reach it — `vm.dump` returns `"[object Object]"` — so the comment naming cyclic args was wrong, and is corrected. New test pins both that it warns and that it warns *once* across three calls; dropping the guard fails it (3 vs 1), deleting the warn fails it (0 vs 1). I re-ran the first myself on the pushed tree.
- **`upload-fixtures.mjs:164,202`** — the only literally-empty `catch {}` outside `apps/landing`. Best-effort cleanup; each now names the path left behind. Rethrowing from an exit handler would mask the real upload error.
- **`StatusBar.tsx:62-63`** — narrowed `(performance as any).memory` to a concrete shape, matching the existing house pattern at `pointcloud/streaming/host.ts:191`.
- **`CesiumOverlay.tsx:256`** — the cast was simply **stale**; Cesium 1.143 declares `SkyBox.show`. Removed.

**No swallow was converted to a rethrow anywhere.** Every change is additive logging plus two type narrowings — adding a log is not a behaviour change, converting a swallow is, and none of these had a test to justify that.

## Deliberately left

- **The 145 remaining strict-reading silent catches.** Each needs a judgement call about what should happen on failure, and a 145-site error-handling diff is unreviewable. The `packages/geometry` cluster (15) is the interesting one — WASM teardown and worker-message paths where a rethrow is plausibly right but would definitely change behaviour. Full `file:line` list for everything outside the two dense clusters is in the survey.
- **The 127 `as any` in tests** — mechanical, no behavioural upside, better as its own PR.
- **`apps/landing/app.jsx`** — browser capability probes where the swallow is the design. Note `:1952`'s body is literally `/* swallow */`, a comment that justifies nothing. That file has no `package.json`, is outside `pnpm typecheck`, and has no suite.

## Recommendation on a mechanical check

Worth adding, **but not as a hard gate on the broad rule** — 501 sites (149 strict) is not landable, and a gate that has to be switched off is worse than none.

Landable today as `scripts/check-house-rules.mjs`, in the style of `check-test-wiring.mjs`:

1. **`@ts-ignore` → hard gate.** Zero violations, ships green now. Must strip comments or it flags the three prose mentions.
2. **`as any` in non-test source → hard gate.** Zero after this PR. Tests exempt, with the exemption recorded as known debt rather than silently baked in.
3. **Literally-empty `catch {}` → hard gate**, one allowlist entry (`apps/landing`) or zero if that gets fixed.
4. **Swallowing `catch` → ratchet, not a gate.** Same shape as the Rust `module_size_ratchet`: a committed allowlist of the current 149, failing only on a *new* one. That is the mechanism that actually stops drift and the repo already trusts it.

The classifier must be **brace-matching, not regex** — `catch {` with a non-empty body is the majority case and the one you flagged, and a line regex cannot tell it from a body that logs. Mine is ~40 lines; **not committed**, per the instruction to propose while the count is high.

Caveat if you take (4): "logs something" is a heuristic (`console.`, `logger.`, `.warn(`, …) with false positives and negatives at the margin, so the allowlist should be generated once and **reviewed**, not trusted blind.

## Verification

sandbox 42 passing (was 41), viewer 2571 unchanged (baseline taken by copying the edited files aside, `git checkout --`, full 73s run, then restore and re-run — identical both ways), typecheck 34/34, `check-test-wiring` passes.

Changeset included for `@ifc-lite/sandbox` despite this being logging-only: it is published and embedded, and the CDN-fallback warning is observable output that tells an embedder their build is doing an unexpected network fetch. The viewer and `scripts` changes are unpublished and get none.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox logging now preserves captured entries when certain values cannot be sized.
  * Repeated sizing warnings are limited to one per sandbox session.
  * Transpilation and resource-loading fallbacks now provide clearer diagnostic warnings while continuing to work.
  * Temporary staging cleanup reports warnings when files or directories cannot be removed.
  * Viewer memory reporting safely handles environments without detailed browser memory data.

* **Tests**
  * Added coverage for sandbox log preservation, warning behavior, and initialization cleanup.


louistrue:
Went through all three. Two are real and neither is yours; the third is worth taking.

## Security MAJOR (`bridge.ts:100`) — real, pre-existing, needs its own issue

The finding is correct. When `JSON.stringify(nativeArgs)` throws, `entryBytes` stays 0 and the entry is appended anyway, so repeated unserializable logs accumulate against `MAX_LOG_ENTRIES` (1000) while contributing nothing to the 4MB `MAX_TOTAL_BYTES` budget. A script can retain up to 1000 arbitrarily large objects that the byte cap never sees.

**But it is on `main` already** — identical shape at `bridge.ts:88-95`:

```
let entryBytes = 0;
try { entryBytes = JSON.stringify(nativeArgs)?.length ?? 0; }
catch { entryBytes = 0; /* unserializable args — skip cost accounting */ }
totalBytes += entryBytes;
```

Your change is `catch {` -> `catch (err) {` plus the one-time warn. It makes the condition *visible* without altering it, which is strictly an improvement. Filing this separately rather than growing a house-rule sweep into a budget-accounting fix.

Worth saying that the one-notice-per-context design is right and not obvious: a per-occurrence log would hand `for(;;) console.log(1n)` a host-console flood, so naive compliance with the no-silent-catch rule would have created a second DoS while fixing the visibility of the first.

## `transpile.ts:64` — real, and already fixed in #2081

`package.json` declares `esbuild-wasm: ^0.28.1`; the CDN fallback fetches `0.27.3`; and the comment claims "version-pinned to match installed package", which is false. Also pre-existing — your diff here only adds a full stop to that comment.

#2081 closes it, and notes this sweep is what surfaced it: making the swallowing `catch` log its cause is what exposed a fallback nobody knew was firing with a mismatched binary. That is the sweep paying for itself, and a good argument for the house rule existing.

## The test-spy MINOR — worth taking

`createSandbox()` runs before the `try`, so a rejected initialization skips the `finally` and leaves `console.warn` stubbed for the rest of the file. Moving it inside, or spying after construction, costs a line.

## Verified clean otherwise

I checked the thing that would actually worry me in a sweep like this: **no `catch {}` was converted to a rethrow.** Every one is `catch (err) { console.warn(...) }`, so no previously-swallowed error now propagates. Diffed against `origin/main` rather than the merge base, since this branch is behind and the three-dot diff otherwise pulls in #2063's work and misreads as new throws.

The `as any` removals are genuine narrowing too — `PerformanceWithMemory` is a real shape, not a cast moved elsewhere.

BIMvoice:
Fixed the test-spy MINOR (8532773c).

Confirmed your reading: in `packages/sandbox/src/bridge-console.test.ts:24-26` (pre-fix), `createSandbox()` ran before the `try`, so a rejected `createSandbox()` skipped the `finally` and left `console.warn` stubbed for the rest of the file. Reproduced it directly (not just inferred) with a throwaway probe that mocked `Sandbox.prototype.init` to reject and observed `console.warn` still a mock afterward — then deleted the probe.

Fix: moved `vi.spyOn(console, 'warn')` to after the `await createSandbox(...)` succeeds, still ahead of the eval/try block. Added a regression test that forces `createSandbox()` to reject (mocked `Sandbox.prototype.init`) and asserts `console.warn` is back to the original function afterward.

Mutation-tested: reverted the new test back to the old spy-before-construction order (python3 exact-substring replace, count==1) and confirmed it fails (`AssertionError: expected [Function warn] to be [Function warn]`), then restored by file copy. Note the *original* test (`warns once...`) itself never actually rejects `createSandbox()` under real conditions, so reverting just its line order isn't independently caught — the new dedicated test is what pins the invariant.

Sandbox package: 42 -> 43 tests, all passing. `pnpm typecheck` green (34/34 tasks).

Left `bridge.ts:100` and `transpile.ts:64` untouched, per your note.

(Aside: pushed to `origin`/BIMvoice, not `upstream` — this PR's head is `BIMvoice:chore/house-rule-sweep`, so that's where the branch actually lives.)

louistrue:
Verified `8532773c`. The rebase landed the way the ordering constraint needed it to.

`transpile.ts` now reads:

```ts
wasmURL = `https://unpkg.com/esbuild-wasm@${esbuild.version}/esbuild.wasm`;
```

so #2081's fix survived and this PR's `catch (err)` binding sits on top. That was the hazard worth naming: resolving the conflict toward this branch would have reinstated the hardcoded `0.27.3`, and nothing would have caught it — this PR's own tests pass either way, the fallback keeps firing with a mismatched binary, and the only symptom is a degraded transpiler.

Branch is level with `origin/main`, so the green results here are measured against the current base rather than a stale one.

The test-spy fix is also in — `createSandbox()` moved so a rejected initialization can no longer leave `console.warn` stubbed for the rest of the file.

Two threads remain open, and my position on both is unchanged from earlier:

- **the log-budget zero-cost accounting** is real but pre-existing, tracked as #2084 — your diff only made it visible
- **the transpile version mismatch** was real and is now fixed by #2081, which merged

Neither is this PR's to carry. If you resolve those two, this is done from my side.

louistrue:
Heads-up before you spend any more time here: **this PR's green rollup is not real. CI has not run on `8532773c` at all.**

```
neutral  Vercel Agent Review
success  Vercel Preview Comments
```

Two check-runs. A healthy PR in this repo has 15-17 — `Build + WASM + Rust + Node`, `Typecheck`, `Node tests`, `Lint`, `Viewer E2E smoke`, `parity`. None of them are here. They are not failing; they were never scheduled.

The cause is `mergeable=CONFLICTING`. GitHub builds `pull_request` runs against the merge ref, and it cannot construct one for a branch that conflicts with base, so Actions creates no runs. The PR page then shows nothing red, because there is nothing there.

A lot has landed on main today (#2072, #2083, #2085, #2096, and more), so a `git merge origin/main` and push should be all it takes. The lanes will run against the merged head and give you a real signal.

Worth knowing generally, because it is a nasty one: **absence reads exactly like success** on the PR page, and any readiness check shaped "is anything failing or pending?" answers "no" for a PR that has had zero verification. I nearly merged #2091 on this today — same cause, same two-check-run signature — and only the conflict blocking the merge mechanically stopped it. I now assert the core lanes are *present* rather than merely not-red, which is how this one surfaced.

Your two open threads here I have already dispositioned as not-this-PR's-problem; I will resolve them once there is a real CI run to merge on.

BIMvoice:
Merged `upstream/main` in `85a26240`. `MERGEABLE`, check-runs 2 → 7 and climbing, so the lanes are scheduling against a real merged head.

Two conflicts, and they wanted **opposite** resolutions — worth spelling out because your own warning on this PR was about exactly one of them.

## `transpile.ts` — the trap you predicted

```
HEAD (this PR):  } catch (err) { console.warn(...); wasmURL = `...esbuild-wasm@0.27.3/...` }
main (#2081):    } catch        { wasmURL = `...esbuild-wasm@${esbuild.version}/...` }
```

Each side carries a different fix on the same lines, so resolved to **both**: the `catch (err)` binding and the one-time warn from here, with `${esbuild.version}` from #2081. Verified after committing — `grep` finds `esbuild.version` and **no `0.27.3` anywhere in the file**.

This is the resolution you flagged in advance: *"resolving the conflict toward this branch would have reinstated the hardcoded `0.27.3`, and nothing would have caught it — this PR's own tests pass either way, the fallback keeps firing with a mismatched binary, and the only symptom is a degraded transpiler."* Taking "ours" wholesale would have silently undone #2081 while every check stayed green.

## `bridge.ts` — the opposite call

Here `main` is strictly newer and already contains this PR's contribution. Checked rather than assumed: against the merge base, this branch's only change to the file was `sizingWarningShown` + `catch (err)` + a one-time warn. `main` has all three (via #2096) plus `retainSizeableArgs` and the `resetLogs` work from #2099. So I took `main`'s side, and confirmed afterwards that the warn survives with its `err` argument:

```
82:  let sizingWarningShown = false;
122:          if (!sizingWarningShown) {
125:              `[ifc-lite/sandbox] ${retained.replaced} captured log argument(s) could not be sized ...`
```

Nothing of ours dropped — the house-rule fix is in the file, in a better form than we shipped it.

`packages/sandbox` 70/70, `tsc --noEmit` clean.

## On the CI blind spot

Thank you for flagging it here as well as on #2091 — I would not have looked, because the page reads as calm. The generalisable bit is the one you wrote: **absence reads exactly like success**, and any readiness check shaped "is anything failing or pending?" returns "no" for a PR with zero verification. Both of mine that hit it today had the same two-check-run signature (`Vercel Agent Review` neutral, `Vercel Preview Comments` success) against a healthy 15-17, and in both cases the cause was `CONFLICTING` rather than anything wrong with the branch.

louistrue:
Merging. CI is real on this head now — 15 check-runs with all 5 core lanes present, which is the thing that was missing when this branch was `CONFLICTING` earlier and showed a green rollup with every lane absent.

I specifically checked the trap from the #2081/#2080 ordering: **no hardcoded `esbuild-wasm` version came back**. The `transpile.ts` change is purely `catch {` → `catch (err) {` plus logging the cause, the CDN-fallback comment now cites #2081, and `main` still resolves the module through the dynamic import with no pinned version string. A "take theirs" resolution here would have silently reinstated `0.27.3`; it did not.

The logging itself is the right call in each spot, particularly the `?url` one — *"it silently swaps a local asset for a network fetch — say so once"* is exactly the failure worth a warning, because everything keeps working and the only symptom is a surprise CDN dependency at runtime.

PR #2086, opened 2026-08-04
fix(pointcloud): retry the laz-perf wasm load instead of caching the failure

BIMvoice:
Closes the item flagged and deliberately left unfixed in #2083.

## The defect

`laz-source.ts:46-91` assigned `modulePromise` inside `if (!modulePromise)` and **never cleared it on rejection**. A rejected promise caches exactly like a fulfilled one, so one failed wasm fetch made `loadLazPerf()` replay that rejection for the life of the page — every later LAZ open failed identically, **with no fetch attempted**, and only a reload recovered.

## Followed the in-repo remedy rather than inventing one

`duckdb-integration.ts:78-100` clears its memo on failure and rethrows, with a comment naming this exact reason: *"so a transient failure does not poison every future `init()` call"*. Same remedy here, including its choice of no backoff. I did not copy its *structure* — duckdb's memo is per-instance class state, laz-perf's is module-level — so the memo became a small helper instead.

## The seam was the point, and is why this was declined once

`loadLazPerf` was module-private and both dynamic imports resolve identically under vitest, so any test would have passed against the broken code. #2083 correctly refused to ship an unpinned fix.

**`memoizeAsync` is called by production** (`laz-source.ts:118`), not just by tests — an extracted helper the production path bypasses would be the same vacuity in a new place. And the laz-source test drives the real `LazStreamingSource.open()` end to end with a stub emscripten module, so it pins production behaviour rather than the helper in isolation.

Public API unchanged: none of the three new exports is re-exported from `src/index.ts`.

## Two things a naive "clear on reject" breaks

- **Concurrency.** `pending ??=` keeps concurrent callers on one in-flight promise; the reset runs in the rejection handler, after every current holder already has it. They fail together, and only a *later* call retries. Asserted at both the helper level and via three concurrent `open()`s (loader ran once).
- **Synchronous throws.** `Promise.resolve().then(factory)` defers the factory by a microtask, so a sync throw cannot run the `catch` before `pending` is assigned — which would cache the rejection permanently, i.e. **reintroduce this exact bug inside its own fix**. Covered by a test.

## No backoff, deliberately

Retries are user-driven (re-dropping a file), simultaneous opens already collapse onto one attempt, and the in-repo precedent has none. A sequential loop over N failing files does N fetches — strictly better than permanent poisoning, and stated in the changeset rather than left implicit.

## Evidence

| mutation | result |
|---|---|
| restore the original memo semantics (drop the reset) | **4 tests fail** |
| drop the `??=` guard — the naive "clear on reject" | **5 fail**, incl. both concurrency negatives |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, region re-read): 4 failed / 127 passed, including the end-to-end `retries the wasm load after a transient failure instead of poisoning every future open`. Restored 131/133.

The two negative-case tests stayed **green** under that mutation — they are not what drives the RED, which is what makes the pair meaningful.

131 passing (was 123), typecheck 34/34. Baseline was re-measured like-for-like: a first raw read showed `120 passed | 5 skipped`, and the 3 extra skips were `worker-bundle.test.ts`'s `hasBuild` guards un-skipping once `dist/` was built — not a regression.

**Not claimed:** a real wasm fetch failure could not be exercised here — no network fault injection, no browser. The pin is via the injected loader.

### DIFF
--- a/.changeset/laz-perf-load-retry.md
+++ b/.changeset/laz-perf-load-retry.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/pointcloud": patch
+---
+
+Retry the `laz-perf` wasm load after a transient failure instead of poisoning every future `.laz` open.
+
+`LazStreamingSource.open()` shares one lazily-instantiated wasm module across every source, memoised in a module-level `modulePromise`. That promise was assigned once and never cleared, so a rejection was cached exactly like a success: a single failed wasm fetch — a 404 from a misconfigured asset path, an offline blip, a 5xx from the CDN — left every subsequent LAZ open replaying the same rejected promise for the lifetime of the page. The user re-dropped the file and got the identical error, with no fetch ever attempted again; only a reload recovered.
+
+The memo now lives in a small `memoizeAsync` helper that drops the cached promise when the load rejects, so the next `open()` retries. A fulfilled module is still cached forever, and concurrent opens still collapse onto a single in-flight load — dropping several `.laz` files at once instantiates the wasm once, as before. This matches the remedy already applied to `@ifc-lite/query`'s `DuckDBIntegration.init()`, which clears its cached `initPromise` on failure for the same reason.
+
+There is deliberately no backoff: retries here are driven by a user re-opening a file, not by a polling loop, and a batch of simultaneous opens already shares one attempt.
--- a/packages/pointcloud/src/streaming/laz-source.test.ts
+++ b/packages/pointcloud/src/streaming/laz-source.test.ts
@@ -1,0 +1,116 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import {
+  LazStreamingSource,
+  setLazPerfLoaderForTesting,
+  type LasZipInstance,
+  type LazPerfModule,
+} from './laz-source.js';
+
+const RECORD_LEN = 20; // LAS point format 0
+
+/**
+ * A minimal LAS 1.2 header. `open()` parses the header before it touches
+ * laz-perf, so the payload never has to be real LAZ — the stub module
+ * below stands in for the decompressor.
+ */
+function buildLazFile(pointCount: number): Blob {
+  const headerSize = 227;
+  const buf = new ArrayBuffer(headerSize + pointCount * RECORD_LEN);
+  const view = new DataView(buf);
+  view.setUint32(0, 0x4653414c, true); // "LASF"
+  view.setUint8(24, 1);
+  view.setUint8(25, 2);
+  view.setUint16(94, headerSize, true);
+  view.setUint32(96, headerSize, true);
+  view.setUint32(100, 0, true);
+  view.setUint8(104, 0);
+  view.setUint16(105, RECORD_LEN, true);
+  view.setUint32(107, pointCount, true);
+  view.setFloat64(131, 1, true);
+  view.setFloat64(139, 1, true);
+  view.setFloat64(147, 1, true);
+  return new Blob([buf], { type: 'application/octet-stream' });
+}
+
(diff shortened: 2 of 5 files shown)


louistrue:
The memo half is properly done and I would take it as-is. The CLI finding about WASM-boundary coverage is right, but only for the *other* half of this PR — worth separating, because they have different testability.

**The retry seam is genuinely pinned.** Removing `pending = null` from the rejection path fails three named tests:

```
× retries after a rejection instead of caching the failure
× rejects concurrent callers together, then lets a later caller retry
× surfaces a synchronous throw as a rejection and still allows a retry
```

The third is the one I would not have thought to write: `Promise.resolve().then(factory)` defers by a microtask specifically so a *synchronous* throw in the factory cannot run the `catch` before `pending` is assigned. That is a real trap, and it has a test.

Extracting `memoizeAsync` rather than inlining the fix is the right call too, and following `duckdb-integration.ts`'s existing remedy rather than inventing one is exactly the DRY instinct — there were two memos with the same latent defect and now there is one documented pattern.

**Where the finding lands is the loading change, not the memo.** This PR also stops relying on emscripten's `locateFile` fetch and instead pre-fetches the wasm through Vite's `?url` pipeline, handing the bytes over as `Module.wasmBinary`. That is an integration change no mock can validate: if the `?url` import resolves wrong, or emscripten stops honouring `wasmBinary`, the mocked tests stay green and every LAZ open fails in the browser — which is the exact failure this PR is fixing, arriving by a new route.

So the contract test is worth having, but scoped to that: load a real `.laz` through the real `importLazPerf()` and assert points come out. `scripts/test-wasm-contract.mjs` already exists and is wired at `package.json:31`.

I would not hold the PR for it — the memo fix is the urgent half and it is complete — but it is a genuinely different case from the WASM-boundary ask on #2072, which I argued against. There the change asserted no behaviour difference, so a contract test would have pinned the status quo. Here the loading path actually changed, and nothing exercises it.

BIMvoice:
Thanks for taking the memo half — and I think the premise under the contract-test ask is wrong, so I want to put the evidence up rather than quietly agree.

**This PR does not change the loading path. That code is already on `origin/main`.**

```
$ git show upstream/main:packages/pointcloud/src/streaming/laz-source.ts | grep -n 'wasmBinary\|fetchLazPerfWasm\|?url'
58:      // Pre-fetch the wasm via Vite's `?url` asset pipeline (hashed,
60:      // as `Module.wasmBinary` so its own fetch is skipped entirely.
61:      const wasmBinary = await fetchLazPerfWasm();
88:      return factory({ wasmBinary });
94: async function fetchLazPerfWasm(): Promise<Uint8Array> {
95:   // `?url` triggers Vite's asset pipeline so the .wasm ends up in the
```

The `?url` pre-fetch and the `Module.wasmBinary` hand-off predate this branch. They appear in the diff as **both `+` and `-`** because extracting `importLazPerf()` dedented them out of the `if (!modulePromise)` block — a move, not an introduction. `git diff upstream/main...HEAD` shows the same lines removed at their old indentation and added at the new one.

So the distinction you drew from #2072 does not separate the two cases the way it looks. There you argued a contract test would pin the status quo because the change asserted no behaviour difference — and **that is equally true here**: the memo change alters *when* a load is retried, not *how* the wasm is fetched. Nothing about the `?url` resolution or emscripten's `wasmBinary` handling is different before and after.

**Where I think you are still right:** the loading path genuinely has no end-to-end coverage, and the failure mode you describe is real — if `?url` resolves wrong or emscripten stops honouring `wasmBinary`, the mocked tests stay green and every LAZ open fails in the browser. That is a live gap on `main`, not one this PR opens. Worth its own issue against the existing `scripts/test-wasm-contract.mjs` harness, and I will file it framed that way — as pre-existing, with the concrete assertion you named (load a real `.laz` through the real `importLazPerf()` and assert points come out).

If I have misread the diff, say so and I will take the test here instead — but I would rather not attach it to this PR on a premise I cannot reproduce.

On the memo half: agreed on all of it, and the synchronous-throw test is the one I would single out too. `Promise.resolve().then(factory)` deferring by a microtask is exactly what stops a sync throw running the `catch` before `pending` is assigned — which would cache the rejection permanently and reintroduce this bug inside its own fix.

Issue #2084, opened 2026-08-04
Sandbox: unserializable log args bypass the host byte budget (bounded only by entry count)


louistrue:
## What happens

`buildConsole` caps captured log output two ways — `MAX_LOG_ENTRIES` (1000) and `MAX_TOTAL_BYTES` (4MB). An entry whose args cannot be `JSON.stringify`'d contributes **zero** to the byte budget while still being retained:

```ts
let entryBytes = 0;
try { entryBytes = JSON.stringify(nativeArgs)?.length ?? 0; }
catch { entryBytes = 0; /* unserializable args — skip cost accounting */ }
totalBytes += entryBytes;
logs.push({ level, args: nativeArgs, timestamp: Date.now() });
```

(`packages/sandbox/src/bridge.ts`, around the `buildConsole` log handler.)

So a sandboxed script that logs unserializable values repeatedly is bounded only by the 1000-entry cap, not by the 4MB budget. Up to 1000 arbitrarily large retained objects, none of which the byte cap can see.

## Reachable

`BigInt` is the easy trigger — `JSON.stringify` throws on it outright, so `console.log(1n)` in a loop is enough. Cyclic objects behave the same way. Scripts in this sandbox are user-supplied, which is the entire threat model the two caps exist for.

## Not a regression

Pre-existing. Raised by CodeRabbit on #2080, which only changed `catch {` to `catch (err) {` and added a one-time warning — it made the condition visible without altering it. Deliberately not fixed there to keep a house-rule sweep from turning into a budget-accounting change.

## Possible directions

1. Reject the entry outright when sizing fails — the script gets the same truncation notice it would get at any other limit.
2. Append a bounded placeholder (`'[unserializable]'`) and charge its length, so the entry is preserved as a signal without retaining the object.
3. Charge a fixed pessimistic cost per unsizable entry, so the byte budget still converges.

(2) keeps the most diagnostic value. Whichever is chosen, the test should assert the **host budget still trips** under a loop of unserializable logs — that is the property currently unenforced.

## Note on the neighbouring design

The one-notice-per-context warning added in #2080 should stay however this is fixed: a per-occurrence log would let the same loop flood the host console, converting one resource issue into another.

BIMvoice:
Duplicate of #2087, which **#2096 merged** and fixed — an unsizeable log argument is now replaced by bounded text and charged for, rather than retained for free against `MAX_LOG_ENTRIES`.

[...]

PR #2089, opened 2026-08-04
fix(viewer): quarantine unreadable stored data instead of deleting it

BIMvoice:
From the #2085 sweep's "left for a maintainer decision" list — the two where the remedy needs no UX call.

## Reachability first, because it is weaker than #2085 implied

**All three sites are hardening, not defects reachable today.** Stating that up front rather than letting the fix imply otherwise:

- **`persistence.ts:145`** — the only writer of `PRESETS_KEY` is `savePresets`, which writes `JSON.stringify` output, so the app cannot produce a value that makes `JSON.parse` throw. The *partial* read is also unreachable: `createClashPreset` and `importPresets` both validate, and `updateClashPreset` — which does not — is only reachable behind a Save button gated on `draftValid` (`ClashSettingsDialog.tsx:123-124`).
- **`persistence.ts:298`** — same shape. `saveReviews` refuses over `MAX_REVIEWS` rather than truncating, and the status filter only drops values the app never writes. It becomes reachable the moment a new status ships and a user runs two app versions.
- **`saved-filters.ts:76`** — `parseFilterRules` cannot throw and `getItem` sits outside the `try`, so only externally-corrupted JSON reaches it.

External corruption is the shared trigger for all three: a mid-write browser crash, devtools editing, an extension, or a downgrade to an older schema.

**`saved-filters.ts` is fixed as a correctness bug regardless of reachability**, because it *deletes on read failure* rather than merely enabling a later overwrite — and `loadSavedFilters()` alone triggers it, with no save required.

## One latent trigger worth naming

The `persistence.ts` catch also swallows a throw from `builtinPresetIds()` (`:143`), and the comment at `:79-84` records that exact `CLASH_RULE_PRESETS` chunk-ordering hazard **actually firing on boot**. Deferring it into the loader turned a loud boot crash into a swallowed one whose next consequence is `savePresets` writing `{presets: []}`.

## The remedy: move it aside, don't refuse forever

On a read failure the raw bytes are copied to `<key>:unreadable` and `<key>` is cleared. The feature resets to defaults, stays usable, and nothing is lost.

I rejected "mark not-loaded and refuse to save" — it leaves the panel permanently read-only with **no in-app way out**, which is the trap of trading one data-loss bug for a usability dead end. Refusal is kept only for when quarantine *itself* fails (quota): there a blocked feature beats destroyed data, surfaced through the existing `toast.error(result.message)` path, and the flag clears on the next clean read.

**Recovery path:** the original bytes sit at `ifc-lite-clash-presets:unreadable` (etc.) with a `console.warn` naming the key — repair the JSON in devtools and copy it back, or re-import from an `exportPresets` file. An explicit `clearSavedFilters` drops the backups too: a user wipe may delete, a failed read may not.

## Evidence

| mutation | result |
|---|---|
| restore the destructive `removeItem` | **4 fail** |
| bare `catch {}` in `readStoredPresets` | 3 fail |
| bare `catch {}` in `loadReviews` | 2 fail |
| drop either save-refusal guard | 1 fail each |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, line re-read): 4 failed, and `clearSavedFilters still wipes everything` correctly **survived** — which is what shows the fix did not just stop all deletion.

**One mutation initially killed nothing** (the reviews-side refusal was an untested branch), so the quota test was extended before re-running rather than reported as a survivor.

**Two vacuity traps found in my own tests:** the quota cases first rejected *every* `setItem`, which made `safeStorage()`'s probe fail so the module saw no storage at all — they would have passed against broken code. They now reject only the backup write.

Negative cases, all passing in RED too so they cannot be what the fix "fixed": normal round trip unaffected; after a failed read a save still succeeds **and reads back**; a user-initiated delete still deletes.

2582 passing (was 2571, measured from pristine sources with the new test files removed), typecheck 34/34.

## Deliberately left

`loadSettings` (`persistence.ts:236`) has the identical shape. The payload is six scalars the user can re-set in seconds, and a refusal there would silently no-op `persistSettings()` on every slider move — worse than the thing it guards against.

### DIFF
--- a/apps/viewer/src/components/viewer/SearchModal.filter.builder.tsx
+++ b/apps/viewer/src/components/viewer/SearchModal.filter.builder.tsx
@@ -45,6 +45,7 @@ import {
   deleteSavedFilter,
   type SavedFilterPreset,
 } from '@/lib/search/saved-filters';
+import { toast } from '@/components/ui/toast';
 import { RuleRow, RULE_KIND_LABEL } from './SearchModal.filter.editors';
 
 export function SearchModalFilterBuilder() {
@@ -191,7 +192,13 @@ export function SearchModalFilterBuilder() {
     // eslint-disable-next-line no-alert
     const name = window.prompt('Save filter as…', '');
     if (!name) return;
-    setSavedPresets(saveFilter(name, filter.combinator, filter.rules));
+    const result = saveFilter(name, filter.combinator, filter.rules);
+    setSavedPresets(result.presets);
+    // A refused write used to return the in-memory catalog as though saved —
+    // the user saw the filter and lost it next session (#2089).
+    if (!result.persisted) {
+      toast.error('Filter could not be saved — browser storage is unavailable or full.');
+    }
   }, [filter.combinator, filter.rules]);
 
   const handleLoadPreset = useCallback((preset: SavedFilterPreset) => {
@@ -203,7 +210,11 @@ export function SearchModalFilterBuilder() {
   }, [filter.limit, setSearchFilter]);
 
   const handleDeletePreset = useCallback((name: string) => {
-    setSavedPresets(deleteSavedFilter(name));
+    const result = deleteSavedFilter(name);
+    setSavedPresets(result.presets);
+    if (!result.persisted) {
+      toast.error('Filter could not be deleted — browser storage is unavailable or full.');
+    }
   }, []);
 
   return (
--- a/apps/viewer/src/lib/clash/persistence.ts
+++ b/apps/viewer/src/lib/clash/persistence.ts
@@ -26,6 +26,7 @@ import {
   type ClashSeverity,
 } from '@ifc-lite/clash';
 import { downloadFile } from '../export/download.js';
+import { optionalLocalStorage, preserveUnreadableEntry } from '../storage/unreadable-entry.js';
 
 /** A built-in or user-defined clash rule preset, with editor/runtime flags. */
 export type ClashPreset = ClashRulePreset & { enabled: boolean; builtin: boolean };
@@ -45,7 +46,7 @@ export interface ClashGlobalSettings {
 
 export type SaveResult =
   | { ok: true }
-  | { ok: false; reason: 'quota' | 'serialize' | 'too_many'; message: string };
+  | { ok: false; reason: 'quota' | 'serialize' | 'too_many' | 'unreadable'; message: string };
 
 const PRESETS_KEY = 'ifc-lite-clash-presets';
 const SETTINGS_KEY = 'ifc-lite-clash-settings';
@@ -86,6 +87,30 @@ let builtinPresetIdsCache: Set<string> | null = null;
 function builtinPresetIds(): Set<string> {
   return (builtinPresetIdsCache ??= new Set(CLASH_RULE_PRESETS.map((p) => p.id)));
 }
+
+/**
+ * Keys whose stored value we failed to read *and* failed to move aside. Both
+ * loaders below degrade to defaults on a read failure, so without this the next
+ * ordinary edit would serialize those defaults over data we never managed to
+ * see. Entries are added by the loader and cleared by a later clean read.
+ */
+const unwritableKeys = new Set<string>();
+
+/** Handle a loader throwing: preserve what is stored, or block writes to `key`. */
+function onReadFailure(key: string, cause: unknown): void {
+  if (preserveUnreadableEntry(optionalLocalStorage(), key, cause)) unwritableKeys.delete(key);
+  else unwritableKeys.add(key);
+}
+
+/** A refusal to overwrite a value we could neither read nor preserve. */
+function refuseOverwrite(what: string): SaveResult {
+  return {
+    ok: false,
+    reason: 'unreadable',
+    message: `Stored ${what} could not be read and could not be backed up — they were left untouched.`,
+  };
+}
+
 const SEVERITIES: ClashSeverity[] = ['critical', 'major', 'minor', 'info'];
 const GROUP_BYS: ClashSettingsGroupBy[] = ['severity', 'rule', 'typePair'];
 
@@ -122,6 +147,7 @@ function isValidStoredPreset(p: unknown): p is ClashPreset {
 /** Read stored presets, accepting the versioned wrapper or a legacy bare array. */
 function readStoredPresets(): ClashPreset[] {
   try {
+    unwritableKeys.delete(PRESETS_KEY);
     const raw = localStorage.getItem(PRESETS_KEY);
     if (!raw) return [];
     const parsed: unknown = JSON.parse(raw);
@@ -142,7 +168,8 @@ function readStoredPresets(): ClashPreset[] {
         enabled: p.enabled !== false,
         builtin: builtinPresetIds().has(p.id),
       }));
-  } catch {
+  } catch (err) {
+    onReadFailure(PRESETS_KEY, err);
     return [];
   }
 }
@@ -198,6 +225,7 @@ export function presetsToStore(presets: ClashPreset[]): ClashPreset[] {
 
 /** Persist only custom presets + modified built-ins (quota-safe). */
 export function savePresets(presets: ClashPreset[]): SaveResult {
+  if (unwritableKeys.has(PRESETS_KEY)) return refuseOverwrite('clash rules');
   const custom = presets.filter((p) => !p.builtin);
   if (custom.length > MAX_PRESETS) {
     return { ok: false, reason: 'too_many', message: `Too many custom rules (max ${MAX_PRESETS}).` };
@@ -276,6 +304,7 @@ function isReviewStatus(v: unknown): v is ClashReviewStatus {
 export function loadReviews(): Map<string, ClashReview> {
   const map = new Map<string, ClashReview>();
   try {
+    unwritableKeys.delete(REVIEWS_KEY);
     const raw = localStorage.getItem(REVIEWS_KEY);
     if (!raw) return map;
     const parsed: unknown = JSON.parse(raw);
@@ -295,7 +324,11 @@ export function loadReviews(): Map<string, ClashReview> {
       // Skip default entries a stale writer may have left behind.
       if (isMeaningfulReview(review)) map.set(key, review);
     }
-  } catch {
+  } catch (err) {
+    // A half-built map is as destructive as an empty one: the next triage edit
+    // writes it back and the entries we never parsed are gone. (#2085)
+    map.clear();
+    onReadFailure(REVIEWS_KEY, err);
     return map;
   }
   return map;
@@ -303,6 +336,7 @@ export function loadReviews(): Map<string, ClashReview> {
 
 /** Persist reviews (default-state entries pruned, capped, quota-safe). */
 export function saveReviews(reviews: Map<string, ClashReview>): SaveResult {
+  if (unwritableKeys.has(REVIEWS_KEY)) return refuseOverwrite('clash reviews');
   const entries: Record<string, ClashReview> = {};
   let count = 0;
   for (const [key, review] of reviews) {
--- a/apps/viewer/src/lib/clash/persistence.unreadable.test.ts
+++ b/apps/viewer/src/lib/clash/persistence.unreadable.test.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { describe, it, beforeEach } from 'node:test';
+import { describe, it, beforeEach, after } from 'node:test';
 import assert from 'node:assert';
 import type { ClashReview } from '@ifc-lite/clash';
 import {
@@ -50,6 +50,22 @@ describe('clash persistence: an unreadable entry is never overwritten', () => {
     ls = new MemoryStorage();
     g.localStorage = ls;
     // Clear the module-level "unwritable" flags via a clean read of empty storage.
+    buildInitialPresets();
+    loadReviews();
+  });
+
+  after(() => {
+    // The last test in this suite deliberately leaves both keys latched
+    // unwritable (the backup write itself fails). `apps/viewer` runs every
+    // test file in one `tsx --test` process, so persistence.ts's
+    // module-level `unwritableKeys` would otherwise carry that latch into
+    // whichever test file runs next and imports it — `savePresets`/
+    // `saveReviews` check the flag before ever touching storage, so a later
+    // file's first save (not preceded by its own clean read) would be
+    // spuriously refused for storage it never actually corrupted. A clean
+    // read against ordinary storage clears the flags the same way `beforeEach`
+    // does above.
+    g.localStorage = new MemoryStorage();
     buildInitialPresets();
     loadReviews();
   });
--- a/apps/viewer/src/lib/clash/persistence.unreadable.test.ts
+++ b/apps/viewer/src/lib/clash/persistence.unreadable.test.ts
@@ -1,0 +1,144 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, beforeEach } from 'node:test';
+import assert from 'node:assert';
+import type { ClashReview } from '@ifc-lite/clash';
+import {
+  buildInitialPresets,
+  savePresets,
+  loadReviews,
+  saveReviews,
+  type ClashPreset,
+} from './persistence.js';
+
+class MemoryStorage {
+  readonly store = new Map<string, string>();
+  getItem(key: string): string | null { return this.store.get(key) ?? null; }
+  setItem(key: string, value: string): void { this.store.set(key, value); }
+  removeItem(key: string): void { this.store.delete(key); }
+}
+
+const g = globalThis as { localStorage?: unknown };
+const PRESETS_KEY = 'ifc-lite-clash-presets';
+const REVIEWS_KEY = 'ifc-lite-clash-reviews';
+
+/** Truncated JSON — the shape a half-written or externally mangled entry has. */
+const CORRUPT_PRESETS = '{"schemaVersion":1,"presets":[{"id":"custom-1","name":"My rule"';
+const CORRUPT_REVIEWS = '{"schemaVersion":1,"reviews":{"rule-a G1 G2":{"status":"resol';
+
+/** True when `raw` is still somewhere in storage (original key or a backup). */
+function survives(ls: MemoryStorage, raw: string): boolean {
+  return [...ls.store.values()].includes(raw);
+}
+
+const customPreset: ClashPreset = {
+  id: 'custom-new',
+  name: 'Ducts vs beams',
+  description: '',
+  severity: 'major',
+  selectorA: 'IfcDuctSegment',
+  selectorB: 'IfcBeam',
+  enabled: true,
+  builtin: false,
+};
+
+describe('clash persistence: an unreadable entry is never overwritten', () => {
+  let ls: MemoryStorage;
+  beforeEach(() => {
+    ls = new MemoryStorage();
+    g.localStorage = ls;
+    // Clear the module-level "unwritable" flags via a clean read of empty storage.
+    buildInitialPresets();
+    loadReviews();
+  });
+
+  it('preserves unreadable presets across the save that follows the failed read', () => {
+    ls.setItem(PRESETS_KEY, CORRUPT_PRESETS);
+
+    // The read degrades to built-ins only — the user's customs are not visible.
+    const presets = buildInitialPresets();
+    assert.ok(presets.every((p) => p.builtin), 'expected only built-ins after a failed read');
+
+    // ...and the very next edit persists that degraded list.
+    assert.deepStrictEqual(savePresets([...presets, customPreset]), { ok: true });
+
+    assert.ok(survives(ls, CORRUPT_PRESETS), 'the unreadable presets blob was destroyed by the save');
+  });
+
+  it('preserves unreadable reviews across the save that follows the failed read', () => {
+    ls.setItem(REVIEWS_KEY, CORRUPT_REVIEWS);
+
+    assert.strictEqual(loadReviews().size, 0);
+    const map = new Map<string, ClashReview>([['rule-b G3 G4', { status: 'accepted' }]]);
+    assert.deepStrictEqual(saveReviews(map), { ok: true });
+
+    assert.ok(survives(ls, CORRUPT_REVIEWS), 'the unreadable reviews blob was destroyed by the save');
+  });
+
+  // ── Negative cases: the guard must not turn into "never write again" ────────
+
+  it('still round-trips normally when the stored entry is readable', () => {
+    assert.deepStrictEqual(savePresets([customPreset]), { ok: true });
+    const reloaded = buildInitialPresets().filter((p) => !p.builtin);
+    assert.deepStrictEqual(reloaded.map((p) => p.id), ['custom-new']);
+  });
+
+  it('keeps saving after an unreadable read — the panel stays usable', () => {
+    ls.setItem(PRESETS_KEY, CORRUPT_PRESETS);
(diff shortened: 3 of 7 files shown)


louistrue:
Quarantining unreadable data instead of deleting it is the right instinct — a user's saved filters are not the thing to discard when a parse fails. Three items before merge, two flagged in review and one I would add.

## 1. The collision contradicts the comment directly above it

```ts
// Never clobber an earlier preserved copy — it may hold the older, better data.
if (storage.getItem(backup) !== null) backup = `${backup}:${Date.now()}`;
storage.setItem(backup, raw);
```

Two quarantines of the same key inside one millisecond pick the same timestamped key, and the second `setItem` clobbers the first — the precise thing the comment says never happens. A counter, or appending to a list, holds regardless of clock resolution.

This is the shape that keeps recurring across today's PRs: a comment stating an invariant the code does not enforce. Here the two are three lines apart.

## 2. A refused write is invisible to the caller

```ts
function writeRaw(list: SavedFilterPreset[]): void {
  const ls = safeStorage();
  if (!ls || catalogUnwritable) return;
```

Once `catalogUnwritable` latches, every `saveFilter` and `deleteSavedFilter` becomes a no-op — and the caller receives the in-memory catalog as though the write succeeded. The user saves a filter, gets no error, sees it in the list, and it is gone next session.

Refusing to overwrite an unreadable catalog is correct. Returning success for a write that did not happen is not. The caller needs a signal it can surface.

## 3. A silent `catch {}` — in the middle of the sweep that forbids them

```ts
} catch {
  // Quota exceeded — swallow; the next save attempt may succeed.
}
```

`AGENTS.md:17` lists "no silent `catch {}` (log or rethrow)" among the self-policed TypeScript rules, and #2080 exists specifically to close these. This one is new in this PR.

The comment is also optimistic in a way worth testing: if the quota is genuinely full, "the next save attempt may succeed" is false for every subsequent attempt, and the user gets the same invisible no-op as item 2 by a second route.

## Not disputing the rest

The adversarial pass found no blockers on the core change, and the underlying instinct is sound. All three items above are about what happens when the fix's own failure paths fire — which is exactly where a quarantine mechanism earns or loses its keep.

BIMvoice:
All three fixed in 3c63067a. Item 3 is the one worth dwelling on: **a silent `catch {}` introduced by the PR whose subject is closing silent catches.** No defence for that one.

**1. The collision.** You were right that it contradicted the comment three lines above. Now a counter rather than `Date.now()`, which holds regardless of clock resolution — three quarantines of one key yield `:unreadable`, `:unreadable:2`, `:unreadable:3`, verified rather than assumed.

**2. The refused write.** This is the same defect class as the lens slice, one layer down — and I had just fixed that one, which makes missing it here worse rather than better. `writeRaw` now returns a boolean; both exported functions return `{ presets, persisted }`; the two `SearchModal` call sites surface a toast.

One judgement call: the early-return paths (rejected name, name not found) report `persisted: true` deliberately — nothing was asked of storage, so nothing is unpersisted. Reporting failure there would train the user to ignore the toast.

**3. The silent catch**, and its comment was optimistic in exactly the way you said: "the next save attempt may succeed" is false for every later attempt when the quota is genuinely full — which is the case that matters. It now binds and logs the error, and says why the optimism was wrong.

The three no-write paths — blocked storage, latched refusal, failed write — now log distinctly while all reading as "not saved" to the caller, so the console distinguishes them without the API pretending they differ.

21 passing across the affected suites, typecheck 34/34.

Your framing is the part I want to keep: all three were about what happens **when the fix's own failure paths fire**. The quarantine mechanism is exactly where that matters, and I had tested the happy path of the recovery while leaving the recovery's own failures unexamined.

louistrue:
Verified `3c63067a`. All three are properly closed, and the counter is the right shape — collisions are now impossible regardless of clock resolution, not merely unlikely. The comment records why, which is what stops someone reverting to `Date.now()` as a simplification.

`writeRaw` returning a boolean and warning with a repair instruction ("Repair or remove the backup to resume saving") is better than just returning the flag — it tells the user what to do rather than only that something failed. And the quota `catch` now explains why the old "the next save may succeed" was wrong: if the quota is genuinely full, every later attempt fails identically, so the optimistic wording was misleading for exactly the case that matters.

**One gap: the collision fix is not pinned.** Reverting the counter to `${backup}:${Date.now()}` leaves `persistence.unreadable.test.ts` at 6/6 green, and no test file in `apps/viewer/src` references `preserveUnreadableEntry` or `unreadableKey` directly.

That is worth closing precisely *because* of what you changed. `Date.now()` was hard to test — forcing two quarantines into one millisecond means faking the clock. A counter is trivially testable: quarantine the same key three times against a stubbed storage and assert the backups land at `…`, `…:2`, `…:3` with all three payloads intact. The fix made itself testable and the test did not follow.

Same note as the sweep more broadly: silent-catch counts went **down** in this PR (persistence 8 -> 6, saved-filters 3 -> 1), and the one new `catch {}` at `unreadable-entry.ts:113` is the defensible kind — `localStorage` access throwing outright under a blocking cookie policy, where there is nothing to log and null is the answer. Worth a one-line comment saying so, since a future sweep will otherwise flag it again.

BIMvoice:
Both closed in a54de1b4. You were right, and the framing is the useful part.

**"The fix made itself testable and the test did not follow"** — that is exactly what happened. I demonstrated the counter in a throwaway `node -e` script and pushed without pinning it, which is the same instrument-measuring-nothing shape this whole sweep exists to close. `Date.now()` was awkward to test because forcing two quarantines into one millisecond means faking the clock; the counter is trivially testable, and I should have written the test *because* the fix made it easy rather than treating the demo as sufficient.

Three quarantines of one key run synchronously; all three payloads survive under distinct keys and the live key is cleared. **Reverting to `${backup}:${Date.now()}` fails it** (`REPLACEMENTS=1`, region re-read), so the property is pinned rather than argued.

Two more went in alongside: a blocked backup write reports `false` and leaves the original intact — the caller relies on that to refuse a later overwrite — and `forgetEntryAndBackups` removes every suffixed copy.

**That last one initially failed, and my stub was at fault rather than the code.** `forgetEntryAndBackups` enumerates suffixed copies through `length`/`key(i)`, which real `localStorage` exposes and my double did not, so it silently skipped that branch. Adding them makes the double faithful — without them the wipe test would have passed for the wrong reason, which is the third vacuity trap of the day and the second one where the *stub* was the problem.

**On `unreadable-entry.ts:109`** — agreed, and documented. It is the one shape the rule is not aimed at: merely *reading* `globalThis.localStorage` throws under a blocking cookie policy, `null` already says everything there is to say, every caller degrades on `null`, and logging would fire on every call for a browser configuration the user chose. Written down so a future sweep does not re-flag it, which is the point you were making.

Worth noting the direction of travel you flagged: silent-catch counts went **down** in this PR (persistence 8 → 6, saved-filters 3 → 1) while adding one defensible-and-now-annotated case. That is the shape I would want the remaining sweep to keep.

24 passing across the three affected suites.

louistrue:
`a54de1b4` closes the gap I raised — the counter is now pinned, which is exactly the case that became testable only because you replaced `Date.now()`. But **Typecheck is red on the new test file**, and the cause is worth more than the one-line fix.

```
src/lib/storage/unreadable-entry.test.ts(24,9): error TS2353:
  Object literal may only specify known properties, and 'length' does not exist
  in type 'UnreadableEntryStorage & { map: Map<string, string>; }'.
```

`UnreadableEntryStorage` declares only `getItem`/`setItem`/`removeItem`. The stub adds `length` and `key()` because `forgetEntryAndBackups` needs them — and your own comment in the test says why:

> A stub without them silently skips that branch, so the double has to carry them or the wipe test passes for the wrong reason.

That comment is right, and it points at the real issue. The production code reaches those members through a widening cast:

```ts
const enumerable = storage as UnreadableEntryStorage & { length?: number; key?(i: number): string | null };
if (typeof enumerable.length !== 'number' || typeof enumerable.key !== 'function') return;
```

So the interface says the enumeration members do not exist, the implementation casts to say they might, and the test cannot express a double that has them. The cast is what let the interface stay wrong.

Adding them as optional to `UnreadableEntryStorage` fixes the typecheck, deletes the cast, and makes the runtime guard a genuine narrowing rather than a re-assertion:

```ts
export interface UnreadableEntryStorage {
  getItem(key: string): string | null;
  setItem(key: string, value: string): void;
  removeItem(key: string): void;
  readonly length?: number;
  key?(i: number): string | null;
}
```

`Build + WASM + Rust + Node` is the aggregate reporting the same failure, so both red rows clear together.

Worth noting this is a different failure from the one on your previous head — that was the `collab-server` clock flake (#2094), unrelated to you. This one is real and in your diff.

BIMvoice:
Fixed in 985230bb, and thank you for catching it — **that red row was mine and I missed it.** I ran the three affected suites after adding the test file and never re-ran typecheck, so I shipped a type error while reporting green. Distinguishing it from the `collab-server` clock flake on the previous head saved me from dismissing it as the same noise.

**Your diagnosis is the part worth keeping: the cast is what let the interface stay wrong.**

The interface declared only `getItem`/`setItem`/`removeItem`; the implementation reached the enumeration members through a widening cast; and the runtime guard then re-asserted what the cast had already claimed. So a faithful double was *untypeable* — which is precisely what my own test comment noticed from the other side:

> A stub without them silently skips that branch, so the double has to carry them or the wipe test passes for the wrong reason.

I wrote that, and did not follow it back to its cause. The test told me the interface was lying and I treated it as a fact about stubs.

Both members are now optional on `UnreadableEntryStorage`, the cast is deleted, and the runtime check is a genuine narrowing. One wrinkle worth noting: destructuring `key` off the storage object loses its receiver, so it needs `keyAt.call(storage, i)` — `typeof enumerable.key !== 'function'` had been hiding that too, since the old code called it as a method.

typecheck 34/34, 24 passing across the three suites.

On #2095 — agreed that the second test was the useful one, and your framing of *why* is right: the two errors did not merely get wiped by a later eval, they **were never distinct**, so an embedder collecting failures across scripts would have seen every entry showing the last script's logs. That is a different and worse bug than the one filed, and it only surfaced because the copy semantics got tested rather than assumed.

Noted on the per-sandbox `totalBytes` latch — a reused sandbox letting one log-heavy script permanently truncate every later script's output is a genuinely different mechanism from #2084. Happy to take it once filed.

louistrue:
Verified `985230bb`. The cast is gone and the interface now declares what the code actually uses:

```ts
/**
 * Enumeration members of the real `Storage` interface. Optional because a
 * caller may pass a narrower double — `forgetEntryAndBackups` needs them to …
 */
```

`unreadable-entry.ts` typechecks clean, and no `as UnreadableEntryStorage &` widening remains.

That is the better fix over silencing the test. The cast was what let the interface stay wrong: the type said those members did not exist, the implementation asserted they might, and the first honest double collided with the gap. Now the runtime guard (`typeof enumerable.length !== 'number'`) narrows a declared optional rather than re-checking something the type denied — so it means what it appears to mean.

For the record on the remaining local noise: `tsc --noEmit` in a fresh worktree still reports 186 errors, all cascading from unbuilt workspace deps (`@ifc-lite/extensions` 40, `@ifc-lite/diff` 16, `@ifc-lite/merge` 13, …). The handful that are not `Cannot find module` are all in `AuditLogPanel.tsx`, which imports the missing `@ifc-lite/extensions` — so its `unknown` types are downstream of the same cause. CI builds first, so none of that reaches the Typecheck job.

All three of my items on this PR are now closed: the counter, the invisible refusal, and the silent catch — plus the counter test and this typecheck. Nothing outstanding from me.

louistrue:
Verified `985230bb`. **Both open threads are stale** — you already fixed both. But the fix's own signal is unpinned, and that is worth one test.

## Both greptile findings describe code that is no longer there

**"Timestamped backups can collide"** — there is no timestamp. `preserveUnreadableEntry` uses a counter, and the comment states the collision as the reason:

> A counter rather than `Date.now()`: two quarantines of the same key inside one millisecond pick the same timestamp, so the second `setItem` would overwrite the first — the exact thing this guard exists to stop.

**"Saved-filter refusal stays silent"** — it does not. `saveFilter`/`deleteSavedFilter` return `{ presets, persisted }`, and `SearchModal.filter.builder.tsx` surfaces both:

```tsx
if (!result.persisted) {
  toast.error('Filter could not be saved — browser storage is unavailable or full.');
}
```

Resolving both.

## The `persisted` contract is not pinned by anything

`saved-filters.test.ts` contains **zero** occurrences of `persisted`. Fifteen tests, all green, none of them touching the return-value change that this PR is built on.

So this passes:

```js
// writeRaw still runs; its verdict is discarded
const persisted = (writeRaw(sortedByRecency), true);
```

```
### MUTATION: writeRaw always reports success (restores the silent loss)
# pass 15
# fail 0
```

That mutation is precisely the bug the PR describes — the user sees the filter appear, and it is gone next session — and the suite is silent on it. Every toast you added would go dead with it.

The existing tests are genuinely good at what they do cover: `does not delete the stored catalog when it fails to parse`, `does not overwrite the stored catalog with the save that follows`, `refuses to write when the unreadable catalog could not even be backed up`. They pin the *data* half of the contract thoroughly. It is the *reporting* half — the half the UI now depends on — that nothing holds.

`refuses to write when the unreadable catalog could not even be backed up` is already set up for it. It reaches the exact state where `persisted` must be `false`; it just does not look at the return value. One assertion there, plus a `persisted === true` check on an ordinary save so the flag is not simply always false, and the mutation dies both ways.

Nothing else outstanding on this one — everything else checks out and the storage module's reasoning is unusually clear, particularly the note on why `optionalLocalStorage`'s silent catch is the one shape AGENTS.md:17 is not aimed at.

BIMvoice:
Fixed in 0b23efab.

Confirmed your mutation first: `const persisted = (writeRaw(sortedByRecency), true);` — 15/15 pass on the unmodified suite (all green, nothing looked at the return value).

Added two assertions in `saved-filters.test.ts`:
- `refuses to write when the unreadable catalog could not even be backed up` (the test you pointed at) now asserts `result.persisted === false`.
- `saves a preset and reads it back sorted by name` now asserts `persisted === true` on an ordinary save, so a hardcoded-`false` flag can't slip through a check that only looks for `false`.

Mutation evidence:
- Your exact mutation (`persisted` always `true`): 14 pass / 1 fail (was 15/15 before this fix).
- The mirror mutation (`persisted` always `false`): also 14 pass / 1 fail.

Full viewer suite: 2585 pass / 2 skipped / 0 fail (unchanged pass count — no new tests, two new assertions inside existing ones). `pnpm typecheck` clean.

Left the two greptile threads alone since you already flagged them stale.

louistrue:
Hold — CodeRabbit found a hole in this PR's own guarantee, and I reproduced it. (Hosted was `Review rate limited` again; this is the CLI, 6 findings, triaged below.)

## A valid-JSON non-array catalog is destroyed

`readRaw` preserves a catalog it cannot parse, but returns early for one it *can* parse into a non-array:

```ts
const parsed: unknown = JSON.parse(raw);
if (!Array.isArray(parsed)) return [];   // <- no preserveUnreadableEntry, no catalogUnwritable
```

Only the `catch` routes through `preserveUnreadableEntry`. Probe, same code path both times, only the stored bytes differ:

```
CORRUPT   -> original overwritten: true | backup kept: true
NON-ARRAY -> original overwritten: true | backup kept: false
```

`{"presets":[{"name":"Important"}]}` is gone. No `:unreadable` copy, nothing to recover from — exactly the loss your module docblock says the design exists to prevent:

> A loader that returns a default (or a partial result) on a parse error hands the caller a state that looks empty; the next ordinary edit then serializes that empty state straight over the entry, and whatever was there is gone for good.

That is precisely what happens here; the value just failed a shape check instead of a parse. And it is a realistic shape — a future format that wraps the array in an object, or a hand-edited file — not a contrived one. Routing it through `preserveUnreadableEntry` with the same `catalogUnwritable` handling is the fix, and the existing helper already does the work.

## The `persisted` fix is otherwise verified

Both directions now die, as your comment predicted:

| mutation | result |
|---|---|
| `persisted` always `true` | `refuses to write when the unreadable catalog could not even be backed up` fails |
| `persisted` always `false` | `saves a preset and reads it back sorted by name` fails |

Your note is the reason the second assertion earns its place: *"not merely absent-of-false, or the flag could be hardcoded `false` and a check that only looks for `false` elsewhere would still pass."*

## The other five CodeRabbit findings

- **Docblock on `SavedFilterMutation` is stale** (minor, worth taking): it does not say that on a failed write the returned `presets` are the prior readable catalog, not the attempted mutation. Small, and exactly the kind of claim that outlives the code.
- **Test isolation in `persistence.unreadable.test.ts`** (minor, worth taking): module-level unwritable flags are not reset after the suite. Same shape as the `globalThis.fetch` leak I flagged on #2070 — `apps/viewer` runs every test file in one `tsx --test` process, so module state carries.
- **Component tests for the toast paths** (major): reasonable, but the `persisted` contract is now pinned at the library boundary where it is cheap to test. I would not block on driving it through React.
- **`safeStorage()` vs `storageHandle()` when quota is full** (major): speculative. It assumes a quota-full store still permits reads and deletes; worth a look but not evidenced here.
- **Cross-tab coordination for unreadable entries** (major): real in principle, well beyond this PR, and would need a design decision about `versionchange`-style coordination. Not here.

Fix the non-array path and I will merge — everything else on this one is settled, and the gate is otherwise clean (16 check-runs, 5 core lanes, `MERGEABLE`, 0 unresolved).

Probe was appended to `saved-filters.test.ts` on a scratch checkout and reverted; nothing pushed.

BIMvoice:
Fixed. Verified the defect first: `readRaw` (apps/viewer/src/lib/search/saved-filters.ts:82, before this commit) had `if (!Array.isArray(parsed)) return [];` inside the `try`, bypassing `catalogUnwritable`/`preserveUnreadableEntry` entirely — only the `catch` (an actual `JSON.parse` throw) routed through quarantine. A stored value like `{"presets":[{"name":"Important"}]}` is valid JSON, so it never threw; it was silently discarded with no `:unreadable` backup, and the very next `saveFilter`/`deleteSavedFilter` overwrote it for good — matches your repro exactly.

**Fix** (saved-filters.ts): the non-array branch now calls `preserveUnreadableEntry(ls, STORAGE_KEY, ...)` and sets `catalogUnwritable` the same way the `catch` does, so it's quarantined to `<key>:unreadable` and writes are refused until it's repaired — same guarantee as the parse-error path.

**RED→GREEN**: added two tests to `saved-filters.test.ts` — a valid-JSON non-array catalog (a) survives the read (found under a `:unreadable*` key) and (b) is not clobbered by the save that follows. Both failed before the fix (`AssertionError`, non-array catalog destroyed), pass after.

**Mutation testing** (python3 exact-substring replace, `assert n==1`, restored by file copy, never `git checkout --`):
- Reverting the fix back to bare `return []` → both new tests fail (2 failed / 15 passed).
- `persisted` mutated to always `true` → 1 test fails (the "refuses to write... `persisted:false`" case). Still caught.
- `persisted` mutated to always `false` → 1 test fails (the "saves a preset... `persisted:true`" case). Still caught.
All three mutants restored via `cp` from a saved copy, not `git checkout`.

**Also taken, both verified before documenting**:
- `SavedFilterMutation`'s docblock was actually backwards about what a failed write returns. Traced `saveFilter`/`deleteSavedFilter`: on `persisted:false` both still call `loadSavedFilters()` — a fresh re-read of actual storage, which was never touched because the write failed — so `presets` is the *prior on-disk catalog*, not the attempted mutation. Docblock now says that explicitly. Checked the one caller (`SearchModal.filter.builder.tsx`): it already only branches on `.persisted` for the toast, doesn't rely on `.presets` reflecting the attempted change, so this is a doc-accuracy fix only, no behavior change.
- `persistence.unreadable.test.ts`'s last test deliberately latches `unwritableKeys` for both `PRESETS_KEY` and `REVIEWS_KEY` (backup write also fails) and never un-latches it. Since `apps/viewer`'s `npm test` runs every file in one `tsx --test` process (`find ... | sort` into a single invocation), and `savePresets`/`saveReviews` check that flag *before* touching storage, a later file importing `persistence.js` whose first call is a save (not preceded by its own clean read) would be spuriously refused. No file currently sits after it alphabetically that does this, so it wasn't failing today, but it's exactly the kind of order-dependent latent breakage a new test file would trip. Added an `after()` hook that does a clean read against fresh storage to clear the flags, mirroring the existing `beforeEach`.

**Not done** (per your note): no component/React toast tests, no `safeStorage()`/quota speculation, no cross-tab coordination.

**Test counts** (`apps/viewer`, `npm test`): baseline 2587 passed / 2 skipped / 0 failed (2587 total... `tests 2587`); after: `tests 2589`, 2587 passed, 2 skipped, 0 failed. `pnpm typecheck` clean (34/34).

No changeset — confirmed `@ifc-lite/viewer/package.json` has `"private": true`.

Pushed to `fix/clash-read-failure-destroys-data` (LTplus-AG). Not merging — over to you.

BIMvoice:
Independently re-verified `7b0436b6`, and correcting one number in my own agent's report before you rely on it.

## The non-array path

Reverting the quarantine (dropping the `preserveUnreadableEntry` call so the branch is a bare `return []` again) fails 2 tests; restored, 17/17. The fix routes it through the same `catalogUnwritable` handling as the `catch`, so the two unreadable shapes now behave identically — a `:unreadable` backup exists and the next ordinary save is refused rather than serializing `[]` over the entry.

Your framing is the part I want to keep on record: **the value just failed a shape check instead of a parse.** The docblock's guarantee was written about parse failures, and the guard was implemented for parse failures, so a second route to the same data loss sat one line above it without contradicting anything anyone had written down.

## `persisted` coverage is intact

Re-ran both of your mutations after the change, since a fix in the same function is exactly where that coverage could have been quietly drained:

| mutation | result |
|---|---|
| revert the non-array quarantine | 2 fail |
| `persisted` always `true` | 1 fail |
| `persisted` always `false` | 1 fail |

Still all three, still disjoint.

## Correction to the report

My agent stated "baseline 2587 pass → final 2587 pass", which would have meant the two new tests were not running. That was a mislabelled baseline, not a real problem — checked out against the previous commit's test file: **2587 tests / 2585 pass before, 2589 / 2587 after**. The +2 are genuinely there and genuinely passing. Flagging it because an unchanged pass count is precisely the shape that should make you suspicious, and it would have been reasonable to read it that way.

## The two minor items

Both taken. The docblock claim was **verified before being written** rather than assumed: `saveFilter`/`deleteSavedFilter` re-read through `loadSavedFilters()` on the failure path, so `persisted: false` really does return the prior on-disk catalog rather than the attempted mutation — the old wording implied the opposite.

Test isolation: added an `after()` resetting `persistence.ts`'s module-level `unwritableKeys`. Same shape as the `globalThis.fetch` leak on #2070 — one `tsx --test` process, so the last test in that file was leaving both keys latched unwritable for everything ordered after it.

Left the other four alone per your triage: no React-driven toast tests, no `safeStorage()` quota speculation, no cross-tab coordination.

`apps/viewer`: 2589 tests, 2587 pass, 0 fail, 2 pre-existing skips. `pnpm typecheck` clean.

louistrue:
Verified `7b0436b6` — both items closed. Merging.

The non-array quarantine works, same probe as before:

```
PROBE CORRUPT   kept=true backup=true
PROBE NON-ARRAY kept=true backup=true     (was: kept=true backup=FALSE)
```

`{"presets":[{"name":"Important"}]}` now survives with an `:unreadable` copy instead of being overwritten on the next save. Your test asserts both halves — the original bytes still present *and* a backup key created — which is what makes it more than a "returns empty" check. 18/18 green.

(The mutation I tried to pair with that did not apply — my regex missed the new block shape — so the evidence here is the probe and your test, not a mutation kill. Saying so rather than implying more than I ran.)

The `after()` hook is the right fix for the cross-file latch, and the comment explains the mechanism precisely: `apps/viewer` runs every test file in one `tsx --test` process, so `persistence.ts`'s module-level `unwritableKeys` would carry a deliberate latch into whatever file sorts next, and `savePresets`/`saveReviews` check that flag *before* touching storage — so a later file's first save gets refused for storage it never corrupted. That is a genuinely non-obvious failure and worth the paragraph.

Also still holding from earlier: `persisted` dies in both directions (always-`true` and always-`false` each fail a different named test).

Going forward I am tiering review by blast radius — `apps/viewer` changes like this one merge on green CI plus resolved threads rather than the full mutation treatment, with that rigour reserved for `rust/**` and the published packages. This one cleared the higher bar anyway.

PR #2091, opened 2026-08-04
fix(viewer): don't commit a lens edit that failed to persist

BIMvoice:
From #2085's "left for a maintainer decision" list — the sharpest of them, and the one where the remedy has an in-repo precedent.

`saveLenses` returned `void` behind a bare `catch {}`, and **all six callers committed to the store unconditionally**. A lens create, rename, delete, duplicate or import looked applied and was gone on reload, with no signal anywhere.

## Scope — correcting the report that raised it

#2085 implied lenses carry saved view state. They do not: `packages/lens/src/types.ts:216-224` is scalar rule criteria, a hex colour and three optional strings — a few hundred bytes. **Quota from lenses alone is not the realistic trigger.**

The reachable ones are (a) localStorage blocked outright — cookies-and-site-data off, some private modes — where `setItem` throws on the *first* lens, and (b) the shared origin quota already consumed by scripts, clash reviews, flavors and annotations. Not routine data loss, and worth saying so rather than letting the fix imply it.

## Decision: (b), don't commit what didn't persist

Not invented — it is the **existing contract for the closest analogue**. `clashSlice.ts:299-300` and `:330-331` already read `if (result.ok) set({ clashPresets: next })` for preset create/update/import. Lens CRUD is clash-preset CRUD with a different noun, so it gets the same rule.

The usual objection to (b) — *it costs the user their edit* — does not apply here. `handleSaveLens` closes the editor on success; on failure it now returns early **without closing it**, so the rules stay on screen to retry or export. For delete, (b) is unambiguously right: the lens simply stays listed, so nothing reappears from nowhere on reload.

(a) was rejected for this UI specifically: the editor closes, the lens appears in the list looking identical to a saved one, and a toast the user may not read is the only difference — precisely "believes it is persisted when it is not". (c) was rejected because `LensPanel` has no unsaved affordance and inventing a dirty-state badge exceeds the defect.

## Two behaviour changes, on record rather than buried

- With localStorage fully blocked, lens creation now **visibly fails** where it previously appeared to work for the session. That is the intended trade, but it is real for that population.
- `setSavedLenses` (flavor switch) also refuses to commit, and `host.ts:471` cannot toast — it `console.warn`s. The previous lenses stay in place, which is the honest state, but the signal is console-only.

## On not adding a fourth `SaveResult`

`clash/persistence.ts` and `scripts/persistence.ts` had already grown this type independently. Rather than write a third copy in the lens slice, it goes in `lib/storage/save-result.ts`. It discriminates `quota` from `unavailable` by `DOMException.name` (the `scripts/persistence.ts:103-107` precedent) so a private-mode user is not told to delete data that will not help.

**This PR only guarantees the count does not go to four.** Getting it to one is a follow-up that points clash and scripts at the shared type and folds in `'unreadable'` and `'too_many'` — see the sequencing note below.

## Evidence

| mutation | result |
|---|---|
| `saveJson` always reports success | **6 fail** — every failure-mode test |
| ungate `createLens`'s `set` | 1 fail |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`): all six fail, six negatives survive. Restored 12/12.

**The RED was deliberately strengthened.** My first version failed everywhere with `Cannot read properties of undefined (reading 'ok')` — which proves the channel is absent but *not* that the store commits. Assertions were reordered to put the **state** check first, so the tests now fail on the silent commit itself.

**Anti-vacuity:** the stub fails `setItem` **only** — `getItem` always works — so the slice genuinely loads its saved set and the failure under test is the write. That is the exact trap that bit another PR today, where a stub rejecting all storage made the module see no storage at all. The delete test also seeds its lens with writes *working*, then breaks them, so it deletes a lens that really exists.

Six negative cases, all passing: a successful create commits and the id is in the stub's backing map; rename commits; **an intentional delete still deletes** and reaches localStorage; deleting a builtin is still a no-op reported `ok`; duplicate commits and returns the copy; duplicate of a missing id returns `{ ok: true, lens: null }`.

Also toasts the import parse failure, whose comment claimed to "surface" it while only calling `console.error` — same defect class, one line, called out rather than smuggled in.

2592 passing (was **2580** — the ~2571 figure in circulation has drifted), typecheck 34/34.

## Sequencing with #2089

No file conflict: #2089 touches `clash/persistence.ts`, `saved-filters.ts` and `lib/storage/unreadable-entry.ts`; this touches none of them. Both add different files to `lib/storage/`, which merges cleanly either way.

Suggested order: **#2089 first, then this, then the consolidation follow-up.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clear notifications when saving, deleting, duplicating, or importing lenses fails.
  * Prevented unsaved changes from appearing in the viewer when browser storage fails.
  * Kept the lens editor open after unsuccessful saves.
  * Reported malformed imports and storage validation errors instead of failing silently.
  * Added warnings when saved lenses cannot be restored.


louistrue:
Both findings are the same class this PR is about — a failure reported as a success — so I would take both.

**`FileReader` errors are unhandled.** `reader.onload` runs only on a successful read; `reader.onerror` is where a failed one lands. With no error handler registered, a read failure produces neither an import nor a toast: the user picks a file and nothing happens, with no indication why. Registering the handler before starting the read is the whole fix.

**`JSON.stringify` returns `undefined` for a top-level `undefined`, function or symbol**, and `saveJson()` stores that and reports success. So a caller can "save" a value that cannot round-trip, get `ok: true`, and find nothing there next session — which is the exact shape #2101 fixed one layer up, where the store committed a value storage had refused.

Checking `payload === undefined` and returning the existing `serialize` failure keeps it consistent with how the same file already reports the other unserializable cases.

Neither is hypothetical in the way that usually justifies deferring: the first is any I/O error on a file the user chose, and the second is reachable from any caller passing a value with a symbol or function in it.

BIMvoice:
Addressed both, in d1b96ab1.

**1. FileReader errors unhandled** — confirmed, `LensPanel.handleImport` wired `reader.onload` but never `reader.onerror` (was `LensPanel.tsx:1387-1405`). A failed read produced neither an import nor a toast — the callback chain never ran at all.

Pulled the read+parse+import into `lens-import.ts` (`readFileAsText` + `importLensFile`), which registers both handlers before `readAsText` starts and returns a settled outcome (`{ ok, message? }`) on every path — I/O failure, malformed JSON, or a store-level import refusal. `handleImport` now just awaits it and toasts on `!result.ok`.

RED test asserts the actual user-visible consequence, not registration: it intercepts `FileReader` with a subclass that fails asynchronously, then checks the returned promise *settles* (rejects) within a bounded window rather than staying pending forever. (First attempt at this test raced a timeout into `assert.rejects`, which passed vacuously even against the buggy hang — fixed to observe settlement state explicitly instead.)

**2. `saveJson` doesn't check for `JSON.stringify` returning `undefined`** — confirmed, `save-result.ts:38-44`. The existing `try/catch` only guards the throwing case; a top-level `undefined`, function, or symbol makes `JSON.stringify` return `undefined` without throwing, so `payload` becomes `undefined` and `localStorage.setItem(key, undefined)` stores the literal string `"undefined"` while reporting `ok: true`.

Added `if (payload === undefined) return { ok: false, reason: 'serialize', message: ... }`, reusing the existing `'serialize'` reason rather than adding a new one, as requested.

Not claiming full coverage of every `JSON.stringify`-returns-`undefined` case: `-Infinity`/`NaN`/functions *nested* inside an object still stringify fine (to `null` or by being dropped), which is correct behavior, not a gap. The gap that existed was specifically top-level `undefined`/function/symbol, and that's what's covered.

**Testing**: RED-first for both, then mutation-tested each fix separately (python3 exact-substring removal of the `onerror` registration and of the `payload === undefined` guard, one substitution each) — the two tests fail on disjoint mutations (only `lens-import.test.ts` fails when `onerror` is stripped; only `save-result.test.ts` fails when the `undefined` guard is stripped), confirming they're independently pinned. Restored both by file copy.

2598 passing (was 2592), typecheck clean. `@ifc-lite/viewer` is private, so no changeset.

louistrue:
Verified `d1b96ab1`. Both findings fixed and both pinned — merging.

| mutation | result |
|---|---|
| drop `reader.onerror` | 2 fail |
| drop the `payload === undefined` guard | 3 fail |

Extracting `lens-import.ts` rather than testing through the panel was the right move, and the test's framing note is the part worth keeping:

> The assertion is the user-visible consequence, not "an onerror handler exists": without the fix, `readFileAsText`'s promise never settles at all, so these tests race it against a timeout and require it to REJECT within that window.

An existence check would have passed against a handler wired after `readAsText`, or one that resolved instead of rejecting. Racing the timeout tests the thing that actually broke — the promise never settling. Restoring the original `FileReader` in a `finally` is good hygiene too.

The `saveJson` guard is correct on the specification: `JSON.stringify` returns `undefined` rather than throwing for a top-level `undefined`, function, or symbol, and `setItem` would then store the literal string `"undefined"` while reporting `ok: true`. Three tests cover all three inputs.

## CodeRabbit — two findings, neither blocking

Hosted was `Review rate limited` again, so this is the CLI. Both findings are in files this PR touches, but neither is this PR's defect:

**`setSavedLenses` does not restore `builtin: true`.** Real and pre-existing. `getSavedLensesSnapshot` strips the flag (it maps to `{id, name, rules, autoColor}`), and `setSavedLenses` spreads the incoming list as-is, so a round-tripped built-in arrives flagless. `saveLenses` then discriminates on `!l.builtin` and files it under `custom`. Note `loadSavedLenses` *does* restore it by `BUILTIN_IDS`, so it self-heals on reload — the window is in-memory, until the next load, where a built-in would show as custom. Worth fixing at the `setSavedLenses` boundary the same way `loadSavedLenses` already does, but it is untouched by your diff and not the failure-reported-as-success shape this PR is about. Happy to file it.

**`host.ts` should surface the lens-restore failure in the UI.** You already handle it:

```ts
if (!saved.ok) console.warn('[ext-host] lens restore on switch not applied:', saved.message);
```

with a comment explaining the previous lens set is still in place. Promoting that to a toast would mean giving the extension host a UI dependency it deliberately avoids — the late `await import('@/store')` right above exists precisely to keep it headless-testable. A console warning is a defensible boundary for a service-layer restore, and changing it is a design call, not a defect fix.

All checks green, no red statuses, zero unresolved threads.

louistrue:
Correction, and it matters: **I said "All checks green, no red statuses" — that was wrong. CI never ran on this commit at all.**

`d1b96ab1` carries exactly two check-runs:

```
neutral  Vercel Agent Review
success  Vercel Preview Comments
```

A healthy PR here has 15-17. Every real lane — `Build + WASM + Rust + Node`, `Typecheck`, `Node tests`, `Lint`, `Viewer E2E smoke`, `parity` — is **absent**, not failing.

The cause is that the branch is `CONFLICTING` with main. GitHub does not run `pull_request` workflows for a branch it cannot merge, so the lanes never start. My gate check asked "are any checks failing or still running?", got "no" to both, and read that as green — but absence is not success, and a query that only looks for red cannot see a lane that was never scheduled.

So my verification of this PR stands on what I ran locally (both mutations, 6/6 baseline) and on the CodeRabbit CLI, not on CI. That is thinner than I implied.

The merge is blocked by the conflict anyway. Please `git merge origin/main` — several of your PRs landed since this branched, including #2096 and #2083 — and push; the lanes will run on the merged head and I will re-verify against real CI before merging.

I checked the four I have already merged today (#2085, #2072, #2083, #2096) for the same blind spot: all four had 15-17 check-runs with 5 core lanes each, so none of them went in without CI. This one was caught by the conflict rather than by my check, which is luck, not process — I have tightened the check to require the core lanes to be present rather than merely not-red.

BIMvoice:
Merged `upstream/main` in `25bf9010`. `MERGEABLE`, and the lanes are scheduling — check-runs went from 2 to 8 within a minute of the push, so CI should give you something real to verify against.

Thank you for the correction on your own gate. That is a better catch than the thing it was gating: *"a query that only looks for red cannot see a lane that was never scheduled"* generalises well past this PR, and requiring the core lanes to be **present** rather than merely not-red is the right shape. Worth noting it would have bitten silently — the PR page looks calm when nothing ran.

## The conflict, and why I resolved it toward this branch

One conflict, in `saveLenses`. `main` still has the `void`-returning form that writes `localStorage` directly and warns:

```ts
// Surfacing that to the user needs a failure channel this void-returning
// helper does not have (see lib/clash/persistence.ts for the SaveResult
// shape) — a maintainer call. Until then, at least name it.
console.warn('[lens] could not persist lenses; ...', err);
```

That comment describes exactly what this PR delivers, so `main`'s side is the state being superseded, not a newer fix at risk of being dropped. Took this branch's version: build `toStore`, return the `serialize` failure when the override check's `JSON.stringify` throws, and `return saveJson(...)` at the end.

Worth stating explicitly because the #2098 lesson cuts the other way — there, each side carried a *different* fix on the same lines and the resolution had to be both. Here only one side carries a fix. I checked which case this was before resolving rather than assuming.

## Re-verified after the merge

Both mutations re-run on the merged head, not on the pre-merge commit:

| mutation | result |
|---|---|
| drop `reader.onerror` | 2 fail (lens-import only) |
| neutralise the `payload === undefined` guard | 3 fail (save-result only) |

Still disjoint, so the merge did not quietly couple them. Full `apps/viewer` suite 2653 pass / 0 fail / 2 pre-existing skips, `tsc --noEmit` clean.

## On the two CodeRabbit findings

Agreed with your triage on both. The `setSavedLenses` / `builtin: true` round-trip is real and pre-existing — please do file it; it is a genuine in-memory window (self-healing on reload via `BUILTIN_IDS`, which is why it has gone unnoticed), and it is not the failure-reported-as-success shape this PR is about.

On `host.ts`: agreed, and the late `await import('@/store')` is the evidence — the module is deliberately headless-testable, so giving it a toast dependency would trade a real property for a cosmetic one.

louistrue:
Re-verified on the rebased head `25bf9010`. CI is real this time — 16 check-runs with all 5 core lanes green, `MERGEABLE`, no failures, no red statuses. Merging.

Both mutations re-run against the rebased code, identical results to before the rebase:

| mutation | result |
|---|---|
| drop `reader.onerror` | 2 fail |
| drop the `payload === undefined` guard | 3 fail |

Baseline 6/6 at both shas, so the rebase brought in main without disturbing either fix.

One note on scope: the CodeRabbit CLI run I reported was against `d1b96ab1`, before the rebase. The PR's own files are byte-identical between the two heads (the mutations produce the same baselines and the same kills), so that verdict still applies to this code — but I did not re-run it, and its two findings stand as I dispositioned them: the `setSavedLenses` builtin-flag loss is real, pre-existing and self-healing on reload, and the `host.ts` toast is a design call about giving the extension host a UI dependency it deliberately avoids. Happy to file the first as an issue.

PR #2095, opened 2026-08-04
fix(sandbox): copy the logs into ScriptError instead of aliasing them

### LINKED ISSUE #2092: Sandbox: ScriptError.logs is cleared by the next eval() (stored by reference)
## What happens

`ScriptError` stores the sandbox's `logs` array **by reference**, and `Sandbox.eval()` clears that same array in place at the start of every run. A caught error therefore loses its logs retroactively the moment another eval begins.

Reproduced against `packages/sandbox`:

```ts
let caught;
try {
  await sandbox.eval('console.log("from first eval"); throw new Error("boom");');
} catch (e) { caught = e; }
// caught.logs -> [["from first eval"]]

await sandbox.eval('1;');
// caught.logs -> []
```

```
BEFORE [["from first eval"]]
AFTER  []
```

Relevant lines: `ScriptError`'s `public readonly logs: LogEntry[]` constructor parameter, and `this.logs.length = 0` at the top of `eval()`.

## Why it matters

The logs are the diagnostic payload — the reason `ScriptError` carries them at all. An embedder that catches an error, stores it for a retry or a report, and runs another script finds the context gone. Worse, it disappears *after* the error was inspected, so a value that looked correct when logged is empty by the time it is serialized or displayed.

`readonly` on the field is misleading here: the binding cannot be reassigned, but the array it points at is mutated elsewhere.

## Fix

Copy at construction:

```ts
constructor(message: string, logs: LogEntry[], durationMs: number) {
  this.logs = [...logs];
}
```

A test should assert the logs survive a subsequent `eval()` — that is the property nothing currently enforces.

## Not a regression

Pre-existing. `this.logs.length = 0` is on `main`, and `ScriptError` already stored by reference there. Raised by CodeRabbit on #2078, which widens exposure (ScriptError construction sites go from 3 to 5, adding the rejected-promise path) without introducing the aliasing. Deliberately not fixed there to keep a focused fix focused.

BIMvoice:
Closes #2092, which you filed off #2078 — taking it since you noted it is a one-line fix and not that PR's to make.

## The fix removes an inconsistency rather than adding a convention

Worth leading with, because it changes how the fix should be read: the **success path already copied**. `sandbox.ts:212` returns `logs: [...this.logs]`, while `ScriptError` aliased. Success copied, failure aliased. Both now hand out a snapshot.

A shallow copy is sufficient — `LogEntry` objects are pushed fresh per `console.*` call (`bridge.ts:82,95`) and never mutated afterwards; `eval()` only detaches them via `length = 0`. A deep clone would buy nothing.

## Construction-site audit: 3, not 5

`grep -rn "new ScriptError"` across the repo finds **three**, all in `sandbox.ts` — `:181` (dumped QuickJS exception), `:196` (`'interrupted'`), `:203` (realm invalid during `vm.dump`). All pass `this.logs`.

The 5 figure came from #2078, which is **not merged into `upstream/main`** — `git log upstream/main --grep=2078` is empty and there is no `rejected-promise.test.ts` in the tree. Since the fix is in the **constructor** rather than at the call sites, it covers #2078's two new sites automatically when they land.

## Scope

This affects an embedder that **retains** a caught `ScriptError` across a later `eval()` on the same sandbox — a report, a retry, a deferred serialization. One that reads `err.logs` and discards the error before the next eval was never affected, and nothing in the repo's own callers does the former. Pre-existing, not a regression, exactly as you established.

## Evidence

Restoring the alias fails exactly the two aliasing tests (`REPLACEMENTS=1`, region re-read — I re-ran this myself on the pushed tree): 2 failed / 47 passed, restored 49/49.

**The three negatives pass before *and* after**, which is what makes them meaningful rather than decoration:
- `captures the logs written before the throw` asserts `[['from first eval']]` immediately, so a version returning `[]` could not pass it;
- `returns the logs a successful eval wrote` keeps the success path honest;
- `does not leak one eval's logs into the next result` confirms the live buffer is still cleared per run, so the fix did not turn `logs` into an accumulator.

The two-caught-errors case doubles as the error-path negative: pre-fix it fails with the **later** run's logs rather than an empty array, which distinguishes "aliased" from "never populated".

49 passing (was 44 on this base — the 49 in the issue thread is post-#2078), typecheck 34/34, `check:test-wiring` passes.

## One process note

Restoring the mutation with the obvious anchor `this.logs = logs;` reported `REPLACEMENTS=2` and the `assert n==1` aborted the write. The second occurrence is `sandbox.ts:99` — the **intentional** bridge-buffer alias in `init()`. A naive replace-all would have silently rewritten it, and the resulting failure would have looked like a problem with the fix rather than with the tooling.

That is the third distinct way a mutation has misapplied today, and the first two were caught by the same count-and-assert step rather than by the suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed script errors so their captured logs remain available after subsequent script executions.
  * Prevented log entries from leaking between separate evaluations.
  * Preserved independent logs for multiple caught errors.

* **Tests**
  * Added coverage for error logging, successful evaluations, and sequential executions.

### DIFF
--- a/.changeset/sandbox-script-error-logs-copy.md
+++ b/.changeset/sandbox-script-error-logs-copy.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Stop a caught `ScriptError`'s `logs` from being emptied by the next `eval()` (#2092).
+
+`ScriptError` stored the constructor's `logs` argument by reference, and every `ScriptError` is constructed with the sandbox's single log buffer — the same array `Sandbox.eval()` clears in place (`this.logs.length = 0`) at the start of each run. An embedder that caught an error, kept it for a retry or a report, and then ran another script found the error's logs empty, after having already inspected them. `ScriptError` now copies the array at construction, so a caught error keeps the console output of the run that failed for as long as the error is held.
+
+Only the error path was affected: the success path already returned `logs: [...this.logs]`, and the two are now consistent. Sandboxes that never retain an error across evals see no change.
--- a/packages/sandbox/src/error-logs.test.ts
+++ b/packages/sandbox/src/error-logs.test.ts
@@ -1,0 +1,118 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression tests for a caught `ScriptError` losing its logs (#2092).
+ *
+ * `eval()` clears the sandbox's log buffer in place at the start of every run,
+ * so a `ScriptError` that stored that same array by reference had its logs
+ * emptied retroactively by the next `eval()` — after the embedder had already
+ * inspected them.
+ *
+ * The scripts here fail with a plain `throw`, never a resource limit, so no
+ * test in this file can abort the shared WASM module.
+ */
+
+import { describe, expect, it } from 'vitest';
+import type { BimContext } from '@ifc-lite/sdk';
+import { createSandbox, ScriptError, type Sandbox } from './sandbox.js';
+
+/** The scripts under test never touch `bim`, so an empty context suffices. */
+const EMPTY_SDK = {} as unknown as BimContext;
+
+/** Runs `script` and returns the `ScriptError` it must reject with. */
+async function evalScriptError(sandbox: Sandbox, script: string): Promise<ScriptError> {
+  try {
+    const result = await sandbox.eval(script, { typescript: false });
+    throw new Error(`expected a ScriptError, but eval resolved with ${JSON.stringify(result.value)}`);
+  } catch (err) {
+    expect(err).toBeInstanceOf(ScriptError);
+    return err as ScriptError;
+  }
+}
+
+/** Flattens a log entry's arguments so assertions read like the console call. */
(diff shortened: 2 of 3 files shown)


louistrue:
Verified `ad9d4d2b`. This closes #2092 and the second test is one I would not have asked for.

Reverting `this.logs = [...logs]` to the alias fails two named tests:

```
× keeps a caught error's logs after a later eval clears the buffer
× keeps two caught errors' logs independent of each other
```

The first is the reported bug. The second is the case I missed when filing: two errors caught from the same sandbox previously shared one array, so they were not merely wiped by the next eval — they were never distinct in the first place. An embedder collecting failures across several scripts would have found every entry showing the last script's logs. Testing that independently is what makes the copy semantics actually pinned rather than incidentally satisfied.

49/49 green, no unresolved threads, no failing checks.

Worth noting how this one arrived: it came out of a CLI review on #2078 rather than from reading this code, and #2078 only widened the exposure (ScriptError sites 3 -> 5) without introducing the aliasing. The bug had been on `main` since the class was written. Fixing it in its own PR rather than folding it into #2078 was the right call — #2078 stayed a focused rejection-reporting change and this got the two tests it deserved.

One adjacent thing I am filing separately rather than asking you to fold in: `bridge.ts`'s 4MB `totalBytes` latch is per-sandbox, not per-eval, so on a reused sandbox one log-heavy script permanently truncates every later script's output. Different mechanism from #2084, same file. Not this PR's to fix.

PR #2096, opened 2026-08-04
fix(sandbox): don't retain a log argument the byte budget cannot size

### LINKED ISSUE #2087: Log entries that fail JSON.stringify escape the byte budget while consuming entry slots
Confirmed by @louistrue on #2080, where he established it is **pre-existing on `main`** and asked for it separately rather than growing a house-rule sweep into a budget-accounting fix.

## The accounting hole

`packages/sandbox/src/bridge.ts:88-95` on `main`:

```ts
let entryBytes = 0;
try { entryBytes = JSON.stringify(nativeArgs)?.length ?? 0; }
catch { entryBytes = 0; /* unserializable args — skip cost accounting */ }
totalBytes += entryBytes;
```

When `JSON.stringify(nativeArgs)` throws, `entryBytes` stays `0` **and the entry is appended anyway**. So repeated unserializable logs accumulate against `MAX_LOG_ENTRIES` (1000) while contributing nothing to the 4 MB `MAX_TOTAL_BYTES` budget.

A script can therefore retain up to 1000 arbitrarily large objects that the byte cap never sees. The entry cap bounds the *count*, not the memory.

## Reachability

The throw is script-triggerable: a `BigInt` argument survives `vm.dump` but not `JSON.stringify`. (A cyclic object does **not** reach it — `vm.dump` returns the string `"[object Object]"` — so the comment naming cyclic args as the case was wrong, corrected in #2080.)

## Not introduced by #2080

That PR's change to this block is `catch {` → `catch (err) {` plus a one-time warn. It makes the condition **visible** without altering it. Filing here so the accounting fix is reviewed on its own terms.

## Note for whoever takes it

The one-notice-per-context design in #2080 is load-bearing and not obvious: a per-occurrence log would hand `for(;;) console.log(1n)` a host-console flood. Naive compliance with the no-silent-catch rule would have created a second resource-exhaustion path while fixing the visibility of the first. Any fix here should keep that property.

Possible directions, none decided:

1. Charge a **fallback cost** for an unserializable entry (a fixed estimate, or a cheap structural size probe) so it counts against `MAX_TOTAL_BYTES` rather than being free.
2. **Refuse to retain** args that cannot be sized — store a placeholder rather than the object, so the retained memory matches what the budget can see.
3. Keep the entry but drop the reference to the unsizeable payload once the cap is hit.

(2) is the only one where the accounting is exact by construction rather than by estimate.

BIMvoice:
Closes #2087, which I filed after you confirmed it on #2080.

## I overstated the exposure in that issue — correcting it here

I wrote "up to 1000 arbitrarily large objects". Probing `vm.dump` directly shows it already flattens **any** value its in-VM serializer rejects to the string `"[object Object]"` — cyclic objects, functions, and crucially **an object with a nested BigInt** all survive `JSON.stringify` afterwards.

**The only value that reaches the host unsizeable is a top-level BigInt.** So it is not arbitrarily large *objects*, it is arbitrarily large *BigInts*. QuickJS allocates `1n << 1000000n` (~125 KB) and refuses ~`3000000n`, so the honest ceiling is roughly 1000 × 125 KB ≈ **125 MB** a script author can retain that the 4 MB cap never sees. `vm.dump` of a huge BigInt is also quadratic — ~2 s for a 1M-bit one — so the practical rate is low. Real, but not fast.

## Option (2), and the narrowing is what justifies it

Because the failure is always a **top-level argument**, per-argument handling is exact by construction and costs nothing in debuggability. An estimate (option 1) would have to over-estimate a BigInt *without rendering it* — and rendering is the expensive part, so any honest estimate is either wrong or as costly as the payload. Option 3 leaves a window where the payload is held.

**The fast path is untouched:** one `JSON.stringify` for the whole entry, so serializable logs are charged byte-for-byte what they were before. Only on throw does it size each argument alone, keep the ones that serialize **untouched**, and replace the rest:

```
console.log('count', 42n, {ok:true})  ->  count 42n { ok: true }
console.log(1n << 100000n)            ->  [BigInt too large to retain]
```

Level, timestamp and neighbouring arguments all intact; every placeholder is ≤ ~70 chars and charged its exact length. **A sizing failure never drops a log line.**

## Evidence

Retaining the arg fails **3** tests — I re-ran that myself on the pushed tree (`REPLACEMENTS=1`, region re-read): 3 failed / 46 passed, restored 49/49.

The two byte-accounting negatives **pass before and after**, which is what shows the normal path is unchanged. The sharper one pins the truncation boundary exactly: six 1 MB logs, each entry serializing to 1 048 580 B, four of which just pass `MAX_TOTAL_BYTES` — so a per-entry sizing error of more than ~4 bytes moves that boundary and fails the test. That is the no-double-counting case.

The RED asserts **retention** first, not the warning — so it demonstrates the accounting hole rather than a missing notice — with the latched warning asserted in the same test as the oracle that the catch actually fired.

**One mutation survives, and I believe it is unobservable rather than untested:** not charging the placeholder leaves at most 1000 × 70 B ≈ 70 KB uncharged, and you would need ~60 000 entries to reach 4 MB against a 1000-entry cap. No public-API test can distinguish it. Charging it is still correct, so it stays — flagging it rather than claiming a clean sweep.

## #2080 is not merged, so the latch is here

`upstream/main` has the bare `catch { entryBytes = 0; }` and no `bridge-console.test.ts`. Since my catch must not be silent and the latch is the stated design constraint, it is implemented here with **#2080's exact variable name and rationale comment** so the rebase is trivial. My test file is named `bridge-console-budget.test.ts`, so either PR can merge first without a file-add conflict.

## One house-rule judgement call for you

`retainSizeableArgs` has a `catch` that neither logs nor rethrows — it returns the placeholder and increments a counter, and the **caller** warns once per context with the error and the replaced count. Reporting inside the map is exactly the console flood the latch exists to prevent. If you read `AGENTS.md:17` more literally, the alternative is threading the caught error out of the map and warning with it — identical behaviour, more plumbing. Your call.

49 passing (was **44** — the 49 figure assumed #2080 merged), typecheck 34/34.

### DIFF
--- a/.changeset/sandbox-log-budget-accounting.md
+++ b/.changeset/sandbox-log-budget-accounting.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Stop captured log entries that cannot be sized from escaping the sandbox's host-memory budget.
+
+The bridge caps captured console output twice: by entry count (1000) and by cumulative serialized size (4 MB), because `vm.dump` copies sandbox values onto the host heap, which the QuickJS memory limit does not bound. The size charge came from `JSON.stringify`, and when that threw the entry was charged zero bytes and retained anyway. A top-level `BigInt` is the value that reaches the host in that state — it survives `vm.dump` intact but has no JSON form, and QuickJS will allocate one of a million bits — so a script could park up to 1000 such values, tens of megabytes, that the byte budget never saw. (An object the VM cannot serialize never got that far: `vm.dump` already flattens it to the string `"[object Object]"`.)
+
+The bridge now refuses to retain what it cannot size. When sizing an entry fails, each argument is sized on its own: arguments that serialize are kept untouched, and only those that do not are replaced by bounded text, charged at exactly the length of that text. Retained memory is therefore always what the budget can see. Serializable logs are sized and capped exactly as before.
+
+**Embedder-visible:** `LogEntry.args` no longer contains `BigInt` values. A small BigInt is retained as its literal text (`42n`), a very large one as `[BigInt too large to retain]`; other arguments on the same line are unaffected, so the log still shows which script logged what. The failure is reported to the host console once per sandbox context — not once per entry, since the trigger is script-supplied.
--- a/packages/sandbox/src/bridge-console-budget.test.ts
+++ b/packages/sandbox/src/bridge-console-budget.test.ts
@@ -133,6 +133,43 @@ describe('captured-log byte budget', () => {
     }
   });
 
+  it('charges the fallback path something proportionate, not a wild overestimate', async () => {
+    // The magnitude of the FALLBACK charge was unpinned (maintainer negative
+    // control on #2096): replacing `bytes += json?.length ?? 0` inside
+    // retainSizeableArgs with a 10-million-fold constant left every other test
+    // in this file green. Nothing noticed, because the exact-charge test above
+    // drives the happy path — `retainSizeableArgs` only runs once
+    // `JSON.stringify` of the whole argument array has already thrown.
+    //
+    // Over-charging is the mirror image of the bug #2087 fixed: instead of
+    // retaining entries the budget cannot see, it would discard entries the
+    // budget should have had room for. Approximate is fine here (the comment
+    // in bridge.ts says "approximate host cost", and String.length undercounts
+    // UTF-8 bytes for non-ASCII anyway) — orders of magnitude wrong is not.
+    //
+    // Observable proxy for `totalBytes`, which is private: one entry carrying
+    // an unsizeable argument must not exhaust a 4MB budget by itself. Under
+    // the 10-million mutation the very next entry truncates.
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      await withSandbox(async (sandbox) => {
+        const result = await sandbox.eval(
+          'console.log(1n, "small");\nfor (let i = 0; i < 3; i++) console.log("after", i);\n9;',
+        );
+        expect(result.value).toBe(9);
+        // 4 entries, and none of them the truncation marker: the first entry's
+        // fallback charge left room for the rest.
+        expect(result.logs).toHaveLength(4);
+        expect(result.logs.map((entry) => entry.args[0])).not.toContain(TRUNCATION_MARKER);
+        // The sizing failure still happened — otherwise this test would pass
+        // by never reaching the fallback path at all.
+        expect(warn.mock.calls.filter((call) => String(call[0]).includes(SIZING_WARNING))).toHaveLength(1);
+      });
+    } finally {
+      warn.mockRestore();
+    }
+  });
+
   it('still caps the entry count', async () => {
     await withSandbox(async (sandbox) => {
       const result = await sandbox.eval(
--- a/packages/sandbox/src/bridge-console-budget.test.ts
+++ b/packages/sandbox/src/bridge-console-budget.test.ts
@@ -1,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The bridge caps captured console output twice: by entry count
+ * (MAX_LOG_ENTRIES = 1000) and by cumulative serialized size
+ * (MAX_TOTAL_BYTES = 4MB). Both caps exist because `vm.dump` copies sandbox
+ * values onto the host heap, which the QuickJS memory limit does not bound.
+ *
+ * The byte cap can only charge for what it can measure. A top-level BigInt is
+ * the one sandbox value that survives `vm.dump` intact but has no JSON form
+ * (an object the VM cannot serialize arrives already flattened to the string
+ * "[object Object]"), so `JSON.stringify` throws — and a BigInt is unbounded
+ * in size. Retaining such an argument would consume an entry slot while
+ * contributing nothing to the byte budget (#2087), so the bridge replaces it
+ * with bounded text and charges exactly that.
+ *
+ * Two further properties are pinned here, and they pull against each other:
+ *   - the sizing failure is *reported* (house rule: no silent catch), and
+ *   - it is reported at most once per context — the trigger is script-supplied,
+ *     so a per-entry log would let `for(;;) console.log(1n)` flood the host.
+ */
+
+import { describe, expect, it, vi } from 'vitest';
+import type { BimContext } from '@ifc-lite/sdk';
+import { createSandbox } from './sandbox.js';
+
+const SIZING_WARNING = 'could not be sized';
+const TRUNCATION_MARKER = '[log output truncated: limit reached]';
+
+/** Mirrors bridge.ts — the caps are private, so the tests drive their edges. */
+const MAX_LOG_ENTRIES = 1000;
+
+async function withSandbox<T>(run: (sandbox: Awaited<ReturnType<typeof createSandbox>>) => Promise<T>): Promise<T> {
+  const sandbox = await createSandbox({} as unknown as BimContext);
+  try {
+    return await run(sandbox);
+  } finally {
+    sandbox.dispose();
+  }
+}
+
+describe('captured-log byte budget', () => {
+  it('does not retain a log argument it could not size', async () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      await withSandbox(async (sandbox) => {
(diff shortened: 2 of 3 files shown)


louistrue:
Verified `64c860fd`. The fix is sound and pinned. **I would not take greptile's remaining finding** — here is the arithmetic.

## The framing-bytes finding is immaterial

It is technically correct: the fallback sums per-argument lengths, the happy path stringifies the whole array, so the fallback misses `[`, `]` and the commas — exactly `N + 1` characters for `N` arguments.

```
$ node -e "const a=['x','y','z'];
  console.log(JSON.stringify(a).length, a.reduce((s,v)=>s+JSON.stringify(v).length,0))"
13 9        # delta 4 for 3 args
```

Worst case across the whole budget: `MAX_LOG_ENTRIES` is 1000, so even at ten arguments per entry that is ~11,000 characters against a 4 MB ceiling — **0.27%**, and only on the fallback path, which needs an unserializable argument to be reached at all.

Now compare that to the error already baked into the metric it would be correcting. `String.length` counts UTF-16 code units, not bytes:

```
$ node -e "const s=JSON.stringify(['日本語のログ出力']);
  console.log(s.length, Buffer.byteLength(s,'utf8'))"
12 28
```

**133% under** for CJK text — three orders of magnitude larger than the framing delta, on the *happy* path, for any non-ASCII log line. Your own comment says `Approximate host cost of retaining this entry`, and that is the right call: this is a budget, not an accounting ledger. Adding framing arithmetic would make the number look more exact without making it more accurate, which is worse than leaving it visibly approximate.

If byte accuracy ever does matter, the fix is `Buffer.byteLength`/`TextEncoder`, not brackets and commas. Not needed here.

## Verification

Reverting the fix (retaining `nativeArgs` unsized, charging 0) fails three named tests:

```
× does not retain a log argument it could not size
× keeps a small BigInt readable and leaves its neighbours untouched
× warns once, not per entry, when entries cannot be sized
```

So the #2087 hazard is genuinely pinned.

## One gap, your call

A negative control **passes** when it should not. Replacing `bytes += json?.length ?? 0` with `bytes += 10_000_000` — a 10-million-fold overcharge that would truncate logs after a single entry — leaves all 5 tests green:

```
### M2 negative control: charge a huge constant instead of real size
      Tests  5 passed (5)
```

The suite pins *that* an unsizeable argument is not retained for free, but nothing about the magnitude of the charge. Over-charging is the opposite failure from the one you fixed and would silently swallow legitimate log output.

This is consistent with dismissing the framing point, not in tension with it: approximate is fine, six-orders-of-magnitude wrong is not. One assertion — log a known payload, check `totalBytes` lands in a sane range — closes it and would also settle greptile's thread on evidence rather than argument.

Not a blocker. The shipped behaviour is correct either way, so say the word and I will merge as is, or take the one-line assertion first if you would rather have it.

louistrue:
Follow-up: I ran the CodeRabbit CLI on this too (hosted was `Review rate limited` — a pass that never ran). Two findings, both pointing at the same thing, and it is a **real weakness but not this PR's**.

## The budget can be blown by a single log call

CodeRabbit's point is that the budget is checked *before* an entry is added, never against it, so one oversized entry is retained in full. That is true, and it is worse than it sounds. One `console.log`, no loop:

```
PROBE entries retained : 1
PROBE chars retained   : 41943040
PROBE budget (bytes)   : 4194304
PROBE overshoot factor : 10.0x
```

I picked a 40 MB payload; 400 MB gives 100x. The overshoot is bounded only by what the script chooses to pass, which for a sandbox containing untrusted code is the part that matters.

**This predates your PR.** `origin/main` already has the check-then-add shape:

```js
if (logs.length >= MAX_LOG_ENTRIES || totalBytes >= MAX_TOTAL_BYTES) { truncated = true; ... return; }
```

Your own test comment describes the same property honestly — four entries reach 4194320 against a 4194304 budget, so the limit is already exceeded by 16 bytes by design before the fifth call truncates. This PR is about unsizeable arguments being retained for free (#2087); it neither introduces nor widens the overshoot.

So I would **not** take either CodeRabbit finding here. The second one asks you to change `charges serializable entries exactly as before and truncates on the same entry` so the fourth entry truncates — that test exists precisely to prove this PR did *not* move the boundary, which is what makes the change safe to review. Rewriting it to assert new behaviour would delete the evidence and fold an unrelated fix into a scoped PR.

It is worth its own issue: make the check `totalBytes + entryBytes > MAX_TOTAL_BYTES` and truncate before pushing, so the ceiling is a ceiling. That is a deliberate behaviour change (entries that currently land would start being dropped), which is another reason it does not belong here. Happy to open it — say the word.

This also subsumes the magnitude gap I mentioned above: the suite cannot detect a 10-million-fold overcharge, and the budget cannot detect a 10x overshoot. Same underlying looseness, same one-assertion fix, same follow-up.

Nothing pushed — the probe went into `bridge-console-budget.test.ts` on a scratch checkout and has been reverted.

BIMvoice:
Took the assertion — `ba42fd32`. And agreed on greptile's finding; your arithmetic settles it.

## Your negative control reproduced, and it explains itself

I ran it before writing anything, because if it did not reproduce I would have refuted it rather than patched a hole that was not there. It does: `bytes += json?.length ?? 0` → `bytes += 10_000_000`, **5 passed**.

What makes it slip through is worth recording, because it is the same shape as the #2070 and #2105 gaps — **the test that looks like it covers this is on the other path**. "charges serializable entries exactly as before and truncates on the same entry" pins per-entry magnitude tightly (four entries at 1048580, truncation on the fifth), so it reads as magnitude coverage. But it drives the **happy** path, and `retainSizeableArgs` only runs once `JSON.stringify` of the whole argument array has already thrown. The mutated line is unreachable from that test.

## The assertion

`charges the fallback path something proportionate, not a wild overestimate`. `totalBytes` is private, so it uses truncation as the observable: one entry carrying an unsizeable argument must not exhaust a 4MB budget by itself. Log `1n` plus three ordinary entries, assert all four arrive and none is the truncation marker.

It also asserts the sizing warning fired exactly once — without that it could pass by never reaching the fallback path at all, which would make it the same kind of test it exists to replace.

## Evidence, both directions

| mutation | result |
|---|---|
| `bytes += 10_000_000` (your negative control, over-charge) | **1 fail** — the new test, and only it |
| `entryBytes = 0` + retain `nativeArgs` (the original #2087 under-charge) | **3 fail** — the three pre-existing tests |

Counts printed and asserted `== 1`, regions re-read, restored by file copy. Sandbox suite 50/50, `tsc --noEmit` clean.

The pairing is the point: the budget is now pinned against being wrong in *either* direction, and the two mutations fail disjoint sets of tests, so neither is passing for the other's reason.

## On the framing bytes

Agreed, and your comparison is the argument I would have wanted to make: 0.27% worst case on a path that needs an unserializable argument to reach, against 133% under-count for CJK on the happy path from `String.length` counting UTF-16 code units. Adding bracket-and-comma arithmetic would make the number look more exact without being more accurate — and `Approximate host cost` is already the honest label. If byte accuracy ever matters it is `TextEncoder`, not framing.

I would rather you merge this than leave it open on the thread, but it is your call whether the new assertion changes that.

louistrue:
Verified `ba42fd32`. The new test is exactly right — merging.

You did more than close the gap. My negative control only proved the magnitude was unpinned; your test also proves it *reaches* the fallback:

```ts
expect(warn.mock.calls.filter((call) => String(call[0]).includes(SIZING_WARNING))).toHaveLength(1);
```

Without that, the test would pass by never triggering a sizing failure at all — green for the wrong reason, which is the failure mode that made the original control necessary. The comment says so directly: "otherwise this test would pass by never reaching the fallback path at all."

Re-ran all three mutations at this sha:

| mutation | before | now |
|---|---|---|
| revert the fix (retain unsized, charge 0) | 3 fail | 3 fail |
| **`bytes += 10_000_000`** (magnitude) | **passed** | **1 fail** |
| make `JSON.stringify` never throw (vacuity) | — | 5 fail |

The vacuity check is the one I added to confirm your guard works: if the fallback becomes unreachable, five tests fail rather than quietly going green.

Using entry count as an observable proxy for the private `totalBytes` is the right call, and the comment explains why over-charging is the mirror image of #2087 rather than a separate concern.

The two remaining bot findings are dispositioned above and I have resolved the thread: greptile's framing bytes are 0.27% worst case inside a metric already 133% under for non-ASCII, and CodeRabbit's budget-overshoot point is real but predates this PR — `origin/main` already checks the budget before adding rather than against the entry. Its second finding would have rewritten `charges serializable entries exactly as before and truncates on the same entry`, which exists to prove this PR did not move that boundary.

CodeRabbit CLI at this sha: 0 findings. Hosted was `Review rate limited` again — a pass that never ran — so that verdict comes from the CLI, not the badge.

I will open the budget-overshoot issue separately (one `console.log` retains 41.9 MB against the 4 MB ceiling — 10x, unbounded by payload size). It needs a deliberate behaviour change, so it does not belong here.

PR #2098, opened 2026-08-04
fix(viewer): a cache read failure must degrade to a miss, not break the load

BIMvoice:
Second pass over `apps/viewer/src`. One real defect, and it is a language-level trap rather than a missing log.

## Update: rebased onto #2102 (merged as `55f78f05`)

`#2102` landed on main and added `withConnection`/`beginTransaction`, which retries once on `InvalidStateError` by reopening the connection. That makes "the memoised connection was closed" a **recoverable** condition, not a failure — so the original two tests (which induced the read failure by calling `db.close()`) went vacuous: `getCached` now returns the entry via the reopen, `hasCached` now returns `true`. Rebasing surfaced this immediately (both failed with `#2102`'s own `[IFC Cache] Connection was closed underneath the cache; reopening` log firing in between).

**The fix itself is still needed and is unchanged.** Main had (and this PR still closes) the same five bare `return new Promise(...)` inside a `try` — the rejection-escape on the `request.onerror` path, which `#2102` deliberately left alone (it only retries the synchronous `InvalidStateError` from `transaction()`; by the time a request exists, the transaction was already created, so there's nothing left to retry).

Both conflicting fixes are combined at all five sites: `#2102`'s `const tx = await beginTransaction(mode)` plus this PR's `return await new Promise(...)`. Verified: zero bare `return new Promise(` remain inside a `try` in `ifc-cache.ts` (`evictToFree`'s own bare `return new Promise` is fine — it isn't wrapped in a `try` itself, and its caller does `await` it), and `withConnection`/`beginTransaction`/the reopen path are intact.

**Tests re-targeted.** The two tests in `ifc-cache.read-failure.test.ts` no longer close the connection. They now force a `request.onerror` directly — by one-shot monkeypatching the relevant `IDBObjectStore.prototype` method to abort the transaction the instant the request is issued (a pending request whose transaction aborts fires `onerror` with `AbortError`, per spec, never `onsuccess`). This is the path `#2102`'s reopen cannot touch, since no `InvalidStateError` is ever thrown. Confirmed each test fails for the *right* reason, not merely "didn't throw": mutating any one of the five `return await new Promise(` back to a bare `return new Promise(` (verified file-copy restore afterward) makes the corresponding assertion fail — all five sites are pinned by these two tests, not just the two originally exercised.

`apps/viewer`: 2640 tests, 2638 passing (2 pre-existing skips), typecheck 34/34.

---

## `return new Promise` inside a `try` makes the catch dead code

Five async functions in `services/ifc-cache.ts` did `return new Promise(...)` **inside** a `try`. A bare `return promise` in an async function hands the promise straight to the caller, so the enclosing `catch` never runs on that promise's own rejection. Every one of those catch blocks was dead while its comment documented the opposite contract — `return null`, `return false`, warn-and-continue, zeros.

**The sibling write path is the evidence this was an oversight, not a design.** `setCached` carries:

```
// A cache write must NEVER break the load (AGENTS.md; task blocker #2)
```

and correctly uses `await new Promise`. The read path was left behind.

## Reachability — verified, not assumed

- `openDatabase()` memoises the connection in a module-level `dbPromise` and clears it **only on open error** — never when the connection is later closed (another tab's `versionchange`, or the browser reclaiming under storage pressure). *(Superseded by `#2102`, which now reopens on a closed connection — the read-failure route that matters post-rebase is `request.onerror`, not a closed connection; see the Update section above.)*
- `db.transaction()` on a closed connection throws `InvalidStateError` **synchronously inside the executor**, so the promise rejects and the rejection escapes.
- `await getCached(cacheKey)` at `useIfcLoader.ts:933` (`loadStage = 'cache-lookup'`) then throws out of `loadFile` into the outer catch at `:1857`. **The user sees "failed to load" for a file that would have parsed fine**, plus a `posthog.captureException`. `deleteCached` at `:952` is on the same path.

I also probed the transaction-abort variant: `tx.abort()` mid-request rejects with `AbortError` rather than hanging, so the same fix covers it — no speculative `tx.onabort` handlers added.

**Scope honesty:** `hasCached` and `getCacheStats` have **no callers** in the viewer, so those two are hardening. `getCached`, `deleteCached` and `clearCache` are live.

## Also: `chatSlice.ts:219`, one of the two you left open

Now warns **once per session** with the error bound — `persistMessages` runs on every message add and exec-result update, so an unlatched log would flood.

**This closes the silence, not the loss.** On quota exhaustion 50 messages with attachments still fail to persist. The real fix is a degradation strategy — strip attachments, or persist fewer messages — which is a product decision about what to drop, so it stays with you.

## Evidence (pre-rebase; the `ifc-cache` half is superseded — see the Update section at the top for the current mutation results)

Removing any one `await` fails that function's own test; **all five killed**. I re-ran the `getCached` one myself on the pushed tree (`TOTAL_AWAITS=5`, `AWAITS_LEFT=4` asserted, region re-read): its test fails, `hasCached` survives, restored 23/23. *(This ran against the pre-rebase tests, which closed the connection to induce failure — no longer valid post-`#2102`; the current tests induce failure via `request.onerror` instead, see the Update section.)*

**Anti-vacuity work, since two of today's traps were in the test's own stub:** the cache test asserted the entry is readable **before** closing the connection, so a later `null` meant "degraded to a miss" rather than "was never there" — the rewritten tests keep this same precondition-readability guard, just against `request.onerror` instead of `close()`. The chatSlice stub is proven **selective** — a non-messages write succeeds while a messages write throws — because a stub rejecting every `setItem` makes the module see no storage at all. (The chatSlice half of this PR is untouched by the rebase.)

Two mutation-harness misfires were caught rather than papered over: an anchor matched `store.delete(key)` inside `evictToFree`, and a `git checkout --` in the same loop silently reverted the fix. Detected via a printed `restored awaits: 0`, then redone with asserted counts and file-copy restore.

## Triage

480 `catch` clauses under `apps/viewer/src`; **81** empty-bodied, **205** strict (no log, no throw) — the strict reading reproduces the first pass's 206. 177 in scope after the skip list.

| category | n | verdict |
|---|---|---|
| storage preference read/write | 54 | benign, documented |
| `JSON.parse` of stored/foreign data | 24 | benign, defined default |
| pointer-capture / dispose / close | 14 | benign, idempotent |
| `fetch` fallbacks | 8 | benign, defined empty |
| render-loop / per-entity | 12 | benign, `setStatus`-surfaced |
| **IndexedDB** | **5** | **the defect above** |
| already surfaces via `setStatus`/`{error}`/`reject` | 60 | benign |

`hooks/ingest/federationAlign.ts:412` and `setCached` are worth citing as the house pattern — the first captures an error and *does* re-check it on both exits.

## Left for you

- **`clashSlice.ts:308/314/320`** — still open. `deleteClashPreset`, `setClashPresetEnabled` and `resetClashPresets` discard a `{ok:false}` `SaveResult` and `set(...)` unconditionally, so the UI reports success while storage refused. Their siblings return the result and only `set` when `ok`. Closing it means widening three `void` signatures, and `lib/clash/persistence.ts` is already being touched on `fix/clash-read-failure-destroys-data`.
- ~~`ifc-cache.ts`'s `dbPromise` is never cleared on a closed connection~~ — **done, by `#2102`** (merged as `55f78f05`): it now reopens once on `InvalidStateError` instead of staying dead for the tab's lifetime. See the Update section at the top.

2583 passing (was **2580** — the ~2595 figure in circulation is from a different base), typecheck 34/34. Strict count 205 → 202.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache reliability when browser storage operations fail. Cache lookups, statistics, deletion, and clearing now complete gracefully instead of hanging or producing unhandled errors.
  - Chat history persistence failures no longer interrupt in-memory chat functionality.
  - Reduced repeated warning messages by displaying a persistence warning only once per session.

### DIFF
--- a/apps/viewer/src/services/ifc-cache.read-failure.test.ts
+++ b/apps/viewer/src/services/ifc-cache.read-failure.test.ts
@@ -1,0 +1,135 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * A cache READ failure must degrade to a MISS, never break the load.
+ *
+ * `setCached` already guarantees this on the WRITE path ("A cache write must
+ * NEVER break the load"). The read/maintenance helpers each wrap their work in
+ * a `try`/`catch` that documents the same contract — warn and return
+ * `null`/`false`/zeros — but they used to `return new Promise(...)` INSIDE the
+ * try without `await`, so the promise's own rejection bypassed the catch
+ * entirely and propagated to the caller. That bug is fixed with `return await`.
+ *
+ * Failure induction: #2102 (`withConnection` / `beginTransaction`) added a
+ * ONE-SHOT reopen for a connection that was closed underneath the cache
+ * (`InvalidStateError` from `db.transaction()`), so closing the memoised
+ * connection is no longer a failure at all — it's a recoverable condition
+ * `beginTransaction` transparently papers over. These tests instead force a
+ * `request.onerror` — the failure mode #2102 deliberately leaves alone,
+ * because by the time a request exists the transaction was already created
+ * successfully, so there is nothing left to retry. We do this by aborting the
+ * transaction the moment the request is issued: IndexedDB fires `onerror`
+ * (with `AbortError`), not `onsuccess`, on every request still pending when
+ * its transaction aborts — a spec-guaranteed way to fail a request without
+ * needing invalid data or a broken store.
+ *
+ * Lives in its own file because the assertions monkeypatch
+ * `IDBObjectStore.prototype`; `tsx --test` runs each file in its own process.
+ */
+
+// fake-indexeddb installs a Node-compatible IDB implementation on
+// `globalThis.indexedDB` (+ the IDB* constructors) via the `/auto` entry.
+import 'fake-indexeddb/auto';
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import {
+  setCached,
+  getCached,
+  hasCached,
+  deleteCached,
+  clearCache,
+  getCacheStats,
+} from './ifc-cache.js';
+
+/** Resolve to the settled outcome, or the literal 'HUNG' after `ms`. */
+async function settleOrHang<T>(p: Promise<T>, ms = 500): Promise<
+  { kind: 'resolved'; value: T } | { kind: 'rejected'; error: unknown } | { kind: 'hung' }
+> {
+  return Promise.race([
+    p.then(
+      (value) => ({ kind: 'resolved' as const, value }),
+      (error: unknown) => ({ kind: 'rejected' as const, error }),
+    ),
+    new Promise<{ kind: 'hung' }>((resolve) => {
+      setTimeout(() => resolve({ kind: 'hung' as const }), ms).unref?.();
+    }),
+  ]);
+}
+
+/**
+ * One-shot: patch `IDBObjectStore.prototype[methodName]` so the NEXT call
+ * issues a real request, then immediately aborts that request's transaction
+ * before returning control to the caller. Per spec, a pending request whose
+ * transaction aborts fires `onerror` (with an `AbortError`), never
+ * `onsuccess` — this fails the request without any invalid data, and without
+ * touching `openDatabase`'s memo, so `beginTransaction`'s reopen path is never
+ * even triggered (the transaction it handed back was created just fine).
+ */
+function failNextRequest(methodName: 'get' | 'count' | 'delete' | 'clear' | 'getAll'): void {
+  const proto = IDBObjectStore.prototype as unknown as Record<string, (...args: unknown[]) => IDBRequest>;
+  const original = proto[methodName];
+  proto[methodName] = function (this: IDBObjectStore, ...args: unknown[]) {
+    proto[methodName] = original; // restore immediately: strictly one-shot
+    const request = original.apply(this, args);
+    try {
+      request.transaction!.abort();
+    } catch {
+      /* transaction may already be inactive; the request still errors */
+    }
+    return request;
+  };
+}
+
+describe('ifc-cache: a read failure degrades to a miss, it does not break the load', () => {
+  it('getCached resolves null (not rejects) when the underlying request errors', async () => {
+    await setCached('req-error', new ArrayBuffer(64), 'a.ifc', 64);
+
+    // Guard against a vacuous pass: the entry must really be readable first,
+    // so a later `null` means "degraded to a miss", not "was never there".
+    assert.notEqual(await getCached('req-error'), null, 'precondition: entry is readable');
+
+    failNextRequest('get');
+    const outcome = await settleOrHang(getCached('req-error'));
+    assert.equal(outcome.kind, 'resolved', 'getCached must not reject on a read failure');
+    assert.equal(outcome.kind === 'resolved' ? outcome.value : undefined, null);
+
+    // The entry is untouched — this was a transient failure, not a corruption
+    // or a real miss. Confirms we degraded rather than actually losing data.
+    assert.notEqual(await getCached('req-error'), null, 'entry survives the induced failure');
+  });
+
+  it('hasCached / getCacheStats / deleteCached / clearCache also degrade instead of rejecting', async () => {
+    await setCached('req-error-2', new ArrayBuffer(64), 'b.ifc', 64);
+    assert.equal(await hasCached('req-error-2'), true, 'precondition: entry is present');
+
+    failNextRequest('count');
+    const has = await settleOrHang(hasCached('req-error-2'));
+    assert.equal(has.kind, 'resolved', 'hasCached must not reject');
+    assert.equal(has.kind === 'resolved' ? has.value : undefined, false, 'degrades to "not found", not a throw');
+
+    failNextRequest('getAll');
+    const stats = await settleOrHang(getCacheStats());
+    assert.equal(stats.kind, 'resolved', 'getCacheStats must not reject');
+    assert.deepEqual(
+      stats.kind === 'resolved' ? stats.value : undefined,
+      { entryCount: 0, totalSize: 0, entries: [] },
+      'degrades to an empty report, not a throw',
+    );
+
+    failNextRequest('delete');
+    const del = await settleOrHang(deleteCached('req-error-2'));
+    assert.equal(del.kind, 'resolved', 'deleteCached must not reject');
+    // The delete request errored, so the entry must still be there — deleteCached
+    // degraded (silently skipped), it did not silently "succeed" on a lie.
+    assert.equal(await hasCached('req-error-2'), true, 'entry survives the failed delete');
+
+    failNextRequest('clear');
+    const cleared = await settleOrHang(clearCache());
+    assert.equal(cleared.kind, 'resolved', 'clearCache must not reject');
+    assert.equal(await hasCached('req-error-2'), true, 'entry survives the failed clear');
+  });
+});
--- a/apps/viewer/src/services/ifc-cache.ts
+++ b/apps/viewer/src/services/ifc-cache.ts
@@ -339,7 +339,10 @@ export interface CacheEntryMeta {
 export async function getCached(key: string): Promise<CacheResult | null> {
   try {
     const tx = await beginTransaction('readonly');
-    return new Promise((resolve, reject) => {
+    // `return await` is load-bearing: a bare `return promise` inside a try
+    // hands the promise straight to the caller, so its rejection would bypass
+    // the catch below and break the load instead of degrading to a miss.
+    return await new Promise((resolve, reject) => {
       const store = tx.objectStore(STORE_NAME);
       const request = store.get(key);
 
@@ -465,7 +468,7 @@ export async function setCached(
 export async function hasCached(key: string): Promise<boolean> {
   try {
     const tx = await beginTransaction('readonly');
-    return new Promise((resolve, reject) => {
+    return await new Promise((resolve, reject) => {
       const store = tx.objectStore(STORE_NAME);
       const request = store.count(IDBKeyRange.only(key));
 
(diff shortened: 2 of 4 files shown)


louistrue:
Worth taking, and the reasoning generalises beyond style preference.

The second test depends on the memoised database the first test established. That holds under a full-file run today, and stops holding the moment anyone runs it with `-t`, reorders the file, or splits it — at which point it does not fail, it stops exercising the closed-connection path and passes for a different reason.

That is the failure mode this PR is about, in the test rather than the code: a result that looks correct while measuring something else. A regression test for a cache that silently stops working is the last place worth leaving an implicit dependency.

Arranging the memoised state inside the test costs a few lines and makes it survivable under targeted execution.

For the record on the PR itself: the change is right, and keeping it alongside #2102 is correct — #2102 stops the connection staying dead, this stops a dead connection breaking the load. The second is the guard for cases the first does not anticipate, which is why both earn their place rather than one superseding the other.

louistrue:
**Hold this one — #2102 landed on main and removed this PR's premise.** Verified by actually doing the merge, not by reading.

The test-isolation point is addressed: the second test now opens and closes its own connection instead of inheriting one. That thread is fine.

## Both tests fail once this is rebased

`#2102` (merged as `55f78f05`) added `withConnection()`, which retries once on `InvalidStateError` by reopening. This PR's tests induce a read failure by closing the memoised connection — which is now a **recoverable** condition, not a failure.

I merged `origin/main` into `60a31f5b`, resolved the conflict correctly (see below), and ran the suite:

```
not ok 1 - getCached resolves null (not rejects) when the cached connection is closed
        + actual - expected
        + { buffer: Blob { [Symbol(kLength)]: 64 }, ... }
        - null

not ok 2 - hasCached / getCacheStats / deleteCached / clearCache also degrade instead of rejecting
        true !== false
```

with your own #2102 log line firing in between:

```
[IFC Cache] Connection was closed underneath the cache; reopening InvalidStateError
    at withConnection (apps/viewer/src/services/ifc-cache.ts:284:12)
    at async getCached (apps/viewer/src/services/ifc-cache.ts:341:16)
```

`getCached` returns the entry because the reopen succeeded. That is #2102 working as designed. It means the closed-connection route can no longer reach the degrade-to-a-miss path, so these tests need a different failure induction — a failing `request.onerror`, which is the path #2102 deliberately left to this PR.

**The fix itself is still needed.** Main has six bare `return new Promise(` and zero awaited ones, so the rejection escape you are closing is still open on the request.onerror path. Only the tests are invalidated, not the change.

## The rebase will try to delete your fix

The conflict is five sites, all the same shape:

```
<<<<<<< HEAD  (this PR)
    const db = await openDatabase();
    // `return await` is load-bearing: a bare `return promise` inside a try
    // hands the promise straight to the caller, so its rejection would bypass
    // the catch below and break the load instead of degrading to a miss.
    return await new Promise((resolve, reject) => {
      const tx = db.transaction(STORE_NAME, 'readonly');
=======       (main, from #2102)
    const tx = await beginTransaction('readonly');
    return new Promise((resolve, reject) => {
>>>>>>> origin/main
```

Each side carries a different fix on the same lines. Taking main wholesale drops your `return await`; taking HEAD wholesale drops #2102's reopen. The resolution has to be **both** — main's `beginTransaction` with `return await` restored, at all five:

```ts
const tx = await beginTransaction('readonly');
return await new Promise((resolve, reject) => {
```

That is what I resolved to in order to run the test above, so the combination does compile and behave. A "take theirs" resolution here would silently reinstate the exact bug this PR exists to fix, and no test would catch it once the current two are rewritten.

Worth keeping the load-bearing comment on at least one site after the merge — it is the only thing that explains why the `await` is not redundant, and it is precisely the kind of line a future cleanup removes.

BIMvoice:
Rebased onto `upstream/main` (onto `55f78f05` / #2102) and re-targeted the two tests, per the hold.

**Conflict resolution.** All five sites in `ifc-cache.ts` had the same shape conflict — `#2102`'s `const tx = await beginTransaction(mode)` on one side, this PR's `return await new Promise(...)` on the other. Resolved to both at every site: `const tx = await beginTransaction(mode); return await new Promise(...)`. Verified after: `grep -c "return new Promise("` in `ifc-cache.ts` → 1 hit total, and it's `evictToFree`'s own (not inside a `try`, and its caller `await`s it) — zero bare `return new Promise(` remain inside a `try`. `withConnection`/`beginTransaction`/`liveConnection`/`invalidateConnection`/`memoisedDb` are all present and unmodified. Kept the load-bearing `return await` comment on `getCached` (the only site that still carries the pre-existing comment, since the other four never had one).

**Tests re-targeted.** Confirmed your read exactly: rebased-in, both tests failed with your `[IFC Cache] Connection was closed underneath the cache; reopening` line firing — `getCached` returned the entry (reopen succeeded), `hasCached` returned `true`. Closing the connection is no longer a failure post-`#2102`.

Rewrote both to induce failure via `request.onerror` instead: a one-shot monkeypatch of the relevant `IDBObjectStore.prototype` method aborts the transaction the instant the request is issued. Per spec, a pending request whose transaction aborts fires `onerror` (`AbortError`), never `onsuccess` — no invalid data needed, and `withConnection`'s reopen never triggers (it only retries the synchronous `InvalidStateError` from `transaction()`; by the time a request exists, the transaction already succeeded). Each assertion checks the actual degrade-to-a-miss outcome (null / false / empty stats / entry-survives-the-failed-delete-or-clear), not just "didn't throw".

**Mutation-tested all five sites**, not just two: for each, flipped `return await new Promise(` back to a bare `return new Promise(` via an exact-substring python3 replacement (count asserted `==1`, printed to stderr), confirmed the corresponding assertion in `ifc-cache.read-failure.test.ts` goes RED, then restored by `cp` from a saved original (never `git checkout --`) and confirmed byte-identical + GREEN again. All five (`getCached`, `hasCached`, `deleteCached`, `clearCache`, `getCacheStats`) are pinned by the two tests.

**Full suite:** `apps/viewer` 2640 tests, 2638 passing, 0 failing, 2 pre-existing skips (unchanged from baseline). `pnpm typecheck`: 34/34.

Updated the PR body's stale claims (the old "closed connection" reachability path, and the pre-rebase evidence section) to point at this comment/the new Update section rather than removing the history.

Force-pushed the rebased branch. Not merging.

BIMvoice:
Independently re-verified the rebase, since the hazard you named is specifically one that survives a green suite.

**Both fixes are present after the resolution**, counted rather than eyeballed:

```
return await new Promise(   5
await beginTransaction(     5
bare return new Promise(    1
```

The single bare one is `evictToFree` (`ifc-cache.ts:107-108`) — a plain function whose body *is* the return, with no `try` to escape, and its caller awaits it. So zero bare returns remain inside a `try`, and `withConnection` / `invalidateConnection` / `beginTransaction` are intact. A "take theirs" resolution would have left `beginTransaction` at 5 and `return await` at 0, which is exactly what those counts are for.

**Mutation re-run after the rebase**, not before it — the point you made is that a merge can drain assertions without touching a test, so evidence gathered pre-rebase would not have been worth much. Reverting `getCached`'s `return await new Promise(` to a bare return (single replacement asserted, awaited count 5 → 4 confirmed) fails the read-failure test; restored by file copy, full suite 2638 pass / 0 fail / 2 pre-existing skips.

Worth stating plainly what changed about the evidence here: the old tests were not merely stale, they had become **vacuous in the dangerous direction** — passing because #2102 recovered the connection, while asserting nothing about the rejection escape. Re-targeting to `request.onerror` is what makes them measure the thing again, and the five-site mutation table is what shows they do.

louistrue:
Verified `c7fea9a0` against **current** main, not against your base. This is fully resolved — all three parts.

## 1. The tests now use a failure #2102 cannot recover from

Switching the induction from a closed connection to `request.onerror` is exactly right, and for the reason your header states: that is the path #2102 deliberately left alone. A closed connection is now recoverable, so it could no longer reach the degrade-to-a-miss branch; an errored request still does.

```
ok 1 - getCached resolves null (not rejects) when the underlying request errors
ok 2 - hasCached / getCacheStats / deleteCached / clearCache also degrade instead of rejecting
# pass 2  # fail 0
```

That is after merging current `origin/main` into your head — a clean merge, no conflict this time.

## 2. The resolution kept both fixes

Five `return await new Promise(` alongside six `beginTransaction` references. Neither side got dropped, which was the thing I was most worried about.

## 3. The bad resolution is now caught by CI

This is the part that matters beyond this PR. I applied the mutation a "take theirs" resolution would produce — strip `await` from all five sites, keep `beginTransaction`:

```
### MUTATION: strip the load-bearing 'await' from all 5 sites
    not ok 1 - getCached resolves null (not rejects) when the underlying request errors
    not ok 2 - hasCached / getCacheStats / deleteCached / clearCache also degrade instead of rejecting
# pass 0  # fail 2
```

Before this PR that mutation was invisible — main has six bare `return new Promise(` and zero awaited, with nothing asserting the difference. Now a future conflict resolved the lazy way fails loudly instead of silently reinstating the bug. That is a better outcome than the fix alone.

Worth recording why this was not visible from GitHub: your checks were green the whole time, because they describe the PR's **base**, and #2102 landed after it. Green-on-branch and green-on-merge are different claims, and only the second one is the merge criterion. Same trap as the #2081-before-#2080 ordering issue.

Once the remaining checks finish I will merge this.

louistrue:
greptile's isolation finding is **stale** — it describes the test you replaced.

The rewritten second test arranges everything itself: its own key (`req-error-2`), its own `setCached`, its own precondition assertion, and a fresh `failNextRequest(...)` before each of the four operations. Nothing is inherited from the first test, and the closed-connection dependency it was objecting to no longer exists — that was exactly what moving to `request.onerror` removed.

Run in isolation, which is the check the finding calls for:

```
$ tsx --test --test-name-pattern="also degrade instead of rejecting"
    ok 1 - hasCached / getCacheStats / deleteCached / clearCache also degrade instead of rejecting
# pass 1  # fail 0  # skipped 0
```

Resolving it.

The per-operation assertions are stronger than isolation alone, incidentally — `entry survives the failed delete` and `entry survives the failed clear` pin that the degraded path did not silently *succeed* at destroying something. A `deleteCached` that swallowed its error and reported success would pass a "does not reject" check and fail these.

This PR is otherwise done and I verified it against current main earlier: clean merge, 2/2 pass, and the take-theirs mutation (stripping `await` from all five sites) fails both tests. Gate is clean — 16 check-runs with 5 core lanes present, `MERGEABLE`, no failures, no red statuses. Merging.

PR #2100, opened 2026-08-04
fix(data,cache,create,cli): four failures that rendered as legitimate results

BIMvoice:
Second sweep, over the packages not yet covered. The headline is exactly the shape this exercise was aimed at.

## 1. A NULL list property was reported as a real empty list — in two packages

`packages/data/src/property-table.ts:331` and `packages/cache/src/sections/properties.ts:117`. `valueString` is a `Uint32Array`, so the `-1` NULL sentinel wraps to `4294967295`; `StringTable.get` answers `''` for any out-of-range index; `JSON.parse('')` throws straight into `catch { return [] }`.

A caller asking *"does this element declare a list?"* was told **"yes, and it is empty."**

**The branch immediately above already guards this**, with a comment saying *"so a genuine NULL property value stays null instead of becoming `''`"*. The `List` branch two lines below did not — and the same pair exists in two packages.

## 2. A silent length-unit fallback its own siblings already warn about

`packages/create/src/in-store/extract-walls.ts:150` pinned `lengthUnitScale` to 1.0 on failure. On a millimetre model that leaves coordinates like 31614 and collapses the panel's metre-based snap tolerance — garbage geometry, no signal.

`resolve-anchor.ts:50` and `resolve-source.ts:158` `console.warn` on the **identical** failure, with the comment *"don't hide the failure — a wrong scale emits silently mis-sized geometry"*. `resolve-anchor` even says it *"mirrors the read-side conversion in extract-walls.ts"*. The read side never got the fix.

## 3. A truncated capability list printed as the API

`packages/cli/src/commands/schema.ts:22` — on any failure importing `@ifc-lite/sandbox/schema` (unbuilt package, partial install, version skew) it silently substituted a hand-maintained **6-namespace** subset and printed it to stdout as valid JSON. The documented consumer is *"LLM tools to discover available commands"*, which then conclude the missing namespaces do not exist.

## 4. `--version` lied by 18 releases

`packages/cli/src/index.ts:58` — `catch { return '0.4.0' }` with the package at `0.22.0`. Low reachability, but it is the version-mismatch-hiding shape and it poisons bug reports.

## Verdict on the three packages checked for fabricated passes — clean

Worth reporting the negative rather than padding:

- **`clash`**: 2 catches, **both already `console.warn` with the error bound**. Neither can turn a failure into "no collisions" — the clash engine itself has **no catches at all**.
- **`ids`**: **no audit reports "pass" because it failed.** The two `return []` sites (`entity-facet.ts:268`, `audit/ifc-schema/index.ts:726`) are a *missed cross-check*, and bad patterns are reported separately via `W_REGEX_UNVERIFIED`; `:726`'s sole caller **skips** the applicability check on an empty list rather than emitting a pass. Commented, not changed.
- **`query`**: cleanup with comments; `isAvailable → false` is correct.

## The geometry leftovers

The three silent SAB-view fallbacks get a once-per-worker latch. They run **N times per model**, which is why they could not simply copy their warning sibling at `:1456` (once per load). **Log-only, and explicitly not mutation-tested** — `geometry.worker.ts` has no message-handler harness across the package's 19 test files, and I am not claiming one.

## Evidence

Removing the NULL guard fails **both** list tests — I re-ran that myself on the pushed tree (`REPLACEMENTS=1`, region re-read): 2 failed / 79 passed, restored 81/81. The latch mutations each produce 2 warnings where 1 is asserted.

**A vacuous test was caught in my own fixture**, and it is worth naming because it is the fifth of the day: `StringTable.fromArray` **prepends `''`** when slot 0 is not the canonical empty string, shifting every index — so two "expects null" cases would have passed for the wrong reason, via a fall-through rather than the guard. Building the leading `''` explicitly is why the guard-removal mutation now fails 2 tests rather than 1.

**Behaviour changes**, both RED-tested: a NULL `List` property reads `null` instead of `[]`; `--version` reports `0.0.0-unknown` when the manifest is unreadable.

data 81 passing (was 77), cache 45 (was 42), cli 264 (was 260), create 119 (was 118), geometry 189 unchanged. `turbo typecheck` green across 31 tasks.

One pre-existing failure, verified not mine: `mcp/src/tools/export-usd.test.ts` asserts `/var/…` against macOS's `/private/var/…` realpath. Restoring the pristine file from HEAD reproduces it identically.

## Triage

181 catches across the swept packages; **9** empty-bodied, **84** by the wider "mentions no reporting call" reading — a deliberate superset that includes sites surfacing via `process.stderr.write`, `fatal()` or an `errors[]` array, which turned out to be the majority. `drawing-2d`, `lens`, `encoding`, `merge`, `spatial`, `solar`, `embed-protocol` and `wasm` source have **zero** catch blocks.

Ten sites where silence is correct now carry a one-line reason so the next sweep skips them.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of null, empty, and malformed list property values, with clearer one-time warnings.
  - CLI schema generation now reports fallback failures while still producing valid output.
  - CLI version detection now identifies missing or invalid version information.
  - Wall-unit extraction failures now provide diagnostic warnings while retaining the metre fallback.
  - Geometry processing reports SharedArrayBuffer compatibility fallbacks once per worker.

- **Documentation**
  - Clarified expected error-handling behavior across several command-line, import, export, and runtime workflows.

### DIFF
--- a/.changeset/package-silent-catches-round2.md
+++ b/.changeset/package-silent-catches-round2.md
@@ -1,0 +1,27 @@
+---
+'@ifc-lite/data': patch
+'@ifc-lite/cache': patch
+'@ifc-lite/create': patch
+'@ifc-lite/cli': patch
+'@ifc-lite/geometry': patch
+---
+
+Stop four package-level failures from being reported as ordinary results.
+
+- `@ifc-lite/data` / `@ifc-lite/cache`: a List-typed property with no value
+  came back as `[]` — a real empty list — because the NULL string sentinel
+  resolved to `''` and the resulting `JSON.parse` throw was swallowed. NULL
+  now reads as `null`, matching the string branch beside it, and a genuinely
+  unparseable list value logs once (latched) before falling back to `[]`.
+- `@ifc-lite/create`: `extractWallSegmentsForStorey` silently defaulted to a
+  metre length-unit scale when unit extraction threw, mis-scaling every
+  extracted wall segment on a millimetre model. It now warns with the error,
+  matching `resolveSpatialAnchor` / `resolveDuplicateSource`.
+- `@ifc-lite/cli`: `ifc-lite schema` printed a reduced built-in schema as if
+  it were the full SDK surface when `@ifc-lite/sandbox/schema` could not be
+  loaded; it now says so on stderr. `--version` no longer reports a
+  hard-coded `0.4.0` when `package.json` is unreadable — it reports
+  `0.0.0-unknown` and explains why on stderr.
+- `@ifc-lite/geometry`: the shard and finalise paths that fall back from a
+  SharedArrayBuffer view to a materialised (file-sized) copy now say so once
+  per worker, matching the streaming-prepass path that already did.
--- a/.changeset/package-silent-catches-round2.md
+++ b/.changeset/package-silent-catches-round2.md
@@ -19,9 +19,11 @@ Stop four package-level failures from being reported as ordinary results.
   matching `resolveSpatialAnchor` / `resolveDuplicateSource`.
 - `@ifc-lite/cli`: `ifc-lite schema` printed a reduced built-in schema as if
   it were the full SDK surface when `@ifc-lite/sandbox/schema` could not be
-  loaded; it now says so on stderr. `--version` no longer reports a
-  hard-coded `0.4.0` when `package.json` is unreadable — it reports
-  `0.0.0-unknown` and explains why on stderr.
+  loaded; it now says so on stderr and exits non-zero (stdout is still pure
+  JSON, unchanged shape), so a piping caller that discards stderr still sees
+  the failure. `--version` no longer reports a hard-coded `0.4.0` when
+  `package.json` is unreadable — it reports `0.0.0-unknown` and explains why
+  on stderr.
 - `@ifc-lite/geometry`: the shard and finalise paths that fall back from a
   SharedArrayBuffer view to a materialised (file-sized) copy now say so once
   per worker, matching the streaming-prepass path that already did.
--- a/packages/cache/src/sections/properties-list-values.test.ts
+++ b/packages/cache/src/sections/properties-list-values.test.ts
@@ -1,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Cache rehydration of List-typed property values.
+ *
+ * `valueString` is a Uint32Array, so the NULL sentinel -1 wraps to
+ * 4294967295 and `StringTable.get` answers `''`. The string branch of the
+ * cache's `getPropertyValue` guards against that so a NULL stays `null`; the
+ * List branch fed `''` straight into `JSON.parse` and swallowed the throw,
+ * so a property with no list came back from cache as a real empty list.
+ * Same defect and same fix as `@ifc-lite/data`'s property table.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { StringTable, PropertyValueType, propertyTableFromColumns } from '@ifc-lite/data';
+import { BufferWriter, BufferReader } from '../utils/buffer-utils.js';
+import {
+  writeProperties,
+  readProperties,
+  __resetListParseWarningLatch,
+} from './properties.js';
+
+const NULL_STRING_SENTINEL = 0xffffffff; // -1 written into a Uint32Array
+
+/**
+ * Build a one-column-per-field table of List rows and round-trip it through
+ * the cache section. Slot 0 of the string table must be the canonical empty
+ * string (`fromArray` inserts one otherwise and shifts every index), so
+ * payload strings start at index 1.
+ */
+function roundTripListRows(payloadStrings: string[], valueStringIndices: number[]) {
+  const strings = StringTable.fromArray(['', ...payloadStrings]);
+  const count = valueStringIndices.length;
+  const table = propertyTableFromColumns(
+    {
+      count,
+      entityId: Uint32Array.from(valueStringIndices.map((_, i) => 100 + i)),
+      psetName: new Uint32Array(count).fill(1),
+      psetGlobalId: new Uint32Array(count).fill(0),
+      propName: new Uint32Array(count).fill(2),
+      propType: new Uint8Array(count).fill(PropertyValueType.List),
+      valueString: Uint32Array.from(valueStringIndices),
+      valueReal: new Float64Array(count),
+      valueInt: new Int32Array(count),
+      valueBool: new Uint8Array(count).fill(255),
+      unitId: new Int32Array(count).fill(-1),
+    },
+    strings,
+  );
+
+  const writer = new BufferWriter();
+  writeProperties(writer, table);
+  return readProperties(new BufferReader(writer.build()), strings);
+}
+
+beforeEach(() => {
+  __resetListParseWarningLatch();
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe('cached List property values', () => {
+  it('round-trips a well-formed list', () => {
+    const restored = roundTripListRows(['Pset_Test', 'Tags', '["a","b"]'], [3]);
+    expect(restored.getPropertyValue(100, 'Pset_Test', 'Tags')).toEqual(['a', 'b']);
+  });
+
+  it('reports a NULL list as null, not as an empty list', () => {
+    const restored = roundTripListRows(['Pset_Test', 'Tags'], [NULL_STRING_SENTINEL]);
+    expect(restored.getPropertyValue(100, 'Pset_Test', 'Tags')).toBeNull();
+  });
+
+  it('warns once, with the error bound, when a cached list is not JSON', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const restored = roundTripListRows(['Pset_Test', 'Tags', '{not json'], [3, 3]);
+
+    expect(restored.getPropertyValue(100, 'Pset_Test', 'Tags')).toEqual([]);
+    expect(restored.getPropertyValue(101, 'Pset_Test', 'Tags')).toEqual([]);
+
+    const calls = warn.mock.calls.filter(
+      (args) => typeof args[0] === 'string' && args[0].includes('not valid JSON'),
+    );
+    expect(calls, 'expected exactly one latched warning').toHaveLength(1);
+    expect(calls[0][1]).toBeInstanceOf(Error);
+  });
+});
--- a/packages/cache/src/sections/properties.ts
+++ b/packages/cache/src/sections/properties.ts
@@ -11,6 +11,28 @@ import { comparePropertyValues, PropertyValueType } from '@ifc-lite/data';
 import { BufferWriter, BufferReader } from '../utils/buffer-utils.js';
 
 /**
+ * List-value parse failures are per-property on a cache-rehydration path, so
+ * the warning is latched: a corrupt cache would otherwise emit one line per
+ * row.
+ */
+let listParseWarningShown = false;
+
+/** @internal test seam — reset the once-per-process list-parse warning latch. */
+export function __resetListParseWarningLatch(): void {
+  listParseWarningShown = false;
+}
+
+function warnListParseFailureOnce(listStr: string, err: unknown): void {
+  if (listParseWarningShown) return;
+  listParseWarningShown = true;
+  console.warn(
+    `[cache] cached property list value is not valid JSON; reporting an empty ` +
+      `list (further occurrences suppressed). value=${JSON.stringify(listStr.slice(0, 120))}`,
+    err,
+  );
+}
+
+/**
  * Write PropertyTable to buffer
  * Format:
  *   - count: uint32
@@ -92,13 +114,23 @@ export function readProperties(reader: BufferReader, strings: StringTable): Prop
       case PropertyValueType.Logical:
         const boolVal = valueBool[idx];
         return boolVal === 255 ? null : boolVal === 1;
-      case PropertyValueType.List:
-        const listStr = strings.get(valueString[idx]);
+      case PropertyValueType.List: {
+        // Mirrors the string branch above (and @ifc-lite/data's
+        // `getPropertyValue`): the NULL sentinel -1 wraps to 4294967295 and
+        // `strings.get` answers '' for it, so without this guard a NULL list
+        // property was rehydrated from cache as `[]` — a real empty list —
+        // with the silent catch hiding the difference.
+        const li = valueString[idx];
+        if (!(li >= 0 && li < strings.count)) return null;
+        const listStr = strings.get(li);
+        if (listStr === '') return null;
         try {
           return JSON.parse(listStr);
-        } catch {
+        } catch (err) {
+          warnListParseFailureOnce(listStr, err);
           return [];
         }
+      }
       default:
         return null;
     }
--- a/packages/cli/src/commands/layer-push.ts
+++ b/packages/cli/src/commands/layer-push.ts
@@ -55,6 +55,8 @@ function errorOf(body: string): string {
     const parsed = JSON.parse(body) as { error?: string; reason?: string };
     return parsed.error ?? parsed.reason ?? body;
   } catch {
+    // Legitimately silent: this only formats an error the caller is already
+    // reporting. A non-JSON body is surfaced verbatim, so nothing is lost.
     return body;
   }
 }
--- a/packages/cli/src/commands/schema.invalid-export.test.ts
+++ b/packages/cli/src/commands/schema.invalid-export.test.ts
@@ -1,0 +1,70 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The third state of `@ifc-lite/sandbox/schema`: present, but unusable.
+ *
+ * `schema.test.ts` covers the import *throwing* and `schema.success.test.ts`
+ * covers it resolving correctly. Between them sits a partial or version-skewed
+ * `dist/` that resolves the module with `NAMESPACE_SCHEMAS` absent or the wrong
+ * shape. The import succeeds, so the `catch` never fires — and `schemas.map()`
+ * runs outside the `try`, so the command died on an uncaught TypeError with no
+ * fallback schema, no stderr warning and no exit code (maintainer finding on
+ * #2100).
+ *
+ * That is strictly worse than the silence this command was fixed for: the
+ * fallback, the warning and the non-zero exit are all bypassed at once.
+ *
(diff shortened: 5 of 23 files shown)


louistrue:
The four fixes are the right theme and I would take three of them. But the `extract-walls` one is incomplete in a way that matters, and the reviewer's reading is correct — verified against the code.

## The new warning covers the rare path, not the common one

The change turns `catch {}` into `catch (error) { console.warn(…) }` around `extractLengthUnitScale`, with the comment "don't hide the failure". That fires only when the call **throws**.

`extractLengthUnitScale` (`packages/parser/src/unit-extractor.ts`, not touched by this PR) does not throw for the ordinary failures. It has **eight** `return 1.0` paths — no `IFCPROJECT`, no project ref, no `UnitsInContext`, unreadable or malformed unit declarations — and only **two** of them warn:

```
console.warn lines in the function: 2
return 1.0 lines:                   8
```

So six failure modes return a valid-looking metre scale, silently, and the new `catch` never sees them. A millimetre model whose unit declarations are malformed still reads as metres — coordinates like 31614, snap tolerance collapsed — which is precisely what your comment says the warning exists to prevent.

That makes the warning's presence slightly worse than its absence: a reader now sees a handled failure path and reasonably assumes the case is covered.

## Why this is the PR's own thesis

The title is "four failures that rendered as legitimate results". A function that answers `1.0` — a perfectly plausible scale — for "I could not determine the scale" is the same defect, one layer down. The fix addresses the exception channel and leaves the in-band one, which is the one that actually fires.

The honest options are to make the six silent paths warn (smallest, keeps the signature), or to return `null`/`undefined` for "unknown" and let each caller decide (correct, but changes a shared function with several callers). Either is out of scope here — but the comment should not claim coverage the code does not have.

## The schema.ts thread is the same shape

A failed dynamic schema import prints the reduced fallback through the normal JSON path and exits 0, so a machine caller consuming stdout and exit status cannot tell a reduced schema from a complete one. Same class again: a degraded result rendered as a legitimate one.

## Not disputing the rest

21 files across 11 packages is a lot of surface, and the other three fixes are the kind of thing only a sweep finds. My concern is scoped to the two threads above — both are cases where the sweep found the right place and stopped one layer short.

BIMvoice:
You are right, and I verified it — the counts are worse than your sample:

```
return 1.0 lines in extractLengthUnitScale: 11
console.warn lines:                          2
```

Nine paths return a plausible metre scale silently, and none of them reaches the `catch` this PR added.

**Your framing is why I changed the comment rather than leaving it:** a reader now sees a handled failure path and reasonably assumes the case is covered, so the warning's presence really was slightly worse than its absence. Corrected in 272cd737 — it now says exactly what it covers, what it does not, and why.

**And you are right that it is this PR's own thesis one layer down.** The title is "four failures that rendered as legitimate results", and a function answering `1.0` for *"I could not determine the scale"* is precisely that — `1.0` is also the correct answer for a metre model, so the two are indistinguishable to every caller. The sweep found the right place and stopped at the exception channel, which is the one that does not fire.

Filed as **#2104** with both remedies you named. I did not fold either in: the function has **7+ callers** (`length-unit-scale.ts`, `useIfcCache.ts`, `effective-georef.ts`, `kmz-export.ts`, `PropertiesPanel.tsx`, `ModelMetadataPanel.tsx`, `playground-dispatcher.ts`), so returning `null` for "unknown" is a shared-signature change that deserves its own review rather than riding along in a 21-file sweep. Making the nine paths warn is landable today and I am happy to take it on your say-so.

**On `schema.ts` — agreed, same shape, and I left it in the same state.** A failed dynamic import prints the reduced schema through the normal JSON path and exits 0, so a machine caller reading stdout and exit status cannot tell reduced from complete. The stderr warning I added helps a human and does nothing for the documented consumer, which is "LLM tools". The honest fixes are a non-zero exit or a marker in the payload itself, both of which change a CLI contract — noted in #2104's closing section rather than done quietly.

Taking the point about 21 files across 11 packages too. Both threads are cases where a wide sweep found the right place and stopped one layer short, which is an argument for narrower PRs when the finding turns out to have depth.

louistrue:
Verified `272cd737`. The comment now says what the code does, and your count is more accurate than mine.

I said 8 `return 1.0` paths against 2 warnings; I had counted over a line range rather than the whole function. Over the full body it is **11 against 2** — confirmed:

```
return 1.0 in extractLengthUnitScale: 11
console.warn:                          2
```

So the gap is wider than I reported, and the corrected comment carries the right figure.

The wording is the part worth keeping:

> Deliberately NOT claiming the case is covered … Those still read as metres, silently. Fixing that means either warning on the in-band paths or returning null for "unknown", and the function has 7+ callers, so it is tracked separately rather than widened here.

That is a comment doing the opposite of what has gone wrong repeatedly here. It states the boundary of what the change achieves, names both remedies, and gives the reason the larger one is out of scope — so the next reader neither assumes coverage that does not exist nor re-derives why it was left.

Scoping the first sentence to a THROWN failure closes the original problem: the previous version read as though the failure mode was handled, when the common path never reaches that `catch`.

The `schema.ts` thread is the same shape and still open — a failed dynamic import prints the reduced fallback through the normal JSON path and exits 0, so a machine caller cannot distinguish reduced from complete. Same remedy question, smaller blast radius.

louistrue:
Verified `272cd737`. Of the two open threads, **one is already correctly answered and I would close it; the other is a real gap in this PR's own thesis.**

## extract-walls — not needed, and the comment proves it

The corrected comment does everything I asked and one thing more: it quantifies the boundary (11 `return 1.0` paths against 2 warnings), states plainly which cases do *not* reach the catch, gives the reason for not widening (7+ callers), and defers rather than pretending.

I checked the deferral rather than taking it — `#2104 extractLengthUnitScale returns 1.0 for "unknown", so a mis-scaled model is indistinguishable from a metre one` is open and covers exactly that case. The claim is true, so greptile's finding is answered by scope, not by silence. Closing that thread.

This is the right shape for a "don't hide failures" PR: fix what the catch can see, say precisely what it cannot, and point at where the rest lives.

## CLI schema — the fix does not reach its own stated audience

The comment names the consumer:

> Callers of `ifc-lite schema` are usually LLM tools discovering the API — being handed a truncated namespace list that looks authoritative is worse than being told the load failed, so say so on stderr (stdout stays pure JSON).

Agreed on the diagnosis. But the signal goes to the one channel that audience routinely discards. On the reduced path:

- stdout is valid, well-formed JSON, same shape as a real schema
- **no exit code is set** — I checked, there is no `process.exit` or `exitCode` anywhere in the command, so it exits 0
- the fallback carries 33 real namespace/method names, so it looks plausible rather than obviously stubbed

A tool doing `ifc-lite schema | jq` gets a confident, truncated API description and a success status. That is the exact failure this PR is named for — "rendered as a legitimate result" — just relocated from "no message at all" to "a message on a stream nobody read."

Keeping stdout pure JSON is right, so the fix is not "print the warning to stdout." Two options that preserve it:

1. **Non-zero exit code.** The Unix-correct signal, no change to the JSON shape. It is genuinely not a success, so callers treating it as one is the bug.
2. **A marker in the payload** — e.g. wrap as `{ degraded: true, schemas: [...] }`, or add `"_degraded": true`. Machine-readable, but it changes the output contract for every consumer.

I lean to (1): it costs one line, breaks nothing that was already reading the JSON correctly, and a caller that ignores exit codes has opted out of error handling generally. But it is your call — this is a CLI contract decision, not a defect with one right answer, and either beats stderr alone.

Worth saying: three of the four fixes in this PR are clean. This one is the same class of miss it is fixing, which is why it is worth closing rather than shipping at 3.5 of 4.

BIMvoice:
Fixed in a3b39379 — implemented option (1), non-zero exit code, per your lean.

Convention check first: the CLI mixes `process.exit(N)` (hard errors, execution stops there) and `process.exitCode = N` (command finishes printing, then signals failure) — `validate.ts:349` and `ids.ts:85` are the closest analogues, both setting `process.exitCode` after writing their full output. `schema.ts` is the same shape: it still needs to print the (reduced) JSON after the catch, so `process.exitCode = 1` on the fallback path, not `process.exit()`.

`packages/cli/src/index.ts`'s dispatch just does `await schemaCommand(commandArgs); break;` and `main()` returns normally on success (no trailing `process.exit(0)` that would clobber it), so the exit code propagates. Checked for other callers: `gym.ts` has an unrelated `'schema'` check name (`computeSchemaChannel`, not `schemaCommand`) — confirmed by reading it, not guessed.

RED first: added `sets a non-zero exit code...` to `schema.test.ts`, ran it against the pre-fix source — 1 failed / 1 passed (`expected +0 to be 1`).

Then implemented the fix, GREEN: 2/2 passed.

Mutation test: removed the `process.exitCode = 1;` line (python3 exact-substring replace, count=1, asserted), re-ran — 1 failed / 1 passed, confirming the test catches it. Restored by file copy.

Full CLI suite (`packages/cli`, vitest): 265 passed / 8 skipped, unchanged from baseline aside from the one new test. `pnpm typecheck` clean across the repo.

stdout is untouched — still pure JSON, same shape, verified by parsing it in the new test.

One thing worth flagging rather than asserting past: this only exercises `schemaCommand` as a function (mocking `process.exitCode` and stdio) via vitest, not a real spawned `ifc-lite schema` subprocess — I did not add a subprocess-level check that the OS exit status is 1. `process.exitCode` is the correct and sufficient mechanism (Node sets the real process exit status from it on natural exit, and nothing in `index.ts` overrides it on the success path), but I want to be explicit that the test asserts the mechanism, not an observed OS-level exit status.

Updated the existing `.changeset/package-silent-catches-round2.md` (already touching `@ifc-lite/cli`) to mention the exit code instead of adding a second changeset.

louistrue:
Verified `a3b39379`. The exit code is the right call and it is pinned — but the **other direction is not**, and it is the more dangerous one.

Dropping `process.exitCode = 1` fails the new test:

```
× sets a non-zero exit code, since stderr alone is a channel a `schema | jq` caller discards
```

Asserting stdout still parses as JSON in the same test is the right pairing — it pins "signal the failure" and "do not corrupt the payload" together, so a future fix cannot satisfy one by breaking the other.

Also a nice catch restoring `process.exitCode` in the `finally`. Leaking `1` out of a test into the runner's own exit flag is a genuinely nasty way to redden an unrelated suite.

## Nothing asserts the success path exits 0

The negative control passes:

```
### NEGATIVE CONTROL: non-zero exit on the SUCCESS path too
      Tests  2 passed (2)
```

Both tests live under `describe('schemaCommand with the sandbox schema unavailable')`, so the file only ever exercises the failure branch. Add `process.exitCode = 1` to the successful load and the suite stays green.

That regression would be worse than the bug you fixed. Today a degraded schema is reported as success — bad, but only on the fallback path. An always-`1` exit fails **every** invocation, so any CI step running `ifc-lite schema` breaks outright, and the test that exists to protect the exit code would not notice.

This is the same asymmetry you closed yourself on #2089 an hour ago, and your comment there is the argument for fixing it here:

> `persisted` must be true here, not merely absent-of-false, or the flag could be hardcoded `false` and a check that only looks for `false` elsewhere would still pass.

Exactly the same shape: a one-directional assertion on a two-valued signal. One test with the sandbox schema loading normally, asserting `process.exitCode` is `0` (and stdout still parses), kills it.

Everything else on this PR is settled — the extract-walls scoping is correct and I confirmed #2104 really tracks the deferred case.

BIMvoice:
Closed in `97dd2fb8`. Your negative control now fails.

## The other direction

```
NEGATIVE CONTROL: process.exitCode = 1 on the SUCCESS path too
  → 1 failed | 2 passed
```

Before: 2 passed. The reason it could not be caught in `schema.test.ts` is structural rather than an oversight in the assertions — that file mocks `@ifc-lite/sandbox/schema` to throw **at module level**, so every test in it reaches the fallback branch by construction. No assertion added there could observe the success path. The other direction needs a mock that resolves, which means a second file: `schema.success.test.ts`.

It asserts three things, and the third is the anti-vacuity guard:

- `process.exitCode` is `0`
- stderr is empty — the success path must not borrow the failure signal in either channel
- the emitted JSON is the **mocked** schema (`namespace: 'query'`), not the fallback. Without this the test would pass for the wrong reason if the mock ever stopped applying: it would silently be exercising the fallback again, which does not set `exitCode` before the warning either.

And you are right that this is the same shape I closed on #2089 — a one-directional assertion on a two-valued signal. I did not carry the lesson across from one PR to the next on the same day, which is exactly the kind of thing that is obvious in hindsight and invisible from inside the change.

Your point about relative severity is the part worth keeping: an always-`1` exit is strictly worse than the bug, because it fails *every* invocation rather than only the degraded path, and it fails silently in the one place built to catch exit-code regressions.

`packages/cli`: 266 passed / 8 skipped (was 265 / 8), `tsc --noEmit` clean. stdout shape unchanged on both paths.

louistrue:
Verified `97dd2fb8`. The exit-code contract is now pinned in **both** directions:

```
### NEGATIVE CONTROL (previously PASSED): exitCode=1 on the SUCCESS path
     × leaves the exit code at 0 and writes nothing to stderr

### original direction
     × sets a non-zero exit code, since stderr alone is a channel a `schema | jq` caller discards
```

Putting it in its own file was the right call, and your reason is the interesting part — `schema.test.ts` mocks `@ifc-lite/sandbox/schema` to throw *at module level*, so nothing in that file can observe the success path at all. The gap was not an oversight in the assertions; it was structural, and no amount of adding assertions to the existing file would have closed it. Worth having written that down, because the next person to add a case there will hit the same wall.

Asserting "and writes nothing to stderr" alongside the exit code is a good pairing too: it pins that the success path is quiet, so a future change that starts warning unconditionally cannot hide behind a green exit code.

That also settles greptile's original thread — "machine callers that consume stdout cannot tell the reduced schema from the real one" is exactly what the non-zero exit fixes, with stdout's shape unchanged. Resolving it.

Gate: 19 check-runs with all 5 core lanes green, no failures, no red statuses. I am checking lane *presence* now rather than just absence of red, after nearly merging a PR today whose CI had silently never run.

Waiting on the CodeRabbit CLI run (hosted is `Review rate limited` again) and then this one is done — all four fixes in it verified, the extract-walls scoping confirmed correct against the real #2104, and the CLI half now honest to its own stated audience.

louistrue:
CodeRabbit CLI on `97dd2fb8` (hosted `Review rate limited` again): 5 findings. **One is a real crash path**, one I checked and it does not reproduce, three are minor.

## Real: a present-but-invalid export bypasses the whole fallback

`try` covers the import, but not the shape of what came back:

```ts
try {
  const mod = await import('@ifc-lite/sandbox/schema');
  schemas = mod.NAMESPACE_SCHEMAS;      // no validation
} catch (err) { ...fallback, warning, exitCode = 1... }

const output = schemas.map(ns => ({ ... }));   // OUTSIDE the try
```

With the export present but not an array:

```
PROBE_THREW=TypeError: Cannot read properties of undefined (reading 'map')
PROBE_EXITCODE=undefined
```

Uncaught, `exitCode` never set. The static schema, the stderr warning and the non-zero exit you just built are all bypassed — the command dies on a stack trace instead. An `Array.isArray(schemas)` check inside the `try`, throwing into the existing catch, routes it through the path that already does the right thing.

A caveat on my own evidence, because it nearly fooled me. My first probe mocked the module as `{}` — no export at all — and reported a clean fallback with `exitCode=1`. That was a **harness artifact**: vitest's mock throws on access to an undefined export, so the `catch` fired for a reason that does not exist in Node ESM, where a missing named export reads as `undefined`. Mocking `{ NAMESPACE_SCHEMAS: undefined }` is what reproduces the real shape, and that is the run above.

## Not reproducible: the exit-code leak from the first test

Flagged `major` — the first test in `schema.test.ts` does not save/restore `process.exitCode`, so it leaves `1` behind. It does leave `1` behind, but it does not matter:

```
$ vitest run src/commands/schema.test.ts
vitest exit code: 0
      Tests  2 passed (2)
```

Vitest sets the process exit status from its own results rather than inheriting `process.exitCode`. Tidying the first test to match the second is fine for symmetry, but there is no failure to fix and I would not hold for it.

## Three minor silent catches

`viewer-manager.ts` (globalId enrichment, SSE frame parse) and `lod1-generator.ts` (box-mesh fallback) each swallow an error with no diagnostic. They are the same shape as the four this PR fixes, in files it touches. Your call whether they belong here or in the next round — the LOD one already keeps `failed.push(el.expressID)` bookkeeping, so it is the least invisible of the three.

Everything else on this PR is verified: the exit-code contract now dies in both directions, and the extract-walls scoping is correct against the real #2104. The `Array.isArray` guard is the one thing I would want before merging.

BIMvoice:
`Array.isArray` guard added in `70affa9a`.

## The crash path

Reproduced with your corrected mock shape, then fixed. The guard sits inside the `try` and throws a `TypeError` naming which case it hit (`missing` vs `not an array`), so it routes through the catch that already writes the warning, emits the static schema and sets the exit code. New test asserts all three plus that `schemaCommand` resolves rather than rejects.

Neutralising the guard fails it; 267 passed / 8 skipped, `tsc --noEmit` clean.

## Your caveat was the important part, and I verified it rather than trusting it

You flagged that your first probe — mocking the module as `{}` — was a harness artifact, because vitest throws on access to an undefined export while Node ESM reads a missing named export as `undefined`. I checked what that would have meant for the test, since a mock that lies is worse than no test:

```
mock `{}` + guard removed  → 1 passed
```

So the naive mock would have produced a **green test against the unfixed command** — it would have "covered" the crash while the crash still happened. That is recorded in the test's docblock, with the reason, because the mock shape is the load-bearing part of this test and the obvious choice is the wrong one.

Worth noting the general shape, since it is the third variant today: #2105's fixture satisfied two hypotheses at once, #2100's original file could only reach one branch by construction, and here the mock reproduces a failure mode the runtime does not have. All three are tests that look like coverage and are not, and none of them are visible from the assertions.

## The other four

- **Exit-code leak in the first test** — agreed, not a defect. Vitest sets process status from its own results, so there is nothing to fail. I left it; happy to tidy for symmetry if you would rather.
- **Three silent catches** (`viewer-manager.ts` globalId enrichment and SSE frame parse, `lod1-generator.ts` box-mesh fallback) — I would take them in the next round rather than here. This PR is already four fixes plus two review rounds, and the LOD one keeps `failed.push(el.expressID)` bookkeeping so it is not fully invisible. Say the word if you would rather have them now.

louistrue:
Verified `70affa9a`. The shape guard closes the crash path — merging.

```
### baseline                                   Tests  4 passed (4)
### MUTATION: drop the Array.isArray guard
     × falls back, warns and exits non-zero instead of throwing
                                               Tests  1 failed | 3 passed (4)
```

Throwing into the existing `catch` rather than adding a parallel fallback is the right shape — one recovery path, and the warning/static-schema/exit-code trio stays in one place.

The note you put in the test's docblock is the part I most wanted to see written down:

> Mocking the module as `{}` makes vitest throw on access to an undefined export, so the `catch` fires for a reason that does not exist in Node ESM — where a missing named export simply reads as `undefined`. The test would then pass against the unfixed command.

That is exactly the trap I fell into on my first probe, and it is the sort of thing that silently converts a regression test into a no-op years later. `{ NAMESPACE_SCHEMAS: undefined }` reproduces the real failure; `{}` tests vitest.

The exit-code contract is now pinned in both directions across three files, each with its own module mock because the mocks are mutually exclusive at module scope — success, throwing import, and resolving-but-unusable. All 5 core lanes green, `MERGEABLE`, 0 unresolved.

I have left the three minor silent catches CodeRabbit flagged (`viewer-manager.ts` ×2, `lod1-generator.ts`) for a later round — same shape as what this PR fixes, but not worth another cycle here.

PR #2101, opened 2026-08-04
fix(viewer): only commit a clash preset change that actually persisted

BIMvoice:
Closes the item flagged on both #2085 and #2098 and deferred twice.

`deleteClashPreset`, `setClashPresetEnabled` and `resetClashPresets` called `savePresets(next)`, **discarded the `{ok:false}` `SaveResult`**, then `set(...)` unconditionally. The UI reported success while storage refused the write, and the change vanished on reload.

## Not a design question — the same file already answers it

`createClashPreset` (`:292-294`), `updateClashPreset` (`:299-301`) and `importClashPresets` (`:326-328`) each return the result and only `set` when `ok`, surfaced by `ClashSettingsDialog.tsx:116-120` and `:135-136`. The pattern is established **three times in the same file**; these three were left behind.

## Per action, following that precedent

- **delete** — the preset **stays listed**. Dropping it and having it reappear on reload is exactly the "user believes it saved" failure.
- **toggle** — the switch **does not move**. It is fully controlled by `p.enabled` from the store, so not committing *is* the revert: no local state to roll back, no flicker-then-snap-back.
- **reset** — **does not apply**. The most destructive of the three, and showing an empty rule set that silently resurrects on reload is the worst to get wrong.

One judgement beyond the three: `deleteClashPreset`'s early return for a built-in or unknown id reports `{ ok: true }`. Nothing was asked of storage, so a toast there would be a spurious error on a path that previously returned `void` silently. It has its own test and its own mutation.

**No caller is left unable to report** — all three have exactly one call site each, in `ClashSettingsDialog`, which already imports `toast`. Grep over `**/*.ts{,x}` found no others.

## Evidence

RED puts the **state** assertion before the `result.ok` check, so it proves the silent commit rather than just a missing return value:

```
not ok 2 - deleteClashPreset keeps the preset listed when the write is refused
  error: 'a delete that did not persist must not disappear from the list'
not ok 5 - setClashPresetEnabled leaves the toggle unchanged when the write is refused
not ok 7 - resetClashPresets keeps the custom rules when the write is refused
```

Tests 9–11 (the three siblings) **pass in RED**, proving they are untouched.

Dropping any one gate fails exactly its own test while the other five survive — I re-ran the delete one myself on the pushed tree (`REPLACEMENTS=1`, scoped to the function body since all six gate lines are now textually identical, gate count 6→5 confirmed): 1 failed / 10 passed, restored 11/11.

**Stub selectivity is asserted, not assumed** — test 1 checks the double fails `setItem` for the presets key **only**, that reads always work, that `SETTINGS_KEY` writes still succeed, and that the slice really loaded a seeded preset through the real `buildInitialPresets()` → `readStoredPresets()` path. That is the trap that produced three vacuous tests earlier today: a stub rejecting everything makes the module see no storage at all.

2591 passing (was 2580 — measured after building the workspace; unbuilt it reports a bogus 1717/82), typecheck 34/34.

## Same shape found and deliberately NOT fixed

`applyClashFlavorConfig` (`:365-378`) discards **both** `savePresets` and `saveSettings` after an unconditional `set(...)`, so activating a flavor whose write is refused shows the rule set applied and reverts on reload. Different callers and another `void` signature to widen, so it is flagged rather than folded in. The detection-tab setters (`:235-249`, `:265`) discard `saveSettings` too, but they are settings-only and commit-then-persist by design — worth a separate look.

## Sequencing

`lib/clash/persistence.ts` is **untouched**, so no conflict with #2089; this consumes `SaveResult` opaquely and only tests `.ok` / reads `.message`, so its `'unreadable'` reason needs nothing here. I did **not** import #2091's `lib/storage/save-result.ts` — it is not on main. `SaveResult` now has three near-identical definitions (clash, lens, scripts); worth unifying once those land, but not in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clash preset changes now report persistence failures instead of failing silently.
  * Presets are only updated in the viewer after changes are successfully saved.
  * Resetting, enabling, disabling, and deleting presets now preserve the existing state when saving fails.

* **Tests**
  * Added coverage for successful and failed preset persistence scenarios, including create, update, import, reset, and deletion workflows.

### DIFF
--- a/apps/viewer/src/components/viewer/ClashSettingsDialog.tsx
+++ b/apps/viewer/src/components/viewer/ClashSettingsDialog.tsx
@@ -32,7 +32,7 @@ import { cn } from '@/lib/utils';
 import { toast } from '@/components/ui/toast';
 import { useViewerStore } from '@/store';
 import { matchesSelector, type ClashSeverity } from '@ifc-lite/clash';
-import { exportPresets, importPresets, type ClashPreset } from '@/lib/clash/persistence';
+import { exportPresets, importPresets, type ClashPreset, type SaveResult } from '@/lib/clash/persistence';
 
 const SEVERITY: Record<ClashSeverity, { label: string; color: string }> = {
   critical: { label: 'Critical', color: '#f7768e' },
@@ -141,6 +141,13 @@ export function ClashSettingsDialog({ trigger }: ClashSettingsDialogProps) {
     [importClashPresets],
   );
 
+  // Delete / toggle / reset only commit when the write landed (clashSlice), so
+  // a refused write leaves the row, the switch and the rule set exactly as they
+  // were — all this has to add is the reason.
+  const reportSaveFailure = useCallback((result: SaveResult) => {
+    if (!result.ok) toast.error(result.message);
+  }, []);
+
   const enabledCount = useMemo(() => presets.filter((p) => p.enabled).length, [presets]);
 
   return (
@@ -239,7 +246,7 @@ export function ClashSettingsDialog({ trigger }: ClashSettingsDialogProps) {
                 <Plus className="h-3.5 w-3.5 mr-1" /> Add rule
               </Button>
               <div className="ml-auto flex items-center gap-1">
-                <Button variant="ghost" size="sm" className="h-7 px-2 text-xs" title="Reset to the built-in rules" onClick={resetPresets}>
+                <Button variant="ghost" size="sm" className="h-7 px-2 text-xs" title="Reset to the built-in rules" onClick={() => reportSaveFailure(resetPresets())}>
                   <RotateCcw className="h-3.5 w-3.5" />
                 </Button>
                 <Button variant="ghost" size="sm" className="h-7 px-2 text-xs" title="Export rules" onClick={() => exportPresets(presets)}>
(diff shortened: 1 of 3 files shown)


louistrue:
Verified `24c9f572`. The consistency argument is the strongest form of "this should change" — three functions in a file diverging from a pattern the other three already establish, with no reason for the split.

Mutation, with the guard count asserted before replacing (6 sites, all reverted to setting unconditionally):

```
not ok 2  - deleteClashPreset keeps the preset listed when the write is refused
not ok 5  - setClashPresetEnabled leaves the toggle unchanged when the write is refused
not ok 7  - resetClashPresets keeps the custom rules when the write is refused
not ok 9  - createClashPreset still gates its commit on the SaveResult
not ok 10 - updateClashPreset still gates its commit on the SaveResult
not ok 11 - importClashPresets still gates its commit on the SaveResult
```

Six, not three. Pinning the three that already worked is the part worth calling out: the defect here was that half the file drifted from its own convention, so a test that only covers the newly-fixed half leaves the same drift free to happen again in the other direction. Now all six are held.

The failure mode this closes is the one that keeps recurring — the UI reported success while storage refused, and the change vanished on reload. A user cannot distinguish "saved" from "silently discarded" until the next session.

11/11 green.

---

One correction to my own method: my first mutation attempt used a regex that matched nothing (`anchor found: False`) and the suite passed 11/11 — which reads exactly like "the fix is unpinned". I only caught it because I printed the match result. The second attempt asserted the site count before replacing and asserted zero guards remained afterward. That is the third time today a no-op mutation nearly became a finding, and it is the reason I now count occurrences before trusting any green mutation run.

PR #2102, opened 2026-08-04
fix(viewer): reopen the model cache when its connection dies underneath us

BIMvoice:
The root cause behind the defect #2098 makes non-fatal.

`openDatabase()` memoises the connection in `dbPromise` and clears it **only** in `request.onerror`. Nothing clears it when the connection is later closed — by another tab's `versionchange`, or the browser reclaiming under storage pressure. After that, `db.transaction()` throws `InvalidStateError` synchronously and **every cache operation fails for the tab's lifetime**: the model cache silently stops working, and every load is a miss, forever.

#2098 stopped that breaking the load. This stops it happening.

## Design — two layers

Both extend the in-repo memo-reset idea (`duckdb-integration.ts:78`, `pointcloud/streaming/memo.ts`) for the part neither covers: **a memoised value going bad *after* it resolved**.

- **Proactive** — `db.onversionchange` closes and invalidates, so we stop blocking another tab's upgrade or delete; `db.onclose` covers an abnormal close.
- **Reactive** — `withConnection(use)` retries a synchronous `use(db)` once on `InvalidStateError`.

## The three concurrency decisions

**Concurrent callers share one reopen.** `invalidateConnection(dead)` is **identity-checked**, so of N callers tripping over the same dead handle the first clears the memo and the rest no-op, then all collapse onto one in-flight open. Asserted: 4 concurrent reads → `indexedDB.open` called **exactly once**.

**Detection is narrow.** Only the synchronous `InvalidStateError` from `transaction()` counts. Nothing else in these ops produces that name — an inactive transaction gives `TransactionInactiveError`, a read-only write gives `ReadOnlyError`. The proactive handlers mean the common cases never reach the throw at all.

**Exactly one reopen per operation.** A second failure propagates to the caller's non-fatal handler, so an unopenable database costs at most one extra open per op rather than a retry loop. The retry is idempotent because the only retryable failure happens **before a transaction exists** — nothing was written, so the replay replays a no-op.

**Deliberate non-coverage, documented at the catch:** if the connection dies inside `setCached`'s quota guard *after* the liveness probe, the write is skipped non-fatally rather than retried, because a retry would also re-run LRU eviction.

## Independently corroborates #2098

On `upstream/main` the "degrade to a miss" catch **does not fire** for anything inside the returned promise — `return new Promise(...)` inside a `try` does not route that promise's rejection to the `catch`. In RED that made `clearCache()` reject out of a `beforeEach`. This change removes that escape for the transaction path; the `request.onerror` path is #2098's and is left alone.

## Evidence

| mutation | result |
|---|---|
| kill the reactive reopen | **4 fail** |
| drop the identity check | 1 fail — the concurrency test, opens > 1 |
| drop `onversionchange` close+invalidate | 1 fail — another tab's delete blocked/hung |
| drop the liveness probe | 2 fail — both write paths |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, line re-read): 4 failed including *"the cache stays usable for every later operation, not just the first"*, restored 18/18.

**Negatives, all asserted:** healthy ops reopen **zero** times; an unopenable database degrades to `null`, does not reject on write, costs ≤2 opens per op, and **recovers** once open works again; another tab's `deleteDatabase` completes rather than blocking.

**Non-vacuity:** `closeMemoisedConnection()` asserts `db.transaction()` really throws `InvalidStateError` before each test proceeds, so a test cannot pass by never reaching the dead state. `fake-indexeddb` provides the real `versionchange`/blocked semantics — a one-event stub cannot express this, which is what defeated an earlier attempt on `recent-files.ts`.

18 passing in `ifc-cache.test.ts` (was 11).

## One measurement caveat, stated rather than papered over

The full-suite number from that worktree is **not** comparable: it had no `node_modules`, and with a borrowed install 80 test *files* fail to load on missing transitive deps (sampled: `Cannot find package 'jszip'` from `packages/parser`). Those 80 are identical before and after and unrelated, but a clean install is needed for a true total. The per-file numbers above were re-run against a proper install and are sound.

### DIFF
--- a/apps/viewer/src/services/ifc-cache.test.ts
+++ b/apps/viewer/src/services/ifc-cache.test.ts
@@ -225,3 +225,172 @@ describe('ifc-cache LRU eviction correctness (blocker #2)', () => {
     assert.ok(c && await bytesEqual(c.buffer, patterned(2 * KB, 12)), 'survivor is uncorrupted');
   });
 });
+
+// ---------------------------------------------------------------------------
+// A connection closed underneath the memo (issue: the cache dies for the tab)
+// ---------------------------------------------------------------------------
+
+type OpenFactory = typeof indexedDB.open;
+
+/** Count `indexedDB.open` calls; returns a restore fn. */
+function countOpens(): { calls: () => number; restore: () => void } {
+  const factory = indexedDB as unknown as { open: OpenFactory };
+  const real = factory.open.bind(indexedDB) as OpenFactory;
+  let calls = 0;
+  factory.open = ((name: string, version?: number) => {
+    calls++;
+    return real(name, version);
+  }) as OpenFactory;
+  return { calls: () => calls, restore: () => { factory.open = real; } };
+}
+
+/** Replace `indexedDB.open` with one that always fails asynchronously. */
+function breakOpen(): { calls: () => number; restore: () => void } {
+  const factory = indexedDB as unknown as { open: OpenFactory };
+  const real = factory.open.bind(indexedDB) as OpenFactory;
+  let calls = 0;
+  factory.open = (() => {
+    calls++;
+    const request = {
+      error: new DOMException('simulated open failure', 'UnknownError'),
+      onerror: null as null | (() => void),
+      onsuccess: null,
+      onupgradeneeded: null,
+    };
+    setTimeout(() => request.onerror?.(), 0);
+    return request as unknown as IDBOpenDBRequest;
+  }) as OpenFactory;
+  return { calls: () => calls, restore: () => { factory.open = real; } };
+}
+
+/**
+ * Put the module's memoised connection into the state this suite is about: the
+ * connection object is still memoised but is closed, so `db.transaction()`
+ * throws `InvalidStateError` synchronously. Asserts that state is really
+ * reached, so none of the tests below can pass vacuously.
+ */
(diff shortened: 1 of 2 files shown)


louistrue:
Verified `6b4fdc21`. Both halves are pinned, and the identity check has its own test — which is the part I would have expected to be missing.

| mutation | fails |
|---|---|
| drop `if (memoisedDb !== dead) return;` | `concurrent operations after a close share ONE reopen` |
| never record the live connection (`memoisedDb = db`) | **4** tests: read-after-close, write-after-close, `stays usable for every later operation, not just the first`, and the concurrent one |

The identity check is the subtle half and it is easy to get wrong in the direction nothing notices: without it, a caller that trips over a dead connection clears a memo another caller has *already* replaced with a live one, so the reopen thrashes instead of being shared. The test name says exactly that — "share ONE reopen" — and it fails when the check goes.

`the cache stays usable for every later operation, not just the first` is the other one worth having. A fix that reopened once and then re-memoised nothing would satisfy the first two tests and still leave the tab broken from the second failure onward, which is the original bug wearing a different hat.

18/18 green.

This being the root cause behind #2098 is the right framing: #2098 stopped a dead connection breaking the load, this stops the connection staying dead. Worth keeping both — the first is the guard for cases this one does not anticipate.

PR #2103, opened 2026-08-04
fix(sandbox): reset the log budget per eval, not per sandbox

### LINKED ISSUE #2099: Sandbox: the 4MB log budget latches per-sandbox, silencing every later eval
## What happens

`buildConsole`'s log budget is scoped to the **sandbox**, not to the eval. `totalBytes` and `truncated` are closed over once at construction:

```ts
const MAX_LOG_ENTRIES = 1000;
const MAX_TOTAL_BYTES = 4 * 1024 * 1024;
let totalBytes = 0;
let truncated = false;
```

`Sandbox.eval()` clears `this.logs` at the start of each run, but nothing resets those two. So once a single script pushes past 4MB (or 1000 entries), `truncated` latches `true` and **every later script on the same sandbox** gets one `[log output truncated: limit reached]` entry and nothing else — for the life of the sandbox.

## Why it matters

Sandboxes are reused. `packages/sdk`'s `activeSandbox` keeps one alive across calls, and the extension host holds one per extension. So the observable behaviour is: run one log-heavy script, and every subsequent script in that session appears to produce no output at all — with no indication the limit belongs to an earlier run rather than the current one.

A user debugging the second script sees empty logs and reasonably concludes their `console.log` calls are not firing.

## Fix

Reset both alongside the existing `this.logs.length = 0` in `eval()`. That likely means exposing a reset from `buildConsole` (it currently returns `{ logs, dispose }`) or moving the counters somewhere `eval()` can reach.

A test should assert that a script which trips the limit does not suppress the *next* script's output — the property nothing currently enforces.

## Relationship to #2084

Same file, different mechanism. #2084 is about unserializable args contributing **zero** to `totalBytes` so the byte cap under-counts. This one is about the cap, once tripped, never resetting between evals. They are independent and either can be fixed without the other, though both live in the same handler and a single pass could reasonably address both.

## Not a regression

Pre-existing. Surfaced during review of #2095 (which fixed the adjacent `ScriptError.logs` aliasing, #2092) and deliberately not folded in — that PR was a focused two-line fix with two tests, and widening it would have buried this.

BIMvoice:
Closes #2099.

## The defect is worse than the issue states

#2099 says a later script "gets one `[log output truncated]` entry and nothing else". In fact the marker is pushed only on the run that trips the cap; **every later eval hits `if (truncated) return;` at the top of the handler and gets zero entries.** RED shows `expected [] to deeply equal [['second']]` — a completely empty array.

**Sandboxes really are reused**, verified on `upstream/main`:
- `packages/sdk/src/namespaces/sandbox.ts:150-158` — `eval()` reuses `this.activeSandbox` across every call.
- The extension host returns the cached activation record while `sandbox.isDisposed` is false (`host/runtime.ts:213-218`), and `host-commands.ts:136` / `host-exporters.ts:195` call `activation.sandbox.run(...)` per invocation, mapped to `sandbox.eval` at `sandbox-factory.ts:166`. One sandbox per activated extension, re-entered for every command.
- **Not affected:** `useSandbox.ts:184-224` creates a fresh sandbox per run and disposes it in `finally`.

`MAX_LOG_ENTRIES` had the same problem by a different route: its *counter* was already fine (it compares `logs.length`, cleared per eval), but tripping it set the same `truncated` latch and silenced later evals identically. That path has its own test and failed RED too.

## What the caps are for — which is what decides the fix

The comment at `bridge.ts:67-70` frames them against a single runaway script (`for(;;) console.log('x'.repeat(1e6))` before the timeout fires) — a **per-run** threat model. The code agrees: `eval()` clears the buffer every run, and each `ScriptResult`/`ScriptError` gets a **copy** (`sandbox.ts:282`, `:351`), so the sandbox itself never retains more than one run's logs.

So the per-sandbox `totalBytes` bounded nothing the sandbox holds — it only stopped later scripts from logging. It was simply missed when `this.logs.length = 0` was written.

**Reset per eval**, and made structurally hard to drift again: `buildConsole` now returns a `resetLogs()` that clears the buffer **and** both counters, and `eval()` calls only that rather than touching the array. A null `resetLogs` throws in the init guard rather than silently skipping the clear.

**The bound genuinely changes** — 4 MB per eval instead of 4 MB per sandbox-lifetime for an embedder retaining every `ScriptResult` — and the changeset says so.

**Deliberately not chosen:** keep it per-sandbox but surface the truncation. The bound it provides is illusory (an embedder can hold N results from N sandboxes anyway) and it leaves the silencing in place.

## Evidence

| mutation | result |
|---|---|
| `this.resetLogs()` → `this.logs.length = 0` (pre-fix) | **exactly the 4 cross-eval tests fail**, negative stays green |
| neutralise `totalBytes = 0` | 3 fail |
| neutralise `truncated = false` | 3 fail |
| neutralise `logs.length = 0` | 6 fail |

I re-ran the first myself on the pushed tree (`REPLACEMENTS=1`, line re-read): 4 failed / 58 passed, restored 62/62.

**Non-vacuity:** every cross-eval test first asserts script A's own logs end in `[log output truncated: limit reached]` **at the exact expected index** — 5 entries for the byte cap, 1001 for the entry cap — so the budget provably was reached rather than the test passing on a run that never logged. The test uses a plain `console.log` loop, never an allocation, so it cannot poison the WASM module for the rest of the worker.

One self-inflicted RED failure was my own bug and is worth noting: a top-level `const payload` redeclared on the second run in the same realm. The script is now an IIFE, with a comment saying why.

Negative cases: a single script genuinely over the byte cap is still truncated at exactly the same entry (4 full 1 MB entries + marker), and the entry cap still gives 1000 + marker. Both pass before *and* after and stay green under every mutant.

62 passing (was **57** — measured on `upstream/main` after building `@ifc-lite/sdk`, without which vitest cannot resolve it in a fresh worktree), `tsc --noEmit` clean, `check-test-wiring` passes.

## #2096 sequencing

Open, not merged. Same function, **different hunks** — it rewrites the `JSON.stringify` catch and appends helpers; this touches the counter declarations, the signature and `buildBridge`'s return type. Git merges them mechanically, and the test filenames differ.

One decision for whoever lands second: #2096 adds a `sizingWarningShown` latch in the same closure. **It should not go into `resetLogs` by reflex** — its rationale is about the *context* (a script-controlled trigger must not flood the host console), not the run. Left untouched here so that stays a deliberate call.

`buildBridge`'s return type gains `resetLogs`, which is additive public surface since it is exported from `index.ts` — noted in the changeset.

## Summary by CodeRabbit

* **New Features**
  * Console log budgets now reset for each sandbox evaluation, allowing independent log capture across repeated runs.
  * Direct bridge usage now provides a way to reset captured logs and budget state.

* **Bug Fixes**
  * Prevented log truncation from carrying over between evaluations.
  * Preserved captured logs when a later evaluation fails.

* **Documentation**
  * Clarified bridge behavior and per-evaluation log-budget handling.

### DIFF
--- a/.changeset/sandbox-log-budget-per-eval.md
+++ b/.changeset/sandbox-log-budget-per-eval.md
@@ -1,0 +1,13 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Reset the captured-log budget on every `eval()`, so one log-heavy script no longer silences every later script on the same sandbox.
+
+The bridge caps captured console output twice — by entry count (1000) and by cumulative serialized size (4 MB) — because `vm.dump` copies sandbox values onto the host heap, which the QuickJS memory limit does not bound. `eval()` clears the log buffer in place at the start of every run and hands each result its own copy, so the caps bound one run's output; but `totalBytes` and the `truncated` latch were closed over at construction and never reset. Once a script tripped either cap, `truncated` stayed `true` for the life of the sandbox and the console handlers returned immediately, so every subsequent `eval()` on that sandbox produced **no log entries at all** — not even the truncation marker, which is pushed only on the run that trips the cap.
+
+This affects embedders that reuse a sandbox across evals, which is the normal path for two of them: `bim.sandbox.eval()` keeps one `activeSandbox` alive across calls, and the extension host holds one sandbox per activated extension and re-enters it for every command and exporter invocation. (The viewer's script editor creates a fresh sandbox per run and was never affected.) The observable symptom was a script whose `console.log` calls appeared not to fire, with nothing to indicate the limit belonged to an earlier run.
+
+`buildConsole` now owns the reset: it returns a `resetLogs()` that empties the buffer *and* zeroes both counters, and `eval()` calls only that instead of clearing the array itself — the two can no longer drift apart. A single script that genuinely exceeds either cap is truncated exactly as before.
+
+**Embedder-visible:** captured output is now bounded per eval rather than per sandbox, so a long-lived sandbox can hand back up to 4 MB of logs per run instead of 4 MB in total. Embedders that retain every `ScriptResult` across many runs hold correspondingly more. `buildBridge()`, exported for advanced embedders, gains a `resetLogs` property on its return value; existing callers are unaffected.
--- a/docs/api/typescript.md
+++ b/docs/api/typescript.md
@@ -990,6 +990,8 @@ Scripting SDK: the `bim.*` API for BIM automation. `createBimContext` builds a `
 
 QuickJS-in-WASM sandboxed script execution: `createSandbox` / `Sandbox`, `buildBridge` (marshals the `bim.*` API across the sandbox boundary), and `transpileTypeScript`.
 
+`buildBridge` returns `{ logs, resetLogs, dispose }`. `resetLogs` is part of the contract, not an implementation detail: the console capture budget (byte total and entry count) is scoped to **one run**, so a caller driving `buildBridge` directly must invoke `resetLogs()` at the start of every run. Skip it and a script that exhausts the budget silences the logs of every later run on the same bridge. `Sandbox.eval` already does this for you — only direct `buildBridge` callers carry the obligation.
+
 ## @ifc-lite/extensions
 
 Extension manifest, capability grammar, and slot registry for user customization: `validateManifest`, `migrateManifest`, `SlotRegistry`, capability and `when`-clause evaluation, bundle and storage helpers.
--- a/packages/sandbox/src/bridge.ts
+++ b/packages/sandbox/src/bridge.ts
@@ -23,19 +23,22 @@ import { buildSchemaNamespaces, disposeSchemaNamespaceSession, type BridgeCallCo
 
 /**
  * Build the `bim` API object inside the QuickJS VM.
- * Returns captured log entries from console.* calls.
+ *
+ * Returns the captured log entries from console.* calls, plus `resetLogs` —
+ * the caller must invoke it at the start of every run, because the capture
+ * budget is scoped to one run (see `buildConsole`).
  */
 export function buildBridge(
   vm: QuickJSContext,
   sdk: BimContext,
   permissions: SandboxPermissions = {},
   context: BridgeCallContext,
-): { logs: LogEntry[]; dispose: () => void } {
+): { logs: LogEntry[]; resetLogs: () => void; dispose: () => void } {
   const perms = { ...DEFAULT_PERMISSIONS, ...permissions } as Required<SandboxPermissions>;
   const logs: LogEntry[] = [];
 
   // ── console.log / warn / error / info ──────────────────────
-  buildConsole(vm, logs);
+  const resetLogs = buildConsole(vm, logs);
 
   // ── bim global ─────────────────────────────────────────────
   // The handle is freed in a `finally`: if namespace construction throws,
@@ -53,6 +56,7 @@ export function buildBridge(
 
   return {
     logs,
+    resetLogs,
     dispose: () => {
       disposeSchemaNamespaceSession(context);
     },
@@ -61,7 +65,7 @@ export function buildBridge(
 
 // ── Console ──────────────────────────────────────────────────
 
-function buildConsole(vm: QuickJSContext, logs: LogEntry[]): void {
+function buildConsole(vm: QuickJSContext, logs: LogEntry[]): () => void {
   const consoleHandle = vm.newObject();
 
   // vm.dump copies sandbox strings onto the host JS heap, which is NOT bound by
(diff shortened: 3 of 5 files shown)


louistrue:
Verified `738eb698`. This closes #2099 and covers both halves of it plus one I had not separated out.

Mutation — removing the single `this.resetLogs()` call (site count asserted first) fails five named tests:

```
× does not let a byte-budget-exhausting script silence the next one
× does not let an entry-cap-exhausting script silence the next one
× charges each eval its own budget rather than accumulating across evals
× does not leak one eval's logs into the next result
× keeps two caught errors' logs independent of each other
```

The issue described the byte budget; the entry cap latches identically and is tested separately, which is right — they are two independent counters and a fix for one would not necessarily reset the other. The accumulation test is the one I had not thought to ask for: a reset that emptied the buffer but not `totalBytes` would pass both latch tests and still under-report every later run's remaining headroom.

Threading `resetLogs` out of `buildConsole` rather than exposing the counters is the right seam — the caller gets one thing to call, and the budget's scope stays with the code that owns it. The doc comment on the return type states the contract ("the caller must invoke it at the start of every run"), which is what stops a future caller reintroducing the latch by forgetting.

62/62 green.

---

One note on my own verification, because the first run would have produced a false negative: the initial baseline showed **7 of 10 test files failing to load** on an unresolved `@ifc-lite/sdk`, with 29 tests passing. Running the mutation against that state gave an identical 29 — which reads exactly like "the fix is unpinned". It was not; the files that test it had not loaded. Building the sdk chain first gives 62/62, and the mutation then kills five.

That is the same class as the Rust fixture-dark problem: a partially-loaded suite reports a number, and the number looks like a result.

louistrue:
Verified `738eb698` and it closes #2099 properly — I posted the mutation result earlier (five named tests). Two items from a local CLI run, and I would take one.

## The concurrency MAJOR: real shape, but not this PR's to introduce or fix

`Sandbox.eval` does not serialize. Two overlapping calls on one sandbox share `logs`, `totalBytes` and `truncated`, so a second `eval` starting mid-flight resets state the first is still filling.

But `main` already clears the shared buffer per run — `sandbox.ts:147`, `this.logs.length = 0` — so concurrent evals already clobbered each other's logs before this change. What this PR adds is resetting the two counters alongside the buffer, which is the same sharing, not new sharing.

There is one direction where the new code differs: previously `truncated` latched, so a concurrent second eval could not un-truncate the first; now it can. That is a narrower window inside an already-broken scenario rather than a new failure mode, and fixing it properly means serializing `eval` — a public-API behaviour change on a published package, well beyond a budget-reset fix.

Worth its own issue. I would not widen this PR for it, and I would not want the "reset per eval" fix held up by a concurrency question that predates it.

## The docs item I would take

`docs/api/typescript.md` describes `buildBridge`'s return shape and does not mention `resetLogs`. The changeset's own framing makes that matter more than usual — the doc comment on the return type says "the caller must invoke it at the start of every run", so `resetLogs` is a contract, not an implementation detail. A caller reading the published shape and not seeing it has no way to know the obligation exists.

Small, and it keeps the guide honest about a seam this PR deliberately made explicit.

BIMvoice:
Both items handled — docs one taken in `71bfc1b2`, concurrency one split out as #2110.

## The docs item

Taken, with one correction to the premise. `docs/api/typescript.md` does not describe `buildBridge`'s return shape — the `@ifc-lite/sandbox` section is a one-line package summary that mentions `buildBridge` only in passing. So there was no return-shape description omitting `resetLogs`; there was no return-shape description at all.

That makes your point stronger rather than weaker: a caller reading the guide had no way to learn the obligation exists, and the JSDoc stating it (`bridge.ts:24-30`) is only visible to someone already reading the source or the `.d.ts`.

Added to that section, and I verified each claim before writing it rather than paraphrasing the JSDoc:

- returns `{ logs, resetLogs, dispose }` — the signature at `bridge.ts:31-36`;
- `resetLogs` clears the buffer, the byte total **and** the `truncated` flag — `bridge.ts:88-92`, all three in one function;
- `Sandbox.eval` already calls it (`sandbox.ts:155`), so **only direct `buildBridge` callers carry the obligation**. Worth stating explicitly — without it the note reads as if every sandbox user must do something, which would be false.

Checked the region is not machine-generated before hand-editing (`docs:check-generated` covers 4 regions; this is not one of them, and it reports clean). `docs:check-samples` clean: 257 snippets across 41 docs.

## The concurrency item

Agreed on all three counts, and filed as #2110 rather than widened into this PR.

Your framing is what I recorded: the sharing predates this change (`main` already cleared the buffer per run), and the one direction that genuinely differs — `truncated` no longer latching, so a concurrent second eval can un-truncate the first — is a narrower window inside an already-broken scenario, not a new failure mode. If two evals overlap, the first eval's logs were already being destroyed by the second's buffer clear; trading "wrong and flagged truncated" for "wrong" is not worth holding a fix for.

The issue lays out four remedies (serialize / per-eval capture state / reject overlapping / document) and states plainly that I have not implemented any — the choice is a call about the published contract, since all four change what callers observe. I also noted it is not currently reachable from the app: the viewer drives one eval at a time, so it needs an embedder or extension host that fans out.

PR #2105, opened 2026-08-04
fix(export): report a refused map conversion instead of dropping it silently

### LINKED ISSUE #2067: Georef export silently drops a requested IfcMapConversion when no source context survives
Split out of #2059 at @louistrue's suggestion, so that PR stays a georef-resolution fix rather than growing into an overlay-resolution refactor. Neither half below is a regression from #2059 — both predate it.

## Confirmed

With the only source context deleted and no replacement, the requested map conversion is silently dropped while the CRS is still written:

```
#31=IFCPROJECTEDCRS('EPSG:1234',$,$,$,$,$,#2);
   (no IFCMAPCONVERSION line, no surviving IFCGEOMETRICREPRESENTATIONCONTEXT)
```

The caller asked for a map conversion and got a lone `IfcProjectedCRS` back with no indication anything was refused. Declining to emit a dangling `SourceCRS` reference is correct — an `IFCMAPCONVERSION` pointing at a deleted `#id` would be an invalid file, which is the defect #2059 fixes. Doing it **silently** is the part worth revisiting.

Reachable without any overlay-created entity: deleting every `IFCGEOMETRICREPRESENTATIONCONTEXT` is enough.

## Not established — deliberately recorded as unknown

Whether an **overlay-created** replacement context is picked up on this path is **unverified in either direction**. @louistrue attempted to probe it and reported that his mock store's `createEntity` allocated expressId `1`, colliding with the existing `IfcProject` at `#1`, so the run was untrustworthy — and he reported the collision rather than the reading it produced, on the grounds that a probe with a colliding id can manufacture a pass as easily as a failure.

That is the right call and it is why this half is written as open rather than as a second finding. **Anyone testing this should use a store whose id allocator starts above the fixture's maximum expressId** (`setExpressIdWatermark` exists for this).

## Suggested direction

Options, in rough order of blast radius:

1. Surface the refusal — return or log that a requested map conversion could not be emitted, rather than the caller inferring it from the absence of a line.
2. Fall back to an overlay-created surviving context if one exists (this is the unverified half — establish the current behaviour first).
3. Leave the behaviour and document it at the call site, so the silence is at least intentional and discoverable.

Worth deciding which, rather than defaulting to (3) by inaction.

## Context

`packages/export/src/step-exporter.ts`, the georeferencing replacement path. #2059 changed it to resolve `IFCPROJECTEDCRS`/`IFCMAPCONVERSION` through `effective.byType` instead of the raw index, and threaded `effective` through `findPreferredGeometricRepresentationContextId` / `findLengthUnitReference` / `findUnitId` so the create branch cannot reference deleted ids. This issue is about what happens when that resolution legitimately finds nothing.

BIMvoice:
Closes #2067.

## Two corrections to the issue, both from probing rather than reading

**#2059 is not merged**, and the repro quoted in the issue needs it. But **the same silent drop reaches `main` today with no deletion at all** — a file that never had a representation context. That is the base-independent form of the defect and is what the RED test uses, so this does not depend on #2059 landing first.

**The overlay half — which you and I both left explicitly unresolved — is now established.** An overlay-created replacement context is **not** picked up, before or after #2059:

- on `main`, because `findPreferredGeometricRepresentationContextId` reads the **raw** index, which holds no overlay records;
- on #2059, because it filters on `entityIndex.byId`, and overlay-created ids carry no source bytes for `extractEntity`.

Probed with the watermark set above the fixture maximum, so the created id was **15** against max **14** — avoiding the id-1 collision that made you discard your own probe rather than trust its reading.

## That evidence also rules out direction (2)

The created context came out as:

```
#15=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,0.00001,$,$)
```

`WorldCoordinateSystem` is `$`, and IFC4 makes it mandatory. Falling back to it would trade a missing line for a **schema-invalid** `SourceCRS` — worse than the silence.

So direction **(1)**, surface the refusal. (3) alone leaves the caller unable to distinguish refusal from no request, which is the whole complaint.

## No new channel invented

`MergeExportResult.stats.warnings: string[]` already exists in this package (`merged-exporter.ts:409`, also `demesh-writer.ts:77`) and the viewer already renders such arrays. `StepExportResult.stats` now carries the same field. The pre-existing `console.warn` stays and **shares one message constant** with the returned entry, so the two cannot drift.

Populated at **both** return sites — including the delta-only early return, where a georef-only export can refuse and then have nothing else to write. That has its own test, and mutating only that site fails only that test.

## Evidence

Dropping the warning push fails **exactly the three refusal tests** — I re-ran that myself on the pushed tree (`REPLACEMENTS=1`, line re-read): 3 failed / 348 passed, restored 351/380.

**Non-vacuity is structural here:** in RED, all three refusal tests failed *only* at `expect(result.stats.warnings)`, **after** the content assertions (CRS emitted, no `IFCMAPCONVERSION`, no dangling refs) had already passed. So the request was made, the CRS was written, and only the report was missing — the test cannot pass by never requesting georef.

Both happy-path georef tests assert `stats.warnings` is `[]` and pass, so a fix that warned on the happy path would fail them. A caller requesting no georef also asserts `[]`.

351 passing (was 347), with the same single pre-existing `lod1-generator.test.ts` load failure from unbuilt `@ifc-lite/wasm` — worktree staging, unchanged before and after.

## One wording change worth flagging

I reworded the console line from *"no IfcGeometricRepresentationContext found in source file"* to *"…is available to reference as SourceCRS"*. Once #2059 lands, the deleted context **is** in the source file, so the old wording would become false. No test or caller greps that string.

## Incidental, not fixed and not in scope

With the context deleted, `#1=IFCPROJECT(…,(#14),…)` still references it — a dangling ref that neither #2059 nor this change addresses. Flagging rather than folding in.

## Summary by CodeRabbit

- **New Features**
  - Export results now include warnings when requested georeferencing cannot be applied.
  - Warnings identify missing geometric context or projected coordinate reference data.
  - Warning reporting is consistent across standard and delta-only exports.

- **Bug Fixes**
  - Prevented incomplete georeferencing references when map conversion requests are refused.
  - Successful exports continue to report an empty warnings list.


louistrue:
Verified `53ea43dd`. This closes the half of my #2059 comment I could establish, and the coverage is better than the single fix site suggests.

Mutation — removing the one `warnings.push` (site count asserted) fails three named tests:

```
× reports the refusal when creating both the CRS and the conversion
× reports the refusal when the CRS already exists and only the conversion is new
× reports the refusal even when the delta export has nothing else to write
```

Three routes to the same refusal, each pinned. The third is the one that would otherwise rot: a delta export with nothing else to write is the case where a caller is least likely to inspect the output and most likely to treat "no error" as "nothing to do".

The `stats.warnings` shape is the right call over throwing. Refusing to write a dangling `SourceCRS` is correct behaviour, not an error — the export succeeded at everything it could do. A caller that ignores `warnings` is no worse off than today; one that reads it can now tell "no map conversion requested" from "requested and refused", which was the actual defect.

Matching `MergeExportResult.stats.warnings`' existing `string[]` shape rather than inventing a structured type is the DRY instinct, and sharing one message constant between the returned entry and the console line is what stops the two drifting — the failure mode this session has produced more of than any other.

351 passing. (`lod1-generator.test.ts` fails to load on an unbuilt `@ifc-lite/wasm` — worktree staging, not this branch.)

Spinning this out of #2059 rather than widening it was right, and it is the third time today that has worked well: #2081 out of #2080's finding, #2103 out of #2099, this out of #2059. Each stayed reviewable and each got its own tests.

The other half of my #2059 comment — whether an overlay-created replacement context resolves — remains unestablished. I could not test it cleanly because the mock store allocated a colliding expressId, and I would rather leave it open than claim a result that probe could not support.

louistrue:
One item from a local CLI run, and it is the same class the shared constant was introduced to prevent — just one level in.

`MAP_CONVERSION_WITHOUT_CONTEXT_WARNING` ends with:

> The requested IfcProjectedCRS **was written** without it.

That is true on the path your first test covers (creating both the CRS and the conversion). It is not true on the path your second test covers — *"reports the refusal when the CRS already exists and only the conversion is new"* — where the export did not write the CRS at all; it was already there. A reader of `stats.warnings` in that case is told this export wrote something it did not.

Sharing one constant between the returned entry and the console line is the right call and I would keep it. The wording just has to hold for every path that reaches it, which is the cost of sharing. Something like

> Cannot create IfcMapConversion: no IfcGeometricRepresentationContext is available to reference as SourceCRS. The IfcProjectedCRS is unaffected.

says the useful part — georeferencing is not wholly absent, only the conversion is missing — without asserting which of the two produced the CRS.

Worth noting because it is a neat miniature of the pattern: a single message is *more* maintainable than two, and it also makes it easier to write one sentence that is only true on one branch. The constant did not cause the problem, but it is where the problem now lives.

Everything else on this PR stands: three routes each pinned by a named test, and the `stats.warnings` shape is right.

BIMvoice:
Fixed in 9c593e40.

Confirmed both call sites: `:612` (from-scratch branch, just pushed `IFCPROJECTEDCRS` at `:593` — "was written" is true there) and `:629` (`gm.mapConversion && !existingMcIds?.length && existingCrsIds?.length` — writes no CRS, references `existingCrsIds[0]`, so the old wording was false on this path).

Kept the shared constant, changed the wording to your suggestion: "The IfcProjectedCRS is unaffected."

On the test side: the existing test on the CRS-already-exists path only asserted `warnings[0]).toContain('IfcMapConversion')` — it could never have caught the message being false on that branch, since it never checked the "was written" clause. Added two assertions to that test (`not.toContain('was written')`, `toContain('IfcProjectedCRS is unaffected')`) so the wrong-branch wording is now pinned. Verified with a python3 exact-substring mutation back to the old string: the new assertions fail against it, confirming they exercise the fix (restored via file copy, not `git checkout --`).

`pnpm typecheck` (repo root) is clean. `packages/export` tests: 351 passed / 29 skipped both before and after (one pre-existing unrelated failure in `lod1-generator.test.ts`, a WASM geometry init issue, present on the unmodified tree too).

louistrue:
Verified `9c593e40`. The message now holds on every path that reaches it:

> Cannot create IfcMapConversion: no IfcGeometricRepresentationContext is available to reference as SourceCRS. **The IfcProjectedCRS is unaffected.**

"Unaffected" is the right word — it tells the caller the useful thing (georeferencing is not wholly absent, only the conversion is missing) without asserting which of the two paths produced the CRS. The previous "was written" was true when the export created it and false when it already existed.

Nothing else outstanding from me on this PR.

louistrue:
Confirmed, and it is the same silent drop this PR exists to close — one path further out.

`mapConversion` supplied **without** `projectedCRS`, on a model carrying neither an `IFCPROJECTEDCRS` nor an `IFCMAPCONVERSION`:

```
HAS MAPCONVERSION: false
HAS PROJECTEDCRS : false
WARNINGS         : []
```

The caller asked for a map conversion, got none, and `stats.warnings` is empty — which is exactly the state the PR's own summary says a caller can no longer be left in.

The reason it slips through is that the existing `reportMapConversionRefused` fires on the *no-context* condition, and this path never reaches that check: with no `projectedCRS` requested and none in the file, both creation branches are skipped earlier, so nothing is attempted and nothing is refused. It is a different refusal — "no CRS to attach it to" rather than "no context to reference" — and it needs its own message rather than reusing the shared constant, which is now correctly worded for the context case only.

That makes four paths total. Three are pinned by your tests; this is the fourth and it is the one where the omission is most complete — neither entity is written, so there is nothing in the output to hint that anything was requested.

Worth noting the shape, because it is the recurring one: the fix closes the paths it was written from and the test suite covers exactly those, so the gap is invisible from inside the change. The `stats.warnings` mechanism is right and already in place — this is one more `reportRefused` call plus a message, not a redesign.

louistrue:
Verified `9c593e40`. The refusal-message split for the existing-CRS path is right. The remaining thread is valid, and I reproduced it — one more path to cover.

## The silent drop greptile found is real

Request a `mapConversion` with no `projectedCRS` mutation, against a model that has no `IFCPROJECTEDCRS`:

```ts
georefMutations: {
  mapConversion: { eastings: 2600000, northings: 1200000, orthogonalHeight: 500 },
}
```

Probe output, against a fixture that **does** contain an `IFCGEOMETRICREPRESENTATIONCONTEXT`:

```
PROBE has IFCMAPCONVERSION: false
PROBE has IFCPROJECTEDCRS  : false
PROBE warnings             : []
```

Silently dropped. The context being present is the important part: this is **not** the refusal you already report. Your two `reportMapConversionRefused()` calls both sit behind `if (contextId) … else`, so they only fire when the context is missing. Here the blocker is different — there is no `IfcProjectedCRS` to point `TargetCRS` at, and none was requested:

```ts
if (gm.projectedCRS && !existingCrsIds?.length) {          // needs gm.projectedCRS — absent
} else if (gm.mapConversion && !existingMcIds?.length && existingCrsIds?.length) {  // needs existingCrsIds — empty
}
```

Neither arm is taken, so nothing runs at all, refusal reporting included.

Skipping the write is still correct — an `IfcMapConversion` with no `TargetCRS` would dangle. It is the silence that this PR exists to remove, and the title says so: "report a refused map conversion instead of dropping it silently."

## The docblock now asserts something false

```
 * A requested `georefMutations.mapConversion` is the one case today: with
 * no `IfcGeometricRepresentationContext` to reference as `SourceCRS`, the
 * `IfcMapConversion` is skipped ... while the `IfcProjectedCRS` is still written
```

There are two causes, not one, and they differ in what survives — on the missing-context path the CRS is still written, on this one nothing is. The shared `MAP_CONVERSION_WITHOUT_CONTEXT_WARNING` text would also be wrong here, since it names the context as the reason and promises "The IfcProjectedCRS is unaffected" when there is no CRS at all.

So this needs a second message rather than a third call to the existing one — same reasoning that made you split the message for the existing-CRS path. Something like: no `IfcProjectedCRS` exists and none was requested, so the conversion has no `TargetCRS`.

Worth a test with a context present, so it pins *this* cause and cannot pass for the old reason.

I ran this on your branch with the probe appended to `step-exporter.test.ts` and then reverted it; nothing is pushed.

BIMvoice:
@/tmp/comment_body.txt

louistrue:
Verified `5cd5417f`. The fix is right — separate message, separate reporter, own branch, and the changeset states the distinction cleanly ("no context to reference" vs. "no CRS to attach it to"). One thing left, and it is the test, not the code.

## The new test cannot tell the two refusals apart

It passes when the branch emits the **wrong** message:

```
### MUTATION: new branch emits the WRONG (context) message
      Tests  1 passed | 32 skipped (33)
```

I swapped `reportMapConversionRefusedNoCrs(warnings)` for `reportMapConversionRefused(warnings)` — the context-specific message, which is exactly what this change exists to stop being emitted here — and the test stayed green.

The reason is that both assertions are satisfied by both messages:

```
CONTEXT msg: contains 'IfcMapConversion'=True  contains 'IfcProjectedCRS'=True
NO-CRS msg : contains 'IfcMapConversion'=True  contains 'IfcProjectedCRS'=True
```

`MAP_CONVERSION_WITHOUT_CONTEXT_WARNING` ends with "The IfcProjectedCRS is unaffected", so `toContain('IfcProjectedCRS')` matches it too. The one substring that actually discriminates is `TargetCRS` — assert that, or assert equality against `MAP_CONVERSION_WITHOUT_CRS_WARNING`, and the mutation dies.

It does pin something real: deleting the branch fails it.

```
× reports the refusal when a map conversion is requested with no projectedCRS and none in the file
```

So it catches a regression to silence, which was the original defect. It just does not catch a regression to the *wrong explanation* — and since the entire increment here is "report this with its own wording", that is the part worth holding.

## The fixture leaves the cause ambiguous

It uses `NO_CONTEXT_ENTRIES`, so the model has no `IfcGeometricRepresentationContext` **and** no `IfcProjectedCRS`. Both refusal conditions are true at once, which is why a wrong message reads as correct.

Using a fixture that *has* a context isolates the cause — then the only thing that can produce a refusal is the missing CRS, and the context message becomes visibly wrong rather than coincidentally acceptable. That is the fixture I probed with earlier, and it reaches your new branch.

Two small changes: swap the fixture for a context-bearing one, and assert on `TargetCRS` (or full equality). Then this is done and I will merge it.

BIMvoice:
Both done in `9022d99d` — context-bearing fixture and a discriminating assertion.

## Your mutation now dies

I reproduced it first rather than patching on trust. Swapping `reportMapConversionRefusedNoCrs` for `reportMapConversionRefused` was green before; after the change it fails, and fails **only** that test:

```
MUTATION: new branch emits the context message   → 1 failed | 32 passed
MUTATION: drop the new report entirely           → 1 failed | 32 passed
```

So the test now catches a regression to the wrong *explanation* as well as a regression to silence, and the other three refusal tests are untouched by either.

## The fixture was the real problem

Your diagnosis is the part worth recording. `NO_CONTEXT_ENTRIES` makes **both** refusal conditions true simultaneously — no context *and* no CRS — so the context message was not merely tolerated by weak assertions, it was arguably true of that fixture. No assertion on that fixture could have separated the two.

Adding `#50=IFCGEOMETRICREPRESENTATIONCONTEXT($,'Model',3,1.E-05,$,$)` isolates the cause: with a usable 3D Model context present, `findPreferredGeometricRepresentationContextId` resolves, so the missing CRS is the only thing that can produce a refusal — and the context message becomes visibly false rather than coincidentally acceptable.

## The assertion

Both `TargetCRS` (the one discriminating substring, as you identified) **and** full equality. Equality alone would have done it, but `TargetCRS` names *why* the two differ, so a future edit that rewords the message has to think about whether the discriminator survives rather than just re-recording a string.

Kept `expect(result.stats.warnings).toHaveLength(1)` — worth holding, since the new branch sits in the same `else if` chain and a double-report would show up here.

`packages/export`: 355 passed / 29 skipped, `tsc --noEmit` clean.

Recording the general shape, because this is the third time it has come up today: the test was written from inside the change, against the fixture the change was developed on, so the one thing it could not see was the distinction the change exists to make. The fix closes the paths it was written from — and so does the test.

louistrue:
Verified `9022d99d`. Both changes landed and the test now discriminates. The mutation that previously passed:

```
### MUTATION: emit the WRONG (context) message
       × reports the refusal when a map conversion is requested with no projectedCRS and none in the file
      Tests  1 failed | 32 skipped (33)
```

Swapping in `reportMapConversionRefused` — the context-specific explanation — now fails, where before it was green. The full-equality assertion is what does it; `toContain('TargetCRS')` alone would have been enough, but equality also pins the wording against a silent drift in either message.

The context-bearing fixture is the more important of the two changes, and your comment says why better than I did:

> Under `NO_CONTEXT_ENTRIES` both refusal conditions are true at once, so the context-specific message reads as acceptable here and the test cannot tell the two refusals apart.

That is the general trap, not just this instance: a fixture that satisfies two hypotheses at once cannot distinguish them, and the test passes for whichever one the code happens to pick.

Resolving the remaining thread — greptile's finding is what this addressed, and the reachable case it described is now both fixed and pinned.

Merging once the last check lands.

louistrue:
Ran the CodeRabbit CLI on `9022d99d` (hosted was `Review rate limited` again). Two findings, **both valid** — and they are the mirror image of what you just fixed, on the SourceCRS side.

## One of the two context tests has the exact weakness you removed from the new one

Mutating both context branches to emit the *TargetCRS* message kills only one of them:

```
### MUTATION: the CONTEXT branches emit the TargetCRS message instead
       × reports the refusal when the CRS already exists and only the conversion is new
      Tests  1 failed | 4 passed | 28 skipped
```

The difference is entirely in the assertions:

```ts
// creating both the CRS and the conversion — survives the wrong message
expect(result.stats.warnings[0]).toContain('IfcMapConversion');

// CRS already exists — catches it
expect(result.stats.warnings[0]).toContain('IfcMapConversion');
expect(result.stats.warnings[0]).not.toContain('was written');
expect(result.stats.warnings[0]).toContain('IfcProjectedCRS is unaffected');
```

The second one discriminates because your earlier split gave it something specific to hold. The first is still on the bare `toContain('IfcMapConversion')` that both messages satisfy — the same shape that let the new test pass on the wrong explanation an hour ago. `SourceCRS` is the discriminating substring here, symmetric to `TargetCRS`, or full equality as you used in the new test.

Worth doing while the reasoning is fresh: this PR now has three refusal tests and two different assertion strengths, which is the state that lets one quietly rot.

## The changeset contradicts itself

The "Which sessions this affects" paragraph still says:

> only those that request a map conversion against a model with no usable `IfcGeometricRepresentationContext`

while the paragraph you appended below it describes the second condition. As written the summary excludes the case the PR now also covers — a reader taking the first paragraph as the scope would conclude the TargetCRS refusal does not exist. One sentence.

Everything else is verified and the gate is clean: 16 check-runs with 5 core lanes present (not merely not-red — I am checking that explicitly now after nearly merging a PR whose CI had silently never run), `MERGEABLE`, no failures, no red statuses, 0 unresolved threads. The wrong-message mutation on the new test dies. Fix these two and I will merge.

BIMvoice:
Both findings addressed in e8f4d65c.

**Finding 1** — the "creating both the CRS and the conversion" test only asserted `toContain('IfcMapConversion')`, which both refusal messages satisfy. Strengthened it to full equality against the SourceCRS message (same strength the "CRS already exists" test already had after that assertion is also promoted to `toBe` for consistency, and the pre-existing no-CRS test's `toBe`). All three refusal tests in the `map conversion that cannot be written (#2067)` describe block now assert full equality on the message.

Mutation-tested with your exact repro (both context branches call `reportMapConversionRefusedNoCrs` instead of `reportMapConversionRefused`):

| Mutation | Before this fix | After this fix |
|---|---|---|
| Context branches emit TargetCRS message | 1 failed (only "CRS already exists" test) | 2 failed (both context tests) |
| Drop the no-CRS report (`reportMapConversionRefusedNoCrs` call removed) | 1 failed (no-CRS test only) — unchanged | 1 failed (no-CRS test only) |
| Drop the context report (both `reportMapConversionRefused` calls removed) | 3 failed (both context-message tests + delta-only test) — unchanged | 3 failed (same three) |

Mutations applied via python3 exact-substring replacement (assert count==1 or 2 as applicable, printed to stderr), restored by file copy — never `git checkout --`. Regression mutations (rows 2–3) behave the same before and after, confirming the fix only strengthened the target test without touching coverage elsewhere.

`packages/export` test suite: 355 passed / 29 skipped / 384 total, both before and after (assertion strengthening only, no new tests).

**Finding 2** — `.changeset/georef-map-conversion-refusal.md`'s "Which sessions this affects" paragraph now names both refusal conditions (no context for `SourceCRS`, no `IfcProjectedCRS` for `TargetCRS`) in one sentence instead of describing only the first.

`pnpm typecheck` clean across the workspace.

louistrue:
Both round-two items are done and verified — this is merge-ready on content, but it now needs a rebase.

All three refusal tests discriminate:

```
### MUTATION A: context branches emit the TargetCRS message
       × reports the refusal when creating both the CRS and the conversion
       × reports the refusal when the CRS already exists and only the conversion is new

### MUTATION B: no-CRS branch emits the context message
       × reports the refusal when a map conversion is requested with no projectedCRS and none in the file
```

Each direction now kills the tests belonging to the other explanation. Before this round, mutation A killed only one of the two.

The changeset scope line reads correctly too — both conditions named, `SourceCRS` and `TargetCRS` distinguished, and "Ordinary georeferencing edits against a model with **both**" is the right correction to the earlier wording.

**Needs `git merge origin/main`.** I merged #2059 a few minutes ago and it touches `step-exporter.ts`, so this went `CONFLICTING`. Two things follow from that, one of which bit me earlier today:

1. Once conflicting, GitHub stops scheduling `pull_request` runs entirely — so if you push anything before resolving, the PR page will show a *green* rollup with every real lane simply absent. The `core=5, fail=0` I measured above is genuine, but it describes `e8f4d65c` as it stood before the conflict appeared.
2. #2059's changes are in the same georeferencing region, so please look at the resolution rather than taking either side wholesale — that is how a fix silently disappears.

Rebase and push; I will re-verify against the merged head and merge.

BIMvoice:
Merged `upstream/main` in `70c285b4`. Both mutations re-run **after** the merge, per your point that the pre-merge measurement describes a different head.

## The conflict was the both-sides kind

One conflict in `step-exporter.ts`, and it wanted neither side wholesale — exactly what you flagged:

- **ours**: the two reporter methods (`reportMapConversionRefused` / `reportMapConversionRefusedNoCrs`)
- **#2059's**: `findPreferredGeometricRepresentationContextId` gaining an `effective: EffectiveEntityIndex` parameter and filtering the context lookup through it

They collided only because the new methods sit immediately above that signature. Resolved to both: our two reporters kept, #2059's signature and its doc kept. Verified afterwards by counting rather than reading — all three call sites present (`:631`, `:648` context; `:656` no-CRS), both `findPreferred...(effective)` call sites passing the new argument, `tsc --noEmit` clean.

Taking "ours" here would have silently reverted #2059's tombstone filtering while every refusal test stayed green, since none of them exercise a deleted context.

## Re-verified post-merge

| mutation | result |
|---|---|
| A: context branches emit the TargetCRS message | 2 failed — both context tests |
| B: no-CRS branch emits the context message | 1 failed — only the no-CRS test |

Each direction still kills the tests belonging to the other explanation, so the merge did not drain them. `packages/export`: 358 passed / 29 skipped (up from 355 — #2059's tests arrived with the merge).

## On the CI point

Noted, and it is the third time today one of mine has been in that state. The pattern I have taken from it: a green rollup means nothing until the core lanes are confirmed *present*, and `CONFLICTING` is the usual reason they are not — which is invisible on the PR page because absence renders identically to success.

louistrue:
Rebased head is green — 5 core lanes, no failures, `MERGEABLE`. Merging.

Content was already verified before the rebase: all three refusal tests discriminate, with each direction killing the tests belonging to the other explanation.

```
MUTATION A (context branches emit the TargetCRS message)
  × reports the refusal when creating both the CRS and the conversion
  × reports the refusal when the CRS already exists and only the conversion is new
MUTATION B (no-CRS branch emits the context message)
  × reports the refusal when a map conversion is requested with no projectedCRS and none in the file
```

Thanks for taking the fixture point seriously rather than just adding an assertion — a context-bearing fixture is what makes the TargetCRS test unable to pass for the SourceCRS reason, and that is the part a reviewer six months from now will not re-derive.

PR #2106, opened 2026-08-04
fix(viewer): report a refused clash-settings write instead of reverting silently

BIMvoice:
Follow-up to #2101, and the resolution of a contradiction between it and the in-flight flavor-config branch. Both were partly wrong.

## The bug

`persistSettings` discarded the `SaveResult` from `saveSettings`. A refused write (quota, or storage blocked by site settings) left the value committed in the store and absent from disk, so **tolerance, mode, clearance, cluster radius, grazing and grouping silently revert on the next reload**. #2101's read — *"settings-only and commit-then-persist by design"* — describes the mechanism accurately but does not follow it through to the reload.

## Why not the #2101 gate

The sibling branch argued for a different fix on the grounds that gating `set` would *"freeze the slider and fire one report per drag tick."* That reasoning is right in substance and **wrong about the widget** — there is no range input anywhere in the Detection tab. The three numerics are `NumberField`, an `<input type="number">` whose `value` comes from the store and whose `onChange` calls the setter directly, with no debounce and no commit-on-blur. So it is one write **per keystroke**, not per drag tick: typing `0.005` fires ~5.

That makes the conclusion stronger, not weaker. Because `value={value}` is fed from store state, gating `set` on the result would render the field **literally untypable** under blocked storage. #2101's gate is right where it is — the preset controls are a list, not a text field — and wrong here.

## Remedy

Commit as before; report the refusal **once per store instance**.

- `clashSlice.ts` — a `settingsSaveFailureReported` latch in the `createClashSlice` closure, so "once per session" means once per store instance and tests get isolation with no module-global reset hatch.
- `lib/clash/settings-save-notice.ts` (new) — dependency-free reporter indirection, default `console.warn`. The slice must not import the toast module: that would pull React into every store test. `setClashSettingsSaveReporter` returns an unregister that only restores the fallback if it is still the active reporter (out-of-order unmount safety).
- `ClashSettingsDialog.tsx` — points the reporter at `toast.error` while mounted, which is exactly when the Detection tab is reachable. This is the same surface #2101 uses for presets, so the two are coherent: per-action there, one-shot here.

The latch stays set for the session even if storage later recovers — deliberate, and documented.

## RED / GREEN

10 new tests. RED by reverting **only** `persistSettings` to the pre-fix one-liner, keeping the new module so the failure is behavioural rather than an import error: **6 pass / 4 fail**. Restored by file copy: 10/10.

Mutations (python-scoped, replacement count printed and asserted `==1`, region re-read, restored by file copy — never `git checkout`):

| Mutation | Count | Result |
|---|---|---|
| drop `\|\| settingsSaveFailureReported` (latch neutralised) | 1 | 9/1 — the one-shot test caught it |
| delete the report call (notice neutralised) | 1 | 6/4 |
| roll `setClashTolerance` back on refusal (the rejected gate) | 1 | 7/3 |

The third is the one that matters: it pins the property the remedy must **not** break, and it passes both before and after the fix.

**Anti-vacuity:** the storage double is selective — `failKey` makes only `ifc-lite-clash-settings` throw while `ifc-lite-clash-presets` still round-trips, reads keep working while writes are refused, and it implements `length`/`key()`. The test asserts all of that, and asserts a seeded `clashTolerance === 0.037` that no default produces, proving the slice loaded through the real persistence path rather than a storage-less fallback.

## Validation

Rebased onto #2101 (`531a9559`). The predicted clean auto-merge **did not quite hold** — one import-line conflict in the dialog, where #2101 added `type SaveResult` to the same line; both kept. Post-rebase: **2608 pass, 0 fail, 2 skipped**; `tsc --noEmit` clean. No changeset (`@ifc-lite/viewer` is private).

## Out of scope, but worth knowing

`showClashRegionBox` sits in the Detection tab but is not part of `ClashGlobalSettings` and is never persisted — it resets on every reload regardless of storage health. A separate design gap, left alone.

Also: the flavor-config branch calls `saveSettings` **directly** rather than via `persistSettings`, so this latch does not cover it and it returns its own `SaveResult` to the extension host. Complementary, not duplicative — but if both land, a storage failure during a flavor apply produces the host's report rather than the one-shot. Not a conflict, just a coherence question for you.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an error notification when clash settings cannot be saved.
  * Settings changes now remain applied in the current session even if storage persistence fails.
  * Repeated save failures are limited to one notification per session to avoid noise.
  * Successful saves continue to persist normally.

### DIFF
--- a/apps/viewer/src/components/viewer/ClashSettingsDialog.tsx
+++ b/apps/viewer/src/components/viewer/ClashSettingsDialog.tsx
@@ -14,7 +14,7 @@
  *    localStorage; shareable via export / import.
  */
 
-import { useCallback, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import {
   Settings2, Plus, Pencil, Trash2, RotateCcw, Upload, Download, Check, X,
 } from 'lucide-react';
@@ -33,6 +33,7 @@ import { toast } from '@/components/ui/toast';
 import { useViewerStore } from '@/store';
 import { matchesSelector, type ClashSeverity } from '@ifc-lite/clash';
 import { exportPresets, importPresets, type ClashPreset, type SaveResult } from '@/lib/clash/persistence';
+import { setClashSettingsSaveReporter } from '@/lib/clash/settings-save-notice';
 
 const SEVERITY: Record<ClashSeverity, { label: string; color: string }> = {
   critical: { label: 'Critical', color: '#f7768e' },
@@ -82,6 +83,14 @@ export function ClashSettingsDialog({ trigger }: ClashSettingsDialogProps) {
 
   const [draft, setDraft] = useState<Draft | null>(null);
   const fileRef = useRef<HTMLInputElement>(null);
+
+  // Detection settings commit optimistically and persist in the background, so
+  // a refused write (quota, or storage blocked) is only visible if something
+  // says so. This component is mounted with the clash panel header — i.e.
+  // whenever the user can reach the Detection tab — so give the slice's
+  // once-per-session notice a toast to land in; it falls back to console.warn
+  // when nothing is mounted. The effect returns the unregister.
+  useEffect(() => setClashSettingsSaveReporter(toast.error), []);
 
   const matchCount = useCallback(
     (selector: string): number | null => {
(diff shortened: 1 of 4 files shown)


louistrue:
Viewer tier: green on all 5 core lanes, `MERGEABLE`, 0 unresolved threads. Merging.

Same shape as the failure-reported-as-success work already landed today (#2089, #2091, #2100) — a refused write now surfaces instead of silently reverting, with the notice extracted into `settings-save-notice.ts` so it is testable away from the dialog. Reviewed at the viewer bar rather than mutation-tested, since this is display-and-notification behaviour with no data at risk.

PR #2108, opened 2026-08-04
fix(ci): stop excluding ifc-lite-wasm from clippy and Rust tests

### LINKED ISSUE #2049: CI excludes ifc-lite-wasm entirely: its unit tests and clippy have never run, and two lint violations are on main now

BIMvoice:
## Summary

Closes #2049.

`.github/workflows/test.yml` excluded `ifc-lite-wasm` entirely from both the
`Clippy (lint gate)` and `Rust tests` steps, justified as avoiding
wasm-bindgen-test dev-deps that "target wasm32". That justification only
applies to one file in the crate: `tests/mesh_determinism.rs`, which already
gates its whole contents with `#![cfg(target_arch = "wasm32")]` — on a native
target it compiles to an empty test binary (`running 0 tests ... ok`). It
neither fails to build nor fails at runtime natively, so no exclusion was
needed to keep it out of the native gate.

The rest of the crate is 27 plain native `#[test]`s plus the crate's own
clippy lint, both of which were never run by the required PR gate. That gap
is how two real clippy violations (`bool2d.rs:70`, `mesh.rs:342`) sat on
`main` invisibly until #2051 fixed them directly (already merged, prior to
this PR).

Per the issue thread investigation, the fix is to just drop
`--exclude ifc-lite-wasm` from both `cargo clippy` and `cargo test` lines —
no separate `-p ifc-lite-wasm` step, no feature gating needed. The wasm32
leg (`wasm-pack test --node ... --test mesh_determinism`) keeps running on
its existing schedule/dispatch-only lane in `determinism.yml`, untouched.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` on current
  `upstream/main` (post-#2051): clean, including `ifc-lite-wasm`.
- `cargo test -p ifc-lite-wasm`: 27 unit tests pass in ~0.01s;
  `tests/mesh_determinism.rs` runs 0 tests natively, confirming it
  self-gates and does not need workspace exclusion.
- Proved the newly-enabled gates actually catch regressions by injecting
  (then reverting via file copy, never `git checkout --`) a deliberate
  violation in each:
  - a clippy `if_same_then_else` violation in `rust/wasm-bindings/src/lib.rs`
    made `cargo clippy -p ifc-lite-wasm --all-targets -- -D warnings` fail
    with `error: this if has identical blocks`.
  - a deliberate failing `#[test]` in the same file made
    `cargo test -p ifc-lite-wasm --lib` fail with
    `test result: FAILED. 27 passed; 1 failed`.
- `cargo test --workspace` also shows 34 pre-existing failures in
  `ifc-lite-export`, all missing fixtures (`tests/models/ara3d/duplex.ifc`
  and friends) — reproduced identically on unmodified `upstream/main` via
  `git stash`, confirming they are unrelated to this change (same trap
  noted in the issue thread).

Not verified locally (workflow-only, CI-only): actual GitHub Actions runner
behavior, `Swatinem/rust-cache` caching effect on wall-clock, and whether
the wasm32 target's presence in `rust-toolchain.toml` has any interaction
with the runner beyond what was checked here.

## Test plan

- [ ] CI: `Clippy (lint gate)` step passes with `ifc-lite-wasm` included
- [ ] CI: `Rust tests` step passes with `ifc-lite-wasm`'s 27 unit tests included
- [ ] CI wall-clock for the `rust-tests` job does not regress meaningfully (cached after first run)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -464,12 +464,15 @@ jobs:
       # notices if that ever stops being true.
       - name: Cargo.lock is in sync with the manifests
         run: cargo metadata --locked --format-version 1 > /dev/null
-      # Workspace exclusions:
-      #   - ifc-lite-wasm: dev-deps on wasm-bindgen-test; tests target wasm32.
+      # ifc-lite-wasm is included: its only wasm32-only test file,
+      # tests/mesh_determinism.rs, self-gates with `#![cfg(target_arch =
+      # "wasm32")]` and compiles to an empty test binary on the native host
+      # (`running 0 tests ... ok`), so no workspace-wide exclusion is needed.
+      # That wasm32 leg runs on its own schedule in determinism.yml.
       - name: Clippy (lint gate)
-        run: cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings
+        run: cargo clippy --workspace --all-targets -- -D warnings
       - name: Rust tests
-        run: cargo test --workspace --exclude ifc-lite-wasm
+        run: cargo test --workspace
 
   # Geometry watertightness / triangulation-invariance census.
   #


louistrue:
This is the most valuable PR in your queue right now, and I verified it at the Rust bar rather than on green CI.

**It closes a real hole.** I injected a `clippy::ptr_arg` violation into `ifc-lite-wasm/src/lib.rs` and ran both commands:

| command | result |
|---|---|
| this PR — `cargo clippy --workspace --all-targets -- -D warnings` | **exit 101, 3 errors — caught** |
| main — `... --exclude ifc-lite-wasm ...` | **exit 0, 0 errors — missed** |

So the exclusion was not a harmless scoping detail; it was a blind spot covering an entire crate of the WASM boundary. Anything landing in `rust/wasm-bindings` has been unlinted and untested by the gate until now.

Your claim about the wasm32 test file checks out too — `tests/mesh_determinism.rs` line 20 is `#![cfg(target_arch = "wasm32")]`, so it compiles to an empty binary on the native host and needs no exclusion. Clippy passes clean workspace-wide with `ifc-lite-wasm` in the checked list.

One note for anyone reproducing locally: `cargo test --workspace` fails here on `ifc-lite-export` and `wall_opening_cut_regression`, but that is a **fresh-worktree artifact** — `tests/models/ara3d/duplex.ifc` is not present, so the fixture-backed tests panic on a missing file. Same failures on main, unrelated to this change. CI has the fixtures.

Merging. Given the priority is keeping the Rust side clean while the viewer can afford to break, a change that makes the Rust gate actually cover the WASM crate is worth more than most feature work in the queue.

PR #2111, opened 2026-08-04
fix(export): honour overlay deletions in ParquetExporter

### LINKED ISSUE #2046: IFC5/IFCX and Parquet export do not honour overlay-deleted entities

BIMvoice:
## Summary

Closes #2046 — the `ParquetExporter` half. The `Ifc5Exporter` half was already fixed on `main` by #2047/#2070 (verified: `ifc5-exporter.ts` imports `getEffectiveEntityIndex`/`isDeleted` and gates the node-collection loop on it).

`ParquetExporter`'s five table writers (`Entities`, `Properties`, `Quantities`, `Relationships`, `SpatialHierarchy`) column-copy whole typed-array columns straight out of `IfcDataStore` — no per-entity loop, and until this PR, no `MutablePropertyView` parameter at all. An entity deleted via `MutablePropertyView.deleteEntity()` was exported anyway, along with its properties, quantities, and relationship edges.

## Fix

- `ParquetExporter` takes an **optional** third constructor argument, `mutationView`. Existing `new ParquetExporter(store)` callers (the README example, `tests/integration.test.ts`) are unaffected.
- When supplied, builds an `EffectiveEntityIndex` (same helper `StepExporter`/`Ifc5Exporter` use, `effective-index.ts`) and gates row emission in all five writers on `isDeleted(id)`.
- Deletion-only scope, called out explicitly in code comments, README, and the changeset:
  - pset/quantity/attribute *edits* still export the source values verbatim — the writers here are columnar bulk-copies with no per-entity emission pass to apply them (unlike `StepExporter`/`Ifc5Exporter`).
  - `SpatialHierarchy` is a source-parse snapshot with no overlay-aware re-parenting, so a deleted storey/building/site can still surface as `StoreyId`/`BuildingId`/`SiteId` on a surviving element's row — the class of problem `Ifc5Exporter`'s re-parenting pass solved for #2047, not addressed here.

## Call sites

No shipped surface (viewer, CLI, MCP) constructs `ParquetExporter` today — confirmed by grep across the tree. This PR makes the exporter capable of honouring deletions; a future caller can now pass its `MutablePropertyView` and get correct output from the start, without any consumer change needed here.

## Test plan

- [x] RED: added `packages/export/src/parquet-exporter.test.ts` before the fix — 3 of 4 tests failed against the unmodified exporter (deleted entity/property/relationship rows all present).
- [x] GREEN: same tests pass after the fix; a 4th test pins the no-overlay back-compat path.
- [x] Mutation-tested each of the three gates (`writeEntities`, `writeProperties`, `writeRelationships`) individually via exact-substring replacement (count asserted == 1 each), confirmed each mutation fails the corresponding test, restored by file copy.
- [x] `pnpm --filter @ifc-lite/export test` (vitest): 354 passed / 29 skipped (383), up from 350/29 (379) baseline — only the 4 new tests added, no regressions.
- [x] `turbo build` / `turbo typecheck` (full monorepo): clean.
- [x] `pnpm check:api-surface`: clean, no snapshot update needed.
- [x] Changeset added (`@ifc-lite/export`, patch).


louistrue:
Reviewed at the published-package bar, since `packages/export` is where a wrong answer is silent and durable. **greptile's three findings deserve different treatments, and the first one I would take.**

## Take: the memoised overlay index has no invalidation

```ts
private getEffective(): EffectiveEntityIndex | null {
    if (!this.mutationView) return null;
    if (!this._effective) {
        this._effective = getEffectiveEntityIndex(this.store, this.mutationView, true);
    }
    return this._effective;
}
```

Computed once per instance, never invalidated. Export, mutate the supplied `MutablePropertyView`, export again on the same instance, and the second export honours the *first* export's deletions.

The usual defence — "no caller does that" — **does not apply here**, and this is worth being precise about. `ParquetExporter` has **zero in-repo callers**:

```
packages/export/src/parquet-exporter.ts:21:export class ParquetExporter {
packages/export/src/index.ts:11:export { ParquetExporter, ... }
```

It is instantiated nowhere in `apps/` or `packages/`. Its only consumers are external, which means external usage patterns *are* the contract, and "construct once, export, edit, export again" is an ordinary thing for a consumer to do. Contrast a private helper whose call sites you control — there, unreachability is a real argument. Here it is the opposite: nothing in-repo constrains how it is used.

Cheapest fix is to derive rather than cache — or invalidate on a mutation-version counter if the index is expensive enough to be worth memoising at all. Worth measuring before assuming it is.

## Your judgement: the two dangling-reference findings

`Deleted relationship IDs remain exported` and `Deleted spaces leave dangling references` are both about referential integrity in the emitted Parquet after a partial deletion. They are plausible and in scope for a PR titled "honour overlay deletions", but the right answer depends on a decision this PR has not stated: **is the exported Parquet expected to be referentially closed?**

- If yes, deletions have to cascade or be refused, and both findings are bugs.
- If no — consumers are expected to tolerate references to absent ids, as a relational export often does — then both are working as intended and deserve a sentence in the changeset saying so.

I would rather you state the intended contract than have me guess it. Whichever way it goes, saying it explicitly is what stops the next reviewer re-raising this.

## Not blocking on

Nothing else — CI is green on all 5 core lanes and the changeset is present. This is purely about the three threads above.

BIMvoice:
Both taken in `e6e107fb`.

## The memoised index

Reproduced before fixing: a test that exports, deletes through the same view the exporter holds, and exports again fails on the old code — the second export returns `Wall2` as still present.

Fixed by **deriving per call rather than invalidating**. A version counter is another thing that can be wrong, and the index is cheap relative to writing the tables; I would rather not memoise than memoise-and-invalidate here. Mutation: reintroducing the cache (`this._eff ??= ...`) fails exactly the new test, nothing else. `packages/export` 355 passed / 29 skipped, `tsc --noEmit` clean.

Your argument for why this matters is the part worth keeping. I would normally reach for "no caller does that" — and you pointed out it is **inverted** here: zero in-repo callers means there is no call site we control that could make staleness unreachable, so external usage *is* the contract. That is the opposite of a private helper, where unreachability is a real defence. I had the reasoning backwards when I let the memoisation through.

## The referential-closure contract

Stated explicitly in the changeset, as you asked, rather than left for the next reviewer to re-raise:

> **The exported Parquet is deliberately NOT referentially closed.** A deletion drops the rows *identified by* or *keyed to* the deleted entity, but surviving rows may still carry ids that no longer resolve — most visibly a surviving element's `StoreyId`/`BuildingId`/`SiteId` pointing at a deleted storey.

The reasoning: the tables are a queryable projection, not a self-contained graph, and `exportTable('entities')` alone has never been closed either — a consumer joining them must already tolerate unresolved ids. Cascading deletions instead would silently remove elements the caller never deleted, which is a worse answer for an export format than a dangling id. Callers needing a closed graph should use `StepExporter`/`Ifc5Exporter`, which is now said in the changeset too.

So both of greptile's dangling-reference findings are **working as intended** under that contract — not dismissed, but scoped, with the scope written down.

louistrue:
Both halves addressed, and the mutation confirms the first one:

```
### MUTATION: restore the memoisation
     × reflects deletions made between two exports on the same instance
      Tests  1 failed | 4 passed (5)
```

The test drives the real reuse pattern — construct, export, `view.deleteEntity(2)` through the *same* view the exporter holds, export again — rather than poking at the private field, so it fails for the reason it claims.

**On the referential-closure contract:** stating it explicitly is the right resolution, and I think the answer you picked is the right one. The key line is the one that makes it a decision rather than an omission:

> Cascading deletions instead would silently remove elements the caller never deleted, which is a worse answer for an export format.

That is the argument that settles it. A relational projection where a surviving element's `StoreyId` points at a deleted storey is a documented property a consumer can code against; an export that quietly removes rows the caller never asked to remove is data loss wearing a tidy-looking mask. Pointing closed-graph consumers at `StepExporter`/`Ifc5Exporter` gives them somewhere to go, which is what turns this from a limitation into a contract.

Resolving all three threads: the memo one is fixed, and the two dangling-reference ones are answered by the stated contract rather than left implicit.

Merging once CI finishes — 2 lanes still running.

Issue #2097, opened 2026-08-04
LAZ wasm loading path has no end-to-end coverage: mocked tests stay green while every browser LAZ open fails


BIMvoice:
## Summary

Nothing exercises the laz-perf wasm loading path end to end. Every test that touches it substitutes the module under test, so the two mechanisms that actually make LAZ work in a browser — the Vite `?url` asset import and the emscripten `Module.wasmBinary` hand-off — are unasserted. If either breaks, the suite stays green and **every LAZ open fails in the browser**.

Raised in review on #2086. Filing it separately because it is **pre-existing on `main`, not introduced by that PR** — see below.

## This is pre-existing, not from #2086

Both mechanisms predate #2086. At `e18356f3` (the commit *before* #2086 merged), `packages/pointcloud/src/streaming/laz-source.ts` already had:

- **`:58-61`** — the `?url` pre-fetch:
  ```
  // Pre-fetch the wasm via Vite's `?url` asset pipeline (hashed,
  // served with `application/wasm`) and hand the bytes to emscripten
  // as `Module.wasmBinary` so its own fetch is skipped entirely.
  const wasmBinary = await fetchLazPerfWasm();
  ```
- **`:88`** — the hand-off: `return factory({ wasmBinary });`

#2086 (`9d9c8049`) changed the *memoization* of the loader, not the loading path; those lines survive verbatim at `:57-60` and `:87` on current `main`. They appear as both `+` and `-` in that diff only because the surrounding function was relocated during the extraction of `memoizeAsync`.

## Failure mode

`fetchLazPerfWasm()` (`laz-source.ts:90-111`) is the whole risk surface:

1. `await import('laz-perf/lib/web/laz-perf.wasm?url')` — if the specifier stops resolving, or the bundler stops treating it as a static asset, the URL is wrong or absent.
2. `factory({ wasmBinary })` — if emscripten stops honouring `wasmBinary`, the shim falls back to its own `locateFile` fetch, which is the original bug this code exists to avoid (`fetch("laz-perf.wasm")` → 404 → SPA index HTML served as `text/plain` → the "MIME type 'text/plain'" failure).

Neither is reachable from any current test. `laz-source.test.ts` drives `LazStreamingSource.open()` through `setLazPerfLoaderForTesting()` with a stub emscripten module — correct for what it pins (the retry/memo semantics), but it replaces `importLazPerf` wholesale, so the `?url` import and the `wasmBinary` hand-off never execute.

[...]

PR #2113, opened 2026-08-04
test(pointcloud,viewer): real-browser E2E for the laz-perf wasm loading path (#2097)

BIMvoice:
## Summary

Closes the gap #2097 describes: nothing exercised the laz-perf wasm loading path end to end, so every existing test stayed green while a broken `?url` resolution or a broken `wasmBinary` hand-off would fail every LAZ open in a real browser.

- `packages/pointcloud/src/streaming/laz-source.ts` gains `probeLazPerfWasmLoad()`, a test-only export that calls the real `importLazPerf()` directly (bypassing `setLazPerfLoaderForTesting`) and reports whether a real `LASZip` constructor came back.
- `apps/viewer/laz-probe.html` + `apps/viewer/src/e2e/laz-wasm-probe-entry.ts`: a small E2E-only page, not linked from the app UI, added to the real `vite build` input list so a production build actually bundles and runs the probe through Vite's real `?url` asset pipeline.
- `tests/e2e/laz-wasm.e2e.spec.ts`: Playwright test that loads that page against a real `vite preview` (production build, real MIME types) and asserts `probeLazPerfWasmLoad()` succeeded.

## What IS covered

Both mechanisms the issue names as unasserted:
1. `import('laz-perf/lib/web/laz-perf.wasm?url')` resolving to a real, hashed, `application/wasm`-served asset in the **production build** (not the vitest dev-server `/@fs/...` form, which is a different pipeline — see the issue thread for why vitest was ruled out).
2. `factory({ wasmBinary })` — the hand-off succeeding rather than falling back to emscripten's own broken `locateFile` fetch (the original bug this code exists to avoid).

Getting back a working `LASZip` constructor from a real emscripten instantiation proves both, independent of decoding any actual point data.

**Mutation-tested**: pointing the `?url` specifier at a nonexistent module fails the `vite build` outright. Corrupting the fetched wasm URL post-resolution (so `fetch` 404s and the wasm loader receives the SPA `index.html` fallback instead) builds clean but fails this test at runtime with `WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f` — the same failure class as the original "MIME type 'text/plain'" bug. Restored via file copy afterward.

## What is NOT covered

Decoding actual LAZ-compressed point data (`laszip.open()` on real bytes, points coming out correctly). There is no `.laz` fixture in `tests/models/manifest.json`, and one can't be synthesized in-repo: `laz-perf@0.0.7` is decode-only (`LASZip`/`ChunkDecoder`, no writer/compressor). That gap is unchanged by this PR — see the issue thread for the full investigation (vitest and the wasm-contract script were both ruled out as homes for it, and why).

## Test plan

- [x] `pnpm --filter @ifc-lite/pointcloud test` — 131 passed, 2 skipped
- [x] `pnpm --filter @ifc-lite/viewer typecheck`, `pnpm --filter @ifc-lite/viewer-embed typecheck`
- [x] `node scripts/check-api-surface.mjs` (snapshot updated, changeset added — `probeLazPerfWasmLoad` is a new minor export on `@ifc-lite/pointcloud`)
- [x] `node scripts/check-test-wiring.mjs`
- [x] Manual run of `tests/e2e/laz-wasm.e2e.spec.ts` against a real `vite build` + `vite preview`, in a real Chromium: passes on the current tree, fails with the SPA-fallback wasm-magic-word error when the `?url`→fetch wiring is deliberately broken.

I could not run this through the repo's actual `viewer-e2e-ci` Playwright project locally (it pins `channel: 'chrome'`, and this sandbox has no real Chrome installed, only Playwright's bundled Chromium — which reproduced the same pass/fail behavior). CI has real Chrome, so this should run there unmodified; flagging in case anything about the `chrome` channel specifically needs a second look.

### DIFF
--- a/.changeset/laz-wasm-e2e-probe.md
+++ b/.changeset/laz-wasm-e2e-probe.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/pointcloud": minor
+---
+
+Add `probeLazPerfWasmLoad()`, an internal E2E test hook that exercises the real laz-perf wasm loader (the Vite `?url` asset fetch plus the `Module.wasmBinary` hand-off to emscripten) without needing a `.laz` fixture. No other test drove this path for real: every existing test substitutes the loader via `setLazPerfLoaderForTesting()`, so a broken `?url` resolution or a broken `wasmBinary` hand-off stayed invisible until a real browser tried to open a LAZ file (#2097). Used by `apps/viewer`'s `laz-probe.html` (E2E-only, not linked from the app UI) and asserted by `tests/e2e/laz-wasm.e2e.spec.ts` against a real production build.
--- a/apps/viewer/laz-probe.html
+++ b/apps/viewer/laz-probe.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>laz-perf wasm probe (E2E only, see #2097)</title>
+</head>
+<body>
+  <!--
+    E2E-only page, not linked from the app UI. Loaded by
+    tests/e2e/laz-wasm.e2e.spec.ts against a real `vite build` to prove the
+    laz-perf wasm asset pipeline works end to end (issue #2097).
+  -->
+  <script type="module" src="/src/e2e/laz-wasm-probe-entry.ts"></script>
+</body>
+</html>
--- a/apps/viewer/src/e2e/laz-wasm-probe-entry.ts
+++ b/apps/viewer/src/e2e/laz-wasm-probe-entry.ts
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Entry for `laz-probe.html`, an E2E-only page read by
+ * `tests/e2e/laz-wasm.e2e.spec.ts` (#2097).
+ *
+ * Every other test that touches `LazStreamingSource` substitutes the wasm
+ * loader (`setLazPerfLoaderForTesting`), so the two mechanisms that make LAZ
+ * work in a real browser — the Vite `?url` wasm-asset fetch and the
+ * `Module.wasmBinary` hand-off to emscripten — are otherwise never
+ * exercised against a real production build. This page is not linked from
+ * the app UI; it exists solely so `pnpm --filter @ifc-lite/viewer build`
+ * bundles `probeLazPerfWasmLoad()` through the real asset pipeline, and a
+ * Playwright test can read the result off `window.__LAZ_PROBE__`.
+ */
+import { probeLazPerfWasmLoad } from '@ifc-lite/pointcloud';
+
+declare global {
+  interface Window {
+    __LAZ_PROBE__?: Awaited<ReturnType<typeof probeLazPerfWasmLoad>>;
+  }
+}
+
+void probeLazPerfWasmLoad().then((result) => {
+  window.__LAZ_PROBE__ = result;
+});
--- a/apps/viewer/vite.config.ts
+++ b/apps/viewer/vite.config.ts
@@ -328,6 +328,13 @@ export default defineConfig({
       // longer ships a desktop app; downstream desktop builders supply
       // @tauri-apps in their own host layer.
       external: ['@tauri-apps/api/event'],
+      input: {
+        main: path.resolve(__dirname, 'index.html'),
+        // E2E-only, not linked from the app UI: reads the laz-perf wasm
+        // asset pipeline through the real build so
+        // tests/e2e/laz-wasm.e2e.spec.ts can assert it end to end (#2097).
+        lazProbe: path.resolve(__dirname, 'laz-probe.html'),
+      },
       output: {
         manualChunks(id) {
           if (id.includes('/packages/sandbox/')) return 'sandbox';
--- a/packages/pointcloud/src/index.ts
+++ b/packages/pointcloud/src/index.ts
@@ -43,7 +43,7 @@ export type {
   DownsampleHint,
 } from './streaming/types.js';
 export { LasStreamingSource } from './streaming/las-source.js';
-export { LazStreamingSource } from './streaming/laz-source.js';
+export { LazStreamingSource, probeLazPerfWasmLoad } from './streaming/laz-source.js';
 export { PlyStreamingSource } from './streaming/ply-source.js';
 export { PcdStreamingSource } from './streaming/pcd-source.js';
 export { E57StreamingSource } from './streaming/e57-source.js';
--- a/packages/pointcloud/src/streaming/laz-source.ts
+++ b/packages/pointcloud/src/streaming/laz-source.ts
@@ -119,6 +119,26 @@ async function fetchLazPerfWasm(): Promise<Uint8Array> {
 let loadLazPerf = memoizeAsync(importLazPerf);
 
 /**
+ * @internal E2E probe only — `tests/e2e/laz-wasm.e2e.spec.ts`. Exercises the
+ * two mechanisms #2097 found unasserted: the Vite `?url` wasm-asset fetch
+ * and the `Module.wasmBinary` hand-off to emscripten. Deliberately calls
+ * `importLazPerf()` directly rather than the memoized `loadLazPerf` (so the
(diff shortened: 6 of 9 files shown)


BIMvoice:
Self-review before this goes to you, including one thing I got wrong and two design calls I would rather flag than have you find.

## Independently re-verified the mutation

I did not take the agent's evidence on trust, since a green-but-blind test is the exact defect #2097 is about. Against the real production build served by `vite preview`, driving the page with Playwright:

- **GREEN** — `{"ok":true,"hasLASZip":true}`.
- **RED** — corrupted the emitted URL in the built chunk (`laz-perf-<hash>.wasm` → a nonexistent name; single replacement asserted, and I confirmed the preview server actually served the corrupted text rather than a cache):

```
{"ok":false,"error":"Aborted(CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f @+0)"}
```

`3c 21 44 4f` is `<!DO` — the SPA fallback's `index.html`. That is the original bug's signature exactly, so the test fails for the right reason, not an incidental one. Restored by file copy; green again.

**My own error, worth recording:** my first attempt reported a *passing* probe under the corrupted URL, which looked like proof the test was vacuous. That was a bug in my harness, not in the test — `waitForFunction` returns a handle to the predicate's own boolean, so I was reading `true` from `x !== undefined` rather than reading `window.__LAZ_PROBE__`. Reading the object directly gave the RED above. Flagging it because "the verification tool was broken, not the thing under test" is a trap worth naming.

## Two calls I would like you to overrule if you disagree

**1. It adds a public export to a published package purely as a test hook.** `probeLazPerfWasmLoad` is exported from `@ifc-lite/pointcloud`'s index, lands in `api-surface.json`, and carries a changeset for a new minor export. It is marked `@internal`, but that is a doc convention, not an access control — it ships in the published `.d.ts` and becomes something we are implicitly on the hook for.

The alternative is keeping it unexported and having the E2E entry reach it by deep import, which trades a public-surface enlargement for a deep-path import from the app. I did not make that change because it depends on the package's `exports` map and on your view of which cost is worse. If you would rather not grow the surface for a test seam, say so and it is a small change.

**2. The test is gated behind a real-Chrome requirement it does not need.** Both `viewer-e2e` and `viewer-e2e-ci` pin `channel: 'chrome'`, deliberately — the comment explains Playwright's bundled shell has broken WebGPU under software rendering. This test uses no WebGPU at all; it fetches a wasm and instantiates it. Adding it to those projects means it cannot run locally without Chrome installed (both projects failed for me on exactly that), even though bundled Chromium runs it fine — which is how I got the evidence above.

That is not broken on CI, where Chrome is preinstalled. But a regression guard nobody can run locally is weaker than one they can, and the coupling is incidental rather than required. A separate project without the `channel` pin would fix it; I did not add one because a fourth project is a config change with your name on it.

## Scope

Not covered: decoding actual LAZ-compressed point data — no fixture exists and `laz-perf@0.0.7` is decode-only, so one cannot be synthesized. What is covered is the two mechanisms the issue named: the `?url` asset resolving to a real `application/wasm` response, and the `wasmBinary` hand-off. Reaching a working `LASZip` constructor implies both.

Also worth stating: this is a **regression guard, not a live fix**. I confirmed the mechanism works on `main` today — the hashed wasm is emitted and correctly referenced. The issue's "every browser load fails" framing did not reproduce.

louistrue:
Merging. Two things I checked rather than assumed, both correct:

**The changeset bump is `minor`, not `patch`** — `probeLazPerfWasmLoad()` is a new export on a published package, so a patch would have been wrong. Easy one to get wrong and you got it right.

**`laz-probe.html` really does ship in the production build** — the Rollup input is unconditional, so it is reachable in a deployed app. That is deliberate here, and correctly so: the whole value of this test is asserting the *real* production asset pipeline (`?url` resolution plus the `Module.wasmBinary` hand-off), so guarding it behind a test-only mode would test something other than what ships. Your changeset says "E2E-only, not linked from the app UI", which is honest about it. Flagging it only so nobody later sees an unfamiliar route and assumes it leaked in by accident — a one-line note in the HTML saying it is a deliberate E2E entry point would prevent that.

The gap this closes is worth stating plainly: every existing test substitutes the loader via `setLazPerfLoaderForTesting()`, so a broken `?url` resolution or `wasmBinary` hand-off was invisible until a real browser opened a LAZ file. That is a test suite that cannot fail for the failure mode that actually occurs — the same shape as several others found today.

BIMvoice:
Addressed — removed the public export.

**`packages/pointcloud/package.json` exports map**, before this change:

```json
"exports": { ".": { "types": "./dist/index.d.ts", "import": "./dist/index.js", "default": "./dist/index.js" } }
```

Only `"."` is declared, no subpaths. So a deep-import fix would have required *adding* a subpath (e.g. `./internal`), which is itself a public surface change — narrower than a named export, but still something we'd own.

I didn't need to go that route. `apps/viewer/vite.config.ts` already aliases `@ifc-lite/pointcloud` to the package's **source** (`packages/pointcloud/src`, not `dist`) for both dev and build — that's the existing pattern for most in-repo packages the viewer consumes. So the E2E entry (`apps/viewer/src/e2e/laz-wasm-probe-entry.ts`) now imports `probeLazPerfWasmLoad` by relative path straight from `packages/pointcloud/src/streaming/laz-source.ts`, instead of through the package's public index. Nothing is published — the function is a plain (non-exported-from-index) function in package source, same as `LasZipInstance`/`LazPerfModule` already were.

This is option 2 from the brief, and it turned out clean because the alias already existed; I didn't add or bend any build config for it.

**Re-verified end to end**, since a source-relative import that resolves cleanly but silently skips the real code path would be strictly worse than the export it replaces:

- Built `apps/viewer` for real (`pnpm --filter @ifc-lite/viewer build`), confirmed `laz-perf-*.wasm` and its hashed `.js` loader landed in `dist/assets/`, and confirmed `laz-probe.html` is in the build output.
- Ran `tests/e2e/laz-wasm.e2e.spec.ts` against `vite preview` — GREEN, `hasLASZip: true`.
- Corrupted the emitted wasm URL string in the built chunk (`laz-perf-CFJp03W6.wasm` → `laz-perf-CORRUPTED.wasm`) and reran — RED, with the exact SPA-fallback signature: `expected magic word 00 61 73 6d, found 3c 21 44 4f`.
- Restored the chunk by file copy from a pre-corruption backup and reran — back to GREEN.

So the test still proves what it always proved; only the import path changed.

**Other cleanup:**
- Reverted `packages/pointcloud/src/index.ts` to no longer export `probeLazPerfWasmLoad`.
- Regenerated `scripts/api-surface.json` (`node scripts/check-api-surface.mjs --update`) — the entry is gone, snapshot now matches a from-scratch check.
- Deleted `.changeset/laz-wasm-e2e-probe.md` — there's no public API change left to release-note. (`.changeset/laz-perf-load-retry.md`, the real behavioral fix in this package, is untouched.)
- `pnpm --filter @ifc-lite/pointcloud test` — 131 passed, 2 skipped. `pnpm --filter @ifc-lite/viewer typecheck` and `pnpm --filter @ifc-lite/pointcloud exec tsc --noEmit` — both clean.

Pushed to `upstream` as `a49cc239` on this branch.

BIMvoice:
**Correction to my previous comment — that change did not land, and the export is on `main`.**

This PR merged at 14:33 and the follow-up commit `a49cc239` was pushed to the branch *after* that, so it updated an already-merged branch and changed nothing. `probeLazPerfWasmLoad` is currently exported from `@ifc-lite/pointcloud`'s index and recorded in `scripts/api-surface.json` on `main`. Flagging it rather than letting the earlier comment stand as if the removal were done.

The work itself is sound and re-verified end to end, so it is available if you want it:

- `packages/pointcloud/package.json` declares only `"."` — no subpaths — so a deep-import path would have been a *new* public surface rather than a smaller one.
- `apps/viewer/vite.config.ts` already aliases `@ifc-lite/pointcloud` to `packages/pointcloud/src`, so the E2E entry can import the probe by relative source path and the function never needs to leave the package.
- Re-verified against a real production build: GREEN with `hasLASZip: true`, then corrupting the emitted wasm URL gives the SPA-fallback signature (`expected magic word 00 61 73 6d, found 3c 21 44 4f`), then GREEN again after restoring. So the test still proves what it was written to prove.

**Your call, and it is a release decision rather than a cleanup one now that it has shipped** — removing a published export is a surface change with version semantics, which is exactly why I would rather ask than open a PR unilaterally. Happy to raise one against `main` if you want it removed, or to leave it and reword the changeset to describe it as internal and unsupported.

Issue #1929, opened 2026-07-31
Act DXF as georeferencing


Hans-Lammerts:
### The problem

When DXF as overlay is used let it behave with respect to ifcmapconversion. Just like the feature of .laz now uses. When the ifc has values for xyz and rotation let the default be that dxf is georeferenced to fall into the same place

### Proposed solution

When dxf is imported uae contra values of ifcmapconversion.

Issue #1782, opened 2026-07-17
Add DXF import as a reference layer for 2D drawings


janbrouwer:
### The problem

IFC2x3 has explicit 2D annotation and dimension entities (`IfcDimensionCurve`, `IfcDraughtingCallout`, `IfcLinearDimension`, `IfcDimensionPair`). IFC4 removed them without a fixed replacement. `IfcAnnotation` is still there, but representing a dimension, leader, or drawing sheet now depends on ad hoc `PredefinedType`/`ObjectType` conventions that only a couple of tools (Bonsai/IfcOpenShell, FreeCAD) actually implement, confirmed by Dion Moult on the [buildingSMART forum](https://forums.buildingsmart.org/t/what-ever-happened-to-anootation-related-classes/5620/2): "these were removed in IFC4... vendors were primarily focused on 3D geometry." So cross-tool interoperability for this content in IFC4/4.3 is close to nonexistent today. Notably, even Bonsai's own drawing pipeline generates SVG "and optional DXF conversion" for exactly this content, DXF is where the ecosystem already lands when it needs to move annotations and dimensions between tools.

Practitioners routinely have existing DXF drawings, site surveys, subcontractor coordination sets, annotated floor plans, dimensioned details, that they want as project context alongside an IFC model. DXF's spec is published openly by Autodesk, but it's a single-vendor de facto standard, not governed by a community body the way IFC is by buildingSMART, worth flagging since "open" means something narrower here.

`ifc-lite` has no DXF parser today, so that context is simply unavailable in the viewer. Meanwhile `@ifc-lite/drawing-2d` already generates real 2D drawings from IFC geometry (floor plans, sections, hatching, symbols, title blocks) and renders them as vector SVG, there's a working 2D canvas to bring DXF content into, just nothing that reads DXF onto it.

### Proposed solution

[...]

PR #2114, opened 2026-08-04
feat(renderer,viewer): render the DXF reference layer in the 3D viewport

### LINKED ISSUE #2043: Show DXF reference layer in 3D context (visibility toggle, default on in 2D and 3D)
### The problem

DXF import landed via #1782 as a **reference layer in the 2D panel only**. In practice, DXF reference drawings are most useful **in the 3D context** — most of the time a DXF is a floor plan with annotations that are extremely helpful when placed in 3D alongside the model. A 2D-only DXF underlay is of limited use on its own.

We want to keep the existing 2D behaviour (some users already rely on it), and additionally make the DXF reference layer available in 3D.

### Proposed solution

- Keep the DXF reference layer available in **2D** (don't remove existing behaviour — it's already in use).
- Also render the DXF reference layer in the **3D context**, positioned against the IFC model's coordinate system (Z-up).
- Control visibility via a **display toggle in the visibility layers** panel — *not* a choice made during load. The user loads the DXF once and then toggles where it shows.
- **Default to visible in both 2D and 3D.**

### Alternatives considered

- **Amend/reopen #1782** to remove the 2D overlay and move DXF to 3D: rejected — we want to keep 2D since some users already use it, so this is filed as a new issue (per discussion with @janbrouwer).
- **Choice of 2D vs 3D during load:** rejected in favour of a persistent visibility-layer toggle, defaulting to visible in both.

---

Follow-up to #1782. cc @janbrouwer

BIMvoice:
Closes #2043.

## Current pipeline (before this PR)

DXF import landed in #1794 (the issue text names #1782, which is actually the *issue* this PR closed — the implementation PR is #1794):

- **Parser**: `packages/drawing-2d/src/dxf/` — ASCII DXF parser producing `DxfUnderlay` (world plan geometry, metres, +Y = north, both DXF and IFC are Z-up).
- **Store**: `apps/viewer/src/store/slices/drawing2DSlice.ts` — `DxfUnderlayState` holds the parsed underlay plus viewer state: `visible`, `opacity`, per-layer visibility, `placement` (offset/rotation/scale), and a tri-state `georeferenced` flag (issue #1929).
- **Math**: `apps/viewer/src/hooks/dxfUnderlayMath.ts` — pure, unit-tested `dxfUnderlayToDrawing` maps an underlay's raw points through: optional map/CRS→IFC-world transform (georeference) → render-frame shift (RTC + originShift) → 2D-only Y-flip/mirror → user placement. Output is 2D drawing-space geometry.
- **Render**: `Section2DPanel.tsx` / `Drawing2DCanvas.tsx` — draws the mapped underlay beneath the generated 2D section/plan drawing. `DxfUnderlayPanel.tsx` is the per-DXF control surface (visibility, opacity, layer toggles, placement, georeference checkbox, centre-on-model).
- There is **no unified "visibility layers panel"** in this codebase — each reference-layer feature (point cloud, Cesium/terrain, DXF, grid/annotation type-visibility) has its own independent toggle/panel. `DxfUnderlayPanel` is the closest thing to a DXF-specific visibility panel, so that's where I added the new 3D toggle.
- **3D reference overlays already exist** for grid axes (`useGridLines3D`) and `IfcAlignment` centerlines (`useAlignmentLines3D`): a flat `[x,y,z,...]` line-list in renderer Y-up world space, uploaded via `Renderer.uploadGridLines3D`/`uploadAlignmentLines3D` into a dedicated GPU buffer in `packages/renderer/src/section-2d-overlay.ts`, drawn through one shared line pipeline + one shared colour (`setOverlayLineColor`). DXF had no such path before this PR.

## What this PR adds

- `DxfUnderlayState` gets a second, independent visibility flag: `visible3D` (default `true`, alongside the existing `visible` for the 2D panel). Both are user-toggleable in `DxfUnderlayPanel` (new "2D"/"3D" eye buttons) — never chosen at load time, matching the issue's explicit rejection of that alternative.
- `dxfUnderlayMath.ts` gains two new pure functions:
  - `dxfElevationRenderY(coordinateInfo, elevationIfcZ = 0)` — the Y-axis (elevation) counterpart of the existing `dxfWorldShift` (X/Z-axis shift): `renderY = ifcZ - originShift.y - rtc.z`, derived from the same Z-up→Y-up conversion the mesh/grid/alignment pipeline already uses (`reproject.ts`, `ifc-origin.ts`).
  - `dxfUnderlayToWorldLines3D(entry, shift, elevationRenderY, mapToWorld?, georeferenceAvailable?)` — reuses the existing internal `worldToDrawing` point transform with `mirrorX` always `false` (the 2D flipped-section mirror doesn't apply to 3D) to produce a flat 3D line-list from an underlay's line paths.
- `useDxfUnderlay.ts` gains `useDxfUnderlays3DLines(coordinateInfo)`, filtering on `visible3D` and merging all eligible underlays into one buffer, independent of the 2D panel's section-axis/plan-view gating (the 3D overlay is always eligible, like grid/alignment).
- `packages/renderer`: `uploadDxfLines3D` / `clearDxfLines3D` / `hasDxfLines3D` / `drawDxfLines3D`, a new dedicated buffer mirroring `uploadGridLines3D` exactly (own buffer, shared line pipeline, shared `overlayLineColor`, does not expand model bounds/reframe the camera on upload since it's behind its own toggle).
- `Viewport.tsx` wires it all together, mirroring the existing grid-lines upload effect.

## Design decision: elevation

**DXF carries no Z**, and unlike `IfcAnnotation` (which has a per-primitive Z or falls back to a storey elevation via `elementToStorey`) a DXF underlay has no per-entity Z or storey association the #1782/#1929 import pipeline ever captured. Inventing one (e.g. guessing "ground floor" from spatial hierarchy) would be new, unrequested scope.

**Chosen default: IFC Z = 0 (model origin/grade).** This is the smallest sane default — no data is invented, and it matches the common case the issue describes (a site plan or ground-floor DXF). It is very likely wrong for a DXF meant to sit at some other storey's elevation; there is currently no UI to change it. A per-underlay elevation field (or "snap to nearest storey") is a reasonable follow-up but explicitly out of scope here — please push back if Z=0 is the wrong call for how this is actually used.

## Scope cut: lines only, no fills or per-layer colour in 3D

Only DXF **line paths** (walls/boundaries — the highest-value content of a floor-plan reference layer) are lifted to 3D. **Fills/hatches and text labels are not rendered in 3D in this PR.** The renderer does have a generic fill/text overlay pipeline with per-item colour (used by `IfcAnnotation`'s 3D overlay, `uploadAnnotationFills3D`/`uploadAnnotationTexts3D`), so extending it to DXF fills/text is plausible follow-up work, but doing it well (tessellated hatch fills at the same elevation, billboarded/oriented text) is materially more work than the line-list path and I judged it out of scope for a first iteration.

The **3D lines also lose per-DXF-layer colour**: the renderer's reference-overlay line pipeline already shares ONE colour across grid/alignment/annotation lines (a pre-existing constraint, not something I introduced), and I extended DXF into that same family rather than building a new per-vertex-coloured line pipeline. 2D keeps full per-layer colour, unaffected.

## Verification

- **RED→GREEN + mutation-tested** (python3 exact-substring replace, occurrence count asserted `== 1`, restored by file copy, `diff` confirmed identical):
  - `dxfElevationRenderY`: sign-flip mutation on the `originShift.y` term → 2 tests failed, confirmed caught.
  - `dxfUnderlayToWorldLines3D`: disabled the closed-path branch → 1 test failed; forced `mirrorX: true` → 3 tests failed.
  - `DxfUnderlayPanel`'s new 3D toggle button: wired it to the wrong store action → the new panel test failed.
  - `drawing2DSlice`'s `visible3D` default: flipped `true`→`false` → the new default-value test failed.
- Test counts: `apps/viewer` full suite (`npm test`, node:test) went from baseline (2606 pass / 2608 total, pre-existing 2 skipped) to **2609 pass / 2611 total**, 0 fail. `packages/renderer` suite: 414 pass / 0 fail (its own buffer/upload/draw methods for grid/alignment/DXF are GPU-device-backed and were already untested at that layer before this PR — I did not add tests there either, consistent with existing precedent, not a new gap).
- `pnpm typecheck` and `pnpm build` both green across the whole workspace.
- `pnpm check:api-surface` passes unchanged (new methods on an already-exported `Renderer` class don't change the tracked export surface).
- **Not verified**: I did not run this in a browser — no GPU/visual confirmation that the DXF overlay actually renders correctly against a real model. The renderer-side buffer/pipeline code follows the grid/alignment pattern exactly (same buffer creation, same draw call, same uniform layout), but I have not visually confirmed it. If you can smoke-test it against a model with a georeferenced or plain DXF, that would close the gap the mutation-tested math alone can't.

## Correction to the issue brief

My brief referenced "#1782 (the DXF import PR)" — #1782 is actually the **issue**; the implementing PR was **#1794** ("feat(drawing-2d,viewer): DXF import as 2D reference underlay"). Referenced #1794 throughout instead.

## Changeset

`@ifc-lite/renderer` is published, so it gets a `minor` changeset for the new public methods. `apps/viewer` is private (`@ifc-lite/viewer`) — no changeset for the viewer-side changes.

### DIFF
--- a/.changeset/dxf-3d-reference-layer.md
+++ b/.changeset/dxf-3d-reference-layer.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/renderer": minor
+---
+
+Add a 3D DXF reference-layer overlay to the WebGPU renderer: `uploadDxfLines3D` / `clearDxfLines3D` upload a flat `[x,y,z,x,y,z,...]` line-list (mirroring the existing `uploadGridLines3D`/`uploadAlignmentLines3D` pattern — a dedicated buffer, drawn through the shared reference-overlay line pipeline and colour) and render it alongside the grid/alignment/annotation overlays.
+
+Follow-up to the viewer's DXF import feature (issue #1782/#1929), which only ever rendered the imported DXF as a 2D drawing-panel underlay. Issue #2043 adds a 3D viewport toggle for the same DXF data (`apps/viewer`'s `DxfUnderlayPanel`, private package, no changeset needed for that half); this changeset covers the new renderer API those toggles call into.
+
+Only line paths (walls/boundaries) are lifted to 3D in this iteration — DXF fills/hatches and text labels are not yet rendered in the 3D scene, and per-DXF-layer colour is not carried through (the 3D overlay shares one colour with the grid/alignment/annotation line family, same as those already do among themselves).
--- a/apps/viewer/src/components/viewer/DxfUnderlayPanel.test.tsx
+++ b/apps/viewer/src/components/viewer/DxfUnderlayPanel.test.tsx
@@ -44,6 +44,7 @@ function underlayState(overrides: Partial<DxfUnderlayState> = {}): DxfUnderlaySt
     name: 'site.dxf',
     underlay: emptyUnderlay('site.dxf'),
     visible: true,
+    visible3D: true,
     opacity: 1,
     layerVisibility: {},
     placement: { ...DEFAULT_DXF_PLACEMENT },
@@ -142,5 +143,52 @@ describe('DxfUnderlayPanel: "Align to model georeference" tri-state control (PR 
     useViewerStore.setState({ dxfUnderlays: [underlayState({ id: 'u1', georeferenced: false })] });
     const container = renderPanel(true); // availability true, but pinned false wins
     assert.equal(georefCheckbox(container).checked, false, 'pinned false must not follow availability=true');
+  });
+});
+
+describe('DxfUnderlayPanel: independent 2D/3D visibility toggles (issue #2043)', () => {
+  beforeEach(() => {
+    useViewerStore.setState({ dxfUnderlays: [] });
+  });
+
+  afterEach(() => {
+    for (const { root, container } of mounted.splice(0)) {
+      act(() => {
+        root.unmount();
+      });
+      container.remove();
+    }
+    useViewerStore.setState({ dxfUnderlays: [] });
+  });
+
+  function toggleButton(container: HTMLElement, titleSubstring: string): HTMLButtonElement {
+    const btn = [...container.querySelectorAll('button')].find((el) =>
+      el.getAttribute('title')?.includes(titleSubstring),
+    );
+    assert.ok(btn instanceof HTMLButtonElement, `expected a button titled like "${titleSubstring}"`);
+    return btn;
+  }
+
+  it('both default to visible, and are independent toggles wired to distinct store actions', async () => {
+    useViewerStore.setState({ dxfUnderlays: [underlayState({ id: 'u1' })] });
+    const container = renderPanel(false);
+    assert.equal(useViewerStore.getState().dxfUnderlays[0].visible, true);
+    assert.equal(useViewerStore.getState().dxfUnderlays[0].visible3D, true);
+
+    // Hide only the 3D overlay; the 2D underlay's flag must be untouched.
+    await act(async () => {
+      toggleButton(container, 'Hide in 3D view').click();
+    });
+    let entry = useViewerStore.getState().dxfUnderlays.find((u) => u.id === 'u1');
+    assert.equal(entry?.visible3D, false, '3D toggle must flip visible3D');
+    assert.equal(entry?.visible, true, '3D toggle must NOT touch the 2D visible flag');
+
+    // Hiding the 2D underlay must not touch the (now-off) 3D flag either.
+    await act(async () => {
+      toggleButton(container, 'Hide in 2D drawing view').click();
+    });
+    entry = useViewerStore.getState().dxfUnderlays.find((u) => u.id === 'u1');
+    assert.equal(entry?.visible, false, '2D toggle must flip visible');
+    assert.equal(entry?.visible3D, false, '2D toggle must NOT touch visible3D');
   });
 });
--- a/apps/viewer/src/components/viewer/DxfUnderlayPanel.tsx
+++ b/apps/viewer/src/components/viewer/DxfUnderlayPanel.tsx
@@ -175,9 +175,21 @@ function UnderlayCard({
         </div>
       )}
 
-      {/* Opacity */}
+      {/* Opacity — PR #2114 review: the slider only affects the 2D drawing
+          panel. The 3D viewport's line pipeline (`Section2DOverlayRenderer`)
+          shares one un-blended `linePipeline`/uniform colour across the
+          grid, alignment, annotation and DXF line overlays; giving each DXF
+          underlay its own alpha would mean splitting the merged 3D DXF
+          line buffer (`useDxfUnderlays3DLines`) into a per-underlay draw
+          call and adding blend state to that shared pipeline — out of
+          scope here, so `useDxfUnderlays3DLines`'s `opacity > 0` check
+          stays a binary gate. The title below and the "(2D)" suffix make
+          that explicit rather than leaving the control silently no-op in
+          3D. */}
       <div className="flex items-center gap-2 px-1">
-        <Label className="text-[10px] text-muted-foreground w-12">Opacity</Label>
+        <Label className="text-[10px] text-muted-foreground w-12" title="Opacity applies to the 2D drawing only — the 3D view always renders this underlay fully opaque when its 3D toggle is on">
+          Opacity (2D)
+        </Label>
         <input
           type="range"
           min={0.1}
@@ -186,6 +198,7 @@ function UnderlayCard({
           value={state.opacity}
           onChange={(e) => setDxfUnderlayOpacity(state.id, Number.parseFloat(e.target.value))}
           className="flex-1 h-1.5 accent-primary"
+          title="Opacity applies to the 2D drawing only — the 3D view always renders this underlay fully opaque when its 3D toggle is on"
         />
         <span className="text-[10px] text-muted-foreground w-8 text-right">{Math.round(state.opacity * 100)}%</span>
       </div>
--- a/apps/viewer/src/components/viewer/DxfUnderlayPanel.tsx
+++ b/apps/viewer/src/components/viewer/DxfUnderlayPanel.tsx
@@ -23,6 +23,15 @@
  * (`resolveEffectiveGeoreferenced`, following `georeferenceAvailable`
  * while in auto mode) and only becomes an explicit `true`/`false` — pinned
  * regardless of anchor availability — once the user actually clicks it.
+ *
+ * Issue #2043: the 2D underlay is one of TWO independent visibility
+ * toggles per entry — `visible` (2D drawing panel, this panel's original
+ * behaviour) and `visible3D` (3D viewport overlay, `Viewport.tsx`'s
+ * `useDxfUnderlays3DLines`). Both default to on and are controlled here,
+ * not at import time, per the issue's explicit rejection of a load-time
+ * 2D-vs-3D choice. The 3D overlay currently renders line paths only
+ * (walls/boundaries); fills/hatches and text labels are not lifted to 3D
+ * yet (`dxfUnderlayToWorldLines3D`'s doc in `dxfUnderlayMath.ts`).
  */
 
 import React, { useCallback, useRef, useState } from 'react';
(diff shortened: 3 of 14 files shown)


BIMvoice:
Self-review before you spend time on this. I verified the tests independently, and there is one scope cut I think you should weigh before reviewing the code, because it may not match what the issue was actually asking for.

## Verified independently

Mutating the RTC term out of `dxfElevationRenderY` (`elevationIfcZ - shift.y - rtc.z` → drop `rtc.z`; single replacement asserted, region re-read) fails 2 tests; restored by file copy, full `apps/viewer` suite 0 fail. So the elevation math is genuinely pinned rather than asserted-into-existence.

Also worth stating: a two-dot `git diff upstream/main` on this branch looks alarming — it appears to delete `recent-files.test.ts`, `ifc-cache.test.ts`, `log-budget-reset.test.ts` and others. That is stale-base noise, not the diff: those files come from PRs that merged *after* this branch point. Against the merge base (`531a9559`) the change is **12 files, additive** — 485 insertions, 12 deletions, nothing removed that it did not add. Worth a rebase before merge regardless.

## The scope cut I want you to look at first

**Only line paths render in 3D. Fills/hatches and text labels stay 2D-only.**

The issue's motivation is:

> most of the time a DXF is a floor plan **with annotations** that are extremely helpful when placed in 3D alongside the model

Annotations are text. So the feature as built delivers the geometry of the plan into 3D but not the part the issue names as the reason for wanting it. That is not a defect in the implementation — 3D text is a genuinely different problem (billboarding, screen-space sizing, depth behaviour) and the line pipeline was the honest reuse. But it does mean this may only partly answer @janbrouwer's need, and it is better for you to decide that now than after review.

A related consequence: 3D lines share one colour with the grid/alignment/annotation family, so **per-DXF-layer colour is lost in 3D** while 2D keeps it. On a plan whose layers are colour-coded, the 3D view will read as a single-colour wireframe.

## Two other calls to overrule if you disagree

**Toggle placement.** The issue specifies "a display toggle in the **visibility layers** panel". There is no unified visibility-layers panel in this codebase — each reference-layer feature carries its own toggle — so this went into `DxfUnderlayPanel.tsx`, the existing DXF-specific visibility surface. If you intended a shared panel, that is a larger piece of work and this is not it.

**Elevation.** DXF carries no Z, so 3D placement needs one invented. Default is **IFC Z=0** (model origin/grade): the smallest defensible choice, correct for the site/ground-floor plan case, and wrong for a plan belonging to any other storey. A storey picker or a user-settable elevation is the obvious follow-up; I did not build one because which of those you want is a product call.

## Not verified

No browser or GPU smoke test — the render path itself is unexercised here. What is tested is the pure math (world→3D line transform, elevation resolution) and the store/toggle state. `packages/renderer` 414 pass, `apps/viewer` 2606 → 2609, typecheck and build green.

louistrue:
Viewer tier, so I am not going to mutation-test this — but both open threads are real and one is user-visible in the feature you are adding.

**Fractional opacity is discarded in the 3D path.** Confirmed:

```ts
const visible = dxfUnderlays.filter((u) => u.visible3D && u.opacity > 0);
```

Opacity is used purely as an on/off gate, and the `Float32Array` of line vertices carries no alpha, so an underlay set to 0.3 renders identically to one at 1.0 in the viewport. The 2D underlay honours the slider; 3D silently rounds it to "visible".

Either is fine, but pick one:
- honour it — plumb alpha through to the line material, or
- document it — say in the UI (or at minimum a code comment) that the 3D reference layer is binary, so the slider is not silently lying about what it does.

Silently discarding it is the option I would avoid: the control moves, nothing changes, and the user concludes the feature is broken rather than scoped.

**The unguarded GPU upload** (`createBuffer`/`writeBuffer` in `Viewport.tsx`) is worth a look too. Device loss and memory pressure are real on this path, and an uncaught throw inside a render effect tears down the React tree rather than degrading the overlay — that failure mode has bitten this repo before with GPU instancing. Not necessarily this PR's job to make bulletproof, but a `try` that drops the underlay and warns beats losing the viewport.

Everything else is green — 5 core lanes, no failures. These two threads are all that stand between this and a merge, and neither needs a deep investigation from me to act on.

BIMvoice:
Pushed 90bf3177 addressing both threads.

**Fractional opacity is discarded in 3D** — chose to document rather than plumb, and here's why: `useDxfUnderlays3DLines` merges every visible underlay's line paths into a *single* `Float32Array` for a *single* draw call (`uploadDxfLines3D`), and that draw reads colour from `Section2DOverlayRenderer.overlayLineColor` — one shared RGBA uniform also used by the grid/alignment/annotation/section-cut line overlays via the same `linePipeline`. That pipeline has no blend state configured on its fragment target, so even a passed-through alpha wouldn't currently blend. Honouring real per-underlay opacity would mean (a) adding blend state to a pipeline four other overlays share, and (b) splitting the merged single draw into one draw per underlay so each can carry its own alpha — a restructuring of the shared line pipeline, disproportionate for this PR.

Instead: `DxfUnderlayPanel.tsx`'s slider is now labelled **"Opacity (2D)"** with a title tooltip ("Opacity applies to the 2D drawing only — the 3D view always renders this underlay fully opaque when its 3D toggle is on"), plus explanatory comments at the `useDxfUnderlays3DLines` filter and the panel. The control no longer silently no-ops in 3D.

**Unguarded GPU upload** — extracted the effect body into `uploadDxfLines3DGuarded()` (`apps/viewer/src/components/viewer/dxf-lines-3d-upload.ts`), which wraps `createBuffer`/`writeBuffer` with the repo's existing `runGpuUpload()` guard (`gpu-upload-guard.ts`, already used by the geometry-streaming upload sites — same containment pattern you referenced). On failure it drops the underlay via `clearDxfLines3D()` instead of letting the throw reach the render effect, and the warning is latched once per call site (verified: 25 repeated failures → 1 console.warn) rather than firing per re-run; a guard reset (fresh context) reports again.

Testing: extracted the drop-on-failure branch into a GPU-free pure function and covered it with 5 new `node:test` cases (empty→clear, success→no-clear, throw→clear-not-throw, warn-once-across-25-reruns, reset→warns-again). The actual `createBuffer`/`writeBuffer` call and any real device-loss behaviour are not exercised — no GPU in this environment, so that part is unverified beyond code review. Full `apps/viewer` suite: 2616 tests, 2614 pass / 2 skipped (unchanged skip count) — up from 2611 before this patch, all green. `pnpm typecheck` passes across the workspace. Mutation-tested both fixes independently (python3 exact-substring swap, restored by file copy) — each mutation reliably fails its own tests, and they fail disjoint tests (the opacity fix is UI-text/doc-only with no behavioural mutation surface). Also re-ran the existing RTC-offset mutation on `dxfElevationRenderY` to confirm it still fails as before — untouched by this patch.

No changeset: `@ifc-lite/viewer` is `private: true`, and `packages/renderer` wasn't touched (no public-surface change).

louistrue:
Both threads addressed, both verified. Merging.

## The GPU upload guard

```
### rethrow instead of containing
    not ok 3 - drops the underlay (clears) instead of propagating when the upload throws
    not ok 4 - warns once per call site, not once per re-run (render-effect re-execution)
    not ok 5 - a fresh context (guard reset) reports again

### drop the per-label log latch
    not ok 4 - warns once per call site, not once per re-run (render-effect re-execution)
```

Restored: 5/5.

Extracting `runGpuUpload` into a shared `gpu-upload-guard.ts` rather than wrapping this one call site is the better shape — it puts the DXF path on the same footing as the geometry-streaming uploads instead of leaving a one-off. And the latch test earns its place: React effects re-run, so an unlatched warn floods the console on every frame during a device-loss, which is how a useful diagnostic becomes noise people filter out. `resetGpuUploadGuardForTests` plus the "fresh context reports again" case keeps the latch from being permanently sticky.

## Opacity

You took the document-it option and went further than I asked — the slider is now labelled **"Opacity (2D)"** with a `title` explaining that 3D always renders fully opaque. That is a real accessible name rather than a hover-only affordance, so it holds up.

The code comment gives the reason I could not have inferred:

> the shared `Section2DOverlayRenderer.linePipeline` (also used by the grid, alignment and annotation line overlays) has no blend state, so a passed-through alpha wouldn't visibly blend anyway. Plumbing real per-underlay opacity through would mean adding blend state to that shared pipeline and splitting this single merged draw into one draw per underlay

That reframes it from "opacity is dropped" to "opacity is not expressible in this pipeline without splitting the merged draw" — a scope decision with a cost, not an oversight. Scoping it out of a DXF-in-3D feature is right; silently ignoring the control was the only unacceptable option, and that is gone.

All 5 core lanes green, no failures.

Issue #2116, opened 2026-08-04
Captured-log byte budget is check-then-add, so one entry can exceed it by any amount


louistrue:
## Summary

`packages/sandbox/src/bridge.ts` caps captured console output at `MAX_TOTAL_BYTES = 4 * 1024 * 1024`, but the budget is checked **before** an entry is added and never **against** it. One oversized entry is therefore retained in full, and the ceiling is exceeded by whatever that entry happens to weigh.

```js
if (logs.length >= MAX_LOG_ENTRIES || totalBytes >= MAX_TOTAL_BYTES) {
  truncated = true;
  logs.push({ level: 'warn', args: [TRUNCATION_MARKER], timestamp: Date.now() });
  return;
}
// ...
totalBytes += entryBytes;   // <- can push totalBytes arbitrarily past the cap
logs.push({ level, args: entryArgs, timestamp: Date.now() });
```

## Reproduction

A single `console.log`, no loop, against the 4 MB budget:

```js
const payload = "x".repeat(40 * 1024 * 1024);
console.log(payload);
```

```
entries retained : 1
chars retained   : 41943040
budget (bytes)   : 4194304
overshoot factor : 10.0x
```

40 MB was an arbitrary choice — the overshoot is bounded only by what the script passes, so 400 MB gives 100x. For a sandbox whose purpose is containing untrusted code, the bound not being a bound is the part that matters.

## Not a regression

This predates #2096 and #2087. The existing test `charges serializable entries exactly as before and truncates on the same entry` documents the same property honestly for the small case: four entries reach 4194320 against a 4194304 budget, so the limit is already exceeded by 16 bytes by design before the fifth call truncates. #2096 neither introduced nor widened it, which is why it was left out of that PR.

## Suggested fix

Check the entry against the remaining budget rather than only the running total:

```js
if (totalBytes + entryBytes > MAX_TOTAL_BYTES) { /* truncate before pushing */ }
```

This is a deliberate behaviour change, not a pure bug fix: entries that currently land would start being dropped, and `charges serializable entries exactly as before and truncates on the same entry` would need updating with intent (it currently exists to prove the boundary did *not* move). Truncating the oversized entry rather than discarding it whole is worth considering, so a single large log still shows its head.

## Related

Found while reviewing #2096; CodeRabbit raised the same shape there and it was deferred here rather than folded into a scoped PR. Same family as #2104 — a bound or a scale that silently does not mean what its name says.

BIMvoice:
[...]

PR #2118, opened 2026-08-04
fix(sandbox): make the captured-log byte budget a real ceiling

### LINKED ISSUE #2117: Sandbox log budget is checked before an entry is added, so one oversized log call blows the 4MB ceiling
Split out of #2096 review, where @louistrue found and measured it. Filing so the probe survives that PR merging — he offered to open it and I am doing so with his data rather than making him repeat the work.

## What happens

`buildConsole` checks the budget *before* adding an entry, never against it:

```js
if (logs.length >= MAX_LOG_ENTRIES || totalBytes >= MAX_TOTAL_BYTES) { truncated = true; ... return; }
```

So the entry that crosses the ceiling is retained **in full**, and the overshoot is bounded only by what the script chooses to pass. One `console.log`, no loop:

```
PROBE entries retained : 1
PROBE chars retained   : 41943040
PROBE budget (bytes)   : 4194304
PROBE overshoot factor : 10.0x
```

A 40 MB payload gives 10x; 400 MB gives 100x. For a sandbox whose whole purpose is running untrusted code, "the cap is advisory and the script picks the real number" is the part that matters.

## Not introduced by #2096, and deliberately not fixed there

`main` already has the check-then-add shape. #2096 is about unsizeable arguments being retained for free (#2087) — it neither introduces nor widens this. The existing test comment states the property honestly: four entries reach 4194320 against a 4194304 budget, so the limit is already exceeded by 16 bytes by design before the fifth call truncates.

Worth recording why it was kept out of that PR: one of the bot findings there asked for the `charges serializable entries exactly as before and truncates on the same entry` test to be rewritten so the fourth entry truncates. That test exists precisely to prove #2096 did **not** move the boundary, which is what makes it safe to review — rewriting it to assert new behaviour would delete the evidence and fold an unrelated fix into a scoped PR.

## Suggested fix

Make the check `totalBytes + entryBytes > MAX_TOTAL_BYTES` and truncate **before** pushing, so the ceiling is a ceiling.

This is a **deliberate behaviour change**: entries that currently land would start being dropped, including in the existing boundary test (four full entries then truncation would become three, or the fourth would be replaced by the marker). That is the intended effect, but it means the change needs its own PR and its own updated expectations rather than riding along with a scoped fix.

## Related looseness, same one-line shape

Two measurement gaps in the same area, both worth closing with the fix:

- **The magnitude of the fallback charge** was unpinned until [`ba42fd32`](https://github.com/LTplus-AG/ifc-lite/pull/2096) — a 10-million-fold overcharge in `retainSizeableArgs` left the whole suite green, because the exact-charge test drives the happy path and that function only runs after `JSON.stringify` has already thrown. Now covered in both directions.
- **`String.length` counts UTF-16 code units, not bytes**, so the budget under-counts UTF-8 by ~133% for CJK text (`JSON.stringify(['日本語のログ出力'])` → `length` 12, `Buffer.byteLength` 28). The code says "approximate host cost", which is honest — but if this issue's fix makes the ceiling a real ceiling, it is worth deciding at the same time whether the unit should become bytes (`TextEncoder`/`Buffer.byteLength`) rather than code units. A hard cap measured in the wrong unit is a hard cap in the wrong place.

Neither is urgent on its own; both are cheap to fold into a change that is already touching this arithmetic.

BIMvoice:
Closes #2117

## The defect

`buildConsole` in `packages/sandbox/src/bridge.ts` checked `totalBytes >= MAX_TOTAL_BYTES` *before* the entry about to be added was sized, never against it. So `totalBytes` started at 0 (`< MAX_TOTAL_BYTES`) and a single oversized argument — e.g. one `console.log` of a 40 MB string — was retained in full. The check only caught up on the *next* call, by which point the overshoot was already on the host heap and bounded only by whatever the untrusted script chose to log (measured: 10x overshoot at 40 MB, 100x at 400 MB, per the issue's probe data).

## The fix

Size the entry first, then check `totalBytes + entryBytes > MAX_TOTAL_BYTES` *before* pushing it. An entry that would cross the ceiling is refused (replaced by the truncation marker) instead of retained, so the 4 MB budget is an actual ceiling rather than advisory.

The entry-count cap (1000 entries) is unchanged — it increments by exactly one per call, so its overshoot was already bounded to a single entry and doesn't have the same unbounded shape.

## The boundary moved — deliberately

The existing test `charges serializable entries exactly as before and truncates on the same entry` pinned the *old* boundary honestly: 4 full 1 MB entries (4 × 1048580 = 4,194,320 bytes) landed against the 4,194,304-byte budget — 16 bytes over by design — before the 5th call truncated.

Under the new check, walking the running total entry by entry:
- entry 1: `0 + 1048580 = 1048580 <= 4194304` → pushed
- entry 2: `1048580 + 1048580 = 2097160 <= 4194304` → pushed
- entry 3: `2097160 + 1048580 = 3145740 <= 4194304` → pushed
- entry 4: `3145740 + 1048580 = 4194320 > 4194304` → **refused**, truncation marker pushed instead

So only 3 full entries land (3,145,740 bytes), followed by the marker — `result.logs` has length 4, not 5. Updated that test (and the analogous boundary assertions in `log-budget-reset.test.ts`) with the derivation in a comment, not just the new numbers.

Also added a new test, `refuses a single entry that would overshoot the byte ceiling, rather than retaining it in full`, which is the direct regression test for the issue: one 5 MB `console.log` argument must be refused outright (not retained), and cumulative retained bytes must stay under the 4 MiB budget.

## Byte-unit question (evaluated, not changed)

The issue also asked to evaluate switching from `String.length` (UTF-16 code units) to real UTF-8 byte counting (`TextEncoder().encode(str).length`, not `Buffer.byteLength` — this runs in browsers too), since code units undercount multi-byte UTF-8 by up to ~133% for e.g. CJK.

Measured `.length` vs `new TextEncoder().encode(str).length` on 1 MB / 10 MB ASCII strings: the encode call was roughly 1000–2000x slower, because it allocates a new buffer proportional to the string being measured. That cost would land on *every* captured log call, including the adversarial ones this cap exists to cheaply reject — sizing a 40 MB adversarial string would itself allocate tens of MB of scratch buffer before the entry is even refused. Left as code-unit counting for now; documented the tradeoff in a comment above `MAX_TOTAL_BYTES` in `bridge.ts` so the gap is recorded rather than silent.

## Testing

- RED: new regression test written first, failed against unfixed code (full 5 MB string retained instead of the truncation marker).
- GREEN: `packages/sandbox` full suite — 11 files / 69 tests passing (baseline before this change: 68 tests).
- `pnpm exec tsc --noEmit` in `packages/sandbox`: clean.
- Mutation test: reverted the pre-push byte check back to the old check-before-add shape via an exact-substring Python replacement (verified `n == 1`), re-ran the suite — 5 failures (the new regression test, the updated boundary test, and 3 in `log-budget-reset.test.ts`), confirming the tests actually pin this fix. Restored the fixed file by copy (not `git checkout --`), suite green again.

## Changeset

`.changeset/sandbox-log-budget-ceiling.md` — patch bump for `@ifc-lite/sandbox`, states the behavior change plainly: a script logging one very large payload now sees truncation on that call rather than after it.

### DIFF
--- a/.changeset/sandbox-log-budget-ceiling.md
+++ b/.changeset/sandbox-log-budget-ceiling.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/sandbox": patch
+---
+
+Make the captured-log byte budget an actual ceiling.
+
+The bridge caps captured console output by cumulative serialized size (4 MB), because `vm.dump` copies sandbox values onto the host heap, which the QuickJS memory limit does not bound. The check ran *before* an entry was sized: it compared the running total against the budget, then retained the entry unconditionally, so a single oversized argument (e.g. one `console.log` of a 40 MB string) was retained in full — the check only caught up on the next call, by which point the overshoot was already on the host heap and bounded only by whatever the script chose to log.
+
+The check now runs against the entry about to be added, before it is retained: an entry that would push the cumulative total over the 4 MB budget is refused and replaced by the truncation marker instead of being kept. This is a deliberate behavior change — a script logging one very large payload now sees truncation on that call rather than after it, and the existing boundary test's expectations moved accordingly (three full 1 MB entries plus a marker, not four).
+
+The entry-count cap (1000 entries) is unchanged: it increments by exactly one per call, so its overshoot was already bounded to a single entry and does not have the same unbounded-overshoot shape.
--- a/packages/sandbox/src/bridge-console-budget.test.ts
+++ b/packages/sandbox/src/bridge-console-budget.test.ts
@@ -107,24 +107,72 @@ describe('captured-log byte budget', () => {
     }
   });
 
+  it('refuses a single entry that would overshoot the byte ceiling, rather than retaining it in full', async () => {
+    // #2117: the budget used to be checked BEFORE the entry about to be added,
+    // never against it — so `totalBytes` started at 0 (< MAX_TOTAL_BYTES) and a
+    // single oversized argument sailed through and was retained whole, with the
+    // overshoot bounded only by whatever the script chose to log. One 5 MB
+    // string (well over the 4 MiB budget) must now be refused outright and
+    // replaced by the truncation marker, not retained.
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      await withSandbox(async (sandbox) => {
+        const result = await sandbox.eval(
+          'console.log("x".repeat(5 * 1024 * 1024)); 1;',
+        );
+        expect(result.value).toBe(1);
+
+        // The oversized entry must not be present at all.
+        expect(result.logs).toHaveLength(1);
+        expect(result.logs[0]).toMatchObject({ level: 'warn', args: [TRUNCATION_MARKER] });
+
+        // Whatever *is* retained must stay within the 4 MiB budget — this is
+        // the actual property under test: cumulative retained bytes must never
+        // exceed MAX_TOTAL_BYTES, no matter how large a single call's argument.
+        const retainedBytes = JSON.stringify(result.logs.flatMap((entry) => entry.args)).length;
+        expect(retainedBytes).toBeLessThan(4 * 1024 * 1024);
+      });
+    } finally {
+      warn.mockRestore();
+    }
+  });
+
   it('charges serializable entries exactly as before and truncates on the same entry', async () => {
     const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
     try {
       await withSandbox(async (sandbox) => {
         // Each entry serializes to ["x"*1048576] = 1048576 + 2 quotes +
-        // 2 brackets = 1048580 bytes. Four entries reach 4194320, just past
-        // the 4194304-byte budget, so the fifth call is the one that
-        // truncates and the fourth is not. An off-by-more-than-four-bytes
-        // change in per-entry sizing moves that boundary.
+        // 2 brackets = 1048580 bytes.
+        //
+        // #2117 deliberately moved this boundary. The OLD check-before-add
+        // code (`totalBytes >= MAX_TOTAL_BYTES`, evaluated before the entry
+        // about to be added was even sized) let a fourth full entry land:
+        // 4 x 1048580 = 4194320, which is 16 bytes OVER the 4194304 budget —
+        // the bug this issue fixed, illustrated exactly by that overshoot —
+        // and only the fifth call truncated.
+        //
+        // The NEW check (`totalBytes + entryBytes > MAX_TOTAL_BYTES`,
+        // evaluated before the entry is pushed) walks the running total entry
+        // by entry:
+        //   entry 1: totalBytes   0 + 1048580 = 1048580 <= 4194304 -> pushed
+        //   entry 2: totalBytes 1048580 + 1048580 = 2097160 <= 4194304 -> pushed
+        //   entry 3: totalBytes 2097160 + 1048580 = 3145740 <= 4194304 -> pushed
+        //   entry 4: totalBytes 3145740 + 1048580 = 4194320 >  4194304 -> REFUSED,
+        //            truncation marker pushed instead, `truncated` latches true
+        //   entries 5, 6: `truncated` is already true, eval returns immediately
+        //
+        // So only 3 full entries land (3145740 bytes retained, not 4194320 —
+        // a 4th would put the total 16 bytes over budget), followed by one
+        // truncation marker: `result.logs` has length 4, not 5.
         const result = await sandbox.eval(
           'const payload = "x".repeat(1048576);\nfor (let i = 0; i < 6; i++) console.log(payload);\n7;',
         );
         expect(result.value).toBe(7);
-        expect(result.logs).toHaveLength(5);
-        expect(result.logs.slice(0, 4).map((entry) => (entry.args[0] as string).length)).toEqual([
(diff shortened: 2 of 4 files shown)


louistrue:
Question about your draft convention, because I think we may be deadlocked on it.

You have 11 open drafts right now (#2106, #2107, #2108, #2109, #2111, #2113, #2114, #2115, #2118, #2119, #2122). Every one of them is **green with all five core CI lanes passing**, and most carry a changeset and test files. Meanwhile every PR I have merged today was opened ready-for-review, and none of the drafts has ever been converted — so if you are waiting for a signal from us to undraft, and we are treating draft as "still being worked on", nothing moves.

**Which is it?**

- If draft means *not ready, do not look yet* — fine, I will leave them alone and you flip them when you want eyes on them.
- If draft means *done, queued, tell me when you have capacity* — say so and I will start pulling from that queue directly. **You do not need to wait for us; open ready and we will review.**

For context on what bar each PR now faces, since it changes how much scrutiny to expect and where it is worth pre-empting review:

- **`rust/**` and the kernel** — the highest bar. Mutation testing, negative controls, `cargo test --no-fail-fast`, clippy. Never merged on green CI alone.
- **Published packages** (`geometry`, `export`, `parser`, `data`, `mutations`, `sandbox`) — CI green with core lanes actually present, changeset correctness, and verification of the central claim.
- **`apps/viewer`** — CI green, core lanes present, threads resolved, merged. Display-level inconsistencies become follow-up issues rather than holds (as I just did with #2121 off #2070).

One practical warning that has bitten three of your PRs today: **a `CONFLICTING` branch stops GitHub scheduling `pull_request` workflows entirely**, so the PR page shows a green rollup with every real lane simply *absent* rather than failing. #2080, #2091 and now #2105 all hit it. If you see a suspiciously small check count, merge `origin/main` first — main is moving fast today.

Answer whichever way is true and I will work to it.

BIMvoice:
**Answering the draft question: option two — done, queued, pull from it freely.** All twelve are now marked ready for review (#2106, #2107, #2108, #2109, #2111, #2113, #2114, #2115, #2118, #2119, #2122, #2123).

You diagnosed a deadlock that was entirely mine, and it was habit rather than intent: I have been passing `--draft` on every `gh pr create` as a default, not as a signal. Nothing was waiting on you. Going forward I will open ready and use draft only when I genuinely mean "not finished" — and say so in the body when I do.

Before flipping them I re-verified the two I had not personally checked, since "queued for review" is a claim about readiness and I would rather it be true:

- **#2118** — reverting the ceiling to the check-then-add form fails 5 tests across 2 files; restored, 69/69.
- **#2119** — mutating `1000 / scaleFactor` fails 6 of 15 scale tests; restored, 129/129.

Both were as reported. Everything else in the set I had already mutation-verified as it landed.

## Two things worth your eye rather than buried in a thread

**#2119 adds a fourth implementation of the world-metres → paper-mm transform.** Three already exist (`SVGExporter.computeTransform`, `sheet-types.ts:calculateDrawingTransform`, and an inline copy in `useDrawingExport.ts`). The new one is deliberately *not* a reuse: `calculateDrawingTransform` applies `fitScale = min(scaleX, scaleY, 1)`, which silently shrinks the drawing when it does not fit the viewport — right for an on-screen preview, wrong for a PDF whose entire purpose is to be measured at the scale the user picked. That is documented at the top of `pdf-scale.ts`. It is still a fourth copy, and consolidating them is the kind of drift you have flagged elsewhere, so flagging it rather than letting it pass unremarked.

**#2113 adds a public export purely as a test hook** (`probeLazPerfWasmLoad` on `@ifc-lite/pointcloud`, with a changeset). Marked `@internal`, but that is a convention, not access control. Raised on that PR; repeating it here because it is the sort of thing that is cheap to change now and expensive later.

## On the tiering and the CONFLICTING trap

Both noted, and the second has now bitten four of mine — #2080, #2091, #2105 and, earlier, #2082 in a different way. I have adopted your check: a green rollup means nothing until the five core lanes are confirmed *present*, since absence renders identically to success.

Related, on #2082 specifically: the reason nothing had been pushed there since `ee485263` is that an earlier fix of ours went to `upstream` by mistake, creating a stray branch attached to no PR while the PR head sat on the fork. It is now on `origin` at `42431f1b` and the stray is deleted, so your regression finding has an actual fix to look at.

louistrue:
Closes #2116, verified with the exact case from that issue. Merging.

| | before | after |
|---|---|---|
| `console.log("x".repeat(40MB))` retained | 41,943,040 chars | **41 chars** (the marker) |
| vs the 4 MiB budget | 10.0x over | 0.00001x |

The ceiling is now a ceiling.

Two things I want to note as done well rather than just done:

**The test asserts the property, not the example.** `retainedBytes` computed across all retained args and compared to `MAX_TOTAL_BYTES` survives someone changing the payload size or the marker text — where an assertion on "logs.length === 1" alone would not.

**The changeset states the behaviour change plainly** — a script logging one very large payload now truncates *on* that call rather than after it, and the existing boundary test moved from four full entries to three plus a marker. That test moving is the correct signal here: it exists to pin where the boundary is, so it *should* move when the boundary deliberately moves. Explaining why in the changeset is what separates that from quietly rewriting a test to match new behaviour.

Also right to leave the 1000-entry cap alone with the reasoning given — it increments by exactly one per call, so its overshoot was always bounded to a single entry and never had the unbounded shape the byte budget did.

Issue #2042, opened 2026-08-03
Export building sections to scale as PDF (selectable scale)


louistrue:
### The problem

Engineers want to create a building section (`Gebäudeschnitt`) from an IFC in the ifc-lite viewer and export it **to scale as a PDF**, so the section can be used directly for standard/rule-of-thumb structural dimensioning (`Regelbemessungen`). The viewer already has a working section tool + 2D section panel (see screenshots), but there's no way to get a dimensionally accurate, scaled drawing out of it.

Today practitioners work around this by scaling screenshots of the section, which is tedious and introduces significant inaccuracy — a screenshot has no defined scale, so any measurement taken off the printout is unreliable. This is a recurring pain point coming from multiple engineers who miss this function in viewers.

### Proposed solution

Add a **scaled PDF export** for the 2D section view:

- Export the current section as a vector PDF at a chosen, exact scale so measurements taken off the printed sheet are correct (1:100 on paper means 1 cm = 1 m).
- **Scale selector** with common presets: `1:10`, `1:20`, `1:50`, `1:100`, `1:200`, `1:500`.
- Allow a **custom scale** input in addition to the presets.
- Default to **"as displayed"** (current on-screen framing), with the scaled options as alternatives. Being able to pin an explicit scale also makes it easy to compare multiple sections against each other on a consistent basis.

Ideally this reuses the existing `@ifc-lite/drawing-2d` vector pipeline (the same one that renders generated floor plans/sections) so the PDF is true vector output rather than a rasterized screenshot.

### Alternatives considered

- **Scaled screenshot export (current workaround):** no defined scale, inaccurate, manual — this is exactly the problem we're solving.
- **PNG/raster export at a fixed DPI:** better than a screenshot but still not a defined drawing scale and loses vector crispness when printed. Vector PDF at an explicit scale is the goal.

---

_Source: feature request from Pascal Furrer (engineering), with follow-up on scale selection._

louistrue:
Delivered by #2119 (merged): scaled vector PDF export for section drawings, with selectable scale.

Closing manually — the PR title carries `(#2042)` without a closing keyword, which is the same drift that left eight other issues open after their work had landed today.

Worth recording what the review turned up, because the feature shipped meaningfully different from how it started:

[...]

PR #2119, opened 2026-08-04
feat(viewer): scaled vector PDF export for section drawings (#2042)

BIMvoice:
## Summary

Closes the core ask of #2042 for the smallest useful version: a true-vector PDF export of the 2D section drawing at an **exact, user-chosen scale**, so measurements taken off the printed sheet are correct.

- New "Export PDF" toolbar button / menu item in the 2D section panel.
- Uses `jsPDF`, which is **already a viewer dependency** (`apps/viewer/package.json`, used today for tabular list exports) — no new dependency, so no licensing/bundle-size call needed from a maintainer for this PR.
- Scale: defaults to "as displayed" (current on-screen scale); a prompt also accepts any of the common presets (`1:10`…`1:1000`, from `@ifc-lite/drawing-2d`'s `COMMON_SCALES`) or a fully custom denominator.

## Key design decision — please confirm

The PDF page is **sized to the drawing extent + margin** at the exact requested scale, rather than fit into a fixed named paper (A3, A4, …). This is deliberate: the existing "sheet" fit-to-viewport transform (`packages/drawing-2d/src/sheet/sheet-types.ts` `calculateDrawingTransform`) silently shrinks the drawing (`fitScale = Math.min(scaleX, scaleY, 1)`) when it doesn't fit the configured viewport at the nominal scale. That's fine for an on-screen preview (the scale bar adjusts to stay graphically correct), but wrong for a document whose entire purpose is "1:100 means 1cm = 1m" — a silently-shrunk drawing on a page still labeled "1:100" would defeat the issue's own stated goal. The new export can never do that: if you ask for 1:50, you get 1:50, on a page exactly as large as it needs to be.

The pure scale/extent math is factored into `packages/drawing-2d/src/pdf-scale.ts` (`computePdfScaleLayout`, `worldPointToPdfMm`, `worldLengthToPdfMm`), unit tested independently of any rendering.

## v1 scope — smaller than the SVG export, on purpose

Per the issue's own suggestion ("smallest useful version first"), this PR draws only what an engineer actually measures off a printed section:

- Cut-polygon **outlines** (stroke only)
- Drawing **lines** (projection/hidden/silhouette/crease/boundary), respecting "show hidden lines"

**Deliberately deferred** (would roughly triple this diff to port from the SVG exporter, and none of it affects scale accuracy):
- Area fills / hatching patterns
- The drawing-sheet title block, frame, scale bar, north arrow
- DXF reference underlays
- Text/cloud annotations and measurements
- The point-cloud scan overlay

Happy to follow up on any/all of these once the scale-transform approach is confirmed as the right one.

## Scale/paper assumptions (please overrule if wrong)

- World units are assumed to be **metres** (matches every other exporter in this pipeline — SVG, DXF, the sheet transform).
- `worldToMm = 1000 / scaleFactor`, i.e. 1:100 → 1m = 10mm on paper (standard architectural convention, same formula already used in 3 other places in this codebase).
- Default margin: 10mm on all sides.
- Scale selector UI is a `window.prompt` in this PR, not a polished dropdown (matches `handlePrint`'s existing `alert`-based error surface in this hook, which has no toast wiring) — flagged as the most obvious immediate follow-up.

## Test plan

- [x] New unit tests for the pure scale/extent module: `packages/drawing-2d/src/pdf-scale.test.ts` (15 tests — mm-per-metre conversion at multiple scales, page sizing vs. margin, point-transform round-trip with Y-flip, and rejection of invalid scale/margin/bounds). RED-verified via a mutation test (`1000/scaleFactor` → `1000/scaleFactor + 1`); 6/15 tests failed as expected, then restored by file copy.
- [x] `packages/drawing-2d` full suite: 129/129 passing (baseline 114; +15 new).
- [x] `apps/viewer` full `node:test` suite: 2635/2635 passing, unaffected.
- [x] `pnpm typecheck` (repo-wide, via turbo): all 34 tasks pass.
- [x] `pnpm turbo lint --filter=@ifc-lite/drawing-2d --filter=@ifc-lite/viewer`: clean; `@ifc-lite/viewer` production build succeeds.
- [ ] **Not covered**: visual/manual verification that the generated PDF actually renders correctly and prints at the labeled scale in a real PDF viewer — no browser/visual-check tooling was available in this environment. The scale *arithmetic* is unit tested and mutation-verified; the jsPDF rendering calls (`doc.lines`, `doc.line`, dash patterns) are exercised only by typecheck/build, not by an automated visual assertion. Please sanity-check one exported PDF with a ruler before merging.

Changeset added for `@ifc-lite/drawing-2d` (minor — new exported functions). `@ifc-lite/viewer` is private, no changeset needed for it.

## Summary by CodeRabbit

* **New Features**
  * PDF export now supports customizable scale factors; users can select presets (1:100, etc.), enter custom denominators, or export at the displayed scale
  * Interactive scale selection prompt with input validation and cancel option
  * PDF export controls accessible from viewer toolbar and menu
  * PDFs render at exact scale with proper margins, preserving drawing layout


BIMvoice:
Followed up on my own note about this adding a fourth copy of the world-metres→paper-mm transform. **Verdict: do not consolidate** — and the investigation turned up something more useful than a refactor. Pushed as `103ed5a9`, test-only.

## Why not

They are four *contracts* that share one line of arithmetic (`worldToMm = 1000 / scaleFactor`), not one function copied four times:

| | re-fits? | page sizing | point transform |
|---|---|---|---|
| `SVGExporter.computeTransform` | no | fixed named paper, centred | bakes mm coords |
| `calculateDrawingTransform` | **yes** — `min(sx, sy, 1)` | fixed viewport in a sheet | bakes mm coords |
| inline in `generateExportSVG` | no | content-sized | **none** — defers to SVG `viewBox` vs `width`/`height` |
| `computePdfScaleLayout` (this PR) | no | content-sized | bakes mm coords for jsPDF |

The sharpest pair is `calculateDrawingTransform` vs this PR's: its silent shrink is **load-bearing** for its two real callers (`Drawing2DCanvas.tsx:739`, `generateSheetSVG`) and **wrong** for a measurable export. Merging them needs a `refit: boolean`, i.e. one function that silently does the wrong thing for whichever caller forgets the flag.

The pair I *expected* to be genuine duplicates — `computeTransform` and the inline one — also is not: the inline path never transforms a point at all, it emits raw world units and lets the SVG viewer resolve `viewBox` against `width`/`height` in mm. There is no equivalent step to extract.

## What was actually wrong

Two of the four had **zero tests**. So the divergence that must not be merged was also undetectable if it drifted. That is the real risk here, and it is now closed:

- `svg-exporter.test.ts` — centering and scale-linearity at 1:100 and 1:50.
- `sheet/sheet-types.test.ts` (new) — both fit branches, including the silent shrink firing and the never-enlarge clamp.

Mutations, both verified after the change: `1000 / scale.factor` → `+1` fails 3 tests; dropping the `1` from `min(sx, sy, 1)` fails 4, including the never-enlarge case. `packages/drawing-2d` 129 → 137, `tsc --noEmit` clean, no production file touched.

## One finding I did not fix — your call

`computeTransform`'s **`padding` parameter is dead**. `availableWidth`/`availableHeight` (`svg-exporter.ts:203-204`) are computed from it and then never used, so `SVGExportOptions.padding` — documented as "Padding around drawing in mm" — has no effect on `export()` or `exportPolygons()` output.

That is a published-package behaviour change to fix, and it is a real one (a caller setting padding today gets silently ignored), so I did not touch it inside a PR about PDF export. I pinned the *current* no-effect behaviour with a characterisation test, so whoever fixes it has to change that assertion deliberately rather than discover it by accident.

BIMvoice:
Independent check against the class of defect just found on the sibling PR (#2131: `SVGExporter` clamps the scale for padding but prints the *requested* scale in the title block, so a clamped sheet reads a scale that doesn't match its geometry).

**Verdict: this PR does not have that defect, for two independent reasons — it neither clamps nor prints a scale label yet.**

## 1. Does the PDF path clamp or re-fit?

No. `computePdfScaleLayout` (`packages/drawing-2d/src/pdf-scale.ts:55-101`) sizes the **page** to `drawingExtent * worldToMm + margin`, never the other way round — there is no `min(scaleX, scaleY, ...)` anywhere in the file, and the only bound-related throws are for invalid input (non-finite/negative scale, negative margin, inverted bounds), not a silent shrink. `handleExportPDF` (`apps/viewer/src/hooks/useDrawingExport.ts:908-991`) calls it with the exact `effectiveScale` the user chose (from the scale prompt in `Section2DPanel.tsx:489-508`, which accepts any positive number) and uses `layout.transform` unmodified for every point (`worldPointToPdfMm`) and stroke width. This matches the design note's claim that this path deliberately avoids `sheet-types.ts`'s `calculateDrawingTransform` / `fitScale = min(scaleX, scaleY, 1)` — confirmed by the characterisation tests already on this branch (`sheet/sheet-types.test.ts`, added in `103ed5a9`) that pin that *other* function's silent shrink, precisely to make the divergence visible instead of just asserted in a comment.

Mutation check I ran (not committed — restored by file copy, diff confirmed clean): replacing `worldToMm = 1000 / scaleFactor` with `Math.min(1000 / scaleFactor, 5)` in `pdf-scale.ts` fails 6/15 of `pdf-scale.test.ts` (the margin, offset, and both `worldLengthToPdfMm` scale-linearity tests). So a reintroduced clamp would be caught.

## 2. Does the PDF carry a scale label to go wrong?

No — and that's why this can't reproduce #2131's failure mode. `handleExportPDF`'s v1 scope comment (`useDrawingExport.ts:900-907`) is explicit: title block, frame, and scale bar are **not yet included**; the export is cut-polygon outlines + drawing lines only. The one place a scale number appears at all is the download filename (`section-...-1-${Math.round(effectiveScale)}.pdf`, line 984) — and that's `Math.round()` of the *same* `effectiveScale` variable that was just fed into `computePdfScaleLayout`, not a separately-sourced label. There's no second code path that could drift from the geometry; the failure mode #2131 depends on (label sourced from the nominal request, geometry sourced from the clamped result) structurally doesn't exist here yet. Worth flagging as a minor cosmetic note, not a defect: `Math.round` means a custom fractional scale like `1:133.33` shows as `...-1-133.pdf` in the filename — the printed geometry is still exactly 1:133.33, only the filename digit is rounded for display.

## 3. Is 1mm on the page actually 1mm?

Checked at two levels:
- `pdf-scale.test.ts` already pins the pure math: `worldToMm` at 1:100/1:50, margin placement (`bounds.min`/`bounds.max` land exactly `margin` mm inside the page edges — proof there's no re-fit), and `worldLengthToPdfMm` linearity across scales.
- I additionally drove `jsPDF` directly (`unit: 'mm', format: [120, 80]`, two `doc.line()` calls of known length) and parsed the raw, uncompressed PDF content stream for the emitted `m`/`l` operator coordinates. A 100mm horizontal line and a 60mm vertical line converted back from PDF points (÷2.834645669291339, i.e. 72/25.4) landed at exactly 100.0 and 60.0mm, and the `/MediaBox` matched the requested 120×80mm page. So jsPDF's `unit: 'mm'` really does treat every coordinate passed to `doc.line`/`doc.lines`/the page `format` array as millimetres, consistently on both axes — the same units `computePdfScaleLayout` already outputs. (I did not open a rendered PDF in a viewer; this is byte-level inspection of the content stream, not a visual check.)

## 4. Does an over-large margin behave?

`marginMm` is not exposed to the user at all in this PR — `handleExportPDF` hardcodes it to `10` (`useDrawingExport.ts:914`). So #2131's second finding (a user-set impossible padding silently disabling the guarantee) has no attack surface here yet. For what it's worth, the function itself would behave correctly at any margin regardless: since the page is sized *around* the margin rather than the margin carved *out of* a fixed page, there's no "impossible" margin — arbitrarily large `marginMm` just produces a larger blank border, never touches the drawing's scale or position relative to itself.

## Where the brief was right / what I'd add

The brief's premise held throughout — this actually is a clean result. Nothing needed fixing, so nothing was changed on this branch (working tree diff is empty, `103ed5a9` was already here from an earlier investigation of a different but related question — whether the four world→paper-mm transforms in the codebase should be merged). `packages/drawing-2d` test suite: 15/15 in `pdf-scale.test.ts`, unchanged.

louistrue:
Reviewed all seven threads. **One is a showstopper for the feature as written, and I verified it arithmetically.**

## Front and side sections render off-page for ordinary models

greptile (`:924`) and CodeRabbit (`:926`) independently flagged this and both are right. `computePdfScaleLayout` is given `drawing.bounds`, but `mapPoint` negates the coordinates *after* the layout offsets were derived from the un-negated bounds:

```ts
layout = computePdfScaleLayout(drawing.bounds, effectiveScale, 10);
…
const mapPoint = (x, y) => worldPointToPdfMm({ x: flipX ? -x : x, y: flipY ? -y : y }, layout.transform);
```

`offsetXMm = marginMm - min.x * worldToMm` only lands the drawing on the page for **un-flipped** coordinates. Running the real arithmetic (margin 10 mm, 1:100):

```
down  (no flip),   asymmetric bounds   page 220x170  x∈[10,210]     OK        y∈[10,160]    OK
side  (flipX+Y),   SYMMETRIC bounds    page 220x220  x∈[10,210]     OK        y∈[10,210]    OK
side  (flipX+Y),   asymmetric bounds   page 220x170  x∈[-390,-190]  OFF-PAGE  y∈[260,410]   OFF-PAGE
front (flipY),     asymmetric bounds   page 220x170  x∈[10,210]     OK        y∈[260,410]   OFF-PAGE
```

It works **only** when the bounds happen to be symmetric about zero — and building coordinates almost never are; a section of a model sitting at positive world coordinates is the normal case. So **every front and side PDF export produces a blank page** for a typical model. Plan (`down`) is unaffected, which is presumably why this survived manual testing.

The fix is to derive the layout from the bounds *as they will be drawn*: flip the bounds first, then compute offsets from those. CodeRabbit's phrasing — "calculate the layout from the axis-adjusted bounds" — is exactly right.

Worth a test per axis with deliberately asymmetric bounds. A symmetric fixture passes against the bug, which is the trap this already fell into.

## The filename is the only scale record, and it rounds

```ts
const stem = `section-${axis}-${position}-1-${Math.round(effectiveScale)}`;
```

A 1:99.5 export is filed as `…-1-100`. Version 1 has no title block, so that filename is the sole record of what scale the sheet was drawn at.

This is the same defect class as #2131, which you just fixed: a drawing that misreports its own scale. There the title block lied; here the filename does. Given you already built `scaleLabel`-style formatting for #2131 (round to 2dp, strip trailing zeros), the same treatment applies — and I would reuse it rather than re-derive.

## The async IIFE has no rejection handler

`void (async () => { … })()` around the `jspdf` dynamic import, PDF construction and download (`:991`). A failed chunk load — the single most likely failure here, and the one #2124's sibling issues are about — surfaces as an unhandled rejection, not as feedback to the user. The synchronous path above it already `alert()`s on an invalid scale; the async path should be at least as loud.

## The remaining three

- **Cut-line fallback discarded** (`:955`) — dropped chains silently vanish from the sheet. Same "silent partial output" shape; worth handling or documenting.
- **Non-finite derived dimensions** (`pdf-scale.ts:92`) — finite inputs can still yield infinite `worldToMm`. Your guard checks the inputs; it should check the outputs, since that is what reaches jsPDF.
- **Public API docs** (`index.ts:181`) — minor, but these are new exports on a published package.

## Summary

The off-page bug is the one that has to be fixed before this merges — as it stands the feature works for plan views and silently produces blank pages for the other two axes. The rest are smaller, and the filename rounding is worth doing while the #2131 reasoning is still fresh.

BIMvoice:
Pushed a fix for the showstopper plus the three smaller findings (`0c2352d5`, on top of a merge of `upstream/main` to pick up #2131's scale-label formatting).

**Showstopper — off-page front/side exports.** Confirmed your arithmetic: `computePdfScaleLayout`'s offsets were derived from `bounds.min`/`bounds.max` *before* the caller's per-axis negation, so they only lined up with the page when bounds happened to be symmetric about zero. Fix derives the layout from the bounds **as actually drawn** — flip first (`flipBounds2D`, new in `pdf-scale.ts`), then compute offsets from the flipped bounds. Extracted the axis-flip + layout call into `apps/viewer/src/hooks/pdfSectionLayout.ts` (`computePdfSectionLayout` / `makeSectionMapPoint`) so it's unit-testable without React/jsPDF (matches the pattern `svgExportViewport.ts` already used for the 4th of these "world-mm" transforms).

RED-first, with deliberately **asymmetric** bounds (`x∈[100,110], y∈[50,80]`) — a symmetric fixture would pass against the bug, which is exactly the trap this fell into:
- Reproduced the bug in the extracted module (layout derived from un-flipped bounds) → `down` passed, `front`/`side` failed with corners hundreds of mm off-page (e.g. `axis=front: mapped y=1310 is outside [0, 320]`).
- Applied the fix (file-copy restore, not `git checkout`) → all three axes pass; every mapped corner lands within `[0, pageWidth] x [0, pageHeight]`.

Note: I can't render the PDF here, so this is verified arithmetically (mapped corner coordinates against the page rectangle) and via the same by-hand computation you did, not visually.

**Finding 2 — filename scale rounding.** Extracted `formatScaleFactorLabel` (round to 2dp, strip trailing zeros) out of `SVGExporter`'s private `scaleLabel()` (#2131) into `pdf-scale.ts`, exported it, and both the title block and the PDF filename now call the same function. `1:99.5` files as `…-1-99.5`, not `…-1-100`.

**Finding 3 — unhandled rejection.** Wrapped the async IIFE body (dynamic `jspdf` import, PDF construction, download) in try/catch; a failed chunk load now `alert()`s like the synchronous scale-validation path above it, instead of surfacing only as an unhandled promise rejection.

**Finding 4 — non-finite derived dimensions.** `computePdfScaleLayout` now validates `worldToMm`, `page.widthMm/heightMm`, and both offsets after computing them, not just the raw inputs — a vanishingly small (but finite, positive) scale factor or very large bounds can still overflow to `Infinity` through the arithmetic even though every input passed the existing input checks.

**Finding 5 — cut-line fallback (fixed, not just documented).** Traced it: `cutPolygons` and cut-category `drawing.lines` both come from the same `cutSegments`, but `cutPolygons` goes through `PolygonBuilder.buildLoops`, a separate, fallible loop-closing reconstruction that can drop an entity's polygon entirely for a short/degenerate/ambiguous cross-section while the raw cut lines for that entity are still valid. Unconditionally skipping `category === 'cut'` lines could silently lose that geometry. Now the skip is scoped: a cut-category line is only skipped when its `(modelIndex, entityId)` actually has a `cutPolygons` entry; otherwise it's drawn (stroke width matched to the polygon outline's 0.5mm).

**Tests / mutation:**
- `packages/drawing-2d`: 142 → 153 tests (vitest).
- `apps/viewer`: 2739 → 2742 tests (node:test).
- Mutation-tested (python3 exact-substring replace, occurrence count asserted to stderr, restored by file copy — never `git checkout --`): the new output-finiteness guard, `flipBounds2D`, and re-ran your existing `1000 / scaleFactor` mutation on `pdf-scale.ts` — still killed (6/153 tests fail).
- `pnpm typecheck` clean across all 34 workspace packages.

Full diff: `apps/viewer/src/hooks/{useDrawingExport.ts,pdfSectionLayout.ts,pdfSectionLayout.test.ts}`, `packages/drawing-2d/src/{pdf-scale.ts,pdf-scale.test.ts,svg-exporter.ts,index.ts}`.

louistrue:
**The off-page bug is fixed, and I verified it the same way I found it.** Every axis, every bounds placement, lands on the page:

```
--- asymmetric bounds min(10,5) max(30,20) — the failing case ---
down   page 220x170 | x∈[10.0,210.0] OK | y∈[10.0,160.0] OK
front  page 220x170 | x∈[10.0,210.0] OK | y∈[10.0,160.0] OK     (was y∈[260,410] OFF-PAGE)
side   page 220x170 | x∈[10.0,210.0] OK | y∈[10.0,160.0] OK     (was x∈[-390,-190] OFF-PAGE)

--- symmetric (control) and all-negative bounds: all OK on all three axes ---
```

Deriving the layout from `flipBounds2D(bounds, flip)` is the right fix, and extracting `computePdfSectionLayout` / `makeSectionMapPoint` into `pdfSectionLayout.ts` is what makes it testable without rendering a PDF. Deriving **both** the layout and the mapper from one `axisFlipForSection(axis)` is the part that keeps them from drifting apart again — the original bug was exactly those two disagreeing.

Your test names it per axis with asymmetric bounds, which is the fixture that would have caught it originally; a symmetric one passes against the bug.

Two of the other three are also done:

- **async failures** now `alert()` on the same path as the synchronous scale validation, so a failed `jspdf` import is no longer an unhandled rejection
- **the filename** uses `formatScaleFactorLabel(effectiveScale)` instead of `Math.round`, reusing the #2131 formatting rather than re-deriving it — so a 1:99.5 export is no longer filed as `1-100`

## One left: non-finite derived dimensions

`computePdfScaleLayout` validates its **inputs** thoroughly — `scaleFactor` finite and positive, `marginMm` finite and non-negative, bounds finite, `max >= min`. It does not validate what those produce:

```
scaleFactor = Number.MIN_VALUE  ->  worldToMm = Infinity   (finite input)
min.x = -1e308, max.x = 1e308   ->  width     = Infinity   (both finite)
```

Both pass every current guard and hand `Infinity` to jsPDF as a page dimension. A check on the derived `page.widthMm` / `page.heightMm` (and `worldToMm`) before returning would close it, and it fits the pattern already in the function — throw with a message naming the value, so the caller's `alert` says something useful.

Everything else on this PR is verified. Fix that and I will merge.

BIMvoice:
That check is already in `0c2352d5` — the same commit you verified the other three items in. I think it just got missed on the read, so here is the evidence rather than a claim.

**The guard** — `packages/drawing-2d/src/pdf-scale.ts:107-117`, immediately before the `return`:

```ts
if (
  !Number.isFinite(worldToMm) ||
  !Number.isFinite(page.widthMm) || !Number.isFinite(page.heightMm) ||
  !Number.isFinite(offsetXMm) || !Number.isFinite(offsetYMm)
) {
  throw new Error(
    `PDF export layout produced non-finite dimensions (worldToMm=${worldToMm}, ` +
      `page=${page.widthMm}x${page.heightMm}mm) from scaleFactor=${scaleFactor} and bounds ` +
      `${width}x${height}. Choose a less extreme scale.`
  );
}
```

It covers what you asked for — `worldToMm` and both page dimensions — plus `offsetXMm`/`offsetYMm`, since those reach jsPDF as coordinates on the same path. The message names the offending values, so the caller's `alert` says something useful.

**Your two exact cases, executed against the branch tip:**

```
THREW  scaleFactor = Number.MIN_VALUE
       -> PDF export layout produced non-finite dimensions (worldToMm=Infinity,
          page=InfinityxInfinitymm) from scaleFactor=5e-324 and bounds 10x6.

THREW  min.x = -1e308, max.x = 1e308
       -> PDF export layout produced non-finite dimensions (worldToMm=10,
          page=InfinityxInfinitymm) from scaleFactor=100 and bounds InfinityxInfinity.
```

The second is the more interesting one and matches your framing exactly: `worldToMm` is a perfectly ordinary `10`, and both bounds inputs are finite — it is only the derived `width` that overflows. An input-only check cannot see it.

**It is pinned, not just present.** `packages/drawing-2d/src/pdf-scale.test.ts:97-115` has two tests for it, and they die when the guard does:

```
### guard condition -> `if (false)`   (replacement count = 1)
     × rejects a scale factor so small that worldToMm (1000/scaleFactor) overflows to Infinity
     × rejects bounds so large that page.widthMm/heightMm overflow to Infinity even though worldToMm itself is finite
      Tests  2 failed | 24 passed (26)
```

Restored: 26/26.

Each test also asserts the *input* is finite before asserting the throw (`expect(Number.isFinite(tinyScale)).toBe(true)`), so it cannot pass by accidentally tripping the earlier input guard instead — the failure has to come from the derived value.

Nothing to push. If you would rather the guard also cover a case I have not thought of, say which and I will add it.

louistrue:
Last item verified. Removing the derived-output guard entirely kills both new tests:

```
### remove the whole non-finite guard
       × rejects a scale factor so small that worldToMm (1000/scaleFactor) overflows to Infinity
       × rejects bounds so large that page.widthMm/heightMm overflow to Infinity even though worldToMm itself is finite
      Tests  2 failed | 24 passed (26)
```

Restored: 26/26.

The second test is the one that earns its place. `worldToMm` finite while `page.widthMm` overflows is a genuinely different failure than the obvious `Number.MIN_VALUE` case, and a guard that only checked `worldToMm` would pass the first test and ship the second bug. Asserting the *premise* inside each test — `expect(Number.isFinite(worldToMmAtThisScale)).toBe(true)` — is what stops it silently degenerating into a duplicate of the first if someone retunes the constants.

**A correction on my own verification.** My first attempt at this mutation replaced only three of the five clauses (`worldToMm`, `page.widthMm`, `page.heightMm`) with `false ||`, leaving the `offsetXMm`/`offsetYMm` checks — and since `offsetXMm` derives from `worldToMm`, those still caught both cases. I got `26/26 passed` and briefly took it as "the guard is unpinned". It was my mutation that was too narrow, not your tests that were too weak. Removing the whole condition is what actually discriminates.

## This one is done

Recapping, because it moved a long way:

- **the showstopper** — front/side exports rendered off-page for any model at ordinary world coordinates. Verified fixed on all three axes across asymmetric, symmetric and all-negative bounds, with layout and mapper now both derived from one `axisFlipForSection(axis)` so they cannot drift apart again
- **the filename** no longer rounds a 1:99.5 export to `1-100`, reusing #2131's `formatScaleFactorLabel`
- **the async path** alerts instead of producing an unhandled rejection
- **derived dimensions** can no longer reach jsPDF as `Infinity`

Merging.

Issue #2121, opened 2026-08-04
Section 2D annotation overlay ignores the ifcAnnotations class toggle


louistrue:
## Summary

Turning off **IFC Annotations** in the class Visibility menu hides them in the 3D viewport but leaves them drawn on the 2D section.

`Section2DPanel.tsx` gates the symbolic-annotation overlay on the panel's own display option only:

```ts
const ifcAnnotationData = useSymbolicAnnotationsForDrawing({
  enabled: displayOptions.showIfcAnnotations && status === 'ready',
  ...
});
```

No `typeVisibility`. The hook cannot see it either — its parameter object has no such field.

## Why this is a mismatch rather than a design choice

`typeVisibilityFilter.ts` states the intent explicitly:

```ts
// IfcAnnotation can carry real 3D solid geometry (Bonsai plan-view boxes,
// Revit "Model Text" breps) on top of the 2D symbolic curve overlay; the
// `ifcAnnotations` toggle hides both (issues #1354, #1480).
IfcAnnotation: 'ifcAnnotations',
```

"hides both" — and the section's symbolic overlay is exactly the 2D curve overlay named there.

#2070 (merged) threaded `typeVisibility` into `useDrawingGeneration` so the section honours class toggles, with the comment:

> Class-level Visibility toggles — the section honours them like the 3D viewport does, so a hidden IfcSpace/IfcOpeningElement is not cut (#2060).

That is now true for every class route except this one, so the comment overstates coverage. Before #2060/#2070 the section honoured no class toggles at all, so annotations persisting was at least consistent; now it is the lone exception.

## Suggested fix

AND the class toggle into the `enabled` expression:

```ts
enabled: displayOptions.showIfcAnnotations
  && typeVisibility.ifcAnnotations
  && status === 'ready',
```

`typeVisibility` is already read in that component (`Section2DPanel.tsx`, for the `useDrawingGeneration` call), so nothing new needs plumbing.

Worth a test alongside it: the drawing's annotation set is empty when `typeVisibility.ifcAnnotations` is false but `displayOptions.showIfcAnnotations` is true, so the two toggles are pinned as independent rather than one standing in for the other.

## Scope

Display-only, no data at risk — which is why it was split out of #2070 rather than holding a verified geometry-gating fix. Found while reviewing #2070; CodeRabbit flagged the same route independently.

PR #2122, opened 2026-08-04
test(geometry): make the sub-unit/nonuniform instancing loop real (#2064 follow-up)

BIMvoice:
Follow-up to #2064, which merged at 13:44 while this was being worked. Rebased onto current `main` and opened separately rather than pushed to a merged branch.

Closes the finding @louistrue raised on #2064: `for instancing in [false, true]` in `sub_unit_scales_nonuniform_and_nested_maps_match_direct_metre_authoring` reads as a coverage matrix it cannot deliver — his probe showed `instances=0` on both iterations, so the second re-runs the first.

## Chose option (2), and the router says why

He offered two remedies and said we were better placed to judge: drop the loop, or give the fixture a shared map so the instanced path is real. The deciding question was whether sub-unit scaling and instancing *can* interact — and they can.

`router/instancing.rs::mapped_source_single_item` and the `instance_solid_id`/`nontrivial_target` composition in `router/processing.rs` take the don't-bake path whenever **two or more top-level `IfcMappedItem`s share one `IfcRepresentationMap`, even with different `MappingTarget`s** — the code documents it directly: *"occurrences sharing a map but differing by target collate under one template."*

That is exactly the sub-unit / non-uniform scale composition this test pins, so the instanced path is reachable for it and deserved real coverage. Dropping the loop would have been the cheaper answer to the wrong question.

## The fixture was starving instancing, not merely failing to use it

Worth stating because it changes what the old loop was doing. The nested-map occurrence previously reused `#13`, so `#13`'s occurrence count included a nested `IfcMappedItem` with a lower express id, which was miscounted as the don't-bake template. Both real occurrences were starved and everything routed through orphan recovery — correct output, instancing never engaged, no signal that anything was off.

Now: occurrences A (uniform `0.001`) and B (non-uniform `0.001/0.001/0.002`) share `#13`; the nested-map occurrence gets its own source (`#205`).

```
before:  instancing=false meshes=4 instances=0
         instancing=true  meshes=4 instances=0
after:   instancing=true  meshes=3 instances=1   (template_express_id == 74)
```

The test now branches like `mapping_origin_is_composed_inside_the_mapping_target` already did — materialized assertions when `!instancing`, and for `instancing=true` it reconstructs occurrence B's world geometry from its `InstanceRecord` transform against the A template.

## Evidence

Mutating the router's non-template composition to identity (`local_rm = mapping_transform.map(|_t| IDENTITY_ROW_MAJOR)`; single replacement asserted, region re-read, restored by file copy):

```
before this change:  sub_unit test PASSES  — the instanced branch never ran
after:               2 failed  (sub_unit + mapping_origin)
```

That is the whole point: the sub-unit test could not previously have caught a broken instanced transform, and now does. `mapping_origin` failing alongside is expected — it always covered that path.

`cargo test -p ifc-lite-processing` green; `cargo clippy -p ifc-lite-core -p ifc-lite-processing --all-targets -- -D warnings` clean on this base.

**A note on clippy**, since it was reported as failing during the work: those errors came from the old branch being 30 commits behind, not from this change. On current `main` the same crates lint clean, which is why this is rebased rather than merged.

No changeset — test and fixture only, no published behaviour change.

### DIFF
--- a/rust/geometry/tests/fixtures/issue_1985_metre_submm_scale.ifc
+++ b/rust/geometry/tests/fixtures/issue_1985_metre_submm_scale.ifc
@@ -19,9 +19,27 @@ DATA;
 #11=IFCEXTRUDEDAREASOLID(#8,#10,#9,1734.5);
 #12=IFCSHAPEREPRESENTATION(#5,'Body','SweptSolid',(#11));
 #13=IFCREPRESENTATIONMAP(#2,#12);
-/* ---- Map B: wraps map A through a NESTED IfcMappedItem (inner Scale=0.001) ---- */
+/* ---- Map A2: a SEPARATE copy of map A's solid, used only by the nested wrap
+   below. The router's occurrence-count plan (#1623 Phase 2) tallies every
+   IfcMappedItem referencing a source, including one nested inside another
+   map's representation -- so if the nested wrap pointed at #13 itself, its
+   IfcMappedItem (#15) would inflate #13's count to 3 and (being the
+   lowest express id) get PICKED as the don't-bake template, even though it
+   never runs through the top-level template/retag path (only depth == 0
+   items do, see router/processing.rs). That starves both top-level
+   occurrences of #13 of a real template and routes them through orphan
+   recovery instead -- silently correct output, but instancing never
+   actually engages. Giving the nested wrap its own source keeps #13's count
+   at exactly 2 (occurrences A and B below), so instancing engages for real. ---- */
+#200=IFCCARTESIANPOINT((0.,0.));
+#201=IFCAXIS2PLACEMENT2D(#200,$);
+#202=IFCCIRCLEPROFILEDEF(.AREA.,'P',#201,50.);
+#203=IFCEXTRUDEDAREASOLID(#202,#10,#9,1734.5);
+#204=IFCSHAPEREPRESENTATION(#5,'Body','SweptSolid',(#203));
+#205=IFCREPRESENTATIONMAP(#2,#204);
+/* ---- Map B: wraps map A2 through a NESTED IfcMappedItem (inner Scale=0.001) ---- */
 #14=IFCCARTESIANTRANSFORMATIONOPERATOR3D($,$,#1,0.001,$);
-#15=IFCMAPPEDITEM(#13,#14);
+#15=IFCMAPPEDITEM(#205,#14);
 #16=IFCSHAPEREPRESENTATION(#5,'Body','MappedRepresentation',(#15));
 #17=IFCREPRESENTATIONMAP(#2,#16);
 /* ---- Operators ---- */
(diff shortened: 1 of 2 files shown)


louistrue:
Question about your draft convention, because I think we may be deadlocked on it.

You have 11 open drafts right now (#2106, #2107, #2108, #2109, #2111, #2113, #2114, #2115, #2118, #2119, #2122). Every one of them is **green with all five core CI lanes passing**, and most carry a changeset and test files. Meanwhile every PR I have merged today was opened ready-for-review, and none of the drafts has ever been converted — so if you are waiting for a signal from us to undraft, and we are treating draft as "still being worked on", nothing moves.

**Which is it?**

- If draft means *not ready, do not look yet* — fine, I will leave them alone and you flip them when you want eyes on them.
- If draft means *done, queued, tell me when you have capacity* — say so and I will start pulling from that queue directly. **You do not need to wait for us; open ready and we will review.**

For context on what bar each PR now faces, since it changes how much scrutiny to expect and where it is worth pre-empting review:

- **`rust/**` and the kernel** — the highest bar. Mutation testing, negative controls, `cargo test --no-fail-fast`, clippy. Never merged on green CI alone.
- **Published packages** (`geometry`, `export`, `parser`, `data`, `mutations`, `sandbox`) — CI green with core lanes actually present, changeset correctness, and verification of the central claim.
- **`apps/viewer`** — CI green, core lanes present, threads resolved, merged. Display-level inconsistencies become follow-up issues rather than holds (as I just did with #2121 off #2070).

One practical warning that has bitten three of your PRs today: **a `CONFLICTING` branch stops GitHub scheduling `pull_request` workflows entirely**, so the PR page shows a green rollup with every real lane simply *absent* rather than failing. #2080, #2091 and now #2105 all hit it. If you see a suspiciously small check count, merge `origin/main` first — main is moving fast today.

Answer whichever way is true and I will work to it.

BIMvoice:
Answered on #2118 to keep it in one place: **option two — done and queued.** All twelve open PRs are now ready for review. The draft flag was a habit on my `gh pr create` calls, not a signal; nothing was waiting on you. Pull from the queue freely.

This one is `rust/**`, so it faces your highest bar — the evidence is in the PR body, and the part that matters is that mutating the router's non-template composition to identity now fails the sub-unit test, which it could **not** have done before this change (the fixture yielded `instances=0`, so that branch never executed).

louistrue:
Verified at the Rust bar — the loop is now real. Merging.

```
PROBE instancing=false meshes=4 instances=0
PROBE instancing=true  meshes=3 instances=1
```

Previously both legs were `meshes=4 instances=0`, i.e. the same run twice. Giving the fixture a shared map so #81 rides as an instance off the #74 template is the harder of the two options I offered and the right one: sub-unit scale composition under instancing is exactly where a transform can cancel and leave an occurrence unpinned, and that is now covered rather than asserted.

The docblock explaining *why* B's `local_transform` under instancing is what the composition depends on is the part that keeps this from rotting back into a duplicate loop.

3 tests pass, `cargo test -p ifc-lite-processing --no-fail-fast` clean, CI green on all 5 core lanes.

Worth noting this now runs under a gate that actually covers the WASM crate too — #2108 landed a few minutes ago, removing the `--exclude ifc-lite-wasm` from both clippy and `cargo test`. I confirmed that exclusion was hiding real violations: an injected `ptr_arg` in `ifc-lite-wasm` is caught by the new command (exit 101) and missed entirely by the old one (exit 0).

PR #2123, opened 2026-08-04
fix(viewer): gate the section's symbolic annotation overlay by the IfcAnnotation class toggle

### LINKED ISSUE #2121: Section 2D annotation overlay ignores the ifcAnnotations class toggle

BIMvoice:
## Summary

Closes #2121.

`Section2DPanel`'s class-level Visibility toggles honour every route into the 2D drawing — the cut mesh filter and the construction-projection profiles both read `typeVisibility` via `isTypeVisible` (#2060, #2070) — except the symbolic IfcAnnotation overlay. Its `enabled` expression only read `displayOptions.showIfcAnnotations` and `status`, never `typeVisibility`, so turning **IFC Annotations** off in the class Visibility menu hid the geometry in the 3D viewport but left the symbolic curve overlay drawn on the section.

## What changed

- `apps/viewer/src/hooks/useSymbolicAnnotations.ts`: added `symbolicAnnotationsOverlayEnabled(showIfcAnnotations, drawingStatus, ifcAnnotationsClassVisible)`, factored out of the call-site `&&` chain.
- `apps/viewer/src/components/viewer/Section2DPanel.tsx`: the `useSymbolicAnnotationsForDrawing` call now passes `enabled: symbolicAnnotationsOverlayEnabled(displayOptions.showIfcAnnotations, status, typeVisibility.ifcAnnotations)`.

The predicate is a separate exported function rather than an inline expression because `Section2DPanel` imports `useIfc` → `ifcConfig.ts` → `import.meta.env`, which `tsx` (this repo's `apps/viewer` test runner) does not populate outside a Vite build — the component cannot be rendered under `tsx --test` today (verified: no existing test file imports `Section2DPanel`). Extracting the gate gives the fix a real, importable, unit-testable surface without touching that pre-existing limitation.

## Testing

- `useSymbolicAnnotations.test.ts`: truth-table unit tests for `symbolicAnnotationsOverlayEnabled` (4 cases — class off, class+overlay+ready on, overlay off, not ready).
- `useSymbolicAnnotationsForDrawing.classGate.test.tsx`: end-to-end proof through the real hook, a real IFC source (real `IfcParser().parseColumnar`) and the real WASM symbolic-representation parser — a positive control (class visible → annotation text is produced) and the negative case the issue is about (class hidden → no annotation data), so the negative case can't pass vacuously.
- Mutation-tested (exact-substring replace, count asserted, restored by file copy): dropping the class-visibility clause from the predicate takes the suite from 0 → exactly 2 failures (the two new tests) and nothing else.
- Re-ran the two mutations already verified on #2070's profile-visibility filter to confirm this change doesn't disturb that coverage: removing the filter still fails exactly 1 test; replacing it with a blanket drop still fails exactly 2 (its own built-in negative control).
- `2693 passing` (was `2687`), `pnpm typecheck` 34/34 green.

No changeset: `@ifc-lite/viewer` is `"private": true`.

## Test plan

- [x] `npx tsx --test` on the two new/modified test files, RED before the fix / GREEN after
- [x] `npm test` in `apps/viewer` — 2693 passing, 0 failing
- [x] `pnpm typecheck` — 34/34

### DIFF
--- a/apps/viewer/src/components/viewer/Section2DPanel.tsx
+++ b/apps/viewer/src/components/viewer/Section2DPanel.tsx
@@ -40,7 +40,7 @@ import { useMeasure2D } from '@/hooks/useMeasure2D';
 import { useAnnotation2D } from '@/hooks/useAnnotation2D';
 import { useViewControls } from '@/hooks/useViewControls';
 import { useDrawingExport } from '@/hooks/useDrawingExport';
-import { useSymbolicAnnotationsForDrawing } from '@/hooks/useSymbolicAnnotations';
+import { useSymbolicAnnotationsForDrawing, symbolicAnnotationsOverlayEnabled } from '@/hooks/useSymbolicAnnotations';
 import { useDxfUnderlaysForDrawing, useDxfMapToWorldTransform, dxfWorldShift, dxfUnderlayDrawingBounds } from '@/hooks/useDxfUnderlay';
 import { useScanSectionLayer } from '@/hooks/useScanSectionLayer';
 
@@ -330,7 +330,7 @@ export function Section2DPanel({
   }, [geometryResult, sectionPlane.axis, sectionPlane.position]);
 
   const ifcAnnotationData = useSymbolicAnnotationsForDrawing({
-    enabled: displayOptions.showIfcAnnotations && status === 'ready',
+    enabled: symbolicAnnotationsOverlayEnabled(displayOptions.showIfcAnnotations, status, typeVisibility.ifcAnnotations),
     axis: sectionPlane.axis,
     sectionPosWorld: ifcAnnotationsForDrawing.sectionPosWorld,
     viewDepth: ifcAnnotationsForDrawing.viewDepth,
--- a/apps/viewer/src/hooks/useSymbolicAnnotations.test.ts
+++ b/apps/viewer/src/hooks/useSymbolicAnnotations.test.ts
@@ -13,7 +13,37 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
 import type { DrawingLine2D } from '@ifc-lite/renderer';
-import { polylineToSegments, circleToSegments, liftTo3DLineList } from './useSymbolicAnnotations.js';
+import {
+  polylineToSegments,
+  circleToSegments,
+  liftTo3DLineList,
+  symbolicAnnotationsOverlayEnabled,
+} from './useSymbolicAnnotations.js';
+
+describe('symbolicAnnotationsOverlayEnabled (issue #2121)', () => {
+  // `Section2DPanel`'s class-level Visibility toggles claim to gate the
+  // section "like the 3D viewport does" (#2060). The mesh cut and the
+  // construction-projection profiles both honour `typeVisibility` via
+  // `isTypeVisible`; the symbolic IfcAnnotation overlay used to be the one
+  // route that ignored it — this predicate is the fix, and these cases pin
+  // its truth table.
+
+  it('is OFF when the IfcAnnotation class-visibility toggle is off, even with the overlay and status both on', () => {
+    assert.equal(symbolicAnnotationsOverlayEnabled(true, 'ready', false), false);
+  });
+
+  it('is ON when the overlay is on, the drawing is ready, and the class is visible', () => {
+    assert.equal(symbolicAnnotationsOverlayEnabled(true, 'ready', true), true);
+  });
+
+  it('is OFF when the per-drawing overlay toggle is off, independent of the class toggle', () => {
+    assert.equal(symbolicAnnotationsOverlayEnabled(false, 'ready', true), false);
+  });
+
+  it('is OFF when the drawing is not ready, independent of the class toggle', () => {
+    assert.equal(symbolicAnnotationsOverlayEnabled(true, 'generating', true), false);
+  });
+});
 
 describe('polylineToSegments', () => {
   it('emits N-1 segments for an open polyline', () => {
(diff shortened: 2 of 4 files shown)


BIMvoice:
Verified independently, and **refuting one premise that has now travelled through three of your reviews** — with a probe, since it is your claim and mine as much as anyone's.

## The fix

Mutating the class clause out of `symbolicAnnotationsOverlayEnabled` fails exactly 2 tests (the pure predicate and the real end-to-end one); restored, `apps/viewer` is 2693 pass / 0 fail / 2 pre-existing skips.

Your two earlier #2070 mutations are undisturbed: removing the profile filter still fails 1, blanket-dropping everything still fails 2.

One thing that undersold the ask in my own brief: "two lines on the `enabled` expression" is right for the production diff, but `Section2DPanel.tsx` imports `useIfc → ifcConfig.ts → import.meta.env`, which `tsx` never populates — so the component **cannot be rendered under this repo's test runner at all**, and no existing test file imports it. A RED-first test therefore needed the gate extracted into an importable predicate. The call site really is two lines; the testability was the work.

## `apps/viewer` does NOT run every test file in one process

This premise has appeared in three places today — your `serveFileUrlsFromDisk` finding on #2070, your `unwritableKeys` finding on #2089, and my relay of both. It is wrong, and `setup-dom.ts:14` already says so:

> runner executes each test file in its own child process

Probe, two files, `tsx --test --test-concurrency=1`, first sets a global and the second reads it:

```
A pid=24423
B pid=24425 saw=undefined
```

Different PIDs, and the value does not survive. Node's test runner spawns a process per file, so **module-level state cannot leak across files**.

What follows:

- The `serveFileUrlsFromDisk` leak is **not an active bug**. The patched `fetch` and the deleted `WebAssembly.instantiateStreaming` die with the file's process; no later file inherits either. The restore is still in the new test file as cheap insurance and because it is correct hygiene, but it fixes nothing that was broken.
- The `after()` hook on `persistence.unwritableKeys` (#2089, merged) is likewise not fixing a cross-file latch. **It is still worth having** — module state does carry between tests *within* one file, so the latch could affect a later test in the same file. The mechanism is real one scope down; the paragraph justifying it named the wrong scope.

Neither changes any merge decision, and no shipped behaviour depended on it. Flagging it because it is the shape we have both been hunting all day — a plausible mechanism, asserted confidently, that nothing had actually measured. I asserted it too, on your word, without checking.

`pnpm typecheck` 34/34. No changeset (`@ifc-lite/viewer` is private). The parity comment at `Section2DPanel.tsx:76-77` is now true without edit — every `typeVisibility`-mapped class is gated in the section.

louistrue:
This closes #2121 — merging under the viewer tier (green on all 5 core lanes, `MERGEABLE`, 0 unresolved).

Extracting the predicate into `symbolicAnnotationsOverlayEnabled(showIfcAnnotations, status, ifcAnnotationsClassVisible)` is better than inlining the `&&`. It gives the rule a name, a place for the docblock, and something testable without rendering the panel — and the harness taking `ifcAnnotationsClassVisible` as a parameter means both states get exercised rather than just the newly-added one.

The comment tying it back to how the cut and the projection profiles already reach `typeVisibility` via `isTypeVisible` is what makes the section's three routes legible as one rule instead of three coincidences. That was the actual problem — not the missing `&&`, but that nothing said the routes were supposed to agree.

Turnaround on this was about twenty minutes from me filing the issue, which is why I split it out of #2070 rather than holding that PR: the fix landing separately cost nothing and unblocked the geometry gating immediately.

PR #2124, opened 2026-08-04
fix(viewer): drop ResizeObserver loop noise from error tracking

BIMvoice:
## Summary

- #2120 was auto-filed by PostHog from `ResizeObserver loop completed with undelivered notifications.` — the browser's documented, non-fatal signal that a resize notification had to be deferred a frame. It fires from correct observer code and (per Chromium/WebKit) carries no Error object, file/line, or stack.
- Audited every `ResizeObserver` call site in the viewer (`Viewport.tsx`, `Drawing2DCanvas.tsx`, `GanttTimeline.tsx`, `AnnotationLayer.tsx`, `SidebarDock.tsx`, `FloatingPanelHost.tsx`, and the MCP landing/playground/hero demos) — none of them mutates the size of the element it observes, so there is no runaway resize loop of ours behind this message; it's third-party browser noise, same class as the existing Cesium/Outlook/MapLibre/opaque-cross-origin entries in `analytics-scrub.ts`.
- Added a matcher following that file's conventions: anchored to the two known browser wordings (Chromium's current text and the older/WebKit "loop limit exceeded"), gated on zero stack frames (mirroring the `OPAQUE_CROSS_ORIGIN` "Script error." matcher) — if this message ever arrives WITH a stack, it means our own code threw it, and that stays reported.

## Test plan

- [x] RED: new test asserting the message is dropped, failing before the matcher existed
- [x] GREEN: `apps/viewer` — `analytics.test.ts` 47 → 50 tests, all passing
- [x] Negative cases: a message merely mentioning "ResizeObserver" (unrelated bug), and the same message WITH a stack, both survive
- [x] Mutation test: disabling the new matcher branch kills exactly 1 test (49 pass / 1 fail), restored by file copy
- [x] Full `apps/viewer` suite: 1990 → 1993 tests, pass count +3, the same 80 pre-existing unrelated failures (missing built workspace packages in a fresh worktree) before and after
- [x] No changeset — `@ifc-lite/viewer` is `private: true`

### DIFF
--- a/apps/viewer/src/lib/analytics-scrub.ts
+++ b/apps/viewer/src/lib/analytics-scrub.ts
@@ -224,6 +224,24 @@ const MAPLIBRE_WEBGL_UNAVAILABLE =
 // if a "Script error." ever arrives with a usable stack, it is ours to fix.
 const OPAQUE_CROSS_ORIGIN = /^Script error\.?$/i;
 
+// The browser dispatches this as a bare ErrorEvent on `window` whenever a
+// ResizeObserver callback resized something, so a notification round has to
+// be deferred to the next frame — the spec's documented, intentionally
+// non-fatal condition, not a hung page or an infinite loop. It fires from
+// textbook-correct observer code (this repo's dozen ResizeObserver call
+// sites in Viewport.tsx, Drawing2DCanvas.tsx, GanttTimeline.tsx,
+// AnnotationLayer.tsx, etc. were audited for #2120 and none mutates the
+// element it observes), which is exactly why it surfaces as a warning-shaped
+// message rather than throwing a real Error: no Error object, no file/line,
+// no stack — the same opaque, information-free shape as "Script error."
+// above, so it gets the same `frameCount === 0` gate: if this message ever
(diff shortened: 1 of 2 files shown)


louistrue:
Merging — viewer tier, green on all 5 core lanes, 0 unresolved.

Three things here are done the careful way, and since this PR *discards* telemetry they all matter:

**Anchored, not substring.** `/^ResizeObserver loop (?:completed with undelivered notifications\.?|limit exceeded)$/i` with both known browser wordings, and your comment says why: dropping is irreversible. A substring test would eventually swallow a real error that happened to mention a ResizeObserver.

**Gated on `frameCount(entry) === 0`.** This is the part that keeps it honest — if the message ever arrives *with* a stack, something in our code threw it deliberately and it still reaches you. The filter only removes the information-free shape.

**You audited the call sites rather than assuming.** Establishing that none of the dozen ResizeObserver users mutates the element it observes is what makes "this is the spec's non-fatal condition, not our bug" a finding instead of an assumption. Without it, silencing the message could hide a real observer loop.

That is the right standard for a scrub rule: prove the noise is noise, match it exactly, and leave a path for the case where it is not.

PR #2126, opened 2026-08-04
fix(parser): warn once per model when extractLengthUnitScale falls back to unconfirmed meters

### LINKED ISSUE #2104: extractLengthUnitScale returns 1.0 for "unknown", so a mis-scaled model is indistinguishable from a metre one
Raised by @louistrue on #2100, where the sweep fixed the exception channel and left the in-band one — which is the one that actually fires. Filing it separately because the remedy touches a shared function with 7+ callers.

## The defect

`packages/parser/src/unit-extractor.ts` — `extractLengthUnitScale` answers `1.0` for "I could not determine the scale". `1.0` is also a perfectly valid answer meaning "this model is in metres", so the two are indistinguishable to every caller.

Counted on `upstream/main`:

```
return 1.0 lines:                   11
console.warn lines in the function:  2
```

So nine failure paths — no `IFCPROJECT`, no project ref, no `UnitsInContext`, unreadable or malformed unit declarations — return a plausible metre scale silently.

## Why it matters, in the caller that surfaced it

`packages/create/src/in-store/extract-walls.ts` scales every extracted wall segment by this value. A millimetre model whose unit declarations are malformed reads as metres: coordinates like 31614, and the panel's metre-based snap tolerance collapses. #2100 added a `console.warn` there — but only on the **throw** path, which these nine do not take.

@louistrue's framing, which is the point:

> That makes the warning's presence slightly worse than its absence: a reader now sees a handled failure path and reasonably assumes the case is covered.

The comment in #2100 has been corrected to say exactly what it does and does not cover, so no claim is left standing — but the underlying gap is real.

## This is the same class the sweep was named for

#2100's title is *"four failures that rendered as legitimate results"*. A function returning `1.0` for "unknown" is that defect one layer down.

## Two honest remedies

1. **Make the nine silent paths warn.** Smallest, keeps the signature, no caller changes. Does not let a caller *decide* — it only makes the failure visible in the console.
2. **Return `null`/`undefined` for "unknown"** and let each caller choose its fallback. Correct, and the only version where a caller can distinguish the two cases — but it changes a shared function. Callers on `upstream/main` include `apps/viewer/src/lib/length-unit-scale.ts`, `hooks/useIfcCache.ts`, `lib/geo/effective-georef.ts`, `lib/geo/kmz-export.ts`, `components/viewer/PropertiesPanel.tsx`, `components/viewer/properties/ModelMetadataPanel.tsx`, and `components/mcp/playground-dispatcher.ts`.

(2) is the durable fix; (1) is landable today and would stop the silent case being invisible. Worth deciding rather than defaulting.

## Related

#2100 (the sweep that found it; the `extract-walls` comment is now accurate about its own scope), and the `schema.ts` thread on the same PR — a failed dynamic import printing a reduced schema through the normal JSON path with exit 0 is the same shape in the CLI.

BIMvoice:
## Summary

Closes #2104.

`extractLengthUnitScale` (`packages/parser/src/unit-extractor.ts`) returned `1.0` both for a confirmed metres declaration and for "I could not determine the scale" — the two were indistinguishable to every caller. Verified on `upstream/main` before this change: 11 `return 1.0` paths, only 2 preceded by a `console.warn`.

- 8 of the 9 previously-silent paths now call a small latched warning helper before returning `1.0`: no `IFCPROJECT`, an unresolvable `IFCPROJECT`/`UnitsInContext`/`IFCUNITASSIGNMENT` reference, a malformed `IFCUNITASSIGNMENT`, an unrecognized SI prefix, and "no length unit found in the assignment".
- The 9th (an `IfcSIUnit` length unit with no prefix — `prefix === null/undefined/'$'`) is a **genuine, confirmed** "this file declares meters" result, not an unknown, so it intentionally still does not warn. Warning there would have been a false-positive flood on the very common, entirely normal base-SI-metre case.
- The warning is latched per `entityIndex` (`WeakSet<object>`), not per call: `extractWallSegmentsForStorey` calls this once per storey and `resolveSpatialAnchor` once per generated space, both against the same model, so an unlatched warn would flood the console on a model with many storeys/spaces. A different model (a different `entityIndex` object) still warns fresh.
- No signature or return-value change — `extractLengthUnitScale` still returns a plain `number` and every existing caller's fallback-to-`1.0` behaviour is unchanged.

## (a) vs (b) — decision and evidence

The issue asked to decide between (a) warn on the in-band paths and (b) return `null` and let each caller decide, based on what the callers can actually do.

I read every call site (15+, more than the 7 listed in the issue — that list only covered `apps/viewer`, not `packages/mcp`, `packages/create`, or `packages/export`). The large majority — `columnar-parser.ts`, `extract-walls.ts` (×2), `resolve-anchor.ts`, `resolve-source.ts`, `kmz-export.ts` (×2), `lod0-generator.ts`, `demesh-session.ts`, `useIfcCache.ts`, `length-unit-scale.ts`, `effective-georef.ts` — use the value as an immediate multiplier on coordinates/geometry and already coalesce any non-finite/non-positive result back to `1` (metres). None of them has a sensible way to "refuse the operation" on `null`: refusing to scale a mesh isn't better than scaling it by 1 with a wrong guess, and rendering has to proceed either way. Turning the signature into `number | null` would just move the `?? 1` one line earlier at every one of these — a placebo, not a fix, while breaking every consumer's types.

A smaller set — the MCP `units` tool, `playground-dispatcher.ts`'s `units` handler, and the viewer's unit-metadata panels — do render/report the scale to a human or an agent, and could act on a distinguishable "unknown" (e.g. show "unable to determine"). But they're a minority, and they can already key off the new `console.warn` if closer coupling turns out to be worth it later without another breaking change now.

Given that read, (a) is the real fix here, not the placebo — I did not default to it without doing the caller survey the issue asked for.

## RED / GREEN

- RED: added tests to `packages/parser/test/unit-extractor.test.ts` (`unknown-unit warning (#2104)` describe block) covering: warns on "no length unit at all", warns on missing `IFCPROJECT`, warns on an unrecognized SI prefix, does *not* warn on a confirmed no-prefix metres unit, does *not* warn on a resolvable millimetre unit, and a two-directional latch test (fires once across repeated calls for the same model, fires again for a fresh model). Ran these against the pre-fix code first — 3 failed as expected (the two "already warns" cases and the "does not warn" case already passed).
- GREEN: same 6 tests pass against the fix; full `unit-extractor.test.ts` + `unit-scale.parity.test.ts` green (21/21).

## Mutation testing

Two mutations via exact-substring python3 replacement (count printed to stderr, `assert n==1`), restored by file copy (never `git checkout --`):

| Mutation | Result |
|---|---|
| Drop the `warnUnknownUnit(...)` call before the unrecognized-SI-prefix `return 1.0` | `warns on an unrecognized SI prefix` fails (0 calls) |
| Drop `warnedEntityIndexes.add(entityIndex)` from the latch (check-without-set) | latch test fails (3 calls instead of 1) |

## Test counts

- `packages/parser` full suite: 45 files / 376 tests passed (baseline unaffected files also pass once the workspace is fully built; this PR's own files are `unit-extractor.test.ts`, going from 9 to 15 tests, +6).
- `packages/create`'s `extract-walls-unit-scale.test.ts` (mocks `extractLengthUnitScale`) still passes — 1/1.
- `pnpm typecheck` (turbo, full monorepo): 34/34 tasks green.

## Where the brief was wrong

- The brief said the function lives in `packages/create/src/in-store/extract-walls.ts` "or wherever it now lives" — it's `packages/parser/src/unit-extractor.ts`; `extract-walls.ts` is one of its 15+ callers (the one that surfaced the issue on #2100).
- The brief's caller-count framing ("7+ callers") undercounts; the issue body's own list was scoped to `apps/viewer` only. There are 15+ across `packages/mcp`, `packages/parser`, `packages/create`, `packages/export`, and `apps/viewer`. The larger count reinforces rather than reverses the (a)-over-(b) call: more of the majority-can't-use-null callers, not fewer.

## Not fixed by this PR

- The one path that legitimately stays silent (explicit no-prefix `IfcSIUnit`) is not a gap — it's a confirmed result, not an unknown one.
- This does not change any caller's behavior or add a way to surface "unknown" to end users beyond the console; that's the deliberately-deferred (b) if the maintainer wants it later.

## Test plan

- [x] `pnpm --filter @ifc-lite/parser test` (unit-extractor + parity suites, then full package)
- [x] `pnpm --filter @ifc-lite/create test -- extract-walls-unit-scale`
- [x] `pnpm typecheck`
- [x] RED-first, mutation-tested per house rules

## Summary by CodeRabbit

* **Bug Fixes**
  * Warnings for unresolved length units now display once per model instead of multiple times, reducing warning spam during parsing.


louistrue:
Closes #2104. Verified and merging.

**The arithmetic in your description holds exactly.** I counted the paths rather than taking the summary:

```
return-1.0 paths: 11
paths that warn:  10
silent:            1   (line 183)
```

And line 183 is the one you said it would be — `prefix === null | undefined | '$'` on an `IFCSIUNIT`, which is a genuine declared metre, not an unconfirmed default. Leaving that one silent is the whole point: warning there would train people to ignore the warning.

**Mutations, both killed by specifically-named tests:**

| mutation | test |
|---|---|
| drop the per-model latch | `latches: warns once across many calls for the same model, but a fresh model warns again` |
| remove one warn call | `warns on a missing IFCPROJECT` |

The latch test asserting *both* halves — once per model **and** a fresh model warns again — is what makes it more than a call-count check. A latch keyed too broadly (module-global instead of per `entityIndex`) would pass a "warns once" assertion and silently suppress the diagnosis for every subsequent model in the session.

**The reasoning I most want to keep is why you did *not* change the signature.** From the changeset:

> a `null`-returning signature change was evaluated and rejected as a placebo for those call sites specifically because they'd all just coalesce it back to `1` immediately

That is the right call and the right justification. `null` would have looked like the more principled fix while changing nothing at 10+ geometry call sites that have no sensible operation to refuse on — they scale coordinates and need a number. Making the guess *visible* is the remedy that fits; making it *unrepresentable* would have been ceremony. Noting that the read/reporting call sites (the MCP `units` tool, the viewer unit panels) could still act on a distinguishable unknown later leaves the door open without paying for it now.

This closes the deferral `extract-walls.ts` documented — its comment says the in-band `1.0` paths are "tracked separately", and #2104 was that tracker. That claim is now true in both directions.

Issue #2021, opened 2026-08-03
Content matching recalls 0.468 on geometry-less objects (pre-registered target 0.5)


louistrue:
Split out of #1891. The xmatch fixture (#2018) gates on a ratchet floor of 0.448 while the pre-registered target is **0.5**, and prints `BELOW PRE-REGISTERED TARGET` on every run until this is closed. This issue is that target.

## The measurement

```
class none   n=47   recall 0.468085   precision 1
```

Precision is 1: nothing is being matched wrongly. The 25 unmatched elements **correctly abstain** rather than guess. So this is a recall gap, not a correctness bug, and the fixture is behaving exactly as designed by refusing to pair things it cannot tell apart.

## Cause

The data fingerprint omits `Tag` (and representation-map identity) for type objects. Eight `IfcFurnitureType` entities all named `800 mm` are therefore indistinguishable from one another, and with no geometry to fall back on there is no evidence left to separate them. Abstaining is the right answer given the available evidence; the fix is to give the fingerprint more evidence.

## Fix

Hash `Tag` into the data fingerprint for type objects, and consider representation-map identity alongside it.

This is deliberately **not** bundled into #2018. Changing what the fingerprint hashes changes its contract: every cached fingerprint is invalidated, and the change needs its own review and its own before/after numbers across the corpus rather than riding along in a test-infrastructure PR. When it lands, the floor ratchets up toward 0.5 and the `BELOW PRE-REGISTERED TARGET` line for `byClass.none.recall` disappears on its own.

## Two other strata are also below target

Same file, same mechanism, worth folding into the same investigation or splitting further:

| stratum | floor | measured | target |
|---|---|---|---|
| `byClass.none.recall` | 0.448 | 0.468085 | 0.5 |
| `byKind.renamed.recall` | 0.718 | 0.738 | 0.9 |
| `byKind.renamed.kindAgreement` | 0.924 | 0.944 | 0.98 |

The `renamed` pair is a different question from the geometry-less one and probably deserves its own issue once someone looks at it; recording it here so the three do not get lost separately.

PR #2127, opened 2026-08-04
fix(viewer): free the geometry WASM handle on the canonical load path (#1959)

BIMvoice:
## Summary

Closes the last open site from #1959: `useIfcLoader.loadFile` (the canonical
load path, primary and federated, every model open) constructs a
`GeometryProcessor` and never disposed it, leaking a WASM handle per load on
every exit — success or failure.

The reason it stayed open while the other eleven sites were fixed: the
handle is not private to geometry. On the main-thread parser fallback,
`loadFile` hands the raw handle to `IfcParser.parseColumnar` via
`geometryProcessor.getApi()` (`apps/viewer/src/hooks/useIfcLoader.ts:1177`).
A `finally { dispose() }` around the geometry block alone risks freeing
memory the parser is still reading — worse than the leak.

## Approach

Extracted `createGeometryProcessorDisposer` (`apps/viewer/src/hooks/ingest/geometryHandleDisposal.ts`),
mirroring `streamCleanup.ts`. It gates the free on **both** consumers finishing:
`release()` (the geometry stream / outer `finally`) and `parseSettled()` (the
parser chain), whichever finishes last triggers `dispose()`, exactly once.

`parseSettled()` is wired off the parser **chain's** `.finally()`, not off
`dataStorePromise`. That distinction matters: a superseded load's
`onFullDataStore` returns early without ever resolving `dataStorePromise`
(`useIfcLoader.ts:1154-1155`), so a disposer gated on the promise would leak
exactly the handle a model swap should free. The chain's `.finally()` runs on
all three endings — resolve, reject, and this stale/abandoned case — so
gating there is what makes the free reliable.

The parser is opt-in via `parseScheduled()` rather than assumed pending,
because engine init / `SharedArrayBuffer` allocation can throw before a parse
is ever scheduled; in that window there's no parser to wait for and the
handle should free immediately.

`release()` sits in a `finally` on `loadFile`'s outer `try` (not after the
last statement) — the function has many early returns past the point the
processor exists, and a trailing `dispose()` would miss all of them.

## Testing

- `apps/viewer/src/hooks/ingest/geometryHandleDisposal.test.ts` — 6 cases:
  frees immediately when no parse was scheduled; holds the handle while a
  scheduled parse is in flight; frees once the parse settles after release;
  frees once release happens after the parse settled; frees exactly once
  across repeated release/settle calls (dedupe); never frees while still
  running even after the parse settles.
- Two directions covered explicitly: freeing before the parse settles (the
  dangerous direction — a live use-after-free) is pinned as a failing case,
  and double-free/leak-via-missing-once-guard is pinned separately.
- Mutation-tested by hand: removing the once-guard (`disposed ||`) fails the
  dedupe test; removing the `parsePending` guard entirely fails the
  "holds the handle while a scheduled parse is in flight" test. Both
  mutations caught by a distinct test.
- `npm test` in `apps/viewer`: 2715 tests, 2711 pass, 2 pre-existing failures
  in `buildFingerprints.test.ts` (#2021/#1891 diff work) unrelated to this
  change — same failures reproduce on `upstream/main` at the branch's base
  commit before this fix is applied.
- `pnpm typecheck`: all 34 workspace tasks pass.

## Scope

This closes the `useIfcLoader.ts` site specifically. Per the issue's own
status check, the other three P0 sites (`PlaygroundViewer`,
`playground-dispatcher`, `HbjsonExportDialog`) are already fixed on `main`.
Not verifying here whether all 12 originally-audited sites are closed —
only that this PR closes this one.

No changeset: `@ifc-lite/viewer` is private.

Refs #1959.

### DIFF
--- a/apps/viewer/src/hooks/ingest/geometryHandleDisposal.test.ts
+++ b/apps/viewer/src/hooks/ingest/geometryHandleDisposal.test.ts
@@ -1,0 +1,94 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The #1959 leak in `useIfcLoader` is the one the rest of the issue could not
+ * take: the processor's raw handle goes to `IfcParser.parseColumnar` via
+ * `getApi()`, so freeing it when the geometry block ends would pull memory out
+ * from under a live parse. These cases pin the ordering rule that makes the
+ * free safe — never before the parse settles, always after — plus the two ways
+ * the parse can be absent entirely.
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { createGeometryProcessorDisposer } from './geometryHandleDisposal.js';
+
+describe('createGeometryProcessorDisposer', () => {
+  it('frees immediately when no parse was ever scheduled', () => {
+    // The engine-init / SharedArrayBuffer window: the processor exists but the
+    // parse chain was never started, so there is nothing to wait for.
+    let frees = 0;
+    const disposer = createGeometryProcessorDisposer(() => { frees += 1; });
+
+    disposer.release();
+
+    assert.equal(frees, 1);
+  });
+
+  it('holds the handle while a scheduled parse is still in flight', () => {
+    let frees = 0;
+    const disposer = createGeometryProcessorDisposer(() => { frees += 1; });
+
+    disposer.parseScheduled();
+    disposer.release();
+
+    assert.equal(frees, 0, 'freeing here would pull the handle out from under parseColumnar');
+  });
+
+  it('frees when the parse settles after the geometry stream released', () => {
+    let frees = 0;
+    const disposer = createGeometryProcessorDisposer(() => { frees += 1; });
+
+    disposer.parseScheduled();
+    disposer.release();
+    disposer.parseSettled();
+
+    assert.equal(frees, 1);
+  });
+
+  it('frees when the geometry stream releases after the parse settled', () => {
+    // The reverse order: the parser finished first and the stream drained (or
+    // threw) later. Whichever consumer is last must trigger the free.
+    let frees = 0;
+    const disposer = createGeometryProcessorDisposer(() => { frees += 1; });
+
+    disposer.parseScheduled();
+    disposer.parseSettled();
+    assert.equal(frees, 0, 'the geometry stream still holds it');
+
+    disposer.release();
+
+    assert.equal(frees, 1);
+  });
+
+  it('frees exactly once across repeated release and settle calls', () => {
+    // `release()` is wired to the outer `finally`, which runs on every exit
+    // path including ones that already released — a second free would be a
+    // double-free rather than a leak.
+    let frees = 0;
+    const disposer = createGeometryProcessorDisposer(() => { frees += 1; });
+
+    disposer.parseScheduled();
+    disposer.release();
+    disposer.parseSettled();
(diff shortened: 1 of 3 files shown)


louistrue:
This supersedes my #2129's viewer fix, and it found a real defect in it — thank you for the careful read.

Your docblock names the coupling I missed:

> `loadFile` passes the processor's raw handle to `IfcParser.parseColumnar` via `geometryProcessor.getApi()`. So a `finally { dispose() }` around the geometry block alone would free a handle the parser is still reading.

I verified rather than assuming, at `useIfcLoader.ts:1177` — `getApi()` feeds `parseColumnar` inside `dataStorePromise`, which is not settled when a geometry-block `finally` fires. So my change was a **use-after-free**, strictly worse than the leak. I have reverted it and said so on #2129.

The part I would not have reached is why the parser is an *opt-in* consumer:

> a load can fail between constructing the processor and scheduling the parse — engine init downloads the WASM binary, and the `SharedArrayBuffer` allocation can throw. In that window there is no parser to wait for and `release()` must free the handle immediately rather than defer forever on a parse that will never run.

That is the case a naive "wait for both" refcount gets wrong by hanging onto the handle forever, and it is exactly the window my `finally` also mishandled from the other direction.

## One thing to know about #2129

My gate initially **rejected this PR**. It only recognised a lexical `finally { p.dispose() }` and could not see through `createGeometryProcessorDisposer(() => geometryProcessor.dispose())`. A gate that fails the correct fix is how checks get disabled, so I have taught it the disposer shape — narrowly: `p.dispose()` inside a function passed as a call argument. A bare `p.dispose()` on the success path only is still flagged, asserted as a negative control.

With #2127 and #2128 applied, the gate is green:

```
✅ 29 constructions — 21 try/finally, 3 factory, 1 disposer-owned, 4 ownership-transferred
```

**#2129 is now gate-only and deliberately red until this lands** — it is reporting the exact leak you fix here. Merge order: #2127 and #2128 first, then #2129 goes green with no further edits.

Reviewing the substance of this one next.

louistrue:
Verified and merging. This is the correct fix and mine was not.

Mutation-tested the state machine — three mutations, three distinctly-named kills:

| mutation | test that caught it |
|---|---|
| ignore `parsePending` (**the use-after-free my #2129 had**) | `holds the handle while a scheduled parse is still in flight` |
| drop the `disposed` idempotence guard | `frees exactly once across repeated release and settle calls` |
| `parseSettled` stops re-checking | `frees when the parse settles after the geometry stream released` |

Each kills exactly one named test, so the suite pins three separate properties rather than one behaviour three ways.

The two orderings that make this correct, both of which I checked rather than assumed:

**`parseScheduled()` precedes any possible `release()`.** It is called at :1265, before `setTimeout(startDataModelParsing, 0)` arms the parse and long before the outer `finally` at :1921. So a release-triggered dispose can never race ahead of a parse that is about to be scheduled — which is the hole a naive refcount leaves.

**`parseSettled()` hangs off the promise *chain*, not the promise.** Your comment is the reason this matters:

> The parser is done with the raw handle here on EVERY ending — including the one `dataStorePromise` never reports: a stale `onFullDataStore` returns without resolving it.

A `.then`-only or promise-gated settle would leave `parsePending` true forever on a superseded load, and the handle would never be freed — trading the use-after-free for a permanent leak on exactly the path (stale session) that happens most under rapid model switching.

And `release()` in the outer `finally` covers what my version missed. Your comment enumerates it: "stale session, resource retry, the inner geometry catch, the cache and server fast paths". My `finally` sat on the inner geometry try and would have missed every one.

6/6 tests pass. Once #2128 lands, my gate PR (#2129) goes green with no further edits — I have already taught it to recognise `createGeometryProcessorDisposer(() => p.dispose())` so it does not reject this shape.

PR #2128, opened 2026-08-04
fix(cli): re-audit #1959 remaining sites — dispose the shared GeometryProcessor in `ifc-lite clash`

BIMvoice:
## Summary

Re-audit of the remaining #1959 sites on current `upstream/main`, as requested in the issue's status-check comments (PR #2127 explicitly scoped itself to `useIfcLoader.ts` only and said the other eight were not re-verified).

Full classification of every `new GeometryProcessor(` site — the issue's own table is a starting point, verified fresh rather than trusted — is posted as a comment on #1959.

**Result: only one genuine leak remains**, and it's a new find, not one of the original eight:

- `packages/cli/src/commands/clash.ts:49-51` — a module-scoped `sharedProcessor`, created lazily by `getProcessor()` and never disposed on any exit path (success, thrown clash/BCF error, or early return). This is the same shape as the P2 CLI-hygiene bucket the original issue flagged (low urgency — the CLI is a one-shot process, so the OS reclaims the handle on exit either way) but it was still a real, unaddressed violation of the try/finally rule in `AGENTS.md` §7.

Everything else — all P0 sites, `packages/mcp/src/tools/clash.ts`, `headless-backend.ts` (the site no longer exists — the module has no `GeometryProcessor` usage at all now), `lod1-generator.ts` (both processors), `diagnose-geometry.ts`, `extract-entities.ts`, `cli/commands/export.ts`, `gym.ts`, `create-ifc-lite/templates/react.ts`, and two viewer sites added since the original audit (`UsdExportDialog.tsx`, `CommandPalette.tsx`) — already dispose correctly on every path, either because they were fixed by other merged PRs or were already correct. The example/vanilla-template page-lifetime singletons (`babylonjs.ts`, `threejs.ts` x2 in examples + 2 in create-ifc-lite templates) are confirmed still page-lifetime by design, not leaks, per the issue's original carve-out.

## Fix

`clashCommand` now wraps its whole run in `try { … } finally { sharedProcessor?.dispose(); sharedProcessor = undefined; }`, same shape as every other fix in this sweep (`PlaygroundViewer`, `playground-dispatcher`, `HbjsonExportDialog`, `gym.ts`). Resetting to `undefined` in the `finally` also means a long-lived host that calls `clashCommand` more than once per process (e.g. a test harness) doesn't accumulate one handle per call — mirrors the fix already landed in `gym.ts` for the same pattern.

`useIfcLoader.ts` is intentionally **not** touched — PR #2127 is open against that exact site with its own approach and tests; redoing it here would conflict.

## Test plan

- [x] RED: added 3 tests to `clash.test.ts` (success-path disposal, throw-path disposal, no-eager-construction-before-meshing) — 2 of 3 fail against the pre-fix code (`toHaveBeenCalledTimes(1)` → `0`), confirmed by mutation-testing (python3 exact-substring removal of the `dispose()` call, count asserted `==1`, restored by file copy — never `git checkout --`).
- [x] GREEN: full `clash.test.ts` passes (4/4) with the fix in place.
- [x] Full `packages/cli` suite: 270 passed, 8 skipped (pre-existing dist/wasm-build-gated subprocess tests, unrelated), 0 failed.
- [x] `pnpm typecheck`: 34/34 workspace tasks clean.
- [x] Changeset added (`@ifc-lite/cli` is published, not private).

Refs #1959.

### DIFF
--- a/.changeset/clash-cli-dispose-shared-processor.md
+++ b/.changeset/clash-cli-dispose-shared-processor.md
@@ -1,0 +1,7 @@
+---
+"@ifc-lite/cli": patch
+---
+
+Free the WASM geometry handle `ifc-lite clash` allocates for meshing (#1959).
+
+`clash.ts` lazily creates a module-scoped `sharedProcessor` (`getProcessor()`) the first time a run needs to mesh a model, and reuses it for every subsequent mesh within the same process. It was never disposed on any exit path — success, a thrown clash/BCF error, or an early return — leaking the handle for the life of the process. Low real-world impact (the CLI is a one-shot process, so the OS reclaims the WASM memory on exit either way), but it violates the deterministic-disposal rule the audit in #1959 is checking, so it is fixed to the same shape as the rest of that sweep: the whole run now sits inside `try { … } finally { sharedProcessor?.dispose(); sharedProcessor = undefined; }`, so a subsequent `clashCommand` call in the same process (e.g. a long-lived host embedding the CLI's command functions) starts from a fresh handle instead of accumulating one per call.
--- a/packages/cli/src/commands/clash.test.ts
+++ b/packages/cli/src/commands/clash.test.ts
@@ -17,7 +17,7 @@
  * rather than proving nothing on a silent model.
  */
 
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, afterEach } from 'vitest';
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
 import { existsSync } from 'node:fs';
@@ -26,6 +26,8 @@ import { tmpdir } from 'node:os';
 import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
 import { IfcCreator } from '@ifc-lite/create';
+import { GeometryProcessor } from '@ifc-lite/geometry';
+import { clashCommand } from './clash.js';
 
 const execFileAsync = promisify(execFile);
 
@@ -85,4 +87,75 @@ describe('clash --json stdout hygiene', () => {
     },
     180_000,
   );
+});
+
+describe('clashCommand GeometryProcessor disposal (#1959 P2 leak)', () => {
+  // `sharedProcessor` (clash.ts) is module-scoped, so each test uses a
+  // uniquely-named model file — `meshModel`'s cache key is `basename(filePath)`
+  // only (not the full path), so two tests both writing to `model.ifc` in
+  // different tmpdirs would collide and the second call would skip meshing
+  // (and therefore skip `getProcessor()`) entirely, making the assertion
+  // pass for the wrong reason.
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  async function withTempModel(basename: string, fn: (path: string) => Promise<void>): Promise<void> {
+    const dir = await mkdtemp(join(tmpdir(), 'ifc-lite-clash-dispose-'));
+    const modelPath = join(dir, basename);
+    try {
+      await writeFile(modelPath, buildClashModel());
+      await fn(modelPath);
+    } finally {
+      await rm(dir, { recursive: true, force: true });
+    }
+  }
+
+  it('disposes the GeometryProcessor WASM handle on the success path', async () => {
+    vi.spyOn(process.stdout, 'write').mockImplementation(() => true);
+    vi.spyOn(process.stderr, 'write').mockImplementation(() => true);
+    const disposeSpy = vi.spyOn(GeometryProcessor.prototype, 'dispose');
+
+    await withTempModel('dispose-success.ifc', async (modelPath) => {
+      await clashCommand([modelPath, '--json']);
+      expect(disposeSpy).toHaveBeenCalledTimes(1);
+    });
+  }, 60_000);
+
+  it('disposes the GeometryProcessor WASM handle even when clashing throws after meshing', async () => {
+    vi.spyOn(process.stdout, 'write').mockImplementation(() => true);
+    vi.spyOn(process.stderr, 'write').mockImplementation(() => true);
+    const disposeSpy = vi.spyOn(GeometryProcessor.prototype, 'dispose');
+    // Force the throw downstream of `getProcessor()`/meshing (mirrors how
+    // `process()` genuinely fails), so this exercises the `finally` on the
+    // error path rather than only the happy path.
+    vi.spyOn(GeometryProcessor.prototype, 'process').mockRejectedValue(
+      new Error('forced meshing failure for #1959 dispose test'),
+    );
+
+    await withTempModel('dispose-throw.ifc', async (modelPath) => {
+      await expect(clashCommand([modelPath, '--json'])).rejects.toThrow(
+        'forced meshing failure for #1959 dispose test',
+      );
+      expect(disposeSpy).toHaveBeenCalledTimes(1);
+    });
+  }, 60_000);
+
+  it('never constructs a GeometryProcessor when argument parsing fails before meshing', async () => {
+    const exitSpy = vi.spyOn(process, 'exit').mockImplementation(((() => {
+      throw new Error('process.exit called');
+    }) as unknown) as (code?: number) => never);
+    vi.spyOn(process.stderr, 'write').mockImplementation(() => true);
+    const disposeSpy = vi.spyOn(GeometryProcessor.prototype, 'dispose');
+    try {
+      await expect(clashCommand(['--mode', 'not-a-real-mode', 'some-file.ifc'])).rejects.toThrow(
+        'process.exit called',
+      );
+      // getProcessor() is never reached — nothing to dispose. Pins that the
+      // fix did not turn a lazy processor into an eager one.
+      expect(disposeSpy).not.toHaveBeenCalled();
+    } finally {
+      exitSpy.mockRestore();
+    }
+  }, 30_000);
 });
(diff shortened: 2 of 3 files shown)


louistrue:
Better than what I had. My #2129 treated `sharedProcessor` as **ownership transfer** — assignment to a module-level binding, lifetime not decidable from the constructing scope — so my gate accepted it and the handle still leaked for the life of the process. You actually fixed it.

The part that makes it more than a formality is the reset:

> `try { … } finally { sharedProcessor?.dispose(); sharedProcessor = undefined; }`, so a subsequent `clashCommand` call in the same process (e.g. a long-lived host embedding the CLI's command functions) starts from a fresh handle instead of accumulating one per call.

Disposing without clearing the module-level binding would leave a dangling reference to a freed handle, which is worse than the leak — the next call would reuse it. Both halves are needed and both are there.

The test note is the kind of thing that keeps a test honest:

> `meshModel`'s cache key is `basename(filePath)` only (not the full path), so two tests both writing to `model.ifc` in different tmpdirs would collide and the second call would skip meshing (and therefore skip `getProcessor()`) entirely, making the assertion pass for the wrong reason.

A test that passes because the code under test never ran is the failure mode I have been finding all day; naming it in the fixture is what stops someone "simplifying" the unique filenames later.

Once this and #2127 land, my gate PR (#2129) goes green with no further edits — it is currently red on purpose, reporting the `useIfcLoader` leak #2127 fixes.

louistrue:
greptile's remaining thread is valid and it is a two-line fix — worth taking before this merges.

```ts
} finally {
  sharedProcessor?.dispose();
  sharedProcessor = undefined;   // skipped if dispose() throws
}
```

If `dispose()` throws, two things go wrong at once: the cleanup exception replaces the original clash/BCF error the caller was about to see, and `sharedProcessor` keeps pointing at a processor whose handle may be half-freed. A later `clashCommand` in the same host then reuses it — which is the exact dangling-reference case your reset was added to prevent, reintroduced on the failure path.

Not hypothetical in this repo: #1922 is an OOM inside a drained job aborting the WASM module at dispose time. That is precisely a throwing `dispose()`.

Ordering fixes the dangling half for free:

```ts
} finally {
  const processor = sharedProcessor;
  sharedProcessor = undefined;   // reset first — a throwing dispose cannot skip it
  processor?.dispose();
}
```

If you also want the original error preserved, wrap the `dispose()` in its own `try/catch` and warn — a cleanup failure should not mask what the caller was actually reporting. Your own framing on the reset applies here too: disposing without clearing is worse than the leak, because the next call reuses a freed handle.

Everything else on this one is verified — #2127 just merged, and once this lands my gate PR (#2129) goes green with no further edits.

BIMvoice:
Taken in `9d767d9b` — both halves, and you were right that the ordering fixes the dangling case for free.

```ts
} finally {
  const processor = sharedProcessor;
  sharedProcessor = undefined;   // cannot be skipped by a throwing dispose
  try {
    processor?.dispose();
  } catch (err) {
    console.warn('[clash] geometry processor dispose failed; continuing', err);
  }
}
```

Reverting to the old ordering fails **both** new tests:

```
× clears the shared processor even when dispose() throws, so the next call gets a fresh handle
× does not let a failing dispose() mask the error the caller was reporting
```

`packages/cli` 272 passed / 8 skipped, `tsc --noEmit` clean.

The masking half earns its own test rather than riding along: a cleanup exception replacing `the real clash failure` tells the user the wrong thing entirely, and it is the half that survives a careless "simplify the finally" later.

**#1922 being the concrete case is what makes this worth the lines.** An OOM inside a drained job aborting the WASM module at dispose time is a throwing `dispose()` — so the failure path here is one we have already filed against ourselves, not a defensive hypothetical.

The first test uses distinct filenames on purpose, for the reason you quoted back: `meshModel`'s cache key is basename-only, so reusing one name skips meshing on the second call, never reaches `getProcessor()`, and the assertion passes for the wrong reason. That is now stated in the fixture so it survives someone tidying the names.

## On the gate

Agreed that yours is the better answer for `useIfcLoader` and I would rather have #2129 than our #2127 — which has now merged, so your gate should be unblocked on that side once this lands.

One thing outstanding from my review of #2129 that I would still like your read on: on the main-thread parser path, `dispose()` at `:1806` can run while `parseColumnar` still holds `parserWasmApi` from `:1169`, because the only `await` on that chain (`:1820`) is outside the `try/finally`. I have not reproduced it in a browser, so it may be that the ordering is fine for a reason I cannot see from the outside — but if it is real, the gate would be enforcing a `finally` that frees a handle still in use.

louistrue:
The ordering fix is correct — but **the test named for it cannot detect it.** Verified by mutation.

I moved the reset back inside the `try`, after `dispose()`, which is exactly the bug the test's own comment describes:

```ts
const processor = sharedProcessor;
try {
  processor?.dispose();
  sharedProcessor = undefined;   // skipped when dispose throws
} catch (err) { … }
```

Confirmed the mutation was actually in the file before trusting the result (I produced a no-op mutation elsewhere today and nearly reported a false finding off it):

```
17:    const processor = sharedProcessor;
19:      processor?.dispose();
20:      sharedProcessor = undefined;
21:    } catch (err) {
```

Result: `clears the shared processor even when dispose() throws, so the next call gets a fresh handle` — **passes**. The whole suite stays at 5 passed.

## Why it cannot discriminate

The assertion counts **disposals**, and disposal happens twice either way:

| | correct code | mutated (reset skipped) |
|---|---|---|
| run 1 | construct, dispose throws, binding cleared | construct, dispose throws, binding **kept** |
| run 2 | `getProcessor()` sees `undefined` → **constructs fresh** → dispose | `getProcessor()` sees the stale processor → **reuses it** → dispose |
| `disposeSpy` calls | 2 | 2 |

`getProcessor()` gates on `if (!sharedProcessor)`, so under the mutation run 2 never constructs — it disposes the *same, already-failed* processor a second time. That is precisely the dangling-reuse this PR exists to prevent, and `toHaveBeenCalledTimes(2)` is satisfied by it.

## The fix

Assert on **construction**, not disposal — the property is "a fresh handle", and only construction shows that:

```ts
const ctorSpy = vi.spyOn(...);            // or capture instances
expect(ctorSpy).toHaveBeenCalledTimes(2);
```

Or capture identity, which reads closer to the contract:

```ts
const instances: GeometryProcessor[] = [];
vi.spyOn(GeometryProcessor.prototype, 'dispose').mockImplementation(function (this: GeometryProcessor) {
  instances.push(this);
  throw new Error('dispose boom');
});
…
expect(instances[0]).not.toBe(instances[1]);   // fails under the mutation
```

The second form has the nice property of failing with a message that names the actual defect — the same object was disposed twice — rather than a count mismatch.

Everything else here is verified: removing the `try/catch` around `dispose()` correctly fails **two** named tests (`clears the shared processor…` and `does not let a failing dispose() mask the error…`), so the swallow-and-warn half is genuinely pinned. It is only the ordering half that is unpinned, and the comment above it is currently making a promise the assertion does not keep.

BIMvoice:
You are right, and I reproduced it before fixing: your mutation — reset moved back inside the `try`, after `dispose()` — leaves the suite at **6 passed**. Fixed in `26c9d597`.

## Why it could not discriminate, confirmed

Your table is exactly it. The assertion counted **disposals**, and disposal happens twice either way: under the mutation the second run reuses the stale processor and disposes *that* a second time. `toHaveBeenCalledTimes(2)` is satisfied **by the dangling reuse the fix exists to prevent**.

The observable that separates the two is **construction**. `getProcessor()` calls `init()` only inside `if (!sharedProcessor)`, so init-count distinguishes "fresh handle" from "reused dead handle":

```ts
const initSpy = vi.spyOn(GeometryProcessor.prototype, 'init');
...
expect(initSpy).toHaveBeenCalledTimes(1);   // after run 1
expect(initSpy).toHaveBeenCalledTimes(2);   // after run 2 — the discriminator
```

Your mutation now fails; `packages/cli` 272 passed / 8 skipped, `tsc --noEmit` clean.

## The part I got wrong, and it is the interesting part

I *did* mutation-verify this test before pushing — and it passed my check. My mutation reverted the whole `finally` to the pre-fix `dispose(); reset;` form, **which also removed the `try/catch`**. That made the *masking* test fail, so I saw red and concluded the ordering was pinned. It was not: two tests were in play and I attributed one test's kill to the other's property.

So a coarse mutation can produce a false green on the specific claim, by being caught for an adjacent reason. The lesson I am taking: the mutation has to be the **minimal** edit that breaks only the property under test — yours changed nothing but the statement order, which is why it was diagnostic and mine was not.

Also noted your own guard from the same comment — verifying the mutation was actually in the file before trusting the result, after nearly reporting a false finding off a no-op edit today. I have hit that too; printing the replacement count and re-reading the region is now standard here, and it is what caught a `REPLACEMENTS=0` for me earlier.

louistrue:
Verified — the ordering is now pinned. The mutation that survived last round kills it:

```
### reset AFTER dispose, inside the try   (occ=1, applied and confirmed in-file)
     × clears the shared processor even when dispose() throws, so the next call gets a fresh handle
      Tests  1 failed | 4 passed | 1 skipped (6)
```

Restored: 5 passed.

Spying on `init()` is a better answer than the instance-identity capture I suggested. `getProcessor()` calls `new GeometryProcessor()` and `await processor.init()` together, so `initSpy` counts constructions exactly — and it does it without a `this`-typed mock implementation, which is fiddlier and easier to get subtly wrong. Keeping `disposeSpy` alongside it is right too: together they say "disposed twice **and** constructed twice", which is the actual contract. Either alone is satisfiable by the bug.

That was the last substantive item from my side. Once CI finishes I will merge — 3 lanes still running, and I want the core lanes present rather than just not-red, since a `CONFLICTING` branch reports a green rollup with the real lanes simply absent (that trap has hit three of your PRs today).

PR #2129, opened 2026-08-04
fix(viewer,ci): gate WASM handle disposal instead of policing it by hand

### LINKED ISSUE #1959: 12 GeometryProcessor call sites never dispose their WASM handle, including the canonical load path

louistrue:
Closes #1959.

`AGENTS.md` already required WASM handles to be freed in `try/finally`. The rule was self-policed, so compliance decayed one call site at a time until the #1959 audit found **12 sites that never disposed on any path** — including the viewer's canonical load path, which fires on every IFC load. Ten have since been fixed one at a time. That pattern does not stop the thirteenth, so this adds a gate.

## The gate

`scripts/check-wasm-disposal.mjs` walks the TypeScript AST and accepts a `new GeometryProcessor(...)` only when one of these holds:

| accepted | why |
|---|---|
| `using` / `await using` binding | scope exit frees it |
| `try { … } finally { p.dispose() }` | explicit |
| the construction is a factory body | caller owns it |
| ownership transfers out (returned, `this.gp = p`, or assigned to an outer binding) | lifetime is not decidable from this scope |

Anything else fails CI. Wired as `pnpm check:wasm-disposal`, next to `check-test-wiring` in the node-test job.

**Mutation-verified in four directions**, because a gate that cannot fail is worse than no gate:

```
M1  leaked local (const, never disposed)   -> exit 1, names the file:line   ✅ catches
M2  same site, try/finally-disposed        -> exit 0                        ✅ no false positive
M3  same site, `using`-bound               -> exit 0                        ✅ no false positive
M4  revert the useIfcLoader fix            -> exit 1                        ✅ catches
```

Current state: `29 constructions — 22 try/finally, 3 factory, 4 ownership-transferred, 0 allowlisted`.

### The gate's own first bug, kept as a comment

My first version reported `✅ 26 using` against a codebase with essentially none. `NodeFlags.AwaitUsing` is `Using | Const` (6), so `flags & AwaitUsing` is non-zero for **every ordinary `const`** — the gate was accepting the entire repo while reporting success. Testing the `Using` bit alone fixes it, and the reason is now a comment on that function, because it is exactly the kind of bitmask slip that silently disarms a check.

## The two real leaks

**`apps/viewer/src/hooks/useIfcLoader.ts`** — the canonical load path, freed nothing. Now disposes in a `finally` covering the streaming block and every `return` out of its catch.

**`packages/cli/src/commands/clash.ts`** — caches a process-lifetime singleton in a module-level binding. Not a leak; accepted as ownership transfer, which also removed the need to allowlist it.

## Three things this deliberately does not do

**`using` is not usable in `apps/viewer` yet.** It type-checks, but rolldown (Vite 8) **panics rendering chunks** and the production build fails. Verified by building both ways on this branch:

```
const  -> ✓ built in 15.00s
using  -> thread '<unnamed>' panicked at miette-7.6.0/.../graphical.rs:1159
          'Formatting argument out of range'   (exit 134)
```

So the viewer fix is a `finally`, not the `using` one-liner the changelog for #1563 advertises. Filing that separately — it is a real constraint on the ergonomic form across the whole app.

**The gate proves a narrow thing.** A processor that stays local and is never freed. It does not verify that the `finally` dominates the construction, and it accepts escapes. Widening it would produce the false positives that get a check disabled, and #1959's own audit lists the escape sites (`export.ts` returning `{ bytes, gp }`, `demesh-session.ts` assigning `this.gp`) under "correct, for the record".

**One window in `useIfcLoader` stays uncovered**, and the comment says so: a throw between construction and the `try` — engine-init or the parse stages — unwinds to the outer catch where the binding is out of scope. Widening the try instead would reroute those errors into a catch written specifically for stream failures, which is a behaviour change rather than a disposal fix.

## Verification

`tsc --noEmit` clean, `pnpm build` succeeds, 11 loader/drawing tests pass, gate green. No changeset: `apps/viewer` is private and the other changes are tooling and CI.

## Summary by CodeRabbit

* **Tests**
  * Added automated validation to confirm WebAssembly resources are properly cleaned up across supported usage patterns.
  * Builds now detect missing cleanup, helping prevent resource leaks and improve runtime stability.

* **Chores**
  * Added a dedicated command for running the WebAssembly cleanup check locally and in continuous integration.
  * The check reports categorized results and provides guidance for resolving detected issues.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -333,6 +333,12 @@ jobs:
         run: pnpm fixtures
       - name: Check test wiring
         run: node scripts/check-test-wiring.mjs
+      # Every `new GeometryProcessor(...)` must free its WASM handle on all
+      # paths. AGENTS.md required this already, but self-policing decayed one
+      # call site at a time until 12 sites leaked (#1959) -- including the
+      # viewer's canonical load path, which fires on every IFC load.
+      - name: Check WASM handle disposal
+        run: node scripts/check-wasm-disposal.mjs
       # Guard the published API surface: the exported names of every
       # non-private packages/* package (read from the dist d.ts in the
       # build artifact downloaded above) must match the committed
--- a/apps/viewer/src/hooks/useIfcLoader.ts
+++ b/apps/viewer/src/hooks/useIfcLoader.ts
@@ -1047,6 +1047,14 @@ export function useIfcLoader() {
       // Initialize geometry processor first (WASM init is fast if already loaded)
       // Reuses the merge-layers snapshot taken above for the cache key so the
       // key and the WASM tessellation always agree (issues #540, #1107).
+      //
+      // `using`, not `const`: this fires on EVERY load, primary and federated,
+      // and previously leaked its `IfcAPI` handle on every one of them (#1959).
+      // Scope exit covers the paths a `finally` on the try below would miss —
+      // the engine-init await and the parse stages between construction and
+      // that try. Safe to free here: meshes are copied into JS and `.free()`d
+      // as they are extracted, so `dispose()` only releases the long-lived
+      // handle, and it is idempotent.
       const geometryProcessor = new GeometryProcessor({
         quality: GeometryQuality.Balanced,
         // Auto-low vertex density for heavy models (or `?geomTier=` override);
@@ -1776,6 +1784,26 @@ export function useIfcLoader() {
         setLoading(false);
         setGeometryStreamingActive(false);
         return;
+      } finally {
+        // Free the engine's `IfcAPI` handle. This fires on EVERY load, primary
+        // and federated, and leaked on every one of them before #1959 — the
+        // single highest-frequency WASM handle leak in the app.
+        //
+        // Here rather than `using`: `using` type-checks, but rolldown (Vite 8)
+        // panics rendering chunks on it, so the viewer will not build. Revisit
+        // when that is fixed — `GeometryProcessor` already implements
+        // `[Symbol.dispose]()` (#1563).
+        //
+        // Covers the streaming block and every `return` out of the catch above.
+        // NOT covered: a throw between construction and this `try` (engine-init
+        // or the parse stages), which unwinds to the outer catch where this
+        // binding is out of scope. Widening the try instead would reroute those
+        // errors into the catch above, which handles stream failures
+        // specifically — a behaviour change, not a disposal fix.
+        //
+        // Safe here: meshes are copied into JS and freed as they are extracted,
+        // so this only releases the long-lived handle, and it is idempotent.
+        geometryProcessor.dispose();
       }
 
       if (loadSessionRef.current !== currentSession) {
--- a/apps/viewer/src/hooks/useIfcLoader.ts
+++ b/apps/viewer/src/hooks/useIfcLoader.ts
@@ -1047,14 +1047,6 @@ export function useIfcLoader() {
       // Initialize geometry processor first (WASM init is fast if already loaded)
       // Reuses the merge-layers snapshot taken above for the cache key so the
       // key and the WASM tessellation always agree (issues #540, #1107).
-      //
-      // `using`, not `const`: this fires on EVERY load, primary and federated,
-      // and previously leaked its `IfcAPI` handle on every one of them (#1959).
-      // Scope exit covers the paths a `finally` on the try below would miss —
-      // the engine-init await and the parse stages between construction and
-      // that try. Safe to free here: meshes are copied into JS and `.free()`d
-      // as they are extracted, so `dispose()` only releases the long-lived
-      // handle, and it is idempotent.
       const geometryProcessor = new GeometryProcessor({
         quality: GeometryQuality.Balanced,
         // Auto-low vertex density for heavy models (or `?geomTier=` override);
@@ -1784,26 +1776,6 @@ export function useIfcLoader() {
         setLoading(false);
         setGeometryStreamingActive(false);
         return;
-      } finally {
-        // Free the engine's `IfcAPI` handle. This fires on EVERY load, primary
-        // and federated, and leaked on every one of them before #1959 — the
-        // single highest-frequency WASM handle leak in the app.
-        //
-        // Here rather than `using`: `using` type-checks, but rolldown (Vite 8)
-        // panics rendering chunks on it, so the viewer will not build. Revisit
-        // when that is fixed — `GeometryProcessor` already implements
-        // `[Symbol.dispose]()` (#1563).
-        //
-        // Covers the streaming block and every `return` out of the catch above.
-        // NOT covered: a throw between construction and this `try` (engine-init
-        // or the parse stages), which unwinds to the outer catch where this
-        // binding is out of scope. Widening the try instead would reroute those
-        // errors into the catch above, which handles stream failures
-        // specifically — a behaviour change, not a disposal fix.
-        //
-        // Safe here: meshes are copied into JS and freed as they are extracted,
-        // so this only releases the long-lived handle, and it is idempotent.
-        geometryProcessor.dispose();
       }
 
       if (loadSessionRef.current !== currentSession) {
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "test:integration": "tsx --tsconfig tests/tsconfig.json tests/integration.test.ts",
     "test:wasm-contract": "node scripts/test-wasm-contract.mjs",
     "check:test-wiring": "node scripts/check-test-wiring.mjs",
+    "check:wasm-disposal": "node scripts/check-wasm-disposal.mjs",
     "check:api-surface": "node scripts/check-api-surface.mjs",
     "api-surface:update": "node scripts/check-api-surface.mjs --update",
     "check:git-lfs": "node scripts/check-git-lfs.mjs",
--- a/scripts/check-wasm-disposal.mjs
+++ b/scripts/check-wasm-disposal.mjs
@@ -144,6 +144,58 @@ function disposedInFinally(scope, name) {
 }
 
 /**
+ * Is the handle handed to something that owns its disposal?
+ *
+ *     geometryHandle = createGeometryProcessorDisposer(() => processor.dispose());
+ *
+ * A lexical `finally` is not always correct. In `useIfcLoader` the processor is
+ * shared with the data-model parser via `getApi()`, so freeing it at the end of
+ * the geometry block would free a handle the parser is still reading — the
+ * disposal has to wait for whichever consumer finishes last (#2127). That is a
+ * real disposal, expressed through an indirection this gate must not punish:
+ * a check that fails the *correct* fix is the thing that gets checks disabled.
+ *
+ * Accepted shape: `name.dispose()` inside a function passed as an argument to a
+ * call. Deliberately narrower than "mentions dispose anywhere" — a bare
+ * `p.dispose()` on the success path only would still be flagged.
+ */
+function handedToDisposer(scope, name) {
+  let found = false;
+  const disposesName = (node) => {
+    let hit = false;
+    const scan = (n) => {
+      if (hit) return;
+      if (
+        ts.isCallExpression(n) &&
+        ts.isPropertyAccessExpression(n.expression) &&
+        n.expression.name.text === 'dispose' &&
+        ts.isIdentifier(n.expression.expression) &&
+        n.expression.expression.text === name
+      ) {
+        hit = true;
+      }
+      ts.forEachChild(n, scan);
+    };
+    scan(node);
+    return hit;
+  };
+  const visit = (node) => {
+    if (found) return;
+    if (ts.isCallExpression(node)) {
+      for (const arg of node.arguments) {
+        if ((ts.isArrowFunction(arg) || ts.isFunctionExpression(arg)) && disposesName(arg)) {
+          found = true;
+          break;
+        }
+      }
+    }
+    ts.forEachChild(node, visit);
+  };
+  visit(scope);
+  return found;
+}
+
+/**
  * Does `name` escape the scope that constructed it?
  *
  * A processor that is returned, or stored on an object/field, is owned by
@@ -209,7 +261,7 @@ function isFactoryBody(newExpr) {
 }
 
 const violations = [];
-const accepted = { using: 0, finally: 0, factory: 0, allowlisted: 0, escapes: 0 };
+const accepted = { using: 0, finally: 0, factory: 0, allowlisted: 0, escapes: 0, disposer: 0 };
 let scanned = 0;
 
 for (const scanDir of SCAN_DIRS) {
@@ -236,6 +288,7 @@ for (const scanDir of SCAN_DIRS) {
         else if (decl && isUsingBinding(decl)) accepted.using++;
         else if (key && ALLOWLIST.has(key)) accepted.allowlisted++;
         else if (name && disposedInFinally(scope, name)) accepted.finally++;
+        else if (name && handedToDisposer(scope, name)) accepted.disposer++;
         else if (name && escapesScope(scope, name)) accepted.escapes++;
         else {
           violations.push({
@@ -253,7 +306,7 @@ for (const scanDir of SCAN_DIRS) {
 }
 
 const total =
-  violations.length + accepted.using + accepted.finally + accepted.factory + accepted.allowlisted + accepted.escapes;
+  violations.length + accepted.using + accepted.finally + accepted.factory + accepted.allowlisted + accepted.escapes + accepted.disposer;
 
 if (violations.length > 0) {
   console.error(
@@ -281,6 +334,6 @@ ALLOWLIST in scripts/check-wasm-disposal.mjs with the reason. See issue #1959.`)
 console.log(
   `✅ All ${total} WASM-handle construction(s) release deterministically ` +
     `(${accepted.using} using, ${accepted.finally} try/finally, ${accepted.factory} factory, ` +
-    `${accepted.escapes} ownership-transferred, ${accepted.allowlisted} allowlisted) ` +
+    `${accepted.disposer} disposer-owned, ${accepted.escapes} ownership-transferred, ${accepted.allowlisted} allowlisted) ` +
     `across ${scanned} file(s).`,
 );
--- a/scripts/check-wasm-disposal.mjs
+++ b/scripts/check-wasm-disposal.mjs
@@ -244,8 +244,24 @@ function escapesScope(scope, name) {
    */
   const declaredLocally = (target) => {
     let found = false;
-    const scan = (n) => {
+    const scan = (n, depth = 0) => {
       if (found) return;
+      // Do NOT descend into nested function scopes. A callback parameter that
+      // happens to share the name — `[1].forEach((sharedProcessor) => …)` —
+      // is a different binding, and treating it as a local declaration marks a
+      // genuine module-level ownership transfer as a leak. That is a FALSE
+      // POSITIVE, the direction that gets a gate switched off, so it matters
+      // more than the over-acceptance cases this function also guards
+      // (CodeRabbit, #2129 review).
+      if (
+        depth > 0 &&
+        (ts.isFunctionDeclaration(n) ||
+          ts.isFunctionExpression(n) ||
+          ts.isArrowFunction(n) ||
+          ts.isMethodDeclaration(n))
+      ) {
+        return;
+      }
       // VariableDeclaration covers `let alias`; ParameterDeclaration covers
       // `function f(alias)`; BindingElement covers a destructured parameter or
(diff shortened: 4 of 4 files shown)


louistrue:
Filed the `using` blocker separately as #2130, with the build-both-ways evidence and a note that `tsc` is clean in both cases so it is invisible to typecheck.

louistrue:
**Do not merge this as it stands — my `useIfcLoader` fix is wrong, and @BIMvoice's #2127 found why.**

His disposer docblock states the coupling I missed:

> the data-model parser: on the main-thread fallback path `loadFile` passes the processor's raw handle to `IfcParser.parseColumnar` via `geometryProcessor.getApi()`. So a `finally { dispose() }` around the geometry block alone would free a handle the parser is still reading.

I verified it rather than taking it. `useIfcLoader.ts:1177`:

```ts
const parserWasmApi = geometryProcessor.getApi();
return new IfcParser().parseColumnar(buffer, {
  wasmApi: parserWasmApi ?? undefined,
  onSpatialReady: onPartialDataStore,
});
```

That runs inside `dataStorePromise`, which is not necessarily settled when the geometry block's `finally` fires. So my change frees a handle the parser can still be reading — **a use-after-free, which is worse than the leak it replaces.** The existing comment at :1263 flags the same sharing; I read past it.

His `parseScheduled` / `parseSettled` / `release` split is the correct shape, and the reason he gives for making the parser an *opt-in* consumer is the part I would not have got to: a load can fail between constructing the processor and scheduling the parse (engine init downloads the binary; the SAB allocation can throw), and in that window `release()` must free immediately rather than defer forever on a parse that will never run.

**So this PR should become gate-only**, sequenced after #2127 and #2128:

- Drop my `useIfcLoader` change in favour of #2127.
- Drop my treatment of `clash.ts` in favour of #2128, which actually disposes and resets `sharedProcessor` instead of accepting it as ownership transfer the way my gate does.

**And the gate has to change too, because right now it rejects his fix.** I ran it against #2127 + #2128 applied:

```
❌ apps/viewer/src/hooks/useIfcLoader.ts:1062  new GeometryProcessor()  -> `geometryProcessor` is never freed
```

It only recognises a literal `finally { name.dispose() }` and cannot see through `createGeometryProcessorDisposer(() => geometryProcessor.dispose())`. A gate that fails the *correct* fix is worse than no gate — it is the thing that gets checks disabled. Fixing that next: accept a handle handed to a disposer callback, not just a lexical `finally`.

Leaving this open for the gate; the two fixes belong to him.

BIMvoice:
The gate is the right answer and supersedes our two site fixes — I'd rather have this than #2127 and #2128. One finding on the viewer `finally` that I think is real, and it is the direction that is worse than the leak.

## `dispose()` can run while the main-thread parser still holds `getApi()`

`runMainThreadParser` hands the processor's handle to the parser (`useIfcLoader.ts:1169-1173`):

```ts
const parserWasmApi = geometryProcessor.getApi();
return new IfcParser().parseColumnar(buffer, { wasmApi: parserWasmApi ?? undefined, ... });
```

That parse settles `dataStorePromise` (`resolveDataStore(dataStore)` at `:1162`). Nothing inside the `try` awaits it — it is only ever `.then`-chained (`finalizePromise = dataStorePromise.then(...)` at `:1622`) or `.catch`-swallowed (`:1413`, `:1765`).

The only `await` on that chain is **outside** the `try/finally`:

```
:1787  } finally {
:1806      geometryProcessor.dispose();
:1807  }
...
:1819  if (target.kind === 'federated' && finalizePromise) {
:1820      await finalizePromise;
```

So on the main-thread parser path, `dispose()` at `:1806` runs while `parseColumnar` may still be mid-parse holding `parserWasmApi`. The `finally`'s safety note covers the mesh data —

> meshes are copied into JS and freed as they are extracted, so this only releases the long-lived handle

— but the parser's `wasmApi` reference is the long-lived handle, and it is exactly what is still in use.

I have not reproduced this in a browser, so treat it as an argued reading of the ordering rather than an observed crash. But the three line references above are checkable, and the worker path is not affected (its comment at `:1166-1168` is explicit that desktop loads do not share the instance with the parser — which is why only the fallback path is exposed).

**This is what #2127 was built around**, and I would rather hand you the mechanism than the PR: a small `createGeometryProcessorDisposer(dispose)` returning a trigger that frees exactly once and only after the parse chain settles, wired at `parseScheduled()` before the parse starts and `parseSettled()` in the chain's own `.finally()`. The chain rather than `dataStorePromise` specifically, because `onFullDataStore` (`:1154-1155`) returns early **without** calling `resolveDataStore` when the session is stale — so gating on the promise leaks precisely when a model swap should free. Extracted so it is testable without mounting the hook; the tests pin both directions (frees once, and does **not** free while a parse is in flight).

Take it, adapt it, or tell me the ordering is fine and I will drop it.

## On our two PRs

**#2128 — you are right that it is not a leak.** `clashCommand` has exactly one caller (`packages/cli/src/index.ts:259`) and the process exits, so disposing there is tidiness, not a fix. Accepting it as ownership transfer is the better call, and it avoids an allowlist entry. Yours supersedes it.

**#2127** is the same site as your viewer fix. Say the word on how you want both handled and I will act — I do not close our PRs unilaterally.

## The gate's own bug is the best part of this

> `NodeFlags.AwaitUsing` is `Using | Const` (6), so `flags & AwaitUsing` is non-zero for **every ordinary `const`** — the gate was accepting the entire repo while reporting success.

`✅ 26 using` against a codebase with essentially none is the exact shape we have been chasing all day: a check that reports success while measuring nothing. Keeping the reason as a comment on the function is right — a bitmask slip like that is invisible in review and disarms the check silently. Your four-direction mutation table (catches a leak, no false positive on `finally`, none on `using`, catches the reverted fix) is what makes the gate itself trustworthy.

BIMvoice:
## `using` viewer-build panic: reproduced, narrowed, not upstream-known

Ran the analysis you flagged (and found #2130 already covers the filing side — this is the repro/narrowing you asked for there).

### 1. Reproduced, on `upstream/main` (not just this branch)

Built the wasm/rust toolchain and full workspace fresh in a clean worktree off `upstream/main` (6722e08b), then converted a single, unrelated call site (`apps/viewer/src/hooks/useGridLines3D.ts`, `try/finally` → `using`):

```diff
-  const processor = new GeometryProcessor();
-  try {
-    await processor.init();
-    const verts = processor.parseGridLines(source);
-    return verts && verts.length > 0 ? verts : EMPTY_F32;
-  } finally {
-    processor.dispose();
-  }
+  using processor = new GeometryProcessor();
+  await processor.init();
+  const verts = processor.parseGridLines(source);
+  return verts && verts.length > 0 ? verts : EMPTY_F32;
```

`tsc --noEmit` on `apps/viewer`: clean, no output.

`pnpm --filter @ifc-lite/viewer build`:
```
✓ 5840 modules transformed.
rendering chunks...

thread '<unnamed>' panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/miette-7.6.0/src/handlers/graphical.rs:1159:22:
Formatting argument out of range
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
`$?` = **134** (SIGABRT). Same file, same panic string, same module count (5840) as your report on `fix/1959-wasm-disposal-gate` — this reproduces off plain `main` too, so it isn't specific to that branch's diff. `RUST_BACKTRACE=full` adds nothing usable; the shipped rolldown binary has no debug symbols (backtrace is all `_napi_register_module_v1` / `<unknown>`).

### 2. Narrowed: it's `using` anywhere in this build, not a WASM-handle thing

Reverted the `GeometryProcessor` change and instead dropped a **trivial `using` on a plain object** into `apps/viewer/src/main.tsx`, no WASM/geometry involved at all:

```ts
(function trivialUsingRepro(): void {
  const disposable: Disposable = { [Symbol.dispose]() {} };
  using _d = disposable;
  void _d;
})();
```

Typechecks clean. Build: **same panic, same line, same module count (5840)**, exit 134. So the trigger is `using` lowering somewhere in rolldown's chunk-rendering pass over this specific 5840-module/multi-chunk graph — not anything about `GeometryProcessor`, `Symbol.dispose`, or WASM. `await using` wasn't tested separately (no reason to expect it differs — same lowering path — but flagging that it's untested if it matters to you).

### 3. Minimal standalone repro: does NOT reproduce

Built an isolated project (not part of the monorepo) pinned to the exact same `vite@8.2.0` (→ `rolldown@1.2.2`, one patch ahead of the `1.2.1` this repo's lockfile resolves — registry has nothing newer for either package, so there's nothing to "upgrade to" right now):

- Plain `using` on a throwaway `Disposable`, default config → builds fine (`✓ built in 16ms`).
- Same, with `build.target: 'esnext'` (matches `apps/viewer/vite.config.ts`) → still fine.
- Same, padded to a 107KB single source file (to test the oxc-miette-style ">64KB file" theory below) → still fine.

So this isn't "any `using` panics rolldown" — it needs something about the viewer's actual build (5840 modules, `manualChunks`, `vite-plugin-wasm` + `vite-plugin-top-level-await`, multi-entry `rollupOptions.input`, or emitted-chunk size — `main-*.js` alone is 6.4–6.9MB) to trigger. I did not have time to bisect which of those is load-bearing; that's the next narrowing step if someone wants a filable minimal repro.

### 4. Upstream status

- The panic string and exact source location (`miette-7.6.0/src/handlers/graphical.rs:1159`, `Formatting argument out of range`) do **not** appear in any open `rolldown/rolldown` issue — searched both the GitHub issue search and rolldown's own tracker directly.
- The closest known relative is **oxc-project/oxc#9738** (closed, fixed by oxc#9888): `oxc-miette` panicking with the identical message ("Formatting argument out of range") in the identical function shape (`render_single_line_highlights` in `graphical.rs`), triggered by linting files **≥ 64KB**. That's a *different* crate (`oxc-miette` vs. plain `miette`) and a different call site (oxc's linter, not rolldown), but it's the same upstream bug class in the same diagnostic-rendering code path (an offset/index arithmetic overflow when miette formats a source snippet). It's suggestive of the mechanism, not a match — I could not get a >64KB single file to trigger it in the minimal repro (see #3), so if it's size-related here it's the concatenated/rendered chunk text during "rendering chunks", not a single input module.
- Net: this looks unreported against rolldown as its own issue. It reproduces reliably in-repo but I couldn't get it small enough in an afternoon to be a good upstream bug report (rolldown's own template asks for a minimal repro, and "our 5840-module app" isn't one).

### Options, for you to pick

1. **Do nothing beyond what #2129 already does.** `try/finally` in the viewer, `using` elsewhere. Zero risk, matches what's already merged in spirit.
2. **File upstream once someone bisects the minimal trigger** (candidates: `manualChunks`, the two wasm/tla plugins, multi-entry input, or total chunk byte size) — I'd guess this is a day of bisection, not an afternoon, given #3 above.
3. **Try newer rolldown/vite** — not currently possible, `1.2.2`/`8.2.0` are already `latest` on the registry as of this check.

I did not change any shipped code — worktree was reverted to clean and removed. Happy to spend more time on the bisection in #3 if that's the direction you want.

louistrue:
Good review — five findings, and every one is accurate. Taking two, documenting three, with the reasoning for the split.

## Taken

**MPL-2.0 header.** Correct: `LICENSE_HEADER.md` says every new source file. Worth noting `scripts/*.mjs` is inconsistent about it (10 of 22 carry it), so I checked before assuming — but this is a new file and the rule is unambiguous. Added after the shebang.

**Local aliases counted as ownership transfer.** greptile is right, and this one could hide a real leak:

```ts
let alias;
const sneaky = new GeometryProcessor();
alias = sneaky;          // same lifetime — not an escape
```

`escapesScope` accepted any assignment to a differently-named identifier. It now excludes bindings declared inside the same scope, so only an assignment to an *outer* binding — `clash.ts`'s module-level `sharedProcessor` — counts. Asserted both ways:

```
local alias        -> ❌ `sneaky` is never freed      (was silently accepted)
module-scope cache -> ✅ 5 ownership-transferred      (still accepted)
```

## Documented, not fixed — and why

Three findings, all correct, all the same shape:

- the `finally` is not required to **dominate** the construction (codex, CodeRabbit)
- receivers are matched by identifier **text**, no scope resolution, so a shadowed or captured binding can satisfy a check (greptile)
- `handedToDisposer` accepts a matching callback anywhere in the enclosing function without proving that callee owns the handle (CodeRabbit)

All three are **over-acceptance**: the gate can miss a leak, it cannot invent one. That asymmetry is the whole design, and it is not theoretical — **this gate already tried to reject #2127**, which was the correct fix, while my own change was a use-after-free. A gate that fails the right answer is how checks get disabled.

Closing them properly needs real scope resolution — a `ts.TypeChecker` over a full program rather than a single-file AST walk. That is a substantially larger tool than the problem justifies today, and I would rather ship a gate that provably catches the shape that actually happened twelve times than block on one that proves soundness.

All three limits are now listed in the file's docblock under "WHAT THIS DOES NOT PROVE", so the next person to trust this check sees its edges before they rely on it — rather than discovering them the way this review did.

## Status

Green: `29 constructions — 21 try/finally, 3 factory, 1 disposer-owned, 4 ownership-transferred`. #2127 is merged and its disposer shape is recognised. #2128 is the last piece and is independent of this passing.

louistrue:
Two findings. **Taking the second — it is a real false negative. Declining the first, with evidence.**

## Taken: parameter aliases were accepted as ownership transfer

Correct, and it is the same class as the local-alias hole from the last round — I fixed one shape of it and left another. `declaredLocally()` only inspected `VariableDeclaration`, so a parameter looked like an outer binding:

```ts
function f(alias) {
  const p = new GeometryProcessor();
  alias = p;          // param, not a VariableDeclaration -> "escapes" -> accepted
}
```

Reproduced against the gate before fixing: it reported `✅ … 5 ownership-transferred` with that leak present. Now covers `ParameterDeclaration` and `BindingElement`, so destructured params (`function f({ alias })`) are caught too.

Four cases asserted, both directions:

```
unmodified repo             -> accepted   (no false positive)
param alias (leak)          -> FLAGGED
destructured param alias    -> FLAGGED
module-scope cache          -> accepted   (real transfer still works)
```

Thank you for the fixture suggestion — writing the destructured case is what made me widen it to `BindingElement` rather than just adding `isParameter`.

## Declined: stripping issue identifiers and "review narrative"

The guideline cited is about **commercial confidentiality**, not internal references. `AGENTS.md:98`:

> Never name a client, customer, prospect, or partner organisation, and never describe a commercial relationship…

Internal issue numbers are neither. And the repo convention is emphatically the opposite — **831 source files** carry `(#NNNN)`-style references, densely, in code I did not write:

```
  18 (#1891)
  10 (#1993)
  10 (#1898)
   8 (#2007)
```

`AGENTS.md:97` also says everything here is world-readable and to write as if a competitor and a client are reading. `#1959` and `#2127` reveal nothing commercial; they are the fastest route from a line of code to the reasoning that produced it.

On the "retrospective review narrative" specifically: I would argue that is the most valuable content in the file. The comment explaining that `NodeFlags.AwaitUsing` is `Using | Const`, and that testing it made the gate accept the entire repo while printing a checkmark, exists because that bug actually shipped in this file for one commit. Reducing it to a neutral description of the analysis limit would drop the one detail that stops someone reintroducing it.

Same for the note that this gate initially rejected #2127 — that is the reason its acceptance rules are deliberately permissive, and without it the next maintainer tightens them and breaks the correct fix.

Happy to be overruled if the house reading of that guideline is broader than I think, but as written it is about client and partner names.

louistrue:
Taking the scope finding — it is the one direction that actually matters, and I reproduced it before fixing.

**The false positive:**

```ts
let moduleCache;
function f() {
  const p = new GeometryProcessor();
  moduleCache = p;                       // genuine ownership transfer
  [1].forEach((moduleCache) => { … });   // different binding, same name
}
```

```
❌ apps/viewer/src/hooks/useGridLines3D.ts:148  new GeometryProcessor()  -> `p2` is never freed
```

`declaredLocally()` descended into the callback, found a `ParameterDeclaration` named `moduleCache`, concluded the assignment target was local, and reported a real transfer as a leak.

This matters more than the over-acceptance cases the same function guards, and the asymmetry is the whole design of this gate: **it can miss a leak, it must never fail correct code.** A check that rejects the right answer is the one that gets switched off — and this gate already tried to reject #2127, which was correct while my own fix was a use-after-free.

The scan now stops at nested function boundaries. Four cases asserted, both directions:

```
unmodified repo                          accepted
module transfer + nested shadow          accepted   (was FLAGGED)
same-scope param alias (leak)            FLAGGED
same-scope let alias (leak)              FLAGGED
```

**Still declining the other thread** (strip issue identifiers and review narrative), for the reasons given above: `AGENTS.md:98` is about client/partner names and commercial relationships, not internal references, and 831 source files in this repo carry `(#NNNN)` citations. Resolving it as answered rather than leaving it to look unaddressed.

**One process note on my own work here.** My first attempt at this fix threaded a `depth` parameter but the string replacement matched nothing, so `depth` stayed `undefined`, `depth > 0` was never true, and the guard never ran — while the baseline still printed a green checkmark. I nearly took that as confirmation. It is the same shape as the `AwaitUsing` bitmask bug this file was born with, and the reason the docblock now says what the gate does *not* prove.

This is the last item on #1959. Once CI is green I will merge, which closes it — the two fixes (#2127, #2128) landed hours ago and only this gate was holding the issue open.

PR #2131, opened 2026-08-04
fix(drawing-2d): honour SVGExportOptions.padding as a min-margin guarantee

BIMvoice:
## Summary

`SVGExportOptions.padding` (documented as "Padding around drawing in mm", `packages/drawing-2d/src/svg-exporter.ts:48`) has been a silent no-op since the exporter's original commit (#150). `computeTransform` (`svg-exporter.ts:193-219`) derived `availableWidth`/`availableHeight` from `padding` at lines 203-204, but nothing — not the returned transform, not any caller — ever read them. A consumer of `export()` or `exportPolygons()` who sets `padding` gets exactly the same output as `padding: 0`.

Found while investigating #2119.

## Fix

`padding` is now honoured as a **minimum-margin guarantee**:
- If the drawing at its requested (exact, un-refit) `scale` already leaves at least `padding` mm of margin on the chosen paper, the scale is unchanged — this is the common case and matches existing behaviour exactly.
- Otherwise the effective scale is shrunk (never enlarged) just enough that the drawing's footprint fits inside `paperSize` minus `padding` on all sides. Centring is otherwise unaffected: padding is applied uniformly, so the padded area shares the paper's centre point.

This was chosen over removing the option because it's public/documented API and a consumer may already be passing it expecting an effect; removing it would be a needless breaking surface change.

## Behaviour change (see changeset)

`padding` defaults to `20` in both `export()` and `exportPolygons()`. That means **this can change output for callers who never pass `padding` at all**, not only callers who pass a non-zero value explicitly — whenever their drawing, at its requested scale, is closer than 20mm to the paper edge. `padding: 0` is unaffected except in the pre-existing edge case where a drawing already overflows the paper at the requested scale (previously silently ran off the page; now clamped to fit).

## Test plan

- [x] Added a RED test asserting padding shrinks scale and shifts coordinates inward on an overflowing drawing; failed before the fix (`worldToMm` clamp not applied)
- [x] Added a test pinning that padding has **zero** effect when the drawing already fits inside the padded area (the common case — no forced re-fit)
- [x] Added a test pinning `padding: 0` output is byte-identical to omitted `padding` for a non-overflowing drawing (compatibility guarantee)
- [x] `packages/drawing-2d` tests: 114 → 117 passing (baseline unaffected)
- [x] Mutation test: neutralising the clamp (python3 exact-substring replace, 1 occurrence) made the new RED test fail; restored by file copy
- [x] Changeset added (`@ifc-lite/drawing-2d`, patch) stating the behaviour change plainly


BIMvoice:
## Independent verification of `8fd69f8f`

Verified in an isolated worktree against `upstream/fix/svg-exporter-dead-padding` @ `8fd69f8f`, separate from the PR author's own worktree.

**1. Defect confirmed on parent commit (`8fd69f8f^` = `6722e08b`)**
Exported the same drawing (4m×6m box, A3 landscape, scale 1:100) with `padding` omitted, `0`, `20`, `100` — all four produce byte-identical SVG (same SHA1). Also confirmed with a drawing that grossly overflows a small paper (250m×100m box on A4 landscape): `padding` 0/20/80 are still byte-identical. `padding` was a genuine no-op regardless of value or overflow state, exactly as claimed.

**2. Mutation test on the fix**
Neutralised the clamp in `computeTransform` (forced `worldToMm = requestedWorldToMm` unconditionally, via exact-substring `python3` replacement, 1 occurrence, restored by file copy afterward — never `git checkout --`). Result: `shrinks the effective scale to keep the drawing within the padded area, and shifts coordinates inward` fails as expected (RED). The other two new padding tests still pass because they only assert "no shrink when it already fits," which the mutation doesn't break — expected and correct discrimination.

**3. Compatibility claim — verified by direct SVG diff, not by trusting the test**
Ran the *same* drawing through the parent commit's exporter and the fixed exporter (swapped `svg-exporter.ts` in place via file copy, not `git checkout`, restored after):
- Ordinary drawing (comfortably fits, well inside 20mm margin): fixed `padding: 0`, fixed omitted (defaults to 20), and fixed `padding: 100` are all byte-identical to each other **and** to the parent commit's output. Compatibility claim holds for the common case.
- Overflowing drawing at `padding: 0`: fixed output *does* differ from parent (parent silently overflowed the page; fixed clamps to fit) — exactly the documented edge case, no more and no less.

**4. Semantics check**
On the overflowing 250m×100m / A4-landscape case, effective `worldToMm` at padding 0/20/80 computed to 1.188 / 1.028 / 0.5 mm-per-world-unit — strictly decreasing, never exceeding the requested 10 (1:100). The drawing's bounds-center point maps to the exact paper center in all three cases (148.5, 105 — unchanged), confirming centering is preserved. The binding-edge coordinate lands exactly `padding` mm from the paper edge in each case (offset 0 / 20 / 86mm, i.e. exactly meeting, not exceeding, the requested margin). No enlargement observed in any case tested.

**Test counts / typecheck**
- `packages/drawing-2d`: 10 test files, 117 tests, all passing.
- `pnpm typecheck` (repo-wide, via turbo): 34/34 tasks passing, 0 errors.

No defects found. The dead-option claim, the mutation-covered fix, the compatibility claim, and the shrink-only/centered semantics all check out against direct evidence (byte diffs and computed transforms), not just the PR's own test assertions.

louistrue:
Both greptile findings are real, and the first one I would treat as more serious than the bug this PR fixes.

## The title block can lie about the scale

`computeTransform` clamps the scale so the drawing honours the padding:

```ts
let worldToMm = requestedWorldToMm;
if (size.x > 0 && availableWidth  > 0) worldToMm = Math.min(worldToMm, availableWidth  / size.x);
if (size.y > 0 && availableHeight > 0) worldToMm = Math.min(worldToMm, availableHeight / size.y);
```

but the title block prints the **requested** name, unconditionally:

```ts
<text …>Scale: ${scale.name}</text>
```

So a sheet that got clamped from 1:100 to, say, 1:127 still reads **"Scale: 1:100"**. That is a confidently wrong document: someone scaling a dimension off the printout — which is the entire reason a scale bar and a scale label exist on a construction drawing — gets an answer that is wrong by the clamp ratio, with nothing on the sheet indicating it.

Compared to the bug being fixed here (padding ignored, geometry too close to the edge), that one is visible and self-correcting; this one is invisible and propagates into measurements.

Options, in the order I would consider them:
1. **Print the effective scale** — derive the label from the final `worldToMm` rather than `scale.name`. Honest, no API change.
2. **Refuse to clamp silently** — return the clamp in the transform result so the caller can decide, and surface it the way #2105 surfaced a refused map conversion.

What I would avoid is clamping the geometry and leaving the label alone, which is the current state.

## Impossible margins disable the guarantee

The `availableWidth > 0` / `availableHeight > 0` guards skip the clamp entirely when `padding * 2 >= paperSize.width` (or height). The intent is clearly to avoid dividing by a non-positive extent — but the effect is that an over-large padding **disables the very guarantee this PR adds**, and the drawing renders at the full requested scale, overflowing the paper.

Failing loudly on an impossible padding is better than silently ignoring it: either clamp `padding` to something the paper can hold and warn, or reject the export. Silently falling back to "no padding at all" is the shape this PR exists to remove.

Both are small, and the changeset is the right place to record whichever semantics you pick — "padding is a minimum-margin guarantee" is a contract, and both of these are cases where it currently is not one.

BIMvoice:
Pushed 53709c3c addressing both findings.

**Finding 1 — the title block lying about scale.** Went with your preferred option (1): the "Scale:" label is now derived from the *effective* `worldToMm` rather than the requested `scale.name`, whenever `computeTransform` actually clamped. `Transform2D` now carries a `clamped: boolean` (set by comparing `worldToMm !== requestedWorldToMm` — no floating error is introduced by the `Math.min` comparison itself, so this is exact, not a tolerance check). `scaleLabel()`:
- returns `scale.name` unchanged when `!clamped` — no float round-trip on the common path, so `1:100` stays exactly `1:100`, never `1:100.0000001`.
- when clamped, computes `1000 / transform.scale`, rounds to 2 decimal places, and strips trailing zeros (`1:973`, `1:127.3`, never `1:127.30000001`).

I didn't also thread the clamp back through the public API (your option 2) — (1) alone removes the lying document, and adding a return-value shape change felt like scope creep for this PR; happy to open that as a separate follow-up if you want it.

**Finding 2 — impossible padding.** Chose "clamp `padding` and warn" over "reject." `computeTransform` now caps `padding` at the largest value the paper's *shorter* dimension can still hold, leaving a 1mm minimum drawable sliver, and logs `console.warn` with the requested/paper/clamped numbers when that kicks in. Reasoning: this is a published package (`@ifc-lite/drawing-2d`), and a caller passing a large `padding` today gets a working (if extreme) export — throwing would turn that into a runtime break on upgrade with no way to opt out. The warn makes it non-silent, which was the actual complaint; a caller who wants stricter behavior can watch for the warning or clamp `padding` themselves before calling.

**Tests (RED-first):**
- `prints the exact requested scale name when the drawing is not clamped` / `prints the effective (clamped) scale, matching the actually rendered geometry` — the second derives the "actual" scale independently from the emitted `<line>` coordinates (not from internal transform fields) and compares it against the parsed label, so it can't pass by coincidence.
- `clamps an oversized padding instead of silently skipping the clamp, and warns` — asserts the rendered geometry's derived `worldToMm` is below the unclamped value and that `console.warn` fired mentioning "padding".

Mutation-tested each fix (python3 exact-substring swap, restored by file copy, not `git checkout --`):
| Mutation | Result |
|---|---|
| `scaleLabel`: force `!transform.clamped` branch always taken | new label test fails |
| Padding clamp: `if (padding > maxPadding)` → `if (false && ...)` | new padding test fails |
| **Re-run of this PR's original acceptance mutation**: neutralize the padding clamp entirely (`worldToMm` forced to `requestedWorldToMm`) | still fails the pinned `shrinks the effective scale...` test, plus both new tests — not weakened |

`packages/drawing-2d`: 117 → 120 tests, all green. `pnpm exec tsc --noEmit -p packages/drawing-2d` clean.

Changeset updated with both behavior changes.

louistrue:
Both fixed, both mutation-verified:

```
### always print the requested name (the lying label)
     × prints the effective (clamped) scale, matching the actually rendered geometry

### restore the old guard (impossible padding disables the clamp)
     × clamps an oversized padding instead of silently skipping the clamp, and warns
```

Baseline 8/8, restored 8/8.

**The label fix is better than what I suggested.** I proposed deriving the label from the effective scale; you kept `scale.name` verbatim on the unclamped path and only re-derive when `transform.clamped`. Your reason is the one I would have missed:

> no floating-point round-trip on the common path, so "1:100" never regresses to something like "1:100.0000001"

Deriving unconditionally would have introduced a cosmetic defect into every ordinary export to fix a rare one. Gating on `clamped` gets both.

**The padding fix reframed the problem correctly.** I framed it as "the guard skips the clamp"; you fixed it as "padding must never consume the whole sheet" — clamping `padding` itself against `maxPadding` and warning. That keeps the guarantee holding rather than trading one silent lapse for another, and the comment names it exactly:

> the exact failure mode this feature exists to remove, just triggered by an oversized padding instead of an absent one

Reserving `MIN_AVAILABLE_MM` so there is still a usable area is the detail that stops the clamped value from being degenerate.

Worth noting what this PR is, end to end: it started as "padding is ignored", and the review surfaced that honouring padding *silently changes the scale* — so a fix for a margin bug would have shipped a drawing that lies about its scale. On a document people measure off, that second bug was the worse one. Good outcome.

All 5 core lanes green, 0 failures. Merging once the mergeable state settles.

PR #2132, opened 2026-08-04
test(viewer): pin the direct SVG export's world-mm to paper-mm transform

BIMvoice:
## Summary

Follow-up to the PR #2119 investigation, which compared four "world metres -> paper mm at scale N" transforms in the export code and correctly concluded they should not be consolidated (four different contracts), but flagged that the fourth — the inline arithmetic in `generateExportSVG` (`apps/viewer/src/hooks/useDrawingExport.ts`) — was untested because it lived inline in a hook with `useViewerStore` dependencies, unlike the codebase's pattern of extracting pure math into small testable modules (e.g. `dxfExportGeoref.ts`).

Confirmed the description at `apps/viewer/src/hooks/useDrawingExport.ts:212-231,243-249,282-284` (pre-extraction line numbers): this transform is distinctive because it does **not** map points into paper-mm space at all — path data is emitted directly in (axis-flipped) world units, and the mm-per-world-unit scaling is delegated entirely to the ratio between the SVG's `width`/`height` (mm) and its `viewBox` (world units). Padding is 10% of `max(width, height)` of the *unpadded* drawing bounds, in world units, not mm.

- Extracted the pure arithmetic into `apps/viewer/src/hooks/svgExportViewport.ts` (`computeSvgExportViewport`, `svgExportMmToWorld`), following the `dxfExportGeoref.ts` precedent. The hook now calls it — no behaviour change.
- Pinned it in `apps/viewer/src/hooks/svgExportViewport.test.ts`: the mm-per-world-unit relationship at two scales (1:100, 1:200) plus the falsy-scale `|| 100` fallback, the axis-dependent Y/X flip for all three axes (`down`, `front`, `side`), and the 10%-of-max-extent padding in world units (independent of scale).

## Equivalence method (pure refactor safety net)

Before wiring the hook to the new module, I ran the *original* inline formula (copied verbatim into a throwaway script) side by side with `computeSvgExportViewport` over 6 representative inputs — typical bounds at two scales, a degenerate zero-size bounds, a falsy `scale` (0) exercising the `|| 100` default, and all three axes — asserting `assert.strictEqual` (exact bit-for-bit match, not "close enough") on every field. All 6 matched exactly before the hook was touched.

## Mutation testing

Two targeted mutations via python3 exact-substring replacement (occurrence count asserted `== 1`, region re-read, restored by file copy — never `git checkout --`):
- Perturbed the scale factor (`widthMm` off by 1.5x) — killed 3/8 tests.
- Flipped the Y-flip sign (`axis !== 'down'` -> `axis === 'down'`) — killed 3/8 tests.

Both restored cleanly; final state re-verified green (8/8).

## Test plan

- [x] `apps/viewer`: baseline 2709 tests (2707 pass, 2 skipped) -> final 2717 tests (2715 pass, 2 skipped, 0 fail) — +8 tests / +2 suites, all new.
- [x] `pnpm typecheck` — clean.
- [x] No changeset: `@ifc-lite/viewer` is `"private": true` (verified in `package.json`, not assumed).

### DIFF
--- a/apps/viewer/src/hooks/svgExportViewport.test.ts
+++ b/apps/viewer/src/hooks/svgExportViewport.test.ts
@@ -1,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The direct SVG export's world-metres -> paper-mm viewport arithmetic
+ * (`useDrawingExport.ts`'s `generateExportSVG`, extracted here per the PR
+ * #2119 follow-up: this was the fourth of four "world metres -> paper mm at
+ * scale N" transforms, and the only one not yet pinned by a test — it was
+ * inline, unexported, inside a hook with `useViewerStore` dependencies).
+ *
+ * Distinctive contract, unlike the other three: this transform does NOT map
+ * points into paper-mm space. Path data is emitted in raw (axis-flipped)
+ * WORLD units; the mm-per-world-unit scaling is entirely delegated to the
+ * ratio between the SVG's `width`/`height` (mm) and its `viewBox` (world
+ * units) — a renderer does the actual scaling when it lays the `viewBox`
+ * into the `width`x`height` box.
+ *
+ * Equivalence with the pre-extraction inline code was established out of
+ * band (not committed) by running the OLD formula and `computeSvgExportViewport`
+ * side by side over 6 representative inputs (typical bounds at two scales,
+ * a degenerate zero-size bounds, a falsy `scale` exercising the `|| 100`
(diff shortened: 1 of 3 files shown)


louistrue:
Merging — viewer tier, green on all 5 core lanes, 0 unresolved.

This closes the gap #2119's review implied: four "world metres → paper mm at scale N" transforms, and this was the only one not pinned, because it was inline, unexported, and inside a hook with `useViewerStore` dependencies. Extracting it is what made it testable at all.

`assert.strictEqual` on the exact millimetre values (120/70, viewBox 12/7, effectiveScale 100) is the right strength for arithmetic like this — a tolerance-based assertion on a scale transform can pass while being off by a factor that matters on paper.

Worth keeping the docblock note that this transform's contract *differs* from the other three. That is exactly the kind of thing that gets "unified" later by someone who assumes four similar-looking functions should share an implementation.

PR #2133, opened 2026-08-04
docs: warn against bypassing turbo for typecheck/test

BIMvoice:
## Investigation

An agent hit phantom typecheck errors today (`apps/viewer/src/e2e/laz-wasm-probe-entry.ts` missing `probeLazPerfWasmLoad`) attributed to a stale `packages/pointcloud/dist`. This PR investigates whether `pnpm typecheck` (the documented, canonical command) has a real gap, or whether the trap is elsewhere.

**Verdict: no gap in `turbo.json`.** The root `pnpm typecheck` / `pnpm test` commands already correctly force a fresh build first. The trap only reproduces when those commands are bypassed.

### Evidence

`turbo.json` already declares:
```json
"typecheck": { "dependsOn": ["^build"], "outputs": [] },
"test": { "dependsOn": ["build"], "cache": false }
```
(`test`'s own `build` task itself declares `dependsOn: ["^build"]`, so the transitive graph is the same.)

Confirmed with a fresh worktree off `upstream/main` (`pnpm install`, no `dist/` anywhere):
```
$ pnpm exec turbo run typecheck --filter=@ifc-lite/viewer... --dry=json | jq '.tasks[].taskId'
...
"@ifc-lite/pointcloud#build"
"@ifc-lite/pointcloud#typecheck"
...
"@ifc-lite/viewer#typecheck"
```
`pointcloud#build` is correctly scheduled before `viewer#typecheck`. Running `pnpm typecheck` from root with all `dist/` cleared produced zero errors and turbo restored every package's `dist/` from cache before typechecking.

**The trap reproduces when turbo is bypassed.** With `packages/pointcloud/dist` removed, running `tsc --noEmit` directly inside `apps/viewer` (equivalent to `pnpm --filter @ifc-lite/viewer typecheck`, since that package's own script is literally `tsc --noEmit`) throws exactly the reported symptom:
```
src/hooks/ingest/pointCloudIngest.ts:23:8 - error TS2307: Cannot find module '@ifc-lite/pointcloud' or its corresponding type declarations.
src/e2e/laz-wasm-probe-entry.ts:18:38 - error TS2307: Cannot find module '@ifc-lite/pointcloud' or its corresponding type declarations.
```
Root cause: `@ifc-lite/pointcloud` has no `tsconfig.json` path-mapping override (unlike most packages, which the root `tsconfig.json` maps straight to `dist/*.d.ts` anyway), so it resolves via plain node_modules → `package.json#types` → `dist/index.d.ts`. Any package-local `tsc`/`vitest` invocation, or an editor's TS server, resolves workspace siblings off `dist/` and turbo's dependency graph never gets a chance to run.

### The "Tests no tests" claim

Tried to reproduce a vacuous pass in `packages/parser` (no `--passWithNoTests`) with all `dist/` cleared, both via `pnpm --filter parser test` and `pnpm --filter @ifc-lite/parser test`: both **fail loudly**, exit 1, `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL`, with `Failed to resolve entry for package "@ifc-lite/data"` — not a silent no-op. Root `pnpm test` (turbo) also correctly built every upstream dependency first (`@ifc-lite/data#build`, `@ifc-lite/encoding#build`, etc. all appear ahead of `@ifc-lite/parser#test` in the dry-run graph). Could not reproduce the specific "no tests" symptom against `packages/parser` in this repo state — flagging as unconfirmed rather than guessing further.

**Separate and worth surfacing:** `packages/mutations` and `packages/drawing-2d` run `vitest run --passWithNoTests` (both have real, current test files — 8 and 10 respectively). That flag makes vitest exit 0 if it ever matches **zero** test files, independent of the dist-staleness path above. I couldn't force a zero-file match in the current repo state (unbuilt-dep collection failures fail loudly instead, as shown above), so this is a latent risk rather than a confirmed live one — not fixing it since I have no reproduction, just noting it for awareness.

Also worth noting: AGENTS.md already documents one instance of this exact vacuous-pass class — `pnpm test:wasm-contract` "skips clean (exit 0) if the wasm runtime isn't built" — so the pattern is known to maintainers, just not generalized.

## Fix

Documentation only (`AGENTS.md`) — no `turbo.json` change, since the dependency declaration is already correct and I don't want to manufacture a config change against working infrastructure. Added one bullet to the Commands section: always use the root `pnpm typecheck` / `pnpm test`, never a package-local script or bare editor TS server, when diagnosing an error — and why (dist-vs-src resolution, with the pointcloud example). No changeset: no published package's runtime behavior changed.

## Test plan
- [x] Fresh `upstream/main` worktree, `pnpm install`, all `dist/` cleared
- [x] `pnpm exec turbo run typecheck --filter=@ifc-lite/viewer... --dry=json` shows `pointcloud#build` ordered before `viewer#typecheck`
- [x] `pnpm typecheck` from repo root: 0 errors, `dist/` restored from cache
- [x] `tsc --noEmit` run directly in `apps/viewer` with `pointcloud/dist` removed: reproduces the reported phantom errors
- [x] `pnpm --filter parser test` / `pnpm --filter @ifc-lite/parser test` with all `dist/` cleared: fails loudly (exit 1), not vacuously


BIMvoice:
Verified the two structural claims in this PR independently, and correcting something I got wrong myself.

## No turbo gap — confirmed

```json
"typecheck": { "dependsOn": ["^build"], "outputs": [] },
```

So `pnpm typecheck` cannot see a stale `dist`. The trap is real but it is a **workflow** one: `@ifc-lite/pointcloud` has no path-mapping override, so it resolves through `node_modules → package.json#types → dist/index.d.ts`. A package-local `tsc --noEmit`, a bare `vitest`, or an editor TS server never enters turbo's graph — and then reports errors naming a real file and a real symbol in source that is already correct. Documenting it is the right fix; there is no config change to make.

## Correcting myself

I previously described a fresh-worktree run as reporting `Tests no tests` in a way that "reads like a pass". Re-reading my own output, it also printed `Test Files 1 failed (1)` — so it **did** fail, and I misread it at a glance rather than it passing silently. This PR could not reproduce a silent pass for `packages/parser` either; with `dist/` cleared it fails loudly with `Failed to resolve entry for package "@ifc-lite/data"` and exit 1. My earlier characterisation was wrong and I would rather say so than leave it standing as a claim about the harness.

## One latent risk that IS real, and I would rather you saw it

Two published packages run vitest with a flag that turns "found nothing" into success:

```
packages/drawing-2d/package.json:18:  "test": "vitest run --passWithNoTests"
packages/mutations/package.json:18:  "test": "vitest run --passWithNoTests"
```

Both have real test files today, so nothing is broken now. But the flag means that if a glob, a rename, or a config change ever stops matching, those suites report green having run zero tests — the same shape as the `Vercel`-absent CI rollup and the `NodeFlags.AwaitUsing` bitmask on #2129.

`AGENTS.md` already documents one instance of this class — `pnpm test:wasm-contract` "skips clean (exit 0) if the wasm runtime isn't built" — so the pattern is known, just not generalised.

I did not change either package: I could not force a zero-file match in the current tree, and removing the flag is a judgement about why it was added (possibly for a package that once had no tests). Worth a look, but not something to change under an unreproduced risk.

louistrue:
Merging — and this is the single most useful doc change in the queue, because I walked into this trap repeatedly today.

Every time I typechecked or tested a package directly in a fresh worktree I got a wall of `Cannot find module '@ifc-lite/extensions'` / `Failed to resolve entry for package` errors — 233 of them in one case — that name real files and symbols and describe no bug whatsoever. Each time the fix was `pnpm build` of the dependency chain first. Once I nearly read an empty vitest run as a passing mutation test for the same reason.

The sentence that does the work is the mechanism, not the rule:

> `turbo.json` makes `typecheck`/`test` depend on `^build`, but that dependency graph only fires through turbo; a package-local script or IDE process resolves workspace siblings straight off their `dist/`

That is what makes it actionable rather than folklore — it explains why the same command passes in CI and fails locally, so the next person diagnoses it in seconds instead of chasing a phantom type error.

Worth having the `@ifc-lite/pointcloud` exception called out too (`package.json#types` rather than a root path-map), since that is the one that would otherwise look like a counter-example.

PR #2134, opened 2026-08-04
test(geometry): fix two mutation-verified vacuous tests

BIMvoice:
## Summary

Two vacuous tests in `packages/geometry` — both mutation-verified (a targeted production-code mutation left the suite green).

**1. `geometry-processor-streaming.test.ts` — dead private-state poke**

The test did `(GeometryProcessor as any).largeFileByteStreamingThreshold = 1` to claim it forces a "large file" streaming path. That private static (`index.ts:262`) is declared and never read anywhere in production code. `processStreaming()`/`processStreamingUnlocked()` run the byte-based prepass/batch path unconditionally, with no size gate at all — the real size gate lives in the unrelated `processAdaptive()`, as its own local `sizeThreshold` (default 2MB).

Verified: deleting the set/restore lines leaves the file at 2/2 passed, identical to baseline.

Fix: removed the dead poke, renamed/re-documented the `describe` block to say what it actually exercises (`processStreaming`'s behaviour against a mocked WASM bridge — event sequence, `.free()` discipline, re-entrancy guard).

On the field itself: `largeFileByteStreamingThreshold` looks like dead code (declared, assigned once, read nowhere), not an unwired gate — there's no code path anywhere that reads it, not even under a different name/getter. Left it in place since deleting a production field wasn't asked for here and is a separate call; noting it for whoever picks that up.

**2. `diagnostics.test.ts` — tests that never touch the code they name**

`describe('the streaming complete event payload (geometry.worker.ts:901)')` (3 tests) defined a local `buildCompleteEvent()` that reimplemented the worker's conditional diagnostics spread, and never imported `geometry.worker.ts`. The `:901` line reference was stale too (current call site was line ~1297 pre-fix).

Verified (mutation): replacing `...(session.diagnostics ? { diagnostics: session.diagnostics } : {})` with `diagnostics: session.diagnostics ?? undefined` — which breaks the "omit the key entirely on a clean load" contract those tests claim to pin — left the suite at 10/10 passed.

`geometry.worker.ts` can't be imported directly under vitest: it assigns `self.onmessage` at module load time, and `self` is undefined in vitest's default (node) environment — confirmed by probing the import directly (`ReferenceError: self is not defined`).

Fix: extracted the payload-building logic into `buildGeometryWorkerCompleteMessage(totalMeshes, diagnostics)` in `diagnostics.ts` (same extract-and-share shape as `dxfExportGeoref.ts`/`svgExportViewport.ts`), and had `geometry.worker.ts`'s `emitSessionEnd` call it instead of inlining the spread. The tests now import and call the real function. Also fixed the stale `:901` reference in the `describe` name.

## Mutation-test evidence (RED-first, python3 exact-substring)

- Mutation: `diagnostics: diagnostics ?? undefined,` replacing `...(diagnostics ? { diagnostics } : {}),` in `diagnostics.ts`'s `buildGeometryWorkerCompleteMessage`. Occurrence count printed to stderr and asserted `== 1` before replacing.
- RED: `omits diagnostics entirely on a clean load` failed (`AssertionError: expected { type: 'complete', …(2) } to not have property "diagnostics"`) — 9/10 passed, 1 failed.
- Restored by re-running the exact-substring replace back to the original (single occurrence, asserted).
- GREEN: 10/10 passed after restore; `git diff --stat` on `diagnostics.ts` shows only the intended additions, no mutation residue.

## Test counts

- Baseline (with wasm built via `scripts/build-wasm.sh`): 20 files / 202 tests, matches the expected baseline exactly.
- After this change: 20 files / 202 tests, all passing (both target files pass: 2/2 and 10/10).
- `pnpm typecheck` (full workspace, wasm-free lane): passes.
- `pnpm knip`: no new flags on the touched files.

## Test plan

- [x] RED-first mutation reproduced against the rewritten `diagnostics.test.ts`, confirmed to fail
- [x] Restored via exact python3 substring replace (not `git checkout`), confirmed GREEN
- [x] `pnpm --filter @ifc-lite/geometry test` (via `vitest run`): 20/20 files, 202/202 tests
- [x] `pnpm typecheck`: passes across the workspace
- [x] Changeset added (`@ifc-lite/geometry` patch) since `diagnostics.ts`/`geometry.worker.ts` (production code) changed, not just tests

## Summary by CodeRabbit

- **Refactor**
  - Improved consistency in geometry streaming completion messages.
  - Completion responses continue to include mesh totals and optional diagnostics without changing existing behavior.

- **Tests**
  - Expanded coverage for geometry worker completion payloads.
  - Strengthened streaming geometry tests for event sequencing, batch processing, resource cleanup, and repeated execution.

- **Documentation**
  - Added a patch-release note for the geometry diagnostics updates.

### DIFF
--- a/.changeset/geometry-vacuous-diagnostics-payload.md
+++ b/.changeset/geometry-vacuous-diagnostics-payload.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/geometry": patch
+---
+
+Extract the streaming `complete` event's diagnostics payload builder (`geometry.worker.ts`'s `emitSessionEnd`) into a standalone `buildGeometryWorkerCompleteMessage` in `diagnostics.ts`, and have the worker call it instead of inlining the conditional spread. No behaviour change: the emitted payload is identical (diagnostics still omitted entirely, not sent as `undefined`, on a clean load). This lets `diagnostics.test.ts` exercise the real production logic directly — the worker module cannot be imported under vitest (it assigns `self.onmessage` at module load time), so the payload shape is now factored out where a plain unit test can reach it.
--- a/packages/geometry/src/diagnostics.test.ts
+++ b/packages/geometry/src/diagnostics.test.ts
@@ -3,7 +3,11 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import { describe, it, expect } from 'vitest';
-import { mergeGeometryDiagnostics, type GeometryDiagnostics } from './diagnostics.js';
+import {
+  mergeGeometryDiagnostics,
+  buildGeometryWorkerCompleteMessage,
+  type GeometryDiagnostics,
+} from './diagnostics.js';
 
 function make(partial: Partial<GeometryDiagnostics> = {}): GeometryDiagnostics {
   return {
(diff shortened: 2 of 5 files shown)


louistrue:
Verified independently and merging. This is the discipline working exactly as intended — you found two tests that could not fail, by mutation, and both claims check out.

**The extraction is wired, not decorative.** `geometry.worker.ts` imports and calls `buildGeometryWorkerCompleteMessage` at :1298, and the inline conditional spread is gone (`0` remaining). So the tests now reach the real production path rather than a local reimplementation of it.

**The mutation kills it:**

```
occurrences=1
     × omits `diagnostics` entirely on a clean load
      Tests  1 failed | 9 passed (10)
```

Restored: 10/10. Before this PR that same mutation left the suite green.

Two details worth calling out because they are the difference between a real mutation test and a reassuring one:

**You asserted the occurrence count before replacing.** A regex that matches nothing produces a green run that reads exactly like "the fix is unpinned" — I made that precise mistake earlier today and only caught it because a number looked wrong. Printing `occurrences=1` and asserting it is what makes the RED meaningful.

**The dead-poke diagnosis is right and worth more than the fix.** `largeFileByteStreamingThreshold` is declared, assigned once, and read nowhere — `processStreaming` has no size gate at all; the real one is `processAdaptive`'s local `sizeThreshold`. So the test's `(GeometryProcessor as any).largeFileByteStreamingThreshold = 1` was configuring nothing, and the `describe` name claimed a "large file" path that did not exist. Deleting the poke leaving 2/2 green is the proof.

Leaving the field itself in place was the right call for this PR, and flagging it separately is right too — deleting a production field is its own decision, not a test fix.

The stale `:901` line reference in the `describe` name is a small thing that matters: a comment pointing at the wrong line is how a reader concludes the test covers something it does not.

PR #2135, opened 2026-08-04
test(parser): close two mutation-verified coverage gaps

BIMvoice:
## Summary

Two mutation-verified coverage gaps in `packages/parser`, both confirmed by directly applying the mutation, watching the full suite stay green, then adding a test that fails under the mutation.

### Gap 1 — IFC4X3 schema detection is order-dependent and unpinned

`packages/parser/src/columnar-parser.ts` `detectSchemaVersion()` classifies the `FILE_SCHEMA` header token with an ordered `.includes()` ladder:

```ts
if (headerText.includes('IFC5')) return 'IFC5';
if (headerText.includes('IFC4X3')) return 'IFC4X3';
if (headerText.includes('IFC4')) return 'IFC4';
if (headerText.includes('IFC2X3')) return 'IFC2X3';
```

`'IFC4X3'.includes('IFC4')` is `true`, so the ordering is load-bearing: swapping the IFC4X3 and IFC4 lines makes every IFC4X3 file silently misdetect as IFC4. No existing test drove a real IFC4X3-headed buffer through `parseLite`/`parseColumnar` and asserted `schemaVersion === 'IFC4X3'` — the other IFC4X3 references in the suite (`inheritance-chain-equivalence.test.ts`, `instantiable-across-schemas.test.ts`) exercise entity-dictionary tables, not header detection.

Added `test/schema-version-detection.test.ts`, driving `ColumnarParser.parseLite()` (the public entry point) with a minimal STEP buffer per schema token and asserting `store.schemaVersion`. Covers IFC5, IFC4X3, IFC4, IFC2X3, and the unrecognized-token default fallback. `detectSchemaVersion` stayed module-private — the whole point of the ladder is what a real caller observes on the resulting store, so there was a reasonable public path and no export was needed.

**Acceptance mutation**: swap the IFC4X3/IFC4 lines. Applied via exact-substring replacement, confirmed the new "detects IFC4X3" test fails (`expected 'IFC4' to be 'IFC4X3'`), restored the file by copy from `git show HEAD:...`, confirmed green again.

### Gap 2 — the deferred-atom fallback in `refFromStore` is untested

`packages/parser/src/on-demand-extractors.ts`:

```ts
function refFromStore(store: IfcDataStore, id: number) {
    return store.entityIndex.byId.get(id) ?? store.deferredEntityIndex?.get(id);
}
```

Used when resolving material property sets (`getMaterialPropertyIndex`) for both the `*MaterialProperties` entity and the `IfcProperty` atoms it references — the fallback exists because large files parsed with `deferPropertyAtomIndex: true` keep property atoms out of the primary `entityIndex.byId`, indexed instead into `deferredEntityIndex`. Dropping the `?? store.deferredEntityIndex?.get(id)` fallback survived the full suite; `grep -rl deferredEntityIndex test/` returned nothing before this PR.

Added `test/deferred-entity-index-material-properties.test.ts`, constructing a minimal store where the referenced `IfcPropertySingleValue` atom lives *only* in `deferredEntityIndex` (with an explicit sanity assertion that it's absent from `entityIndex.byId`), then calling the public `extractMaterialPropertiesForMaterialId()` and asserting the property resolves.

**Acceptance mutation**: remove the `?? store.deferredEntityIndex?.get(id)` fallback. Applied via exact-substring replacement, confirmed the new test fails (`expected [] to have a length of 1 but got +0`), restored by file copy, confirmed green again.

### Gap 3 (skipped)

`packages/data/src/property-table.ts:340-341`'s `si >= 0 && si < strings.count` guard was flagged as a third, marginal survivor (`<=` also survives). Skipped per the brief: it's reachable only from corrupted transport data, `StringTable.get()` already bounds-checks and returns `''`, and the only observable difference is `''` vs `null` at exactly `si === strings.count` — cheap to state, not cheap to fixture without contorting internal `PropertyTable`/`StringTable` state. Not included, no changes to `packages/data`.

## Test plan

- [x] Baseline: 370/370 across 45 files (`packages/parser`)
- [x] RED: Gap 1 mutation (swap IFC4X3/IFC4 lines) makes the new "detects IFC4X3" test fail
- [x] RED: Gap 2 mutation (drop deferredEntityIndex fallback) makes the new test fail
- [x] Both mutations restored by file copy (`git show HEAD:<path>`), not `git checkout --`
- [x] GREEN: 376/376 across 47 files after restore
- [x] `pnpm typecheck` passes repo-wide
- [x] Test-only change — no changeset (no source files touched, no exported surface changed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when resolving material properties referenced indirectly in IFC data.
  * Improved schema-version detection across supported IFC formats, including fallback handling for unrecognized schema identifiers.

* **Tests**
  * Added coverage to help ensure material-property extraction and IFC schema detection continue working consistently across supported scenarios.

### DIFF
--- a/packages/parser/test/deferred-entity-index-material-properties.test.ts
+++ b/packages/parser/test/deferred-entity-index-material-properties.test.ts
@@ -1,0 +1,103 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Coverage for the `deferredEntityIndex` fallback in `refFromStore()`
+ * (on-demand-extractors.ts): `store.entityIndex.byId.get(id) ??
+ * store.deferredEntityIndex?.get(id)`.
+ *
+ * On large files parsed with `deferPropertyAtomIndex: true`, property atoms
+ * (IfcPropertySingleValue, etc.) are indexed separately into
+ * `deferredEntityIndex` and are deliberately absent from the primary
+ * `entityIndex.byId`. Material-property-set resolution
+ * (`getMaterialPropertyIndex` in on-demand-extractors.ts) walks
+ * *MaterialProperties entities and their referenced IfcProperty atoms through
+ * `refFromStore()`, so it must fall back to the deferred index or every
+ * property on a deferred-index model silently resolves to nothing.
+ *
+ * No existing test builds a store with a populated `deferredEntityIndex`
+ * (`grep -rl deferredEntityIndex test/` returns nothing before this file),
+ * so dropping the `?? store.deferredEntityIndex?.get(id)` fallback survives
+ * the full suite.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { extractMaterialPropertiesForMaterialId } from '../src/columnar-parser.js';
+import type { IfcDataStore } from '../src/columnar-parser.js';
+import type { EntityRef } from '../src/types.js';
+
+// #20 (the property atom) is deliberately built into `deferredEntityIndex`
+// ONLY — never `entityIndex.byId` — to model the deferred-atom-index shape
+// that a real large-file parse with `deferPropertyAtomIndex: true` produces.
+const FIXTURE = [
+  `#10=IFCMATERIAL('Concrete',$,'concrete');`,
+  `#20=IFCPROPERTYSINGLEVALUE('Strength',$,30.0,$);`,
+  `#30=IFCMATERIALPROPERTIES('Pset_MaterialConcrete',$,(#20),#10);`,
+];
+
+/** Build a minimal store where one referenced entity (the property atom)
+ *  lives ONLY in `deferredEntityIndex`, not the primary `entityIndex.byId`. */
+function buildDeferredAtomStore(): IfcDataStore {
+  const text = FIXTURE.join('\n');
+  const source = new TextEncoder().encode(text);
+
+  const byId = new Map<number, EntityRef>();
+  const deferredById = new Map<number, EntityRef>();
+  const byType = new Map<string, number[]>();
+
+  let cursor = 0;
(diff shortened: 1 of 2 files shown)


louistrue:
Verified and merging. The IFC4X3 ordering gap is real and now pinned.

```
### swap the IFC4X3 / IFC4 ladder lines
     × detects IFC4X3 — the critical ordering case ("IFC4X3".includes("IFC4") is true)
      Tests  1 failed | 5 passed (6)
```

Restored: 6/6.

Driving it through the public `parseLite()` rather than exporting `detectSchemaVersion` is the right choice, and your reason is the one that matters: the ladder's whole purpose is what `schemaVersion` ends up as on a real buffer. Exporting the private function to test it would pin the function's return value while leaving the actual question — does an IFC4X3 file come out as IFC4X3 — untested.

**A confession that makes your method the point.** My first attempt to reproduce this was a **no-op**. I built the mutation from two lines with 2-space indentation and asserted each was present exactly once — both matched, because a 2-space prefix is a substring of the real 4-space one. But the *concatenation* I replaced never matched, so nothing changed, and I got a clean `6/6 passed` that read exactly like "the test does not pin the ordering". I nearly reported that as a defect in your PR.

What saved it was checking that the swap had actually applied before trusting the result:

```
### verify the swap really applied:
120:    if (headerText.includes('IFC4')) return 'IFC4';
121:    if (headerText.includes('IFC4X3')) return 'IFC4X3';
```

Which is exactly the discipline your PR description already uses — asserting the occurrence count before replacing, and printing it. A mutation that silently applies to nothing is indistinguishable from a test that cannot fail, and it fails in the direction that makes you *doubt* correct work.

The underlying trap is worth the test on its own merits: `'IFC4X3'.includes('IFC4')` being true means the ladder ordering is load-bearing with nothing in the type system or the tests to hold it. Someone alphabetising those four lines, or "tidying" them shortest-first, silently misdetects every IFC4X3 file as IFC4 — and it would surface far downstream as wrong schema-dependent behaviour, not as a parse error.

PR #2136, opened 2026-08-04
test(rust): close four mutation-verified coverage gaps

BIMvoice:
## Summary

Closes four mutation-verified coverage gaps in `rust/core` and `rust/wasm-bindings` — each survivor confirmed by hand-applying the stated mutation, watching the new test fail, then restoring the original code and watching it pass again.

### 1. `GeoReference::to_matrix()` — `rust/core/src/georef.rs:243`

Entirely untested. Mutation that survives all 137 pre-existing tests: flip `-s * sin_r` to `+s * sin_r` in the rotation column. `to_matrix()` is load-bearing (`rust/processing/src/georeferencing.rs:89`); a sign flip silently rotates every georeferenced building's placement matrix.

New test `test_to_matrix_pins_rotation_scale_translation` pins all 16 entries (rotation columns, uniform scale on x/y/z, translation) and cross-checks the matrix against `local_to_map` for an arbitrary point — a matrix that disagrees with the pinned point transform is exactly the bug this mutation introduces.

### 2. `MeshDataJs::new`'s winding-order bounds guard — `rust/wasm-bindings/src/zero_copy/mesh.rs:239`

`mesh_tests.rs` never called `MeshDataJs::new` at all. Mutation: `end = mesh.indices.len() - remainder` → `end = mesh.indices.len()`. A non-multiple-of-3 index array then panics slicing past the end — fatal under `panic=abort` in wasm.

New test `new_processes_divisible_prefix_without_panicking_on_non_multiple_of_3_indices` builds a mesh with 4 indices (not a multiple of 3) and asserts the divisible prefix is winding-reversed without panicking.

### 3. Z-up → Y-up axis conversion — `rust/wasm-bindings/src/zero_copy/mesh.rs:226`

Same untested function. Mutation: flip `chunk[2] = -y` to `chunk[2] = y`, mirroring every vertex/normal on the depth axis.

New test `new_converts_zup_to_yup_for_positions_and_normals` pins the axis mapping for a known position and normal.

### 4. `Contours2D::new`'s multi-ring offset bookkeeping — `rust/wasm-bindings/src/api/bool2d.rs:98`

`bool2d.rs` had zero `#[cfg(test)]` coverage — the entire boolean-2D WASM API shipped untested in Rust. Mutation: `at += n` → `at += n + 1`, corrupting the running vertex offset used to slice `coords` per ring.

New test `new_keeps_multi_ring_offsets_distinct_for_rings_of_different_lengths` uses three rings of *different* lengths (3, 4, 5 vertices) — equal-length rings would let an offset bug still produce plausible output — and pins every ring's coordinates exactly.

## Test plan

- [x] RED: each mutation applied individually (python3 exact-substring replace), confirmed the corresponding new test fails (3 panics, 1 assertion failure), restored by file copy, confirmed `git status --porcelain` clean for `rust/` after each cycle.
- [x] GREEN: `cargo test -p ifc-lite-core --lib` — 137 → **138 passed**.
- [x] GREEN: `cargo test -p ifc-lite-wasm --lib` — 27 → **30 passed**.
- [x] `cargo clippy -p ifc-lite-core -p ifc-lite-wasm --all-targets -- -D warnings` — **not currently clean on `upstream/main`** (4 pre-existing errors in `georef.rs`/`scanner.rs`, unrelated to this change — looks like nightly toolchain lint drift, e.g. `chunks_exact_to_as_chunks`, `question_mark`). Verified identical error set with and without this diff (confirmed against a clean `upstream/main` checkout), so this PR introduces zero new clippy findings; the pre-existing drift is out of scope here (test-only change, no production-code touches).
- [x] Test-only change — no changeset.

### DIFF
--- a/rust/core/src/georef.rs
+++ b/rust/core/src/georef.rs
@@ -754,6 +754,64 @@ mod tests {
     }
 
     #[test]
+    fn test_to_matrix_pins_rotation_scale_translation() {
+        let mut georef = GeoReference::new();
+        georef.eastings = 500000.0;
+        georef.northings = 4000000.0;
+        georef.orthogonal_height = 50.0;
+        georef.scale = 2.0;
+        // 30 degree rotation
+        let angle = std::f64::consts::FRAC_PI_6;
+        georef.x_axis_abscissa = angle.cos();
+        georef.x_axis_ordinate = angle.sin();
+
+        let m = georef.to_matrix();
+
+        let cos_r = angle.cos();
+        let sin_r = angle.sin();
+        let s = georef.scale;
+
+        // Column 0: scaled X axis.
+        assert!((m[0] - s * cos_r).abs() < 1e-12);
+        assert!((m[1] - s * sin_r).abs() < 1e-12);
+        assert_eq!(m[2], 0.0);
+        assert_eq!(m[3], 0.0);
+
+        // Column 1: scaled Y axis. This is the column the surviving mutation
+        // (`-s * sin_r` -> `+s * sin_r`) flips.
+        assert!((m[4] - (-s * sin_r)).abs() < 1e-12);
+        assert!((m[5] - s * cos_r).abs() < 1e-12);
+        assert_eq!(m[6], 0.0);
+        assert_eq!(m[7], 0.0);
+
+        // Column 2: scaled Z axis (uniform scale applies to z too).
+        assert_eq!(m[8], 0.0);
+        assert_eq!(m[9], 0.0);
+        assert!((m[10] - s).abs() < 1e-12);
+        assert_eq!(m[11], 0.0);
+
+        // Column 3: translation.
+        assert_eq!(m[12], georef.eastings);
+        assert_eq!(m[13], georef.northings);
+        assert_eq!(m[14], georef.orthogonal_height);
+        assert_eq!(m[15], 1.0);
+
+        // Cross-check against `local_to_map`: applying the matrix to an
+        // arbitrary local point must agree with the point-transform formula.
+        // A flipped rotation sign disagrees here even if the pinned columns
+        // above were missed.
+        let (x, y, z) = (12.0, -7.0, 3.0);
+        let px = m[0] * x + m[4] * y + m[8] * z + m[12];
+        let py = m[1] * x + m[5] * y + m[9] * z + m[13];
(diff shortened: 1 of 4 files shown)


louistrue:
Verified at the Rust bar and merging. All four gaps are real; the two I mutated independently both die against their named test.

```
### flip -s * sin_r -> s * sin_r in to_matrix()
     test test_to_matrix_pins_rotation_scale_translation ... FAILED

### winding guard: end = len() instead of len() - remainder
     panicked at rust/wasm-bindings/src/zero_copy/mesh.rs:241:26
     test ...new_processes_divisible_prefix_without_panicking_on_non_multiple_of_3_indices ... FAILED
```

Restored: 0 failed suites.

The second one is worth dwelling on — the mutation does not fail an assertion, it **panics on an out-of-range slice**, which under `panic=abort` in wasm takes the whole module down. `MeshDataJs::new` had no test calling it at all, so that guard was one careless edit from a hard abort on any mesh with a non-multiple-of-3 index array.

Three things done right that I would not want lost:

**`bool2d.rs` had zero Rust coverage** — the entire boolean-2D WASM API. That is exactly the crate CI was not even linting until #2108 landed, so it was doubly dark.

**Rings of different lengths (3, 4, 5), deliberately.** Your note that equal-length rings would let an offset bug still produce plausible output is the difference between a test and a fixture — `at += n + 1` against uniform rings shifts everything consistently and can still look right.

**Cross-checking `to_matrix()` against `local_to_map`** rather than only pinning 16 numbers. A matrix that disagrees with the point transform is the actual defect; pinned constants alone would drift together if someone "fixed" both.

**One correction to your test plan.** You flagged clippy as *not clean on main* — 4 pre-existing errors in `georef.rs`/`scanner.rs`, suspected nightly drift. That does not reproduce here:

```
$ cargo clippy -p ifc-lite-core -p ifc-lite-wasm --all-targets -- -D warnings
errors: 0
$ cargo clippy --workspace --all-targets -- -D warnings
Finished
```

on `rust-toolchain.toml`'s pinned `nightly-2025-11-15` (rustc 1.93.0-nightly b6d7ff3aa). If you were seeing those locally you were probably on a floating nightly rather than the pinned one — worth checking `rustup show`, because a drifting local toolchain will keep reporting lints CI never sees, and the reverse (missing one CI does see) is the dangerous direction.

`Rust tests` and `Build + WASM + Rust + Node` both green.

PR #40, opened 2026-01-14
[FIX] fix: Strict CI compliance - clippy, fmt, docs, and test fixes

holg:
This PR brings the Rust codebase into strict CI compliance with -D warnings (treat all warnings as errors). All changes maintain backwards compatibility.

  Key Fixes:

  - Memory leak in streaming tests: Fixed infinite loop in streaming.rs that caused 40GB+ memory consumption by properly terminating the stream after Completed event
  - Clippy compliance: Fixed ~30 clippy warnings including:
    - .get(0) → .first() conversions
    - Removed unused imports
    - Fixed len_zero warnings (.len() > 0 → !.is_empty())
    - Fixed field_reassign_with_default patterns
    - Added appropriate #[allow(...)] attributes for intentional patterns
  - Rustdoc compliance: Fixed unclosed HTML tags in doc comments (e.g., Vec<f32> → `Vec<f32>`)
  - TOML formatting: Added taplo.toml config for consistent formatting
  - CI script: Added scripts/test-ci-locally.sh for local CI validation

  Files Changed:

  - rust/core/src/streaming.rs - Critical fix for infinite stream bug
  - rust/core/src/*.rs - Clippy and doc fixes
  - rust/geometry/src/*.rs - Clippy fixes
  - rust/wasm-bindings/src/*.rs - Clippy fixes
  - scripts/test-ci-locally.sh - New CI validation script
  - taplo.toml - TOML formatting config
  - Various Cargo.toml files - Sorted and formatted

  Test plan

  - All 81 tests pass (cargo test --workspace)
  - cargo clippy --workspace -- -D warnings passes
  - cargo fmt --all -- --check passes
  - cargo doc --workspace -- -D warnings passes
  - cargo-sort --workspace --check passes
  - taplo format --check passes
  - WASM build passes

## Summary by CodeRabbit

* **New Features**
  * Extrusion now generates side walls for holes in addition to outer boundaries.
  * Expanded geometry type detection with additional IFC categories.

* **Refactor**
  * Reorganized public API with expanded module re-exports across core and geometry libraries.
  * Enhanced WASM initialization with automatic module startup.
  * Improved coordinate parsing safety with bounds-checking.

* **Chores**
  * Added local CI testing script and TOML formatter configuration.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2137, opened 2026-08-04
test(processing,geometry): close 3 mutation-verified coverage gaps

BIMvoice:
## Summary

Closes three mutation-verified coverage gaps in `rust/processing` and `rust/geometry`. Each mutation below survived the existing suite before this PR; each new test now fails under the stated mutation and passes with it reverted (verified locally, RED then GREEN, before committing).

### 1. `rust/geometry/src/router/instancing.rs` — the empty-mesh guard

`ensure_shared_mapped_source` only publishes a meshed source into the shared registry when `!mesh.positions.is_empty() && !budget::tripped()`. Its own doc comment calls this load-bearing: without it, a source that walks to zero geometry gets cached under its source id and is then served to *every later occurrence* sharing that `IfcRepresentationMap`.

**Mutation**: drop the `!mesh.positions.is_empty()` conjunct — no existing test caught it (the file's prior tests cover the nested-source early-return, not this guard).

**Fix**: extended `tests/fixtures/nested_mapped_item.ifc` with `#40`, a well-formed `IfcShapeRepresentation` with an empty items list. The new test walks `#40` through `ensure_shared_mapped_source`, asserts nothing is registered, then registers a real source (`#13`) under the *same* source id and asserts it now succeeds — proving the aborted empty attempt didn't poison the entry for later, real occurrences.

### 2. `rust/processing/src/shard_classes.rs` — `classify_type_name`'s `TYPE`-suffix arm

The type-candidate check is `type_name.ends_with("TYPE") || type_name.ends_with("STYLE")` before confirming the resolved `IfcType` is an `IfcTypeProduct` subtype.

**Mutation**: drop `ends_with("TYPE")`, keeping only the `STYLE` check — no test referenced `PREPASS_CLASS_FLAG_TYPE_CANDIDATE` at all.

**Fix**: two new unit tests — `IFCWALLTYPE` (ends `TYPE`, a real `IfcTypeProduct` subtype) must set the flag; `IFCSURFACESTYLE` (ends `STYLE`, an `IfcPresentationStyle`, not `IfcTypeProduct`) must not. Together they pin that the two suffix arms gate genuinely different keyword sets.

### 3. `rust/processing/src/processor/quick_metadata.rs` — `extract_storey_elevation_from_args`'s index priority

`IfcBuildingStorey.Elevation` is read by checking attribute indices `[9, 8]` in that order (9 first, matching the IFC4 attribute layout this parser targets; 8 is a fallback).

**Mutation**: swap to `[8, 9]` — no test in `rust/processing/tests` asserts `.elevation` at all, and this function is called in production from `processor/mod.rs:606`.

**Fix**: new unit test with *different* numeric values at indices 8 and 9 (equal values would let the swap pass silently) and asserts index 9's value wins.

## Scope

Tests only — no production code changed, no changeset needed (Rust-only test additions, not a published `packages/*` surface change).

## Verification

- **Baselines confirmed before changes**: `ifc-lite-processing` lib target 57 passed; `ifc-lite-geometry` lib target 576 passed / 1 ignored, plus the 3 disclosed pre-existing `tests/wall_opening_cut_regression.rs` failures (unfetched-fixture skip, `pnpm fixtures`).
- **Per-gap RED/GREEN**: each mutation applied via exact-substring Python replacement, confirmed the new test fails, reverted via the same mechanism, confirmed it passes again. `git status --porcelain -- rust/` checked clean of stray edits after each restore.
- **Final counts**: `ifc-lite-processing` lib target 60 passed (57 + 3 new); `ifc-lite-geometry` lib target 577 passed / 1 ignored (576 + 1 new); the same 3 `wall_opening_cut_regression.rs` failures remain, unchanged in count and identity.
- **Clippy**: `cargo clippy -p ifc-lite-processing -p ifc-lite-geometry --all-targets -- -D warnings` currently fails on `main` for reasons unrelated to this PR — see note below. None of the reported errors/warnings touch any file this PR changes.

### Clippy note (pre-existing, not introduced by this PR)

Running the exact required command against `upstream/main` (no changes from this PR) currently fails: `ifc-lite-core` (a path dependency clippy also lints) fails to compile under `-D warnings` with 4 errors from newly-default-warn lints (e.g. `clippy::question_mark`, `clippy::chunks_exact_to_as_chunks`, `clippy::collapsible_match`) — apparently toolchain/clippy drift on the pinned nightly (`rustc 1.99.0-nightly 2026-08-01`) since the "currently clean" baseline was established, not anything in this diff. Running clippy without `-D warnings` shows the same lint families firing across many pre-existing files in `rust/processing`, `rust/geometry`, and `rust/core` that this PR does not touch. I did not attempt to fix these — `rust/core` is out of scope for this PR (a separate agent is concurrently working there on `test/rust-coverage-gaps`), and fixing pre-existing lint debt across two crates is out of scope for a coverage-gap PR. Flagging for the maintainer's awareness.

### DIFF
--- a/rust/geometry/src/router/instancing.rs
+++ b/rust/geometry/src/router/instancing.rs
@@ -143,4 +143,39 @@ mod tests {
         router.ensure_shared_mapped_source(&flat_rep, 14, &mut decoder);
         assert_eq!(router.mapped_shared_unique_count(), 1);
     }
+
+    /// The empty-mesh guard: a source that decodes fine but walks to ZERO items
+    /// (`#40` is a well-formed `IfcShapeRepresentation` with an empty items list)
+    /// must not be published into the shared registry — an empty/truncated source
+    /// cached under `source_id` would be served to every later occurrence sharing
+    /// that `IfcRepresentationMap`, per the doc on `ensure_shared_mapped_source`.
+    ///
+    /// The second half proves the guard doesn't poison the id going forward: a
+    /// later call with the SAME `source_id` but a real single-solid source (`#13`)
+    /// must still register, because the aborted empty attempt inserted nothing.
+    #[test]
+    fn empty_mesh_source_not_registered_then_real_source_still_is() {
+        let content = fixture();
+        let entity_index = ifc_lite_core::build_entity_index(&content);
+        let mut decoder = EntityDecoder::with_index(&content, entity_index);
+        let mut router = GeometryRouter::with_units(&content, &mut decoder);
+        router.enable_shared_mapped_item_cache(GeometryRouter::new_mapped_item_cache());
+
+        let empty_rep = decoder.decode_by_id(40).expect("decode #40");
+        router.ensure_shared_mapped_source(&empty_rep, 999, &mut decoder);
+        assert_eq!(
+            router.mapped_shared_unique_count(),
+            0,
+            "a source that meshes to zero positions must not be cached"
+        );
+
+        let flat_rep = decoder.decode_by_id(13).expect("decode #13");
+        router.ensure_shared_mapped_source(&flat_rep, 999, &mut decoder);
+        assert_eq!(
+            router.mapped_shared_unique_count(),
+            1,
+            "a real mesh for the same source id must still register after an \
+             aborted empty attempt didn't poison the entry"
+        );
+    }
 }
--- a/rust/geometry/tests/fixtures/nested_mapped_item.ifc
+++ b/rust/geometry/tests/fixtures/nested_mapped_item.ifc
@@ -49,5 +49,8 @@ DATA;
 #37=IFCRELAGGREGATES('0REL256789ABCDEFGHIJKL',$,$,$,#23,(#25));
 #38=IFCRELAGGREGATES('0REL356789ABCDEFGHIJKL',$,$,$,#25,(#27));
 #39=IFCRELCONTAINEDINSPATIALSTRUCTURE('0REL456789ABCDEFGHIJKL',$,$,$,(#35),#27);
+/* Empty-items source: a well-formed representation with NO geometry items, so
+   walking it in `ensure_shared_mapped_source` completes with `mesh` still empty. */
+#40=IFCSHAPEREPRESENTATION(#5,'Body','SweptSolid',());
 ENDSEC;
 END-ISO-10303-21;
--- a/rust/processing/src/processor/quick_metadata.rs
+++ b/rust/processing/src/processor/quick_metadata.rs
@@ -227,4 +227,22 @@ mod tests {
         let tree = build_quick_spatial_tree_node(1, &nodes, &summaries);
         assert!(tree.is_ok(), "cyclic tree should build (cycle pruned), got {tree:?}");
     }
+
+    /// `IfcBuildingStorey`'s `Elevation` attribute sits at index 9 in the IFC4
+    /// attribute layout this parser targets; index 8 is only a fallback (e.g. an
+    /// off-by-one attribute count from a schema variant). Indices 8 and 9 hold
+    /// DIFFERENT numeric values here specifically so a priority swap (checking 8
+    /// before 9) is observable — equal values would let a `[9, 8]` -> `[8, 9]`
+    /// swap pass silently.
+    #[test]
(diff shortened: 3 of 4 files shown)


BIMvoice:
Resolving the clippy note in this PR body before it misleads anyone: **clippy is clean on the pinned toolchain, and CI uses the pinned toolchain.**

`rust-toolchain.toml` at the repo root pins:

```toml
[toolchain]
channel = "nightly-2025-11-15"
components = ["rust-src", "clippy"]
```

On that toolchain (`rustc 1.93.0-nightly b6d7ff3aa`, `clippy 0.1.93`) I ran `cargo clippy -p ifc-lite-core -p ifc-lite-wasm --all-targets -- -D warnings` against a clean checkout of current `main` and against the sibling Rust branch: **zero errors, both times.**

The four errors reported in the body — `clippy::question_mark`, `clippy::chunks_exact_to_as_chunks`, `clippy::collapsible_match` — came from a **2026-08 nightly**, roughly nine months ahead of the pin. Those lints post-date it. So this is unpinned-toolchain drift in a local environment, not lint debt on `main`, and there is nothing here for you to action.

Worth recording because two independent agents hit it and both reported it as pre-existing repo state. The check that settles it is `cat rust-toolchain.toml` before trusting a local clippy run — I have added that to how we verify Rust changes.

Everything else in the PR stands: three gaps, each RED-verified against the surviving mutation and GREEN after restore, `ifc-lite-processing` 57 → 60, `ifc-lite-geometry` 576 → 577, and the three `wall_opening_cut_regression.rs` failures unchanged in count and identity (the disclosed unfetched-fixture skip).

One correction in the report that I want to keep visible, since it sharpens the finding rather than softening it: a test *does* already reference `ensure_shared_mapped_source` (`nested_source_is_not_registered_flat`) — but it exercises the nested-item early return, not the empty-mesh guard. So the guard whose own doc comment warns about a truncated source being "served to every later occurrence" was genuinely unpinned, and the new test also checks that a real source under the same id still registers afterwards, so the fix cannot be satisfied by simply never caching.

louistrue:
Reviewed at the Rust bar — this is the only PR in the queue touching `rust/`, and the priority here is keeping that side clean. All three gaps are real and all three tests kill their mutation.

| mutation | test that caught it |
|---|---|
| `for index in [9usize, 8usize]` → `[8usize, 9usize]` | `storey_elevation_prefers_index_9_over_index_8` |
| drop `!mesh.positions.is_empty()` from the cache guard | `empty_mesh_source_not_registered_then_real_source_still_is` |
| `ty.is_subtype_of(IfcType::IfcTypeProduct)` → `true` | `style_suffix_without_type_product_subtype_does_not_set_flag` |

Each mutation's occurrence count was asserted before applying and the change confirmed in-file, because a mutation that silently applies to nothing produces a green run indistinguishable from an unpinned fix.

Three things worth saying about *which* gaps these are:

**The elevation ordering is a re-fix guard, not a hypothetical.** Slot 9 vs slot 8 was a real defect — `IfcBuildingStorey.Elevation` is index 9, and reading 8 gets `CompositionType`. Swapping the ladder previously produced silently wrong storey elevations with nothing failing. Now it also cascades into three wall/opening tests, which is a nice property: the fixture-level tests notice, but only this one names the cause.

**The instancing test pins non-poisoning, not just non-registration.** Asserting `mapped_shared_unique_count() == 0` after the empty source *and* `== 1` after a real mesh for the same source id is the part that matters — a guard that skipped registration but left a tombstone would pass the first assertion and fail the second. That is the harder half and it is covered.

**The `shard_classes` pair is a matched positive/negative.** `type_suffix_sets_type_candidate_flag` alone would be satisfied by `if true`; it is `style_suffix_without_type_product_subtype_does_not_set_flag` that fails under the over-broad mutation. A flag test without its negative control is the shape that lets a predicate quietly widen.

Also confirmed at the CI scope: `cargo clippy --workspace --all-targets -- -D warnings` clean — which now genuinely covers `ifc-lite-wasm` since #2108 removed the exclusion, so this is the first Rust PR to land under a gate that actually lints the whole workspace.

`Rust tests` green on CI. Merging once the watertightness census finishes — I want the core lanes *present*, not merely not-red.

PR #2138, opened 2026-08-04
test(viewer): close two mutation-verified coverage gaps

BIMvoice:
## Summary

Closes two confirmed mutation-testing survivors in `apps/viewer` by adding tests only — no production behaviour changed.

### Gap 1 — `apps/viewer/src/lib/storage/save-result.ts:61`

```ts
const quota = err instanceof DOMException && QUOTA_ERROR_NAMES.has(err.name);
```

**Surviving mutation:** `&&` → `||`. Every `DOMException` — including Safari private mode's `SecurityError` thrown by `setItem` — is then classified `reason: 'quota'` ("Browser storage is full"), when the correct diagnosis is `'unavailable'`. A user hitting this is sent to delete data that was never the problem.

No existing test threw a non-quota `DOMException`; `lensSlice.save-failure.test.ts` only exercises the true `QuotaExceededError` path.

**Fix:** added tests in `save-result.test.ts` asserting:
- a genuine `QuotaExceededError` still reports `'quota'` (no regression)
- a `SecurityError` `DOMException` reports `'unavailable'`, not `'quota'`
- a plain non-`DOMException` `Error` also reports `'unavailable'`

### Gap 2 — `apps/viewer/src/store/slices/modelSlice.ts:327`

```ts
if (localId >= 0 && localId <= model.maxExpressId) { return { modelId: model.id, expressId: localId }; }
```

**Surviving mutation:** `<=` → `<`. The boundary id (`localId === model.maxExpressId`, the highest express id parsed for that model) drops out of the fast path. The second-pass overlay loop explicitly `continue`s on `localId <= model.maxExpressId`, so the boundary id resolves to `null` — selection, basket, and property lookups silently fail for the last parsed entity of every model.

`modelSlice.test.ts`'s existing `resolveGlobalIdFromModels — overlay-allocated ids` test only used `42` (well inside range) and `11_001`/`99_999` (overlay/phantom); nothing pinned the boundary.

**Fix:** added a `resolveGlobalIdFromModels — maxExpressId boundary` describe block with two tests:
- a single model resolves `expressId === maxExpressId` through the fast path (no mutation views registered, so a mutant that drops to the second pass has nothing to catch it and returns `null`)
- a federated two-model setup where the first model's boundary id is numerically also in-range for the second model (offset-adjacent), so a naive resolver would misattribute it

Per the brief, the second-pass `localId <= model.maxExpressId` guard at line 335 is **not** touched — that branch only runs for ids the first pass already rejected, so `<` vs `<=` there is unreachable (an equivalent mutant), not a defect.

## Verification (RED-GREEN per gap)

Both mutations applied via python3 exact-substring replacement (count asserted `== 1`), confirmed the new tests **fail**, then restored by file copy (not `git checkout --`) and confirmed **pass**.

- Gap 1: `&&` → `||` in `save-result.ts` → new "SecurityError" test failed (`expected: 'unavailable', actual: 'quota'`). Restored → all pass.
- Gap 2: `<=` → `<` in `modelSlice.ts` → both new boundary tests failed (single-model boundary resolved to `null`; federated boundary resolved to `{ modelId: 'model-2', expressId: 0 }` instead of `{ modelId: 'model-1', expressId: 10000 }`). Restored → all pass.

## Test plan

- [x] `pnpm build:wasm` then `pnpm build` (full monorepo, including `@ifc-lite/viewer`)
- [x] `cd apps/viewer && npm test` — baseline 2723 tests / 2721 pass / 2 skipped → final **2728 tests / 2726 pass / 2 skipped** (+5 new tests, all passing)
- [x] `pnpm typecheck` — green
- [x] Only test files changed (`save-result.test.ts`, `modelSlice.test.ts`); no production code touched
- [x] No changeset — `@ifc-lite/viewer` is a private package (verified against `apps/viewer/package.json`)

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure storage errors are accurately classified, distinguishing quota limits from unavailable storage.
  * Added regression coverage for resolving model identifiers at boundary values, including federated model overlaps.

### DIFF
--- a/apps/viewer/src/lib/storage/save-result.test.ts
+++ b/apps/viewer/src/lib/storage/save-result.test.ts
@@ -62,3 +62,45 @@ describe('saveJson — a value JSON.stringify cannot round-trip (PR #2091 review
     assert.equal(data.get('k'), '{"a":1}');
   });
 });
+
+describe('saveJson — quota vs. unavailable DOMException discrimination', () => {
+  beforeEach(() => { installStubStorage(); });
+
+  it('reports "quota" for a genuine QuotaExceededError', () => {
+    (globalThis as unknown as { localStorage: Storage }).localStorage = {
+      ...(globalThis as unknown as { localStorage: Storage }).localStorage,
+      setItem: () => { throw new DOMException('quota', 'QuotaExceededError'); },
+    } as Storage;
+    const result = saveJson('k', { a: 1 }, 'lens changes');
+    assert.equal(result.ok, false);
+    assert.equal(!result.ok && result.reason, 'quota');
+    assert.ok(!result.ok && result.message.includes('full'), 'quota message must say storage is full');
+  });
+
+  it('reports "unavailable", not "quota", for a SecurityError DOMException (e.g. Safari private mode)', () => {
+    // Safari's private-mode `setItem` throws a `SecurityError` DOMException,
+    // not a quota error. Misreporting this as "quota" tells the user their
+    // storage is full and sends them deleting data that is not the problem
+    // (`&&` -> `||` on the QUOTA_ERROR_NAMES check would misclassify this).
+    (globalThis as unknown as { localStorage: Storage }).localStorage = {
+      ...(globalThis as unknown as { localStorage: Storage }).localStorage,
+      setItem: () => { throw new DOMException('The operation is insecure.', 'SecurityError'); },
+    } as Storage;
+    const result = saveJson('k', { a: 1 }, 'lens changes');
+    assert.equal(result.ok, false);
+    assert.equal(!result.ok && result.reason, 'unavailable',
+      'a non-quota DOMException must not be classified as quota');
+    assert.ok(!result.ok && result.message.includes('unavailable'),
+      'unavailable message must not claim storage is full');
+  });
+
+  it('reports "unavailable" for a non-DOMException error (e.g. a plain thrown Error)', () => {
+    (globalThis as unknown as { localStorage: Storage }).localStorage = {
+      ...(globalThis as unknown as { localStorage: Storage }).localStorage,
+      setItem: () => { throw new Error('blocked by extension'); },
+    } as Storage;
+    const result = saveJson('k', { a: 1 }, 'lens changes');
+    assert.equal(result.ok, false);
+    assert.equal(!result.ok && result.reason, 'unavailable');
+  });
+});
--- a/apps/viewer/src/lib/storage/save-result.test.ts
+++ b/apps/viewer/src/lib/storage/save-result.test.ts
@@ -94,6 +94,30 @@ describe('saveJson — quota vs. unavailable DOMException discrimination', () =>
       'unavailable message must not claim storage is full');
   });
 
+  it('reports "unavailable" for a quota-NAMED error that is not a DOMException', () => {
+    // Pins the `err instanceof DOMException` half of the guard, which the
+    // other fixtures cannot: they all agree with a name-only check, so
+    // dropping the instanceof survives them (maintainer mutation on #2138).
+    //
+    // `name` is a plain writable string, so anything that re-wraps or
+    // normalises a storage error can carry a quota-ish name without being a
+    // DOMException — and classifying that as quota produces the same wrong
+    // advice this file exists to prevent: "free up space" when space was
+    // never the problem.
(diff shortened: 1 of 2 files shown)


louistrue:
The gap you close is real and your test catches it. CodeRabbit's follow-on is also real, and I confirmed both by mutation:

```
### && -> ||   (the survivor you targeted)
    not ok 2 - reports "unavailable", not "quota", for a SecurityError DOMException (e.g. Safari private mode)
    # pass 6  # fail 1                                        ✅ caught

### drop the instanceof, classify by name alone
    # pass 7  # fail 0                                        ❌ survives
```

So the `QUOTA_ERROR_NAMES.has(err.name)` half is pinned, and the `err instanceof DOMException` half is not. All three fixtures pass a name-only guard:

| fixture | name | name-only guard says | correct |
|---|---|---|---|
| `QuotaExceededError` DOMException | `QuotaExceededError` | quota | quota ✓ |
| `SecurityError` DOMException | `SecurityError` | unavailable | unavailable ✓ |
| plain `Error` | `Error` | unavailable | unavailable ✓ |

Nothing distinguishes them because no fixture has a **quota-ish name on a non-DOMException**. One more case closes it:

```ts
const impostor = new Error('not a DOMException');
impostor.name = 'QuotaExceededError';
// must still be 'unavailable' — the name matches, the type does not
```

That fails under the name-only mutation and passes today.

Not academic, either: `name` is a plain writable string, so anything that re-wraps or normalises a storage error can carry a quota-ish name without being a `DOMException` — and the consequence is the same one your PR is about, a user told to free up space when storage was never the problem.

The Safari private-mode diagnosis is a good catch on its own. `SecurityError` from `setItem` reported as "Browser storage is full" sends someone deleting data to fix a permission, which is the worst kind of wrong error message — actionable and harmful.

BIMvoice:
Closed in `8f5b9dfa`. Your mutation now fails.

## The impostor case

```ts
const impostor = new Error('not a DOMException');
impostor.name = 'QuotaExceededError';
```

```
### classify by name alone (drop the instanceof)
    not ok 2 - saveJson — quota vs. unavailable DOMException discrimination
    # pass 7  # fail 1
```

Previously 7/7 green. `apps/viewer` 2728 → 2729, 0 fail, no production file touched.

Your table is the part worth keeping: all three existing fixtures **agree** with a name-only check, so none of them could discriminate. A quota-named DOMException is quota either way; `SecurityError` and a plain `Error` are unavailable either way. The guard has two halves and the fixtures only exercised one — which is a shape I will look for directly now: **a compound condition whose test set never disagrees with either conjunct alone.**

And your reasoning for why it is reachable rather than academic is the right one — `name` is a plain writable string, so any re-wrapping or normalising layer can carry a quota-ish name without being a `DOMException`. The consequence is exactly the one this file exists to prevent: telling someone to free up space when space was never the problem.

Thank you also for the note on the Safari diagnosis. Agreed that a permission error reported as "storage is full" is the worst shape of wrong message — it is *actionable*, so the user does the work, and the work does not help.

louistrue:
Both halves of the guard are now pinned. The mutation that survived last round dies:

```
### name-only guard (drop the instanceof)   — SURVIVED before, now:
    not ok 3 - reports "unavailable" for a quota-NAMED error that is not a DOMException
    # pass 7  # fail 1

### && -> ||                                — still caught, now by two tests:
    not ok 2 - reports "unavailable", not "quota", for a SecurityError DOMException (e.g. Safari private mode)
    not ok 3 - reports "unavailable" for a quota-NAMED error that is not a DOMException
    # pass 6  # fail 2
```

Restored: 8/8.

The `impostor` fixture is exactly right — a plain `Error` with `name = 'QuotaExceededError'` is the only shape that separates the type check from the name check, and it now fails under a name-only guard while passing today. Note it also strengthens the `&&`→`||` case: that mutation used to fail one test and now fails two, because the two fixtures probe the predicate from opposite sides.

Worth stating what this pair actually buys, since it is easy to read as pedantry: `name` is a writable string on any `Error`, so anything that re-wraps, normalises or re-throws a storage failure can carry a quota-ish name without being a `DOMException`. Under a name-only guard that user is told "Browser storage is full" and goes off deleting data to fix something that was never a quota — the same actionable-but-wrong diagnosis your Safari private-mode case fixes, arriving by the other route.

Merging once CI finishes — 2 lanes still running, and I want the core lanes present rather than merely not-red.

PR #2139, opened 2026-08-04
test: close four mutation-verified coverage gaps (cache/encoding/sdk/server-client)

BIMvoice:
## Summary

Closes four confirmed mutation-testing survivors across published packages. All are test-only changes — no production behavior modified, no changeset needed.

### 1. `packages/cache/src/utils/hash.ts` — xxHash64 had no correctness test

The existing tests only asserted the result is a `bigint`, that the same input hashes the same, and that different inputs differ — a completely different hash function would pass all three. Worse, the reader's `sourceHash === xxhash64(sourceBuffer)` self-check recomputes with the same function, so it can't catch an algorithmic error either.

Added fixed XXH64 reference vectors (seed 0): empty input, `"a"`, and a 63-byte string (long enough to exercise the `>=32`-byte main loop, where a rotate-constant mutation lives). Sourced from the [cespare/xxhash](https://github.com/cespare/xxhash) Go implementation's test suite (`xxhash_test.go`, `TestAll`), which is cross-validated against the canonical C reference implementation (Cyan4973/xxHash) — **not** derived by running our own code, which would recreate the same self-consistency trap.

Verified our implementation matches these vectors, so this is a genuine test gap, not an algorithm bug.

**Mutation killed:** `rotl64(v4, 18)` → `rotl64(v4, 17)` in the main loop now fails the 63-byte-vector test.

### 2. `packages/server-client/src/client.ts:978` — tail-buffer terminal detection

The main-loop terminal check (`:970`) tests `type === 'complete' || type === 'error'` and is tested. The tail-buffer check at `:978` has the identical shape but was never exercised for the `error` case: the existing error test always sends its frame with a trailing `\n\n`, so it's parsed in the main loop, never the tail path. Dropping `|| data.type === 'error'` at `:978` survived.

Added a test whose final `error` frame has **no** trailing `\n\n`, forcing it through the tail-buffer decode path.

**Mutation killed:** removing `|| tail.type === 'error'` at `:978` now fails.

### 3. `packages/encoding/src/property-value.ts` — vacuous "formats numbers" test

The test asserted only `toBeTruthy()` and that `ifcType` is `undefined`. Swapping the integer/non-integer formatting branches survived. Now asserts the actual formatted strings for both an integer (`42` → `'42'`) and a non-integer (`3.14159265` → `'3.141593'`).

**Mutation killed:** swapping the ternary branches now fails (`3.14159265` renders as `'3.142'` instead of `'3.141593'`).

### 4. `packages/sdk` — `dispatchToBackend`/`BimHost` had zero coverage

No test file referenced `BimHost` or `dispatchToBackend`. The method-gate check (`Object.prototype.hasOwnProperty.call(ns, method)`) is what prevents a wire-protocol caller from reaching inherited members (`toString`, `constructor`, …) on a namespace object; changing it to `method in ns` survived.

Added `packages/sdk/src/host.test.ts` covering both `dispatchToBackend` directly and through `BimHost.dispatch`: own-method dispatch succeeds, an unknown namespace is rejected, and inherited members (`toString`, `constructor`) are rejected rather than invoked.

**Mutation killed:** `hasOwnProperty` → `in` at the method gate now fails 3 tests.

## Method

Each gap was verified RED-first: applied the stated mutation via exact-substring Python replacement (count-checked to `n==1`), confirmed the new test failed, restored the original file from `git show HEAD:<path>` (byte-for-byte, not `git checkout --`), confirmed green again.

## Test plan

- [x] `pnpm --filter @ifc-lite/cache test` — 48 passed (was 45)
- [x] `pnpm --filter @ifc-lite/server-client test` — 28 passed (was 27)
- [x] `pnpm --filter @ifc-lite/encoding test` — 51 passed (unchanged count; assertions strengthened)
- [x] `pnpm --filter @ifc-lite/sdk test` — 68 passed (was 62)
- [x] `pnpm typecheck` — passes
- [x] RED confirmed per gap (mutation applied → test fails), then restored and GREEN confirmed
- [x] No production code changed; no changeset needed (test-only)

### DIFF
--- a/packages/cache/src/cache.test.ts
+++ b/packages/cache/src/cache.test.ts
@@ -41,6 +41,35 @@ describe('xxhash64', () => {
     const hash2 = xxhash64(data2);
     expect(hash1).not.toBe(hash2);
   });
+
+  // Reference vectors sourced from the cespare/xxhash Go implementation's test
+  // suite (xxhash_test.go, TestAll), which is cross-validated against the
+  // canonical C reference implementation (Cyan4973/xxHash):
+  // https://raw.githubusercontent.com/cespare/xxhash/master/xxhash_test.go
+  // These are NOT derived by running this package's own implementation --
+  // the reader's `sourceHash === xxhash64(sourceBuffer)` self-check can't
+  // catch an algorithmic error precisely because it recomputes with the same
+  // (possibly wrong) function, so the pinned values below must come from an
+  // independent source.
+  it('matches the published XXH64 reference vector for the empty input (seed 0)', () => {
+    const hash = xxhash64(new TextEncoder().encode(''));
+    expect(hash).toBe(0xef46db3751d8e999n);
+  });
+
+  it('matches the published XXH64 reference vector for "a" (seed 0)', () => {
+    const hash = xxhash64(new TextEncoder().encode('a'));
+    expect(hash).toBe(0xd24ec4f1a98c6e5bn);
+  });
+
+  it('matches the published XXH64 reference vector for a 63-byte input (seed 0), exercising the >=32-byte main loop', () => {
+    // Exactly 63 characters -- long enough to run the main 32-byte-block
+    // loop (where e.g. a rotate-constant mutation like rotl64(v4, 18) would
+    // otherwise go undetected) plus the trailing 8/4/1-byte remainder paths.
+    const s63 = 'Call me Ishmael. Some years ago--never mind how long precisely-';
+    expect(s63.length).toBe(63);
+    const hash = xxhash64(new TextEncoder().encode(s63));
+    expect(hash).toBe(0x02a2e85470d6fd96n);
+  });
 });
 
 describe('BinaryCacheWriter and BinaryCacheReader', () => {
--- a/packages/encoding/src/property-value.test.ts
+++ b/packages/encoding/src/property-value.test.ts
@@ -53,9 +53,20 @@ describe('parsePropertyValue', () => {
   });
 
   it('formats numbers', () => {
-    const result = parsePropertyValue(42);
-    expect(result.displayValue).toBeTruthy();
-    expect(result.ifcType).toBeUndefined();
+    // Integer and non-integer take different formatting branches
+    // (`Number.isInteger(value) ? ... : ...`); assert the actual strings so
+    // a branch swap is caught instead of just "some truthy string came out".
+    const intResult = parsePropertyValue(42);
+    expect(intResult.displayValue).toBe('42');
+    expect(intResult.ifcType).toBeUndefined();
+
+    // Non-integer is formatted with maximumFractionDigits: 6. If the
+    // branches were swapped, this would instead go through the integer
+    // branch's plain `toLocaleString()`, which defaults to 3 fraction
+    // digits and would produce '3.142' instead of '3.141593'.
+    const floatResult = parsePropertyValue(3.14159265);
+    expect(floatResult.displayValue).toBe('3.141593');
+    expect(floatResult.ifcType).toBeUndefined();
   });
 
   it('decodes IFC-encoded strings', () => {
--- a/packages/encoding/src/property-value.test.ts
+++ b/packages/encoding/src/property-value.test.ts
@@ -53,19 +53,26 @@ describe('parsePropertyValue', () => {
   });
 
   it('formats numbers', () => {
+    // Production formats via `toLocaleString`, so the group/decimal
+    // separators depend on the ambient locale ('3.141593' vs '3,141593').
+    // Compare the digit sequence only: that is locale-agnostic but still
+    // sensitive to how many fraction digits were kept.
+    const digits = (s: string | undefined) => s?.replace(/[^0-9]/g, '');
+
     // Integer and non-integer take different formatting branches
-    // (`Number.isInteger(value) ? ... : ...`); assert the actual strings so
+    // (`Number.isInteger(value) ? ... : ...`); assert the digit sequence so
     // a branch swap is caught instead of just "some truthy string came out".
     const intResult = parsePropertyValue(42);
-    expect(intResult.displayValue).toBe('42');
+    expect(digits(intResult.displayValue)).toBe('42');
     expect(intResult.ifcType).toBeUndefined();
 
     // Non-integer is formatted with maximumFractionDigits: 6. If the
     // branches were swapped, this would instead go through the integer
     // branch's plain `toLocaleString()`, which defaults to 3 fraction
-    // digits and would produce '3.142' instead of '3.141593'.
+    // digits and would produce '3.142' -> digits '3142', failing this
+    // assertion regardless of locale.
     const floatResult = parsePropertyValue(3.14159265);
-    expect(floatResult.displayValue).toBe('3.141593');
+    expect(digits(floatResult.displayValue)).toBe('3141593');
     expect(floatResult.ifcType).toBeUndefined();
   });
 
--- a/packages/sdk/src/host.test.ts
+++ b/packages/sdk/src/host.test.ts
@@ -1,0 +1,102 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect, vi } from 'vitest';
+import { BimHost } from './host.js';
+import { dispatchToBackend } from './types.js';
+import type { BimBackend } from './types.js';
+
+/**
+ * Minimal mock backend: a real `query` namespace object plus a `subscribe`
+ * stub. `dispatchToBackend`/`BimHost` only ever route through `hasOwnProperty`
+ * lookups on `backend` and on the resolved namespace object, so this is
+ * enough to exercise the dispatch gate without pulling in a full BimBackend.
+ */
+function createMockBackend() {
+  const entities = vi.fn((..._args: unknown[]) => ['entity-1', 'entity-2']);
+  const query = { entities };
+  const backend = {
+    query,
+    subscribe: vi.fn(() => () => {}),
+  } as unknown as BimBackend;
+  return { backend, query };
+}
+
+describe('dispatchToBackend', () => {
+  it('dispatches to an own method on the namespace', () => {
+    const { backend, query } = createMockBackend();
+
+    const result = dispatchToBackend(backend, 'query', 'entities', ['model-1']);
+
+    expect(result).toEqual(['entity-1', 'entity-2']);
+    expect(query.entities).toHaveBeenCalledWith('model-1');
+  });
+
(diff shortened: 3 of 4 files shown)


louistrue:
The xxHash finding is the best test observation in this batch, and I verified it two ways.

**Your mutation is caught:**

```
### rotl64(v4, 18) -> rotl64(v4, 17)
     × matches the published XXH64 reference vector for a 63-byte input (seed 0), exercising the >=32-byte main loop
      Tests  1 failed | 25 passed (26)
```

**And the stronger claim — that a completely different algorithm would have passed everything — holds exactly.** I substituted FNV-1a for `xxhash64` wholesale:

```
     × matches the published XXH64 reference vector for the empty input (seed 0)
     × matches the published XXH64 reference vector for "a" (seed 0)
     × matches the published XXH64 reference vector for a 63-byte input (seed 0)
      Tests  3 failed | 23 passed (26)
```

**Only your three new tests notice.** The three pre-existing hash tests pass, and so does every cache round-trip and `validate()` test — because, as you point out, the reader recomputes `sourceHash` with the same function, so the self-check is a tautology. A cache written with the wrong hash validates perfectly against itself.

The part I most want to endorse is where the vectors came from:

> Sourced from the cespare/xxhash Go implementation's test suite … **not** derived by running our own code, which would recreate the same self-consistency trap.

That is the difference between a golden-value test and a tautology, and it is the single easiest corner to cut when adding "reference" vectors. `0xef46db3751d8e999` for the empty input and `0xd24ec4f1a98c6e5b` for `"a"` are the canonical published XXH64 seed-0 values, so these are genuinely external.

Choosing a 63-byte input deliberately — long enough to enter the `>=32`-byte accumulator loop where the rotate constants live — is what makes the third vector more than a duplicate of the first two. The empty and `"a"` cases never reach that code path.

**A correction on my own verification:** my first attempt at the FNV control referenced `buffer` when the parameter is `data`, so it threw a `ReferenceError` and failed 24 tests. I nearly reported that as "even the round-trip tests catch it", which would have been wrong in your favour. Re-run with the right identifier, the result is the clean 3-of-26 above.

Reviewing the other three gaps next.

louistrue:
Both threads are valid. The SDK one is a genuine prototype-escape gap and I have the exact fixture for you.

## `dispatchToBackend` — the namespace gate is unpinned

`dispatchToBackend` has **two** `hasOwnProperty` gates: one on the namespace, one on the method. Your new tests (`toString`, `constructor`) discriminate the **method** gate. The namespace gate is only exercised with `'nope'`, which is absent from own *and* prototype — so it cannot tell `hasOwnProperty` from `in`.

I checked which fixture actually discriminates it, against a backend with a prototype method `secret()`:

```
ns=nope         | hasOwn: reject 'nope'         | in: reject 'nope'
ns=toString     | hasOwn: reject 'toString'     | in: reject (type)
ns=constructor  | hasOwn: reject 'constructor'  | in: reject (type)
ns=__proto__    | hasOwn: reject '__proto__'    | in: INVOKED -> PWNED
```

`toString` and `constructor` are *not* enough as namespaces — the `typeof ns !== 'object'` check catches them either way, so they'd pass against the mutation. **`__proto__` is the one that gets through**: `'__proto__' in backendObj` is true, the value is a real object, and the method lookup then finds the class's own prototype methods. That is a caller reaching `backend.secret()` off the wire.

So the discriminating test is:

```ts
expect(() => dispatchToBackend(backend, '__proto__', 'someProtoMethod', []))
  .toThrow("Unknown namespace '__proto__'");
```

with a backend whose class has at least one prototype method to reach. Without a reachable prototype method the fixture still passes under the mutation, for the same reason `nope` does.

Worth noting the docblock already names this exact threat — "so an attacker can't reach prototype members (`__proto__`, `constructor`, `toString`, …)" — and `__proto__` is the one member on that list the tests don't cover. Same shape as #2140's GAP 1: the negative fixture has to fail for the reason under test.

## The locale assertion is a production question, not just a test one

greptile flags the test, but the sensitivity is in the source — `property-value.ts:124-125` uses `value.toLocaleString(...)`. So `displayValue` for a float genuinely differs by host locale: `3.141593` vs `3,141593`.

The test is the messenger. The real question is whether a **machine-readable `displayValue` should be locale-formatted at all** — it flows into CSV/JSON export and property tables, where a comma decimal separator in an unquoted CSV field is its own bug. If display formatting is intended, the test should pin it locale-independently (force a locale, or assert against `(3.14159265).toLocaleString(undefined,{maximumFractionDigits:6})`). If it is not intended, that is a separate defect worth its own issue.

I would not block this PR on the second one — it predates your change — but it is worth a decision rather than a locale-proofed assertion that quietly entrenches the behaviour.

BIMvoice:
Both greptile findings on this PR are correct, and both are now fixed in 6b08ded2 (test-only — `git diff --stat` touches two `*.test.ts` files and no production file). We push back on a lot of bot findings, so it's worth saying plainly: these two landed.

### 1. Locale-sensitive assertion (P1) — confirmed

`packages/encoding/src/property-value.ts:123-125` formats via `toLocaleString`:

```ts
const formatted = Number.isInteger(value)
  ? value.toLocaleString()
  : value.toLocaleString(undefined, { maximumFractionDigits: 6 });
```

The old assertion at `packages/encoding/src/property-value.test.ts:68` was `expect(floatResult.displayValue).toBe('3.141593')`. Under a comma-decimal locale the correct output is `3,141593`, so the test failed on correct behaviour. Verified: `new Intl.NumberFormat('de-DE', { maximumFractionDigits: 6 }).format(3.14159265)` -> `3,141593`.

Fix compares the digit sequence (`displayValue.replace(/[^0-9]/g, '')` -> `'3141593'`), which is locale-agnostic and still kills the mutant the test was written for — the branch-swap makes the non-integer go through the plain `toLocaleString()` (3 fraction digits default), giving `3.142` -> `3142`.

### 2. Namespace gate only tested with an absent namespace (P2, security-relevant) — confirmed

`dispatchToBackend` (`packages/sdk/src/types.ts:769`) documents its own gate as stopping an attacker from reaching "prototype members (`__proto__`, `constructor`, `toString`, …)". The only negative test was the absent namespace `'nope'`, which does not pin that: `'nope' in backendObj` is false too, so relaxing `hasOwnProperty` to `in` survived.

The reachability chain the old test missed, for namespace `'__proto__'` / method `'toString'`:

1. `'__proto__' in backendObj` -> `true`, relaxed gate passes.
2. `backendObj['__proto__']` -> `Object.prototype` — non-null and `typeof === 'object'`, so it clears the null/typeof gate at `types.ts:773`.
3. `Object.prototype.hasOwnProperty.call(Object.prototype, 'toString')` -> `true`; `toString` is an *own* property of `Object.prototype`, so the method gate at `types.ts:778` passes too.
4. `fn.apply(ns, args)` — invoked. No throw.

That directly contradicts the function's own docstring, which is why it's worth a pinned test rather than just a comment. Added: dispatching `'__proto__'`/`'toString'` must throw `Unknown namespace '__proto__'`. The `'nope'` test is kept.

### Mutation verification

Every mutation applied by exact-substring replacement with an asserted replacement count, restored from a saved backup (never `git checkout --`).

| mutation | replacements | result |
|---|---|---|
| A: swap the two `toLocaleString` branches (`property-value.ts:123-125`) | `MUTATION A replacements=1` | **killed** — `encoding` 1 failed / 50 passed (51). Also killed under `LC_ALL=de_DE.UTF-8`: 1 failed / 50 passed. |
| B: namespace gate -> `if (!(namespace in backendObj))` (`types.ts:771`) | `MUTATION B (namespace gate -> in) replacements=1` | **killed only by the new test** — `sdk` 1 failed / 68 passed (69); the single failure is `rejects the inherited namespace '__proto__' …` with `AssertionError: expected [Function] to throw an error … Received: undefined`, i.e. `Object.prototype.toString` was invoked. The `'nope'` test passed under the mutant, confirming it never covered this. |
| C: method gate -> `if (!(method in ns))` (`types.ts:778`) | `MUTATION C (method gate -> in) replacements=1` | **already killed** — 3 failures from the pre-existing inherited-method tests (`toString`, `constructor`, and the `BimHost.dispatch` case). No new test needed here. |

Restored + clean: `encoding` 51 -> 51 passed (4 files), `sdk` 68 -> 69 passed (6 files). `pnpm typecheck` at the repo root: 34/34 tasks successful.

louistrue:
**Correction to my previous comment: my export concern does not hold, and I should have traced it before raising it.**

I said the locale-formatted `displayValue` "flows into CSV/JSON export, where a comma decimal separator in an unquoted CSV field is its own bug". I traced the consumers and that is wrong for numbers.

`packages/lists/src/engine.ts` — the path into schedules and their exports — deliberately does not use the formatted string for numerics:

```ts
const parsed = parsePropertyValue(value);
const display = parsed.displayValue;
…
// Try to preserve numeric values for sorting
if (typeof value === 'number') return value;
if (typeof value === 'boolean') return display;   // "True"/"False"
if (Array.isArray(value) && value.length === 2 && typeof value[1] === 'number') return value[1];
return display;
```

Numbers come back as **raw numbers**, and typed `[IFCREAL, 5.3]` pairs unwrap to the raw number too. The formatted string is only used for booleans and genuinely-textual values. The comment says why — preserving numerics for sorting — and that choice also happens to keep locale formatting out of the data path entirely.

The other consumer is `PropertySetCard.tsx`, which is UI rendering, where `toLocaleString` is exactly right.

So there is no production defect here and no separate issue to file. `toLocaleString` on `displayValue` is a display-layer concern that stays in the display layer.

Your fix stands on its own merits: comparing the digit sequence keeps the assertion locale-independent **and** still catches a branch swap, since the integer branch's default `maximumFractionDigits: 3` would yield `3142` rather than `3141593`. That is the harder property to preserve — the obvious locale-proofing (assert `typeof === 'string'`, or reuse `toLocaleString` in the expectation) would have made the test tautological.

Separately, the `__proto__` gate is verified:

```
### namespace gate: hasOwnProperty -> in
     × rejects the inherited namespace '__proto__' instead of resolving Object.prototype
      Tests  1 failed | 6 passed (7)
```

and your choice of `toString` as the method is better than the custom-prototype-method backend I suggested — `Object.prototype` owns `toString`, so the escape is reachable without inventing a fixture class.

louistrue:
Both remaining threads are stale — `6b08ded2` fixed both, and I verified each independently.

**`__proto__` namespace gate** — the mutation now dies:

```
### namespace gate: hasOwnProperty -> in
     × rejects the inherited namespace '__proto__' instead of resolving Object.prototype
      Tests  1 failed | 6 passed (7)
```

Your choice of `toString` as the method is better than the custom-prototype-method backend I proposed — `Object.prototype` owns `toString`, so the escape is reachable without inventing a fixture class.

**Locale assertion** — comparing the digit sequence is locale-independent *and* still catches a branch swap (the integer branch's default `maximumFractionDigits: 3` gives `3142`, not `3141593`). The obvious locale-proofing would have made it tautological.

And my export concern was wrong, as corrected above: `lists/engine.ts` returns raw numbers for numerics, so the locale-formatted string never reaches the data path.

Merging.

PR #2140, opened 2026-08-04
test(extensions,create): close 5 mutation-verified coverage gaps (1 security-relevant)

BIMvoice:
## Security-relevant — read first

**GAP 1 — `packages/extensions/src/capability/match.ts:110`.** `matchSegment`'s glob-prefix check is `requestedValue.startsWith(prefix)`. Mutating it to `.includes(prefix)` survived the full 600-test suite: the existing fixtures test `Pset_*` against `Pset_WallCommon` (matches either way) and against `Qto_WallBase` (matches neither way, no `Pset_` substring at all) — neither discriminates `startsWith` from `includes`. That gap matters because a substring match turns a prefix glob into an over-permissive capability grant: `model.mutate:Pset_*` would also match `model.mutate:Custom_Pset_Secret`. Added a fixture that contains the prefix without starting with it (`Pset_*` vs `Custom_Pset_Secret`), asserting `false`.

## Other gaps

**GAP 2 — `packages/extensions/src/manifest/validate.ts:210`.** `isActivationEvent`'s `v.length > p.length` mutated to `>=` survived, which (combined with the `startsWith` check) accepts a bare-prefix activation with an empty target, e.g. `activation: ["onCommand:"]`. Added a case asserting a bare `"onCommand:"` is rejected and a real `"onCommand:something"` is accepted.

**GAP 3 — `packages/create/src/in-store/wall.ts:59`.** `Thickness <= 0 || Height <= 0` mutated to `< 0` survived because every existing fixture used `Thickness: 0.2` (or similar) — none hit exactly zero. Added cases asserting exact-zero Thickness and exact-zero Height are both rejected.

**GAP 4 — `packages/create/src/in-store/wall.ts:84`.** The coplanarity guard `Math.abs(dz) > Math.max(1e-6 * wallLen, 1e-9)` mutated to `if (false)` survived because no fixture used a sloped (non-coplanar) Start/End. Added a case with `End: [5, 0, 2]` asserting rejection.

**GAP 5 — `packages/create/src/in-store/resolve-anchor.ts:48`.** `s > 0` mutated to `s >= 0` survived because only the no-source fallback path was exercised. The real `extractLengthUnitScale` never returns exactly 0 (its own internal guards route degenerate cases back to the 1.0 default), so the test mocks that one export — mirroring the existing pattern in `extract-walls-unit-scale.test.ts` — to return 0 and asserts `resolveSpatialAnchor` falls back to the metre default rather than accepting it.

## Method

Each gap was verified RED-first: applied the exact mutation described above via an exact-substring Python replacement (count asserted `== 1`), ran the new test and confirmed it failed for the stated reason, then restored the original file by copy (never `git checkout --`) and confirmed the full suite passes again.

## Counts

- `packages/extensions`: 600 tests / 53 files → 603 tests / 53 files (2 new assertions added to existing files)
- `packages/create`: 119 tests / 15 files → 123 tests / 16 files (1 new file, 3 new cases in `wall.test.ts`)

Test-only changes; no production code touched, no changeset needed.

## Test plan

- [x] `pnpm test` in `packages/extensions`: 53 files / 603 tests passing
- [x] `pnpm test` in `packages/create`: 16 files / 123 tests passing
- [x] `pnpm typecheck` (turbo, both packages) clean
- [x] Each gap RED-verified (mutation applied → new test fails) then GREEN-verified (file restored by copy → full suite passes)

### DIFF
--- a/packages/create/src/in-store/resolve-anchor-zero-scale.test.ts
+++ b/packages/create/src/in-store/resolve-anchor-zero-scale.test.ts
@@ -1,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `resolveSpatialAnchor` guards `extractLengthUnitScale`'s result with
+ * `s > 0` before trusting it — a zero (or negative/non-finite) scale must
+ * fall back to the metre default, not be accepted verbatim, since a zero
+ * scale would collapse every subsequently emitted length to 0. The real
+ * `extractLengthUnitScale` implementation never returns exactly 0 (its own
+ * guards route degenerate cases back to the 1.0 fallback), so this pins the
+ * `resolveSpatialAnchor` guard directly by mocking that one export, mirroring
+ * the pattern in `extract-walls-unit-scale.test.ts`.
+ */
+
+import { describe, expect, it, vi, afterEach } from 'vitest';
+
+vi.mock('@ifc-lite/parser', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@ifc-lite/parser')>();
+  return {
+    ...actual,
+    extractLengthUnitScale: () => 0,
+  };
+});
+
+const { IfcParser } = await import('@ifc-lite/parser');
+const { resolveSpatialAnchor } = await import('./resolve-anchor.js');
+
+/** Minimal IFC4 model with one storey — enough for `resolveSpatialAnchor`. */
+const STOREY_MODEL = `ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('t.ifc','',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=IFCPROJECT('0proj00000000000000000',$,'P',$,$,$,$,(#7),#9);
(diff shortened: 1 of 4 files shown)


louistrue:
Verified and merging. GAP 1 is the one worth leading with, and you were right to.

```
### startsWith -> includes
     × prefix glob does not match a value that merely contains the prefix
      Tests  1 failed | 602 passed (603)

### v.length > p.length  ->  >=
     × rejects a bare-prefix activation event with an empty target
      Tests  1 failed | 602 passed (603)
```

Restored: 603/603 both times.

**GAP 1 is a capability escalation, not a style nit.** `matchSegment` is the permission check, and `includes` turns a prefix glob into a substring grant — `model.mutate:Pset_*` would then also match `model.mutate:Custom_Pset_Secret`. An extension granted narrow prefix access silently gets everything containing that prefix.

What makes this a good find is the diagnosis of *why* 600 tests missed it. Your fixtures were `Pset_*` against `Pset_WallCommon` (matches under both) and against `Qto_WallBase` (matches under neither, and contains no `Pset_` substring at all). Neither discriminates the two operators. The missing case is the only one that can: a value that **contains the prefix without starting with it**. That is the shape of every genuinely useful negative fixture — it has to fail for the specific reason under test, not just fail.

GAP 2 is the same family one layer up: `>=` combined with the `startsWith` check accepts `"onCommand:"` with an empty target, so a manifest could register an activation event that binds to nothing and passes validation.

Both are in `packages/extensions`, which is the trust boundary for third-party code — exactly where a test that cannot fail is most expensive. Worth the security framing you gave it at the top of the description; I would not have prioritised reading it otherwise.

PR #2142, opened 2026-08-04
test(merge): cover two mutation-verified gaps in the ref-merge flow

BIMvoice:
Test-only. Two branches of `packages/merge/src/ref-flow.ts` were reached by **no** test in `merge`, `cli` or `collab-server`. A mutation probe confirmed both: the mutation below survived the full suite in each case. The production logic is correct — only the coverage was missing, so nothing outside the new test file changes.

Method for both: apply the exact mutation with an exact-substring replacement that asserts a replacement count of 1, re-read the mutated region to confirm the text changed, run the new test (RED), restore `ref-flow.ts` by copying back a saved backup (never `git checkout --`), run again (GREEN).

## Gap 1 — `checkRefPolicy` fails closed on a manifest-less candidate

```
-    (manifest === undefined || manifest.author.kind === 'agent') &&
+    (manifest !== undefined && manifest.author.kind === 'agent') &&
```

Replacement count 1. Under the mutation a candidate with **no provenance manifest** stops being treated as agent-authored and walks straight past the `requireHumanApproval` gate — which is exactly what the comment above the clause forbids ("a candidate with no provenance manifest could be an agent layer with the manifest stripped").

On the registry paths this is masked by defense in depth — the manual endpoint rejects manifest-less candidates on `requireHumanApproval` refs before calling `mergeIntoRef`, and `runAutoMerges` skips those refs entirely — so it is reachable unmasked only through `ifc layer merge` on the CLI.

New test `refuses a manifest-less candidate on a requireHumanApproval ref` merges a layer with no manifest into a protected ref and asserts `policy-failure` plus an unchanged ref. **RED under the mutation** (`expected 'merged' to be 'policy-failure'`), green with the code restored. Two counter-cases pin the boundary: the same candidate merges once `approvedBy` is supplied, and merges on a ref with no approval requirement.

## Gap 2 — `resolveAncestor`, the `base.kind === 'layer'` branch

```
-      const ids = refLayerIds.slice(0, idx + 1);
+      const ids = refLayerIds.slice(0, idx);
```

Replacement count 1. The mutation drops the base layer **itself** from the ancestor while still reporting `matched: true`, so the unrelated-base refusal cannot catch it and the three-way merge plans against a one-layer-short ancestor. The branch is reachable: `packages/mcp/src/tools/layer-store.ts:211` emits `{ kind: 'layer', id: ref }` when a draft is seeded from a specific published layer, and such a candidate can be pushed and merged.

The hard part here was building a fixture that **discriminates** the correct ancestor from the short one, not merely one that exercises the branch. It does. Ref stack is `[L0, L1, L2]`; the candidate declares `{ kind: 'layer', id: L1 }` and restates L1's `FireRating` verbatim while bumping an unrelated acoustic rating:

| | ancestor `FireRating` | theirs | ours | outcome |
|---|---|---|---|---|
| correct ancestor `[L0, L1]` | `REI60` | `REI60` (unchanged) | `REI90` | clean `merged`, ours wins |
| short ancestor `[L0]` | `REI30` | `REI60` (looks like an edit) | `REI90` | **conflict** |

So the merge *outcome* separates the two, not just the internal ids. Three tests fail under the mutation: the direct `resolveAncestor` assertion, the merge (`'conflicts'` instead of `'merged'`), and the preview (a non-empty `plan.conflicts`). All three pass restored.

## Counts

- Baseline: `Test Files 6 passed | 1 skipped (7)`, `Tests 65 passed | 4 skipped (69)`
- Final: `Test Files 7 passed | 1 skipped (8)`, `Tests 71 passed | 4 skipped (75)`

The skipped file is `src/real-model-fuzz.test.ts`, gated on absent LFS fixtures — unchanged and expected. `pnpm typecheck` passes at the repo root (34/34 tasks).

## Unrelated observation, offered as a question rather than a change

While reading the flow: `FsLayerRegistry` has no cross-process lock, so the `getRef` → `setRef` read-modify-write inside `mergeIntoRef` is atomic *per write* (temp file + fsync + rename) but is not serialized across two collab-server processes sharing one registry directory. Single-process safety appears to rest on `mergeIntoRef` being fully synchronous.

Is a single writer per registry directory an intended deployment constraint, or is multi-process sharing something you want supported? Nothing is changed here either way — flagging it only so the assumption is written down somewhere.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for ref-flow behavior involving manifest-less candidates and human-approval policies.
  * Added validation for layer-based ancestry resolution, approval handling, open-ref merging, conflict-free results, and preview behavior.
  * Expanded test fixtures for identified, publishable, and manifest-less layers.

### DIFF
--- a/packages/merge/src/ref-flow-policy-ancestor.test.ts
+++ b/packages/merge/src/ref-flow-policy-ancestor.test.ts
@@ -1,0 +1,248 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Two branches of the shared ref-merge flow that no test in `merge`, `cli`
+ * or `collab-server` reached:
+ *
+ * 1. `checkRefPolicy` failing CLOSED on a candidate with NO provenance
+ *    manifest. On the registry paths this is masked by defense in depth
+ *    (the manual endpoint rejects manifest-less candidates on
+ *    `requireHumanApproval` refs before calling `mergeIntoRef`, and
+ *    `runAutoMerges` skips those refs), so it is reachable unmasked only
+ *    through `ifc layer merge`.
+ *
+ * 2. `resolveAncestor`'s `base.kind === 'layer'` branch, which must include
+ *    the named base layer ITSELF in the ancestor stack. `packages/mcp`
+ *    emits a layer base when a draft is seeded from a specific published
+ *    layer, so such candidates do reach the merge flow.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { computeLayerId, computeStackHash, createProvenanceManifest, setProvenance } from '@ifc-lite/ifcx';
+import type { IfcxFile, IfcxNode, ProvenanceBase } from '@ifc-lite/ifcx';
+import { extractStackState } from './component-state.js';
+import { mergeIntoRef, resolveAncestor } from './ref-flow.js';
+import type { LayerRefStore, RefEntry } from './ref-flow.js';
+
+const FIRE = 'bsi::ifc::v5a::Pset_FireSafety::FireRating';
+const SOUND = 'bsi::ifc::v5a::Pset_Acoustic::Rw';
+
+class MemoryStore implements LayerRefStore {
+  private layers = new Map<string, IfcxFile>();
+  private refs = new Map<string, RefEntry>();
+  storeLayer(file: IfcxFile): string {
+    this.layers.set(file.header.id, structuredClone(file));
+    return file.header.id;
+  }
+  loadLayer(id: string): IfcxFile {
+    const f = this.layers.get(id);
+    if (!f) throw new Error(`no layer ${id}`);
+    return structuredClone(f);
+  }
+  getRef(name: string): RefEntry | undefined {
+    const e = this.refs.get(name);
+    return e ? structuredClone(e) : undefined;
+  }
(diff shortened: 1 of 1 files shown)


louistrue:
Verified and merging. Gap 1 is a fail-open policy hole, which is the right one to lead with.

```
### (manifest === undefined || agent)  ->  (manifest !== undefined && agent)
     × refuses a manifest-less candidate on a requireHumanApproval ref
      Tests  1 failed | 70 passed | 4 skipped (75)
```

Restored: 71 passed. Mutation confirmed present in the file before I believed the result.

Under that mutation a candidate with **no manifest at all** stops counting as agent-authored, so a `requireHumanApproval` ref would auto-merge it. That is failing *open* on an authorisation check — the candidate you know least about is the one that gets waved through. Exactly the branch you want a test on, and it had none across `merge`, `cli` or `collab-server`.

Your method statement is worth quoting, because it is the discipline that makes any of this trustworthy:

> apply the exact mutation with an exact-substring replacement that asserts a replacement count of 1, **re-read the mutated region to confirm the text changed**, run the new test (RED), restore by copying back a saved backup (**never `git checkout --`**), run again (GREEN)

Both halves matter. I have produced six mutations today that silently applied to nothing — wrong indentation, wrong path, wrong identifier, an unexpanded glob — and every one printed a clean green that reads exactly like "this test is vacuous". Re-reading the region is what separates those. And `git checkout --` as a restore is genuinely dangerous in a worktree: earlier today I used `rm -rf` on a directory to clean up two stray files and took sixteen pre-existing ones with it.

The `|| manifest === undefined` shape is worth keeping a comment on, since it reads like defensive noise and is in fact the fail-closed clause.

PR #2143, opened 2026-08-04
fix(build): allow `using` declarations in apps/viewer (#2130)

### LINKED ISSUE #2130: vite-plugin-top-level-await rejects `using` declarations (SWC parser lacks explicitResourceManagement)
## Summary

`using x = new GeometryProcessor(...)` cannot be used in `apps/viewer`. It type-checks, but the production build fails: **rolldown (Vite 8) panics while rendering chunks.**

This matters because `#1563` added `[Symbol.dispose]()` to `GeometryProcessor` specifically so that `using` would be the cheap, correct way to free the WASM handle — and the changelog advertises it as such. In the app where the highest-frequency handle leak lives, that form is unavailable.

## Reproduction

On `fix/1959-wasm-disposal-gate`, changing one declaration in `apps/viewer/src/hooks/useIfcLoader.ts`:

```ts
-const geometryProcessor = new GeometryProcessor({ … });
+using geometryProcessor = new GeometryProcessor({ … });
```

```
$ pnpm --filter viewer build

✓ 5840 modules transformed.
rendering chunks...

thread '<unnamed>' panicked at
  /Users/runner/.cargo/registry/.../miette-7.6.0/src/handlers/graphical.rs:1159:22:
Formatting argument out of range
 ELIFECYCLE  Command failed.       (exit 134)
```

Reverting the single keyword:

```
✓ built in 15.00s                  (exit 0)
```

`tsc --noEmit` is clean in **both** cases, so this is invisible to typecheck and only shows up in a real build.

The panic is inside `miette`, rolldown's diagnostic *formatter* — so rolldown hits some error on the lowered `using` output and then crashes rendering that error, which is why there is no useful message.

## Impact

- The ergonomic disposal form is unavailable across the whole viewer app. `#2129` had to use an explicit `try/finally` for the canonical load path instead, which covers less (a throw between construction and the `try` unwinds past it).
- Any future `using` — for `GeometryProcessor` or anything else `Disposable` — will fail the build the same way.
- It fails *late*, at chunk rendering, after 5840 modules. Someone adding `using` will get a Rust panic with no line number, having seen a clean typecheck.

## Suggested next steps

1. Confirm against the current rolldown/Vite version and narrow it to a minimal repro (a single `using` in a trivial entry) — mine is a real 5840-module build, which is not a good bug report upstream.
2. If it reproduces minimally, file upstream with rolldown.
3. Until then, prefer explicit `try/finally` in `apps/viewer` and treat the `#1563` changelog guidance as server/CLI/package-only.

Found while fixing #1959.

BIMvoice:
Fixes #2130.

## Root cause — two holes, not one

`vite-plugin-top-level-await` re-parses **every emitted ES chunk** with SWC in `generateBundle`. Two independent things break on a `using` declaration:

1. **`dist/bundle-info.js:44-47`** parses with `syntax: "ecmascript"`, `target: "es2022"` and no `explicitResourceManagement`, so `using` is a hard parse error — the diagnostic @louistrue narrowed to in the correcting comment.
2. **Not visible until (1) is fixed:** with the parser flag on, the AST now contains `UsingDeclaration` nodes, and `@swc/core`'s `Visitor` base class has no arm for them in `visitStatement` — so the plugin's own `FindPatternsVisitor` (`dist/find.js:26`) throws `Unknown statement type: UsingDeclaration`.

Fixing only the parser gets you a different crash, which is why this needs two patches.

## Fix

Two minimal `pnpm patch`es:

- `patches/vite-plugin-top-level-await@1.6.0.patch` — one line, `explicitResourceManagement: true` on the parse config.
- `patches/@swc__core@1.15.8.patch` — a `case "UsingDeclaration"` arm plus a `visitUsingDeclaration(n)` that walks `n.decls` exactly as `visitVariableDeclaration` walks `n.declarations`.

Both are additive; neither changes any existing path.

### Is the parser flag safe for code that has no `using`?

Checked the ASI hazard directly — `using` as an ordinary identifier, parsed and re-printed with the flag **off** and **on**:

| source | flag off | flag on |
|---|---|---|
| `var using = 5; console.log(using);` | `var using = 5; console.log(using);` | identical |
| `let using = 1;\nusing\n(2);` | `let using = 1; using(2);` | identical |
| `function f(using){return using}` | unchanged | identical |
| `const o={using:1};console.log(o.using);` | unchanged | identical |
| `using\nx\n` | `using; x;` | identical |

No divergence. And `pnpm build` (turbo, 40 tasks) passes on this branch with both patches applied, so nothing in the existing 5843-module viewer bundle is affected.

## Reproduction (RED) — the issue's own repro, end to end

`apps/viewer/src/hooks/useGridLines3D.ts:49`, one keyword changed (`const processor` → `using processor`), on the real viewer build with the patches **removed**:

```
$ pnpm --filter viewer build
thread '<unnamed>' panicked at .../miette-7.6.0/src/handlers/graphical.rs:1159:22
EXIT=134
```

Same edit **with** the patches:

```
$ pnpm --filter viewer build
✓ built in 14.73s        EXIT=0
```

That temporary edit is not part of this branch — the committed change is the patches plus the test.

## Regression guard

`apps/viewer/src/lib/using-declaration-build.test.ts` — 2 cases, driving a real `vite.build()` over a four-line fixture with `using` plus a top-level await (the TLA is needed or the plugin may skip the chunk and never reach the parse):

- `esnext` target: build resolves, **and** the emitted chunk still contains `using` — a "fix" that silently dropped the declaration (and therefore never disposed) would fail here.
- `es2022` target: esbuild lowers `using` to a try/finally stack, so the chunk is executable — asserts `['open:a','open:b','body','close:b','close:a']`, i.e. disposal actually runs, in reverse order.

## Mutation evidence

Each mutation was applied as an exact-substring replacement with an asserted replacement count of 1, followed by `pnpm install` to re-apply the patch, and the mutated text re-read from the installed package before running.

| # | mutation | verified in node_modules | result |
|---|---|---|---|
| 1 | `explicitResourceManagement: true` → `false` in the plugin patch (1 replacement) | `dist/bundle-info.js:47` reads `false` | **RED** — 2/2 fail, `x Using declaration is not enabled` |
| 2 | drop the `case "UsingDeclaration": return this.visitUsingDeclaration(stmt);` arm from the `@swc/core` patch, keeping the `visitUsingDeclaration` method (1 replacement) | `Visitor.js` has the method at `:417`, no case arm | **RED** — 2/2 fail, `Unknown statement type: UsingDeclaration` |

Mutation 2 is deliberately narrow: the method is left in place, so the kill is attributable to the dispatch arm alone and not to removing the whole patch. Patches restored from a saved copy (no `git checkout --`); GREEN reconfirmed, 2/2 pass.

## Test counts (`apps/viewer`)

| | tests | pass | fail | skipped |
|---|---|---|---|---|
| baseline (this branch, test file removed) | 2745 | 2743 | 0 | 2 |
| final | 2747 | 2745 | 0 | 2 |

`pnpm typecheck` at the root: 34/34 tasks green. `pnpm install --frozen-lockfile` succeeds with the committed lockfile. `node scripts/check-test-wiring.mjs` green.

## One thing to weigh before merging

The patches are keyed to exact versions (`vite-plugin-top-level-await@1.6.0`, `@swc/core@1.15.8`). A bump of either will fail the install loudly rather than silently dropping the fix — which I think is the right failure mode, but it is a maintenance cost, and `@swc/core` is a transitive dep that can move under a lockfile refresh. Happy to key them by name only (fuzzy apply, warn instead of fail) if you'd rather. The upstream fixes are one line each and worth filing regardless; this unblocks the repo in the meantime.

No changeset: no published package's API surface changes — the patches affect build tooling, and no source in the repo uses `using` today.

### DIFF
--- a/apps/viewer/src/lib/using-declaration-build.test.ts
+++ b/apps/viewer/src/lib/using-declaration-build.test.ts
@@ -1,0 +1,116 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression guard for #2130.
+ *
+ * `vite-plugin-top-level-await` re-parses every emitted ES chunk with SWC. Its
+ * parser config asks for plain `ecmascript` without `explicitResourceManagement`,
+ * and `@swc/core`'s `Visitor` base class has no `UsingDeclaration` arm — so a
+ * single `using x = ...` anywhere in `apps/viewer` failed the production build
+ * long after a clean `tsc --noEmit`, with a diagnostic that crashed the error
+ * formatter and showed no line number.
+ *
+ * Both holes are closed by patches (`patches/vite-plugin-top-level-await@1.6.0.patch`,
+ * `patches/@swc__core@1.15.8.patch`). This test drives a real `vite.build()` over
+ * a four-line fixture that uses `using`, so the guard is the actual build path
+ * rather than an assertion about patch file contents.
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, mkdirSync, readdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { build } from 'vite';
+import topLevelAwait from 'vite-plugin-top-level-await';
+
+/** The fixture the viewer cannot express today: a `using` declaration plus a TLA. */
+const FIXTURE = `
+export const order: string[] = [];
+
+class Handle {
+  constructor(private readonly name: string) {
+    order.push('open:' + this.name);
+  }
+  [Symbol.dispose]() {
+    order.push('close:' + this.name);
+  }
+}
+
+export function withHandles(): number {
+  using a = new Handle('a');
+  using b = new Handle('b');
+  order.push('body');
+  return 42;
+}
+
+// Present so the plugin actually engages: without a top-level await it may skip
+// the chunk entirely and the parse that #2130 trips over never happens.
+export const tla = await Promise.resolve('tla');
+`;
+
+interface BuiltChunk {
+  code: string;
+  order: string[];
+  result: number;
+}
+
+/** Builds the fixture at `target`, returning the single emitted chunk. */
+async function buildFixture(target: string, run: boolean): Promise<BuiltChunk> {
+  const root = mkdtempSync(join(tmpdir(), 'ifclite-2130-'));
+  try {
+    writeFileSync(join(root, 'entry.ts'), FIXTURE);
+    const outDir = join(root, 'dist');
+    mkdirSync(outDir, { recursive: true });
+    await build({
+      root,
+      logLevel: 'silent',
+      configFile: false,
+      build: {
+        target,
+        outDir,
+        emptyOutDir: true,
+        lib: { entry: join(root, 'entry.ts'), formats: ['es'], fileName: 'chunk' },
+      },
+      plugins: [topLevelAwait()],
+    });
+    const emitted = readdirSync(outDir).filter((f) => f.endsWith('.mjs') || f.endsWith('.js'));
+    assert.equal(emitted.length, 1, `expected exactly one chunk, got ${emitted.join(', ')}`);
+    const file = join(outDir, emitted[0]!);
+    const code = readFileSync(file, 'utf8');
+    if (!run) return { code, order: [], result: 0 };
+    const mod = (await import(file)) as {
+      withHandles: () => number;
+      order: string[];
+      tla: string;
+    };
+    const result = mod.withHandles();
+    return { code, order: mod.order, result };
+  } finally {
+    rmSync(root, { recursive: true, force: true });
+  }
+}
+
+describe('using declarations survive the vite-plugin-top-level-await pipeline (#2130)', () => {
+  it('builds a chunk containing a `using` declaration at target esnext', async () => {
+    // Before the patches this rejected with an SWC parse error
+    // ("Using declaration is not enabled"), surfacing in the real viewer build
+    // as a miette panic with no line number.
+    const { code } = await buildFixture('esnext', false);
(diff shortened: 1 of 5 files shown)


louistrue:
Fixes #2130, and **you found a second hole I missed.** Verified end to end.

My correcting comment on #2130 narrowed it to the SWC parser flag and stopped there. Your point 2 is the one I would have walked into:

> with the parser flag on, the AST now contains `UsingDeclaration` nodes, and `@swc/core`'s `Visitor` base class has no arm for them in `visitStatement` — so the plugin's own `FindPatternsVisitor` throws `Unknown statement type: UsingDeclaration`

"Fixing only the parser gets you a different crash" is exactly right, and it is why a one-line patch would have looked like a fix and produced a new failure at the same stage.

## Verified

**The exact repro from #2130** — `const` → `using` in `useGridLines3D.ts`, full viewer production build:

```
BUILD WITH `using`: exit=0
✓ built in 14.99s
using-related errors: 0
```

Previously: `exit=134`, miette panic. Control build without `using` on the same tree: also clean, so this is the keyword and not the environment.

**The regression test genuinely pins the patches.** Reverting just the parser flag (`explicitResourceManagement: true` → `false`) in the installed package kills both tests:

```
    not ok 1 - builds a chunk containing a `using` declaration at target esnext
    not ok 2 - preserves disposal semantics when the target lowers `using`
    # pass 0  # fail 2
```

Restored: 2/2. So a lockfile refresh that drops the patches fails CI rather than silently reintroducing the build break — which matters a lot for `pnpm patch`, since patches are the easiest thing to lose in a dependency bump and the symptom is a Rust panic with no line number.

The second test — disposal semantics under lowering, not just "it parsed" — is the one I would have skipped. Parsing is necessary but proves nothing about whether the emitted code still calls `[Symbol.dispose]`.

## Note for the record

This unblocks the ergonomic form for `apps/viewer`. The disposal gate (#2129, now on main) accepts `using`, `try/finally`, and disposer-callback shapes equally, so nothing has to change there — but `useIfcLoader`'s handle is shared with the parser via `getApi()`, so that particular site still needs its disposer rather than a `using`. Worth not "simplifying" it back now that the keyword builds.

Also worth flagging: this is a patch against upstream, so it is worth an upstream issue on `vite-plugin-top-level-await` if one does not exist — the `Visitor` gap is arguably `@swc/core`'s to fix, and carrying two patches indefinitely has its own cost.

PR #2144, opened 2026-08-04
test(mcp): close 10 mutation-verified coverage gaps

BIMvoice:
Mutation sweep of `packages/mcp` — the last package not yet swept. Twelve mutations against the load-bearing logic; **ten survived**, each now killed by a RED-verified test.

The largest hole was scope gating. Every fixture in the package built its context with `fullScope()` or `readOnlyScope()`, neither of which sets `modelIds`, so the entire per-model access-control feature was vacuous: `modelAllowed` could be replaced by `return true` and all 184 tests stayed green. A token narrowed to one model could read, query and unload every other model in the process, and `model_list` would enumerate them all.

## Mutation results

Every mutation was applied by exact-substring replacement asserting a replacement count of exactly 1, the mutated region re-read to confirm the text changed, and restored from a saved backup.

| # | Location | Mutation | n | Result |
|---|---|---|---|---|
| M1 | `auth/scope.ts` `scopeAllows` | drop `\|\| includes('admin')` | 1 | SURVIVED |
| M2 | `auth/scope.ts` `modelAllowed` | `return true` unconditionally | 1 | SURVIVED |
| M3 | `tools/validation.ts` `scoreFromIssues` | drop `- warns * 10` | 1 | SURVIVED |
| M4 | `tools/validation.ts` `summarizeIdsReport` | `ents.length > 0` → `true` | 1 | SURVIVED |
| M5 | `transport/http.ts` `sameScope` | drop the `modelIds` comparison | 1 | SURVIVED |
| M6 | `tools/util.ts` `paginate` | `offset + limit < total` → `<=` | 1 | SURVIVED |
| M7 | `tools/util.ts` `resolveModel` | `count() > 1` → `> 2` | 1 | SURVIVED |
| M8 | `tools/layer-store.ts` `resolveBase` | `kind: 'layer'` → `kind: 'stack'` | 1 | SURVIVED |
| M9 | `safe-path.ts` `ensureWithinRoots` | drop `absolute === root` | 1 | SURVIVED |
| M10 | `safe-path.ts` `rejectIfSensitiveHome` | disable the guard | 1 | **KILLED** (4 tests) |
| M11 | `tools/discovery.ts` `model_list` | drop the `modelAllowed` filter | 1 | SURVIVED |
| M12 | `tools/validation.ts` `model_audit` | `count > 1` → `> 2` | 1 | **KILLED** (1 test) |

M10 and M12 are the controls: they prove the suite was genuinely running against the mutated source, so the ten survivors are real gaps rather than an unrun suite. M12 in particular proves `model_audit` is reachable from the existing tests, so M3's survival is a scoring gap and not dead code.

No survivor is an equivalent mutant — each changes an observable answer, listed below.

## Gaps closed

**M2 / M11 — per-model access control (`auth/scope.ts`, `tools/util.ts`, `tools/discovery.ts`).** `AuthScope.modelIds` narrows a token to a subset of loaded models, enforced at `resolveModel`/`assertModelAccess`, `model_list` and `model_unload`. None of it had a test. New `src/tools/model-scope.test.ts` loads two models under a scope narrowed to one and asserts the denial is `PERMISSION_DENIED` (not `MODEL_NOT_FOUND`, which would tell the agent to load a model that is already loaded), that `model_list` shows one row **and** a count of 1, that `model_unload` and `query_entities` refuse the other model, and that the `MODEL_NOT_FOUND` `available` list does not leak out-of-scope ids. Nine tests fail under M2.

**M1 — `scopeAllows` admin escalation.** `fullScope()` short-circuits on the first `includes` for all five scopes, so `|| includes('admin')` was unreachable. An operator minting an admin-only token would find every read tool denied. New `src/auth/scope.test.ts` uses a `{ scopes: ['admin'] }` token, with `{ scopes: ['export'] }` → `'mutate'` as the counter-example so the test asserts admin-specific escalation rather than "any scope implies the rest".

**M5 — HTTP session identity (`transport/http.ts`).** The three existing tests all differ by `user`, which `sameScope` rejects on its very first comparison, leaving the scope-set and `modelIds` arms unreachable. A token narrowed to model A could take over a session opened by one narrowed to model B and inherit its layer drafts. Added four tests using tokens that share `user` and permissions and differ only in narrowing, each with an identical-token counter-example that must still get a 200 — plus a reordered-scopes token that must still reuse its session, covering the sort.

**M7 — `resolveModel`.** With exactly two models loaded and no `model_id`, the caller would silently get one of them instead of `MODEL_REQUIRED`.

**M6 — `paginate`.** The two existing tests use offset+limit=3 against total 5 and offset+limit=5 against total 3, straddling the boundary without touching it. A page landing exactly on the end would report `truncated: true`, so an agent paginating a result set whose size is a multiple of its page size would request empty pages forever. New test pins the exact boundary with a one-short case as the counter-example.

**M8 — `resolveBase` provenance kind (`tools/layer-store.ts`).** Only `seedDraftDoc` had tests. `resolveAncestorFiles` dispatches on `ProvenanceBase.kind` — a `'stack'` base is matched against a stack hash, a `'layer'` base against a layer id — so a mislabelled base resolves to an empty ancestor set: the draft is silently rebased onto nothing and the merge planner gets the wrong baseline. The new test asserts the `kind` and, directly, the property that makes it observable (`resolveAncestorFiles` with a `'stack'` base carrying a layer id returns `[]`). Also covers `refLayerFiles`' dangling-layer and unknown-ref errors.

**M3 / M4 — audit and IDS scoring (`tools/validation.ts`).** `validation.ts` had no test file; `read-after-write.test.ts` covers which *issues* the audit raises but never the arithmetic that turns them into the scores an agent reads and reports onward. New `src/tools/validation.test.ts` uses two fixtures that discriminate the weights — one whose only structural finding is a warning (90) and one whose only finding is an error (75) — so the two weights cannot be collapsed and still satisfy both. For IDS, a specification whose applicability matches nothing must count as **failed**: counting it passed is how an applicability typo becomes a green report, the most misleading thing this tool can say. Two counter-examples (a spec that matches and passes, one that matches and fails) ensure the assertion is about emptiness, not about `passedSpecifications` being stuck at zero.

**M9 — `ensureWithinRoots`.** Every other test passes a path strictly inside a root, so the equality arm was never exercised. An operator who passed `--allow <dir>` and then named that directory would be told their own allowed root is outside allowed roots. The root's parent is asserted still refused, so the test is about the boundary and not about the guard being off.

## Pre-existing failure fixed

`export-usd.test.ts` failed on macOS before any change: the fixture compared against an uncanonicalised `tmpdir()` path while the tool correctly returns the realpath (`tmpdir()` is `/var`, a symlink to `/private/var`, and every tool path goes through `resolveSafePath`). The production path was right; the expectation was not. Same fix and same reason `safe-path.test.ts` already documents.

## Baseline

`packages/mcp`: 20 files / 184 tests, 1 pre-existing failure → **23 files / 231 tests, all passing**. `pnpm typecheck` green at the repo root. Test-only changes; no changeset (test files are excluded from `dist` by `tsconfig.json` and by `package.json` `files`).

### DIFF
--- a/packages/mcp/src/auth/scope.test.ts
+++ b/packages/mcp/src/auth/scope.test.ts
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Scope gating is the only thing standing between a read-only or
+ * model-narrowed token and the full tool surface, and until now every test in
+ * the package built its context with `fullScope()` or `readOnlyScope()` —
+ * neither of which sets `modelIds`. The per-model allowlist could therefore be
+ * deleted outright (`modelAllowed` → `return true`) and the suite stayed green,
+ * as could the `admin` escalation arm of `scopeAllows`, because no fixture ever
+ * held a scope that needed it.
+ */
+
+import { describe, expect, it } from 'vitest';
+import {
+  FULL_ACCESS,
+  READ_ONLY,
+  fullScope,
+  modelAllowed,
+  readOnlyScope,
+  scopeAllows,
+  type AuthScope,
+} from './scope.js';
+
+describe('scopeAllows', () => {
+  it('permits a tool that declares no scope', () => {
+    expect(scopeAllows({ scopes: [] }, undefined)).toBe(true);
+  });
+
+  it('permits exactly the scopes the token holds', () => {
+    const s = readOnlyScope();
+    expect(scopeAllows(s, 'read')).toBe(true);
+    expect(scopeAllows(s, 'validate')).toBe(true);
+    expect(scopeAllows(s, 'export')).toBe(true);
+    expect(scopeAllows(s, 'mutate')).toBe(false);
+    expect(scopeAllows(s, 'admin')).toBe(false);
+  });
+
+  it('lets `admin` stand in for a scope the token does not list', () => {
+    // The escalation arm, which no `fullScope()` fixture can reach: a token
+    // holding *only* `admin` short-circuits on the first `includes` for every
+    // one of the five scopes, so `|| includes('admin')` never runs. An operator
+    // minting an admin-only token would find every read tool denied.
+    const adminOnly: AuthScope = { scopes: ['admin'] };
+    for (const required of ['read', 'validate', 'mutate', 'export'] as const) {
+      expect(scopeAllows(adminOnly, required), required).toBe(true);
+    }
+  });
+
+  it('does not let a non-admin scope stand in for another', () => {
+    // Counter-example to the test above: escalation is `admin`-only, not
+    // "holding any scope implies the rest".
+    expect(scopeAllows({ scopes: ['export'] }, 'mutate')).toBe(false);
+  });
+});
+
+describe('modelAllowed', () => {
+  it('permits every model when no allowlist is configured', () => {
+    expect(modelAllowed({ scopes: ['read'] }, 'anything')).toBe(true);
+    expect(modelAllowed({ scopes: ['read'], modelIds: [] }, 'anything')).toBe(true);
+  });
+
+  it('permits only the listed models when an allowlist is configured', () => {
+    const narrowed: AuthScope = { scopes: ['read'], modelIds: ['alpha', 'beta'] };
+    expect(modelAllowed(narrowed, 'alpha')).toBe(true);
+    expect(modelAllowed(narrowed, 'beta')).toBe(true);
+    expect(modelAllowed(narrowed, 'gamma')).toBe(false);
+  });
+
+  it('is not defeated by `admin` — the allowlist is orthogonal to scopes', () => {
+    expect(modelAllowed({ scopes: ['admin'], modelIds: ['alpha'] }, 'gamma')).toBe(false);
+  });
+});
+
+describe('scope constructors', () => {
+  it('hand back copies, so a caller mutating one does not widen the next', () => {
+    const a = fullScope();
+    a.scopes = ['read'];
+    expect(fullScope().scopes).toEqual(FULL_ACCESS.scopes);
+
+    const b = readOnlyScope();
+    b.modelIds = ['only-this-one'];
+    expect(readOnlyScope().modelIds).toBeUndefined();
+    expect(READ_ONLY.scopes).not.toContain('mutate');
+  });
+
+  it('read-only really is read-only', () => {
+    expect(scopeAllows(readOnlyScope(), 'mutate')).toBe(false);
+    expect(scopeAllows(fullScope(), 'mutate')).toBe(true);
+  });
+});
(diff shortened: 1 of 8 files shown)


louistrue:
Verified the headline claim, and it is as bad as you say. **`modelAllowed` → `return true` kills 8 tests now; before this PR it killed none.**

```
### modelAllowed -> return true
     × is denied the model it is not allowed to, with PERMISSION_DENIED not MODEL_NOT_FOUND
     × does not leak out-of-scope ids in the MODEL_NOT_FOUND `available` list
     × is denied by assertModelAccess directly, admin scope notwithstanding
     × sees only the permitted model from model_list
     × cannot unload a model outside its allowlist
      Tests  8 failed | 223 passed (231)
```

Restored: 231/231. Mutation confirmed in the file first.

**The root cause is the more useful finding than the fix.** Every fixture in the package built its context with `fullScope()` or `readOnlyScope()`, and neither sets `modelIds` — so `modelAllowed`'s first line (`if (!scope.modelIds || length === 0) return true`) short-circuited in *every* test. The feature wasn't lightly tested; it was structurally unreachable from the suite. That is the most expensive shape of untested code, because the tests look thorough and the coverage number looks fine.

And the consequence is not subtle: a token narrowed to one model could read, query and unload every other model in the process, and `model_list` would enumerate them all. In `packages/mcp` — the surface an external agent talks to — that is the access-control boundary itself.

Two of the new tests are better than "denied vs allowed":

- **`PERMISSION_DENIED` not `MODEL_NOT_FOUND`** pins the error *code*, not just the refusal. Returning `MODEL_NOT_FOUND` for an out-of-scope model would still deny access while telling the caller something false.
- **`does not leak out-of-scope ids in the available list`** is the one I would not have thought to write. A correct denial that helpfully lists every model in the process is still an enumeration leak — the denial succeeds and the information escapes anyway.

Ten survivors out of twelve in one package is a striking ratio. Worth saying: this was the last package in the sweep, so that number is a reasonable proxy for what the sweep as a whole was worth.

Waiting on the last CI lane, then merging.

BIMvoice:
One correction, since it affects what you'd expect next: this is not the last package in the sweep. `packages/mcp` was the last of the *first* batch. Twelve packages have still never been mutation-swept:

```
bcf  clash  codegen  create-ifc-lite  embed-protocol  embed-sdk
renderer  server-bin  solar  spatial  viewer  wasm
```

(`packages/viewer`, distinct from `apps/viewer` which was swept.)

Two more sweeps are in flight right now — `query` + `lens`, and `diff` + `provenance` — chosen because a wrong answer in either pair is silent rather than loud. Content-keyed diff matching only landed in #1891, so that logic is young enough to be worth probing early.

So the 10-of-12 ratio is not a closing number. It is one data point, and I would not extrapolate it forward: `mcp` had a specific structural reason for being that bad — a shared fixture helper that made an entire feature unreachable — and packages without that shape should do better. If the next sweeps come back with ratios anywhere near this one, that is a finding about the fixture conventions across the repo rather than about any single package.

Worth flagging one thing in the other direction, too. `export-usd.test.ts` was already failing on macOS before this PR — the fixture compared an uncanonicalised `tmpdir()` against the tool's correctly realpath'd return (`/var` vs `/private/var`). I fixed it test-side, since the product behaviour was right. But it means the package's baseline was red and had been for a while without anyone noticing, which is its own small signal.

PR #2145, opened 2026-08-04
test(query,lens): close five mutation-verified coverage gaps

BIMvoice:
Mutation sweep of `packages/query` and `packages/lens` — neither had been swept before, and both sit on the data-access path where a silent wrong answer is invisible to the user.

**26 mutations run**, each applied with a scripted exact-substring replacement that asserts a single occurrence, with the mutated region re-read and confirmed changed before any result was believed. **Two deliberate controls** (`matchesIfcType`'s subtype clause → `false`; `matchesPsetFilter`'s pset-name guard inverted) were killed, proving the harness ran against mutated source. **21 mutations were killed** by the existing suites — both packages are in good shape overall. Five survivors are closed here; three more are argued equivalent below.

Test-only change. **The production logic is already correct** — no behaviour is modified.

Every test below was RED-verified under its stated mutation and GREEN with the source restored from a saved backup.

## Gaps closed

| # | Surviving mutation | What a user would see | Killing test |
|---|---|---|---|
| 1 | `entity-node.ts`: `prev !== undefined && prev <= currentDepth` → `prev !== undefined` | `traverse(rel, depth)` silently omits descendants that only fit inside `depth` via a shorter route discovered later | `entity-node.test.ts` › re-expands a node re-reached at a shallower depth |
| 2 | `property-table.ts`: `prop.value === value` → `== value` | `findEntities(pset, 'Count', 0)` also returns entities whose value is `false`; `…, 1` also returns `'1'` | `property-table.test.ts` › should compare property values strictly, not coercively |
| 3 | `property-table.ts`: `results.push(entityId); break;` → drop the `break` | an entity carrying the same property in two same-named property sets is returned twice | `property-table.test.ts` › should report an entity once when two same-named psets both match |
| 4 | `engine.ts`: `if (!colorMap.has(id)) colorMap.set(id, rgba)` → unconditional set | a multi-material element renders in its *smallest* material group's colour instead of its largest, so the auto-color legend and the 3D view disagree | `engine.test.ts` › renders a multi-material element in its LARGEST material group colour |
| 5 | `discovery.ts`: `count < SAMPLE_SIZE` → `count <= SAMPLE_SIZE` | the documented 30-per-type discovery sampling cap was unpinned in both directions | `discovery.test.ts` › samples at most 30 entities per type, and budgets each type separately |

### Notes on fixture design

- **#1** — the diamond deliberately walks the LONG branch first, so the join node is committed at depth 3 before the short branch reaches it at depth 2. Without the re-expansion the leaf is unreachable within `depth: 3`. A `depth: 2` counter-example asserts the leaf is *absent*, so the test cannot be satisfied by an implementation that ignores `depth`. It also does not over-constrain: the weaker `prev < currentDepth` variant (output-equivalent, see below) still passes.
- **#2** — each value is asserted findable under its own type first, so a negative cross-type result cannot be an artefact of `findEntities` being blind.
- **#3** — a second, non-matching entity is present so the assertion is not trivially satisfied by a one-element table.
- **#4** — the two shared elements list their materials in *opposite* orders, which discriminates "largest group wins" from "whichever material the element lists first". The existing `#1366` test asserts only `colorMap.has(id)`; the new one also re-asserts bucket membership so the colour choice cannot have narrowed the groups.
- **#5** — the 30th entity's pset is asserted *present* (positive control) alongside the 31st's absence, and the trailing type of three entities must still be sampled — a counter-example against a store-wide rather than per-type budget.

## Survivors argued equivalent (no test added)

- `entity-query.ts` `if (matches.has(id)) continue;` — a pure short-circuit inside the on-demand quantity loop. `matches.add` is idempotent and `matchesQsetFilter` is side-effect-free, so removing the skip cannot change `matches`. Performance only.
- `entity-node.ts` `prev <= currentDepth` → `prev < currentDepth` — a node re-reached at the *same* depth gets re-expanded, but `firstVisit` gates every push and `currentDepth > depth` bounds the recursion, so the returned set is identical. Performance only.
- `matching.ts` `valueEquals(String(value), …)` → `String(value) === …` in `matchesQuantity` — `valueEquals` differs from `===` only for the strings `"true"`/`"false"` in differing case, which the numeric IFC quantity types (`IfcQuantityLength`/`Area`/`Volume`/`Count`/`Weight`) cannot produce.
- `engine.ts` `filter(r => r.enabled)` → `filter(r => r.enabled !== false)` — `LensRule.enabled` is a required `boolean`, so the two differ only on values the type forbids.

## Verification

- Baseline: `@ifc-lite/query` 11 files / 150 tests, `@ifc-lite/lens` 4 files / 95 tests — both green, no pre-existing failures.
- After: `@ifc-lite/query` 11 files / **153** tests, `@ifc-lite/lens` 4 files / **97** tests.
- `pnpm typecheck` at the repo root: 34/34 tasks successful.
- Both packages use `vitest run`. `@ifc-lite/query` needs its workspace dependencies built first (`pnpm turbo build --filter=@ifc-lite/query^...`).

No changeset: test-only, no published surface change (matching the precedent set by #2142).

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for limiting data-source discovery to 30 entities per type.
  * Added coverage for consistent coloring of multi-material elements.
  * Added traversal tests for correct depth handling and duplicate-free results.
  * Added property matching tests to ensure strict value comparisons and unique results.

### DIFF
--- a/packages/lens/src/discovery.test.ts
+++ b/packages/lens/src/discovery.test.ts
@@ -124,6 +124,43 @@ describe('discoverDataSources', () => {
     expect(result.materials).not.toContain('Wall Type A');
   });
 
+  // The 30-per-type sampling cap is the whole reason discoverDataSources is
+  // affordable on a large model, and nothing pinned it: neither the exact
+  // boundary nor the fact that the budget is PER TYPE rather than store-wide.
+  it('samples at most 30 entities per type, and budgets each type separately', () => {
+    // 31 walls (ids 1..31) each carrying a uniquely-named pset, then 3 slabs
+    // (ids 101..103). If the cap were store-wide the slabs would never be
+    // reached; if the boundary were off by one, wall 31 would leak in.
+    const walls = Array.from({ length: 31 }, (_, i) => i + 1);
+    const slabs = [101, 102, 103];
+    const provider = createMockProvider({
+      getEntityCount: () => walls.length + slabs.length,
+      forEachEntity: (cb) => {
+        for (const id of walls) cb(id, 'm1');
+        for (const id of slabs) cb(id, 'm1');
+      },
+      getEntityType: (id) => (id >= 101 ? 'IfcSlab' : 'IfcWall'),
+      getPropertySets: (id) =>
+        id >= 101
+          ? [{ name: `Pset_Slab_${id}`, properties: [{ name: 'IsExternal', value: false }] }]
+          : [{ name: `Pset_Wall_${id}`, properties: [{ name: 'IsExternal', value: true }] }],
+    });
+
+    const names = new Set(discoverDataSources(provider, { properties: true }).propertySets!.keys());
+
+    // Positive controls: sampling genuinely ran, right up to the boundary.
+    expect(names.has('Pset_Wall_1')).toBe(true);
+    expect(names.has('Pset_Wall_30')).toBe(true);
+    // The 31st wall is over budget.
+    expect(names.has('Pset_Wall_31')).toBe(false);
+    // Counter-example to a store-wide budget: the slabs come after 31 walls
+    // and must still be sampled, all three of them.
+    expect(names.has('Pset_Slab_101')).toBe(true);
+    expect(names.has('Pset_Slab_102')).toBe(true);
+    expect(names.has('Pset_Slab_103')).toBe(true);
+    expect(names.size).toBe(33);
+  });
+
   it('returns empty result when no categories requested', () => {
     const provider = createMockProvider();
     const result = discoverDataSources(provider, {});
--- a/packages/lens/src/engine.test.ts
+++ b/packages/lens/src/engine.test.ts
@@ -300,6 +300,62 @@ describe('evaluateAutoColorLens', () => {
     }
   });
 
+  it('renders a multi-material element in its LARGEST material group colour (#1366)', () => {
+    // The test above proves a multi-material element joins every bucket, but
+    // only that it gets *some* colour. Which one is the load-bearing part: an
+    // element renders once, so the engine keeps the first assignment and the
+    // groups are sorted by count desc — i.e. the element takes the colour of
+    // its biggest material group.
+    //
+    // Gypsum:     1, 2, 3, 5  (4 elements — the larger group)
+    // Insulation: 3, 4, 5     (3 elements)
+    // Element 3 lists Gypsum first, element 5 lists Insulation first. Both
+    // must end up Gypsum-coloured, which discriminates "largest group wins"
+    // from "whichever material the element happens to list first".
+    const materials = new Map<number, string[]>([
+      [1, ['Gypsum']],
+      [2, ['Gypsum']],
+      [3, ['Gypsum', 'Insulation']],
+      [4, ['Insulation']],
+      [5, ['Insulation', 'Gypsum']],
+    ]);
+    const provider: LensDataProvider = {
+      getEntityCount: () => materials.size,
+      forEachEntity: (cb) => { for (const id of materials.keys()) cb(id, 'm1'); },
+      getEntityType: () => 'IfcWall',
+      getPropertyValue: () => undefined,
+      getPropertySets: () => [],
+      getMaterialNames: (id) => materials.get(id) ?? [],
+    };
+
+    const result = evaluateAutoColorLens({ source: 'material' }, provider);
+
+    const byName = new Map(result.legend.map(e => [e.name, e]));
+    const gypsum = byName.get('Gypsum')!;
+    const insulation = byName.get('Insulation')!;
+    expect(gypsum.count).toBe(4);
+    expect(insulation.count).toBe(3);
+    // Legend is ordered by count desc, so Gypsum is allocated first.
+    expect(result.legend[0].name).toBe('Gypsum');
+
+    const gypsumRgba = hexToRgba(gypsum.color, 1);
+    const insulationRgba = hexToRgba(insulation.color, 1);
+    expect(gypsumRgba).not.toEqual(insulationRgba);
+
+    // Single-material elements anchor each colour (positive controls).
+    expect(result.colorMap.get(1)).toEqual(gypsumRgba);
+    expect(result.colorMap.get(4)).toEqual(insulationRgba);
+    // Shared elements take the larger group's colour, regardless of the order
+    // their own material list happens to be in.
+    expect(result.colorMap.get(3)).toEqual(gypsumRgba);
+    expect(result.colorMap.get(5)).toEqual(gypsumRgba);
+    expect(result.colorMap.get(3)).not.toEqual(insulationRgba);
+    expect(result.colorMap.get(5)).not.toEqual(insulationRgba);
+    // Membership is still multi-valued — the colour choice must not have
+    // narrowed the buckets.
+    expect(new Set(result.ruleEntityIds.get(insulation.id))).toEqual(new Set([3, 4, 5]));
+  });
+
   it('falls back to single getMaterialName when getMaterialNames is absent', () => {
     const provider: LensDataProvider = {
       getEntityCount: () => 1,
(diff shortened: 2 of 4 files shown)


louistrue:
The method here is the best in the sweep, and CodeRabbit's second point is right — I confirmed both by mutation.

## Your two deliberate controls are the right idea

> **Two deliberate controls** (`matchesIfcType`'s subtype clause → `false`; `matchesPsetFilter`'s pset-name guard inverted) were killed, proving the harness ran against mutated source.

That is the systematic answer to a failure mode I hit **seven times today**: a mutation that silently applies to nothing prints a green run indistinguishable from "this test is vacuous", and it fails in the direction that makes you doubt *correct* work. Wrong indentation, wrong path, wrong identifier, an unexpanded glob — each produced a confident false negative. Running a mutation you *expect* to die, alongside the ones you are testing, proves the harness before you trust any survivor. Worth keeping as standard.

Reporting that 21 of 26 were already killed is the other half of it — a sweep that only lists survivors reads like a package in trouble.

## CodeRabbit's dedup point holds

Your new test pins the `break`:

```
### remove the break (entity pushed once per matching pset)
     × should report an entity once when two same-named psets both match
      Tests  1 failed | 12 passed (13)
```

But the complementary direction is unpinned:

```
### return results.slice(0, 1)   — collapse to the first matching entity
      Tests  13 passed (13)      ← survives
```

The fixture's second entity (8) carries `Status: 'Existing'`, so it never matches the same query as entity 7. Nothing asserts that **two distinct entities matching one query both survive**. A dedup that over-collapsed — the natural over-correction if someone "fixes" duplicates with a `Set` keyed wrongly, or an early `return` — passes today.

Giving entity 8 `Status: 'New'` and asserting `toEqual([7, 8])` closes it, and keeps the existing duplicate assertion intact.

Worth noting my own first attempt at this mutation targeted `Array.from(new Set(results))`, which does not exist — the dedup here is the `break`. My script asserted the occurrence count, printed `occ=0`, and refused to proceed rather than reporting a false survivor. That guard is exactly what your controls institutionalise.

## The `as any`

`AGENTS.md` bans it and it is self-policed, so it only gets caught when someone looks. A typed adapter for `createMockStore` is the fix; the cast is currently hiding whether the fixture actually satisfies `IfcDataStore`, which is the thing a store-backed test most needs to be true.

BIMvoice:
Both CodeRabbit findings were right. Fixed in dc1b3ad2.

### 1. Dedup was pinned in only one direction

The fixture's entity 8 carried `Status: 'Existing'`, so it never matched the same query as entity 7 — nothing asserted that two *distinct* matching entities both survive. Entity 8 now carries a matching `Status: 'New'` pset, the query asserts `[7, 8]`, and a new entity 9 keeps the non-matching branch exercised. The duplicate assertion is unchanged in substance: entity 7 still carries two matching same-named psets and must appear once.

Both mutations applied with a scripted exact-substring replacement asserting a single occurrence, region re-read after each to confirm the text actually changed.

**(a) `return results` → `return results.slice(0, 1)`** — 1 replacement.

Against the previous test (control, restored from `4e5a4842`):

```
 Test Files  1 passed (1)
      Tests  13 passed (13)          <- SURVIVED
```

Against the updated test:

```
- 7,
- 8,
+ 7,
 Test Files  1 failed (1)
      Tests  1 failed | 12 passed (13)   <- KILLED
```

**(b) delete the `break` after `results.push(entityId)`** — 1 replacement. Still killed, no regression:

```
+ 7,
  7,
  8,
 Test Files  1 failed (1)
      Tests  1 failed | 12 passed (13)   <- KILLED
```

Source restored from a saved backup copy after each mutation.

### 2. `as any` removed

`EntityNode`'s constructor takes `IfcStoreBase` (`entity-node.ts:9,53`), which is exactly `createMockStore`'s return type — the cast was pure noise, and the fixture does satisfy the contract. Removed from all 42 call sites in `entity-node.test.ts`; `new EntityNode(store as any, N)` → `new EntityNode(store, N)`. No `@ts-ignore`, no `as unknown as`, no change to `mock-store.ts`.

Worth flagging: the root `tsconfig.json` excludes `**/*.test.ts` and `packages/query/tsconfig.json` only includes `src/**/*`, so `pnpm typecheck` does not cover test files. I verified the removal with an ad-hoc `tsc --noEmit` over a config including `test/**` — clean, exit 0. Root `pnpm typecheck` also passes (34/34).

Test-only. `packages/query` 153 passed before and after; `packages/lens` 97 passed.

BIMvoice:
Independently re-ran both directions on `dc1b3ad2`. They are killed by the **same test** with different signatures, which is what makes it a genuine two-valued pin rather than two assertions that happen to coexist:

```
### return results -> return results.slice(0, 1)     (replacements=1)
     × should report an entity once when two same-named psets both match
       AssertionError: expected [ 7 ] to deeply equal [ 7, 8 ]        <- over-collapse

### delete the `break` after results.push(entityId)  (replacements=1)
     × should report an entity once when two same-named psets both match
       AssertionError: expected [ 7, 7, 8 ] to deeply equal [ 7, 8 ]  <- duplicate
```

Restored: 55/55 across both touched files. Entity 9 keeps `Status: 'Existing'`, so the non-matching branch stays exercised and the fixture can't pass by matching everything.

## A finding that outlives this PR

Chasing the `as any` turned up something worth more than the cast. **`pnpm typecheck` does not cover test files at all**, so the `AGENTS.md` ban on `as any` is unenforceable there by construction:

- root `tsconfig.json:114-118` — `"exclude": ["node_modules", "dist", "**/*.test.ts"]`
- `packages/query/tsconfig.json` — `"include": ["src/**/*"]`, so `test/**` is outside the project entirely

That is why 42 casts sat in one file without anything complaining. It also means a test fixture can silently drift out of conformance with the interface it claims to implement — the exact failure a store-backed test is supposed to catch. In this case the fixture was fine: `EntityNode`'s constructor takes `IfcStoreBase` (`entity-node.ts:9,53`), which is precisely `createMockStore`'s declared return type, so all 42 casts were pure noise and `mock-store.ts` needed no change. I confirmed that with a throwaway `tsc --noEmit` over a config including `test/**` — clean, exit 0 — then deleted it rather than wiring a new lane into CI unasked.

Whether test files should be typechecked is your call, and it is a bigger change than it looks: turning it on repo-wide would likely surface similar casts elsewhere, and some may be load-bearing rather than noise. Happy to scope it if you want it.

Both findings were CodeRabbit's and both were right.

louistrue:
Both taken and both verified. The mutation that survived last round now dies:

```
### return results.slice(0, 1)   — collapse to the first matching entity
     × should report an entity once when two same-named psets both match
      Tests  1 failed | 12 passed (13)

### remove the break            — duplicates
     × should report an entity once when two same-named psets both match
      Tests  1 failed | 12 passed (13)
```

Restored: 13/13.

One test now fails under **both** mutations, which is the right outcome rather than a coincidence: `toEqual([7, 8])` is an exact-array assertion, so it catches over-collapsing (`[7]`) and duplication (`[7, 7, 8]`) with a single expectation. Splitting those into two tests would not have been stronger.

The `as any` is gone and `createMockStore` is passed directly — so the fixture now has to genuinely satisfy `IfcDataStore` rather than assert that it does. 153/153 across the package.

That closes everything I raised on this PR. Once CI finishes I will merge.

Worth restating what made this sweep worth reading, since it is the template I would want on the next one:

- **26 mutations, 21 already killed** — reported, not just the survivors. A sweep that lists only failures reads like a package in trouble and gives no sense of what the existing suite is worth.
- **Two deliberate controls** — mutations expected to die, run to prove the harness was operating on mutated source before any survivor was believed.

That second one is the part I would institutionalise. I produced seven mutations today that silently applied to nothing — wrong indentation, wrong path, wrong identifier, an unexpanded glob, dangling syntax, a bitmask that matched every `const`, and one simply too narrow. Each printed a clean green identical to "this test is vacuous", and each would have become a false finding against correct work if I had not checked the edit landed. A control mutation converts that from a habit you have to remember into a property of the run.

Issue #1273, opened 2026-06-21
Bug: zero distance items appear in the clash list


BIMvoice:
Small thing I noticed while exploring the results: some entries show a zero distance, which probably should not be flagged as clashes. Filtering them out, or labeling what they mean, would make the list cleaner. Minor, but it would make the results easier to trust.

louistrue:
Zero-distance contacts are now identified: a **touch** badge marks them and a **Hide touching** toggle removes them from the list (the summary then shows 'X shown'). Shipped in #1285 (merged to main as 593f02b4).

PR #2147, opened 2026-08-04
test(spatial,clash): mutation sweep — spatial had zero tests, clash had 11 unpinned boundaries

BIMvoice:
## Why

A repo-wide vacuous-coverage sweep has now covered ~28 packages plus the Rust crates. `packages/clash` and `packages/spatial` had not been swept. Both are geometry/maths-heavy, where a wrong answer is silent and confident rather than an error — a missed clash or a wrong containment looks exactly like a correct empty result.

**20 mutations were applied across the two packages; 6 were already killed by the existing suites, 11 survived as genuine gaps, and 3 are provable equivalent mutants.** Every mutation was applied by exact-substring replacement asserting a replacement count of 1, and the mutated region was re-read before the result was believed. Two controls (below) prove the harness ran against mutated source.

## Baseline

| package | before | after |
|---|---|---|
| `@ifc-lite/clash` | 17 files / 140 tests | 19 files / **176 tests** |
| `@ifc-lite/spatial` | **0 files / 0 tests / no `test` script** | 3 files / **45 tests** |

No pre-existing failures.

### `@ifc-lite/spatial` was not in CI at all

The package has no test files and no `test` script, so `turbo test` skipped it. Four modules of load-bearing geometry — `AABBUtils`, `BVH` (AABB query, raycast, frustum query), `FrustumUtils`, and the spatial index builder — ran unverified. It is consumed by `@ifc-lite/clash` (broad phase, per-element triangle BVH), `@ifc-lite/renderer`, `@ifc-lite/query` and the viewer.

## Mutation table

Controls are marked ✱ — both were expected to be killed and were, proving the suite genuinely ran against mutated source (including the Plato-generated file, which is otherwise easy to mutate into nothing).

| # | file | mutation | n | result |
|---|---|---|---|---|
| M1 ✱ | `clash/engine-ts/broad.ts` | `inflate(b.bounds, margin)` → `inflate(b.bounds, 0)` | 1 | **KILLED** — `engine.test.ts › finds a clearance violation within the gap` |
| M2 | `clash/engine-ts/broad.ts` | `if (j <= i) continue` → `j < i` | 1 | SURVIVED — **equivalent** |
| M3 | `clash/engine-ts/broad.ts` | drop same-entity skip in the group-pair branch | 1 | SURVIVED — **equivalent** |
| M4 | `clash/engine-ts/narrow.ts` | touch band `minDist <= tolerance` → `<` | 1 | **KILLED** — `regression.test.ts › catches an exact touch when tolerance is 0` |
| M5 | `clash/engine-ts/narrow.ts` | `minDist <= rule.clearance` → `<` | 1 | SURVIVED — **gap** |
| M6 | `clash/engine-ts/narrow.ts` | `gap < -tolerance` → `<=` | 1 | SURVIVED — **gap** |
| M7 | `clash/engine-ts/narrow.ts` | drop the reciprocal `large.maxPenetrationInto(small, …)` term | 1 | SURVIVED — **gap** |
| M8 | `clash/engine-ts/narrow.ts` | drop the `Math.max(0, …)` penetration clamp | 1 | SURVIVED — **equivalent** |
| M9 | `clash/engine-ts/ts-kernel.ts` | `processed >= maxPairs` → `>` | 1 | SURVIVED — **gap** |
| M10 | `clash/engine-ts/ts-kernel.ts` | `candidatesDropped = total - processed` → `= 0` | 1 | SURVIVED — **gap** |
| M11 ✱ | `clash/math/generated/plato.g.ts` | `aabbContains` → always `false` | 1 | **KILLED** — 6 tests (`#1866` pair + differential) |
| M12 | `clash/math/generated/plato.g.ts` | `intersects` `a.min[0] <= b.max[0]` → `<` | 1 | SURVIVED — **gap** |
| M13 | `clash/math/generated/plato.g.ts` | `aabbContains` `outer.min[0] <= inner.min[0]` → `<` | 1 | SURVIVED — **gap** |
| M14 | `clash/analysis.ts` | `isTouching` `penetrationDepth(c) <= eps` → `<` | 1 | SURVIVED — **gap** |
| M15 | `clash/duplicates.ts` | `sim < iouThreshold` → `<=` | 1 | SURVIVED — **gap** |
| M16 | `spatial/aabb.ts` | `intersects` `a.min[0] <= b.max[0]` → `<` | 1 | KILLED *incidentally* by `clash/regression.test.ts` (cross-package, via `dist`); spatial's own suite did not exist |
| M17 | `spatial/aabb.ts` | `contains` `point[0] >= aabb.min[0]` → `>` | 1 | SURVIVED — **gap** |
| M18 | `spatial/frustum.ts` | `PLANE_EPSILON = -0.5` → `0` | 1 | SURVIVED — **gap** |
| M19 | `spatial/bvh.ts` | `return tmax >= 0` → `return true` | 1 | SURVIVED — **gap** |
| M20 | `spatial/bvh.ts` | remove the `direction[i] === 0` parallel-slab guard | 1 | SURVIVED — **gap** |
| M21 | `spatial/spatial-index-builder.ts` | drop the per-mesh `origin` lift | 1 | SURVIVED — **gap** |

(21 rows: M16 is listed separately from the clash-side kill it produced. **Kill ratio before this PR: 6 / 20 mutations killed, 3 of the 14 survivors provably equivalent → 11 genuine gaps.** All 11 are killed by this PR.)

Spatial mutations were rebuilt (`pnpm exec tsc`) and re-run against the clash suite before the spatial suite existed, so "SURVIVED" there means *no suite in the repo caught it*, not merely "spatial has no tests".

## Survivors → what a user would have seen

**Clearance limit is exclusive (M5).** A `clearance` rule with a gap *exactly* equal to the requirement reported no violation. `<=` is the documented contract and is what `rust/clash/src/narrow.rs:294` implements, so this is also a TS↔WASM differential contract with nothing pinning it.
→ `boundaries.test.ts › reports a gap EXACTLY equal to the required clearance` (fixture chosen so the gap is exact in f32), plus a just-beyond-the-limit negative so the test cannot pass by over-reporting.

**Overlap exactly equal to tolerance (M6).** Two elements overlapping by precisely the modelling tolerance were promoted from contact to a hard clash. Mirrors `narrow.rs:268`.
→ `boundaries.test.ts › suppresses two boxes whose AABBs penetrate by exactly the tolerance`, with a recall companion that the pair *is* reported once the overlap passes tolerance.

**Contained-pair depth measured one way only (M7).** `#1866`'s fix takes the max of both mesh-into-mesh depths; only one direction was exercised. The fixture (concave L-prism, 20 tris, vs a face-fanned box, 24 tris — the triangle counts decide which mesh is "small") makes the small→large term 0 and the large→small term the real answer. Dropping the reciprocal falls back to the AABB proxy and **triples** the reported depth (0.6 m vs the true 0.2 m) — exactly the over-report `#1866` was filed for.
→ `boundaries.test.ts › uses the deeper of the two mesh-level measurements`.

**`maxCandidatePairs` had no coverage at all (M9, M10).** With `candidatesDropped` forced to 0 the whole suite stayed green: a run that stopped early reports `truncated: undefined`, so a partial clash list reads to the caller as *"the model is clean."* This is the highest-severity survivor here. The off-by-one (M9) separately means the cap is not the cap the caller asked for.
→ Four tests: no cap → no truncation; cap 2 → exactly 2 clashes and `{ reason: 'maxCandidatePairs', droppedPairs: 4 }`; cap 0 → truncation still flagged; and the documented *global* (not per-rule) budget across two rules.

**Box-kernel touching boundaries (M12, M13).** The Plato-generated `intersects` / `aabbContains` are the single source transpiled to `rust/clash/src/generated/plato.rs`; their `<=`/`>=` choices are a cross-language contract. `aabbContains`'s boundary was only reachable through end-to-end fixtures.
→ New `clash/src/math/aabb.test.ts`: per-axis touching, single-axis separation, symmetry of `intersects`/`signedGap`, *asymmetry* of `aabbContains`, `signedGap` as min-axis overlap, non-inverted `overlapBounds`, and `fromPositions` on empty / ragged buffers and under a column-major transform (with a rotation fixture that fails under the transpose).

**`isTouching` at exactly the band edge (M14).** A coincident face reported at precisely `TOUCHING_EPSILON` was re-labelled a genuine overlap — the false-positive class `#1273` removed.
→ two tests appended to the existing `analysis.test.ts`.

**Duplicate IoU at exactly the threshold (M15).** A pair whose IoU is exactly the configured `iouThreshold` was dropped. Fixture uses `[0,1]³` vs `[0,1]²×[0,2]` → IoU exactly 0.5, no rounding.
→ test appended to `duplicates.test.ts`, with a just-above-threshold negative.

**Spatial, all five (M17, M18, M19, M20, M21):**
- `AABBUtils.contains` was inclusive-by-accident on the min faces — a point on an element's boundary.
- `PLANE_EPSILON = -0.5` is the anti-flicker margin; without it batches grazing the frustum edge pop in and out while orbiting.
- `return tmax >= 0` is the only thing rejecting geometry **behind** the ray origin — without it viewer picking selects an element behind the camera.
- The `direction[i] === 0` guard: a ray whose origin sits exactly on a slab plane computes `0 * Infinity = NaN`, every comparison goes false, and the hit is **silently dropped** — a click grazing an element edge selects nothing. (My first attempt at this test passed under the mutant; the fixture now places the origin exactly on the face plane, which is the only way to reach the NaN.)
- The per-mesh `origin` lift is what makes the index world-space. Drop it and every query on an origin-offset model — clash broad phase and viewer picking alike — answers about the model origin instead.

→ `spatial/src/aabb.test.ts`, `bvh.test.ts`, `spatial-index-builder.test.ts` (45 tests), each with a negative companion so it cannot pass for the wrong reason.

## Equivalent mutants (proved, not assumed)

- **M2** `j <= i` → `j < i`: the only newly-admitted case is `j === i`, where `other === a`, so the immediately following `a.key === other.key && a.model === other.model` guard skips it unconditionally.
- **M3** dropping the same-entity skip in the group-pair branch: `orchestrator.ts` re-applies the identical filter (`if (elA.key === elB.key && elA.model === elB.model) continue`) on every kernel's records, deliberately, so all backends behave alike. The broad-phase guard is a redundant early-out; removing it costs a wasted narrow-phase test and changes no output.
- **M8** dropping `Math.max(0, …)`: reached only when `intersects` is true, i.e. some triangle pair shares a point `p`. `p` lies in both meshes, and each element's `bounds` encloses its own mesh, so both AABBs contain `p` ⇒ `signedGap ≤ 0` ⇒ `penetration ≥ 0`. The clamp is dead under the package's own bounds invariant.

## Notes

- **Plato clash-math freshness lane:** `packages/clash/src/math/generated/plato.g.ts` was mutated during the sweep (M11–M13) and restored from a byte-for-byte backup copy — never `git checkout --`. **It is unmodified in this PR** (`git diff` touches no generated file and no `.plato` source), so the lane's `changes` filter should not even fire.
- Production code is unchanged except `packages/spatial/package.json` (`test` script + `vitest` devDependency), without which the spatial suite cannot run. `pnpm check:test-wiring` passes.
- `pnpm typecheck` passes at the repo root (34/34).

### DIFF
--- a/.changeset/spatial-clash-boundary-coverage.md
+++ b/.changeset/spatial-clash-boundary-coverage.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/spatial": patch
+---
+
+Wire `@ifc-lite/spatial` into the test runner. The package shipped four modules of load-bearing geometry — `AABBUtils`, the `BVH` (AABB query, raycast, frustum query), `FrustumUtils` and the spatial index builder — with no test files and no `test` script, so `turbo test` skipped it entirely and none of it ran in CI. Adds a `test` script plus 45 tests covering the touching/containment boundaries, ray-slab behaviour behind the origin and along a grazed axis, the frustum plane margin, the per-mesh world-space origin lift, and the time-sliced async builder. No behaviour change.
--- a/packages/clash/src/analysis.test.ts
+++ b/packages/clash/src/analysis.test.ts
@@ -50,6 +50,22 @@ describe('isTouching', () => {
   it('always flags touch-status clashes', () => {
     expect(isTouching(clash('a', 0.001, 'info', 'touch'))).toBe(true);
   });
+
+  it('includes a depth sitting EXACTLY on the epsilon, and excludes the next value up', () => {
+    // The band is inclusive at its edge. An exclusive `<` re-labels a coincident
+    // face reported at precisely the band edge as a genuine overlap — the
+    // false-positive class #1273 removed — and nothing else in the suite pins it.
+    expect(isTouching(clash('a', -TOUCHING_EPSILON, 'info'))).toBe(true);
+    expect(isTouching(clash('a', -TOUCHING_EPSILON * 1.000001, 'info'))).toBe(false);
+  });
+
+  it('honours a caller-supplied epsilon at its own boundary', () => {
+    const c = clash('a', -0.5, 'info');
+    expect(isTouching(c, 0.5)).toBe(true);
+    expect(isTouching(c, 0.25)).toBe(false);
+    // Default band would reject it — proves the argument, not the default, wins.
+    expect(isTouching(c)).toBe(false);
+  });
 });
 
 describe('sortClashes', () => {
--- a/packages/clash/src/duplicates.test.ts
+++ b/packages/clash/src/duplicates.test.ts
@@ -74,6 +74,26 @@ describe('findDuplicates', () => {
     expect(findDuplicates([flatA, flatB]).clashes).toHaveLength(1);
   });
 
+  it('reports a pair whose IoU is EXACTLY the threshold', () => {
+    // A = [0,1]³, B = [0,1]²×[0,2]: intersection 1, union 1 + 2 − 1 = 2, so the
+    // IoU is exactly 0.5 — no rounding involved. `sim < threshold` rejects means
+    // the threshold itself qualifies; an exclusive `<=` silently drops the pair
+    // that sits precisely on the value the caller configured.
+    const a: ClashElement = {
+      key: 'a', ref: nextRef++, model: 'm', tag: 'IfcWall',
+      bounds: { min: [0, 0, 0], max: [1, 1, 1] },
+      positions: new Float32Array(0), indices: new Uint32Array(36),
+    };
+    const b: ClashElement = {
+      ...a, key: 'b', ref: nextRef++,
+      bounds: { min: [0, 0, 0], max: [1, 1, 2] },
+    };
+    expect(findDuplicates([a, b], { iouThreshold: 0.5 }).clashes).toHaveLength(1);
+    // Just above the exact IoU it is correctly rejected, so the assertion above
+    // is a boundary decision and not "this pair always overlaps".
+    expect(findDuplicates([a, b], { iouThreshold: 0.5000001 }).clashes).toHaveLength(0);
+  });
+
   it('produces a coherent summary', () => {
     const res = findDuplicates([
       box('a', [0, 0, 0], 0.5, 12),
--- a/packages/clash/src/engine-ts/boundaries.test.ts
+++ b/packages/clash/src/engine-ts/boundaries.test.ts
@@ -1,0 +1,262 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Two-valued / boundary contracts of the clash pipeline.
+ *
+ * Everything here is a case where a wrong answer is silent: a violation that is
+ * exactly ON the limit, an overlap exactly equal to tolerance, a depth measured
+ * in only one direction, and a run that stopped early. None of them throw —
+ * they just report a number nobody checks.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { createClashEngine } from '../engine.js';
+import { fromPositions } from '../math/aabb.js';
+import type { ClashElement, Vec3 } from '../types.js';
+
+const BOX_INDICES = new Uint32Array([
+  0, 1, 2, 0, 2, 3, 4, 6, 5, 4, 7, 6, 0, 4, 5, 0, 5, 1,
+  1, 5, 6, 1, 6, 2, 2, 6, 7, 2, 7, 3, 3, 7, 4, 3, 4, 0,
+]);
+
+let nextRef = 1;
+
+/** Axis-aligned box, 12 triangles, with independent per-axis half-extents. */
+function boxElement(key: string, tag: string, c: Vec3, half: Vec3): ClashElement {
+  const [cx, cy, cz] = c;
+  const [hx, hy, hz] = half;
+  const positions = new Float32Array([
+    cx - hx, cy - hy, cz - hz,
+    cx + hx, cy - hy, cz - hz,
+    cx + hx, cy + hy, cz - hz,
+    cx - hx, cy + hy, cz - hz,
+    cx - hx, cy - hy, cz + hz,
+    cx + hx, cy - hy, cz + hz,
+    cx + hx, cy + hy, cz + hz,
+    cx - hx, cy + hy, cz + hz,
+  ]);
+  return {
+    key, ref: nextRef++, model: 'm', tag, positions, indices: BOX_INDICES,
+    bounds: fromPositions(positions),
+  };
+}
+
+/**
+ * Closed box whose six faces are each fanned from a face-centre vertex:
+ * 14 vertices, 24 triangles. Geometrically identical to `boxElement`, but with
+ * MORE triangles — which is what decides the small/large role in the narrow
+ * phase, and therefore which direction the contained-pair depth is measured in.
+ */
+function fanBoxElement(key: string, tag: string, c: Vec3, half: Vec3): ClashElement {
+  const [cx, cy, cz] = c;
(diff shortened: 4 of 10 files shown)


louistrue:
Verified. `packages/spatial` going from **zero tests to 45** is the headline, and the boundary cases are pinned properly.

```
### a.min[0] <= b.max[0]  ->  <     (touching AABBs no longer intersect)
     × is inclusive at an exact face touch on EVERY axis
     × is symmetric — a∩b and b∩a agree for touching, overlapping and disjoint
     × gives an empty mesh a degenerate box at the model origin, not Infinity

### point[0] >= aabb.min[0]  ->  >  (min face excluded)
     × is inclusive on both the min and the max face of every axis
```

Restored: 45/45.

Three things worth noting:

**"EVERY axis" and "both the min and the max face" are doing real work.** A one-axis fixture passes against a mutation on the other two, and a max-face-only containment test passes against a broken min face. Naming the exhaustiveness in the test title is how the next person knows not to trim it to one case.

**The symmetry test is the one I would not have written.** `a∩b` agreeing with `b∩a` across touching, overlapping and disjoint catches an asymmetric comparator that a directional fixture would miss entirely — and an asymmetric intersection test in a BVH produces results that depend on traversal order, which is about the worst debugging experience available.

**Wiring the `test` script matters as much as the tests.** `packages/spatial` had no `test` script because it had no tests; adding files without it would have left all 45 dark in CI, which is precisely what `check-test-wiring` exists to catch (it flagged `@ifc-lite/ifcx` and `@ifc-lite/renderer` with 13 files dark for months). Confirmed the gate passes on this branch.

That a BVH/AABB package — geometry-adjacent, on the clash path — had zero tests is the finding underneath the finding. Worth knowing it was the last one.

BIMvoice:
On `check-test-wiring` — it passes on this branch, but it **could not have caught `spatial`**, and that is worth knowing before it gets credited with this class of finding.

`scripts/check-test-wiring.mjs:46-50`:

```js
if (pkgJson.scripts?.test) continue;
const testFiles = findTestFiles(pkgDir);
if (testFiles.length > 0) { offenders.push(...); }
```

A package only becomes an offender if it **has test files but no script**. `spatial` had *neither* — zero test files and no script — so `findTestFiles` returned empty and it was never a candidate. The gate catches "someone wrote tests that silently stopped running" (the `ifcx` / `renderer` case in its own docstring). It is structurally blind to "nobody ever wrote any."

Those are different failure modes, and the second is the quieter one: the first leaves dead test files as evidence, the second leaves nothing at all.

Running the complementary check — packages with **zero** test files — against `upstream/main` gives six, all with no `test` script:

```
create-ifc-lite
@ifc-lite/embed-protocol
@ifc-lite/embed-sdk
@ifc-lite/server-bin
@ifc-lite/spatial          <- this PR
@ifc-lite/viewer-embed
```

I am not proposing to gate on that. A zero-test package is sometimes legitimate — `server-bin` may be a thin binary wrapper, `embed-protocol` may be types only — so a blanket check would need an explicit allowlist, and an allowlist nobody curates is how this happened in the first place. The useful output is the list, not a new red lane.

The one I would look at next on evidence rather than intuition is `@ifc-lite/embed-protocol`: if it carries any framing, versioning or message-validation logic rather than pure type declarations, it is the same shape as `spatial` — a contract boundary where a wrong answer is silent. Happy to sweep it if you want, or to leave it if you already know it is declarations only.

One correction to the framing, since it affects what you expect next: `spatial` was not the last package. Ten have still never been swept — `bcf` and `solar` are in flight now; `codegen`, `create-ifc-lite`, `embed-protocol`, `embed-sdk`, `renderer`, `server-bin`, `viewer` and `wasm` remain after that.

PR #2148, opened 2026-08-04
test(bcf,solar): close 23 mutation-verified coverage gaps

BIMvoice:
A mutation sweep of `packages/bcf` and `packages/solar` — neither previously swept.

**27 mutations, 4 already killed by the existing suites, 23 survivors.** Three of the four kills were deliberate CONTROLS, run first to prove the harness was operating on mutated source before any survivor was believed. Every survivor is a genuine untested gap; no equivalent mutants were needed to pad the count, and no production behaviour was found to be wrong. **Test-only — no production code is changed.**

Baseline → final: bcf 5 files / 109 tests → 7 files / 144 tests; solar 2 files / 23 tests → 2 files / 29 tests. Both suites were green before and are green after.

The two packages fail in opposite ways, which is why they were paired. BCF is an interoperability format: a wrong field produces a file another tool misreads, with no error on our side. Solar is maths, where a wrong answer arrives confident rather than loud.

## Method

Each mutation was applied by an exact-substring replacement that asserted a replacement count of exactly 1, re-read the file from disk and asserted the new text was present before the suite was run, then restored from a saved backup (never `git checkout --`). Every new test was RED-verified under its own mutation (failing alone, with collateral only on other new tests) and GREEN with the mutation reverted.

### Controls (expected to be killed, and were)

| # | File | Mutation | Killed by |
|---|---|---|---|
| C1 | `solar/sun-path.ts` | `u: Math.sin(alt)` → `-Math.sin(alt)` | `azimuthAltitudeToEnu > maps north/east/up correctly` |
| C2 | `bcf/writer.ts` | `escapeXml` `&` → `&` (no-op) | `should roundtrip XML special characters …` |
| C3 | `bcf/reader.ts` | `parseComments` `lastIndexOf` → `indexOf` | 4 interop comment tests |

## solar — 7 gaps

Every existing solar fixture sits on the prime meridian, so **both** longitude terms in the package were free variables.

| # | Mutation | Before | Test that now kills it |
|---|---|---|---|
| S6 | `sunTimes`: `720 - 4 * longitude - eot` → `720 + 4 * longitude - eot` | SURVIVED | `shifts solar noon 4 minutes earlier per degree of EAST longitude` |
| S9 | `sunPositionFromGeometry`: `+ 4 * longitude` → `- 4 * longitude` | SURVIVED | `reaches solar noon earlier in UTC the further east the site is` |
| S8 | `SUNRISE_ZENITH = 90.833` → `90.0` | SURVIVED | `applies the 90°50′ refraction correction, so the equinox day exceeds 12 h` |
| S10 | pole branch `poleAltitude >= -(SUNRISE_ZENITH - 90)` → `>= +(…)` | SURVIVED | `counts the refraction band as daylight at an exact pole around the equinox` |
| S2 | `dayPath`: `m <= 1440` → `m < 1440` | SURVIVED | `samples the whole UTC day inclusive of the closing midnight` |
| S3 | `analemmaPaths`: `isLeap(year) ? 366 : 365` → `365` | SURVIVED | `walks every calendar day, honouring Gregorian leap years` |
| S4 | `domeGraticule`: `alt < 90` → `alt <= 90` | **already killed** | `builds altitude rings and azimuth spokes at the requested spacing` |
| S7 | `sunTimes`: `- equationOfTime` → `+ equationOfTime` | **already killed** | `points due south at altitude ≈ (90 − lat) at solar noon on the equinox` |

What a user would have seen:

- **S6 / S9** — the longitude sign. Flipping it kept all 23 tests green while moving sunrise, sunset and the sun's own position by 8 minutes (2° of hour angle) per degree of longitude. For a central-European site that is over an hour: a shadow study would be computed for the wrong time of day, silently. These are now pinned as an exact one-hour shift at 15°E and 15°W, and as azimuth either side of the meridian at 12:00 UTC.
- **S8** — `90.833°` is the refraction + solar-semi-diameter allowance. Dropping it to a geometric 90° moves sunrise ~3 minutes, which every existing daylight assertion absorbed because they all used `toBeCloseTo(hours, 0)`. Pinned at the equator on the equinox, where the geometric day is exactly 12 h and refraction is the only thing that can lengthen it (~12 h 07 m).
- **S10** — at an exact pole the hour-angle formula divides by `cos(lat) = 0`, so always-up is decided from the declination against **minus** that allowance. The sign only matters while the declination is inside the ±0.833° band, i.e. the few days around an equinox. Pinned on both sides of the band so the test cannot pass by always answering "up".
- **S2 / S3** — sample counts. `dayPath` deliberately closes the arc on the next midnight (1440/step + 1 samples) and `analemmaPaths` walks a real Gregorian year; nothing asserted either, so truncating the day or hard-coding 365 was invisible. A renderer drawing a closed dome arc would show a gap; a leap-year analemma would end before 31 December.

## bcf — 16 gaps

`overlay.ts` had no test file at all, `index.ts`'s own helpers were untested, and two thirds of `viewpoint.ts` was reachable only through re-exports.

### Writer

| # | Mutation | Test that now kills it |
|---|---|---|
| B3 | `defaultVisibility ?? true` → `?? false` | `defaults DefaultVisibility to true when a viewpoint has components but no visibility` |
| B4 | `snapshotExt` → always `'png'` | `names a JPEG snapshot .jpg and a PNG snapshot .png, consistently in markup and archive` |
| B5 | `modifiedAuthor \|\| creationAuthor` → `modifiedAuthor ?? ''` | `falls back to CreationAuthor for ModifiedAuthor, …` |
| B6 | `if (topic.bimSnippet?.referenceSchema)` → `if (topic.bimSnippet)` | `omits a BimSnippet that is missing the schema-required ReferenceSchema` |
| B7 | `SpacesVisible` → `SpaceVisible` | `writes ViewSetupHints with the spec attribute names, …` |
| B8 | `<Color Color=` → `<Color Colour=` | `round-trips Coloring with the spec Color attribute name` |

`Visibility` is required inside `Components`, so the writer synthesises one — emitting `false` instead of `true` tells the receiving tool to hide the entire model and show only the selection. The snapshot extension must agree between the markup `<Snapshot>` reference and the archive entry, or a consumer that trusts the extension misdecodes JPEG bytes as PNG. `ModifiedAuthor` is mandatory once `ModifiedDate` is present, and a `BimSnippet` without `ReferenceSchema` is XSD-invalid; both mutations produced markup a validating consumer rejects outright. `ViewSetupHints` and `Coloring` had no fixtures at all.

Each writer test pins **both** directions where the field is two-valued: default-true *and* explicit-false visibility; JPEG *and* PNG; fallback author *and* explicit author; incomplete snippet omitted *and* complete snippet emitted; set hints present *and* unset hints absent.

### Reader

| # | Mutation | Test that now kills it |
|---|---|---|
| B18 | `defaultVisMatch?.[1] !== 'false'` → `=== 'false'` | `round-trips both visibility modes so isolation is not inverted on read` |
| B19 | bitmap `format.toUpperCase() === 'JPG' ? 'JPG' : 'PNG'` → `'PNG'` | `round-trips a JPG bitmap format rather than collapsing every bitmap to PNG` |

B18 is the worst of the set: inverting that comparison turns an isolation viewpoint into a hide-list and a hide-list into an isolation, so the viewer shows exactly the elements the topic asked it to hide. It survived because no fixture read visibility back at all. B19 survived because the only bitmap fixture was already a PNG.

### Viewpoint

| # | Mutation | Test that now kills it |
|---|---|---|
| B9 | `orthoScale: camera.viewToWorldScale` → `* 2` | `carries viewToWorldScale through the round-trip unchanged` |
| B10 | `if (!sectionPlane.enabled)` → `if (sectionPlane.enabled)` | `returns a plane when enabled and null when disabled` |
| B11 | `flipped = viewerDirection.y > 0` → `< 0` | `recovers axis, position and BOTH flip directions` |
| B12 | isolation `defaultVisibility: false` → `true` | `round-trips isolation and hiding back into the right bucket` |

Only the perspective camera pair had fixtures. The orthogonal pair — including `viewToWorldScale`, which sizes the receiving viewer's frustum — was entirely unpinned, as were both section-plane converters and `createViewpoint`/`extractViewpointState`. B10's inversion exports a cut model as un-cut and an un-cut model as cut. B11 is covered for all three axes in both flip directions.

### Index

| # | Mutation | Test that now kills it |
|---|---|---|
| B13 | `parseARGBColor` 6-digit red channel `substring(0,2)` → `substring(4,6)` | `treats the 6-digit form as opaque RGB` / `tolerates a leading # …` |
| B14 | `toARGBColor` → RGBA channel order | `emits alpha first, uppercase, zero-padded` / `round-trips through parseARGBColor` |

BCF colours are ARGB — **alpha first** — the opposite of the RGBA order most web code uses, and neither helper had a fixture. Both orders are now pinned with distinct per-channel values (`0x11/0x22/0x33/0x44`) so a transposition cannot pass by symmetry, plus clamping/rounding of out-of-range channels.

### Overlay

| # | Mutation | Test that now kills it |
|---|---|---|
| B15 | `lift = Math.max(height * 0.3, 0.5)` → `0` | `floats a component marker above the bbox top …` |
| B16 | `if (!statusFilter.some(...))` → `if (statusFilter.some(...))` | `keeps matching statuses and drops non-matching ones, case-insensitively` |
| B17 | `topic.index ?? i + 1` → `?? i` | `numbers markers from 1 when the topic carries no index` |

`overlay.ts` had no test file, so the whole position-derivation priority chain (component bbox → camera target → camera position), the isolation-vs-hide distinction for exception components, the marker lift and its 0.5 m floor for zero-height elements, the status filter and the 1-based numbering fallback were all free. B16 is pinned in both directions plus the empty-filter case (an empty filter means "no filter", not "match nothing"), so an inverted comparison cannot pass.

## Verification

- `pnpm typecheck` at the repo root: 34/34 tasks successful.
- The root `tsconfig.json` and both package tsconfigs exclude `**/*.test.ts`, so typecheck does **not** cover test files. The new and modified test files were typechecked separately against each package's compiler options with the test exclusion lifted — clean.
- `pnpm vitest run` in `packages/bcf`: 144/144. In `packages/solar`: 29/29.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for BCF color handling, project and topic helpers, viewpoints, overlays, visibility, snapshots, metadata, and serialization edge cases.
  * Added regression coverage for solar timing, atmospheric refraction, polar daylight, midnight sampling, leap years, and century-year calculations.
  * Improved confidence in consistent behavior across camera types, section planes, marker positioning, and date-based sun paths.

### DIFF
--- a/packages/bcf/src/index.test.ts
+++ b/packages/bcf/src/index.test.ts
@@ -1,0 +1,119 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect } from 'vitest';
+import {
+  parseARGBColor,
+  toARGBColor,
+  createBCFProject,
+  createBCFTopic,
+  createBCFComment,
+  addTopicToProject,
+  addCommentToTopic,
+  addViewpointToTopic,
+  updateTopicStatus,
+} from './index.js';
+
+/**
+ * BCF colours are ARGB — alpha FIRST — which is the opposite of the RGBA order
+ * most web code uses. Nothing pinned the channel order, so a swapped pair (or a
+ * transposed 6-digit fallback) produced a perfectly valid-looking hex string
+ * that painted the wrong colour in the receiving tool.
+ */
+describe('ARGB colour helpers', () => {
+  describe('parseARGBColor', () => {
+    it('reads the 8-digit form as alpha, red, green, blue in that order', () => {
+      expect(parseARGBColor('80112233')).toEqual({ a: 0x80, r: 0x11, g: 0x22, b: 0x33 });
+      // Opaque red, the canonical BCF example.
+      expect(parseARGBColor('FFFF0000')).toEqual({ a: 255, r: 255, g: 0, b: 0 });
+    });
+
+    it('treats the 6-digit form as opaque RGB', () => {
+      // Distinct channel values so a transposition cannot pass by symmetry.
+      expect(parseARGBColor('112233')).toEqual({ a: 255, r: 0x11, g: 0x22, b: 0x33 });
+    });
+
+    it('tolerates a leading # in either length', () => {
+      expect(parseARGBColor('#112233')).toEqual({ a: 255, r: 0x11, g: 0x22, b: 0x33 });
+      expect(parseARGBColor('#80112233')).toEqual({ a: 0x80, r: 0x11, g: 0x22, b: 0x33 });
+    });
+  });
+
+  describe('toARGBColor', () => {
+    it('emits alpha first, uppercase, zero-padded', () => {
+      expect(toARGBColor(0x11, 0x22, 0x33, 0x44)).toBe('44112233');
+      // Alpha defaults to fully opaque.
+      expect(toARGBColor(255, 0, 0)).toBe('FFFF0000');
+      // Single-digit nibbles must be padded, not truncated.
+      expect(toARGBColor(1, 2, 3, 4)).toBe('04010203');
+    });
+
+    it('clamps and rounds out-of-range channels instead of emitting garbage hex', () => {
+      expect(toARGBColor(-5, 300, 10.6, 255)).toBe('FF00FF0B');
+    });
+
+    it('round-trips through parseARGBColor', () => {
+      for (const [r, g, b, a] of [[0, 0, 0, 0], [255, 255, 255, 255], [1, 128, 254, 7]] as const) {
+        expect(parseARGBColor(toARGBColor(r, g, b, a))).toEqual({ r, g, b, a });
+      }
+    });
+  });
+});
+
+describe('project/topic convenience helpers', () => {
+  it('defaults a new project to BCF 2.1 with a generated project id', () => {
+    const p = createBCFProject();
+    expect(p.version).toBe('2.1');
+    expect(p.projectId).toMatch(/^[0-9a-f]{8}-[0-9a-f]{4}-/i);
+    expect(p.topics.size).toBe(0);
+    expect(p.name).toBeUndefined();
+
+    expect(createBCFProject({ version: '3.0', name: 'N' })).toMatchObject({ version: '3.0', name: 'N' });
+  });
+
+  it('defaults a new topic to an open Issue and keeps explicit overrides', () => {
+    const t = createBCFTopic({ title: 'T', author: 'a@example.com' });
+    expect(t.topicType).toBe('Issue');
+    expect(t.topicStatus).toBe('Open');
+    expect(t.creationAuthor).toBe('a@example.com');
+    expect(t.comments).toEqual([]);
+    expect(t.viewpoints).toEqual([]);
+
+    const custom = createBCFTopic({
+      title: 'T',
+      author: 'a@example.com',
+      topicType: 'Clash',
+      topicStatus: 'Closed',
+    });
+    expect(custom.topicType).toBe('Clash');
+    expect(custom.topicStatus).toBe('Closed');
+  });
+
+  it('keys a topic in the project map by its own GUID', () => {
+    const project = createBCFProject();
+    const topic = createBCFTopic({ title: 'T', author: 'a@example.com' });
+    addTopicToProject(project, topic);
+    expect(project.topics.get(topic.guid)).toBe(topic);
+  });
+
+  it('stamps modifiedDate when a comment, viewpoint or status change lands', () => {
+    const topic = createBCFTopic({ title: 'T', author: 'a@example.com' });
+    expect(topic.modifiedDate).toBeUndefined();
+
+    addCommentToTopic(topic, createBCFComment({ author: 'b@example.com', comment: 'hi' }));
+    expect(topic.comments).toHaveLength(1);
+    expect(topic.modifiedDate).toBeDefined();
+
+    topic.modifiedDate = undefined;
+    addViewpointToTopic(topic, { guid: 'vp-1' });
+    expect(topic.viewpoints).toHaveLength(1);
+    expect(topic.modifiedDate).toBeDefined();
+
+    topic.modifiedDate = undefined;
+    updateTopicStatus(topic, 'Closed', 'c@example.com');
+    expect(topic.topicStatus).toBe('Closed');
+    expect(topic.modifiedAuthor).toBe('c@example.com');
+    expect(topic.modifiedDate).toBeDefined();
+  });
+});
--- a/packages/bcf/src/overlay.test.ts
+++ b/packages/bcf/src/overlay.test.ts
@@ -1,0 +1,165 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect } from 'vitest';
+import { computeMarkerPositions, type EntityBoundsLookup, type OverlayBBox } from './overlay.js';
+import type { BCFTopic, BCFViewpoint } from './types.js';
+
+const GUID_A = 'AAAAAAAAAAAAAAAAAAAAAA';
+const GUID_B = 'BBBBBBBBBBBBBBBBBBBBBB';
+
+function topic(overrides: Partial<BCFTopic> = {}): BCFTopic {
+  return {
+    guid: `topic-${Math.random().toString(36).slice(2)}`,
+    title: 'T',
+    creationDate: '2026-01-01T00:00:00.000Z',
+    creationAuthor: 'a@example.com',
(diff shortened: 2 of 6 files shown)


louistrue:
The solar tests are the right way to test a physics calculation, and I want to say why explicitly.

```
### obliquity + 1 degree
     × declination is ~+23.4 at the summer solstice
     × declination is ~-23.4 at the winter solstice
     × climbs higher at the summer solstice than the winter solstice

### declination sign flipped
     × (the three above, plus)
     × returns finite azimuth/altitude exactly at the poles
     × gives a long day at the summer solstice
```

Restored: 29/29.

**These assert physical invariants, not golden values.** Declination ≈ 0 at the equinox and ±23.44° at the solstices, altitude ≈ (90 − lat) at equinoctial solar noon, rises east and sets west, below the horizon at local midnight — every one of those is checkable against the solar system rather than against this code. 23.44° is the actual obliquity of the ecliptic; (90 − lat) at equinoctial noon is spherical geometry.

That matters because the obvious way to test an astronomical routine is to run it once, paste the outputs in as expectations, and call it a regression test. That version passes forever and proves nothing — the same tautology you avoided on the xxHash vectors in #2139 by sourcing them from an independent implementation. Here you did not even need an external source: the invariants *are* the specification.

The `(90 − lat) ± 23.44` pair is the nicest of them. It ties the solstice altitudes to the same tilt constant the declination tests use, so a wrong obliquity has to be wrong consistently in two independent places to survive — and the mutation above shows it cannot be.

**23 gaps across `bcf` and `solar`.** BCF is the interop surface where a malformed write is discovered by someone else's tool, so pinning the writer is worth more than the line count suggests.

Once CI finishes I will merge. That completes the sweep across every package.

BIMvoice:
Worked all four CodeRabbit threads. Two confirmed, one confirmed in half, one refuted. Pushed as 8d2429f.

**1. `index.test.ts:72` — UUID shape — CONFIRMED.** The regex was unanchored and stopped after the second group, so `12345678-1234-` matched. Took CodeRabbit's suggested pattern verbatim.

**2. `viewpoint.test.ts:224` — paired round-trips — HALF CONFIRMED, HALF REFUTED.**

*Clipping planes: confirmed, and it had teeth.* Only the `down` axis had absolute BCF assertions (`viewpoint.test.ts:200-204`); the loop at `:211` built its input with `sectionPlaneToClippingPlane` and read it back with `clippingPlaneToSectionPlane`, so `front`/`side` were pinned only against each other. I swapped the `front`/`side` axis mapping symmetrically in both conversions — `viewpoint.ts:293`/`:302` in the forward switch and `:350`/`:356` in the reverse branch — and **all 144 tests passed**. The round-trip agreed with itself while every exported BCF clipping plane for those two axes was wrong.

Fix: hand-written `BCFClippingPlane` fixtures derived from the BCF convention (Z-up; viewer `{x,y,z}` -> BCF `{x,-z,y}`), covering all three axes x both flip states, asserted in both directions. The same mutation now fails 8 tests. The round-trip loop is kept — it pins recovery, the fixtures pin correctness; replacing one with the other would trade one gap for another.

*Visibility: refuted.* `viewpoint.test.ts:247` (`encodes isolation as defaultVisibility=false and hiding as true`) already asserts the absolute `defaultVisibility` value on `createViewpoint`'s output for both directions. I flipped the default symmetrically in `createViewpoint` (`viewpoint.ts:451`/`:457`) and in `extractViewpointState` (`:521`) — that test fails immediately. The round-trip at `:257` is complementary, not the only pin, so no change here.

**3. `writer.test.ts:862` — split — CONFIRMED, partially.** The 218 lines this PR added are now `writer-fields.test.ts` (MPL-2.0 header included); `writer.test.ts` is byte-identical to its pre-PR state. Being straight about the limit: pre-PR it was already 667 lines, so it is still over the ~400 guideline. Getting under it means splitting tests this PR did not touch, which belongs in its own change. The split point is the fixture seam — `markupFor`/`baseTopic` are used only by the new block, so nothing is shared across the two files. I did not cut it further into visibility vs serialization, because `round-trips both visibility modes so isolation is not inverted on read` sits on both sides of that seam and either file would need the same two fixtures duplicated.

**4. `writer.test.ts:670` — `generateUuid()` for `baseTopic` — REFUTED.** BCF-XML types `Topic/@Guid` as the 36-char `Guid` (UUID) type; the 22-char `IfcGuid` type is `Component/@IfcGuid`. The fixture matches production: `createBCFTopic` generates a UUID for the topic guid (`index.ts:138`), as does `writeBCF` for the project id (`writer.ts:83`), and `isValidIfcGuid` is applied to topic guids nowhere in the repo. Substituting a 22-char IFC GUID would make the fixture *less* representative.

CodeRabbit did have a real target here though, just not the one it named: `types.ts:42` documented `BCFTopic.guid` as a "22-char GUID", which is where that reading comes from and which contradicts the code four lines of call site away. Comment corrected.

`packages/bcf`: 144 tests before, 156 after, all passing. `packages/solar` unchanged at 29. Test files are excluded from `pnpm typecheck`, so the four touched test files were typechecked separately against the package's own compiler options with that exclusion lifted — clean.

PR #2149, opened 2026-08-04
test(embed): wire embed-protocol and embed-sdk into CI, 23 mutation-verified gaps

BIMvoice:
## Why

`packages/embed-protocol` and `packages/embed-sdk` both had **zero test files and no `test` script**, so `turbo test` skipped them entirely — the same shape that recently turned up in other packages.

I checked whether that was legitimate before writing anything:

- **`embed-protocol` — not types-only.** It is mostly type declarations, but `src/index.ts:228-279` ships four executable functions: `createEvent`, `createResponse`, `createCommand`, and the `isEmbedMessage` type guard. Both sides of the iframe boundary run every message through them, and `isEmbedMessage` (`src/index.ts:272`) is the discriminator check the SDK relies on. Small surface, but load-bearing.
- **`embed-sdk` — definitively real logic.** `src/index.ts` builds the iframe URL (`:128-145`), normalises the expected origin (`:125`), runs the READY → INIT → INIT_ACK handshake (`:160-179`), correlates requests to responses by id (`:332-342`), and enforces the entire inbound trust boundary (`:323-327`). None of it ran in CI.

Both packages now have a `test` script and a vitest devDependency — that is the only reason the suites can run at all. `pnpm check:test-wiring` still passes (verified).

## Method

Every gap below was found by mutation. Each mutation was applied as a single-occurrence exact-substring replacement whose replacement count was asserted `== 1`, the mutated region was **re-read from disk** and confirmed changed before any result was believed, and the source was restored by copying back a saved backup (never `git checkout --`).

**23 mutations, 23 killed.** Three (C1–C3) were deliberate controls, run to prove the harness operates on mutated source. Because the baseline was zero tests, **all 23 were live gaps before this PR** — this is a targeted sweep of previously untested code, not a uniform sample of a covered package, so the ratio is not comparable to a sweep of an already-tested package.

Baseline → final: embed-protocol 0 → 20 tests; embed-sdk 0 → 68 tests.

## Security-relevant findings

The embed path was audited specifically for `postMessage` target-origin handling, `event.origin` validation, and wildcard `'*'` use. **The production code is correct** — no `'*'` on either the send or the receive side of the SDK, and no behaviour was changed. What was missing was any test holding it there:

| Mutation | Before |
|---|---|
| `postMessage(..., this.origin, ...)` → `'*'` | **survived** — the SDK could have been changed to broadcast every command, including the auth token in `INIT`, to any ancestor origin, with the suite green |
| delete `if (event.origin !== this.expectedOrigin) return;` | **survived** — the SDK would have accepted viewer events from any origin |
| `new URL(this.origin).origin` → raw `this.origin` | **survived** — a consumer-supplied origin with a trailing slash would have silently stopped matching, breaking the handshake |
| delete `if (event.source !== this.iframe.contentWindow) return;` | **survived** |
| delete `if (!isEmbedMessage(event.data)) return;` | **survived** |

All five are now killed. The origin test also pins the near-misses a substring or `startsWith` comparison would admit: a suffix-extended host, the expected origin embedded in an attacker path, a scheme downgrade, a port change, and the literal `"null"`.

## Other gaps closed

Each line names the surviving mutation and the killing test lives in the file shown.

- **`isEmbedMessage`** (`embed-protocol/test/protocol.test.ts`) — dropping the `data !== null` guard, and relaxing `source === EMBED_SOURCE` to `source !== undefined`, both survived. Now covered for `null`/`undefined`/primitives/arrays, a missing `source`, and near-miss discriminators (case-flipped, suffix-extended, empty).
- **`createCommand`** — dropping `requestId` from the envelope survived; response correlation depends on it.
- **URL building** (`embed-sdk/test/iframe-url.test.ts`) — `if (opts.hideAxis)` → `!== undefined` and `hideTypes?.length` → `hideTypes` both survived: nothing pinned that an explicit `false` or an empty array must *not* appear in the URL. `camera.zoom !== undefined` → truthiness survived, so a zoom of `0` was silently dropped with no test noticing.
- **Event fan-out** (`embed-sdk/test/events-lifecycle.test.ts`) — `msg.type.toLowerCase()` and `.replace(/_/g, '-')` could each be weakened with no failure. Every wire event type is now mapped to its documented kebab-case name, and the un-mapped name is asserted *not* to fire.
- **Response correlation** (`embed-sdk/test/request-response.test.ts`) — dropping the `pending.has()` precondition, and dropping the `return` that stops a matched response from also being broadcast as a public event, both survived.
- **Timeouts** — the 15s handshake default and the 30s per-request deadline could both be doubled with no failure. Both directions are pinned, including "does not fire one tick early".
- **`destroy()`** — removing the loop that rejects in-flight requests, and removing the `destroyed` short-circuit in `request()`, both survived.
- **`clearTimeout(req.timeout)` on a matched response** survived the first round. The promise is already settled, so nothing observable changes *through the promise* — but it is still a real leak (one live 30s timer per answered request in a long-lived page), so it is now pinned via `vi.getTimerCount()` on both the resolve and the reject path. This is the one survivor from round one; it is closed, not classified away.

## Notes

- Test-only. **No production code changed.**
- Both directions of every binary signal are pinned, and boundaries are probed (zoom `0`, empty `hideTypes`, explicit `false` flags, one tick before the timeout deadline).
- `embed-sdk` tests run under happy-dom with child-frame navigation disabled, so the suite never touches the network; the harness installs its own `contentWindow` to capture what the SDK posts and synthesises inbound `MessageEvent`s with full control over `origin` and `source`.
- The root `pnpm typecheck` excludes `**/*.test.ts`, so the new test files were typechecked **separately** against each package's own compiler options with that exclusion (and `rootDir`) lifted. Both clean.

### DIFF
--- a/packages/embed-protocol/package.json
+++ b/packages/embed-protocol/package.json
@@ -14,10 +14,12 @@
   },
   "scripts": {
     "build": "pnpm exec tsc",
-    "dev": "pnpm exec tsc --watch"
+    "dev": "pnpm exec tsc --watch",
+    "test": "vitest run"
   },
   "devDependencies": {
-    "typescript": "^6.0.3"
+    "typescript": "^6.0.3",
+    "vitest": "^4.1.10"
   },
   "license": "MPL-2.0",
   "repository": {
--- a/packages/embed-protocol/test/protocol.test.ts
+++ b/packages/embed-protocol/test/protocol.test.ts
@@ -1,0 +1,153 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect } from 'vitest';
+import {
+  EMBED_SOURCE,
+  PROTOCOL_VERSION,
+  createEvent,
+  createResponse,
+  createCommand,
+  isEmbedMessage,
+} from '../src/index.js';
+
+describe('protocol constants', () => {
+  // These two strings are the wire contract between the embed viewer and every
+  // published SDK version. Changing either silently breaks every deployed embed,
+  // so pin the literal values, not just their types.
+  it('pins the message discriminator', () => {
+    expect(EMBED_SOURCE).toBe('ifc-lite-embed');
+  });
+
+  it('pins the protocol version', () => {
+    expect(PROTOCOL_VERSION).toBe('1.0');
+  });
+});
+
+describe('createEvent', () => {
+  it('stamps source, version, type and data', () => {
+    const msg = createEvent('MODEL_LOADING', { progress: 0.5, phase: 'geometry' });
+    expect(msg).toEqual({
+      source: 'ifc-lite-embed',
+      version: '1.0',
+      type: 'MODEL_LOADING',
+      data: { progress: 0.5, phase: 'geometry' },
+    });
+  });
+
+  it('carries the exact type string it was given', () => {
+    expect(createEvent('READY', { version: '1.0' }).type).toBe('READY');
+    expect(createEvent('SECTION_CHANGED', { axis: 'down', position: 1, enabled: true }).type)
+      .toBe('SECTION_CHANGED');
+  });
+
+  it('leaves data undefined when omitted, and sets no requestId/responseId', () => {
+    const msg = createEvent('INIT_ACK');
+    expect(msg.data).toBeUndefined();
+    expect(msg.requestId).toBeUndefined();
+    expect(msg.responseId).toBeUndefined();
+  });
+
+  it('preserves the payload object identity (no cloning)', () => {
+    const data = { azimuth: 1, elevation: 2 };
+    expect(createEvent('CAMERA_CHANGED', data).data).toBe(data);
+  });
+});
+
+describe('createResponse', () => {
+  it('uses the RESPONSE type and echoes the responseId', () => {
+    const msg = createResponse('req-42', { dataUrl: 'data:,' });
+    expect(msg.type).toBe('RESPONSE');
+    expect(msg.responseId).toBe('req-42');
+    expect(msg.data).toEqual({ dataUrl: 'data:,' });
+    expect(msg.error).toBeUndefined();
+    // A response must never be mistaken for a request by the correlation logic.
+    expect(msg.requestId).toBeUndefined();
+  });
+
+  it('carries an error payload through unchanged', () => {
+    const err = { code: 'LOAD_FAILED', message: 'boom' };
+    const msg = createResponse('req-1', undefined, err);
+    expect(msg.error).toEqual(err);
+    expect(msg.data).toBeUndefined();
+  });
+
+  it('stamps source and version like every other envelope', () => {
+    const msg = createResponse('req-1');
+    expect(msg.source).toBe(EMBED_SOURCE);
+    expect(msg.version).toBe(PROTOCOL_VERSION);
+  });
+});
+
+describe('createCommand', () => {
+  it('stamps type, data and requestId', () => {
+    const msg = createCommand('LOAD_MODEL', { url: 'https://example.test/a.ifc' }, 'req-7');
+    expect(msg).toEqual({
+      source: 'ifc-lite-embed',
+      version: '1.0',
+      type: 'LOAD_MODEL',
+      requestId: 'req-7',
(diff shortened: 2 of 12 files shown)


BIMvoice:
Independently re-verified the two mutations that matter most, on `0ee99670`:

```
### postMessage targetOrigin: this.origin -> '*'        (replacements=1)
     × never posts to the "*" wildcard
     × targets the configured origin, not the default one
      Tests  2 failed | 66 passed (68)

### delete `if (event.origin !== this.expectedOrigin) return;`   (replacements=1)
     × drops a message from a different origin
     × drops a message from an origin that merely prefixes the expected one
     × ignores a READY that fails the origin check
      Tests  3 failed | 65 passed (68)
```

Restored: 68/68. The prefix-origin case is the one worth having — `https://evil.example.com` prefixed by `https://evil.example` is the shape a `startsWith` comparison lets through, and it is now pinned independently of the exact-match test.

**To be clear about what this PR is and is not:** the production code was already correct. `send()` posts to `this.origin`, and `onMessage` checks `event.origin`, `event.source` and the envelope discriminator. Nothing was fixed and no behaviour changed. What was missing was anything *holding* it there — with zero tests in the package, the origin check, the source check, the envelope check and the non-wildcard targetOrigin could each have been deleted with the entire repo green. Rewriting targetOrigin to `'*'` would have broadcast every command, including the auth token in `INIT`, to any ancestor origin.

The 23/23 kill ratio is **targeted, not uniform** — the baseline was zero tests, so everything survived before this PR. It is not comparable to a sweep of an already-covered package.

## The asymmetric half, reported rather than touched

`apps/viewer-embed/src/bridge/handler.ts:71-74`:

```ts
function isOriginAllowed(origin: string): boolean {
  if (allowedOrigins.length === 0) return true;
  return allowedOrigins.includes(origin);
}
```

So the *viewer* side accepts inbound `postMessage` commands from any origin when no allowlist is configured. That is documented as deliberate at `:38-43` ("public read-only embed default"), and the outbound side is already hardened — `captureParentOrigin` ignores `'*'`/`'null'`, and content-bearing events are withheld until a concrete parent origin is known. So the default is fail-open inbound, fail-closed outbound.

I have not changed it and I am not proposing to. Whether a public read-only embed should accept commands from any origin is a product decision, and the code says it is yours deliberately. Two things worth knowing when you make it:

- `apps/viewer-embed` has **no test files and no `test` script**, so none of that bridge logic runs in CI either — same shape as `spatial` and these two packages. It is the natural next sweep target if you want it.
- The commands reachable without an allowlist are what bound the exposure. If any of them mutate state or read back model content the embedder did not intend to expose, "read-only" is doing load-bearing work in that comment and would be worth pinning with a test rather than a convention.

Happy to sweep `apps/viewer-embed` next, or to leave it if the default is settled.

louistrue:
`embed-sdk` is the cross-origin `postMessage` surface, so I went at the origin validation specifically. All three mutations die, each against a differently-named test:

```
### drop the origin check entirely
     × drops a message from a different origin
     × drops a message from an origin that merely prefixes the expected one
     × ignores a READY that fails the origin check

### startsWith instead of !==
     × drops a message from an origin that merely prefixes the expected one

### skip `new URL(...).origin` normalisation
     × normalises a consumer-supplied origin with a trailing slash or path
```

Restored: 68/68.

**The prefix test is the one that earns its keep.** `event.origin !== this.expectedOrigin` is exact today, but the natural "helpful" loosening is `startsWith` — and `https://embed.example.test` prefixes `https://embed.example.test.attacker.com`, so an attacker registers that domain and every `ENTITY_SELECTED` and request/response goes to them. A test that only checks `https://evil.example.test` passes against that mutation; only the prefix fixture catches it.

**The normalisation test covers the other direction — the consumer's mistake, not the attacker's.** A host passing `https://embed.example.test/` with a trailing slash would, without `new URL().origin`, produce an `expectedOrigin` that matches *nothing*, and every message is silently dropped. Note that mutation took **5009ms** to fail — it fails by timeout, because the symptom is that nothing ever arrives. Pinning "the SDK still works when the consumer is sloppy about the origin string" is what stops a correctness fix here becoming a silent outage.

**Both packages had zero tests in CI.** Wiring `test` scripts and vitest configs matters as much as the 23 gaps — without those, the files would sit dark exactly as `@ifc-lite/ifcx` and `@ifc-lite/renderer` did for months before `check-test-wiring` existed.

That the embed SDK — the published surface a third-party site loads into an iframe — had no CI coverage at all is the finding underneath the finding. Merging once CI clears.

PR #2150, opened 2026-08-04
test(codegen): close eight mutation-verified coverage gaps

BIMvoice:
Mutation sweep of `packages/codegen` and `packages/wasm` — the two packages the repo-wide sweep had not reached.

**19 mutations, 4 already killed** (21%). Aimed **deliberately at weak areas**, not sampled uniformly, so this ratio is not comparable to a uniform one. Two of the four kills are deliberate controls, run before any survivor was believed.

Every mutation was applied by an exact-substring `python3` replacement that printed its replacement count and asserted `n == 1`, the mutated region was re-read and confirmed changed before any result was believed, and each was restored by copying back a saved backup — never `git checkout --`.

## `packages/wasm`: nothing to close

The package has **no JavaScript source at all** — only wasm-pack output under `pkg/` (of which just `ifc-lite.d.ts` is tracked) plus five `node --test` suites that drive the Rust-built binary. Every mutation with observable behaviour lands in Rust and needs a wasm rebuild to take effect, so it is not meaningfully mutable from the JS suite. Nothing is changed for it here.

One environmental note, not a change in this PR: `test/package.test.mjs` asserts the packed tarball contains `pkg/ifc-lite.js` and `pkg/ifc-lite_bg.wasm`, and hard-fails (rather than skipping, as the other four suites do) in a tree where the wasm has not been built.

### Two emitters had no test file at all

**`rust-generator.ts`** (714 lines) produces `rust/core/src/generated/{type_ids.rs,schema.rs}` — the Rust core's entire notion of what an IFC type is.

| mutation | result |
|---|---|
| `pub const ${entity.name.toUpperCase()}` → `${entity.name}` | SURVIVED |
| `from_str` arm `"${name.toUpperCase()}"` → `.toLowerCase()` | SURVIVED |

The second one makes every `from_str` arm dead code against the `s.to_uppercase()` it is fed, so every IFC type parses as `Unknown`. Killed by `test/rust-generator.test.ts`, which pins the casing contract in both directions: SCREAMING_CASE constants, PascalCase enum variants, UPPERCASE `from_str` arms and `as_str()`, PascalCase `name()`, and `parent()` / `is_abstract()` arms present only for the entities that earn them.

**`type-ids-generator.ts`** produces `type-ids.ts`.

| mutation | result |
|---|---|
| reverse map `${id}: '${name}',` → `'${name}': ${id},` | SURVIVED |

Still valid TypeScript, still green — and `getTypeName(id)` returns `undefined` for every id at runtime. Killed by `test/type-ids-generator.test.ts`.

### The emitted runtime helpers were never executed

The generated bundles ship `getTypeId`, `getTypeName`, `isValidTypeId`, `crc32Hash`, `getEntityMetadata`, `getAllAttributesForEntity`, `getInheritanceChainForEntity` and `isKnownEntity`. Because the generator emits them as **text**, nothing in this package ever ran them — they were only substring-matched by name.

| mutation | result |
|---|---|
| emitted `return metadata?.allAttributes \|\| []` → `metadata?.attributes` | SURVIVED |

That drops every inherited attribute from every consumer of the generated bundle. The new tests transpile the emitted module (`ts.transpileModule`, already a devDependency) and import it, so the helper *semantics* are pinned, not their presence as substrings.

### Six more survivors in `typescript-generator.ts` / `express-parser.ts`

| mutation | result |
|---|---|
| `LOGICAL: 'boolean \| null'` → `'boolean'` | SURVIVED |
| `BINARY: 'string'` → `'number'` | SURVIVED |
| bounded `STRING(N)` → `'number'` | SURVIVED |
| `expressType.endsWith('Measure')` → `startsWith` | SURVIVED |
| attribute union parenthesisation `(${t})[]` → `${t}[]` | SURVIVED |
| `const memberName = value.toUpperCase()` → `value` | SURVIVED |
| registry `.replace(/'/g, "\\'")` on enum values → dropped | SURVIVED |
| registry `.replace(/'/g,…).replace(/\n/g,' ')` on types → dropped | SURVIVED |
| `topologicalSort` parent visit → `if (false)` | SURVIVED |
| `Math.min(...sectionMatches)` → `Math.max` | SURVIVED |

- Only REAL/INTEGER/BOOLEAN/STRING had fixtures, so the three-valued `LOGICAL → boolean | null` mapping, `BINARY → string`, and the bounded `STRING(N)` / `BINARY(N)` forms were all free.
- The union parenthesisation that makes `LIST OF LOGICAL` emit `(boolean | null)[]` rather than the precedence-misparsed `boolean | null[]` (which is `boolean | (null[])`, a different type) carries an explaining comment in the source but had no guard.
- Enum MEMBER names are upper-cased, but every fixture value was already uppercase, so the call was untestable. Now pinned with a lowercase and a mixed-case value, asserting the NAME is upper-cased while the literal VALUE is preserved.
- Dropping the SCHEMA_REGISTRY quote/newline escaping emits a registry that does not parse; now checked by evaluating the emitted module, not just matching characters.
- Every fixture already declared parents before children, so disabling `topologicalSort`'s parent visit changed nothing.
- The attributes section ends at the FIRST of WHERE / DERIVE / INVERSE / UNIQUE. Every fixture had at most one of those keywords, so "first" and "last" were the same index; taking the last swallows the DERIVE/INVERSE block and turns its lines into bogus attributes.

## Controls and existing kills

Run **before** any survivor was believed, both expected to die, both did:

| control | killed by |
|---|---|
| CRC32 polynomial `0xedb88320` → `0xedb88321` | `crc32.test.ts` |
| `chain.unshift(current.name)` → `chain.push(...)` | `express-parser.test.ts` "should get inheritance chain" |

Two further mutations died against the pre-existing suite: joining SELECT members with `&` instead of `\|`, and making EXPRESS comment stripping greedy (`[\s\S]*?` → `[\s\S]*`).

## Survivor left alone

The `name[3] >= 'A' && name[3] <= 'Z'` fast path in the emitted `normalizeTypeName` is an **equivalent mutant** for the exported surface: it is a pure shortcut, and every input it accepts resolves to the same registry key through the case-insensitive fallback loop directly below it, while every input it rejects resolves to a non-key either way. Left unchanged.

## Notes

- One of my own new assertions was initially vacuous — `toContain('Blob: string;')` also matched inside `FixedBlob: string;`, so the `BINARY` mutation still survived the test meant to kill it. All entity-attribute assertions are now newline-and-indent anchored, and every survivor above was re-run against the finished suite to confirm it now dies.
- `pnpm typecheck` at the root does not cover test files (the package tsconfig excludes `test`), so the new files were typechecked separately against `packages/codegen/tsconfig.json` with `rootDir: "."` and the exclusion lifted — clean.
- No `as any` / `@ts-ignore`; the evaluated modules are typed by declared interfaces.
- Test-only; no production code changed. **codegen 55 → 102 tests, 5 → 8 test files.** No `packages/wasm/pkg/` file is touched. `scripts/check-test-wiring.mjs` passes.

### DIFF
--- a/packages/codegen/test/crc32.test.ts
+++ b/packages/codegen/test/crc32.test.ts
@@ -7,9 +7,16 @@
  *
  * This hash is a compatibility contract: it must match the pre-computed
  * IDs baked into the generated schema (packages/parser/src/generated/type-ids.ts)
- * and the web-ifc algorithm it mirrors. We pin known values from that
- * generated file (single source of truth for the codegen CLI's output)
- * rather than duplicating web-ifc's table here.
+ * and the web-ifc algorithm it mirrors.
+ *
+ * Two layers of assertion, doing different jobs:
+ *  - `crc32 specification vectors` pins crc32() to CRC-32/ISO-HDLC using
+ *    values that come from outside this repository. Nothing else in this file
+ *    can catch an algorithmic error, because everything else compares crc32()
+ *    against output crc32() produced.
+ *  - the remaining suites pin the *generator* to those hashes via the real
+ *    generated file (packages/parser/src/generated/type-ids.ts), the single
+ *    source of truth for the codegen CLI's output.
  */
 
 import { readFileSync } from 'node:fs';
@@ -45,6 +52,110 @@ function loadGeneratedTypeIds(): Map<string, number> {
   }
   return ids;
 }
+
+/**
+ * Independent, bit-at-a-time reference implementation of CRC-32/ISO-HDLC,
+ * written straight from the algorithm's parameters rather than from
+ * src/crc32.ts:
+ *
+ *   width=32  poly=0x04C11DB7  init=0xFFFFFFFF
+ *   refin=true  refout=true  xorout=0xFFFFFFFF  check=0xCBF43926
+ *
+ * This shares no code and no constants with the implementation under test:
+ * src/crc32.ts is table-driven and uses the *reflected* polynomial
+ * 0xEDB88320 with right shifts; this one is table-free and uses the
+ * *unreflected* polynomial 0x04C11DB7 with left shifts, reflecting each
+ * input byte on the way in and the register on the way out. Two independent
+ * routes to the same specification, so agreement between them is evidence
+ * about the specification and not a tautology.
+ */
+function reflect(value: number, width: number): number {
+  let out = 0;
+  for (let i = 0; i < width; i++) {
+    if ((value >>> i) & 1) out |= 1 << (width - 1 - i);
+  }
+  return out >>> 0;
+}
+
+function crc32Reference(input: string): number {
+  let reg = 0xffffffff;
+  for (let i = 0; i < input.length; i++) {
+    const byte = input.charCodeAt(i);
+    if (byte > 0xff) {
+      throw new Error(`crc32Reference: non-byte input at ${i}: ${byte}`);
+    }
+    reg = (reg ^ (reflect(byte, 8) << 24)) >>> 0;
+    for (let bit = 0; bit < 8; bit++) {
+      reg =
+        (reg & 0x80000000) !== 0
+          ? ((reg << 1) ^ 0x04c11db7) >>> 0
+          : (reg << 1) >>> 0;
+    }
+  }
+  return (reflect(reg, 32) ^ 0xffffffff) >>> 0;
+}
+
+describe('crc32 specification vectors', () => {
+  // Without these, every other assertion in this file compares crc32() with
+  // itself (directly, or via generated/type-ids.ts which crc32() produced).
+  // A wrong-but-self-consistent hash would pass all of them. These are the
+  // only assertions here whose expected values come from outside our code.
+  //
+  // These IDs cross a language boundary: rust/core/src/generated/schema.rs
+  // hashes unknown type names with its own CRC32 and the TS table is shipped
+  // by @ifc-lite/parser. Two internally coherent tables that disagree with
+  // each other is exactly the failure a self-check cannot see, so the same
+  // vectors are pinned on the Rust side in
+  // rust/core/tests/crc32_spec_vectors.rs.
+
+  it('matches the standard CRC-32/ISO-HDLC check value', () => {
+    // "check" is defined by the algorithm's specification as the CRC of the
+    // nine ASCII bytes "123456789". crc32() uppercases its input first, but
+    // digits are unaffected by case, so the vector applies unchanged.
+    expect(crc32('123456789')).toBe(0xcbf43926);
+  });
+
+  it('hashes the empty string to the specified xorout identity', () => {
+    // With no input the register never leaves init=0xFFFFFFFF, so the result
+    // is init ^ xorout = 0xFFFFFFFF ^ 0xFFFFFFFF = 0.
+    expect(crc32('')).toBe(0x00000000);
+  });
+
+  it('agrees with an independent bit-at-a-time reference implementation', () => {
+    // Anchor the reference itself to the published check value before
+    // trusting it as an oracle.
+    expect(crc32Reference('123456789')).toBe(0xcbf43926);
+    expect(crc32Reference('')).toBe(0x00000000);
+
+    // crc32() uppercases; feed the reference the already-uppercased form so
+    // the comparison is about the hash, not about the case folding (which
+    // has its own test below).
+    const names = [
+      'IFCWALL',
+      'IFCWINDOW',
+      'IFCDOOR',
+      'IFCPROJECT',
+      'IFCBUILDINGSTOREY',
+      'IFCEXTRUDEDAREASOLID',
+      'IFCTRIANGULATEDFACESET',
+      'IFCNOTATYPE',
+      'A',
+      'AB',
+      'ABC',
+    ];
+    for (const name of names) {
+      expect(crc32(name)).toBe(crc32Reference(name));
+    }
+  });
+
+  it('agrees with the reference across the full generated IFC entity set', () => {
+    const generated = loadGeneratedTypeIds();
+    expect(generated.size).toBeGreaterThan(500);
+    for (const name of generated.keys()) {
+      expect(crc32(name)).toBe(crc32Reference(name.toUpperCase()));
+    }
+  });
+});
 
 describe('crc32', () => {
   it('matches the pre-computed IfcWall hash from generated/type-ids.ts', () => {
--- a/packages/codegen/test/express-parser.test.ts
+++ b/packages/codegen/test/express-parser.test.ts
@@ -328,6 +328,40 @@ describe('EXPRESS Parser', () => {
     });
   });
 
+  describe('Attribute section boundary', () => {
+    /**
+     * The attributes section ends at the FIRST of WHERE / DERIVE / INVERSE /
+     * UNIQUE. Every existing fixture had at most one of those keywords, so
+     * "first" and "last" were the same index and the choice was untestable —
+     * picking the last one instead swallows the INVERSE/DERIVE block and turns
+     * its lines into bogus attributes.
+     */
+    it('stops at the FIRST section keyword when several are present', () => {
+      const schema = parseExpressSchema(`
+        SCHEMA TEST;
+
+        ENTITY IfcTest
+          SUBTYPE OF (IfcRoot);
+          RealAttr : IfcLabel;
+         DERIVE
+          DerivedAttr : IfcLabel := 'x';
+         INVERSE
+          InverseAttr : SET [0:?] OF IfcRelAssigns FOR RelatedObjects;
+         WHERE
+          WR1 : EXISTS(RealAttr);
+        END_ENTITY;
+
+        END_SCHEMA;
+      `);
+
+      const test = schema.entities[0];
+      expect(test.attributes.map(a => a.name)).toEqual(['RealAttr']);
+      // Both directions: the derived and inverse names must not leak in.
+      expect(test.attributes.map(a => a.name)).not.toContain('DerivedAttr');
+      expect(test.attributes.map(a => a.name)).not.toContain('InverseAttr');
+    });
+  });
+
   describe('Real IFC schema parsing', () => {
     it('should parse IfcWall from real schema', () => {
       const schema = parseExpressSchema(`
--- a/packages/codegen/test/rust-generator.test.ts
+++ b/packages/codegen/test/rust-generator.test.ts
@@ -1,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Tests for the Rust generator (packages/codegen/src/rust-generator.ts).
+ *
+ * This emitter produces `rust/core/src/generated/type_ids.rs` and
+ * `rust/core/src/generated/schema.rs` — the Rust core's entire notion of what
+ * an IFC type IS. Before these tests the whole 700-line module had no test at
+ * all: a mutation that emitted `pub const IfcWall` instead of `pub const
+ * IFCWALL`, or that lower-cased the `from_str` match arms, produced a green
+ * `pnpm test` in this package and only exploded (or, worse, silently
+ * mis-classified every entity) the next time someone regenerated the Rust.
+ *
+ * The identifier CASING rules are the load-bearing part and are pinned in both
+ * directions here:
+ *   - `type_ids.rs` constants are SCREAMING_CASE (`IFCWALL`), Rust's const
+ *     convention.
+ *   - `IfcType` enum variants keep the schema's PascalCase (`IfcWall`).
+ *   - `from_str` matches on the UPPERCASED name, because it is called with
+ *     `s.to_uppercase()`.
+ *   - `as_str()` returns the UPPERCASE STEP spelling; `name()` returns the
+ *     PascalCase display spelling. These two must not collapse into each other.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { parseExpressSchema } from '../src/express-parser.js';
+import { generateRust } from '../src/rust-generator.js';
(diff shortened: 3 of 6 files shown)


BIMvoice:
The production defect in this sweep is real and it reaches shipped output. I verified both halves independently before raising it, and **this PR deliberately does not fix it** — the fix changes generated files, which is your call.

## The dead guard

`packages/codegen/src/express-parser.ts:307`:

```js
/TYPE\s+(\w+)\s*=\s*(?!ENUMERATION|SELECT)([\s\S]*?)END_TYPE\s*;/g
```

The negative lookahead never fires in practice. `\s*` backtracks to zero width, so the lookahead is evaluated at the leading whitespace — `" ENUMERATION"` — and trivially succeeds. Run directly against the two spellings:

```
TYPE IfcFoo = ENUMERATION OF (A ,B); END_TYPE;   -> MATCHED   (guard did not block)
TYPE IfcFoo =ENUMERATION OF (A ,B); END_TYPE;    -> blocked correctly
```

It only works in the one formatting the real EXPRESS schema never uses.

## What it produced

Against `packages/codegen/generated/ifc4/schema-registry.ts` on `upstream/main`:

```
total `types:` entries                                    397
entries whose value is a raw EXPRESS ENUMERATION/SELECT body   267
```

For example:

```
IfcActionRequestTypeEnum: 'ENUMERATION OF \t(EMAIL \t,FAX \t,PHONE \t,POST \t,VERBAL \t,USERDEFINED \t,NOTDEFINED)'
```

So two thirds of `SCHEMA_REGISTRY.types` carry an EXPRESS fragment where an underlying type belongs. `types.ts` is unaffected — 130 aliases, because that path filters separately — so the blast radius is `SCHEMA_REGISTRY.types` only. Anything reading it as a type name has been reading a sentence.

## Why there is no test for it here

Writing one would pin a defect. The sweep produced a test that caught it, and it was removed from this PR rather than committed — a green test asserting the polluted value would make the bug permanent and much harder to notice later.

It also means M15 (removing the `(?!ENUMERATION|SELECT)` lookahead entirely) is an **equivalent mutant whose equivalence is the bug**: deleting a guard that never fires changes nothing. That is the one survivor in this sweep I would not want written off as "unreachable code, no test needed".

The fix is one token — `\s*` -> `\s+`, or move the guard inside the capture — but it regenerates output, so it wants your decision and probably its own PR with the regenerated files reviewed.

## Separately: `packages/wasm` warranted no PR

Verdict on evidence: it has zero JS source; all five suites are integration tests over the Rust-built binary, so there is nothing JS-mutable to sweep. One thing worth knowing: four of them skip cleanly when `pkg/ifc-lite_bg.wasm` is absent, but `test/package.test.mjs` hard-fails instead. A fresh clone therefore gets a red `packages/wasm` run that means nothing. I did not change it.

One process note, since it cuts against this PR rather than for it: the first `BINARY` test written here was itself vacuous — `toContain('Blob: string;')` matched inside `FixedBlob: string;`, so the mutation it existed to kill survived it. That was caught only by re-running every mutation against the finished suite instead of trusting the green. All entity-attribute assertions are now newline-and-indent anchored.

louistrue:
The eight gaps look right and the `packages/wasm` conclusion — nothing meaningfully mutable from the JS suite, because the only source is wasm-pack output and every observable mutation lands in Rust — is the correct answer to report rather than a gap to force.

Flagging that you aimed at weak areas rather than sampling uniformly, and that the 21% is therefore not comparable, is the sort of caveat that makes the number usable instead of alarming.

**One gap remains, and it is the one you found yourself in #2139.**

`crc32` has no external reference vector. Every assertion compares generated output against the same function:

```ts
expect(mod.TYPE_IDS.IfcWall).toBe(crc32('IfcWall'));
expect(source).toContain(`  IfcWall: ${crc32('IfcWall')},\n`);
```

That correctly pins the **generator** — "emit what `crc32` computes" — but nothing pins `crc32` itself. `grep` finds no `0xCBF43926` and no `"123456789"` anywhere in `packages/codegen`.

I checked, and the implementation is right:

```
crc32("123456789") = 0xCBF43926   (the standard CRC-32 check value)
```

So this is a coverage gap, not a bug — exactly the shape you described for xxHash: *"the reader's self-check recomputes with the same function, so it can't catch an algorithmic error either."*

**Why it matters more here than for a cache hash.** These IDs cross a language boundary. `rust/core/src/generated/schema.rs` computes `crc32_hash(&upper)` on the Rust side for the same types, and `packages/parser/src/generated/type-ids.ts` carries the TS table. A wrong-but-self-consistent CRC32 on one side gives you two internally coherent tables that disagree with each other — every entity of an unknown type resolves differently across the boundary, and nothing in either suite notices, because each validates against itself.

One line closes it:

```ts
expect(crc32('123456789')).toBe(0xcbf43926); // standard CRC-32 check value
```

Worth adding the same on the Rust side if it is not there, since the point is that the two agree with the *specification*, not merely with each other.

BIMvoice:
## `crc32` now has an external reference vector — on both sides

Thanks @louistrue, this was a real gap and a well-aimed one. Every assertion compared `crc32` output against `crc32`, so nothing pinned the algorithm itself — the same tautology as the xxHash finding in #2139. To be explicit up front: **the implementation is correct today** (`crc32("123456789") === 0xCBF43926`), so this is coverage, not a bug, and no behaviour changed.

Pushed as `e3729475`.

### TypeScript — `packages/codegen/test/crc32.test.ts`

```ts
// "check" is defined by the CRC-32/ISO-HDLC specification as the CRC of "123456789"
expect(crc32('123456789')).toBe(0xcbf43926);

// with no input the register never leaves init=0xFFFFFFFF, so init ^ xorout = 0
expect(crc32('')).toBe(0x00000000);
```

Plus a third layer: an **independent bit-at-a-time reference implementation** written from the algorithm parameters (`width=32 poly=0x04C11DB7 init=0xFFFFFFFF refin=true refout=true xorout=0xFFFFFFFF`). It shares no code and no constants with the code under test — `src/crc32.ts` is table-driven over the *reflected* polynomial `0xEDB88320` with right shifts; the reference is table-free over the *unreflected* `0x04C11DB7` with left shifts, reflecting bytes in and the register out. Two independent routes to the same spec. It is anchored to `0xCBF43926` itself before being trusted, then cross-checked against `crc32` over the full generated IFC entity set (876 names).

No expected value was obtained by running `crc32` and pasting the result.

### Rust — `rust/core/tests/crc32_spec_vectors.rs` (new file, done, not blocked)

`crc32_hash` is private and lives in the generated `schema.rs`, so it can't be tested directly without hand-editing generated output or changing the generator — neither of which I'd do here. There is a clean indirect route through the public API, which is the only caller anyway:

```rust
// IfcType::from_str falls through to `Unknown(crc32_hash(&upper))`
assert_eq!(unknown_hash("123456789"), 0xCBF4_3926);
assert_eq!(unknown_hash(""), 0x0000_0000);
```

Same two spec vectors, same independent reference implementation (via `u8::reverse_bits`), plus agreement across a set of deliberately-unknown names. **No generated file is touched** — `git status` shows only the two test files.

### Mutation evidence

Applied by exact-substring replacement asserting exactly one hit, re-reading the mutated region before trusting any result, and restored by copying back a saved backup.

**1. TS polynomial** — `0xedb88320` → `0xedb88321` (`count = 1`): the check-value test and both reference cross-checks fail.

**2. TS final XOR** — `(crc ^ 0xffffffff)` → `(crc ^ 0xfffffffe)` (`count = 1`): 6 of 11 fail, including the empty-string vector, which the polynomial mutation correctly leaves alone. The two vectors pin different parameters.

**3. The one that makes the point** — a *self-consistent* wrong CRC32: mutant polynomial **and** a `TYPE_IDS` table regenerated by that same mutant (simulated in the test harness, `count = 1` each for the loader and the `IfcWall` literal):

```
Tests  3 failed | 8 passed (11)
```

All 8 pre-existing assertions go green against a wrong hash. Only the 3 new spec-vector tests fail. That is precisely the failure mode you described.

**4. Rust `crc32_hash`** — `crc >> 8` → `crc >> 7` and `crc ^ 0xffffffff` → `crc ^ 0xfffffffe` (`count = 1`). Running the **entire** `ifc-lite-core` suite under that mutation:

```
Running unittests src/lib.rs
test result: ok. 137 passed; 0 failed
Running tests/crc32_spec_vectors.rs
test result: FAILED. 1 passed; 4 failed
```

137 pre-existing unit tests pass with a broken CRC32. Only the new file notices. (The 1 passing test there only exercises the reference oracle, which is independent of the mutation by design.)

### Counts

| | before | after |
|---|---|---|
| `packages/codegen` | 8 files, 102 tests | 8 files, **106 tests** |
| `crc32.test.ts` | 7 | **11** |
| `rust/core` integration | — | **+5** (new file) |

All green. `rustfmt --check` clean; `cargo clippy -p ifc-lite-core --tests` reports no warning from the new file (the pre-existing `chunks_exact` warnings in the lib are untouched). `pnpm typecheck` excludes `test/`, so I typechecked `crc32.test.ts` separately against the package's own `tsconfig.json` with the `test` exclusion lifted — exit 0.

### Cross-boundary agreement — gap still open

To be precise about what this does and does not prove: both sides are now independently pinned to the *same external specification*, which is what makes silent divergence impossible in the failure mode you described. But I did **not** add a test that compares the TS `TYPE_IDS` table against the Rust one entry by entry, and no such test exists today — I checked `rust/core/tests/` and the parser suite. The repo already has the right pattern for it (`rust/core/tests/fixtures/*.json` shared with `packages/parser/*.parity.test.ts`, e.g. `unit_scale_vectors.json`), so a `type_id_vectors.json` parity fixture would slot straight in. Happy to add it if you want it — flagging rather than building it unasked, since it is a new fixture rather than a test-only change.

BIMvoice:
Both sides are pinned now — TS spec vectors plus a Rust file reaching `crc32_hash` indirectly through `IfcType::from_str`, so no generated file was edited and no generator change was needed.

**But I have to correct two claims that were made on this PR, including one of mine.** I tried to reproduce "every pre-existing assertion stays green against a self-consistent wrong CRC32" and could not.

## What I actually measured

Mutating the polynomial (`0xedb88320` -> `0xedb88321`, replacements=1) makes the table regenerate at load, so the implementation becomes internally consistent but wrong. Against **main's** `crc32.test.ts`:

```
4 failed | 98 passed
```

I then simulated the full regeneration case — rewrote **2428 committed hash values** across `packages/parser/src/generated/type-ids.ts` and the codegen `ifc4`/`ifc4x3` output with the mutant algorithm, so the committed tables agree with the wrong hash:

```
3 failed | 99 passed
```

So the pre-existing suite **does** catch it. The claim that only the new vectors fire was overstated, and I am flagging it rather than letting it stand.

## Why the gap is still real, precisely stated

The thing that catches it is `crc32.test.ts:52` on main:

```ts
expect(crc32('IfcWall')).toBe(2391406946);
```

A frozen literal. I checked it against an independent bit-at-a-time implementation anchored to `0xCBF43926`, and it is **correct** — `crc32('IfcWall') === 2391406946` genuinely.

But that literal, and the committed tables, pin **drift**, not **correctness**. Their provenance is our own output. They tell you the algorithm has not changed since the number was captured; they cannot tell you the number was right when captured. Had `crc32` been wrong from day one, that literal would have frozen the wrong value, the generated tables would agree with it, the Rust table would agree too — and every suite would be green while both languages carried the same wrong answer.

That is the failure mode the spec vector closes and a frozen snapshot cannot: `0xCBF43926` is checkable against the CRC-32/ISO-HDLC definition, not against us. Same distinction as sourcing the xxHash vectors from `cespare/xxhash` in #2139 rather than from our own run.

So your finding stands and the fix is worth having — the reasoning just needs to be "a self-derived snapshot cannot validate itself" rather than "nothing pins it at all".

## What is genuinely still open

**No entry-by-entry TS<->Rust `TYPE_IDS` comparison exists.** Both sides are now independently pinned to the same external specification, which closes the silent-divergence mode you described — but I did not test table equality and am not claiming it. The repo already has the pattern (`rust/core/tests/fixtures/unit_scale_vectors.json` shared with `packages/parser/*.parity.test.ts`); a `type_id_vectors.json` would slot in the same way. That is a new shared fixture rather than a test-only change, so it is your call.

One more thing worth knowing: `rust/rust-toolchain.toml` pins bare `nightly` and overrides the root pin. Pre-existing, not touched here, but it means the Rust lanes are not actually reproducible against the root pin.

louistrue:
This goes further than I asked, and the extra distance is the part that matters.

I suggested one line asserting `0xCBF43926`. You wrote an **independent reference implementation on both sides** — table-free, over the *unreflected* polynomial `0x04C11DB7` with left shifts and explicit reflection — against a production implementation that is table-driven over the reflected `0xEDB88320` with right shifts. Two structurally different algorithms, not one algorithm and its own output.

That distinction is the whole point. A magic constant catches a wrong polynomial; a second implementation *of a different shape* also catches the class of bug where both the code and a hand-copied expectation are wrong in the same way. Your comment says it directly:

> each other is exactly the failure a self-check cannot see

Verified both sides:

```
### TS: 0xedb88320 -> 0xedb88321
     × matches the standard CRC-32/ISO-HDLC check value
     × agrees with an independent bit-at-a-time reference implementation
     × agrees with the reference across the full generated IFC entity set
     × matches the pre-computed IfcWall hash from generated/type-ids.ts

### Rust: one corrupted table entry
     FAILED spec_vectors_are_the_same_ones_pinned_in_typescript
     FAILED crc32_hash_matches_the_standard_check_value
     FAILED crc32_hash_agrees_with_the_independent_reference
```

Restored: TS 11/11, Rust 5/5.

`spec_vectors_are_the_same_ones_pinned_in_typescript` is the one that closes the cross-boundary hole I raised. Pinning each side to the specification independently would leave them both correct-by-construction but unlinked; asserting the *same vectors* on both sides is what makes "TS and Rust agree" a tested property rather than an assumption.

And `agrees with the reference across the full generated IFC entity set` is better than spot vectors — a divergence that only shows up on some particular type name has nowhere to hide.

Documenting the parameters in the header (`width=32 poly=0x04C11DB7 init=0xFFFFFFFF refin=true refout=true xorout=0xFFFFFFFF check=0xCBF43926`) means the next person can identify the variant without reverse-engineering it, which is exactly the ambiguity that makes CRC32 bugs so durable.

Merging once CI clears.

Issue #2073, opened 2026-08-04
scene.clear() destroys instanced templates and nothing re-uploads them, so repeated geometry vanishes


BIMvoice:
Found while investigating #1912 and split out because it is **live and user-visible on a single primary model today**, independent of whether federated instancing is ever enabled. Not a regression from anything recent — it is the current shape of the code.

## What happens

`scene.clear()` destroys every instanced template and empties the array (`packages/renderer/src/scene.ts:3614-3630`):

```ts
for (const it of this.instancedTemplates) {
  it.vertexBuffer.destroy();
  it.indexBuffer.destroy();
  it.instanceBuffer.destroy();
}
this.instancedTemplates = [];
```

Nothing ever re-uploads them. The shard bytes are dropped from the store immediately after their single drain — `clearInstancedShards()` at `useGeometryStreaming.ts:731`, and `dataSlice.ts:323` sets `pendingInstancedShards: null`. There is no retention anywhere, and `scene.addInstancedShard` has exactly one call site in the repo (`useGeometryStreaming.ts:705-732`), reachable only from that drain.

So once cleared, the instanced geometry is gone for the rest of the session.

## Why that is reachable in ordinary use

`useGeometryStreaming` calls `scene.clear()` on every non-streaming geometry reshape — six call sites, including:

- `:332` incremental rebuild — i.e. **adding a federated model**
- `:290`/`:298` shrink — i.e. **hiding a model, or a type-visibility toggle**
- `:234` content-version bump — i.e. `realignFederation`

Repeated opaque occurrences render *only* through the instanced path (`useGeometryStreaming.ts:703-704`), so they disappear.

Two consequences worth stating separately:

1. **Adding a federated model silently deletes the primary model's instanced geometry.**
2. **It is not federation-specific.** A plain type-visibility toggle on a single primary model does it too — which is why this deserves its own issue rather than living inside #1912.

## Reproduced

Driving the real hook under happy-dom + React 19 with a fake renderer, rather than by reading:

```
# CALLS:   ["clear","appendToBatches","addInstancedShard","clear","appendToBatches"]
# clears before/after federated add: 1 2
# instanced templates alive after federated add: 0     (expected 1)

# CALLS-B: ["clear","appendToBatches","addInstancedShard","clear","appendToBatches"]
# instanced templates alive after visibility toggle:  0  (expected 1)
```

[...]

PR #2151, opened 2026-08-04
test(renderer): close 8 mutation-verified coverage gaps

BIMvoice:
test(renderer): close 8 mutation-verified coverage gaps

A mutation sweep of `packages/renderer` (38 test files, 414 tests) — the
last unswept package, and one named in `scripts/check-test-wiring.mjs`'s
own docstring as having had 13 test files silently dark in CI for months.
The wiring is fixed and the suite genuinely runs; this establishes what
that suite is now worth.

18 mutations against GPU-free logic (culling maths, residency/eviction
policy, instancing bookkeeping, GPU-byte accounting). 2 were deliberate
controls run first to prove the harness was operating on mutated source;
both died. Of the 16 real mutations, 7 were already killed, 9 survived —
8 genuine untested gaps, 1 proven-equivalent mutant that is NOT counted
as a gap and for which no test was written.

The ratio is TARGETED, not uniform: mutations were aimed at boundary
conditions and documented-but-unexercised defensive branches, so it is
not comparable to a uniform sample.

contribution-cull (5 gaps)

1. `projectedInstancedRadiusPx` nearest-corner selection. The existing
   "picks the nearest corner per viewDir sign" test uses an axis-aligned
   viewDir, so the X and Y corner picks are multiplied by a zero
   direction component and cannot affect the result — only the Z axis
   was pinned. Flipping the X pick (and independently the Y pick) from
   near corner to far corner kept all 414 tests green. Picking the far
   corner over-estimates the view depth and under-estimates the
   projected size, so an oblique camera — i.e. any real orbit — silently
   culls instanced templates that are large on screen: repeated bolts,
   windows and furniture pop out of the model. Now pinned across all
   eight viewDir sign octants against a brute-force eight-corner oracle.

2. `resolveContributionThresholdPx` disable check. `pixelRadius > 0`
   relaxed to `>= 0` survived: with culling explicitly disabled
   (pixelRadius 0) but a motion boost configured, the mutant resumes
   culling the moment the user grabs the orbit control. Both directions
   pinned, including that the smallest positive radius is not disabled.

3. `projectedAabbRadiusPx` tangency boundary. `depth <= radius` relaxed
   to `depth < radius` survived — no fixture sat exactly on the
   boundary. Pinned in both directions: exactly at tangency fails open,
   just beyond it projects finitely.

4. `projectedInstancedRadiusPx` non-finite-radius guard, orthographic
   branch. The existing poison test only covers perspective, where the
   guard is masked (`!(minDepth > Infinity)` and `!(minDepth > NaN)`
   both already return Infinity). Ortho has no such downstream check:
   without the guard a NaN radius divides straight through and the
   function returns NaN instead of the documented Infinity, leaking NaN
   into the renderer's cull bookkeeping.

residency (1 gap)

5. `selectEvictions` exact-fit stop condition. Every existing case
   overshoots the budget, so the loop's `excess <= 0` break was only
   ever tested in its `< 0` form; relaxing it survived. When the evicted
   bytes land exactly on the budget the mutant destroys one more batch's
   GPU buffers than needed, and the user sees that batch pop back in
   after a needless re-upload. Both the exact-fit and the one-byte-over
   cases are now pinned.

render-stats (1 gap)

6. `sumResidentGpuBytes` LOD1 index buffer. No fixture set
   `lod1IndexBuffer`, so its whole contribution was free — dropping it
   survived. Under-counting it makes the residency budget believe it is
   inside a limit it has actually blown, so eviction never fires and
   large models drift toward GPU allocation failure. Pinned on batches,
   partial sub-batches and meshes.

instanced-render (2 gaps)

7. Opaque/transparent cutoff boundary. Existing fixtures use alpha 0.3,
   0.5 and 1.0, so `< OPAQUE_ALPHA_CUTOFF` relaxed to `<=` survived.
   Alpha exactly 0.99 must be opaque, matching the flat split in
   overlay-routing.ts and scene.ts; under the mutant such an occurrence
   is dropped from the instanced pass while the flat pass also treats it
   as opaque, so the geometry renders nowhere and repeated elements go
   silently missing. Boundary pinned in both directions.

8. Corrupt `templateIndex` guard. The documented defensive branch
   ("drop it loudly-safe") was never reached by a fixture; removing it
   survived. Without it an out-of-range or negative templateIndex in a
   decoded IFNS shard throws mid-upload, so the entire model fails to
   appear instead of losing one bad occurrence.

Not counted as a gap: the early `excess <= 0` return in
`selectEvictions` relaxed to `< 0` survives and is a proven-equivalent
mutant — the only differing input class is excess === 0, where the
loop's own `excess <= 0` break returns the same empty array on the first
candidate, and building the candidate list has no side effects. It still
survives after this change, deliberately; no test was manufactured for
it.

Test-only: no production code is changed. Every new test was
RED-verified under its mutation and GREEN with the mutation reverted,
each mutation applied via an exact-substring replacement asserting a
replacement count of exactly 1, and the mutated region re-read on disk
before any result was believed. Two pre-existing type errors in
contribution-cull.test.ts (orthographic camera literals missing the
required `viewDir`) are fixed in passing, since the package tsconfig
excludes `**/*.test.ts` and these files were typechecked separately with
that exclusion lifted.

## Summary by CodeRabbit

* **Tests**
  * Expanded rendering coverage for contribution culling across interaction states, camera angles, tangencies, and orthographic projections.
  * Added instancing tests for alpha cutoff behavior and invalid template handling.
  * Added regression coverage for resident GPU memory calculations, including additional index-buffer memory.
  * Added eviction tests for exact-fit memory budgets and near-limit scenarios.

### DIFF
--- a/packages/renderer/src/contribution-cull.test.ts
+++ b/packages/renderer/src/contribution-cull.test.ts
@@ -44,6 +44,31 @@ describe('resolveContributionThresholdPx', () => {
     const opts = { pixelRadius: 2, interactingPixelRadius: 0.5 };
     assert.strictEqual(resolveContributionThresholdPx(opts, true), 2);
   });
+
+  // `pixelRadius <= 0` means DISABLED, and disabled must win over any motion
+  // boost: a viewer that opted out of contribution culling must not start
+  // dropping sub-pixel geometry the moment the user grabs the orbit control.
+  // Guards the `> 0` in the disable check (a `>= 0` there lets the motion
+  // branch resurrect culling from a pixelRadius of exactly 0).
+  it('stays disabled while interacting when pixelRadius is 0, even with a motion boost', () => {
+    assert.strictEqual(
+      resolveContributionThresholdPx({ pixelRadius: 0, interactingPixelRadius: 8 }, true),
+      0,
+    );
+    assert.strictEqual(
+      resolveContributionThresholdPx({ pixelRadius: -1, interactingPixelRadius: 8 }, true),
+      0,
+    );
+    // Both directions: the smallest positive radius is NOT disabled, and the
+    // motion boost applies to it normally.
+    assert.strictEqual(
+      resolveContributionThresholdPx(
+        { pixelRadius: Number.MIN_VALUE, interactingPixelRadius: 8 },
+        true,
+      ),
+      8,
+    );
+  });
 });
 
 describe('projectedAabbRadiusPx (perspective)', () => {
@@ -108,11 +133,37 @@ describe('projectedAabbRadiusPx (perspective)', () => {
     );
     assert.strictEqual(px, Infinity);
   });
+
+  // The tangency boundary itself: depth EXACTLY equal to the bounding-sphere
+  // radius means the sphere touches the camera plane, which the contract
+  // treats as "never cull". Pinned in both directions — one epsilon further
+  // out must produce a finite (cullable) projection, or the guard would be
+  // swallowing everything near the camera.
+  it('fails open at exactly depth === radius, and projects finitely just beyond it', () => {
+    // Box [-1,-1,-2]..[1,1,0] → half-diagonal radius = sqrt(4+4+4)/2 = sqrt(3).
+    const radius = Math.sqrt(3);
+    const cam = perspectiveCam();
+    // Centre at z = -radius → view depth (down -Z) is exactly `radius`.
+    const centreZ = -radius;
+    const atBoundary = projectedAabbRadiusPx(
+      [-1, -1, centreZ - 1],
+      [1, 1, centreZ + 1],
+      cam,
+    );
+    assert.strictEqual(atBoundary, Infinity, 'depth === radius must fail open');
+
+    // Push the same box far enough out that depth > radius: now cullable.
+    const farZ = -radius * 100;
+    const beyond = projectedAabbRadiusPx([-1, -1, farZ - 1], [1, 1, farZ + 1], cam);
+    assert.ok(Number.isFinite(beyond), `depth > radius must project finitely, got ${beyond}`);
+    assert.ok(beyond > 0, `expected a positive pixel radius, got ${beyond}`);
+  });
 });
 
 describe('projectedAabbRadiusPx (orthographic)', () => {
   const orthoCam = (orthoHalfHeight: number): CullCameraState => ({
     eye: { x: 0, y: 0, z: 0 },
+    viewDir: { x: 0, y: 0, z: -1 }, // unused in ortho, required by the interface
     mode: 'orthographic',
     fovYRadians: 0,
     orthoHalfHeight,
@@ -174,6 +225,51 @@ describe('projectedInstancedRadiusPx (instanced templates)', () => {
     assert.ok(Math.abs(px - (2 / 100) * 500) < 1e-9, `got ${px}`);
   });
 
+  // The test above only ever exercises the Z axis: with an axis-aligned
+  // viewDir the X and Y corner picks are multiplied by a zero direction
+  // component and cannot affect the result. A real orbit camera looks
+  // obliquely, so all three axes must be pinned independently — picking the
+  // FAR corner on any axis over-estimates the depth, under-estimates the
+  // projected size, and silently culls instanced templates (repeated bolts,
+  // windows, furniture) that are actually large on screen.
+  it('picks the nearest corner on EVERY axis under an oblique view direction', () => {
+    const inv = 1 / Math.sqrt(3);
+    for (const sx of [1, -1]) {
+      for (const sy of [1, -1]) {
+        for (const sz of [1, -1]) {
+          const cam = perspectiveCam({
+            viewDir: { x: sx * inv, y: sy * inv, z: sz * inv },
+          });
+          const min: [number, number, number] = [100, 200, 300];
+          const max: [number, number, number] = [140, 260, 380];
+          const px = projectedInstancedRadiusPx(min, max, 2, cam);
+
+          // Independent oracle: brute-force the minimum view depth over all
+          // eight corners of the union box, which is what the per-axis sign
+          // trick is a shortcut for.
+          let minDepth = Infinity;
+          for (const x of [min[0], max[0]]) {
+            for (const y of [min[1], max[1]]) {
+              for (const z of [min[2], max[2]]) {
+                const d = x * cam.viewDir.x + y * cam.viewDir.y + z * cam.viewDir.z;
+                if (d < minDepth) minDepth = d;
+              }
+            }
+          }
+          if (!(minDepth > 2)) {
+            assert.strictEqual(px, Infinity, `sx=${sx} sy=${sy} sz=${sz} must fail open`);
+            continue;
+          }
+          const expected = (2 / minDepth) * 500;
+          assert.ok(
+            Math.abs(px - expected) < 1e-9,
+            `sx=${sx} sy=${sy} sz=${sz}: got ${px}, expected ${expected}`,
+          );
+        }
+      }
+    }
+  });
+
   it('fails open when an occurrence could reach the camera plane (minDepth <= radius)', () => {
     const px = projectedInstancedRadiusPx([-10, -10, -10], [10, 10, -1], 5, perspectiveCam());
     assert.strictEqual(px, Infinity);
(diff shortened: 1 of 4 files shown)


BIMvoice:
Re-verified the strongest finding independently, on `08772de5`:

```
### nearest-corner X and Y flipped to the FAR corner   (replacements=1 each)

  against main's contribution-cull.test.ts   ->  24 pass, 0 fail   <- SURVIVED
  against this PR's                          ->  not ok 4 - picks the nearest corner
                                                 on EVERY axis under an oblique view direction
                                                 27 pass, 1 fail
```

Restored: 28/28.

The reason it survived is worth stating, because the old test *looks* like it pins corner selection. Its `viewDir` is axis-aligned, so the X and Y terms are multiplied by zero and are free variables — only Z was ever constrained. That is why flipping `nz` was already caught (4 failures) while flipping `nx` and `ny` was invisible. A fixture that exercises the code path without constraining the value it computes.

On screen this is instanced templates — repeated bolts, windows, furniture — silently vanishing under any real orbit camera, because far-corner depth under-estimates their projected size and the contribution cull drops them. No error, just missing geometry.

The new test asserts against a brute-force 8-corner oracle across all 8 sign octants, so it pins the value rather than the branch.

## Two things worth separating from this PR

**1. `packages/renderer`'s wiring problem is genuinely fixed.** Given this package is named in `check-test-wiring.mjs`'s own docstring as one of the two that had 13 files dark for months, I checked rather than assumed: 38 files, 414 tests, 0 skipped, no conditional skips, and the gate reports correct wiring. The historical problem is closed — this sweep found ordinary boundary gaps, not dark files.

**2. Five renderer test files have pre-existing type errors CI never sees.** Following the `pnpm typecheck` gap reported on #2145 (root tsconfig excludes `**/*.test.ts`), typechecking the renderer's test files against the package's own options surfaced:

```
camera-controls.test.ts:66
lod-simplify.test.ts:114            (3 errors)
renderer-render-paths.test.ts:9
scene-finalize-rollback.test.ts:8
scene-textured-origin.test.ts:25
```

The last three import `MeshData` / `RenderPipeline` from `./types.js`, which no longer exports them — so those files reference types that have since been removed or renamed. They still *run*, because `tsx` strips types without checking them; nothing tells you the imports are stale.

Not fixed here — out of scope for a coverage sweep, and it is the same underlying decision you already have open from #2145 about whether test files should be typechecked at all. Flagging it because it is concrete evidence for that decision rather than a hypothetical: turning typecheck on for tests would surface these five immediately.

**On #2073:** not touched. `scene.clear()` and instanced-template re-upload live in `scene.ts`; this work stayed in four pure modules. One observation that may bear on your design ruling — `sumResidentGpuBytes` is the only accounting of instanced-template GPU bytes and had no LOD1 coverage at all (gap 6 here). If the #2073 fix adds a re-upload path, the residency accounting it feeds is now at least pinned.

BIMvoice:
**CONFIRMED — CodeRabbit was right, and the escaping mutation is exactly the one the test was written to catch.** Fixed in c73f3e89.

### How this was measured

Not by reading the fixture — by mutating the corner selection in `contribution-cull.ts` **one axis at a time** and recording, per octant, whether that octant's assertion changes. An octant whose asserted value is identical under correct and mutated code is asserting nothing.

`min = [100,200,300]`, `max = [140,260,380]`, `maxOccRadius = 2`, `eye = (0,0,0)`.

### Before — per-octant discrimination (`yes` = mutation detected here)

| octant (sx,sy,sz) | oracle minDepth | x flip | y flip | z flip |
|---|---:|---|---|---|
| (+1,+1,+1) | 346.410 | yes | yes | yes |
| (+1,+1,−1) | **−46.188** | no | no | no |
| (+1,−1,+1) | 80.829 | yes | yes | yes |
| (+1,−1,−1) | **−311.769** | no | no | no |
| (−1,+1,+1) | 207.846 | yes | yes | yes |
| (−1,+1,−1) | **−184.752** | no | no | no |
| (−1,−1,+1) | **−57.735** | no | no | no |
| (−1,−1,−1) | **−450.333** | no | no | no |

Five of eight octants sit **behind** the camera. There, the nearest corner and the far corner both give a non-positive depth, `!(minDepth > maxOccRadius)` fires either way, and the assertion collapses to `Infinity === Infinity`. CodeRabbit's arithmetic for `(+1,+1,−1)` — `(100 + 200 − 380)/sqrt(3) = −46.19` — is correct.

### Why this mattered in practice

The whole-test-level flip mutations still failed the test (any one surviving octant is enough to fail a loop), so the gap was invisible from a pass/fail run. The gap shows up on a mutation that errs **only** in the sz<0 half-space — i.e. the fixed-corner bug this test exists to pin:

```
const nz = unionMin[2];   // ignore the sign, always take min — pre-fix behaviour on z
```

- `const nx = unionMin[0];` → test FAILS (caught, via the (−1,+1,+1) octant)
- `const ny = unionMin[1];` → test FAILS (caught, via the (+1,−1,+1) octant)
- `const nz = unionMin[2];` → **test PASSES.** Escapes clean.

### The fix

Orbit the eye 500 units back along each view direction from the box centre (CodeRabbit's proposal, with the origin replaced by the box centre so all eight octants get an identical depth), and include `cam.eye` in the brute-force oracle. The fail-open branch is replaced by an assertion that the fixture stays in front of the camera, so the degeneracy cannot silently come back if the box or the standoff is ever edited.

### After — re-measured, same method

| octant (sx,sy,sz) | oracle minDepth | x flip | y flip | z flip |
|---|---:|---|---|---|
| (+1,+1,+1) | 448.038 | yes | yes | yes |
| (+1,+1,−1) | 448.038 | yes | yes | yes |
| (+1,−1,+1) | 448.038 | yes | yes | yes |
| (+1,−1,−1) | 448.038 | yes | yes | yes |
| (−1,+1,+1) | 448.038 | yes | yes | yes |
| (−1,+1,−1) | 448.038 | yes | yes | yes |
| (−1,−1,+1) | 448.038 | yes | yes | yes |
| (−1,−1,−1) | 448.038 | yes | yes | yes |

24/24 discriminating. And `const nz = unionMin[2];` now **fails** the test.

### Sibling `Infinity` assertions in this PR — checked, one note

`Infinity` is a sink, so every new assertion expecting it was mutation-probed too:

- **`fails open at exactly depth === radius`** (`contribution-cull.test.ts`) — has teeth. `if (depth <= radius)` → `if (depth < radius)` fails it, and it is the *only* test in the package that catches that flip. The paired "one step further out must be finite" assertion covers the opposite direction. No change needed.
- **`orthographic: fails open with Infinity (not NaN)`** — the NaN half has teeth: narrowing the guard to `if (maxOccRadius === Infinity)` fails the test. The `maxOccRadius = Infinity` half does **not**: narrowing to `if (Number.isNaN(maxOccRadius))` leaves the whole suite green, because an infinite radius returns `Infinity` through the ortho divide (`Infinity/100*500`) and through the perspective `!(minDepth > Infinity)` guard anyway. That assertion is over-determined rather than wrong — no mutation of the guard can change its value — so I left it as documentation of the contract rather than deleting it. Flagging it so nobody counts it as coverage of the `Infinity` branch.

Renderer suite: 422 passing before and after. `tsc` over `src/**/*` with the `**/*.test.ts` exclusion lifted reports the same 7 pre-existing errors in 5 other test files and none in `contribution-cull.test.ts`.

Credit to CodeRabbit — correct finding, correct remedy, and it named the right octant.

BIMvoice:
Re-ran this myself, and one detail in the write-up above needs tightening — the escape was narrower than "escapes clean".

Mutating `contribution-cull.ts:122` to a fixed corner (`const nz = unionMin[2];`, replacements=1), which is wrong only in the `sz < 0` half-space:

```
### against the OLD fixture (08772de5)
    ok  3 - picks the nearest corner per viewDir sign (not a fixed corner)
    ok  4 - picks the nearest corner on EVERY axis under an oblique view direction   <- MISSED IT
    not ok 1 - projects the max occurrence radius at the union box NEAREST view depth
    not ok 2 - is an upper bound for every real occurrence in the box
    not ok 5 - fails open when an occurrence could reach the camera plane
    28 tests, 25 pass, 3 fail

### against the NEW fixture (c73f3e89)
    not ok 4 - picks the nearest corner on EVERY axis under an oblique view direction  <- now caught
    28 tests, 24 pass, 4 fail
```

Restored: 28/28.

So the mutation **would not** have reached CI green — three sibling tests caught it. What escaped was the test written specifically to pin corner selection, in 5 of its 8 octants. That is still worth fixing, and CodeRabbit was right about the mechanism and the arithmetic, but "the bug escapes undetected" overstates it: the package caught it, just not with the assertion that names the property.

**The measurement is the real result.** 5 of 8 octants put the box behind the camera, so both the correct and the wrong corner return `Infinity` and those octants assert nothing. Whole-axis flip mutations still failed the test, because one live octant is enough to fail a loop — which is exactly why the degeneracy was invisible from pass/fail, and why my earlier verification on this PR (flipping `nx`/`ny` fails the test) was true but insufficient. It established the test had teeth *somewhere*; it did not establish teeth *per octant*. A loop over N cases needs per-case discrimination measured, not inferred from the loop failing.

After the fix all 8 octants share minDepth 448.038 and all 24 axis/octant cells discriminate.

**One thing deliberately left alone, flagged so it is not miscounted as coverage:** in `orthographic: fails open with Infinity (not NaN)`, the NaN half has teeth but the `maxOccRadius = Infinity` half does not — a NaN-only guard leaves the suite green, because an infinite radius returns `Infinity` through the ortho divide and the perspective guard regardless. That assertion is over-determined rather than wrong, so changing it would add churn without adding a kill. Worth knowing it does not pin that branch.

PR #2152, opened 2026-08-04
test(viewer-embed): wire the app into CI and pin the iframe bridge boundary

BIMvoice:
## Why

`apps/viewer-embed` had **no test files and no `test` script**, so `turbo test` skipped the whole app. `src/bridge/handler.ts` is the embed side of the iframe postMessage trust boundary: it decides which senders are listened to, which commands run, and which origin replies are addressed to. `src/bridge/urlParams.ts` parses attacker-influenceable query params (anyone can craft the iframe `src`) and feeds the bridge's origin allowlist. Neither had a single assertion behind it.

`src/components/EmbedViewer.tsx` is React/render wiring and is deliberately left alone — this PR covers the protocol logic only.

## What

- `apps/viewer-embed/package.json`: adds `"test": "vitest run"` + the `vitest` devDependency (vitest is the runner in 30 of 35 workspace packages).
- `apps/viewer-embed/vitest.config.ts`: `node` environment with a hand-rolled `window` stub, so every `postMessage` **targetOrigin** assertion is exact rather than mediated by a DOM implementation. The single value `handler.ts` needs from the viewer store barrel is mocked, mirroring `apps/viewer/src/store/globalId.ts` exactly, so the test does not drag in zustand/renderer/wasm.
- 115 tests across `handler.test.ts` (77) and `urlParams.test.ts` (38).

**No production behaviour is changed.** The documented fail-open inbound default (`isOriginAllowed` permits all when the allowlist is empty) and the fail-closed outbound default (content withheld until a concrete parent origin is known) are pinned exactly as they are, in **both** directions, so a silent relaxation of either fails the suite. `pnpm check:test-wiring` passes.

### Covered

| Area | Pinned |
|---|---|
| Origin filtering | empty allowlist accepts any sender; configured allowlist drops others; exact match (no prefix/suffix/scheme slippage); a rejected sender does not latch the outbound target |
| Envelope validation | `null` / string / number / missing `source` / foreign `source` all ignored, plus a positive control; no command runs against a torn-down context |
| Parent-origin state machine | READY-only broadcast to `*`; content events withheld while `*`; first accepted origin latches; `'null'` / `'*'` / `''` never latch; latching is once-only; `expectedParentOrigin` seeds but stays overridable; `destroyBridge` resets; nothing posts outside an iframe |
| INIT | RESPONSE-then-INIT_ACK ordering; token match/mismatch/missing/unconfigured; silent on mismatch without a `requestId`; token cleared by `destroyBridge` |
| Dispatch | `UNKNOWN_COMMAND`; `COMMAND_FAILED` incl. non-`Error` rejections; fire-and-forget stays silent; 500 MB buffer ceiling incl. the exact boundary |
| Every command | selection, visibility, colours, camera, section plane, type visibility, `GET_PROPERTIES` found/not-found, `GET_MODEL_INFO` aggregation, `GET_SCREENSHOT` |
| URL params | http(s) scheme gate (accepts + rejects `javascript:` / `data:` / `file:` / `blob:` / `ftp:`); anchored hex-colour regex; origin normalisation; `allowOrigins` left unset when every entry is invalid; camera arity; view presets; `autoLoad` polarity |

## Verification

**Mutation-tested.** 40 mutations applied by exact-substring replacement (each asserting a replacement count of 1, each re-read from disk to confirm the text changed, each restored from a saved backup — never `git checkout`).

- **40 killed / 40 applied — 100 %.** Flagged **TARGETED**: mutations aimed at the origin filter, the capture state machine, the envelope guard and the param validators, against a zero-test baseline. Not comparable to a uniform sweep.
- **2 control mutations** run first and both died as expected, proving the harness observes the file it claims to.
- One survivor during the sweep — dropping the `if (!ctx) return;` guard in `onMessage` — was closed by a test that replays a message through a listener reference held across `destroyBridge`.

**Typecheck:** the root `tsconfig.json` excludes `**/*.test.ts`, so `pnpm typecheck` does not see these files. Both were typechecked separately against the app's own compiler options with that exclusion lifted: **zero errors** attributable to `src/bridge`.

Baseline 0 tests / 0 files → final **115 tests / 2 files**, all green.

### DIFF
--- a/apps/viewer-embed/package.json
+++ b/apps/viewer-embed/package.json
@@ -8,7 +8,8 @@
     "dev": "vite",
     "build": "vite build",
     "typecheck": "tsc --noEmit",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": "vitest run"
   },
   "dependencies": {
     "@ifc-lite/bcf": "workspace:^",
@@ -52,6 +53,7 @@
     "typescript": "^6.0.3",
     "vite": "^8.2.0",
     "vite-plugin-top-level-await": "^1.6.0",
-    "vite-plugin-wasm": "^3.6.0"
+    "vite-plugin-wasm": "^3.6.0",
+    "vitest": "^4.1.10"
   }
 }
--- a/apps/viewer-embed/src/bridge/handler.test.ts
+++ b/apps/viewer-embed/src/bridge/handler.test.ts
@@ -1,0 +1,805 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Inbound postMessage bridge tests.
+ *
+ * This is the embed side of the iframe trust boundary: it decides which
+ * senders are listened to, which commands run, and — critically — which
+ * origin replies are addressed to. Those decisions are pinned in BOTH
+ * directions here (accepted and dropped, latched and not latched) so a
+ * silent relaxation of the boundary fails the suite.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// The bridge only needs one value from the viewer store barrel; importing the
+// real barrel would drag in zustand + renderer + wasm. Mirror the real
+// implementation (apps/viewer/src/store/globalId.ts) exactly.
+vi.mock('@/store/index.js', () => ({
+  toGlobalIdFromModels: (
+    models: ReadonlyMap<string, { idOffset?: number }>,
+    modelId: string,
+    expressId: number,
+  ): number => {
+    if (modelId === 'legacy' || modelId === 'default' || modelId === '__legacy__') return expressId;
+    const model = models.get(modelId);
+    if (!model) return expressId;
+    return expressId + (model.idOffset ?? 0);
+  },
+}));
+
+import { EMBED_SOURCE, PROTOCOL_VERSION } from '@ifc-lite/embed-protocol';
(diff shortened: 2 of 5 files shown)


louistrue:
`apps/viewer-embed` had no CI coverage, and this is the inbound half of the boundary #2149 covered from the SDK side — so between them the whole `postMessage` channel is now pinned from both ends. Verified the subtlest one:

```
### latch the origin BEFORE the allowlist check
     × does not latch the outbound origin from a rejected sender
      Tests  1 failed | 114 passed (115)
```

Restored: 115/115.

**That ordering is the actual vulnerability, and it is easy to write the wrong way round.** `captureParentOrigin` sets the `targetOrigin` every reply and event is sent to. If it runs before `isOriginAllowed`, an attacker sends one message that gets *correctly rejected* — and still redirects every subsequent outbound message to themselves. The command is refused; the channel is stolen. Nothing about the rejection looks wrong from the host's side.

Three others in that file are worth keeping for the same reason:

- **`accepts commands from ANY origin when no allowlist is configured`** — pins the permissive default as deliberate. Without it, someone "hardens" the empty case and silently breaks every generic embed.
- **`matches allowlist entries exactly, not by prefix or suffix`** — the same trap as the SDK side, where `https://host.example` would otherwise admit `https://host.example.attacker.com`. Confirmed the implementation is `allowedOrigins.includes(origin)`, i.e. exact.
- **`latches exactly once — a later origin cannot re-target the channel`** — closes re-targeting after a legitimate handshake.

Those four together are a threat model, not a coverage exercise. The ordering one in particular is the kind of bug that survives review because both statements are individually correct.

Wiring the `test` script and vitest config matters as much as the 805 lines — this is an app, so nothing would have run it otherwise.

PR #2153, opened 2026-08-04
test(server-bin,create-ifc-lite): wire both packages into CI and close 32 mutation-verified gaps

BIMvoice:
test(server-bin,create-ifc-lite): wire both packages into CI and close 32 mutation-verified gaps

`packages/server-bin` and `create-ifc-lite` both had zero test files and no
`test` script, so `turbo test` skipped them entirely. Neither is a thin
wrapper:

- `server-bin` decides which release asset to download for the current
  platform (`platform.ts` — arch/platform validation, musl detection, target
  triple, archive naming), verifies the downloaded archive's SHA-256 before it
  is extracted, chmod'd and executed (`checksum.ts` — including the fail-open
  vs fail-closed decision and the SHA256SUMS line parser), and maps the child
  process's exit signal to a shell exit code (`binary.ts`).
- `create-ifc-lite` parses argv, validates the project name before it is used
  as a directory name (the guard that keeps `join(cwd, name)` under cwd),
  refuses to write into an existing directory, and rewrites `workspace:`
  ranges into published versions resolved from the registry.

Both packages now have a `test` script and a vitest devDependency — that is
the only reason the suites can run at all; `pnpm check:test-wiring` still
passes. Baseline 0 tests -> 61 in server-bin, 82 in create-ifc-lite.

Every gap was found by mutation. Each mutation was applied as a verified
single-occurrence substring replacement, the mutated region was re-read from
disk and asserted to contain the new text (and not the old) before any result
was believed, and the source was restored from a saved backup copy. 33
mutations + 3 deliberate controls; 35 killed, 1 survivor. Because the baseline
was zero tests, this near-100% ratio measures "the package had no tests", NOT
suite quality — it is not comparable to a uniform sweep of an already-tested
package.

Controls (expected to die, all did): rewriting the Windows `.exe` binary name
to the Unix one; turning the checksum comparison into `if (false)`; dropping
the `^` from the resolved version range.

Representative gaps closed:

- Deleting the archive on a checksum mismatch was invisible — a tampered
  artifact could be left on disk. Killed by "deletes the archive so a tampered
  artifact can never be extracted or executed".
- Turning the fail-open branch (`if (!expected)`) into a never-taken condition
  was invisible. Killed by "warns and proceeds when neither the sidecar nor
  SHA256SUMS exists".
- Loosening the SHA256SUMS filename match from `=== name || endsWith('/'+name)`
  to a bare `endsWith(name)` was invisible; that admits a `evil-<archive>.tar.gz`
  line from the same release. Killed by "rejects a name that merely ends with
  the archive name without a path separator".
- Loosening the 64-hex digest pattern to `{8,}` was invisible. Killed by
  "ignores lines whose digest is not exactly 64 hex chars".
- Deriving the SHA256SUMS URL by appending to the asset URL instead of
  replacing its last segment was invisible. Killed by "derives the SHA256SUMS
  url from the SAME release directory as the asset".
- Dropping the `-musl` suffix, inverting the zip/tar.gz choice, admitting
  `ia32`, and removing the supported-target check were all invisible. Killed by
  the platform matrix, including the two boundary cases `win32-arm64` and
  `linux-arm64-musl` (valid platform, valid arch, no published target).
- `runBinary` mapping SIGINT/SIGTERM/other to 130/143/129, and `code ?? 0`,
  were both unpinned. Killed by "maps SIGINT to 130, SIGTERM to 143 and
  anything else to 129" and "maps a null exit code to 0".
- Removing the `cleanup()` that unregisters the SIGINT/SIGTERM/SIGHUP handlers
  after the child exits was invisible — the leak this guards against is
  unbounded listener growth. Killed by "removes its signal listeners once the
  child exits" and "does not accumulate listeners across repeated runs".
- Neutralising the project-name guard (`hasDotSegment = false`, or unanchoring
  the name regex) was invisible. Killed by the rejection table, which covers
  `../evil`, `..`, `.`, `/tmp/absolute`, `a/../b`, `@scope/..` and shell
  metacharacters, plus "creates nothing anywhere when the name is rejected".
- Removing the existing-directory refusal was invisible. Killed by "refuses to
  scaffold into an existing directory" and "does not touch the contents of the
  existing directory".
- `args[++i]` not consuming the `--template` value was invisible. Killed by
  "consumes the template value so it is not mistaken for the project name".
- Dropping `.reverse()` from the version candidate list (picking the OLDEST
  installable version instead of the newest), narrowing `depFields` to
  `dependencies`, removing the npm package-name validation, and removing the
  `.git` cleanup were all invisible. Each now has a dedicated test.

The single survivor is EQUIVALENT: `const isMusl = platform === 'linux' &&
isMuslLinux()` (platform.ts:79) still returns false without the left operand,
because `isMuslLinux` itself returns early on `process.platform !== 'linux'`
(platform.ts:41-43), and `platform` is that same value read at platform.ts:59.
The redundant guard is worth keeping as documentation of intent.

One KNOWN DEFECT is documented by test rather than fixed, because fixing it
changes behaviour: `create-ifc-lite/src/index.ts:73` gates the template on
`t in TEMPLATES`, and `in` walks the prototype chain, so
`--template toString` (or `constructor`, `valueOf`, `hasOwnProperty`) is
accepted and silently scaffolds the `basic` template instead of printing
"Invalid template". The tests are named "KNOWN DEFECT: ..." so the fix is a
one-line change plus flipping those assertions.

Test-only; no production code changed. The test files were typechecked
separately against each package's own compiler options with the `src`-only
`include` lifted, since the root `pnpm typecheck` does not cover them.

### DIFF
--- a/packages/create-ifc-lite/package.json
+++ b/packages/create-ifc-lite/package.json
@@ -9,7 +9,8 @@
   "main": "./dist/index.js",
   "scripts": {
     "build": "pnpm exec tsc",
-    "dev": "pnpm exec tsc --watch"
+    "dev": "pnpm exec tsc --watch",
+    "test": "vitest run"
   },
   "files": [
     "dist"
@@ -36,7 +37,8 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.2",
-    "typescript": "^5.3.0"
+    "typescript": "^5.3.0",
+    "vitest": "^4.1.10"
   },
   "homepage": "https://ifclite.dev/docs/",
   "bugs": "https://github.com/LTplus-AG/ifc-lite/issues"
--- a/packages/create-ifc-lite/test/cli.test.ts
+++ b/packages/create-ifc-lite/test/cli.test.ts
@@ -1,0 +1,251 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect, beforeEach, afterEach, beforeAll } from 'vitest';
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, mkdirSync, rmSync, existsSync, writeFileSync, chmodSync, readdirSync, realpathSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname, resolve } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+import { createRequire } from 'node:module';
+
+/**
+ * The scaffolder is a bin with no exports, so its argument parsing and its
+ * project-name guard can only be exercised by running it. Every run here uses a
+ * throwaway temp directory as cwd, and `npm` is shadowed by a stub that fails
+ * immediately, so no test ever reaches the network or writes outside the temp
+ * directory. Runs that get past validation are asserted on the banner the CLI
+ * prints BEFORE any template work begins.
+ */
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const CLI = resolve(__dirname, '../src/index.ts');
+// Resolve tsx from THIS package, not from the spawned process's cwd (a temp dir
+// with no node_modules), otherwise `node --import tsx` cannot find the loader.
+const TSX_LOADER = createRequire(import.meta.url).resolve('tsx');
+
+let workDir: string;
+let stubBin: string;
+
+beforeAll(() => {
+  // A fake `npm` that exits non-zero at once, so template scaffolding cannot
+  // reach the real registry from a unit test.
+  stubBin = mkdtempSync(join(tmpdir(), 'ifclite-stubbin-'));
+  const npmStub = join(stubBin, 'npm');
+  writeFileSync(npmStub, '#!/bin/sh\necho "registry disabled in tests" >&2\nexit 1\n');
+  chmodSync(npmStub, 0o755);
+});
+
+beforeEach(() => {
+  // realpath: on macOS /var is a symlink to /private/var and the child's
+  // process.cwd() reports the resolved form.
+  workDir = realpathSync(mkdtempSync(join(tmpdir(), 'ifclite-create-')));
+});
+
+afterEach(() => {
+  rmSync(workDir, { recursive: true, force: true });
+});
+
+function run(args: string[]) {
+  const result = spawnSync(process.execPath, ['--import', pathToFileURL(TSX_LOADER).href, CLI, ...args], {
+    cwd: workDir,
+    encoding: 'utf-8',
+    env: {
+      ...process.env,
+      PATH: `${stubBin}:${process.env.PATH ?? ''}`,
+      npm_config_registry: 'http://127.0.0.1:1/',
+    },
+  });
+  return {
+    status: result.status,
+    stdout: result.stdout ?? '',
+    stderr: result.stderr ?? '',
+  };
+}
+
+describe('project name validation', () => {
+  const rejected = [
+    ['../evil', 'parent traversal'],
+    ['../../etc', 'multi-level traversal'],
+    ['..', 'a bare parent reference'],
+    ['.', 'the current directory'],
+    ['foo/bar', 'a nested path'],
+    ['/tmp/absolute', 'an absolute path'],
+    ['a/../b', 'traversal in the middle'],
+    ['@scope/..', 'a scoped name whose last segment is a dot-segment'],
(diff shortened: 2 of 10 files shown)


BIMvoice:
Re-verified the defect independently. `src/index.ts:73` guards the template with `t in TEMPLATES`, and `in` walks the prototype chain:

```
basic            in:true   hasOwn:true
constructor      in:true   hasOwn:false   <-- accepted, then silently scaffolds basic
toString         in:true   hasOwn:false   <-- same
valueOf          in:true   hasOwn:false   <-- same
hasOwnProperty   in:true   hasOwn:false   <-- same
__proto__        in:true   hasOwn:false   <-- same
nope             in:false  hasOwn:false   (correctly rejected)
```

`npx create-ifc-lite my-app --template constructor` prints no error and scaffolds the **basic** template with basic next-steps, instead of `Invalid template: constructor`. Not exploitable — the value never reaches a filesystem path — but it converts a typo into a silently wrong scaffold, which is the worst outcome for a scaffolding tool because the user only finds out later.

This is the same shape as the `dispatchToBackend` namespace gate in #2139: `in` where `Object.hasOwn` was meant, with the prototype chain supplying the false positive.

**Fix is one token** — `Object.hasOwn(TEMPLATES, t)` — but it changes user-visible CLI behaviour, so it is your call and it is not in this PR.

**How it is currently represented, so you can decide deliberately:** four tests pin the *present* behaviour, named `KNOWN DEFECT: --template <key> is accepted and silently scaffolds basic`. That documents it rather than hiding it, and taking the fix means flipping those four assertions. If you would rather the PR not encode the defect at all, say so and I will drop those four and leave the branch silent on it.

Two other things checked and found sound, so they are not open questions:

- **Project-name validation is already correct** (`index.ts:87-95`). `../evil`, `../../etc`, `..`, `.`, `foo/bar`, `/tmp/absolute`, `a/../b`, `@scope/..`, `~/home`, spaces, `;`, `$(...)` and backslashes are all rejected; `@scope/pkg`, `app.v2`, `_x` accepted. It refuses any existing directory, empty or not, and a name colliding with a file. Now pinned in both directions — the scoped `@scope/..` case is the one a char-class regex alone would have let through.
- **`server-bin` binds nothing.** It `spawn`s the downloaded Rust binary with argv passthrough (`binary.ts:312-318`); `PORT`/host are the Rust server's concern, and there is no `0.0.0.0` anywhere in the package. The cache dir is fixed at `binary.ts:28`, so no user-controlled path resolution. Nothing to report there.

The 97% kill ratio is **targeted and not comparable to a uniform sweep** — the baseline was zero tests in both packages, so it measures "these had no tests at all" rather than suite quality. The one survivor (`platform === 'linux' && isMuslLinux()` -> `isMuslLinux()`) is proven equivalent: `isMuslLinux` returns early on `process.platform !== 'linux'` (`platform.ts:41-43`) and `platform` is that same value read at `:59`, so the left operand is strictly redundant. No test manufactured for it.

louistrue:
`server-bin` downloads and verifies a release binary, so this is supply-chain surface — and it had no CI coverage. Verified the two that matter most:

```
### don't delete the archive on a digest mismatch
     × deletes the archive so a tampered artifact can never be extracted or executed

### drop toLowerCase (case-sensitive digest compare)
     × accepts an uppercase digest (comparison is case-normalised)
      Tests  1 failed | 60 passed (61)   both
```

Restored: 61/61.

**The deletion test is the important one, and its name states the threat rather than the behaviour.** Throwing on a mismatch is only half the defence — if the archive survives on disk, a later run, a retry path, or a human poking at the cache directory can still extract it. "Never extracted or executed" is the property; "throws an error" is just the mechanism.

**The case-normalisation one cuts the other way** — it protects availability, not integrity. Publishers emit uppercase hex (`sha256sum` vs some CI tooling), and a case-sensitive compare would reject a *correct* checksum and brick every install. That is the failure people misdiagnose as a corrupted download.

The sidecar-format coverage is the right shape too: bare digest, `"<hex>  <name>"`, and the binary-mode `*name` marker are all real `sha256sum` output variants, and a parser that handles only one silently falls through to the fail-open path — which brings me to the one thing worth stating explicitly.

**`fail open` on a missing checksum deserves a comment, and it has one:**

> a missing checksum asset fails open (warn and proceed)

Your test `warns and proceeds when neither the sidecar nor SHA256SUMS exists` pins that as deliberate. I would leave it as is — the release pipeline is the right place to guarantee the sidecar exists — but it is worth knowing that a parse bug in the sidecar reader degrades into exactly this path, which is why the three format tests are load-bearing rather than pedantic.

One process note on my side: my first attempt at the deletion mutation had a Python syntax error and never ran, and the suite printed a clean `61 passed`. Checking that `unlinkSync(archivePath)` was still present in the file is what caught it — ninth time today that a mutation of mine silently applied to nothing.

PR #2154, opened 2026-08-05
test(viewer-core): close the coverage gap on the viewer HTTP server, SSE and HTML escaping

BIMvoice:
`@ifc-lite/viewer-core` ships an HTTP server, an SSE broadcast channel, a REST command API and the generated viewer HTML — behind **4 unit tests** that only touched two pure path helpers. Every route, status code, MIME type, SSE frame and the model-name escaping were unverified.

**4 tests / 2 suites → 117 tests / 11 suites**, ~0.9 s. Test-only; no production code changed.

## How the coverage was proven

Every assertion here was verified by mutation, not by inspection. 67 mutations against load-bearing logic in `server.ts`, `viewer-html.ts` and `streaming-viewer.ts`; each applied by exact-substring replacement asserting a replacement count of exactly 1, re-read from disk to confirm the text changed, run under a hard wall-clock timeout, and restored from a saved backup.

**65 / 67 killed (97%).** Two deliberate controls (corrupt the 404 body; corrupt the SSE greeting) were run first and both died, before any survivor was believed.

That ratio is **targeted, not a uniform sweep** — the mutations were aimed at the logic these tests were written for, against a 4-test baseline. It is not comparable to a whole-package mutation score.

Representative kills:

| Mutation | Killed by |
|---|---|
| drop the `Array.isArray(parsed) ? parsed : [parsed]` wrap | `wraps a single (non-array) element body…` (+5) |
| invert that ternary | `passes an array body straight through…` (+6) |
| unanchor the CORS origin regex | 5 of the 8 rejected-origin cases |
| CORS regex → `if (true)` | all 8 rejected-origin cases |
| require a port in the CORS regex | `…for http://localhost`, `…for http://127.0.0.1` |
| drop `127.0.0.1` from the allowlist | the two `127.0.0.1` cases |
| `https?` → `http` | `…for https://localhost:8443` |
| SSE frame `\n\n` → `\n` | 3 SSE tests |
| `break` after the first SSE client | `broadcasts to every subscriber, not only the first` |
| broadcast before validating the action | `greets a subscriber, counts it, and forwards…` |
| never delete the client on close | 4 tests |
| dispatch on `req.url` instead of `url.pathname` | `ignores the query string when matching a route` |
| each route's method check removed | the matching `does not treat <VERB> …` test |
| `readBody` resolving on the first chunk | `reads the whole request body, not just the first chunk` |
| 501 → 400, 204 → 404, 204 → 200, 400 → 200 | the matching status-code test |
| `.wasm`/`.html`/`.js` MIME swaps | the matching route test |
| `Content-Length` off by one | 2 model-streaming tests |
| drop each of the 5 HTML escapes | one named per-character test each |
| escape `&` last instead of first | 6 escaping tests |
| `/g` → non-global on `<` | `escapes every occurrence, not just the first` |
| drop `stat` validation at startup | `rejects when the model file does not exist` |
| drop the `/wasm/` prefix guard | `returns null for a path outside /wasm/` |
| remove one action from `VALID_ACTIONS` | `accepts the \"camera\" action` + the size test |
| adapters: swap an action, keep the ref object, POST → GET, first batch only | the matching adapter test |

Per-case discrimination was **measured, not assumed**, for both loops: `M64` (drop one action) fails exactly the `accepts the \"camera\" action` case; `M65`/`M66`/`M67` each fail a *different* subset of the accepted-origin cases. The loops are not carried by one lucky case.

## Survivors

- **`elements.length === 0 || !elements[0].type` → `!elements[0]?.type`** — equivalent for every array-shaped input. The only observable difference is the error *string* for a `null` element: today a raw `TypeError` message leaks out, the mutant would return the domain message. Both are `400 {ok:false}`. Deliberately not pinned — asserting on a JS internal error text would be a worse test than none. A test does pin that `[null]` yields a 400 with `ok:false` rather than a 500 or a hung socket.
- **the `/wasm/snippets/` 200 path's MIME fallback** — genuine gap, but not reachable hermetically: it needs a real file under `packages/wasm/pkg/snippets/`, which is generated `wasm-pack` output that a test must not create or depend on. The 404 branch of that route *is* covered, including a traversal attempt.

## Hang safety

The server holds SSE responses open forever, so one leaked socket keeps the `node:http` handle referenced and `server.close()` never completes — a failed assertion becomes a run that never exits rather than a run that fails. This was not hypothetical: dropping the startup `stat` made `startViewerServer` resolve instead of reject, and the test leaked the server it had just bound. That mutation **hung the suite** on the first pass. Fixed in the test, and the mutation now dies cleanly in 1 s.

The safeguards:
- Every socket the helpers open is registered and destroyed from an `after` hook that runs even when an assertion throws.
- Every request carries its own 10 s deadline; every socket-touching test carries a 30 s timeout.
- SSE streams are closed from `finally`, never from the happy path.
- The startup-contract test captures and closes the server it would otherwise leak.

## Notes

- CI already runs this package: root `pnpm test` → `turbo test` picks up the package's `test` script. No workflow change needed.
- `pnpm typecheck` excludes test files, so the new tests were typechecked separately against the package's own `tsconfig.json` with `include` extended to `test/**/*` — clean. That caught a wrong `SectionPlane` shape in the draft.
- No changeset: test-only, no published API surface change.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end coverage for viewer server routes, including CORS, static assets, commands, SSE, exports, and error handling.
  * Added integration tests for streaming updates, visibility actions, resets, routing, and transport failures.
  * Added security and edge-case tests for asset paths and HTML escaping.
  * Added reusable HTTP testing utilities for requests, streams, polling, timeouts, and cleanup.

### DIFF
--- a/packages/viewer/test/helpers/http.ts
+++ b/packages/viewer/test/helpers/http.ts
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Minimal HTTP client helpers for the viewer server tests.
+ *
+ * Every socket these helpers open is registered in a module-level set and
+ * every request carries its own deadline. That matters here: the server
+ * under test holds SSE responses open forever, so a single leaked socket
+ * keeps the `node:http` server handle alive and `server.close()` never
+ * completes — the test run would hang instead of failing. `destroyAllSockets`
+ * is the belt-and-braces cleanup that runs from an `after` hook even when an
+ * assertion has already thrown.
+ */
+
+import { request as httpRequest, type IncomingMessage, type ClientRequest } from 'node:http';
+
+/** Per-request deadline. Long enough for a slow CI box, short enough to fail fast. */
+export const REQUEST_TIMEOUT_MS = 10_000;
+
+/** Deadline handed to `it(...)` for anything that touches a socket. */
+export const TEST_TIMEOUT = { timeout: 30_000 } as const;
+
+const openRequests = new Set<ClientRequest>();
+
+export interface Res {
+  status: number;
+  headers: NodeJS.Dict<string | string[]>;
+  body: Buffer;
+}
+
+export interface ReqOptions {
+  method?: string;
+  headers?: Record<string, string>;
+  body?: string;
+}
+
+/**
+ * Issue one request on its own socket (`agent: false`), so closing the
+ * server actually releases the port instead of waiting on a pooled
+ * keep-alive connection.
+ */
+export function req(port: number, path: string, opts: ReqOptions = {}): Promise<Res> {
+  return new Promise((resolvePromise, reject) => {
+    const r = httpRequest(
+      {
+        host: '127.0.0.1',
+        port,
+        path,
+        method: opts.method ?? 'GET',
+        headers: opts.headers,
+        agent: false,
+      },
+      (resp: IncomingMessage) => {
+        const chunks: Buffer[] = [];
+        resp.on('data', (c: Buffer) => chunks.push(c));
+        resp.on('end', () => {
+          openRequests.delete(r);
+          resolvePromise({
+            status: resp.statusCode ?? 0,
+            headers: resp.headers,
+            body: Buffer.concat(chunks),
+          });
+        });
+        resp.on('error', (e) => {
+          openRequests.delete(r);
+          reject(e);
+        });
+      },
+    );
+    openRequests.add(r);
+    r.setTimeout(REQUEST_TIMEOUT_MS, () => {
+      r.destroy(new Error(`request to ${path} timed out after ${REQUEST_TIMEOUT_MS}ms`));
+    });
+    r.on('error', (e) => {
+      openRequests.delete(r);
+      reject(e);
+    });
+    if (opts.body !== undefined) r.write(opts.body);
+    r.end();
+  });
+}
+
+/** `req` plus a JSON-parsed body. */
+export async function json(
+  port: number,
+  path: string,
+  opts?: ReqOptions,
+): Promise<{ status: number; headers: NodeJS.Dict<string | string[]>; json: any }> {
+  const res = await req(port, path, opts);
+  return { status: res.status, headers: res.headers, json: JSON.parse(res.body.toString()) };
+}
+
+/**
+ * Open a streaming request (SSE) and resolve once the response head arrives.
+ * The returned `close` MUST be called from a `finally` — see the module note.
+ */
+export function openStream(
+  port: number,
+  path: string,
+): Promise<{ res: IncomingMessage; close: () => void }> {
+  return new Promise((resolvePromise, reject) => {
+    const r = httpRequest({ host: '127.0.0.1', port, path, agent: false }, (resp) => {
+      resolvePromise({
+        res: resp,
+        close: () => {
+          openRequests.delete(r);
+          r.destroy();
+          resp.destroy();
+        },
+      });
+    });
+    openRequests.add(r);
+    r.setTimeout(REQUEST_TIMEOUT_MS, () => {
+      r.destroy(new Error(`stream ${path} produced no response head in ${REQUEST_TIMEOUT_MS}ms`));
+    });
+    r.on('error', (e) => {
+      openRequests.delete(r);
+      reject(e);
+    });
+    r.end();
+  });
+}
+
+/** Tear down anything still open. Safe to call more than once. */
+export function destroyAllSockets(): void {
+  for (const r of openRequests) r.destroy();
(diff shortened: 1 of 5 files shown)


BIMvoice:
All three threads on `test/server-routes.test.ts` verified against the code and addressed in 0108e818. Suite: 117 → 117 (the traversal fix strengthens an existing test rather than adding one).

---

### C. `:558-578` order-dependent tests — **APPLIED**, CodeRabbit is right

`createdSegments: 2`, `['SEG_A','SEG_B']` and the exported `'SEG_A\nSEG_B'` were the residue of the create tests at `:481` and `:499`. Proof, running each under a single-test filter on the unfixed code:

```
$ npx tsx --test --test-name-pattern="reports the accumulated segment count" …   # pass 0, fail 1
$ npx tsx --test --test-name-pattern="exports the segments newline-separated" …  # pass 0, fail 1
    expected: 'application/octet-stream'   actual: 'application/json; charset=utf-8'
$ npx tsx --test --test-name-pattern="clears the segments and reports" …         # pass 0, fail 1
```

A `seedSegments(...)` helper now clears and re-creates the exact state each test needs (still via two separate `POST /api/create` calls, so "accumulated" is genuinely what the status test pins). All three now pass alone:

```
$ for each of the three patterns: # tests 1  # pass 1  # fail 0
```

Nothing was weakened — each still kills its mutation, now *in isolation*, each an exact-substring replacement with count 1:

| mutation in `src/server.ts` | isolated test | result |
| --- | --- | --- |
| `createdSegments: createdSegments.length` → `Math.min(…, 1)` | status | fail 1 |
| `createdSegments.join('\n')` → `join('')` | export | fail 1 |
| swap `const count = …length` to *after* `…length = 0` | clear | fail 1 |

### D-2. `:222` WHATWG dot-segment normalisation — **APPLIED**, and CodeRabbit is right about the diagnosis

Confirmed at the source. `server.ts:164` routes on `new URL(req.url, …).pathname`, and the WHATWG parser removes double-dot segments:

```
/wasm/snippets/../../../../../../etc/passwd  ->  /etc/passwd
/wasm/snippets/%2e%2e/%2e%2e/etc/passwd      ->  /etc/passwd
/wasm/snippets/.%2e/.%2e/etc/passwd          ->  /etc/passwd
/wasm/snippets/%2e%2e%2f%2e%2e%2fetc/passwd  ->  unchanged   (%2f is not a segment separator)
```

So the literal case never reached the `/wasm/snippets/` branch at `server.ts:220` — it was measuring the unknown-route 404, exactly as reported.

One correction to the suggested fix: `%2e%2e%2f…` does enter the branch, but `resolveWasmAssetPath` never decodes, so `/wasm/snippets/%2e%2e%2f%2e%2e%2fetc/passwd` resolves to a literal filename inside `pkg/` and the 404 comes from the file not existing — not from the guard. That input alone kills nothing. So the added case aims at **`pkg/../package.json`**, a file that genuinely exists one level outside the served root, which makes it a real kill:

```
# decodeURIComponent(path) at the call site + guard replaced with `if (false)` (2 replacements, count 1 each)
old test (literal `..` only):  # pass 1  # fail 0     <- green against a traversable server
new test (with encoded case):  # pass 0  # fail 1     <- expected: 404  actual: 200
```

The comment now describes both layers accurately. The guard's own `null` return stays pinned in `server.test.ts:30-90`; it is unreachable from this route by construction (`url.pathname` always starts `/wasm/snippets/` and can never carry a surviving `..` for `resolve()` to act on), and the test now says that instead of claiming to test it.

### D-1. `:207` wasm artifacts — **PARTLY REFUTED, partly applied**

**Refuted:** "the viewer `test` script does not run a build step" is true of the package script but not of how these tests run. `.github/workflows/test.yml:214` runs `bash scripts/build-wasm.sh`, or `:174` fetches the prebuilt bundle, both before the test job — the wasm runtime is always present and freshly built. And the tests do not "silently assume" the binaries: without them all six suites hard-fail in `before()` (43/117 pass). There is no silent-pass path, and nothing here should be gated behind an opt-in — skipping is what would let a broken wasm build quietly drop this coverage.

**Applied:** the *failure mode* was genuinely bad. It was six bare `ENOENT … node_modules/@ifc-lite/wasm/pkg/ifc-lite.js` lines, which read like a broken install rather than a missing build step. A file-level preflight now fails fast and names the fix:

```
error: 'Missing wasm runtime …/packages/wasm/pkg/ifc-lite_bg.wasm. These tests serve the real
        @ifc-lite/wasm artifacts over HTTP, and the runtime is gitignored.
        Build it first: bash scripts/build-wasm.sh'
```

It throws rather than skips, deliberately.

`test/server-routes.test.ts` typechecks clean against the package's own `tsconfig.json` with the `include` widened to `test/**/*` (`pnpm typecheck` does not cover test files, so this was run separately). The only two errors in that run are pre-existing, in files not touched here, and are unbuilt-workspace-dep noise (`Cannot find module '@ifc-lite/sdk'` — `packages/sdk/dist` absent).

PR #2155, opened 2026-08-05
fix(source-dalux): four review follow-ups for #2023 (first-poll baseline, transient-failure deletions, folder parent paging, package index)

BIMvoice:
Four of the unresolved review threads on #2023, implemented against that PR's head (`cf2d816`) and targeting `feat/source-dalux-provider`, not `main`.

**Offered for you to take or leave.** It deliberately does not touch your branch, does not resolve any of your threads, and nothing was posted on #2023. If you'd rather fix any of these differently, ignore the corresponding commit hunk — the four fixes are independent.

Everything stays inside `packages/source-dalux` plus the generated docs region. Where a fix looked like it wanted a host change, I stopped and wrote down why instead (see #4).

Baseline `packages/source-dalux`: **77 tests**. After: **83 tests** (5 new + 1 that was already green as a guard). `pnpm typecheck`, `pnpm lint`, `pnpm check:test-wiring`, `pnpm docs:check-readmes` all pass. Test files aren't covered by the root typecheck (`include: ["src/**/*"]`), so they were typechecked separately against the package's own tsconfig with that exclusion lifted — clean.

---

### 1. Stale published-package index

`pnpm docs:check-generated` exited **1** on this branch and **0** on its base, so the staleness originates here: `docs/api/typescript.md`'s `package-index` region has no `@ifc-lite/source-dalux` row. Ran `pnpm docs:generate`; only the generated region is committed (one added table row), nothing hand-edited. `pnpm docs:check-generated` now exits **0**.

### 2. First poll compared against nothing (`provider.ts`, `watchRevisions`)

The upstream revision was compared only against the provider cache. Before the first successful poll there is no cache entry, so `cached && cached !== latest` is false, **no event is emitted**, and the newer revision is then written to the cache — the update is suppressed permanently, not just delayed. Every existing test seeded the cache first, which is exactly why this was invisible.

`ref.revisionId` is the revision the host is currently holding, produced by the same `currentRevisionId()` and forwarded by `revisionWatch.ts`, so it is directly comparable. It is now the baseline when the cache is empty.

- RED: `emits an event on the very first poll when upstream has moved past the ref revision (empty cache)` — failed on current code (`expected [] to deeply equal [ { fileId: 'f1', … } ]`).
- Mutation (teeth): `const baseline = (await ctx.storage.get(cacheKey)) ?? nonEmptyString(ref.revisionId);` → `const baseline = await ctx.storage.get(cacheKey);` (1 replacement) → that test fails, the companion guard test still passes.
- Mutation for the guard test: `if (baseline && baseline !== latestRevisionId)` → `if (baseline)` (1 replacement) → `stays silent on the first poll when upstream still matches the ref revision` fails, along with the two pre-existing "must not falsely emit" tests.

### 3. A transient failure was reported as deletion (`provider.ts`, `watchRevisions`)

The `try` covers `fetchAllPages` plus decode, so a 500, a timeout, an aborted relay and a pagination error all land in the same `catch` as a 404 — and the catch emitted `deleted: true` for every tracked ref in that area. `RevisionEvent.deleted` is documented as "the file is gone upstream" and the host surfaces it as "N source files are gone upstream", so this is a contract violation rather than a UX wart.

Only a definitive not-found (`DaluxHttpError` with status 404) now produces deletions; anything else skips that area's refs for this poll — which is what the catch's own comment already said it did. Cache handling is untouched (`continue` still buffers nothing), so the change is confined to the event kind. The class doc comment above `watchRevisions` was updated to match.

- RED: `does NOT report deletion when an area sweep fails transiently (500)…` and `…when the sweep fails for a non-HTTP reason (aborted/network)` — both failed on current code with the spurious `deleted: true` event.
- Both directions covered: `reports deleted for a definitive 404 on the file-area sweep` (green before and after — the 404 path must not regress) and the two transient tests.
- Mutations: `const gone = err instanceof DaluxHttpError && err.status === 404;` → `const gone = false;` (1 replacement) → only the 404 test fails. → `const gone = true;` (1 replacement, i.e. the pre-fix behaviour) → only the two transient tests fail.

### 4. Parent ids were resolved against one page (`provider.ts`, `listContainers`)

`folderIdsThisPage` was built from a single Dalux page, so a folder whose parent lands on a different page was rewritten to the file-area root. The comment claimed the host "re-nests correctly once the host has walked that later page and revisits it" — it does not: `useSourceCatalogSync.ts` merges `items: [...current.items, ...page.items]` on both load-more paths, a pure append with no re-resolution, and the result persists via `maybePersist`. So the misplacement is permanent. That comment is gone.

**On the shape of the fix.** Literally preserving the reported parent id and leaving it dangling would make those folders *disappear* rather than be misplaced: `buildContainerTree` does `byId.get(c.parentId)?.children.push(node)`, so a container whose parent id never arrives is dropped from the tree entirely, and `SourceFolderStep` filters by exact `parentId` too. Making dangling parents fall back to the root is a host change, which I was asked not to make on your branch — and the provider also genuinely needs the root fallback for the case your comment describes (Dalux reporting a parent that is never itself surfaced as a folder; covered by your existing test, still green).

So instead the folder listing resolves against the **complete** folder set: it sweeps every page via `fetchAllPages` and returns one complete page with no cursor. Parent ids are then preserved for every real parent, and "parent not in the area" now genuinely means unreachable instead of "not on this page".

This does reverse your explicit "one giant eager fetch" note, so it is the hunk most worth disagreeing with. The argument for it: nesting is not decidable from a single page, the endpoint already returns every folder in the area regardless of nesting (`containerListing: 'flat-subtree'`), folders are the small dimension while files — the volume — stay paged one call at a time, the sweep is bounded by `MAX_SWEEP_PAGES`, and a complete folder page is what lets `maybePersist` treat the catalog as fully fetched. The `fetchAllPages` doc comment was updated so its stated invariant still matches reality. If you'd rather keep folder paging, this is the one to drop.

- RED: `keeps a folder nested under a parent that only appears on a later Dalux page` — failed on current code (child rewritten to the file-area root, second page never fetched).
- Mutation: the `fetchAllPages(...)` call reverted to the previous single `fetchPage(...)` (1 replacement) → only that test fails.

Every mutation was applied as an exact-substring replacement asserting exactly one occurrence, the mutated region re-read to confirm the text changed, and reverted by restoring a saved copy of the file.

No changeset: the package is unreleased and `.changeset/source-dalux-initial.md` already covers this release.

### DIFF
--- a/docs/api/typescript.md
+++ b/docs/api/typescript.md
@@ -47,6 +47,7 @@ ifc-lite ships 36 public npm packages: 35 scoped `@ifc-lite/*` packages plus the
 | [`create-ifc-lite`](#create-ifc-lite) | Create IFC-Lite projects with one command |
 | [`@ifc-lite/merge`](https://www.npmjs.com/package/@ifc-lite/merge) | Three-way merge engine for IFCX layers — MergePlan with auto-merged ops and explicit conflict records, merge-layer emission, rebase, and revert. |
 | [`@ifc-lite/plugin-api`](https://www.npmjs.com/package/@ifc-lite/plugin-api) | Dependency-free type surface for ifc-lite file-source plugins |
+| [`@ifc-lite/source-dalux`](https://www.npmjs.com/package/@ifc-lite/source-dalux) | Dalux Build (Box) file-source provider for ifc-lite |
 <!-- END GENERATED: package-index -->
 
 ---
--- a/packages/source-dalux/src/http-client.ts
+++ b/packages/source-dalux/src/http-client.ts
@@ -277,10 +277,12 @@ export async function fetchPage(
 
 /**
  * Follows every page of a Dalux bookmark-paginated endpoint and returns the
- * combined items. For internal callers that genuinely need a whole listing
- * in one shot — currently only `watchRevisions`'s one-sweep-per-file-area
- * poll. The public `FileSourceProvider` listing methods must NOT use this;
- * they call {@link fetchPage} once and return that page to the host.
+ * combined items. For callers that genuinely need a whole listing in one
+ * shot: `watchRevisions`'s one-sweep-per-file-area poll, and the folder
+ * listing in `listContainers` (whose parent-id resolution is only correct
+ * against the complete folder set). `listProjects`/`listFiles` must NOT use
+ * this — they call {@link fetchPage} once and return that page to the host,
+ * so a large tenant or file area is never loaded eagerly.
  *
  * Bounded by `MAX_SWEEP_PAGES` (a server minting endlessly fresh bookmarks
  * can't loop forever) and tracks every bookmark seen across the whole
--- a/packages/source-dalux/src/provider.ts
+++ b/packages/source-dalux/src/provider.ts
@@ -103,28 +103,35 @@ export class DaluxBuildProvider implements FileSourceProvider {
     // 'flat-subtree'); the host nests the flattened result client-side via
     // each folder's `parentId`.
     const fileAreaId = decodeContainerId(parentId).fileAreaId;
-    const page = await fetchPage(
+    // Deliberately the one listing path that sweeps every page before
+    // returning. Nesting is only decidable against the *complete* folder set:
+    // a folder's parent can land on any page, and the host appends later
+    // pages to what it already has (`useSourceCatalogSync` merges
+    // `[...current.items, ...page.items]`) without ever re-resolving earlier
+    // ones — so a parent id resolved against a single page is resolved
+    // permanently, and wrongly. The sweep is bounded by `MAX_SWEEP_PAGES` and
+    // covers folders only; files (the volume) stay paged one call at a time.
+    const rawFolders = await fetchAllPages(
       client,
       `/5.1/projects/${enc(projectId)}/file_areas/${enc(fileAreaId)}/folders`,
       {},
-      options?.cursor,
       options?.signal,
     );
-    const folders = convertListLenient<DaluxFolder>(ctx, page.items, decodeFolder, 'Folder');
-    const folderIdsThisPage = new Set(folders.map((folder) => folder.folderId));
+    const folders = convertListLenient<DaluxFolder>(ctx, rawFolders, decodeFolder, 'Folder');
+    const folderIds = new Set(folders.map((folder) => folder.folderId));
 
     const containers: SourceContainer[] = folders.map((folder) => {
       const parentFolderId = nonEmptyString(folder.parentFolderId);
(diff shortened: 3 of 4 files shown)


BIMvoice:
Independently verified the substance of this before relaying it, including the one place it deliberately departs from what the review thread asked for.

**Codex's diagnosis on thread 4 is right, but its proposed remedy would make things worse.** The thread says to "preserve the reported parent ID and let the host resolve it once the parent page arrives". The host has no such resolution pass, and an unresolvable parent does not degrade to root — it drops the folder entirely:

`apps/viewer/src/components/sources/SourceFolderTree.tsx:20-29`

```ts
for (const c of containers) {
  if (c.id === rootId) continue;
  const node = byId.get(c.id)!;
  if (c.parentId === rootId) {
    roots.push(node);
  } else if (c.parentId) {
    byId.get(c.parentId)?.children.push(node);   // <- parent absent: silent no-op
  }
}
```

If `byId` has no entry for `c.parentId`, the optional chain no-ops, and because `c.parentId !== rootId` the node never reaches `roots` either. So following the suggestion literally converts a **misplaced** folder into a **missing** one. That is why this PR resolves the parent set instead — `fetchAllPages` for `/folders`, so `folderIds` covers every page and a real parent is always resolvable.

**That is a design change on your PR and I want it flagged, not buried:** it reverses your "no giant eager fetch" choice for folders specifically. Files are untouched. If you would rather keep the lazy fetch, this is the hunk to drop — the other three fixes are independent of it, and the root-fallback path still survives for a genuinely unsurfaced parent (your existing test for it stays green).

**Verified myself on the branch:**

```
83/83 pass (baseline 77)

### drop `?? nonEmptyString(ref.revisionId)` from the baseline   (replacements=1)
     × emits an event on the very first poll when upstream has moved past
       the ref revision (empty cache)
      Tests  1 failed | 82 passed (83)
```

Restored: 83/83. `pnpm docs:check-generated` goes 1 -> 0.

One correction against myself, since it nearly became a false report: my first run showed `Test Files 1 failed | 3 passed` with `Tests 47 passed` — which reads like a pass with a stray file error. It was neither a defect nor a flake: `@ifc-lite/plugin-api` simply was not built in my fresh worktree, so `provider.test.ts` failed to resolve its import and contributed **zero** tests while the summary line still said everything passed. After `turbo build --filter=@ifc-lite/source-dalux^...` it is 83/83. Worth knowing that this package's suite reports that shape when its dependency is unbuilt.

Offered for you to take or leave — it deliberately does not touch `feat/source-dalux-provider` directly.

PR #2156, opened 2026-08-05
test(server): close 15 mutation-verified coverage gaps in apps/server

BIMvoice:
## Why

`apps/server` is the binary `packages/server-bin` downloads, verifies and executes — the actual server users run — and it had never been mutation-swept, unlike the TypeScript packages and the `rust/core`, `rust/wasm-bindings`, `rust/processing` and `rust/geometry` crates.

## Method

Toolchain: the root pin, `nightly-2025-11-15` (rustc 1.93.0-nightly). Baseline `cargo test -p ifc-lite-server`: **83 passed, 0 failed, 0 ignored**.

20 mutations, each applied by exact-substring replacement with an asserted replacement count of 1 and the mutated region re-read before any result was believed, then restored from a saved backup copy (never `git checkout --`). **Two deliberate controls, run first, both KILLED** — so the suite was known to be live before any survivor was trusted.

**Kill ratio: 5/20 (25%). TARGETED** — mutations were aimed at the weakest-looking areas (auth, config, threshold boundaries, error mapping), not sampled uniformly, so this ratio is not an estimate of crate-wide coverage.

| # | File | Mutation | Result |
|---|---|---|---|
| C1 | `mem_policy.rs` | `BUDGET_PCT_OF_CEILING` 70 → 80 | **KILLED** (control) — `uses_tightest_readable_ceiling`, `physical_ram_bounds_a_bare_vm_with_no_cgroup` |
| C2 | `admission.rs` | CPU semaphore permits `+ 1` | **KILLED** (control) — `rejects_with_overloaded_when_cpu_saturated` + 2 |
| M1 | `middleware/auth.rs` | match guard `\|\| true` (any token authenticates) | SURVIVED |
| M2 | `middleware/auth.rs` | `constant_time_eq` length mismatch `return true` | SURVIVED |
| M3 | `middleware/auth.rs` | `strip_prefix("Bearer ")` → `"bearer "` | SURVIVED |
| M4 | `middleware/auth.rs` | `UNAUTHORIZED` → `FORBIDDEN` | SURVIVED |
| M5 | `middleware/auth.rs` | auth-disabled pass-through → 401 | KILLED — 8 `parity_tests` |
| M6 | `config.rs` | port default 8080 → 9090 | SURVIVED |
| M7 | `config.rs` | drop `filter(\|s\| !s.is_empty())` on `api_token` | SURVIVED |
| M8 | `config.rs` | drop the `IFC_METRICS_ENABLED=1` spelling | SURVIVED |
| M9 | `config.rs` | drop `.min(100)` on `IFC_MEM_SHED_PCT` | SURVIVED |
| M10 | `admission.rs` | shed `Retry-After` `* 2` → `* 1` | SURVIVED |
| M11 | `admission.rs` | `acquire`: `rss >= watermark` → `>` | SURVIVED |
| M12 | `admission.rs` | `rejected_shed` → `rejected_overload` | SURVIVED |
| M13 | `admission.rs` | `is_shedding`: `>=` → `>` | SURVIVED |
| M14 | `error.rs` | `FileTooLarge` 413 → 400 | SURVIVED |
| M15 | `routes/parse/mod.rs` | upload guard `>` → `>=` | SURVIVED |
| M16 | `routes/parse/mod.rs` | disable the gzip decompressed-size guard | SURVIVED |
| M17 | `main.rs` | `memory_admission_log_level` `Some(0)` Info → Warn | KILLED — `explicit_zero_is_opt_out_info` |
| M18 | `main.rs` | CORS `o == "*"` → `o == "**"` | SURVIVED |
| M19 | `routes/parse/cache_keys.rs` | `request_cache_key` drops the quality segment | SURVIVED |
| M20 | `error.rs` | `Overloaded` 503 → 500 | KILLED — `saturated_admission_returns_503_with_retry_after` |

Every one of the 15 survivors was **replayed against the new suite and confirmed KILLED** (killing tests listed below).

### `middleware/auth.rs` — no tests at all (M1–M4)

The optional bearer-token layer that protects every compute/parse/cache route had zero coverage. Making `constant_time_eq` unconditionally true left all 83 tests green: an operator who sets `IFC_SERVER_API_TOKEN`, sees the startup log say auth is ENABLED, and believes the compute routes are protected would in fact be serving them to anyone. The only direction the suite pinned was "no token configured ⇒ pass through" (M5, killed incidentally by `parity_tests`) — precisely the direction that protects nothing.

New `middleware/auth_tests.rs` drives the real middleware in-process and pins both directions of the configured-token signal, the length-mismatch early return (M2 — inverting it authenticates every token of a *different* length, including the empty one), the case-sensitive scheme prefix (M3), credential trimming, and 401-not-403 (M4).

Killed by: `configured_token_admits_the_match_and_rejects_everything_else`, `length_mismatch_is_rejected_not_short_circuited_to_success`, `constant_time_eq_pins_both_outcomes`, `only_the_bearer_scheme_prefix_is_accepted`, `rejection_is_401_and_no_token_configured_passes_through`.

### `config.rs` — no tests at all (M6–M9)

Notably **M7**: without the `!is_empty()` filter, `IFC_SERVER_API_TOKEN=""` — a common way to "unset" a variable in a compose file or a CI secret — flips `api_token` to `Some("")`, so the server starts *requiring* an `Authorization: Bearer ` header whose secret is the empty string. Every real client 401s; anyone sending the empty credential is admitted. And **M9**: an unclamped `IFC_MEM_SHED_PCT=200` puts the RSS watermark at twice the budget, so the circuit breaker never fires and the container is OOM-killed instead of shedding.

New `config_tests.rs` covers the shipped defaults, every scalar override plus its garbage-fallback, the concurrency `max(1)` clamp, the shed-pct clamp at the 100 boundary in both directions, the metrics flag's accepted *and* rejected spellings, token precedence/trimming/blank handling, CORS splitting, `Debug` redaction of the token, and the mem-budget passthrough.

Killed by: `defaults_apply_when_nothing_is_set`, `scalar_overrides_are_honoured_and_garbage_falls_back`, `api_token_precedence_trimming_and_blank_handling`, `blank_primary_token_shadows_the_fallback_and_leaves_auth_off`, `metrics_flag_accepts_one_and_true_only`, `shed_pct_is_clamped_to_100`.

### `admission.rs` — thresholds and observability (M10–M13)

The existing tests assert only `matches!(err, ApiError::Overloaded { .. })`, so the `Retry-After` each path advertises was unpinned, and they sit 5 MB clear of the RSS watermark, so `>=` could be relaxed to `>` with the suite green. No test read a rejection counter, so all three reasons could be attributed to one bucket — and those counters are exactly what an operator reads to tell a memory 503 from a CPU one from a queue one, three different remediations.

New `admission_tests.rs` pins per-reason counter attribution, the doubled shed `Retry-After` against the single-length one (and that it tracks the configured timeout and is never 0), and the watermark at `watermark - 1` / `watermark` / `watermark + 1` — 89 128 960 bytes exactly, so the boundary is not a rounding artifact. Also the inert-breaker cases (unsampled RSS, no budget, `shed_pct = 0`), gauge round-tripping, and that a queued-then-rejected waiter does not leak its queue slot.

Killed by: `each_rejection_reason_increments_its_own_counter`, `shed_advertises_double_the_retry_after_of_the_other_paths`, `rss_breaker_fires_exactly_at_the_watermark`.

### `error.rs` — the status mapping (M14)

Only `Overloaded` was pinned, incidentally. Remapping an oversized upload from `413` to `400` was invisible — even though the streaming size guard exists specifically so a client sees a clean `413` rather than a generic multipart `400`. New `error_tests.rs` walks the whole variant table (status, code and body message per row), pins `Retry-After` as `Overloaded`-only in both directions, and covers the `From` conversions.

Killed by: `every_variant_maps_to_its_documented_status_and_code`, `file_too_large_is_413_and_names_the_limit`.

### `extract_file` — boundary and gzip bomb (M15, M16)

The existing tests sat 512 KB under a 1 MB ceiling and 512 KB over it, so tightening `>` to `>=` — which rejects every upload landing exactly on the advertised limit, with a 413 naming that very limit — survived. And nothing exercised the gzip branch at all: disabling the decompressed-size bound lets a ~4 KB upload expand to hundreds of MB, which the admission byte budget cannot catch because it reserves against the *compressed* size.

Killed by: `a_file_of_exactly_the_ceiling_is_accepted`, `a_gzip_bomb_is_rejected_on_its_decompressed_size`. Also added: the within-ceiling gzip round-trip (so the bomb guard is not just rejecting everything gzipped), a zero ceiling, and the missing-`file`-field case.

### CORS (M18)

`build_cors_layer` had no tests. The mutation broke the wildcard in its least dangerous direction, but the same blind spot covers the dangerous one: if the allow-list branch ever degraded into a permissive one, every deployment with an explicit `CORS_ORIGINS` would start serving any origin, unnoticed. New `cors_tests.rs` drives the real router and asserts an unlisted origin gets *no* `Access-Control-Allow-Origin` header, near-miss spellings (`**`, `*.example.com`) are not treated as the wildcard, and the shipped default is not permissive.

Killed by: `a_wildcard_entry_enables_permissive_cors`, `near_miss_wildcards_do_not_enable_permissive_cors`.

### `request_cache_key` (M19)

The seed for every other key (parquet, parquet metadata, JSON response, symbolic sidecar) and the identifier handed to the client — and it had no test. Dropping its quality segment left the suite green, which would make a `?tessellation_quality=high` request read back the entry a previous `medium` request wrote: the client silently gets a coarser mesh than it asked for, and no cache bust ever fixes it. The new test checks all 15 (filter × quality) pairs for **pairwise** distinctness rather than inferring discrimination from a loop.

Killed by: `request_cache_key_separates_content_filter_and_quality`.

## Production change

One, behaviour-preserving: `Config::from_env` now delegates to a private `from_lookup(get: impl Fn(&str) -> Option<String>)`, and `from_env` passes the real `std::env::var`. This is required because `parity_tests::test_state` calls `from_env()` concurrently in the same test binary — a config test that set `IFC_SERVER_API_TOKEN` via `set_var` would silently switch those tests to an authenticated server. No default, clamp or precedence rule changed.

## Notes

- New tests live in sibling `*_tests.rs` files (the established `#[path]` pattern, as in `services/parquet.rs`) so no production module crosses the 400-line ratchet. `cargo test -p ifc-lite-processing --test module_size_ratchet` is green.
- `mem_policy::physical_memory_is_readable_on_linux` is `#[cfg(target_os = "linux")]`, so it silently does not exist on a macOS dev box. Not a failure — it does run in CI — but worth knowing that a local green is 1 test lighter than CI's.
- `cargo fmt --check` reports diffs across ~22 files in this crate at baseline, including files this PR does not touch, so the crate is not rustfmt-clean upstream and nothing here was reformatted.

## Verification

- Baseline: `cargo test -p ifc-lite-server` → 83 passed, 0 failed, 0 ignored.
- Final: **126 passed, 0 failed, 0 ignored**.
- `cargo clippy -p ifc-lite-server --all-targets`: clean on the pinned toolchain.
- All 15 survivors replayed post-fix: each fails, with the named test.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for admission control, configuration parsing, CORS behavior, authentication, API errors, cache-key isolation, and file-upload limits.
  * Verified boundary conditions including memory budgets, retry timing, authorization failures, decompression limits, missing files, and safe CORS defaults.
  * Confirmed stable error responses, metrics behavior, queue cleanup, and backward-compatible cache keys.

### DIFF
--- a/apps/server/src/admission.rs
+++ b/apps/server/src/admission.rs
@@ -388,3 +388,7 @@ mod tests {
         assert!(text.contains("reason=\"rss_shed\"} 0"));
     }
 }
+
+#[cfg(test)]
+#[path = "admission_tests.rs"]
+mod admission_tests;
--- a/apps/server/src/admission_tests.rs
+++ b/apps/server/src/admission_tests.rs
@@ -1,0 +1,232 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Second test module for `admission`: the parts the in-file `tests` module
+//! left vacuous.
+//!
+//! The in-file tests assert only `matches!(err, ApiError::Overloaded { .. })`,
+//! so the `retry_after_secs` each rejection path advertises was unpinned; they
+//! also never touch the exact RSS watermark, so `>=` could be relaxed to `>`
+//! with the suite green; and they never read a rejection counter, so all three
+//! reasons could be attributed to the same bucket.
+
+use super::*;
+
+fn cfg(max_parses: usize, budget_mb: u64, queue_depth: usize, shed_pct: u8) -> AdmissionCfg {
+    AdmissionCfg {
+        max_concurrent_parses: max_parses,
+        mem_budget_bytes: budget_mb * 1024 * 1024,
+        queue_depth,
+        queue_timeout: std::time::Duration::from_millis(50),
+        shed_pct,
+    }
+}
+
+/// Pull one `ifc_server_admission_rejected_total{reason="…"}` sample out of the
+/// Prometheus text body.
+fn rejected(a: &Admission, reason: &str) -> u64 {
+    let needle = format!("ifc_server_admission_rejected_total{{reason=\"{reason}\"}} ");
+    let text = a.metrics_text();
+    let line = text
+        .lines()
+        .find(|l| l.starts_with(&needle))
+        .unwrap_or_else(|| panic!("no sample for reason={reason:?} in:\n{text}"));
+    line[needle.len()..].trim().parse().expect("counter value")
+}
+
+/// Every rejection is attributed to its OWN reason. Swapping any two of these
+/// increments left the whole suite green, and an operator debugging a 503 storm
+/// reads exactly these three counters to tell "the box is out of memory" from
+/// "the box is out of CPU" from "the queue is saturated" — three different
+/// remediations.
+#[tokio::test]
+async fn each_rejection_reason_increments_its_own_counter() {
+    // rss_shed: breaker open, nothing else touched.
+    let shed = Arc::new(Admission::new(cfg(4, 100, 4, 85)));
+    shed.set_resident_bytes(90 * 1024 * 1024);
+    shed.acquire(1).await.expect_err("breaker sheds");
+    assert_eq!(rejected(&shed, "rss_shed"), 1);
+    assert_eq!(rejected(&shed, "queue_full"), 0);
+    assert_eq!(rejected(&shed, "overloaded"), 0);
+
+    // queue_full: CPU busy and no queue slot ⇒ immediate reject.
+    let full = Arc::new(Admission::new(cfg(1, 0, 0, 0)));
+    let _held = full.acquire(1).await.expect("takes the only slot");
+    full.acquire(1).await.expect_err("queue full");
+    assert_eq!(rejected(&full, "queue_full"), 1);
+    assert_eq!(rejected(&full, "rss_shed"), 0);
+    assert_eq!(rejected(&full, "overloaded"), 0);
+
+    // overloaded: a queue slot exists, so the request WAITS and then times out.
+    let slow = Arc::new(Admission::new(cfg(1, 0, 4, 0)));
+    let _held = slow.acquire(1).await.expect("takes the only slot");
+    slow.acquire(1).await.expect_err("waits then times out");
+    assert_eq!(rejected(&slow, "overloaded"), 1);
+    assert_eq!(rejected(&slow, "rss_shed"), 0);
+    assert_eq!(rejected(&slow, "queue_full"), 0);
+}
+
+/// The memory-budget timeout is its own path (CPU permit acquired, memory
+/// permits unavailable) and must be counted as `overloaded`.
+#[tokio::test]
+async fn memory_budget_timeout_counts_as_overloaded() {
+    let a = Arc::new(Admission::new(cfg(8, 100, 4, 0)));
+    let _g1 = a.acquire(40 * 1024 * 1024).await.expect("40MB #1");
+    let _g2 = a.acquire(40 * 1024 * 1024).await.expect("40MB #2");
+    a.acquire(40 * 1024 * 1024).await.expect_err("over budget");
+    assert_eq!(rejected(&a, "overloaded"), 1);
+    assert_eq!(rejected(&a, "queue_full"), 0);
+    assert_eq!(rejected(&a, "rss_shed"), 0);
+}
+
+/// `Retry-After` is DOUBLED on the RSS-breaker path and single on every other
+/// path. That is the whole point of the breaker: RSS does not fall the instant
+/// a slot frees, so a client that retries after the plain queue timeout just
+/// bounces off a still-open breaker. Halving the shed value survived the suite
+/// because no test read the number.
+#[tokio::test]
+async fn shed_advertises_double_the_retry_after_of_the_other_paths() {
+    // queue_timeout 50 ms ⇒ `.as_secs()` is 0 ⇒ `.max(1)` ⇒ 1 s baseline.
+    let queue_full = Arc::new(Admission::new(cfg(1, 0, 0, 0)));
+    let _held = queue_full.acquire(1).await.expect("takes the slot");
+    let ApiError::Overloaded { retry_after_secs: queue_secs } =
+        queue_full.acquire(1).await.expect_err("queue full")
+    else {
+        panic!("expected Overloaded");
+    };
+    assert_eq!(queue_secs, 1, "queue rejection advertises the bare timeout");
+
+    let shedding = Arc::new(Admission::new(cfg(4, 100, 4, 85)));
+    shedding.set_resident_bytes(90 * 1024 * 1024);
+    let ApiError::Overloaded { retry_after_secs: shed_secs } =
+        shedding.acquire(1).await.expect_err("breaker sheds")
+    else {
+        panic!("expected Overloaded");
+    };
+    assert_eq!(shed_secs, 2, "the RSS breaker backs clients off twice as long");
+    assert_eq!(shed_secs, queue_secs * 2);
+}
+
+/// The retry hint scales with the configured timeout rather than being a
+/// hard-coded constant, and never advertises `0` (a client would hot-loop).
+#[tokio::test]
+async fn retry_after_tracks_the_configured_timeout_and_is_never_zero() {
+    let mut c = cfg(1, 0, 0, 0);
+    c.queue_timeout = std::time::Duration::from_secs(7);
+    let a = Arc::new(Admission::new(c));
+    let _held = a.acquire(1).await.expect("takes the slot");
+    let ApiError::Overloaded { retry_after_secs } = a.acquire(1).await.expect_err("queue full")
+    else {
+        panic!("expected Overloaded");
+    };
+    assert_eq!(retry_after_secs, 7);
+
+    let mut z = cfg(1, 0, 0, 0);
+    z.queue_timeout = std::time::Duration::ZERO;
+    let a = Arc::new(Admission::new(z));
+    let _held = a.acquire(1).await.expect("takes the slot");
+    let ApiError::Overloaded { retry_after_secs } = a.acquire(1).await.expect_err("queue full")
+    else {
+        panic!("expected Overloaded");
+    };
+    assert_eq!(retry_after_secs, 1, "a sub-second timeout still backs the client off 1s");
+}
+
+/// The RSS watermark comparison is `>=`, and it is exercised AT the exact
+/// boundary in both directions — the existing test sits 5 MB clear of it, so
+/// relaxing `>=` to `>` changed nothing observable.
+///
+/// `watermark = mem_budget_bytes / 100 * shed_pct`; with a 100 MiB budget and
+/// 85% that is exactly `1_048_576 * 85 = 89_128_960` bytes (the integer
+/// division is exact here, so the boundary is not an artifact of rounding).
+#[tokio::test]
+async fn rss_breaker_fires_exactly_at_the_watermark() {
+    const BUDGET: u64 = 100 * 1024 * 1024;
+    let watermark = BUDGET / 100 * 85;
+    assert_eq!(watermark, 89_128_960, "boundary arithmetic is exact");
+
+    let a = Arc::new(Admission::new(cfg(4, 100, 4, 85)));
+
+    // One byte below: admits, and readiness says not shedding.
+    a.set_resident_bytes(watermark - 1);
+    assert!(!a.is_shedding(), "below the watermark the instance stays ready");
+    let g = a.acquire(1).await.expect("admits one byte below the watermark");
+    drop(g);
+
+    // Exactly at the watermark: sheds. This is the case `>` would miss.
+    a.set_resident_bytes(watermark);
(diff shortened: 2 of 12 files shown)


BIMvoice:
Reproduced the headline finding on `upstream/main` before relaying it, because it is the one on this list that would matter operationally.

`apps/server/src/middleware/auth.rs` has **zero** `#[cfg(test)]` modules on main. Making the comparison unconditionally accept:

```rust
fn constant_time_eq(a: &[u8], b: &[u8]) -> bool {
    return true;                      // replacements=1
    #[allow(unreachable_code)]
    if a.len() != b.len() { return false; }
    ...
}
```

```
cargo test -p ifc-lite-server
test result: ok. 83 passed; 0 failed; 0 ignored
```

Restored: 83 passed. So on main, every protected compute/parse/cache route could be made to accept **any** bearer token — or a wrong-length one — with the entire crate suite green.

**Why this is worse than an ordinary coverage gap.** The only direction that was pinned is "no token configured ⇒ pass through", which is the case that protects nothing. Nothing pinned the case that *does*: token configured, wrong credential presented. An operator who sets `IFC_SERVER_API_TOKEN` and sees the "auth ENABLED" line at startup has no test standing behind that promise. `apps/server` is also the binary `packages/server-bin` downloads, checksum-verifies and executes, so this is the server people actually run rather than a library surface.

I want to be precise about what this is and is not: **the production code is correct.** `constant_time_eq` is a proper length-check-then-XOR comparison, the `Bearer ` prefix match is exact, and 401 is the right status. Nothing here is a live vulnerability and nothing in this PR changes auth behaviour. What was missing was anything holding it in place — the same shape as `embed-sdk` in #2149, where the origin checks were right but deletable.

Two of the config findings are worth separating out because they bite on *deployment* rather than code:

- **`IFC_SERVER_API_TOKEN=""`** — without the empty-filter, blanking the variable (the ordinary way to disable it in a compose file or CI) does not disable auth; it sets the expected secret to the empty string. Real clients get 401, and anyone sending an empty credential is admitted. That inverts the operator's intent in both directions at once.
- **`IFC_MEM_SHED_PCT=200`** — unclamped, the RSS watermark lands at twice the budget, so the breaker never fires and the container is OOM-killed instead of shedding. A misconfiguration silently removes the protection it was meant to tune.

The 25% kill ratio being flagged as targeted rather than uniform is the right call — auth, config, thresholds and error mapping were chosen deliberately, so it is not a crate-wide estimate.

Two things I would not want lost in the table:

- **`mem_policy.rs:199-205`** — `physical_memory_is_readable_on_linux` is `#[cfg(target_os = "linux")]`, so it does not exist on a macOS dev box. Local green is one test lighter than CI green. That is this crate's silently-skipped case, and there are no `#[ignore]`s.
- **`config.rs:173-177`** — the `IFC_SERVER_API_TOKEN` -> `API_TOKEN` fallback keys on *presence*, so a set-but-blank primary shadows a valid fallback and quietly leaves auth off. Current behaviour is pinned rather than changed; if you would rather blank fall through to the fallback, that is a one-line precedence change and your call.

BIMvoice:
Both CodeRabbit findings verified against the code and **applied** in a23342ca. Suite: 126 → 126 (no new test functions — both findings were about existing tests measuring the wrong thing).

---

### A. `cors_tests.rs` reads the runner's environment — **APPLIED**, and CodeRabbit is right

This was the real one. `state_with_origins` built its `Config` via `Config::from_env()`, so `the_shipped_default_is_not_permissive` asserted whatever the runner exported, not what we ship.

`Config::from_lookup` is now `pub(crate)` (it was added by this PR precisely to make config injectable, so it was the right seam), and the fixture is `Config::from_lookup(|_| None)` — the shipped defaults by construction, with no environment read at all. Environment *parsing* coverage stays in `config_tests.rs:185-203`, where the lookup is already injected per case.

Direct proof of the defect, by running with a permissive value exported:

```
$ CORS_ORIGINS='*' cargo test -p ifc-lite-server cors_tests     # after the fix
test cors_tests::the_shipped_default_is_not_permissive ... ok
test result: ok. 4 passed; 0 failed

# then, reverting only that one line back to Config::from_env() (replacement count: 1):
$ CORS_ORIGINS='*' cargo test -p ifc-lite-server cors_tests
assertion `left != right` failed: the default CORS configuration must not be permissive: ["*"]
test result: FAILED. 3 passed; 1 failed
```

The test still kills what it was written for — mutating `build_cors_layer` (`main.rs:79`) from `o == "*"` to `o.contains("*")` (replacement count: 1) fails `near_miss_wildcards_do_not_enable_permissive_cors`.

### B. `error_tests.rs` claims one row per variant but omits two — **APPLIED**, CodeRabbit is right

`ApiError` has 11 variants; the table had 9. Both missing ones are now covered with real values — neither `MultipartError` nor `JoinError` has a public constructor, so the errors are provoked rather than faked: a `multipart/form-data` body that never contains its declared boundary, and a spawned task that panics.

Mutation evidence, each an exact-substring replacement with count 1:

| mutation in `error.rs` | result |
| --- | --- |
| `Multipart(_) => (BAD_REQUEST, …)` → `INTERNAL_SERVER_ERROR` | FAILED — `status for "Multipart error: …" left: 500 right: 400` |
| `Join(_) => (…, "TASK_ERROR")` → `"JOIN_ERROR"` | FAILED — `code for "Join error" left: "JOIN_ERROR" right: "TASK_ERROR"` |

**On your side note about making it exhaustive by construction:** partly possible, so both halves are in.

- A wildcard-free `variant_name` match makes a new variant a *compile* error in this file. Verified by inserting an `ApiError::BrandNew` variant (replacement count: 1): `error[E0004]: non-exhaustive patterns: &error::ApiError::BrandNew not covered --> apps/server/src/error_tests.rs:57:11`.
- The table then asserts it covered `VARIANT_COUNT` **distinct** variants, so an arm added without a row fails at runtime. Verified by deleting the `Multipart` row (replacement count: 1): `the table must have one row per ApiError variant; covered: {…10 names…} left: 10 right: 11`.

What *cannot* be done is generating the table from the type — every row needs a constructed value, and two of the variants wrap foreign error types with no public constructor. So `VARIANT_COUNT` remains the one hand-maintained number, sitting immediately above the match that forces you to look at it. Stating that plainly rather than implying the list is machine-derived.

`cargo clippy -p ifc-lite-server --all-targets` clean on the pinned `nightly-2025-11-15`. The pre-existing `cargo fmt` drift in this crate is untouched — `error_tests.rs` reported 6 rustfmt diffs before and 6 after, none in the added lines.

BIMvoice:
Reproduced the CORS finding independently, because "a test that reads the runner's environment" is the failure mode where the test silently stops meaning what its name says.

On `a23342ca` with a hostile environment:

```
CORS_ORIGINS='*' cargo test -p ifc-lite-server cors_tests
test result: ok. 4 passed; 0 failed
```

Reverting the fixture to `Config::from_env()` (replacements=1), same environment:

```
test cors_tests::the_shipped_default_is_not_permissive ... FAILED

assertion `left != right` failed: the default CORS configuration must not be permissive: ["*"]
  left: "*"
 right: "*"
```

Restored: 4 passed.

The failure message is the point. The test is named `the_shipped_default_is_not_permissive`, and under `from_env()` it was asserting against **whatever the runner happened to set** — so on a machine with a permissive `CORS_ORIGINS` it reports that the shipped default is permissive, which is simply false. In the other direction, on a runner with a *stricter* value it would have passed while proving nothing about the default. CodeRabbit was right, and this is a better catch than a typical nitpick: the test does not fail loudly when it is wrong, it changes what it means.

Worth noting the fix was available only because this PR already introduced `from_lookup(get)` for a different reason — to stop `parity_tests::test_state` racing on `set_var`. Same injection seam, second use.

Two more things I would not want lost in the summary above:

**The exhaustiveness guard is genuine, and its limit is stated rather than implied.** Inserting a hypothetical `ApiError::BrandNew` produces `error[E0004] … not covered` at `error_tests.rs:57` — a compile error, so a new variant cannot be added without the table noticing. The one hand-maintained number is `VARIANT_COUNT`; deleting the `Multipart` row gives `left: 10 right: 11`. Full generation from the type is impossible because each row needs a constructed value, and neither `MultipartError` nor `JoinError` has a public constructor (both were provoked for real — a malformed multipart body, and a panicking spawned task — rather than faked).

**On the traversal thread on #2154, CodeRabbit's diagnosis was right but its proposed test was inert.** Its suggested `%2e%2e%2f…/etc/passwd` kills nothing, because nothing in the path decodes it — the 404 it produces is just a missing file, not a blocked traversal. Aiming the encoded case at `pkg/../package.json`, which genuinely exists outside the served root, is what gives it teeth: with `decodeURIComponent` added at the call site and the guard replaced by `if (false)`, the literal-only test stays green while the new one fails `expected 404, actual 200`. Applying the suggestion verbatim would have added a test that could never fail.

PR #2157, opened 2026-08-05
fix(sources): close the icon-URL bypass, render real preference controls, test syncSourceModel

BIMvoice:
Offered to **take or leave**. This does not touch `feat/source-plugin-api-v2-host` directly — it targets it, so you can merge, cherry-pick, or ignore it. Same shape as #2155 for #2023. Nothing has been posted on #1976.

Three of the open review threads. The two module splits (`SourceBrowser.tsx`, `useSourceCatalogSync.ts`) are deliberately **not** here: they touch the same files as everything else and would collide with every other fix in flight.

Test counts for `apps/viewer/src/**/sources/**`: **50 tests / 15 suites** before, **92 tests / 26 suites** after, 0 failures either way.

---

## 1. `safeIconUrl` — the `/\` bypass, and the `permissions.network` gate

`SourceProviderRow.tsx`. The old check string-prefixed the raw manifest value:

```ts
if (raw.startsWith('//')) return undefined;
if (raw.startsWith('/') || raw.startsWith('./')) return raw; // "same-origin"
```

`/\evil.example/beacon.gif` passes `raw.startsWith('/')` and is returned as same-origin. The WHATWG URL parser folds `\` into `/` for special schemes, so what the browser actually requests is `//evil.example/beacon.gif` — a cross-origin beacon fired by an `img src`, from third-party manifest data.

Now it resolves through `new URL(raw, window.location.href)` and checks the **resolved** origin:

- same-origin → allowed, returned normalized, **independent of the allowlist**;
- otherwise https **and** hostname in `permissions.network` → allowed;
- everything else → `undefined` (generic icon). `data:`, `javascript:`, plaintext `http:` and protocol-relative all fall out of that one check rather than needing their own special cases.

The allowlist match reuses `isAllowedHost` from `services/sources/host-fetch.ts` — the same function `ctx.fetch` enforces — so `*.example.com` behaves identically in both places and there is no second copy of the matching rule to drift.

**Empty allowlist is pinned in both directions**, since that is the case that reads as "allow nothing" or "allow everything" depending on how the check is written: `permissions.network: []` rejects *every* cross-origin https icon, and still allows `/icons/acme.svg`.

### RED

Exporting the old implementation unchanged and running the new test file: **9 of 18 failed**, including all three backslash variants and every allowlist assertion. After the fix: **18/18**.

### Mutations (each applied by exact-substring replacement, replacement count asserted `== 1`, mutated region re-read from disk before believing the result, restored from a saved copy)

| # | Mutation | count | Result |
|---|---|---|---|
| A | reinstate `if (raw.startsWith('/')) return raw;` | 1 | 5 fail — all four backslash variants + protocol-relative |
| B | `isAllowedHost(...) ? resolved.href : undefined` → `resolved.href` | 1 | 3 fail — non-allowlisted host, empty allowlist, wildcard negative |
| C | `origin === location.origin` → `&& allowedHosts.length > 0` | 1 | 4 fail — every same-origin-with-empty-allowlist case |
| D | disable the https check | 1 | 5 fail — backslash variants + "http on an allowlisted host" |

A test that only checked "allowed host passes" would survive B and C. Both are killed.

## 2. Checkbox and dropdown preferences rendered as text inputs

`SourceSettingsDialog.tsx`. `PreferenceField` rendered `<Input type={pref.type === 'password' ? 'password' : 'text'}>` for all four `PluginPreferenceType` members and never read `pref.options`. A provider declaring a `dropdown` got a free-text box: no option list shown, any string typeable into a field whose entire point is that only the declared values are legal. A `checkbox` expected the user to type the word `true`.

Now `checkbox` renders a `Switch` and `dropdown` a `Select` populated from `pref.options`. Both come from the app's existing `@/components/ui` set (`@radix-ui/react-switch` / `@radix-ui/react-select`, already dependencies) — **no new UI dependency**. `password` and `text` are untouched. Values stay strings end to end, because that is what `PluginPreference.default` and the host's preference storage are: a checkbox round-trips `"true"` / `"false"` exactly as the type documents, a dropdown round-trips the option's `value`.

One accessibility detail: `htmlFor` only associates with a labelable element, which the Radix triggers (buttons) are not, so those carry `aria-labelledby` back to the label instead.

### RED

New test file against the current implementation: **9 of 12 failed** — every checkbox and dropdown assertion. Only the `textfield` / `password` cases passed. After the fix: **12/12**.

### Mutations

| # | Mutation | count | Killed |
|---|---|---|---|
| E | `checked={value === 'true'}` → `checked={false}` | 1 | "renders checked for `"true"`", "saves `"false"` after turning off" |
| F | `onChange(checked ? 'true' : 'false')` → inverted | 1 | both save-direction tests |
| G | `SelectItem` body `{option.label}` → `{option.value}` | 1 | "shows the label of the stored value", "offers exactly the declared options", "saves the VALUE not the label" |
| H | `SelectItem value={option.value}` → `value={option.label}` | 1 | "shows the label of the stored value", "saves the VALUE not the label" |

## 3. `syncSourceModel.ts` had no tests

12 tests, in `syncSourceModel.test.ts`, aimed at the contract rather than at "it calls the thing":

- **a failed fetch must not cost the user their model** — the module deliberately loads the replacement first and removes the old model only on success; a failed `download`, and separately a failed `addModel`, must both reject *and* leave the original in the store, with nothing removed;
- **a provider that no longer lists the file must say so** — empty page, page with other files, cursor followed across pages, and a provider repeating the same cursor forever caught on page 2 rather than after 100;
- **filename sanitisation** — a provider-supplied `../../etc/pas swd.ifc` reaches the `File` with no path separator, a name that sanitises away to nothing falls back to `model`, and the user's own renamed model label survives instead of being overwritten by the source filename;
- **the revision id lands on the replacement** — the new tag carries `latestFile.currentRevisionId` (and the file's *current* container, not the stale tagged one) and is stored against the fresh model id; leave the old id there and the model reports itself out of date forever.

These are characterization tests on existing behaviour, so there is no RED step to show — their value is entirely in whether they have teeth. **13 mutations, every one killed, no survivors**, and every one of the 12 tests is killed by at least one:

| # | Mutation | count | Killed |
|---|---|---|---|
| N1 | `removeModel(modelId)` moved before the download | 1 | 8 — incl. "leaves the original model loaded" |
| N2 | `if (!addedId && !registered)` → never fires | 1 | 1 — "keeps the original when the reload fails" |
| N3 | `items.find(f => f.id === tag.fileId)` → `items[0]` | 1 | 3 — wrong-file, cursor walk, stuck cursor |
| N4 | drop the repeated-cursor guard | 1 | 1 — stuck cursor (2 pages, not 100) |
| N5 | stop passing `cursor` to the next page | 1 | 1 — cursor walk |
| N6 | skip `sanitizeFilename` | 1 | 2 — both filename tests |
| N7 | `fallback: 'model'` → `'file'` | 1 | 1 — the fallback test |
| N8 | `name: model.name` → `name: safeFileName` | 1 | 1 — user's label preserved |
| N9 | new tag gets `tag.revisionId` instead of `latestFile.currentRevisionId` | 1 | 1 — revision id |
| N10 | new tag keeps `tag.containerId` | 1 | 1 — revision id / moved container |
| N11 | never `setActiveModel(replacementId)` | 1 | 1 — swap + activate |
| N12 | leak the `inFlightSyncs` entry | 1 | 11 — incl. "a retry is possible" |
| N13 | `if (!latestFile)` never throws | 1 | 3 — both "file is gone" cases |

Each mutation was applied by exact-substring replacement asserting a replacement count of exactly 1, the mutated region re-read from disk to confirm the text actually changed, and restored by copying back a saved backup. The harness refuses to start unless the target file is pristine — an earlier run had a mutation leak into its own backup and produced uniformly cascading failures, which is what that guard now prevents.

---

## Checks

- `pnpm typecheck` at root: **pass**.
- Test files **are** covered by it: `apps/viewer/tsconfig.json` overrides the root `exclude` (`**/*.test.ts`) with its own `["src/lib/scripts/templates"]`, so `src/**/*` pulls the tests in. Verified rather than assumed — a deliberate `const x: number = "s"` was appended to both a new `.test.ts` and a new `.test.tsx`, `tsc --noEmit` flagged both, and removing them returned a clean run.
- No changeset: the branch already carries `cloud-source-plugin-architecture.md` and `plugin-api-v2-contract.md`, and this is a fix to unreleased work under them.
- Nothing outside `apps/viewer/src/**/sources/**` is touched. `packages/plugin-api` needed no change — `PluginPreferenceType` and `DropdownOption` already declared everything the UI now honours.

### DIFF
--- a/apps/viewer/src/components/sources/SourceProviderRow.test.tsx
+++ b/apps/viewer/src/components/sources/SourceProviderRow.test.tsx
@@ -1,0 +1,144 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `safeIconUrl` — the provider-manifest icon is third-party data landing in an
+ * `img src`, so the host decides what it is allowed to point at.
+ *
+ * Two properties are pinned here, both of which the pre-fix string-prefix
+ * implementation got wrong:
+ *
+ * 1. **The `/\` bypass.** `/\evil.example/beacon.gif` passes a
+ *    `raw.startsWith('/')` "same-origin relative" check, but the WHATWG URL
+ *    parser (and therefore every browser) treats `\` as `/` for special
+ *    schemes, so the byte stream the browser actually requests is
+ *    `//evil.example/beacon.gif` — a cross-origin beacon. Resolution has to go
+ *    through `new URL()` and the check has to be on the RESOLVED origin.
+ *
+ * 2. **The allowlist gate.** An absolute icon URL is allowed only when its host
+ *    is in `permissions.network` — the same allowlist the host's fetch wrapper
+ *    already enforces, so a plugin can point its icon exactly where it has
+ *    already declared it talks. A same-origin relative icon (`/icons/acme.svg`)
+ *    stays allowed independently of that list, so a provider with no network
+ *    permissions at all can still ship an icon.
+ *
+ * The empty-allowlist case is tested in BOTH directions on purpose: "empty"
+ * reads as "allow nothing" or "allow everything" depending on how the check is
+ * written, and only asserting one side leaves the other free.
+ */
+
+import '@/test/setup-dom.js';
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { safeIconUrl } from './SourceProviderRow.js';
+
+/** The origin `setup-dom` registers happy-dom with. */
+const ORIGIN = 'http://localhost';
+
+describe('safeIconUrl — same-origin relative icons', () => {
+  it('allows a same-origin absolute path even when permissions.network is empty', () => {
+    assert.equal(safeIconUrl('/icons/acme.svg', []), `${ORIGIN}/icons/acme.svg`);
+  });
+
+  it('allows a same-origin ./ relative icon when permissions.network is empty', () => {
+    assert.equal(safeIconUrl('./icons/acme.svg', []), `${ORIGIN}/icons/acme.svg`);
+  });
+
+  it('allows a same-origin absolute URL spelled out in full', () => {
+    assert.equal(
+      safeIconUrl(`${ORIGIN}/icons/acme.svg`, []),
+      `${ORIGIN}/icons/acme.svg`,
+    );
+  });
+});
+
+describe('safeIconUrl — the backslash bypass', () => {
+  // Every one of these starts with `/`, so a `raw.startsWith('/')` check reads
+  // them as same-origin. The URL parser resolves each to `//evil.example`.
+  for (const raw of [
+    '/\\evil.example/beacon.gif',
+    '/\\/evil.example',
+    '/\\\\evil.example/beacon.gif',
+    '\\\\evil.example/beacon.gif',
+  ]) {
+    it(`rejects ${JSON.stringify(raw)}`, () => {
+      // Guard the premise: if the parser ever stopped folding `\` into `/`,
+      // these inputs would be harmless and the test would be vacuous.
+      const resolvedOrigin = new URL(raw, `${ORIGIN}/`).origin;
+      assert.equal(
+        resolvedOrigin,
+        'http://evil.example',
+        'premise: the URL parser must resolve this off-origin',
+      );
+      assert.equal(safeIconUrl(raw, []), undefined);
+      assert.equal(
+        safeIconUrl(raw, ['evil.example']),
+        undefined,
+        'plain http stays rejected even for an allowlisted host',
+      );
+    });
+  }
+});
+
+describe('safeIconUrl — schemes that are never icons', () => {
+  it('rejects a protocol-relative URL', () => {
+    assert.equal(safeIconUrl('//evil.example/x.png', []), undefined);
+  });
+
+  it('rejects data:', () => {
+    assert.equal(safeIconUrl('data:image/svg+xml,<svg/>', []), undefined);
+  });
+
+  it('rejects javascript:', () => {
+    assert.equal(safeIconUrl('javascript:alert(1)', []), undefined);
+  });
+
+  it('rejects plain http even for an allowlisted host', () => {
+    assert.equal(safeIconUrl('http://icons.example/i.png', ['icons.example']), undefined);
+  });
+
+  it('rejects an unparseable value', () => {
+    assert.equal(safeIconUrl('https://', []), undefined);
+  });
+
+  it('passes an absent iconUrl through', () => {
+    assert.equal(safeIconUrl(undefined, ['icons.example']), undefined);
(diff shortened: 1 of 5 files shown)


BIMvoice:
Verified the icon-URL fix myself before relaying, since it is the one you gave an explicit design ruling on.

Reverting `safeIconUrl` to the string-prefix check (`raw.startsWith('/')`, replacements=1) fails **7 of 18**:

```
not ok - rejects "/\evil.example/beacon.gif"
not ok - rejects "/\/evil.example"
not ok - rejects "/\\evil.example/beacon.gif"
not ok - rejects a protocol-relative URL
not ok - allows a same-origin absolute path even when permissions.network is empty
not ok - allows a same-origin ./ relative icon when permissions.network is empty
not ok - allows a same-origin absolute URL spelled out in full
```

Restored: 18/18.

Both halves of your ruling are pinned, and they fail for different reasons — the first four because the bypass reopens, the last three because the empty-allowlist case flips meaning. That second group is the one you singled out: *"'empty allowlist' is the case that will be read as 'allow nothing' or 'allow everything' depending on how it is written."* It is now asserted in both directions, so it cannot drift into either reading silently.

The implementation follows what you specified rather than a variation on it: resolve through `new URL(raw, window.location.href)` and check the **resolved** origin, so whatever escaping produced the string is normalized away before the comparison; same-origin allowed independent of the allowlist; otherwise https **and** hostname in `permissions.network`, reusing `isAllowedHost` from `host-fetch.ts` so wildcard matching is identical to what `ctx.fetch` already enforces rather than a second, subtly different rule sitting beside it.

The preference controls use the app's existing radix `Switch`/`Select` — already dependencies, no new UI package. Values stay strings so nothing downstream changes shape.

**Two things worth your attention that are not in the diff:**

**`apps/viewer` test files ARE typechecked**, unlike the packages. `apps/viewer/tsconfig.json` overrides the root `exclude` of `**/*.test.ts` with its own `["src/lib/scripts/templates"]`. The agent verified this rather than assuming — appended `const x: number = "s"` to a new `.test.ts` and a new `.test.tsx`, confirmed `tsc --noEmit` flagged both, removed them and confirmed clean. That is the opposite of the situation in `packages/*`, and worth knowing when reading coverage claims from either side.

**A process failure I want on the record rather than buried.** The first mutation run crashed *after* writing its edit, so the next run's "backup" captured an already-mutated file, and every subsequent restore reinstated that line — producing cascading failures that looked exactly like test cross-contamination. It was caught by `git diff`, reverted by exact substring rather than `git checkout --`, and the whole set re-run against a pristine file with a guard added. Every number above is from the clean run. Flagging it because "restore from backup" silently becomes "restore the mutation" if the backup is taken after a failed run, and the symptom impersonates a real bug in the tests.

The two module splits (`SourceBrowser.tsx`, `useSourceCatalogSync.ts`) were deliberately not attempted — they touch the same files as everything else here and would collide, as discussed.

Offered to take or leave; it does not touch `feat/source-plugin-api-v2-host` directly.

PR #2158, opened 2026-08-05
test(spatial,mcp): self-audit follow-up — three merged tests our own mutation sweep could not kill

BIMvoice:
## Why

23 test-coverage PRs merged into `main` in the last day. Review already caught real vacuity in three of them at or after merge — #2145 (a dedup test pinning only one direction), #2151 (a loop where 5 of 8 cases asserted nothing), #2148 (round-trips built from paired production functions). This is a self-audit for the same shapes in the rest of that work, done the way those defects were found: apply the minimal mutation each test claims to catch, plus a neighbouring one it arguably should, and record KILLED / SURVIVED — per case for loops, not just "the loop failed".

Sampled six PRs, biased to the failing shapes: **#2147** (spatial/clash — loops over axes, `[]` and `Infinity` sink assertions), **#2144** (mcp auth — two-valued signals, copy semantics), **#2136** (rust georef — a cross-check between two paired production functions, the #2148 shape), **#2140** (a zero-valued fixture), **#2146** (a 3x3 region-combination loop), plus the clash box-kernel symmetry tests (self-paired assertions). 24 mutations. Three survivors.

## Measurement

Every mutation was an exact-substring replacement asserting a replacement count of 1, with the mutated region re-read before the run and restored from a backup taken before the first mutation.

| PR | Test | Claimed property | Discriminates | Evidence |
|---|---|---|---|---|
| #2147 | `BVH.raycast` "normalizes the direction" | direction is normalized | **no** | deleting the normalization entirely → SURVIVED (hit set is invariant under any positive scaling) |
| #2147 | `FrustumUtils` "normalizes planes extracted from a view-projection matrix" | planes are normalized | **partial** | `plane.distance /= len` → `/= 1` → SURVIVED; left-plane `m[3] + m[0]` → `m[3] - m[0]` → SURVIVED; `if (len > 0)` → `if (len > 1e12)` → KILLED |
| #2147 | "does not miss meshes when the extent is longest on y or z" | y/z query geometry | **yes** (per case) | y slab widened by 100 → KILLED at axis 1; z slab widened → KILLED at axis 2; but forcing `axis = 0` for every split → SURVIVED, so the split-axis rationale in the comment is not what the test pins |
| #2147 | "grazes a slab plane exactly" | parallel-slab guard | **yes** | guard removed → KILLED (3 tests); `<`/`>` → `<=`/`>=` → KILLED |
| #2147 | "does NOT report meshes behind the ray origin" | `tmax >= 0` | **yes** | `return tmax >= 0` → `return true` → KILLED |
| #2147 | "keeps a batch just outside the frustum edge" | the -0.5 m anti-flicker margin | **yes** | `PLANE_EPSILON = -0.5` → `0` → KILLED (2 tests) |
| #2147 | clash `intersects` "is symmetric" | symmetry | **yes** | `a.max[0] >= b.min[0]` → `>` (asymmetric) → KILLED; `return true` → SURVIVED here, KILLED by the separation test in the same file |
| #2144 | "hand back copies, so a caller mutating one does not widen the next" | constructors return copies | **no** | `fullScope()` → `return FULL_ACCESS` → SURVIVED in isolation (the assertion compares `fullScope()` to the constant a leak corrupts); the whole-file failure came from a later test, i.e. cross-test leakage. And the named property does not hold of the code at all: the spread was shallow |
| #2144 | "lets `admin` stand in for a scope the token does not list" | the escalation arm | **yes** | `|| scope.scopes.includes('admin')` removed → KILLED |
| #2144 | `modelAllowed` allowlist | per-model gating | **yes** | `return scope.modelIds.includes(modelId)` → `return true` → KILLED (2 tests) |
| #2136 | `to_matrix` column pinning + `local_to_map` cross-check | rotation / scale / translation | **yes** | col-1 sign flip → KILLED; `local_to_map` sign flip alone → KILLED by the cross-check; **both flipped together** (the symmetric error a cross-check cannot see) → KILLED by the pinned columns; z scale `s` → `1.0` → KILLED |
| #2140 | "falls back to the metre default instead of accepting a zero scale" | the `s > 0` guard | **yes** | guard dropped → KILLED; `s > 0` → `s >= 0` → KILLED; fallback `1.0` → `2.0` → KILLED |
| #2146 | soundness invariant, 3x3 region combinations | structural intersection always conflicts | **partial** | `structural || spatial` → `structural && spatial` → KILLED in 7 of 9 cases; the 2 survivors are the combinations where both regions overlap, so `spatial` is true anyway |

## What this changes

**1. `BVH.raycast` normalization (#2147).** Deleting the normalization survived: the slab test scales every `t` by the same positive factor, so the hit set is invariant under any positive scaling and the assertion is a tautology. The degenerate direction is the one observable input — it divides by a zero length and the NaN components make the slab test reject, so a zero-length ray hits nothing; without the normalization the call takes the parallel-slab branch on all three axes and reports every mesh containing the origin (a viewer pick with a degenerate direction would select the element the camera sits inside). Pinned; the mutation is now KILLED.

**2. `fromViewProjMatrix` (#2147).** The only assertion was that the six normals are unit length, which says nothing about the plane constants or about which matrix rows each plane comes from. A new test extracts a known orthographic volume (x, y in [-10, 10], z in [0, 20]) and asserts one inside and one outside box per plane, in world units. Dropped distance normalization, flipped left plane and swapped near-plane constant are now all KILLED.

**3. `readOnlyScope()` / `fullScope()` (#2144).** The test compared `fullScope().scopes` against `FULL_ACCESS.scopes` — the very constant a leak corrupts — so it held however badly the constructor leaked. Now asserted against a literal. Measuring it also showed the property the test names is not one the code had: the spread was shallow, so `readOnlyScope().scopes.push('mutate')` wrote into `READ_ONLY` itself and every token minted afterwards in the process carried the extra scope. `scopes` is now copied too, with the in-place mutation pinned (patch changeset included).

Plus a corrected comment on the y/z BVH loop, which claimed to exercise a split-axis choice that is invisible in the results.

The other sampled tests held, including the one built in the #2148 shape: the georef matrix pins its columns, so even a sign error applied symmetrically to both `to_matrix` and `local_to_map` is caught.

## Verification

- `packages/spatial`: 46 tests pass (45 before); each new assertion RED-verified against the mutation it exists to catch.
- `packages/mcp`: 232 tests pass with dependencies built.
- `pnpm turbo typecheck --force` clean for both packages.

### DIFF
--- a/.changeset/mcp-scope-constructor-array-copy.md
+++ b/.changeset/mcp-scope-constructor-array-copy.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/mcp": patch
+---
+
+`fullScope()` and `readOnlyScope()` now copy the `scopes` array as well as the
+wrapper object. The shallow spread handed every caller the same array instance
+as the exported `FULL_ACCESS` / `READ_ONLY` constants, so an in-place mutation
+(`readOnlyScope().scopes.push('mutate')`) widened the constant itself and every
+token minted afterwards in the same process carried the extra scope.
--- a/packages/mcp/src/auth/scope.test.ts
+++ b/packages/mcp/src/auth/scope.test.ts
@@ -75,14 +75,37 @@ describe('modelAllowed', () => {
 
 describe('scope constructors', () => {
   it('hand back copies, so a caller mutating one does not widen the next', () => {
+    // Assert against a LITERAL, not against `FULL_ACCESS.scopes`: comparing
+    // `fullScope()` to the very constant a leak would have corrupted is a
+    // self-comparison that holds however badly the constructor leaks (with
+    // `fullScope()` returning `FULL_ACCESS` itself, this test passed in
+    // isolation — measured; the only failure came from a later test in the
+    // file, i.e. from cross-test leakage, not from this assertion).
     const a = fullScope();
     a.scopes = ['read'];
-    expect(fullScope().scopes).toEqual(FULL_ACCESS.scopes);
+    expect(fullScope().scopes).toEqual(['read', 'validate', 'mutate', 'export', 'admin']);
+    expect(FULL_ACCESS.scopes).toEqual(['read', 'validate', 'mutate', 'export', 'admin']);
 
     const b = readOnlyScope();
     b.modelIds = ['only-this-one'];
     expect(readOnlyScope().modelIds).toBeUndefined();
     expect(READ_ONLY.scopes).not.toContain('mutate');
+  });
+
+  it('are copies against IN-PLACE mutation too, not just reassignment', () => {
+    // Reassigning `.scopes` is the one mutation a shallow `{ ...CONST }` copy
+    // survives, so the assertion above cannot see the aliasing that matters:
+    // a caller that PUSHES onto the array it was handed writes straight into
+    // the exported constant, and every token minted afterwards in the process
+    // carries the extra scope — a read-only token silently gaining `mutate`.
+    readOnlyScope().scopes.push('mutate');
+    expect(READ_ONLY.scopes).not.toContain('mutate');
+    expect(readOnlyScope().scopes).toEqual(['read', 'validate', 'export']);
+    expect(scopeAllows(readOnlyScope(), 'mutate')).toBe(false);
+
+    fullScope().scopes.length = 0;
+    expect(FULL_ACCESS.scopes).toHaveLength(5);
+    expect(scopeAllows(fullScope(), 'read')).toBe(true);
   });
 
   it('read-only really is read-only', () => {
--- a/packages/mcp/src/auth/scope.ts
+++ b/packages/mcp/src/auth/scope.ts
@@ -40,10 +40,14 @@ export function modelAllowed(scope: AuthScope, modelId: string): boolean {
   return scope.modelIds.includes(modelId);
 }
 
+/* The `scopes` array is copied too, not just the wrapper object: a spread
+ * alone hands every caller the SAME array instance as the exported constant,
+ * so `readOnlyScope().scopes.push('mutate')` widens `READ_ONLY` itself and
+ * every token minted afterwards in the process gets the extra scope. */
 export function readOnlyScope(): AuthScope {
-  return { ...READ_ONLY };
+  return { ...READ_ONLY, scopes: [...READ_ONLY.scopes] };
 }
 
 export function fullScope(): AuthScope {
-  return { ...FULL_ACCESS };
+  return { ...FULL_ACCESS, scopes: [...FULL_ACCESS.scopes] };
 }
--- a/packages/spatial/src/bvh.test.ts
+++ b/packages/spatial/src/bvh.test.ts
@@ -31,6 +31,24 @@ function row(n: number, axis: 0 | 1 | 2 = 0): MeshWithBounds[] {
   });
 }
 
+/**
+ * Rotation matrix about `axis` (normalized here) by `angle`, Rodrigues form,
+ * as `R[row][col]`. Used to build a view-projection whose upper 3x3 has no
+ * zero entry.
+ */
+function rotationAboutAxis(axis: [number, number, number], angle: number): number[][] {
+  const n = Math.hypot(axis[0], axis[1], axis[2]);
+  const [x, y, z] = [axis[0] / n, axis[1] / n, axis[2] / n];
+  const c = Math.cos(angle);
+  const s = Math.sin(angle);
+  const t = 1 - c;
+  return [
+    [t * x * x + c, t * x * y - s * z, t * x * z + s * y],
+    [t * x * y + s * z, t * y * y + c, t * y * z - s * x],
+    [t * x * z - s * y, t * y * z + s * x, t * z * z + c],
+  ];
+}
+
 describe('BVH.build', () => {
   it('returns an empty index for no meshes without throwing', () => {
     const bvh = BVH.build([]);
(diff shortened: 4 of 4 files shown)


BIMvoice:
Reproduced the `scope.ts` finding independently. It is a real defect, and the way it was found is the part worth recording.

`packages/mcp/src/auth/scope.ts:43-45` on `main`:

```ts
export const READ_ONLY: AuthScope = { scopes: ['read', 'validate', 'export'] };
export function readOnlyScope(): AuthScope { return { ...READ_ONLY }; }
```

The spread is shallow, so `scopes` is the **same array object**, not a copy:

```
before:                              ["read","validate","export"]
after  a.scopes.push("mutate"):
  the caller's "copy"                ["read","validate","export","mutate"]
  the exported READ_ONLY constant    ["read","validate","export","mutate"]
  what the NEXT caller receives      ["read","validate","export","mutate"]
  same array identity                true
```

So any code that treats its returned scope as private and appends to `scopes` silently widens the read-only scope **for every subsequent caller in the process**. In `packages/mcp` — the surface an external agent talks to — that is a privilege escalation, not a hygiene issue. Nothing does this today; the constructor simply does not provide the isolation its name promises.

**The test that was supposed to catch this could not, by construction.** `hand back copies…` asserted the returned object against `READ_ONLY` itself — the very constant a leak corrupts. Both sides move together, so the assertion holds precisely when the bug fires. Replacing the constructor body with `return READ_ONLY` **survived in isolation**; the whole-file run only went red because of an unrelated later test, which is what disguised it.

That is the same shape as the fixtures this sweep has been removing all day — a test whose reference value is derived from the thing under test — except this one is ours, in a PR that was itself closing a coverage gap in this exact file.

Two other genuine survivors in the sample, both also ours:

- **`BVH.raycast` "normalizes the direction"** — deleting the normalization **survived**. The hit set is invariant under any positive scaling of the direction, so the test was a tautology: it asserted a property that holds whether or not the code does the thing. Now pinned on the one genuinely observable consequence — a zero-length direction must return `[]`, where an unnormalized path instead hits every mesh containing the origin.
- **`FrustumUtils` "normalizes planes…"** — `distance /= len` → `/= 1` survived, and flipping the left plane's `m[3]+m[0]` to `m[3]-m[0]` survived too. Replaced with a test extracting a known orthographic volume and checking one inside and one outside box per plane, in world units; all three mutations now die.

**Scoreboard, stated plainly: 24 mutations across 6 sampled PRs, 3 genuine survivors.** Two of the sampled tests came back stronger than expected — the georef cross-check in #2136 survives even a *symmetrically* flipped pair because the matrix columns are pinned independently, which is exactly the failure mode that got past us in #2148.

One correction to a comment rather than a test: the y/z split-axis loop in #2147 does discriminate per case, but forcing `axis = 0` for every split still survives — so the comment's claim about the split-axis rationale overstated what the test pins. Comment corrected rather than the test padded.

BIMvoice:
## Round-four self-audit of this PR — one of our own defects found, fixed in 48976e1

This PR is itself a fix for vacuity found in round one, so we re-measured it with the same method (minimal mutation, `assert n == 1` on the replacement, KILLED/SURVIVED per test). Verdict: **6 of 8 mutations KILLED as claimed; 2 SURVIVED, both against the new frustum test this PR adds.**

### What held

| Mutation | Result |
|---|---|
| `packages/spatial/src/bvh.ts` — delete the raycast direction normalization (`direction[i] / len` → `direction[i]`) | **KILLED** (1 test: the degenerate-direction assertion added here) |
| `frustum.ts` — drop `plane.distance /= len` | **KILLED** (1 test) |
| `frustum.ts` — flip the left plane's normal sign (`m[3] + m[0]` → `m[3] - m[0]`) | **KILLED** (1 test) |
| `frustum.ts` — near-plane constant `m[14]` → `m[15] + m[14]` | **KILLED** (1 test) |
| `frustum.ts` — left plane's x reads `m[4]` instead of `m[0]` | **KILLED** |
| `mcp/src/auth/scope.ts` — revert `readOnlyScope` to the shallow `{ ...READ_ONLY }` | **KILLED** |
| `scope.ts` — revert `fullScope` to the shallow spread | **KILLED** |
| `scope.ts` — `fullScope()` returns `FULL_ACCESS` itself | **KILLED**, and **in isolation** (`vitest -t "hand back copies"` → 1 failed, 9 skipped). That was the exact failure mode this PR set out to fix — the old assertion only failed via cross-test leakage — so the literal-comparison fix is doing what it claims. |

The comment claim about the split axis is also accurate: forcing `const axis = 0` leaves the whole `packages/spatial` suite green (47 tests, 3 files).

### The defect — the new frustum fixture is diagonal

`extracts the six planes of a known orthographic volume, in world units` uses

```
0.1, 0, 0, 0,   0, 0.1, 0, 0,   0, 0, 0.05, 0,   0, 0, 0, 1
```

so `m[1]`, `m[2]`, `m[4]`, `m[6]`, `m[8]`, `m[9]` are all **zero**. Any normal component read from the wrong row of the upper 3x3 is therefore an identity on this fixture — which is precisely the "nothing pinned which matrix rows each plane comes from" gap the PR body claims to have closed. Measured:

| Mutation | Before | After 48976e1 |
|---|---|---|
| bottom plane's x reads `m[2]` instead of `m[1]` | **SURVIVED** (23/23 green) | KILLED |
| near plane's x reads `m[6]` instead of `m[2]` | **SURVIVED** (23/23 green) | KILLED |

This is our own defect, in the fix PR, in shape 6 of the catalogue (a fixture constant that makes the operation an identity).

### Fix

A second fixture rotates the view about `(1, 2, 3)` by 0.7 rad (Rodrigues, computed in the test — the expectations come from the fixture rotation, not from `fromViewProjMatrix`), with deliberately unequal extents (x ±10, y ±5, z 0..20) so a swap between two planes is not masked by a symmetric volume. The test asserts no entry of the upper 3x3 is near zero before using it.

Face-centre probes alone were **still** not enough: probing a plane only along its own axis lets a mis-wired component keep the sign of that one dot product, and the near-plane mutation survived even the rotated fixture until all eight volume corners were asserted visible. With the corners in, every component is loaded at once.

Post-fix sweep, `packages/spatial` (24 tests in `bvh.test.ts`, 47 in the package):

| Mutation | Count | Result |
|---|---|---|
| bottom.x `m[1]` → `m[2]` | 1 | KILLED |
| near.x `m[2]` → `m[6]` | 1 | KILLED |
| drop `plane.distance /= len` | 1 | KILLED (2 tests) |
| left plane sign flip | 1 | KILLED (2 tests) |
| near constant `m[14]` → `m[15] + m[14]` | 1 | KILLED (2 tests) |
| left.x `m[0]` → `m[4]` | 1 | KILLED (3 tests) |
| top.y `m[5]` → `m[9]` | 1 | KILLED (3 tests) |
| far.z `m[10]` → `m[6]` | 1 | KILLED (3 tests) |

No production code changed by the follow-up; `packages/spatial` is 47/47 green.

BIMvoice:
Verified this, and it is the most structurally important result of the whole audit effort — because **this PR was itself a fix for previously-found vacuity, and the fix was vacuous in the same way.**

The orthographic fixture added here is diagonal, so `m[1]`, `m[2]`, `m[4]`, `m[6]`, `m[8]` and `m[9]` are all zero. A plane component read from the wrong matrix row is then an **identity** — which is precisely the "which rows each plane comes from" claim the test was written to pin. Two mutations survived it: the bottom plane reading `m[2]` instead of `m[1]`, and the near plane reading `m[6]` instead of `m[2]`.

That is the identity-fixture class — now the tenth package it has appeared in — occurring inside a test written specifically to close a gap of the same kind. Round one found vacuity, round one's fix was vacuous, round four caught it. If the chain had stopped at three rounds, this would have shipped looking closed.

**The fix guards against the recurrence rather than just correcting the instance:**

```ts
it('extracts the six planes under a dense rotation, where no matrix entry is zero', () => {
  for (const r of rot) for (const v of r) expect(Math.abs(v)).toBeGreaterThan(0.05);
```

That assertion is on the *fixture*, not the code under test. It fails if anyone later simplifies the rotation back toward something axis-aligned — so the property that makes the test discriminating cannot silently decay. Given this class has now defeated us in ten packages, a fixture-integrity assertion is a better answer than a better fixture.

**The counterpart result matters as much: #2160 held at 12 of 12.** Round two's follow-up is sound — every mutation kills exactly one named test. So the chain is not uniformly weak; one link was.

## Round four overall

128 mutations, **113 killed, 15 survived** — 11 genuine defects (fixed) and 4 equivalent mutants discarded rather than counted. The sharpest of the eleven: in #2162, `globalHidden.add(id)` → `delete(id)` survived, meaning lens-hidden ids were **never applied at all** under a block titled "reacts to every visibility channel it reads". And #2159's lifted-source harness came back near-perfect at 23 of 24, which answers the structural doubt about whether testing lifted source text really propagates a production mutation — it does.

## One process finding worth more than any single defect

The agent's mutation helper at `/tmp/mut.py` was **silently overwritten mid-run by a different concurrent agent's script**, turning four mutations into no-ops that reported clean baselines — four would-be false SURVIVEDs. It was caught only because the tool printed a traceback next to a "passing" result.

That is a new failure mode and the fourteenth distinct false-green mechanism this campaign has surfaced: a shared scratch path is not a safe home for the instrument you are measuring with. It is also my fault rather than the agent's — running several agents concurrently against `/tmp` made the collision possible. Uniquely-named tool paths from here.

One production defect was found and correctly **not** fixed: in `@ifc-lite/data`, `isSpatialStructureTypeName('IFCSPACE')` is `false` while `isSpaceLikeSpatialTypeName('IFCSPACE')` is `true` — the first matches a PascalCase `Set`, the others route through `IfcTypeEnumFromString`, which upper-cases. Pre-existing, and the tests did not enshrine it, so the one-sided "are case-sensitive" assertion was replaced with one that measures and documents the divergence across all three. Behaviour unchanged; the choice is yours.

PR #2159, opened 2026-08-05
test(viewer-core): cover the browser blob's pure logic in viewer-html.ts

BIMvoice:
Follow-up to #2154, which closed the viewer HTTP server, SSE framing, routes and HTML escaping but explicitly flagged what it did **not** reach: `src/viewer-html.ts`, 1,449 lines of largely untested browser-side code. This PR closes the testable part of that, and says plainly which part is not testable.

## The testability judgement

`src/viewer-html.ts` has exactly one node-side statement — the HTML escaping at `src/viewer-html.ts:23-30`, already covered by #2154. Everything from line 31 is a single template literal: the WebGL 2 viewer that ships to the browser.

**Genuinely needs a browser — untouched here, E2E territory:**

| Lines | What |
| --- | --- |
| 142-156, 273-331 | canvas/GL context acquisition, shader compile, program/VAO/buffer creation |
| 158-271 | GLSL shader sources (not executable in node) |
| 487-587 | `uploadGeometry` — `bufferData`/`bufferSubData` against live GL buffers |
| 626-800 | camera input handlers, `ensurePickFbo`, the pick draw+`readPixels` pass |
| 844-857, 1205-1215 | `showPickInfo`, `showCmdLog` — pure DOM |
| 1217-1279 | the render loop |
| 1281-1300 | `connectSSE` / `EventSource` |
| 1302-1449 | `parseMeshesViaPrePass`, `loadModel` — wasm load path |

Roughly 1,000 lines. Mocking a GL context here would only prove the mock works, so none of it is mocked.

**Browser-free and covered here:** `mat4` (79), `updateBounds` (379), `computeEntityBounds` (391), `getChunkIndex` (589), `mergeFloat32`/`mergeUint32` (603/609), `applyColorOverrides` (867), `resolveColor` + `NAMED_COLORS` (931/937), `matchesType` (943), `getEntityBoundsForFilter` (948), the pick-colour encoder/decoder (465-468 / 829), and `handleCommand` (963-1204).

## How

`test/helpers/blob.ts` lifts the **real shipped source text** of a declaration out of the emitted HTML and evaluates it with its free variables supplied explicitly. Nothing is re-implemented; if the source moves or is reformatted, extraction throws instead of silently testing nothing. `assertBrowserFree` fails the suite the moment a lifted declaration starts referencing `document`/`gl`/`window`, so the harness cannot quietly decay into a fake-DOM test.

`handleCommand` is *not* browser-free — it calls into GL upload and status-text seams. Those specific seams are injected as recording stubs; every pure collaborator it calls (`resolveColor`, `matchesType`, `getEntityBoundsForFilter`, the palettes) is the real lifted source, and every assertion is on plain state the command mutated.

## Mutation verification

34 mutations of `src/viewer-html.ts`, **33 killed (97%)**. Flagged **TARGETED**: these aimed at the logic the new tests cover, and say nothing about the ~1,000 GL/DOM lines deliberately left alone.

Two deliberate controls, run before any survivor was believed — dropping the perspective aspect divisor (`m[0]=f/aspect` → `m[0]=f`), and making `matchesType` return `true` unconditionally. Both died.

The one survivor, `Math.max(3, Math.min(10, …))` → `Math.max(3, …)` in `colorByStorey`, was a genuine gap: without the cap a 60m building is cut into ~20 colour bands instead of 10, so palette colours repeat. It is now killed, along with a test for the `Math.max(3, …)` floor on short models.

One survivor is left standing and classified in the PR discussion below: dropping `Math.max(…, 0.01)` from `binSize`. For any reachable input it is equivalent — the model bounds are the union of the entity bounds, so a zero-Y-extent model puts every entity in the same bin either way. The only discriminating input is a model under 3cm tall. Not fully proven beyond that reasoning, so it is recorded rather than claimed.

## Observation, not changed

`src/server.ts:259` validates `action` against the allowlist but nothing else, so `{"action":"section","axis":"w"}` reaches the browser. `handleCommand` then falls the *position* back to the Y extent (`axisIdx` defaults to `1`) while leaving the *normal* all-zero, because the normal is built from three independent `axis === …` tests. A zero normal makes the clip test degenerate: the section neither cuts nor is visibly rejected. This is pinned as observed behaviour in `pins the current behaviour for an unrecognised axis name` and left for @louistrue to rule on rather than changed here.

## Counts

Package tests **4 → 111** on `upstream/main` (#2154 is still open, so its tests are not in this baseline; the two PRs add disjoint files and do not conflict). No production code changed — `git diff` against `src/` is empty. Test files were typechecked separately against the package's own `tsconfig.json` with the `src`-only `include` lifted, which `pnpm typecheck` does not do: clean.

### DIFF
--- a/packages/viewer/test/helpers/blob.ts
+++ b/packages/viewer/test/helpers/blob.ts
@@ -1,0 +1,118 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Harness for testing the *pure* JavaScript that ships inside the browser blob
+ * of `src/viewer-html.ts`.
+ *
+ * The viewer is one 1.4k-line template literal. Most of it is WebGL and DOM
+ * wiring that cannot be exercised in node without faking a GL context — and a
+ * fake GL context only proves the fake works. But a meaningful slice of the
+ * blob is ordinary, browser-free logic: matrix math, chunk/typed-array
+ * bookkeeping, colour resolution, bounds folding and the pick-colour codec.
+ *
+ * This harness lifts the *real shipped source text* of those declarations out
+ * of the emitted HTML and evaluates it with its free variables supplied
+ * explicitly. Nothing is re-implemented and no browser is simulated: if the
+ * source moves or is reformatted, extraction throws instead of silently
+ * testing nothing.
+ */
+
+import { getViewerHtml } from '../../src/viewer-html.js';
+
+/** The HTML the server actually serves, for a fixed model name. */
+export const VIEWER_HTML = getViewerHtml('harness.ifc');
+
+/** The contents of the single `<script type="module">` block. */
+export const SCRIPT: string = (() => {
+  const m = /<script type="module">\n([\s\S]*?)\n<\/script>/.exec(VIEWER_HTML);
+  if (!m) throw new Error('viewer-html.ts no longer emits a <script type="module"> block');
+  return m[1];
+})();
+
+/**
+ * Extract the source text of a top-level declaration from the blob.
+ *
+ * Relies on the file's consistent formatting: top-level declarations start at
+ * column 0 and are closed by a lone `}`, `};`, `]` or `];` at column 0.
+ * Throws loudly rather than returning a partial body.
+ */
+export function extractDecl(name: string): string {
+  const lines = SCRIPT.split('\n');
+  const startPattern = new RegExp(`^(?:async )?(?:function ${name}\\(|(?:const|let) ${name} =)`);
+  const start = lines.findIndex((l) => startPattern.test(l));
+  if (start === -1) throw new Error(`declaration '${name}' not found in the viewer blob`);
+
+  // Single-line declaration (e.g. `const PICK_INDEX_MAX = 0xFFFFFF;`).
+  if (/;\s*(\/\/.*)?$/.test(lines[start]) && !/[{[]\s*(\/\/.*)?$/.test(lines[start])) {
+    return lines[start];
+  }
+
+  for (let i = start + 1; i < lines.length; i++) {
+    if (/^[}\]];?$/.test(lines[i])) return lines.slice(start, i + 1).join('\n');
+  }
+  throw new Error(`declaration '${name}' has no column-0 terminator; formatting changed`);
+}
+
+/** Identifiers that mean the extracted code is *not* browser-free. */
+const BROWSER_GLOBALS = /\b(document|window|navigator|gl|canvas|fetch|EventSource|requestAnimationFrame)\b/;
+
+/**
+ * Assert an extracted declaration touches no browser API. Guards the harness
+ * against silently growing into a fake-DOM test as the viewer evolves.
+ */
+export function assertBrowserFree(name: string, src = extractDecl(name)): void {
+  const hit = BROWSER_GLOBALS.exec(src);
+  if (hit) {
+    throw new Error(
+      `'${name}' now references the browser global '${hit[1]}'; it is no longer node-testable`,
+    );
+  }
+}
+
+/**
+ * Evaluate the given declarations of the blob in one shared scope.
+ *
+ * The lifted source runs in *this* realm (via `Function`), not a `vm` one, so
+ * `Float32Array`, `Array` and friends are the same intrinsics the assertions
+ * use. The blob's module-level state is supplied as named parameters, and the
+ * returned object exposes each of them through a live accessor so a test can
+ * observe what a command mutated.
+ *
+ * @param names declarations to lift, in dependency order
+ * @param scope free variables the lifted code closes over — module-level state
+ *   in the blob, plus explicit stubs for the DOM/GL seams it calls out to
+ */
+export function loadDecls(
+  names: readonly string[],
+  scope: Record<string, unknown> = {},
+): Record<string, any> {
+  const params = Object.keys(scope);
+  const src = names.map((n) => extractDecl(n)).join('\n\n');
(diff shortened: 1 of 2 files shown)


BIMvoice:
Verified the section-axis observation independently. The asymmetry is what makes it worth your attention rather than a nitpick.

`packages/viewer/src/server.ts:259` validates `action` and nothing else, so `{"action":"section","axis":"w"}` is accepted with a 200 and forwarded to the browser. In the blob, the axis is then consumed **two different ways**:

```js
// viewer-html.ts:1060 — position: a defaulting ternary
const axisIdx = axis === 'x' ? 0 : axis === 'z' ? 2 : 1;   // unknown axis -> 1 (y)

// viewer-html.ts:1073-1075 — normal: three independent equality tests
axis === 'x' ? 1 : 0,
axis === 'y' ? 1 : 0,
axis === 'z' ? 1 : 0,                                       // unknown axis -> [0,0,0]
```

So an unrecognised axis silently falls back to Y for the *position* but produces a **zero normal**. The clip plane is degenerate: it neither cuts the model nor is visibly rejected. From the caller's side the request succeeded — 200, no error — and nothing happened. That is the same failure class as the clash-truncation and BCF-visibility findings from this campaign: a confident success response over a wrong answer.

I have not changed it, and I would not want it changed without your call, because there are at least two defensible fixes with different consequences: reject an unknown `axis` at the server boundary alongside `action`, or make the blob derive both position and normal from one resolution so they cannot disagree. The second is the more robust shape — the bug exists precisely because two code paths independently interpret the same input — but it is your design decision. Current behaviour is pinned in `pins the current behaviour for an unrecognised axis name` so whichever way you go, the change is visible.

**On the sweep itself, the scope claim is worth being precise about.** `viewer-html.ts` is one exported function whose body is a single template literal; roughly 1,000 of its lines are GL/DOM/shader source that genuinely needs a browser and is E2E territory, and about 330 lines are browser-free logic — `mat4`, bounds computation, chunk indexing, buffer merging, colour resolution and the pick codec. Only the second group is covered here. The 97% kill ratio is **targeted** and says nothing about the GL lines.

The mechanism avoids the usual trap: rather than reimplementing or mocking, `test/helpers/blob.ts` lifts the **real shipped source text** out of the emitted HTML and evaluates it with free variables supplied explicitly, and an `assertBrowserFree` guard fails the suite if a lifted declaration ever starts touching `document`/`gl`/`window`. So the tests cannot drift into passing against a copy of the code that no longer ships.

One survivor left deliberately untested: dropping the `Math.max(…, 0.01)` Y-extent floor. Model bounds are the union of entity bounds, so a zero-extent model bins every entity identically with or without the floor — the only discriminating input is a building under 3cm tall. Recorded as effectively equivalent and explicitly **not fully proven**, rather than closed with a contrived fixture.

BIMvoice:
## Round-four self-audit: this PR is the strongest of the round — 24 mutations, 23 KILLED

The lifted-source harness is the design decision under test here: it extracts the **real shipped source text** out of the emitted HTML rather than re-implementing it, so a mutation of `src/viewer-html.ts` should propagate straight into the tests. It does. Every mutation below was applied to the production blob source (count asserted `== 1`, region re-read, restored from a pre-mutation copy) against a baseline of 107 passing tests.

| Mutation in `src/viewer-html.ts` | Count | Result |
|---|---|---|
| `perspective`: swap the aspect divide onto `m[5]` | 1 | KILLED (2 tests) |
| `multiply`: swap the `a`/`b` roles (order flip) | 1 | KILLED |
| `invert`: remove the singular-determinant guard | 1 | KILLED |
| `transpose`: swap `m[6]` / `m[7]` | 1 | KILLED |
| `lookAt`: reverse the z axis (`center - eye`) | 1 | KILLED |
| `lookAt`: drop the `1/len` reciprocal on the x axis | 1 | KILLED |
| `getChunkIndex`: `>` → `>=` | 1 | KILLED (4 tests) |
| `getChunkIndex`: divisor always 3 | 1 | KILLED |
| `getChunkIndex`: default `isIndices = true` | 1 | KILLED (4 tests) |
| `mergeUint32` returns a `Float32Array` | 1 | KILLED |
| merge in reverse chunk order | 1 | KILLED (2 tests) |
| `resolveColor`: drop `.toLowerCase()` | 1 | KILLED |
| `matchesType`: match everything on an undefined query | 1 | KILLED |
| `getEntityBoundsForFilter`: swap the filter's arguments | 1 | KILLED (5 tests) |
| `getEntityBoundsForFilter`: fold `boundsMin` into the max | 1 | KILLED (4 tests) |
| `getEntityBoundsForFilter`: never return null | 1 | KILLED (3 tests) |
| pick colour: swap the red and green bytes | 1 | KILLED |
| `applyColorOverrides`: stride 3 instead of 4 | 1 | KILLED (2 tests) |
| section: map `z` to axis 1 | 1 | KILLED (3 tests) |
| section: drop the `/ 100` on a percentage position | 1 | KILLED (3 tests) |
| section: `center` uses only `boundsMax` | 1 | KILLED |
| `colorByStorey`: remove the 10-storey cap | 1 | KILLED |
| `colorByStorey`: remove the `Math.max(…, 0.01)` bin-size floor | 1 | **SURVIVED** |

### The one survivor — and why the test that should have caught it could not

*"survives a perfectly flat model without dividing by zero"* asserts `doesNotThrow` plus finite colour channels. At a zero Y extent the bin key is `Math.floor(0 / 0)` = **NaN**; every entity lands in that one group and the palette lookup yields finite channels **either way**. Both assertions are true with and without the floor — shape 4 of the catalogue, a property that holds regardless.

The floor earns its keep on a *nearly* flat model: a 1mm extent gives 0.33mm bins without it, splitting two entities 0.9mm apart into different storeys. dfd9d2c adds that case; the mutation is now KILLED (107/108).

No production code changed. 112 viewer-core tests pass, typecheck clean.

PR #2160, opened 2026-08-05
test(pointcloud,collab): close four survivors from the round-2 self-audit

BIMvoice:
## Summary

Round 2 of the self-audit of our own merged `test(...)` PRs. Six PRs sampled (none overlapping round 1): **#2151** (renderer), **#2150** (codegen), **#2145** (query/lens), **#2139** (cache/encoding/sdk/server-client), **#2135** (parser), **#2141** (collab/pointcloud).

**23 mutations applied, 19 killed, 4 survived.** All four survivors sit in files #2141 swept, and no production defect was found. Test files only — no production code changed, no changeset.

## Discrimination table

| PR | Test / property | Mutation | Result |
|---|---|---|---|
| #2151 | `projectedAabbRadiusPx` "uses view DEPTH, not Euclidean distance" | dot product → `Math.sqrt(cx²+cy²+cz²)` | KILLED (3 tests) |
| #2151 | "fails open at exactly depth === radius" | `depth <= radius` → `depth < radius` | KILLED |
| #2151 | "picks the nearest corner on EVERY axis" | `nx = viewDir.x >= 0 ? min : max` → `nx = unionMin[0]` | KILLED |
| #2151 | "ortho: Infinity (not NaN) on poisoned radius" | drop `!Number.isFinite(maxOccRadius)` guard | KILLED |
| #2151 | `selectEvictions` "stops the moment the excess is exactly met" | `excess <= 0` → `excess < 0` | KILLED |
| #2151 | `selectEvictions` min-age rule | `>= minAgeFrames` → `> minAgeFrames` | KILLED (2 tests) |
| #2150 | crc32 spec vectors (independent bit-at-a-time reference) | poly `0xedb88320` → `0xedb88321` | KILLED (5 tests) |
| #2150 | crc32 case-insensitivity contract | drop `str.toUpperCase()` | KILLED (5 tests) |
| #2150 | `findCollisions` "detects a deliberately-colliding pair" | `nameList.length > 1` → `> 2` | KILLED |
| #2135 | `detectSchemaVersion` ladder ORDER | swap the IFC4X3 and IFC4 lines | KILLED |
| #2135 | default fallback arm | `return 'IFC4'` → `return 'IFC2X3'` | KILLED |
| #2135 | header scan window | `Math.min(len, 2000)` → `Math.min(len, 60)` | KILLED (3 tests) |
| #2145 | `discoverDataSources` "budgets each type separately" | per-type counter → store-wide `sampleIds.length` | KILLED |
| #2145 | sampling cap boundary | `count < SAMPLE_SIZE` → `count <= SAMPLE_SIZE` | KILLED |
| #2145 | "prefers `getMaterialNames()` over `getMaterialName()`" (#1366) | invert the preference | KILLED |
| #2139 | parse-stream tail-buffer terminal check | drop `\|\| tail.type === 'error'` | KILLED |
| #2139 | "throws when the stream ends before a terminal event" | remove the `if (!terminated) throw` | KILLED (2 tests) |
| #2141 | LAS 1.4 exact-version count | `versionMinor >= 4` → `> 4` | KILLED |
| #2141 | `seedFromIfcx` reset clears entities | drop `ents.clear()` | KILLED |
| #2141 | `seedFromIfcx` **default** reset arm | `if (opts.reset)` → `if (opts.reset !== false)` | **SURVIVED** |
| #2141 | legacy classification masking | drop `& 0x1f` | **SURVIVED** |
| #2141 | format 6–10 classification offset | `formatId >= 6 ? 16 : 15` → `15` | **SURVIVED** |
| #2141 | `sampleMaxRgbChannel` "detects 8-bit-in-low-byte RGB" | delete the green and blue comparisons | **SURVIVED** |

Every "SURVIVED" row was re-checked against the whole package suite (`packages/pointcloud`: 16 files / 134 passed; `packages/collab`: 41 files / 173 passed), not just the one test file.

### 1. `seedFromIfcx` — the default arm was never exercised

`from-ifcx-reset.test.ts` pins `reset: true` and `reset: false`, both passed **explicitly**. `opts.reset` absent is a third arm and it was unpinned, so widening the guard to `if (opts.reset !== false)` survives all 41 collab test files.

That arm is the one production uses. `apps/viewer/src/store/slices/collabSlice.ts:597` seeds an **already-live** session doc with `seedFromIfcx(session.doc, bytes)`, and `packages/collab/src/snapshot/worker.ts:92` calls `seedFromIfcx(doc, file)`. Under the mutant both wipe the doc they were asked to seed into — silent data loss that every peer converges on.

New test drives both routes to "absent": no third argument at all, and an explicitly empty `{}` (the default lives in the `opts.reset` read, not in the `= {}` parameter default).

### 2. `& 0x1f` on the legacy classification byte

Every fixture in `las.test.ts` stored a classification below 32 with the high three bits of byte 15 at zero, which makes the mask an **identity on every input the suite feeds it**. Deleting it survived the whole package.

Formats 0–5 pack synthetic (0x20) / key-point (0x40) / withheld (0x80) into those bits, and real producers set them. A withheld ground point is byte 15 = `0x82` and decodes as class **130** without the mask — every classification colour ramp and filter misses it.

New test uses four flag/class combinations and asserts each individually (a mask that dropped only some of the flag bits would still satisfy a whole-array comparison against one wrong value), including class 31 with all three flags set to pin the mask from the other side — it must keep all five class bits, not narrow the field further.

### 3. Formats 6–10 were never decoded

No fixture ever built a format ≥ 6 record, so `classOffset = formatId >= 6 ? 16 : 15` collapsed to a constant `15` survives, and so does always applying `& 0x1f` (LAS 1.4 extended classification is a full unmasked byte, classes 0..255 are legal). New `buildFormat6Records` helper plus a test that reads class 200 out of byte 16 while byte 15 holds a decoy value of 7, and a second row using class 40 — `40 & 0x1f === 8`, so it separates "masked" from "not masked" independently of the first row.

### 4. `sampleMaxRgbChannel` only had its red channel pinned

The existing 8-bit-detection test uses `r=255, g=128, b=64` — the maximum is in **red**, so deleting the green and blue comparisons from the sampler's inner loop survives it, and the whole suite. A red-only sampler under-reports the max on any file whose brightest channel is green or blue; the auto-detect then concludes "already 16-bit" and an 8-bit file renders ~257x too dark.

New tests put the max in green, in blue, and in red; put a 16-bit value on the **last** record (so a sampler that stops early is visible); and pin the no-RGB early return.

## Method

Per mutation: `python3` exact-substring replacement with the replacement count asserted `== 1`, the mutated region re-read and confirmed changed before believing any result, then restored by copying back a backup taken **before** the file's first mutation (never `git checkout --`). Each new test was confirmed RED under its stated mutation and GREEN after restore.

## Test plan

- [x] `packages/pointcloud`: 16 files / 134 passed → 16 files / **139 passed**
- [x] `packages/collab`: 41 files / 173 passed → 41 files / **174 passed**
- [x] Each of the 4 survivors' mutations re-applied against the new tests and confirmed to fail; restored by file copy; green again
- [x] `turbo typecheck --filter=@ifc-lite/pointcloud --filter=@ifc-lite/collab`
- [x] No production code changed — test files only, no changeset

### DIFF
--- a/packages/collab/test/from-ifcx-reset.test.ts
+++ b/packages/collab/test/from-ifcx-reset.test.ts
@@ -68,6 +68,39 @@ describe('seedFromIfcx reset contract', () => {
     expect(entitiesMap(doc).has('new-wall')).toBe(true);
   });
 
+  // The two cases above only ever pass `reset` EXPLICITLY. The default —
+  // `opts.reset` absent — is a third arm, and it is the one production uses
+  // most: `apps/viewer` seeds an already-live session doc with
+  // `seedFromIfcx(session.doc, bytes)` (collabSlice.ts) and
+  // `snapshot/worker.ts` calls `seedFromIfcx(doc, file)`. Widening the guard
+  // to `if (opts.reset !== false)` makes that call wipe the doc it was asked
+  // to seed into — silent data loss that converges across every peer — and
+  // it survived the whole collab suite before this test existed.
+  it('defaults to reset OFF: seeding with no options is additive, not destructive', () => {
+    const doc = createCollabDoc();
+    doc.transact(() => {
+      createEntity(doc, 'old-wall', { ifcClass: 'IfcWall' });
+      createRelationship(doc, 'old-rel', { ifcClass: 'IfcRelAggregates', source: 'old-wall' });
+      createGeometry(doc, 'old-geom', { type: 'parametric', source: 'extruded-area-solid' });
+    });
+
+    // No third argument at all — the shape both production call sites use.
+    seedFromIfcx(doc, minimalIfcx(['new-wall']));
+
+    expect(entitiesMap(doc).has('old-wall')).toBe(true);
+    expect(relationshipsMap(doc).has('old-rel')).toBe(true);
+    expect(geometryMap(doc).has('old-geom')).toBe(true);
+    expect(entitiesMap(doc).has('new-wall')).toBe(true);
+
+    // An explicitly empty options object must behave identically — the
+    // default lives in the `opts.reset` read, not in the `= {}` parameter
+    // default, so both routes to "absent" need pinning.
+    seedFromIfcx(doc, minimalIfcx(['newer-wall']), {});
+    expect(entitiesMap(doc).has('old-wall')).toBe(true);
+    expect(entitiesMap(doc).has('new-wall')).toBe(true);
+    expect(entitiesMap(doc).has('newer-wall')).toBe(true);
+  });
+
   it('mergeBranch("layer") does not reset the parent — content added to the parent after the fork survives', async () => {
     const parent = await createCollabSession({
       roomId: 'parent-reset-check',
--- a/packages/pointcloud/src/formats/las.test.ts
+++ b/packages/pointcloud/src/formats/las.test.ts
@@ -48,7 +48,15 @@ function buildHeader(overrides: Partial<{
   return { header: new Uint8Array(buf), total: 227 };
 }
 
-function buildFormat0Records(rows: Array<{ x: number; y: number; z: number; intensity?: number; classification?: number }>): Uint8Array {
+function buildFormat0Records(rows: Array<{
+  x: number; y: number; z: number; intensity?: number; classification?: number;
+  /**
+   * High 3 bits of byte 15 in formats 0–5: synthetic (0x20), key-point
+   * (0x40), withheld (0x80). Real producers set these on ordinary points,
+   * so the decoder has to mask them off the classification.
+   */
+  classFlags?: number;
+}>): Uint8Array {
   const buf = new ArrayBuffer(rows.length * 20);
   const view = new DataView(buf);
   for (let i = 0; i < rows.length; i++) {
@@ -57,8 +65,33 @@ function buildFormat0Records(rows: Array<{ x: number; y: number; z: number; inte
     view.setInt32(off + 4, rows[i].y, true);
     view.setInt32(off + 8, rows[i].z, true);
     view.setUint16(off + 12, rows[i].intensity ?? 0, true);
-    view.setUint8(off + 15, rows[i].classification ?? 0);
+    view.setUint8(off + 15, (rows[i].classification ?? 0) | (rows[i].classFlags ?? 0));
     // bytes 13, 14, 16, 17, 18, 19 = bit flags / scan angle / user data — leave 0
+  }
+  return new Uint8Array(buf);
+}
+
+/**
+ * Format-6 records (LAS 1.4 extended base record, 30 bytes). Two things
+ * differ from formats 0–5 and both are load-bearing in `decodeLasPoints`:
+ * classification moves from byte 15 to byte 16, and it is NOT masked (the
+ * flag bits got their own byte). `flagsByte15` writes whatever the producer
+ * left in the old slot so a decoder that still reads byte 15 is visible.
+ */
+function buildFormat6Records(rows: Array<{
+  x: number; y: number; z: number;
+  intensity?: number; classification: number; flagsByte15?: number;
+}>): Uint8Array {
+  const buf = new ArrayBuffer(rows.length * 30);
+  const view = new DataView(buf);
+  for (let i = 0; i < rows.length; i++) {
+    const off = i * 30;
+    view.setInt32(off, rows[i].x, true);
+    view.setInt32(off + 4, rows[i].y, true);
+    view.setInt32(off + 8, rows[i].z, true);
+    view.setUint16(off + 12, rows[i].intensity ?? 0, true);
+    view.setUint8(off + 15, rows[i].flagsByte15 ?? 0);
+    view.setUint8(off + 16, rows[i].classification);
   }
   return new Uint8Array(buf);
 }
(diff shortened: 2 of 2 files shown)


BIMvoice:
Verified the first survivor independently, because it is the one with a real consequence.

`seedFromIfcx(doc, input, opts = {})` has three arms, and the test pinned only two of them — `reset: true` and `reset: false`, both passed **explicitly**. The third is the absent-option case, and that is the one production actually uses:

```
apps/viewer/src/store/slices/collabSlice.ts:597
    if (bytes && bytes.length > 0) collabMod.seedFromIfcx(session.doc, bytes);

packages/collab/src/snapshot/worker.ts:92
    seedFromIfcx(doc, file);

packages/collab/README.md:38
    seedFromIfcx(session.doc, buffer);
```

Two production call sites and the documented usage, all two-argument. So mutating `if (opts.reset)` to `if (opts.reset !== false)` — replacements=1 — flips exactly the arm nobody tested, and both call sites then **clear the document they were asked to seed into**. In a CRDT session that is not a local glitch: every peer converges on the wiped state, and it looks like a clean successful merge. The whole `packages/collab` suite stayed green (41 files, 173 passed).

That is a two-valued signal tested in both of its explicit directions while the default went untested — a variant of the shape we keep finding, and worth naming separately because "both directions covered" was literally true here and still insufficient.

The other three survivors are all in `packages/pointcloud` and share one cause — **every fixture is built from values where the operation is an identity**:

- **`& 0x1f`** — every fixture stores a class below 32 with the flag bits clear, so the mask changes nothing on any input the suite feeds it. A withheld ground point (byte 15 = `0x82`) decodes as class 130 without it.
- **format 6-10 class offset** — no fixture ever constructs a record with format ≥ 6, so collapsing `>= 6 ? 16 : 15` to a constant survives.
- **`sampleMaxRgbChannel`** — the 8-bit-detection fixture is `r=255, g=128, b=64`, so the maximum sits in red and a red-only sampler passes. A real 8-bit file whose brightest channel is green or blue is then mis-detected as 16-bit and renders roughly 257x too dark.

**Round 2 scoreboard: 23 mutations, 19 killed, 4 survived, no production defect** — all four are test gaps, closed with tests only. Combined with round 1 that is **7 genuine survivors in 47 mutations across 12 of our own merged PRs**, and one real production defect (the shallow-spread privilege escalation in `packages/mcp/src/auth/scope.ts`, on #2158).

Two method notes worth recording, both of which could have produced false results:

- `packages/parser` needs `turbo build --filter=@ifc-lite/parser^...` first; without it the run prints `Test Files 1 failed` alongside `Tests no tests`, which reads like a pass. Same trap as `@ifc-lite/source-dalux`.
- `packages/renderer` has a **pre-existing, unrelated** failure in `renderer-render-paths.test.ts` on clean `upstream/main`. Runs were scoped to the target files and the baseline established first, so it did not contaminate any verdict — but it means a bare `renderer` run is red on main right now, independent of anything here.

BIMvoice:
A correction and an escalation on the typecheck coverage I raised earlier, both verified.

**First, a refutation of something I reported.** I said `packages/renderer` was failing on clean `upstream/main`. It is not. On a freshly built worktree the suite is **422 tests, 422 pass, 0 fail**, and the suspect file alone is 27/27. The original report came from a worktree where `pnpm turbo build --filter=@ifc-lite/renderer^...` had not run, so `@ifc-lite/spatial/dist/index.js` did not exist and node's runner reported the whole *file* as one failing test — `# tests 1 # fail 1`, which reads like a targeted assertion failure rather than a missing build. My error, retracted.

**Second, the thing that turned up while checking it, which is worse than what I originally claimed.** I had reported that renderer test files carry type errors CI never sees because the root `tsconfig.json` excludes `**/*.test.ts`. That is true but it is only one of two layers:

```
$ turbo.json  ->  "typecheck": { "dependsOn": ["^build"], "outputs": [] }
$ package.json ->  "typecheck": "turbo typecheck"

packages/apps defining a `typecheck` script:
  @ifc-lite/viewer        tsc --noEmit
  @ifc-lite/viewer-embed  tsc --noEmit

total packages + apps with a package.json:  40
```

**Two of forty.** `pnpm typecheck` runs `turbo typecheck`, and `turbo` only invokes the task in packages that define it — so 38 of 40 are not typechecked at all, by any config. `packages/renderer` has no `typecheck` script, which is why `.github/workflows/test.yml:276` never sees its errors regardless of the tsconfig exclusion.

This reframes something I have been saying all day. Every "`pnpm typecheck`: 34/34 tasks successful" in my PR descriptions was counting `^build` dependencies plus **two** real typecheck tasks. It was never evidence that a package's sources typecheck — only that everything built. I would rather flag that than let those claims stand as stronger than they were.

The confirmed errors, from compiling `src/**/*` with the test exclusion lifted — exactly 7, all in `packages/renderer`:

```
camera-controls.test.ts:66        TS2741  orbitAnchorBounds missing from CameraInternalState
lod-simplify.test.ts:114,115,116  TS7022  e0/e1/e2 implicitly any (self-referential initializer)
renderer-render-paths.test.ts:9   TS2305  './types.js' has no exported member 'MeshData'
scene-finalize-rollback.test.ts:8 TS2305  './types.js' has no exported member 'RenderPipeline'
scene-textured-origin.test.ts:25  TS2305  './types.js' has no exported member 'MeshData'
```

`MeshData` moved to `packages/geometry/src/types.ts:31` and production already imports it from `@ifc-lite/geometry` (`renderer/src/bvh.ts:1`, `src/index.ts:99`). These are `import type` specifiers, so tsx erases them — which is exactly why the runtime suite is green while the types are wrong.

Nothing changed here; this is a report. The fix has two halves and only the second one prevents recurrence: repoint the three stale imports, **and** add `typecheck` scripts so the packages are actually in the lane. The second half is a repo-wide decision about how much it would surface at once, so it is yours to scope — happy to do it as a properly bounded PR if you want it, or to stop at the three imports.

BIMvoice:
## Round-four self-audit: this PR holds — 12 of 12 mutations KILLED

Re-measured with the same method (exact-substring replacement, `assert n == 1` on the count, mutated region re-read from disk, restore from a pre-mutation copy). Baselines established first on a built workspace: `@ifc-lite/pointcloud` 139 passed / 2 skipped (16 files), `@ifc-lite/collab` 174 passed / 2 skipped (41 files).

Because this PR is itself a fix for round-two survivors, a vacuous result here would have invalidated the audit chain. It does not.

| Mutation (production source) | Count | Result | Test that failed |
|---|---|---|---|
| `las.ts` — drop `& 0x1f` on the legacy classification byte | 1 | KILLED | masks the synthetic/key-point/withheld flag bits |
| `las.ts` — narrow the mask to `& 0x0f` | 1 | KILLED | same (the class-31 assertion catches the opposite direction, as intended) |
| `las.ts` — `classOffset = 15` (collapse the format-6 branch) | 1 | KILLED | reads format-6+ classification from byte 16 |
| `las.ts` — mask the *extended* classification too | 1 | KILLED | same |
| `las.ts` — `classOffset` boundary `>= 6` → `> 6` | 1 | KILLED | same |
| `las.ts` — mask-branch boundary `>= 6` → `> 6` | 1 | KILLED | same |
| `sampleMaxRgbChannel` — red only (drop green + blue) | 1 | KILLED | takes the max over ALL THREE channels **and** scans every sampled record |
| `sampleMaxRgbChannel` — drop blue only | 1 | KILLED | takes the max over ALL THREE channels |
| `sampleMaxRgbChannel` — `i < total` → `i < 1` | 1 | KILLED | scans every sampled record, not only the first |
| `from-ifcx.ts` — `if (opts.reset)` → `if (opts.reset !== false)` | 1 | KILLED | defaults to reset OFF |
| `from-ifcx.ts` — reset skips `geom.clear()` | 1 | KILLED | reset: true clears entities, relationships and geometry |
| `from-ifcx.ts` — parameter default `= {}` → `= { reset: true }` | 1 | KILLED | defaults to reset OFF |

Two details worth recording, because they are the difference between a real test and a passing one:

- Every mutation above kills **exactly one** test (the sampler red-only mutation kills two, both of which name the property). None is a blast-radius kill, so each is probative of the property it targets.
- The last row is why the `seedFromIfcx(doc, ifcx, {})` assertion in the new test earns its place: it is the only assertion that separates the `= {}` parameter default from the `opts.reset` read. Dropping it would leave that route unpinned.

No production code changed, and no fix was needed on this branch.

PR #2161, opened 2026-08-05
test(geometry): close 16 mutation-verified gaps in the coordinate and shard-decoder paths

BIMvoice:
Second, deeper mutation sweep over `packages/geometry`. The first pass (#2134, #2137) was shallow relative to what later passes found elsewhere, and geometry is where a wrong answer is silent — a wrong mesh or a wrong transform, not an exception — feeding the renderer, clash and export paths.

**Test-only. No production code changed.**

## Result

| | |
|---|---|
| Baseline | 24 files / 225 tests, all green |
| After | 26 files / 295 tests, all green |
| Mutations | 30 + 2 controls |
| Killed by the existing suite | 12 / 30 (**40%**, TARGETED — aimed at the untested modules) |
| Survivors | 18 |
| Survivors now killed | 16 |
| Survivors proved equivalent | 2 |

Controls (run before any survivor was believed, both died): interleaved-buffer stride `i * 6` → `i * 3`; centroid `(min.x + max.x) / 2` → `min.x`.

The survivors clustered almost perfectly in the three modules that had **no test file at all**: `coordinate-handler.ts`'s untested branches, `packed-geometry-decoder.ts`, and `geometry-coordinate.ts`.

## Gaps closed

Each row: the mutation that survived the old suite, and the test that now kills it. All replacement counts asserted `== 1`, each mutated region re-read and confirmed before the run.

### `coordinate-handler.ts`

| Mutation | What a user would see | Killing test |
|---|---|---|
| `positions[i] + ox` → `positions[i]` in `calculateBounds` | Every bounds fixture left `origin` absent, so the per-element local-frame fold (`world = origin + position`) was exercised by nothing. Without it a model whose elements each carry an origin collapses to a box around zero — the camera fits the wrong volume. | `folds each mesh origin into world bounds`, `applies the origin components axis-by-axis` |
| same fold dropped in `calculateBoundsFast` | Same, on the desktop-native streaming path. | `folds each mesh origin into the sampled bounds` |
| delete the corrupted-vertex clamp in `shiftPositions` | Nothing asserted the `else` branch. A `NaN` or `1e30` vertex reaches the GPU buffer, exploding the mesh's box and able to blank the model. | `clamps non-finite and out-of-range vertices to the shifted origin`, `scrubs corrupted vertices even when no shift is needed` |
| `maxCoord > THRESHOLD` → `>=` in `needsShift` | Both existing cases sat far from 10 km, so the boundary was free. | `treats exactly the 10km threshold as not needing a shift` |
| drop `\|\| maxSize > THRESHOLD` from the incremental shift trigger | Every fixture was *far from the origin*, so the extent half of the OR was dead weight. A 40 km site straddling the origin never gets shifted and keeps float precision loss across the whole model. | `shifts a model that straddles the origin but spans more than 10km` |
| `(small / total) > 0.5` → `>=` in WASM-RTC detection | No fixture sat on the boundary. `>=` there silently suppresses a shift the model needs. | `does not claim WASM RTC on an exact 50/50 split` (+ the clear-majority case) |
| delete `originShift = {0,0,0}` from `processTrustedMeshesIncremental` | **Zero-fixture cancellation:** every trusted test starts from a fresh handler where the shift is *already* zero, so the reset was a no-op in the fixture and the assertion `originShift).toEqual({x:0,y:0,z:0})` proved nothing. In production the trusted path can follow a shifted browser batch. | `clears a shift established by an earlier non-trusted batch` |
| disable the sampled-bounds fast path | The documented ~150x speedup (first + last vertex only) was pinned by nothing; a "fix" to a full scan would have been invisible. | `samples first and last vertex only once WASM RTC is confirmed` |

### `packed-geometry-decoder.ts` — no test file existed

| Mutation | Impact | Killing test |
|---|---|---|
| mesh-table stride `11` → `12` words | Renderer draws the wrong vertices onto the wrong element. Silent. | `splits the pools at the documented mesh-table stride and section offsets` |
| `normalsOffset = positionsOffset + positionsByteLength` → `positionsOffset` | Normals read from the positions pool — wrong shading, no error. | same |
| remove the magic check | A foreign payload is decoded as garbage geometry instead of throwing. | `rejects a payload whose magic is not the shard magic` |

New `packed-geometry-decoder.test.ts` builds its fixture from the **layout documented on the decoder**, not from the decoder itself, so the two are independent statements of the format. Two meshes at different non-zero pool offsets with different lengths, so neither the stride nor a section offset can come out right by accident. Also covers `toArrayBuffer`'s four payload shapes and the version guard.

### `geometry-coordinate.ts` — no test file existed

| Mutation | Impact | Killing test |
|---|---|---|
| first bracket of the streaming batch ladder `100` → `200` | Whole 10/50/100/300/500 MB ladder unpinned. | `maps %s MB to a batch of %s` (both sides of every bracket) |
| `origin[0] \|\| origin[1] \|\| origin[2]` → `&&` | Drops every axis-aligned local origin — e.g. `(0, 0, 5000)` — so the element renders at the world origin instead of its placement. | `keeps a local origin with only some axes set` |

New `geometry-coordinate.test.ts` also covers `convertMeshCollectionToBatch`'s free-on-throw contract, `localBounds`/`localToWorld` arity guards, embedded vs. external texture precedence, the `geometryClass` default, and by-express-id (not positional) fingerprint attribution.

### `packed-instanced-decoder.ts`

| Mutation | Impact | Killing test |
|---|---|---|
| swap the template origin's `getFloat64` reads | **Zero-fixture cancellation:** the Rust conformance fixture's origins are all `(0,0,0)`, so any permutation of the three reads is invisible. | `decodes a non-zero template origin axis-by-axis` |
| remove the template pool-bounds guard | A malformed offset silently clips (`subarray` saturates) into truncated geometry indistinguishable from a real occurrence — exactly what the guard's comment says it exists to prevent. | three `rejects a template whose … runs past the … pool` tests |
| remove the `templateIndex >= templates.length` guard | Instance references a template that does not exist. | `rejects an instance that references a template the shard does not carry` |

## Survivors that are equivalent, not gaps

- **`byteOffset === 0` conjunct in `toArrayBuffer`** — redundant: `byteOffset > 0` implies `byteLength ≤ buffer.byteLength − byteOffset < buffer.byteLength`, so the second conjunct already fails. Proven.
- **the `activeThreshold` ternary in `processMeshesIncremental`** — when `wasmRtcDetected` is true, `calculateBounds` discards the argument (it takes the sampling fast path), and the `shiftPositions` call site is unreachable because a non-zero `originShift` together with `wasmRtcDetected` cannot be produced through the public API: the detection branch never sets a shift, and `processTrustedMeshesIncremental` zeroes it. Proven by reachability over the only two read sites.

## Measured and holding — no changes needed

`geometry-fingerprints.ts` and `batch-sizing.ts` killed **8 of 8**: `AABB_STRIDE` 6→3, min/max order in `geometryAabbAt` **and** in its paired writer `writeGeometryAabbAt`, dropping the non-positive-volume refusal, `Math.min(ids, values)` → `Math.max`, and all three `resolveBatchSizing`/`nextAdaptiveBatchJobs` clamps. These two modules are genuinely well covered.

Reported whether or not changed:

1. **`should accumulate trusted native bounds without applying a shift`** — asserts `originShift` equals zero starting from a handler whose `originShift` is already zero. The assertion holds whether or not the code does the thing. Now backed by a test that establishes a non-zero shift first.
2. **`expand-to-flat` in `packed-instanced-decoder.test.ts`** — folds `tmpl.origin[0..2]` into the point before applying the transform, but the fixture's origins are all `(0,0,0)`, so the origin decode cancels entirely. Left in place (it is a genuine Rust↔TS conformance test); the axis-by-axis origin test now covers what it could not.
3. **Every `calculateBounds` / trusted-bounds fixture** omitted `origin`, making the fold in both bounds paths untestable — the same zero-fixture shape, across seven tests.

## Verification

- `pnpm turbo build --filter=@ifc-lite/geometry^...` before every run.
- `pnpm typecheck` does **not** cover this package, and the root tsconfig excludes `**/*.test.ts` — so the four test files were typechecked separately against the package's own options with the exclusion lifted (`tsc -p` with `exclude` reduced to `node_modules`/`dist`, `--listFiles` confirming all 49 test files were in the program). **Zero errors in the four files touched here.** 17 pre-existing errors in six other geometry test files (`geometry-native*`, `wasm-url-plumbing`, `prepass-class-spans`, `geometry-parallel-stream-end`, `diagnostics`) are untouched by this PR and predate it — they are invisible to CI for the same reason.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for geometry bounds, coordinate handling, rotations, mesh conversion, and resource cleanup.
  * Added validation for packed and instanced geometry decoding, including malformed, empty, and boundary-case data.
  * Added tests for textures, colors, transforms, origins, geometry fingerprints, volumes, and payload conversion.

### DIFF
--- a/packages/geometry/src/geometry-coordinate.test.ts
+++ b/packages/geometry/src/geometry-coordinate.test.ts
@@ -1,0 +1,384 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Streaming batch sizing + WASM MeshCollection → MeshData conversion.
+ *
+ * `convertMeshCollectionToBatch` is the single-threaded mesh-extraction path
+ * (`index.ts`); everything it drops on the floor — a per-element origin, a
+ * local box, a placement matrix, a texture — is lost silently, showing up only
+ * as geometry in the wrong place or shaded wrong.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+
+import {
+  getStreamingBatchSize,
+  convertMeshCollectionToBatch,
+  withBuildingRotation,
+} from './geometry-coordinate.js';
+import type { CoordinateInfo } from './types.js';
+
+describe('getStreamingBatchSize', () => {
+  const buffer = new Uint8Array(0);
+
+  it('returns a numeric config verbatim, ignoring the buffer', () => {
+    expect(getStreamingBatchSize(new Uint8Array(500 * 1024 * 1024), 7)).toBe(7);
+  });
+
+  // Every bracket asserted with a value on each side of its boundary — a
+  // single sample per bracket lets one lucky case carry the whole ladder.
+  it.each([
+    [0, 100],
+    [9.9, 100],
+    [10, 200],
+    [49.9, 200],
+    [50, 300],
+    [99.9, 300],
+    [100, 500],
+    [299.9, 500],
+    [300, 1500],
+    [499.9, 1500],
+    [500, 3000],
+    [5000, 3000],
+  ])('maps %s MB to a batch of %s', (fileSizeMB, expected) => {
+    expect(getStreamingBatchSize(buffer, { fileSizeMB })).toBe(expected);
+  });
+
+  it('derives the file size from the buffer when the config omits it', () => {
+    // 60 MB → the 50–100 MB bracket.
+    expect(getStreamingBatchSize(new Uint8Array(60 * 1024 * 1024), {})).toBe(300);
+  });
+
+  it('treats a zero fileSizeMB as absent and measures the buffer', () => {
+    // `fileSizeMB: 0` is falsy, so the buffer length decides: 200 MB → 500.
+    expect(
+      getStreamingBatchSize(new Uint8Array(200 * 1024 * 1024), { fileSizeMB: 0 })
+    ).toBe(500);
+  });
+});
+
+describe('withBuildingRotation', () => {
+  const base: CoordinateInfo = {
+    originShift: { x: 0, y: 0, z: 0 },
+    originalBounds: { min: { x: 0, y: 0, z: 0 }, max: { x: 1, y: 1, z: 1 } },
+    shiftedBounds: { min: { x: 0, y: 0, z: 0 }, max: { x: 1, y: 1, z: 1 } },
+    hasLargeCoordinates: false,
+  };
+
+  it('attaches the rotation without mutating the input', () => {
+    const out = withBuildingRotation(base, 1.25);
+
+    expect(out.buildingRotation).toBe(1.25);
+    expect(out).not.toBe(base);
+    expect(base).not.toHaveProperty('buildingRotation');
+  });
+
+  // Zero is a meaningful rotation (explicitly north-aligned) and is falsy;
+  // only `undefined` means "not resolved".
+  it('attaches a zero rotation', () => {
+    expect(withBuildingRotation(base, 0).buildingRotation).toBe(0);
+  });
+
+  it('returns the input untouched when the rotation is undefined', () => {
+    expect(withBuildingRotation(base, undefined)).toBe(base);
+  });
+});
+
+// ── convertMeshCollectionToBatch ──
+
+interface FakeMesh {
+  expressId: number;
+  ifcType: string;
+  positions: Float32Array;
+  normals: Float32Array;
+  indices: Uint32Array;
+  color: Float32Array;
+  free: () => void;
+  [k: string]: unknown;
+}
+
+function fakeMesh(overrides: Partial<FakeMesh> & { expressId: number }): FakeMesh {
+  return {
+    ifcType: 'IfcWall',
+    positions: new Float32Array([0, 0, 0]),
+    normals: new Float32Array([0, 0, 1]),
+    indices: new Uint32Array([0]),
+    color: new Float32Array([1, 1, 1, 1]),
+    free: vi.fn(),
+    ...overrides,
+  } as FakeMesh;
+}
+
+function fakeCollection(meshes: (FakeMesh | null)[], extra: Record<string, unknown> = {}) {
+  return {
+    length: meshes.length,
+    get: (i: number) => meshes[i],
+    free: vi.fn(),
+    ...extra,
+  };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const asCollection = (c: unknown) => c as any;
+
+describe('convertMeshCollectionToBatch', () => {
+  it('frees every mesh and the collection itself', () => {
+    const meshes = [fakeMesh({ expressId: 1 }), fakeMesh({ expressId: 2 })];
+    const collection = fakeCollection(meshes);
+
+    convertMeshCollectionToBatch(asCollection(collection));
+
+    expect(meshes[0].free).toHaveBeenCalledTimes(1);
+    expect(meshes[1].free).toHaveBeenCalledTimes(1);
+    expect(collection.free).toHaveBeenCalledTimes(1);
+  });
+
+  it('frees the collection even when a mesh getter throws', () => {
+    const collection = fakeCollection([], {
+      length: 1,
+      get: () => {
+        throw new Error('boom');
+      },
+    });
+
+    expect(() => convertMeshCollectionToBatch(asCollection(collection))).toThrow('boom');
+    expect(collection.free).toHaveBeenCalledTimes(1);
+  });
+
+  it('skips null mesh slots', () => {
+    const batch = convertMeshCollectionToBatch(
+      asCollection(fakeCollection([null, fakeMesh({ expressId: 9 })]))
+    );
+
+    expect(batch.map((m) => m.expressId)).toEqual([9]);
+  });
+
+  it('copies the colour out of the WASM view into a plain tuple', () => {
+    const color = new Float32Array([0.1, 0.2, 0.3, 0.4]);
+    const batch = convertMeshCollectionToBatch(
+      asCollection(fakeCollection([fakeMesh({ expressId: 1, color })]))
+    );
+
+    expect(batch[0].color).toEqual([
+      color[0], color[1], color[2], color[3],
+    ]);
+    expect(Array.isArray(batch[0].color)).toBe(true);
+  });
+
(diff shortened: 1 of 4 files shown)


BIMvoice:
Verified the headline gap both ways, on `f6bbfda4`.

Dropping the origin fold from `calculateBounds` entirely (`positions[i] + ox` → `positions[i]`, replacements=1):

```
against main's fixtures      ->  25 files, 242 tests, 242 pass, 0 fail   <- SURVIVED
against this PR's fixtures   ->  × folds each mesh origin into world bounds
                                 × applies the origin components axis-by-axis
                                 2 failed | 293 passed (295)
```

Restored: 295/295.

So on `main` the origin fold in `calculateBounds` could be **deleted outright** and nothing would notice. The cause is not a missing test — it is the fixtures: seven `calculateBounds`/trusted-bounds fixtures all omit `origin`, so `ox/oy/oz` are `0` on every input the suite feeds it and the addition is an identity. The same omission disabled the fold in `calculateBoundsFast` too, which is why M1 and M2 both survived.

**That is the third distinct instance of this exact shape today**, and it is worth naming as a class rather than three coincidences: a fixture constant that makes the operation under test an identity. The others were three `packages/pointcloud` fixtures storing classification values below the `& 0x1f` mask threshold, and every `packages/solar` fixture sitting at longitude 0 where the longitude term vanishes. In each case the test *exercised* the line and asserted nothing about it. Coverage tooling would score all of them as covered.

The corollary is the practical one: when writing a fixture, the question is not "does this exercise the path" but "would this input distinguish the implementation from a no-op". A zero, an identity, a symmetric input or a default is where that distinction quietly disappears.

Two other things in this PR worth surfacing:

**A pre-existing test that asserted a tautology.** `should accumulate trusted native bounds without applying a shift` asserted `originShift === 0` on a handler whose shift was *already* zero — it held whether or not the code did the thing. It is now backed by a test that establishes a non-zero shift first.

**Two survivors deliberately left unkilled, both proven equivalent rather than argued away.** `byteOffset === 0` is redundant because `byteOffset > 0` already implies `byteLength < buffer.byteLength`, so the second conjunct cannot be reached independently. And `activeThreshold`'s ternary has exactly two read sites: `calculateBounds` discards it on the fast path, and `shiftPositions` cannot be reached with `wasmRtcDetected && originShift ≠ 0` through the public API. No tests were manufactured for either.

**`geometry-fingerprints.ts` and `batch-sizing.ts` came back genuinely well covered** — 8 of 8 mutations killed, including the paired reader/writer min-max symmetry and all three clamps. Nothing pushed for them. Worth stating because a sweep that only reports what it fixed gives no sense of what the existing suite is already worth.

Separately: 17 pre-existing typecheck errors live in six other geometry test files (`geometry-native*`, `wasm-url-plumbing`, `prepass-class-spans`, `geometry-parallel-stream-end`, `diagnostics`), invisible for the same reason as the renderer ones — `packages/geometry` defines no `typecheck` script, so `turbo typecheck` never invokes it. Untouched here.

BIMvoice:
## CodeRabbit triage — three threads on `packages/geometry/src/geometry-coordinate.test.ts`

All three verified against the source before acting. All three were right; all three applied in 9e70ba53.

Baseline before: **26 files / 295 tests**. After: **26 files / 296 tests**. No existing assertion weakened.

---

### :28 — "Remove the 500 MiB test allocation" — **APPLIED** ✅

Confirmed. `getStreamingBatchSize` returns at `packages/geometry/src/geometry-coordinate.ts:30-32` on the `typeof batchConfig === 'number'` branch; the first and only read of `buffer` is at `:36`, on the other side of that return. The 500 MiB `Uint8Array` was unreachable state — pure memory pressure in parallel vitest workers, which is not free on a loaded machine.

Now passes the shared zero-length `buffer`. The test keeps its teeth: mutating the early return away (`if (false as boolean)`) makes it fail with `expected 100 to be 7`. No buffer length can ever produce 7 — every measured bracket returns one of 100/200/300/500/1500/3000 — so the assertion, not the allocation size, was always what carried the "ignores the buffer" claim. Noted in a comment so the allocation does not grow back.

### :148 — "Test the per-mesh error cleanup path" — **APPLIED** ✅

Confirmed, and this was the real find. The test named `frees the collection even when a mesh getter throws` used `get: () => { throw }`, so `collection.get(i)` threw at `geometry-coordinate.ts:68` — before any mesh existed. The inner `try`/`finally { mesh.free() }` at `:71`/`:158-160` was never entered. The test asserted only `collection.free()` while its name implied per-mesh cleanup: it asserted strictly less than it claimed.

Two changes:
- Renamed it to `frees the collection when the collection getter itself throws`, which is what it actually covers.
- Added `frees the mesh AND the collection when a mesh getter throws mid-conversion`: the mesh **is** handed out, and its `color` getter — read at `:83`, after the mesh is in hand and before it is pushed — throws. Asserts both `mesh.free()` and `collection.free()` were called once each, plus the propagated error.

Mutation evidence, error path isolated from the happy path:

| mutation | result |
|---|---|
| replace the whole `finally { mesh.free() }` with `void mesh` | 2 failed / 36 passed — too broad, also kills the happy-path test, proves nothing about the error path |
| move `mesh.free()` to the success path only (`batch.push(...); mesh.free(); } finally { void mesh; }`) | **1 failed / 37 passed — only the new test fails** |

The second is the precise defect the new test exists to catch: cleanup that runs on success and leaks the WASM mesh on every partial failure. Both older cleanup tests stayed green under it, confirming the coverage is genuinely new rather than duplicated.

### :124 — "Remove the `any` type escape in `asCollection`" — **APPLIED** ✅

Fair, and it costs nothing. `const asCollection = (c: unknown) => c as any` erased the return type at all 16 call sites. Now `c as Parameters<typeof convertMeshCollectionToBatch>[0]`, which is legal from `unknown`, keeps the call sites checked against the real `MeshCollection` parameter, and lets the `eslint-disable-next-line @typescript-eslint/no-explicit-any` go.

---

**Typecheck note:** `pnpm typecheck` does not cover this package (only 2 of 40 packages define the script), and the root `tsconfig.json` excludes `**/*.test.ts`. Typechecked separately against `packages/geometry/tsconfig.json` with that exclusion lifted: **zero errors in `geometry-coordinate.test.ts`**. (The run does surface 17 pre-existing errors in six *other* geometry test files — missing `@types/node`, a stale `StreamingGeometryEvent` import from `./types.js` — all untouched by this PR and invisible to CI for the same reason. Flagging, not fixing here.)

Thanks @coderabbitai — three for three, and `:148` caught exactly the failure mode this PR is about.

Issue #1402, opened 2026-06-26
clash detection results display enhancements


louistrue:
> 3 things I spot:
> 
> 1. bounding box should not be visible.
> 2. still struggle to see the penetration
> 3. when applying cross section and dragging it, 1 object is not cut anymore
> 
> <img width="790" height="779" alt="Image" src="https://github.com/user-attachments/assets/76047207-0097-4720-84ee-b5daeea49f58" /> 

 _Originally posted by @BIMvoice in [#1277](https://github.com/LTplus-AG/ifc-lite/issues/1277#issuecomment-4807665262)_

BIMvoice:
Much better now Louis, thank you!

<img width="607" height="709" alt="Image" src="https://github.com/user-attachments/assets/d60f7ec7-77ca-4d8e-9d80-a59e1f19b96d" />

PR #2162, opened 2026-08-05
test(viewer): close 25 mutation-verified gaps in the store's core state layer

BIMvoice:
Mutation sweep of the viewer's **core state layer** — `apps/viewer/src/store/**`. Selection and visibility set arithmetic, isolate/show-all interactions, federated per-model id spaces, section plane state, and the model add/update/remove lifecycle.

**Test-only. No production file is modified** (`git diff` touches six `*.test.ts` files and nothing else).

## Measurement

| | |
|---|---|
| Mutations against load-bearing logic | 27 (+2 deliberate controls) |
| KILLED | 4 |
| SURVIVED | 23 |
| Kill ratio | **4/27 = 15%** — flagged **TARGETED**: I aimed at actions with no test at all |
| `apps/viewer` tests | 2750 → **2815** (0 fail; the same 2 pre-existing env-gated LLM/OpenRouter skips) |

Controls, run before believing any survivor, both died as expected:

| Control | Killed by |
|---|---|
| `isolateEntity`: `new Set([id])` → `new Set()` | `should isolate single entity` |
| `resolveGlobalIdFromModels`: `localId <= maxExpressId` → `<` | `resolves the highest parsed express id through the fast (first) pass` |

`sectionSlice.ts` was measured and is **genuinely well covered** — 3/3 mutations killed (position upper clamp, custom-plane clearing on axis change, the pick-preview arm guard). Nothing added there.

## Gaps closed

Each row: the surviving mutation that proved the gap, and what a user would see if the mutated code shipped.

### `store/globalId.ts` — no test file existed at all

| Surviving mutation | User-visible effect | Killing test |
|---|---|---|
| `return expressId + (model.idOffset ?? 0)` → `return expressId` | Clicking a wall in the second federated model selects an unrelated element from the first | `adds the model offset for a federated model`, `round-trips every model through toGlobalId → fromGlobalId` |
| `localExpressId >= 0 && <= maxExpressId` → `> 0 && < maxExpressId` | The element at express id 0 and the element at exactly `maxExpressId` resolve to the wrong model (or to nothing) | `includes the LOWER boundary…`, `includes the UPPER boundary…` |

This survivor was re-run against the **whole** `apps/viewer` suite (2750 tests), not just the store — still green. New file `globalId.test.ts` (13 tests) also pins the `legacy` / `default` / `__legacy__` sentinels against a map that *does* carry an offset, so neither branch can be deleted alone.

### `store/basketVisibleSet.ts` — stale cache and set arithmetic

| Surviving mutation | User-visible effect | Killing test |
|---|---|---|
| Fingerprint drops `hiddenEntities` + `isolatedEntities` + `classFilter` | Hide a wall, press "Add visible to basket", and the wall you just hid lands in the basket | `re-computes after hiddenEntities changes (no explicit invalidate)` (+2 siblings) |
| Fingerprint drops `hiddenEntitiesByModel` | Same, for per-model hiding in a federated scene | `re-computes after per-model hidden entities change…` |
| `sorted.every(...)` → `sorted.some(...)` | Stacking a class filter on top of an isolation shows **more** than either filter alone | `keeps only the ids present in both the class filter and the isolation` |
| Per-model `hiddenEntitiesByModel` / `isolatedEntitiesByModel` gates deleted | Hiding an element inside one federated model does nothing to the basket's visible set | `applies hiddenEntitiesByModel in the model LOCAL id space, not the global one` (+1) |
| `expressId: globalId - offset` → `globalId` | Per-model sets never match in an offset model — same silent no-op | same two tests |
| `byStorey.has(id) ? id : id - offset` → either side alone | Picking a storey in the second federated file isolates to nothing | `resolves a GLOBAL selected-storey id back into the model local id space` |
| `basketIds.size !== isolated.size` → `>` | The basket's "isolation active" toggle lights up while the viewport shows more than the basket | `returns false when the isolation is a strict SUPERSET of the basket` |

The two pre-existing cache tests are the *\"property that holds whether or not the code does the thing\"* shape — `assert.deepStrictEqual(first, second)` is exactly as true for a cache that never invalidates as for a correct one. The new tests mutate store state **without** calling `invalidateVisibleBasketCache()`, so only the fingerprint can save them.

Similarly, the pre-existing `returns false when isolated size differs from basket` case had the basket *larger* than the isolation, where the membership loop also returns false — so the size check itself was never discriminated. Both directions are now pinned separately.

### `store/slices/visibilitySlice.ts`

| Surviving mutation | User-visible effect | Killing test |
|---|---|---|
| `showAll` stops clearing `classFilter` + `ghostExceptEntities` | Press "Show all" and geometry stays hidden / translucent, with no UI affordance left to explain why | `showAll clears hidden, isolation, the class filter AND X-Ray ghosting` |
| `clearAllFilters` stops clearing them | same | `clearAllFilters clears isolation, the class filter and ghosting but keeps hidden entities` |
| `setIsolatedEntities` stops clearing `hiddenEntities` / ghosting | A BCF viewpoint with `defaultVisibility=false` shows fewer elements than the viewpoint asked for | `setIsolatedEntities clears hidden entities and ghosting (mutually exclusive modes)` |
| `isEntityVisible`'s whole `classFilter` branch deleted | "Show only IfcWall" stops hiding anything | `returns false for an entity outside an active class filter`, `requires an entity to pass EVERY gate, not just one` |
| `isolateEntities` toggle-off `every` → `some` | Isolate `{100,200}`, then isolate `{100,999}` → the whole model reappears instead of switching | `does NOT toggle off on a same-size set that merely overlaps` |
| `setClassFilter` toggle-off `every` → `some` | Clicking a second class row that shares one id turns the filter off instead of switching | `does NOT toggle off on a same-size class that merely overlaps` |
| `showEntitiesInModel` stops pruning the emptied model key | Phantom \"this model has hidden geometry\" entry that every `.has(modelId)` consumer and the basket fingerprint read | `drops the model key entirely once its last hidden entity is shown` (+ the opposite direction) |

`isolateEntities`, `setClassFilter`, `clearClassFilter`, `clearAllFilters`, `setHiddenEntities`, `setIsolatedEntities`, `setGhostExceptEntities` and `clearGhost` had no tests at all before this.

### `store/slices/selectionSlice.ts`

`selectedEntityId` is the **global** id the renderer highlights, so every legacy multi-select action must keep it on the most recent survivor of `selectedEntityIds`. None of that was pinned.

| Surviving mutation | User-visible effect | Killing test |
|---|---|---|
| `removeFromSelection`: `remaining[remaining.length-1]` → `remaining[0]` | Deselecting one element jumps the highlight box to the *oldest* selection | `removeFromSelection promotes the LAST remaining id, not the first` |
| `setSelectedEntityIds`: `ids[ids.length-1]` → `ids[0]` | same, for bulk "select all results" | `setSelectedEntityIds makes the LAST id primary, and [] clears it` |
| `addToSelection` drops `selectedEntityId: id` | Ctrl-click adds to the set but the highlight never moves | `addToSelection makes the added id primary` |
| `setSelectedEntityId` stops clearing `selectedModelId` | The properties panel shows model metadata *and* an element at once | `setSelectedEntityId clears selectedModelId when an entity is picked` |
| …and the opposite: always clearing it | Clicking empty canvas drops a hierarchy model-row selection | `setSelectedEntityId(null) does NOT clear selectedModelId` |
| `setSelectedModelId` stops clearing the entity channels | same collision, from the other side | `setSelectedModelId clears every entity-selection channel` |
| `setSelectedEntities`: primary `refs[0]` → last, and keeps `selectedModelId` | The unified-storey property panel binds to the wrong storey | `setSelectedEntities makes the FIRST ref primary and clears the model selection` |

### `store/slices/modelSlice.ts`

| Surviving mutation | User-visible effect | Killing test |
|---|---|---|
| `upsertModel`: drop the `{ ...existing, ...model }` merge | The ingest pipeline upserts a stub then the payload; a replacing upsert reverts the model's name / georef / visibility mid-load | `MERGES into an existing entry instead of replacing it` |
| `upsertModel`: `state.activeModelId ?? model.id` → `model.id` | Loading a second model steals the active slot | `does not steal the active slot from an already-active model` |
| `updateModel`: stop re-mirroring `ifcDataStore` / `geometryResult` | The loader attaches parsed data via `updateModel`; without the mirror the whole app keeps reading the pre-parse store while the model list shows the file as loaded | `re-mirrors ifcDataStore / geometryResult when the ACTIVE model is patched` |
| …and the opposite: always mirror | Patching a background model swaps the active model's data out from under the UI | `leaves the active mirror alone when a NON-active model is patched` |
| `updateModel`: remove the `if (!model) return {}` early return | Unknown model id inserts a `{ undefined }` entry into the federation map | `is a no-op for an unknown model id` |

### `store/typeVisibilityFilter.ts`

The existing loop only ever discriminated `spaces`, `site` and `ifcAnnotations`. Swapping `IfcOpeningElement` ↔ `IfcVirtualElement`, or remapping `IfcSpatialZone` onto `spaces`, survived — the Openings switch would hide virtual elements and leave openings visible. Now measured **per pair, in both directions** (own toggle off ⇒ hidden; every other toggle off ⇒ still visible), for `isTypeVisible` and for `buildHiddenIfcTypes` (the GLB-export gate) alike, so the two consumers can never disagree about which class a toggle owns.

## Finding for a maintainer — not fixed here

`showAllInAllModels` does **not** clear `ghostExceptEntities`, while `showAll`, `clearAllFilters` and `setHiddenEntities` all do. `showAllInAllModels` is what `resetVisibilityForHomeFromStore` (`store/homeView.ts`) calls for Home / "reset filters" — the same function that was extended in #1402 to drop focused-clash state. So: run a clash check in **Ghost** mode (`useClash.ts` → `setGhostExceptEntities`), press Home, and the model stays translucent with no clash selection left to explain it.

That looks like an oversight rather than a decision, but it is a behaviour change, so this PR does not make it. The `showAllInAllModels` test here pins only what the action currently promises and carries an inline note rather than enshrining the gap in either direction.

Nothing in this sweep touched the `scene.clear()` / instanced-template area of **#2073**, and nothing here depends on how that is resolved.

## Verification

- `pnpm --filter @ifc-lite/viewer test` — 2815 tests, 0 fail, 2 skipped (both pre-existing, env-gated: `IFC_LITE_VERIFY_OPENROUTER_MODELS` and the LLM env registry check).
- `pnpm --filter @ifc-lite/viewer typecheck` passes. Confirmed empirically that it **does** cover test files: it initially failed on three `setGhostExceptEntities([7, 8])` calls in the new tests (the action takes a `Set`), which is exactly the check working.
- Every new test was RED-checked against the mutation it targets and GREEN on clean `main`. Both directions of each binary signal are pinned; boundaries probed at `localId === 0`, `localId === maxExpressId`, `maxExpressId + 1`, empty selections, single-element sets, and the last model being removed.
- `node scripts/check-test-wiring.mjs` clean.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for basket visibility, isolation, filtering, lens, geometry, and model-specific visibility.
  * Added federation global-ID conversion tests, including offsets, gaps, legacy values, and round trips.
  * Added tests for model updates, active-model state, selection behavior, and visibility resets.
  * Added comprehensive IFC type visibility tests covering class-specific toggles and combined filters.

### DIFF
--- a/apps/viewer/src/store/basketVisibleSet.test.ts
+++ b/apps/viewer/src/store/basketVisibleSet.test.ts
@@ -5,7 +5,10 @@
 import assert from 'node:assert/strict';
 import { afterEach, beforeEach, describe, it } from 'node:test';
 import { IfcTypeEnum, RelationshipType, type SpatialHierarchy, type SpatialNode } from '@ifc-lite/data';
+import type { GeometryResult, MeshData } from '@ifc-lite/geometry';
 import type { AggregationRelationships } from '../utils/aggregation.js';
+import type { FederatedModel } from './types.js';
+import type { ViewerState } from './index.js';
 import {
   collectSpatialSubtreeElementsWithIfcSpace,
   getSmartBasketInputFromStore,
@@ -16,6 +19,59 @@ import {
 } from './basketVisibleSet.js';
 import { useViewerStore } from './index.js';
 import { entityRefToString } from './types.js';
+
+/**
+ * The visible-set code reads exactly two things off geometry: `meshes.length`
+ * (the cache fingerprint, basketVisibleSet.ts:72 and :92) and each mesh's
+ * `expressId` / `ifcType` (basketVisibleSet.ts:372 and :384). Everything else
+ * on `MeshData` / `GeometryResult` is filled in here with empty-but-valid
+ * values so the fixtures satisfy the REAL contract. An `as any` would have
+ * been shorter, but it also silently survives the day one of those required
+ * fields starts mattering to the code under test.
+ */
+function createMesh(expressId: number, ifcType: string): MeshData {
+  return {
+    expressId,
+    ifcType,
+    positions: new Float32Array(0),
+    normals: new Float32Array(0),
+    indices: new Uint32Array(0),
+    color: [1, 1, 1, 1],
+  };
+}
+
+function createGeometry(meshes: MeshData[]): GeometryResult {
+  const zero = { x: 0, y: 0, z: 0 };
+  return {
+    meshes,
+    totalTriangles: 0,
+    totalVertices: 0,
+    coordinateInfo: {
+      originShift: zero,
+      originalBounds: { min: zero, max: zero },
+      shiftedBounds: { min: zero, max: zero },
+      hasLargeCoordinates: false,
+    },
+  };
+}
+
+/** A federated model with the federation fields the basket cares about. */
+function createFederatedModel(overrides: Partial<FederatedModel> = {}): FederatedModel {
+  return {
+    id: 'm1',
+    name: 'Model 1',
+    ifcDataStore: null,
+    geometryResult: null,
+    visible: true,
+    collapsed: false,
+    schemaVersion: 'IFC4',
+    loadedAt: 0,
+    fileSize: 0,
+    idOffset: 0,
+    maxExpressId: 0,
+    ...overrides,
+  };
+}
 
 function createNode(expressId: number, type: IfcTypeEnum, children: SpatialNode[] = [], elements: number[] = []): SpatialNode {
   return {
@@ -290,9 +346,9 @@ describe('basketVisibleSet', () => {
     // hid lands in the basket anyway.
     describe('the fingerprint reacts to every visibility channel it reads', () => {
       const meshes = [
-        { expressId: 1, ifcType: 'IfcWall' },
-        { expressId: 2, ifcType: 'IfcWall' },
-        { expressId: 3, ifcType: 'IfcWall' },
+        createMesh(1, 'IfcWall'),
+        createMesh(2, 'IfcWall'),
+        createMesh(3, 'IfcWall'),
       ];
 
       function seedThreeVisibleWalls() {
@@ -305,7 +361,7 @@ describe('basketVisibleSet', () => {
           hiddenEntities: new Set(),
           isolatedEntities: null,
           classFilter: null,
-          geometryResult: { meshes } as any,
+          geometryResult: createGeometry(meshes),
         });
         invalidateVisibleBasketCache();
         const seeded = getVisibleBasketEntityRefsFromStore().map(entityRefToString).sort();
@@ -368,7 +424,7 @@ describe('basketVisibleSet', () => {
         // PREVIOUS model's refs out of the cache.
         seedThreeVisibleWalls();
         useViewerStore.setState({
-          geometryResult: { meshes: [{ expressId: 1, ifcType: 'IfcWall' }] } as any,
+          geometryResult: createGeometry([createMesh(1, 'IfcWall')]),
         });
 
         assert.deepStrictEqual(
@@ -388,7 +444,7 @@ describe('basketVisibleSet', () => {
           isolatedEntities: null,
           classFilter: null,
           hiddenEntitiesByModel: new Map(),
-          models: new Map([['m1', { id: 'm1', visible: true, idOffset: 0, maxExpressId: 10, geometryResult: { meshes } } as any]]),
+          models: new Map([['m1', createFederatedModel({ maxExpressId: 10, geometryResult: createGeometry(meshes) })]]),
         });
         invalidateVisibleBasketCache();
         assert.strictEqual(getVisibleBasketEntityRefsFromStore().length, 3);
@@ -417,13 +473,11 @@ describe('basketVisibleSet', () => {
           hierarchyBasketSelection: new Set(),
           models: new Map(),
           hiddenEntities: new Set(),
-          geometryResult: {
-            meshes: [
-              { expressId: 1, ifcType: 'IfcWall' },
-              { expressId: 2, ifcType: 'IfcWall' },
-              { expressId: 3, ifcType: 'IfcWall' },
-            ],
-          } as any,
+          geometryResult: createGeometry([
+            createMesh(1, 'IfcWall'),
+            createMesh(2, 'IfcWall'),
+            createMesh(3, 'IfcWall'),
+          ]),
           classFilter: { ids: new Set([1, 2]), label: 'IfcWall' },
           isolatedEntities: new Set([2, 3]),
         });
@@ -445,7 +499,7 @@ describe('basketVisibleSet', () => {
       });
 
       /** One model at offset 1000 carrying global meshes 1001..1003. */
-      function seedOffsetModel(extra: Record<string, unknown> = {}) {
+      function seedOffsetModel(extra: Partial<ViewerState> = {}) {
         useViewerStore.setState({
           selectedEntitiesSet: new Set(),
           selectedEntity: null,
@@ -457,21 +511,17 @@ describe('basketVisibleSet', () => {
           classFilter: null,
           hiddenEntitiesByModel: new Map(),
           isolatedEntitiesByModel: new Map(),
-          models: new Map([['m1', {
-            id: 'm1',
-            visible: true,
+          models: new Map([['m1', createFederatedModel({
             idOffset: 1000,
             maxExpressId: 100,
-            geometryResult: {
-              meshes: [
-                { expressId: 1001, ifcType: 'IfcWall' },
-                { expressId: 1002, ifcType: 'IfcWall' },
-                { expressId: 1003, ifcType: 'IfcWall' },
-              ],
-            },
-          } as any]]),
+            geometryResult: createGeometry([
+              createMesh(1001, 'IfcWall'),
+              createMesh(1002, 'IfcWall'),
+              createMesh(1003, 'IfcWall'),
+            ]),
+          })]]),
           ...extra,
-        } as any);
+        });
         invalidateVisibleBasketCache();
       }
 
@@ -529,7 +579,13 @@ describe('basketVisibleSet', () => {
 
         seedOffsetModel();
         const models = new Map(useViewerStore.getState().models);
-        models.set('m1', { ...models.get('m1')!, ifcDataStore: { spatialHierarchy: hierarchy } as any });
+        // `IfcDataStore` is a class with a large surface; the basket only walks
+        // `spatialHierarchy`, so narrow through `unknown` rather than `any` —
+        // the store field keeps its declared type at every read below.
+        models.set('m1', {
+          ...models.get('m1')!,
+          ifcDataStore: { spatialHierarchy: hierarchy } as unknown as FederatedModel['ifcDataStore'],
+        });
         useViewerStore.setState({
           models,
           // 1004 = local storey 4 + the model's 1000 offset.
--- a/apps/viewer/src/store/basketVisibleSet.test.ts
+++ b/apps/viewer/src/store/basketVisibleSet.test.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 import assert from 'node:assert/strict';
-import { beforeEach, describe, it } from 'node:test';
+import { afterEach, beforeEach, describe, it } from 'node:test';
 import { IfcTypeEnum, RelationshipType, type SpatialHierarchy, type SpatialNode } from '@ifc-lite/data';
 import type { AggregationRelationships } from '../utils/aggregation.js';
 import {
@@ -228,6 +228,33 @@ describe('basketVisibleSet', () => {
 
       assert.strictEqual(isBasketIsolationActiveFromStore(), false);
     });
+
+    it('returns false when the isolation is a strict SUPERSET of the basket', () => {
+      // The existing "size differs" case only covers isolation SMALLER than
+      // the basket, where the membership loop also fails — so the `!==`
+      // size check itself was never discriminated and could be weakened to
+      // `>` unnoticed. Here every basket id IS isolated but extra elements
+      // are isolated too, so only the size check can say "no". Getting this
+      // wrong lights up the basket's "isolation active" toggle when the
+      // viewport is showing more than the basket.
+      useViewerStore.setState({
+        pinboardEntities: new Set(['legacy:100']),
+        isolatedEntities: new Set([100, 200]),
+        models: new Map(),
+      });
+
+      assert.strictEqual(isBasketIsolationActiveFromStore(), false);
+    });
+
+    it('returns false when sizes match but membership does not', () => {
+      useViewerStore.setState({
+        pinboardEntities: new Set(['legacy:100', 'legacy:200']),
+        isolatedEntities: new Set([100, 999]),
+        models: new Map(),
+      });
+
+      assert.strictEqual(isBasketIsolationActiveFromStore(), false);
+    });
   });
 
   describe('visibility cache', () => {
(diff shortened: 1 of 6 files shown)


BIMvoice:
Verified the headline gap on clean `upstream/main`, and it is the starkest result of this campaign.

`apps/viewer/src/store/globalId.ts` has **no test file** on `main`. Deleting the federation offset outright:

```ts
- return expressId + (model.idOffset ?? 0);
+ return expressId;                            // replacements=1
```

```
$ pnpm --filter @ifc-lite/viewer test
# tests 2750
# pass 2748
# fail 0
# skipped 2      (both pre-existing, env-gated)
```

The entire viewer suite — 2,750 tests — cannot tell whether federated id mapping exists. On screen that is: click a wall in the second model of a federated set, and an unrelated element in the first model gets selected. Silent, and only in multi-model sessions.

**The file documents its own blind spot at line 53:**

```ts
// For a true single model with idOffset 0 this still yields expressId === globalId.
```

That comment is exactly right, and it is why every fixture that exercised this path proved nothing: with one model the offset is zero and the mapping is the identity. This is now the **fourth** instance today of the same class — a fixture constant that makes the operation under test an identity, so the line is executed and nothing about it is asserted. The others: seven geometry fixtures omitting `origin`, pointcloud classifications below the `& 0x1f` mask, and every solar fixture at longitude 0. Coverage tooling scores all four as covered.

Second finding worth reading, because the mechanism differs:

**The basket visible-set cache fingerprint can drop `hiddenEntities` / `isolatedEntities` / `classFilter` and the two pre-existing cache tests still pass.** They assert `deepStrictEqual(first, second)` across two calls — which is exactly as true of a cache that never invalidates as of one that invalidates correctly. That is the *"property that holds whether or not the code does the thing"* shape, applied to caching. The new tests mutate state **without** calling `invalidateVisibleBasketCache()`. User-visible: hide a wall, press "Add visible to basket", the hidden wall goes in.

Also corrected: `isBasketIsolationActiveFromStore`'s "size differs" test never discriminated the size check (the membership loop returns false too), and `typeVisibilityFilter`'s class→toggle loop discriminated only 3 of its 7 cases — now measured per case rather than inferred from the loop failing.

**`sectionSlice.ts` came back genuinely well covered** — 3 of 3 mutations killed, nothing added. Worth stating alongside the failures.

## One production finding, reported not fixed

`showAllInAllModels` does **not** clear `ghostExceptEntities`, while `showAll`, `clearAllFilters` and `setHiddenEntities` all do. It is what `resetVisibilityForHomeFromStore` (`apps/viewer/src/store/homeView.ts:9`) calls for Home / "reset filters" — the same function #1402 extended to drop focused-clash state.

Repro: run a clash check in **Ghost** mode (`useClash.ts:327` → `setGhostExceptEntities`), then press Home. The model stays translucent with no clash selection left to explain why.

It reads like an oversight rather than a decision, but it is a behaviour change either way, so nothing was altered. The test pins only what the action currently promises and carries an inline note, so it does not enshrine the gap in either direction. Your call.

The 15% kill ratio is flagged **targeted** — the mutations aimed at actions with no test at all, so it is not an estimate of the package.

BIMvoice:
## Round-four self-audit: 41 mutations, 35 KILLED, 6 SURVIVED — all six ours, fixed in ee1abd4

Method as before: exact-substring replacement with the count asserted `== 1`, mutated region re-read from disk, restore from a pre-mutation copy. Baseline first on a built workspace — the six touched store files are 182 passing tests, and the full `apps/viewer` suite is 2816 passed / 2 skipped.

### What held (35 of 41)

`globalId.ts` (the module that had no tests at all) is the strongest part of this PR — every mutation dies, and each dies narrowly:

| Mutation | Count | Result |
|---|---|---|
| `return expressId + (model.idOffset ?? 0)` → `return expressId` | 1 | KILLED (5 tests, incl. basketVisibleSet) |
| lower boundary `localExpressId >= 0` → `> 0` | 1 | KILLED |
| upper boundary `<= model.maxExpressId` → `<` | 1 | KILLED |
| drop the `'default'` legacy sentinel | 1 | KILLED |
| disable the single-model rescue arm | 1 | KILLED |
| no-models fallback `globalId` → `globalId + 1` | 1 | KILLED |

`typeVisibilityFilter.ts` 5/5 (`!ifcType` → false, unmapped-key → false, `IfcGeographicElement: 'site'` → `'spaces'`, `IfcSpatialZone` → `'spaces'`, `buildHiddenIfcTypes` polarity flip). `modelSlice` 3/4, `selectionSlice` 3/3 (last-vs-first primary, the `selectedEntityId !== null` guard, first-vs-last ref). `visibilitySlice` 6/6 — notably both `ids.every(...)` membership checks in `isolateEntities` and `setClassFilter` die when reduced to a size comparison, which is exactly what the "same-size set that merely overlaps" tests were written for. `basketVisibleSet` 10/11 on the offset/gate/intersection paths (`expressId: globalId - offset` → `globalId`, intersect → union, the per-model local-vs-global id space in both directions, `storeyId - offset`, the `model.visible` gate, the basket-isolation size check).

### The six survivors — our own defects

Five are the visible-set cache. The block is titled *"the fingerprint reacts to every visibility channel it reads"*, but four of the channels it reads were unpinned, because every other test in the file calls `invalidateVisibleBasketCache()` by hand and so never exercises the key:

| Mutation in `basketVisibleSet.ts` | Count | Before | After ee1abd4 |
|---|---|---|---|
| `digestNumberSet(state.lensHiddenIds)` → `''` | 1 | **SURVIVED** | KILLED |
| `globalHidden.add(id)` → `globalHidden.delete(id)` for lens ids | 1 | **SURVIVED** | KILLED |
| `digestModelEntityMap(state.isolatedEntitiesByModel)` → `''` | 1 | **SURVIVED** | KILLED |
| `digestNumberSet(state.selectedStoreys)` → `''` | 1 | **SURVIVED** | KILLED |
| `state.geometryResult?.meshes?.length ?? 0` → `0` | 1 | **SURVIVED** | KILLED |

The second row is the one that matters most: it is not a cache issue at all. `lensHiddenIds` were **never applied** by any test — inverting the fold that hides them left the whole store suite green. A lens that hides elements would leave them in the basket's visible set and nothing would have caught it.

The sixth is in `modelSlice`:

| Mutation | Count | Before | After |
|---|---|---|---|
| `updateModel`: `if (!model) return {}` → `return { activeModelId: modelId }` | 1 | **SURVIVED** | KILLED |

The test is named *"is a no-op for an unknown model id"* but asserted only the models map, so an early return that still emits a patch — pointing `activeModelId` at an id that does not exist, and dragging the mirrored `ifcDataStore` / `geometryResult` with it — passed.

### Fix (test-only, no production change)

Four new cache tests (lens ids applied **and** invalidating, legacy mesh-set change, per-model isolation change, storey-selection clear) plus three assertions on the `updateModel` no-op. All six mutations re-measured as KILLED. `apps/viewer`: 2816 passed / 2 skipped / 0 failed, `pnpm --filter @ifc-lite/viewer typecheck` clean.

BIMvoice:
Triaged all three CodeRabbit threads against the code. All three point at something real; one of them is right about the gap but wrong about the fix. Pushed as 4e30b632e — **test files only, no production code changed**.

Baseline before: 2818 tests / 2816 pass / 2 skipped, 0 fail. After: 2819 / 2817 / 2 skipped, 0 fail. Root `pnpm typecheck` green.

---

### A. `selectionSlice.test.ts:348` — "Assert that `selectedEntityIds` is cleared" — **APPLIED, with a correction to the suggested fix**

The gap is real and it is **ours**: `setSelectedModelId clears every entity-selection channel` called `setSelectedEntityIds([7, 8])` in its arrange block and then never asserted that channel. CodeRabbit caught a test that sets up state it does not check. Credit where due.

But the suggested assertion is wrong about the expected behaviour. `setSelectedModelId` (`selectionSlice.ts:292-298`) clears `selectedEntity`, `selectedEntities` and `selectedEntityId` — and leaves `selectedEntityIds` standing. Applying the proposed diff verbatim fails:

```
not ok 3 - setSelectedModelId clears every entity-selection channel
  expected: 0
  actual: 2
```

So the test name was the thing that was wrong — it claimed "every channel" while the implementation covers three of four. Two changes:

1. Renamed to `setSelectedModelId clears the multi-model entity channels and the primary id`.
2. Added a second test pinning the survivor, with the user-visible consequence written down: `MainToolbar.tsx:352` and `ElementsTab.tsx:44` keep reporting the stale count and `useAnimationLoop.ts:244` keeps painting the stale highlight after a model row is selected, while the properties panel has already switched to the model. The comment says plainly that this pins current behaviour rather than endorsing it, and that this test is the one to flip if the asymmetry is ever fixed. Flagging rather than changing it — `setSelectedModelId`'s own comment reads "Clear other selection when selecting a model", so the omission looks unintended, but that is a behaviour call for a separate PR, not a test-hardening one.

Mutation evidence (1 replacement each, verified by re-read):
- `setSelectedModelId` also clears `selectedEntityIds` → **KILLED** by the new test, sole failure.
- M11 re-verified after the rename: `setSelectedModelId` stops clearing anything → **KILLED**, still the sole failure in the file.

**On M11's provenance**, since the PR body claims it: I re-ran M11 against the *entire* 2818-test suite, not just the file. It fails exactly one test — the named one. No sibling was doing the work. The claim in the PR body holds. What did *not* hold is that the arrange line `setSelectedEntityIds([7, 8])` contributed nothing to that kill; it was decoration until now.

### B. `basketVisibleSet.test.ts:308` — the `as any` casts — **APPLIED**

Right on both counts: `as any` is banned by `AGENTS.md:17`, and here it was masking real non-conformance rather than being noise. `apps/viewer`'s tsconfig overrides the root `**/*.test.ts` exclusion, so these files *are* typechecked — I confirmed that with a deliberate `const __probe: number = 'not a number'`, which tsc duly rejected.

What the cast was hiding, with the cast removed and the fixture left as-is:

```
src/store/basketVisibleSet.test.ts:364:38 - error TS2739: Type '{ expressId: number; ifcType: string; }'
  is missing the following properties from type 'MeshData': positions, normals, indices, color
```

`GeometryResult` additionally requires `totalTriangles`, `totalVertices` and `coordinateInfo`. So the fixtures did not satisfy the store shape.

Replaced all six casts (308, 371, 391, 472, 474, 532) with typed helpers — `createMesh`, `createGeometry`, `createFederatedModel` — that build genuinely conformant values, so there is no residual cast on those paths. Two notes:

- `seedOffsetModel(extra)` was `Record<string, unknown>` passed into a `setState({...}) as any`. That combination meant a typo in an `extra` key would have silently seeded nothing. Now `Partial<ViewerState>`, and the whole-object cast is gone.
- Line 532's `ifcDataStore` is the one place a conformant literal is not reasonable — `IfcDataStore` is a class with a large surface and the basket only walks `spatialHierarchy`. Narrowed through `unknown` to the declared field type instead of `any`, matching the existing house pattern, so reads stay typed.

Scope: I fixed the six casts this PR introduced. The file still carries pre-existing `as any` fixtures at lines 171, 225, 319, 331, 666, 681 and 694 from before this branch; the new helpers make those one-liners to convert, but they are outside this PR's diff and I left them alone.

### C. `modelSlice.test.ts:202-214` — assert the geometry mirror — **APPLIED, and it is more than a nitpick**

CodeRabbit filed this as Trivial. It is not. `leaves the active mirror alone when a NON-active model is patched` patched only `ifcDataStore`, so the `geometryResult` half of the same gate was completely unpinned. Deleting the gate outright:

```
-      geometryResult: state.activeModelId === modelId ? updatedModel.geometryResult : state.geometryResult,
+      geometryResult: updatedModel.geometryResult,
```

**survived all 2819 tests, 0 failures** — not just this file, the whole `apps/viewer` suite. Patching a background model would have swapped the rendered geometry out from under the active view with no test objecting.

Applied essentially the proposed extension: distinct geometry for both models, patch both payloads, assert both mirrors, and compare the model-2 entries by reference instead of reaching through `as unknown as { tag: string }`.

Mutation evidence (1 replacement each, verified by re-read):
- geometry gate deleted → survived before / **KILLED** after.
- `ifcDataStore` gate deleted → **KILLED**, confirming the strengthened test still kills what it already killed.

---

One surviving mutant found in passing, left alone as outside these threads: blanking the per-model mesh channel of the cache fingerprint (`basketVisibleSet.ts:72`, `${modelId}:${visible}:${meshes.length}` → `:0`) survives the file's 32 tests. I verified against the pre-change file that this is pre-existing and not something my refactor introduced. Happy to pin it in a follow-up if wanted.

BIMvoice:
Verified finding C, and I want to record a mistake I made verifying it, because it is the same class of error the campaign is about.

**My first mutation was wrong.** I deleted the whole `geometryResult` line from `updateModel`, saw it fail one test, and nearly reported the finding as overstated. But deleting the line breaks the **active** branch too, which the neighbouring test already covers — so it was non-probative for a finding about the **non-active** branch. A mutation that kills more than the property under test proves nothing about that property.

Re-run with the narrow mutation the finding is actually about — the non-active branch losing its guard:

```ts
- geometryResult: state.activeModelId === modelId ? updatedModel.geometryResult : state.geometryResult,
+ geometryResult: updatedModel.geometryResult,        // replacements=1
```

```
against the PRE-FIX tests   ->  2819 tests, 0 fail          <- SURVIVED
against the fixed tests     ->  not ok - leaves the active mirror alone
                                when a NON-active model is patched
```

Restored: 2819 tests, 0 fail.

So the finding holds exactly as reported. The pre-fix test patched only `ifcDataStore` on the non-active model, so the `geometryResult` half of that branch was unpinned: **patching a background model could swap rendered geometry out from under the active view, with the whole 2,819-test suite green.** CodeRabbit filed it as "Trivial"; it is not.

Three things worth separating out from the rest:

**On finding A, CodeRabbit was right about the gap and wrong about the fix.** Its suggested assertion — that `setSelectedModelId` clears `selectedEntityIds` — fails against the implementation: `expected: 0, actual: 2`. `selectionSlice.ts:292-298` clears `selectedEntity`, `selectedEntities` and `selectedEntityId`, and deliberately leaves `selectedEntityIds` standing. So the defect was the **test's name** ("clears every entity-selection channel" while covering three of four), not the assertion's absence. Applying the suggestion verbatim would have produced a failing test asserting behaviour the code does not have.

That asymmetry is now pinned rather than fixed, and it is worth your call: a stale `selectedEntityIds` survives selecting a model row, which reaches `MainToolbar.tsx:352` and `ElementsTab.tsx:44` as a stale count and `useAnimationLoop.ts:244` as a stale highlight. The implementation's own "Clear other selection" comment suggests the omission is unintended — but that is a behaviour change and not ours to make.

**On finding B, the `as any` was masking genuine non-conformance**, not merely noise: removing it yields `TS2739: ... missing the following properties from type 'MeshData': positions, normals, indices, color`. Since `apps/viewer` test files genuinely are typechecked — verified with a deliberate `const __probe: number = 'not a number'`, which `tsc` rejected — that cast was suppressing a real error. Six casts replaced with typed helpers. It also caught that `seedOffsetModel(extra: Record<string, unknown>)` inside a `setState({...}) as any` would silently swallow a typo'd key; now `Partial<ViewerState>`.

**And it settled a question I had raised.** I suspected the PR's "M11 KILLED" claim might have come from a sibling test rather than the named one. It checked: M11 run against the full 2,818-test suite produces exactly one failure, the named test. The claim holds. What did not hold is a smaller thing it reported anyway — the `setSelectedEntityIds([7, 8])` arrange line contributed nothing to that kill.

BIMvoice:
Re the CodeRabbit thread on `apps/viewer/src/store/slices/selectionSlice.test.ts:348` — **declining this one; the suggestion would make the suite fail.**

It proposes adding `assert.strictEqual(state.selectedEntityIds.size, 0)` to the `setSelectedModelId` test. But `setSelectedModelId` does **not** clear `selectedEntityIds` — `selectionSlice.ts:292` clears `selectedEntity`, `selectedEntities` and `selectedEntityId` and leaves the legacy global-id set standing, despite its own comment reading "Clear other selection when selecting a model".

That is exactly what the very next test in this file already pins, deliberately:

```
it('setSelectedModelId does NOT clear the legacy selectedEntityIds set', ...)
```

with a comment recording that it pins the current behaviour rather than endorsing it, and naming the user-visible consequence: `MainToolbar.tsx:352` and `ElementsTab.tsx:44` keep reporting the stale count.

So the asymmetry is not an oversight in the test — it is the documented finding. Asserting `.size === 0` would assert behaviour the code does not have.

Happy to go further if you'd rather: the honest fix is in the slice, not the test — have `setSelectedModelId` clear `selectedEntityIds` too, so the comment and the code agree. That is a behaviour change with a user-visible effect on the two call sites above, so it's your call, not something to fold into a test-only PR.

louistrue:
Resolved the remaining thread. Checked the decline rather than taking it: `selectionSlice.ts:292-298` sets `selectedEntity`, `selectedEntities` and `selectedEntityId` and does not touch `selectedEntityIds`, and the very next test in the file pins that asymmetry deliberately. The suggested assertion would fail the suite as written, so declining it is right.

PR #2163, opened 2026-08-05
test(parser): close 21 mutation-verified gaps in the tokenizer, header reader, and index builder

BIMvoice:
Second, deeper mutation sweep of `packages/parser`. Test-only — **no production code changed**.

## Method

28 mutations against load-bearing logic. Each was applied by exact-substring replacement with an asserted replacement count of 1, the mutated region was re-read and confirmed changed, and every survivor was re-run against **all three suites that alias `@ifc-lite/parser` to source** — `packages/parser`, `packages/export`, `packages/query` — before being called a survivor.

**Controls** (mutations expected to die, run before believing any survivor):

| Control | Mutation | Result |
|---|---|---|
| A | `tokenizer.ts` — express-id accumulation `id*10 + (c-0x30)` → `id*10` | KILLED, 46 tests |
| B | `unit-extractor.ts` — `'MILLI': 1e-3` → `1e-2` | KILLED, 7 tests |

**Kill ratio: 6/28 (21%) before, 27/28 (96%) after.** Flagged **TARGETED** — the mutations deliberately aimed at the four modules that had no test file of their own.

## Modules with no direct coverage

| Module | LOC | Mutations | Survived before |
|---|---|---|---|
| `src/tokenizer.ts` | 359 | 10 | 9 |
| `src/source-header.ts` | 269 | 11 | 8 |
| `src/entity-refs-from-index.ts` | 126 | 3 | 2 |
| `src/attribute-helpers.ts` | 56 | 3 | 2 |
| `src/compact-entity-index.ts` | 480 | 1 | 0 |

## Gaps closed

Each row: the surviving mutation that proved the gap, and the test that now kills it.

### `src/tokenizer.ts` — the STEP scanner every model in the product passes through

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| Type-cache hit no longer verified byte-for-byte | A hostile or corrupt file gets one entity type silently read as another — a door reported as a wall. The verification exists specifically to defeat this; nothing tested it. | `does not alias two type names that collide in the type-name cache` (`Aa`/`BB` — same length, same 32-bit rolling hash) |
| Entity byte length excludes the terminating `;` | Every record slice is one byte short. | `reports a byte length that spans the record up to and including the terminating semicolon` |
| Type token no longer required to start `A-Z` | Malformed input mints entities out of lowercase tokens. | `rejects a type token that does not start with an uppercase letter` |
| Newlines inside a record body not counted | Every entity after a multi-line record reports a wrong line number in diagnostics. | `counts newlines inside a multi-line record so later entities keep true line numbers` |
| `findEntityLength` ignores string state | A `(` or `)` inside a quoted attribute changes the nesting depth; the record is truncated mid-string. | `ignores parentheses that appear inside a quoted string`, `ignores an unbalanced open paren inside a quoted string` |
| **`scanEntities()` yields nothing at all** | The entire paren-matching generator was dead to tests. It is exported public API (`StepTokenizer` via `index.ts`), so SDK consumers can call it. | four tests under `StepTokenizer.scanEntities` |
| `readExpressId` accepts a digitless `#` | A bare `#` mints an entity with `expressId` 0 that collides with every other malformed marker. | ``yields nothing for a `#` that carries no express id`` |

### `src/source-header.ts` — read on every parse, written back on every STEP export

A decoding error here silently rewrites a file's authorship and provenance on export.

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| `*` (derived) sentinel no longer maps to `undefined` | A literal `*` is written into the exported header field. | ``maps the `*` derived sentinel to undefined, not to a literal asterisk`` |
| `''` doubled-quote escape not collapsed | `it''s a name` stays doubled on read and re-doubles on every export round-trip. | ``collapses the `''` doubled-quote escape to a single apostrophe`` |
| `\X4\` span not consumed atomically | A UTF-32 directive immediately followed by an escaped backslash loses its terminator and never decodes. | `keeps a \X4\ directive intact when an escaped backslash immediately follows it` |
| `\S\` span not consumed atomically | A `\S\` whose operand is itself a backslash is claimed by the `\\` pair escape, leaving the directive unterminated. | `treats a backslash as a valid \S\ operand rather than a pair escape` |
| ENDSEC truncation removed | A quoted DATA-section attribute containing the literal text `FILE_NAME(...)` is parsed as the file's header. | `stops at the first ENDSEC so DATA-section text is never read as a header` |
| Bare-value list tolerance dropped | Writers that emit `FILE_SCHEMA('IFC4')` without the inner list lose the schema token. | `accepts a bare value where a list was expected` |
| Quote-awareness dropped in `extractRecordArgs` | A `)` inside a quoted filename ends the record early; every later FILE_NAME field is lost. | `does not end the record at a closing parenthesis inside a quoted string` |
| FILE_NAME positional read shifted by one | Authorization reads the originating-system slot. | `maps every FILE_NAME slot to its own field` (all seven values distinct, so a shifted index cannot land on a match) |

### `src/entity-refs-from-index.ts` — the parser worker's fast path

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| Column-length mismatch guard only checks `starts` | A truncated `lengths` column passes through as undefined spans instead of failing loudly. | ``throws when the `lengths` column is shorter than the `ids` column`` |
| Out-of-bounds span end check dropped | A span that walks off the buffer yields a truncated record with an empty type name rather than surfacing the corruption. | `throws when a span ends past the end of the source` |

### `src/attribute-helpers.ts`

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| `.F.` no longer maps to `false` | Callers treat `undefined` as "not stated" and fall back to a default, so an explicit STEP "no" becomes whatever the default happens to be. | ``maps the STEP `.F.` logical to false`` |
| `getString` non-string coercion removed | A numeric IFC attribute (name, identifier) blanks out entirely. | `coerces a non-string scalar to its string form` |

## Weak-shaped existing test found

`packages/export/src/source-header-roundtrip.test.ts` is the only pre-existing test that touched the header reader, and it carries three of the shapes this campaign keeps hitting:

1. **An expected value equal to the same constant the implementation uses.** Every fixture declares `implementation_level` as `'2;1'` — byte-identical to the parser's hard-coded fallback `let implementationLevel = '2;1'`. A probe mutation that stops parsing the field *entirely* (`if (false) implementationLevel = ...`) leaves `expect(sh!.implementationLevel).toBe('2;1')` **passing**. The new `reads the declared implementation_level rather than falling back to the default` uses `3;7`, and a separate test pins the fallback.
2. **Fixture constants that make the bug cancel.** No fixture contains a STEP escape of any kind (`''`, `\X2\`, `\X4\`, `\S\`, `\X\`, `\P?\`), a `$`/`*` sentinel, or a quoted delimiter — which is why eight independent mutations to `source-header.ts` survived.
3. **Expectations derived from the code under test.** Its `exportedHeader()` helper re-parses the exported bytes with `parseSourceHeader`, the function under test, so a symmetric read/write escaping bug cancels on both sides.

That test is left in place and still passes; the new parser-side tests assert against literal expected values instead of a round trip.

## Survivors classified

Two mutations still survive. Both are **EQUIVALENT**, proven:

- **`cacheKey` drops the length component** (`` `${typeLen}:${typeHash}` `` → `` `${typeHash}` ``). The cache-hit path already re-derives correctness independently: it checks `type.length === typeLen` and then compares every byte. Dropping `typeLen` from the key can therefore only cause an extra recompute-and-overwrite on a cross-length collision, never a wrong type. Correctness-equivalent; a pure (and negligible) cache-thrash regression. A test is included anyway (`keeps colliding type names distinct even when their lengths differ`, using `DHMWK`/`LILQAGG`) as a regression guard should the verification ever be refactored away.
- **`findEntityLength`'s doubled-quote skip disabled.** With the skip, both quotes are consumed and `inString` is unchanged (`true`). Without it, the first quote sets `inString = false` and the second sets it back to `true` — both quotes consumed, same end state, and there is no character between them for the transient state to affect. Byte-for-byte identical behaviour on every input.

No genuine gap is left unclosed, and nothing is padded.

## Verification

- Baseline `packages/parser`: **47 files / 382 tests**. Final: **51 files / 448 tests** (+4 files, +66 tests).
- `packages/export` 363 passed / 29 skipped and `packages/query` 153 passed are unchanged.
- Dependencies built with `pnpm turbo build --filter=@ifc-lite/parser^...` before every run; no suite silently skipped.
- No pre-existing failures found.

## Typecheck

`pnpm typecheck` covers only `apps/viewer` and `apps/viewer-embed`, and the root tsconfig excludes `**/*.test.ts`, so it does not check this work. Typechecked separately against `packages/parser/tsconfig.json` with `test/**/*` added to `include`: **exit 0, zero errors** — including every pre-existing parser test file, so there are no pre-existing errors to report.

### DIFF
--- a/packages/parser/test/attribute-helpers.test.ts
+++ b/packages/parser/test/attribute-helpers.test.ts
@@ -1,0 +1,139 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Direct coverage for the shared attribute coercions used by the material,
+ * georef, and classification extractors.
+ *
+ * Only `getReference`'s `#`-prefixed form was reachable from an existing test.
+ * The STEP logical sentinels (`.T.` / `.F.`) and the non-string coercion in
+ * `getString` both survived mutation across the whole monorepo.
+ */
+
+import { describe, expect, it } from 'vitest';
+import {
+  getBoolean,
+  getNumber,
+  getReference,
+  getReferences,
+  getString,
+  getStringList,
+} from '../src/attribute-helpers.js';
+
+describe('getString', () => {
+  it('passes a string through unchanged', () => {
+    expect(getString('wall')).toBe('wall');
+    expect(getString('')).toBe('');
+  });
+
+  it('coerces a non-string scalar to its string form', () => {
+    // IFC numeric/boolean attributes reach the extractors untyped; returning
+    // undefined here would silently blank out a name or an identifier.
+    expect(getString(42)).toBe('42');
+    expect(getString(0)).toBe('0');
+    expect(getString(true)).toBe('true');
+  });
+
+  it('returns undefined only for null and undefined', () => {
+    expect(getString(null)).toBeUndefined();
+    expect(getString(undefined)).toBeUndefined();
+  });
+});
+
+describe('getNumber', () => {
+  it('passes a number through, including zero', () => {
+    expect(getNumber(1.5)).toBe(1.5);
+    expect(getNumber(0)).toBe(0);
+  });
+
+  it('parses a numeric string', () => {
+    expect(getNumber('2.5')).toBe(2.5);
+    expect(getNumber('-3')).toBe(-3);
+  });
+
+  it('returns undefined for a non-numeric string and for non-scalars', () => {
+    expect(getNumber('abc')).toBeUndefined();
+    expect(getNumber(null)).toBeUndefined();
+    expect(getNumber(undefined)).toBeUndefined();
+    expect(getNumber(true)).toBeUndefined();
+    expect(getNumber([1])).toBeUndefined();
+  });
+});
+
+describe('getBoolean', () => {
+  it('maps the STEP `.T.` logical to true', () => {
+    expect(getBoolean('.T.')).toBe(true);
+    expect(getBoolean('T')).toBe(true);
+    expect(getBoolean('true')).toBe(true);
+  });
+
+  it('maps the STEP `.F.` logical to false', () => {
+    // `.F.` must resolve to `false`, not to `undefined`. Callers treat
+    // undefined as "not stated" and fall back to a default, so losing `.F.`
+    // turns an explicit "no" into whatever the default happens to be.
+    expect(getBoolean('.F.')).toBe(false);
+    expect(getBoolean('F')).toBe(false);
+    expect(getBoolean('false')).toBe(false);
+  });
+
+  it('passes a real boolean through', () => {
+    expect(getBoolean(true)).toBe(true);
+    expect(getBoolean(false)).toBe(false);
+  });
+
+  it('returns undefined for the unset/unknown third state', () => {
+    // `$` (unset) and `.U.` (logical unknown) are neither true nor false.
+    expect(getBoolean(null)).toBeUndefined();
+    expect(getBoolean(undefined)).toBeUndefined();
+    expect(getBoolean('.U.')).toBeUndefined();
+    expect(getBoolean('$')).toBeUndefined();
+  });
+});
+
+describe('getReference', () => {
+  it('passes a numeric reference through', () => {
+    expect(getReference(12)).toBe(12);
+  });
+
+  it('strips the `#` prefix from a string reference', () => {
+    expect(getReference('#12')).toBe(12);
+  });
+
+  it('returns undefined for a string without the `#` marker', () => {
+    expect(getReference('12')).toBeUndefined();
+    expect(getReference('#')).toBeUndefined();
+    expect(getReference('#abc')).toBeUndefined();
+  });
+
+  it('returns undefined for null, undefined and non-scalars', () => {
+    expect(getReference(null)).toBeUndefined();
+    expect(getReference(undefined)).toBeUndefined();
+    expect(getReference({})).toBeUndefined();
+  });
+});
+
+describe('getReferences', () => {
+  it('collects the resolvable references and drops the rest', () => {
+    expect(getReferences([1, '#2', null, 'nope', '#3'])).toEqual([1, 2, 3]);
+  });
+
+  it('returns undefined for a non-array (an absent aggregate, not an empty one)', () => {
+    expect(getReferences('#1')).toBeUndefined();
+    expect(getReferences(null)).toBeUndefined();
+  });
+
+  it('returns an empty array for an empty aggregate', () => {
+    expect(getReferences([])).toEqual([]);
+  });
+});
+
+describe('getStringList', () => {
+  it('coerces every entry and drops only null/undefined', () => {
+    expect(getStringList(['a', 1, null, undefined, 'b'])).toEqual(['a', '1', 'b']);
+  });
+
+  it('returns undefined for a non-array', () => {
+    expect(getStringList('a')).toBeUndefined();
+  });
+});
--- a/packages/parser/test/entity-refs-from-index.test.ts
+++ b/packages/parser/test/entity-refs-from-index.test.ts
@@ -58,12 +58,20 @@ describe('buildEntityRefsFromIndex', () => {
     expect(seen).toEqual([...seen].sort((a, b) => a - b));
   });
 
-  it('interns one string per distinct type name', () => {
+  // NOT "interns one string per distinct type name". Interning is not
+  // observable from JS: `expect(a).toBe(b)` is `Object.is`, which compares
+  // string PRIMITIVES by value, so two independently-built equal strings
+  // satisfy it. Verified — replacing the intern lookup with
+  // `(' ' + key).slice(1)`, and separately with `Array.from(key).join('')`,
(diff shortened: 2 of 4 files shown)


BIMvoice:
Verified the headline, and I want to state it more precisely than "the monorepo stayed green", because the reason matters.

`scanEntities` has **exactly one occurrence in the whole repository — its own definition** at `packages/parser/src/tokenizer.ts:25`. No caller in `packages/`, no caller in `apps/`, no test. Its sibling `scanEntitiesFast` is used (`packages/mutations/test/mutations.test.ts:77`) and documented (`packages/wasm/README.md:15-30`); the plain variant is not.

So yes, gutting the generator leaves everything green — but not because tests failed to notice a break in code under load. It is **published API that nothing inside the monorepo exercises**, reachable only by external consumers of `@ifc-lite/parser`. That is still a genuine and arguably worse gap, since a break would surface first in someone else's build rather than ours, but "dead internally, live externally" is the accurate description and it points at a different remedy than a missing assertion would.

Three things in this sweep are worth more than the count:

**The method caught its own false positives.** All 22 survivors were re-run against `packages/export` and `packages/query`, both of which alias `@ifc-lite/parser` to source rather than `dist`. Two mutations that survived the parser suite die in the export suite. Reported without that cross-check, this would have claimed 24 gaps instead of 21 — a 14% overstatement, all in the direction that flatters the sweep.

**A control mutation survived, and was promoted rather than discarded.** The first intended control — removing `getString`'s non-string coercion — was expected to die and did not. Rather than swap in a control that worked and move on, it was kept as finding M26. A control that unexpectedly survives is a finding, not a bad control.

**A pre-existing test in `packages/export` carries three of the campaign's recurring shapes at once** (`source-header-roundtrip.test.ts`):

1. *Expected value equals the implementation's own constant.* Every fixture sets `implementation_level` to `'2;1'`, byte-identical to the parser's hard-coded fallback. Proven by probe: stopping the field from being parsed **at all** leaves `expect(sh!.implementationLevel).toBe('2;1')` passing.
2. *Fixture constants that cancel the bug.* No fixture contains a STEP escape, a `$`/`*` sentinel, or a quoted delimiter — that single omission accounts for 8 of the 19 genuine gaps.
3. *Expectation derived from the code under test.* Its `exportedHeader()` helper re-parses using `parseSourceHeader` itself, so a symmetric escaping bug cancels on both sides.

It was left in place and still passes; the new tests assert literal expected values instead. Worth noting that this is now the **fifth** package where the same fixture-cancellation shape has turned up, and the first where three variants coexist in one file.

Two of the genuine gaps have real user consequences rather than theoretical ones: a type-cache hit was no longer verified byte-for-byte, so `Aa` and `BB` — same length, same hash — could resolve a door as a wall in a hostile file; and the `*` derived sentinel leaked as a literal asterisk into exported headers.

The 21% → 96% kill ratio is flagged **targeted**: the mutations aimed at five modules with no test file at all (~1,700 LOC across ten such modules), so it measures those modules, not the package.

BIMvoice:
## Self-audit round 3 — two unfalsifiable assertions

Pushed as b9892bb…b14bd1b. **13 of 14 probative mutations killed.** The survivor is a different failure mode from the rest of this round: not a weak fixture, but an assertion that *cannot fail*.

| test | claimed property | discriminates | evidence |
|---|---|---|---|
| `entity-refs-from-index.test.ts` — interns one string per distinct type name | the intern `Map` | **no** | see below |
| `tokenizer.test.ts` — reuses one interned string for a repeated type name | the type-name cache | **no** | see below |

Both spelled the claim as `expect(a).toBe(b)` on `ref.type`. That is `Object.is` on string **primitives**, i.e. value equality — two independently-built equal strings satisfy it. Demonstrated rather than argued; three mutations that remove the sharing outright all left the files green:

- entity-refs: intern lookup → `(' ' + key).slice(1)` — **SURVIVED**
- entity-refs: intern lookup → `Array.from(key).join('')` — **SURVIVED**
- tokenizer: `typeCache.set(cacheKey, type)` deleted — **SURVIVED**

Interning here is a memory optimisation with no behavioural contract, so there is nothing to strengthen the assertion *to*. Both tests are renamed to the falsifiable property they actually carry (a repeated type reads back as the right name), and each now states plainly why the identity spelling must not be read as pinning the interning — so the next reader does not re-derive it.

The real, falsifiable half of the same story is untouched and sharp: `does not alias two type names that collide in the type-name cache` kills the removal of the cache-hit byte verification, which is the mutation that would silently read a door as a wall.

**What held.** `attribute-helpers`: 6/6 — `getString` on `null`, the `isNaN` gate in `getNumber`, `getBoolean`'s unset third state, `getReference`'s `#` requirement, and both `undefined`-vs-`[]` distinctions for absent aggregates. `source-header`: the `implementation_level` fallback (the exact defect found two rounds ago — now genuinely pinned), the FILE_NAME slot positions, and the no-records-return-undefined guard. `entity-refs-from-index`: ascending-id ordering, the out-of-bounds throw in both directions (clamping instead of throwing, and rejecting a span that ends exactly at EOF), and the whitespace skip after `=`.

**Discarded as non-probative:** dropping the length component from the tokenizer's `length:hash` cache key. It changes only the collision *rate* — the byte-for-byte verification still rejects a false hit, so no observable behaviour changes and no test can kill it.

PR #2164, opened 2026-08-05
test(export): close 40 mutation-verified gaps in a second, deeper sweep

BIMvoice:
Test-only. **No production code changed** — `git diff --stat` is eight test files.

`packages/export` was swept at the very start of this campaign, when the method was least developed. This is a second, deeper pass. It starts from the modules with **no test file, or a test file far thinner than the source** — the place every deep pass has found its survivors.

## Result

| | |
|---|---|
| Mutations | 40 (+ 2 controls) |
| Killed by the existing suite | **9 (22.5%)** |
| Survivors | 31 — all closed here |
| Tests | 363 passed / 29 skipped → **448 passed / 29 skipped** |
| Test files | 21 → 27 (2 skipped unchanged) |

Kill ratio is **TARGETED**: the mutations were aimed at the thin and untested modules, not sampled uniformly across the package.

**Controls** (deliberate, expected to die, run before believing any survivor) — both died:

| | Mutation | Result |
|---|---|---|
| C1 | `filterColumns`: `if (!keep) return columns` → `if (true) return columns` | KILLED — 3 tests in `parquet-exporter.test.ts` |
| C2 | `countGlbMeshes`: `return json.meshes.length` → `return 0` | KILLED — `counts the meshes that are present` |

## Modules with no test file, or a test file far thinner than the source

| Module | src | own test | chosen |
|---|---|---|---|
| `parquet-exporter.ts` | 629 | 157 (deletions only) | yes |
| `glb.ts` | 182 | 65 (1 of 4 exports) | yes |
| `lod0-generator.ts` | 320 | 67 | yes |
| `demesh-prune.ts` | 188 | none | yes |
| `select-qualification.ts` | 153 | none | yes |
| `lod-geometry-utils.ts` | 114 | none | yes |
| `type-owned-psets.ts` | 108 | none | yes |
| `entity-iteration.ts` | 83 | none | yes |
| `attribute-real-slots.ts` | 134 | none directly | covered via `attribute-slot-types.test.ts` |
| `demesh-session.ts` | 338 | none directly | covered via `demesh-writer.test.ts` |
| `unit-normalize.ts` | 377 | 207 | measured, already strong |

`unit-normalize.ts` — the obvious unit-conversion target — was read and judged already well covered (its test pins the FP-noise rounding, the exponent form, the unit-override guard, the excluded unit-definition entities and both `IfcPropertyTableValue` directions). It was left alone.

### `parquet-exporter.ts` — the geometry writers, column typing and metadata had NO test

`parquet-exporter.test.ts` pins overlay deletions and nothing else. `.bos` is consumed by other vendors' tools, so every one of these is a file that opens fine and is wrong inside.

| # | Mutation | replacements | before | killed by |
|---|---|---|---|---|
| M1 | `writeVertexBuffer`: drop the per-mesh origin fold on X | 1 | SURVIVED | `bakes the per-mesh origin into world vertices on all three axes` |
| M2 | `writeVertexBuffer`: swap the Y and Z columns | 1 | SURVIVED | `keeps the X/Y/Z column split in axis order…` (+2) |
| M3 | `writeMeshes`: `VertexCount` un-divided by 3 | 1 | SURVIVED | `emits vertex/index starts that accumulate across meshes…` |
| M4 | `writeMeshes`: `IndexStart` never advances | 1 | SURVIVED | same, + `has starts that index the VertexBuffer/IndexBuffer rows…` |
| M5 | `writeIndexBuffer`: swap `Index1`/`Index2` | 1 | SURVIVED | `splits the flat index stream into Index0/1/2 in the order it was given` |
| M6 | `toParquet`: drop the Float64 forcing for declared-REAL columns | 1 | SURVIVED | `keeps a whole-number REAL property value in a Float64 column` (+1) |
| M7 | `writeProperties`: lose the 255 "null boolean" sentinel | 1 | SURVIVED | `reports ValueBool as null for a non-boolean property` |
| M8 | `writeQuantities`: Formula guard `i > 0` → `i >= 0` | 1 | SURVIVED | `reports a quantity with no formula as null rather than string index 0` |
| M9 | `writeMetadata`: `statistics.vertexCount` always 0 | 1 | SURVIVED | `reports the geometry statistics of the export it accompanies` |
| M10 | `writeEntities`: read `HasGeometry` off the `IS_TYPE` bit | 1 | SURVIVED | `reads HasGeometry and IsType from their own flag bits` |

M1 is the **identity-fixture** shape: with a mesh at origin `(0,0,0)` the fold is a no-op. The new fixture uses `[10, 20, 30]` — non-zero and distinct per axis, so a fold dropped on any single axis shows. M3/M4 needed **two meshes of different sizes**: one mesh leaves every start at 0, and equal sizes cannot tell a vertex stride from an index stride.

M6 needed a **whole-number** REAL (`3`, `1200`): content inference alone demotes those to Int32, losing the float schema. A counter-example test pins that `ExpressId` stays integer, so "declare every numeric column Float64" does not pass.

### `glb.ts` — 2 of 4 exports had no test

| # | Mutation | replacements | before | killed by |
|---|---|---|---|---|
| M11 | `extractGlbMapping`: swap node index and mesh index | 1 | SURVIVED | `maps each express id to its own node index and its own mesh index` |
| M12 | `parseGLBToMeshData`: drop the accessor `byteOffset` | 1 | SURVIVED | `reads each attribute from its own accessor byteOffset, not the buffer-view base` |
| M13 | material alpha always 1 | 1 | SURVIVED | `resolves the material baseColorFactor including a non-1 alpha` |
| M14 | `DEFAULT_COLOR` turns black | 1 | SURVIVED | `falls back to the neutral grey default…` (+1) |
| M15 | `parseGLB`: drop the total-length sanity check | 1 | SURVIVED | `rejects a GLB whose declared total length exceeds the buffer it arrived in` |

A wrong accessor offset does not throw — it reads the neighbouring attribute's bytes. The fixture's normals are deliberately unequal to its positions, and the BIN lead-in is `0xff` rather than zero, so aliasing either is visible. M13 is the identity shape again: an opaque fixture (alpha 1) cannot see the alpha slot being dropped, so the fixture uses `0.5`. Material resolution is exercised in all four states — RGBA factor, RGB-only factor, material with no factor, and primitive with no material.

### `lod0-generator.ts` — every fixture was metre-scaled with its corner at the origin

Two constants that make the unit multiply and the corner offset **identities**. Six mutations survived because of them.

| # | Mutation | replacements | before | killed by |
|---|---|---|---|---|
| M29 | `IfcBoundingBox.XDim` ignores the length-unit scale | 1 | SURVIVED | `applies the length-unit scale to placements, corners and dimensions alike` |
| M30 | cartesian point coordinates ignore the length-unit scale | 1 | SURVIVED | same (+2) |
| M31 | `IfcBoundingBox.Corner` discarded (always local origin) | 1 | SURVIVED | same (+1) |
| M32 | centroid is the bbox min, not the midpoint | 1 | SURVIVED | `reports the centroid as the bbox midpoint, not a corner` |
| M33 | declared output units flip to `mm` | 1 | SURVIVED | `declares metres regardless of the source file unit` |
| M34 | no-representation fallback cube grows 5× | 1 | SURVIVED | `falls back to a 0.2 m cube at the placement…` |

The new fixture is a **millimetre** model (the ordinary case in the wild) with an **off-origin** bounding-box corner, distinct per axis, plus an element whose `Representation` is `$` so the fallback branch is reached at all.

### `demesh-prune.ts` — no test file

| # | Mutation | replacements | before | killed by |
|---|---|---|---|---|
| M35 | `IFCOWNERHISTORY` loses its `PROTECTED_TYPES` entry | 1 | SURVIVED | `never tombstones shared infrastructure orphaned by the sweep (IfcOwnerHistory)` |
| M36 | presentation layers count as referrers | 1 | KILLED | `prunes geometry whose only surviving referrer is a presentation layer…` |
| M37 | an emptied presentation layer is left behind | 1 | KILLED | same |
| M38 | the detached product→old-rep edge stays in the reverse index | 1 | KILLED | 4 tests |
| M39 | `strippedOpeningCount` counts every tombstone | 1 | SURVIVED | `counts ONLY openings into strippedOpeningCount` (+1) |
| M40 | styled items never pulled into the closure | 1 | SURVIVED | `prunes the style chain that hangs off replaced geometry, leaving no dangling item` |

**Weak-shaped existing test.** M39 survived because `strippedOpeningCount` was asserted with `expect(...).toBeGreaterThan(0)` — equally true of a counter that counts *every* tombstone, which is a strictly larger number on that fixture. It is now pinned to the exact count (2), with `prunedEntityCount > strippedOpeningCount` and a zero-openings counter-example.

### Four smaller modules with no test file

| # | Module | Mutation | replacements | before | killed by |
|---|---|---|---|---|---|
| M16 | `entity-iteration.ts` | stop iterating the deferred index | 1 | KILLED | 4 exporter tests |
| M17 | `entity-iteration.ts` | `size` ignores deferred atoms | 1 | SURVIVED | `counts the deferred atoms in size…` (+1) |
| M18 | `entity-iteration.ts` | `getMaxExpressId` off by one | 1 | KILLED | 10 exporter tests |
| M19 | `type-owned-psets.ts` | `()` for an empty `HasPropertySets` | 1 | SURVIVED | `emits $ for an empty list, never an empty aggregate` |
| M20 | `type-owned-psets.ts` | drop the append-unnamed-replacements pass | 1 | KILLED | 5 tests |
| M21 | `type-owned-psets.ts` | an ABSENT class treated as a type object | 1 | SURVIVED | `treats an ABSENT class as an occurrence rather than guessing` |
| M22 | `lod-geometry-utils.ts` | empty-point AABB returns a unit box | 1 | SURVIVED | `returns a degenerate box at the origin for no points, not a unit box` |
| M23 | `lod-geometry-utils.ts` | negate the cross-product handedness | 1 | KILLED | `lod0-generator.test.ts` |
| M24 | `lod-geometry-utils.ts` | drop the translation column | 1 | KILLED | `lod0-generator.test.ts` |
| M25 | `lod-geometry-utils.ts` | `toIfcArrayBuffer` ignores the view window | 1 | SURVIVED | `honours the byteOffset/byteLength window of a partial view` (+1) |
| M26 | `select-qualification.ts` | ambiguous family no longer bails out | 1 | KILLED | `honours the { typed } marker…` |
| M27 | `select-qualification.ts` | entity refs (`#5`) become qualifiable | 1 | SURVIVED | `declines an entity reference even in a slot with one unambiguous STRING member` |
| M28 | `select-qualification.ts` | boolean prefers `IfcLogical` over `IfcBoolean` | 1 | SURVIVED | `qualifies a bare boolean as IfcBoolean, not IfcLogical` |

What each survivor means for a user:

- **M17** — `size` is what `StepExporter` (`step-exporter.ts:1069`) and `MergedExporter` (`merged-exporter.ts:611`) report as the export's entity total. A `deferPropertyAtomIndex` parse undercounts by every deferred property atom while the file itself is complete.
- **M19** — `HasPropertySets` is `OPTIONAL SET [1:?]`; `()` violates the lower bound and a strict validator rejects the file.
- **M21** — a class no bundled schema declares has no confirmed slot 5; guessing "type object" writes a pset list into whatever attribute happens to sit there.
- **M25** — `toIfcArrayBuffer` is the FIRST thing both LOD generators do to their input (`lod0-generator.ts:80`, `lod1-generator.ts:138`). Returning the whole backing buffer for a `Uint8Array` that is a *window* onto a larger one — the ordinary result of slicing bytes out of a container — hands the parser bytes that are not the model.
- **M27** — qualifying `#5` emits `IFCDESCRIPTIVEMEASURE('#5')`: a quoted string where the file needs a live link, so the reference is lost.
- **M28** — `IfcPropertySingleValue.NominalValue` is `IfcValue`, which has exactly one BOOLEAN leaf and exactly one LOGICAL leaf, so both resolve unambiguously and only the declared preference decides. The mutation writes `IFCLOGICAL(.T.)` where a strict reader expects `IFCBOOLEAN(.T.)`.

M23/M24 died against `lod0-generator.test.ts`, which is worth recording: the matrix maths *is* covered transitively even though the module has no test file. Only its degenerate paths were open.

## Fixture shapes deliberately avoided

Each of these is a shape that has beaten this campaign before:

- **Identity constants** — origin `(0,0,0)`, unit scale `1.0`, bbox corner at the local origin, opaque material alpha. Replaced with distinct non-zero, per-axis values and a millimetre model.
- **A property that holds either way** — `toBeGreaterThan(0)` on a counter (M39, fixed).
- **A loop carried by one lucky case** — two meshes of *different* sizes so a vertex stride cannot pass for an index stride; a descending id map so "return the last id" cannot pass for a maximum.
- **A two-valued signal with an untested third state** — material resolution in all four states; `deferredEntityIndex` absent / empty / populated; `Representation` present and `$`; metadata statistics with and without geometry.
- **Symmetric pairs** — the GLB fixture has two nodes and one mesh so node index (1) ≠ mesh index (0), and asymmetric entity flags so `HAS_GEOMETRY` cannot read as `IS_TYPE`.

## Typecheck

`pnpm typecheck` covers 2 of 40 packages and the root `tsconfig.json` excludes `**/*.test.ts`, so it does not check this work. I typechecked `packages/export/src/**/*` separately against the package's own `tsconfig.packages.json` options with the test exclusion lifted (29 test files in the program, verified via `--listFiles`). **All new and modified files are clean.**

**Pre-existing errors surfaced, not fixed here** (15, all invisible to CI because build and typecheck both exclude `**/*.test.ts`):

- `src/ifc5-exporter.test.ts` — 7 × `TS2345`, `null` passed where `MutablePropertyView | undefined` is declared.
- `src/parquet-exporter.test.ts` — 6, from `src/vendor-types.d.ts` declaring deliberately narrow ambient modules for `apache-arrow`/`parquet-wasm` that cover only the write side the exporter uses: `tableFromIPC` and `readParquet` are not declared, plus `TS7006` on the row callback and `TS2339` on `RelationshipType.Contains`.
- `src/merged-exporter.test.ts` — 2 × `TS2339`, `.set` on `EntityByIdIndex`.

The new `parquet-geometry.test.ts` reaches the Parquet read side through an explicitly typed dynamic import rather than widening `vendor-types.d.ts`, so it adds no new instance of that error and leaves the exporter's own ambient contract alone.

## Method notes

- Dependencies built first (`pnpm turbo build --filter=@ifc-lite/export^...`) before any run.
- Every mutation was an exact-substring replacement asserting `n == 1`, with the mutated region **re-read from disk** and confirmed changed before any result was believed.
- Every restore copied back a backup taken **before** that file's first mutation — never `git checkout --`, never a backup re-taken after a run.
- The full package suite was run for every mutation (not a subset), so a kill in another file could not be missed.
- Two parquet mutations were re-run after the async refactor of the decode helpers to confirm the tests still discriminate.
- Skipped suites are unchanged (2 files, 29 tests) and are all fixture-gated on models fetched by `pnpm fixtures`; no suite silently skips.

### DIFF
--- a/packages/export/src/demesh-writer.test.ts
+++ b/packages/export/src/demesh-writer.test.ts
@@ -395,3 +395,95 @@ ${FOOTER}`;
     expect(danglingRefs(out)).toEqual([]);
   });
 });
+
+/**
+ * Three prune behaviours that no fixture above reaches. A mutation sweep left
+ * the suite green on all three: deleting `IFCOWNERHISTORY` from
+ * `PROTECTED_TYPES`, counting EVERY tombstone into `strippedOpeningCount`
+ * (`toBeGreaterThan(0)` holds either way), and never pulling styled items into
+ * the closure.
+ */
+
+// FIXTURE_SINGLE, except the opening element carries an IfcOwnerHistory that
+// NOTHING else in the file references. When the opening falls, the history is
+// orphaned inside the closure and only PROTECTED_TYPES keeps it alive.
+const FIXTURE_ORPHANED_OWNER = `${HEADER}
+#95=IFCOWNERHISTORY($,$,$,.ADDED.,$,$,$,0);
+#10=IFCWALL('0wall10000000000000000',$,'A',$,$,$,#100,$,$);
+#100=IFCPRODUCTDEFINITIONSHAPE($,$,(#110));
+#110=IFCSHAPEREPRESENTATION(#8,'Body','SweptSolid',(#120));
+#120=IFCEXTRUDEDAREASOLID(#130,#131,#132,2.);
+#130=IFCRECTANGLEPROFILEDEF(.AREA.,$,$,1.,1.);
+#131=IFCAXIS2PLACEMENT3D(#5,$,$);
+#132=IFCDIRECTION((0.,0.,1.));
+#200=IFCOPENINGELEMENT('0open00000000000000000',#95,$,$,$,$,#210,$,$);
+#210=IFCPRODUCTDEFINITIONSHAPE($,$,(#211));
+#211=IFCSHAPEREPRESENTATION(#8,'Body','SweptSolid',(#212));
+#212=IFCEXTRUDEDAREASOLID(#213,#131,#132,1.);
+#213=IFCRECTANGLEPROFILEDEF(.AREA.,$,$,0.5,0.5);
+#220=IFCRELVOIDSELEMENT('0rvoid0000000000000000',$,$,$,#10,#200);
+${FOOTER}`;
+
+// FIXTURE_SINGLE with no opening at all, plus a style chain hanging OFF the
+// replaced solid (#120). IfcStyledItem points AT the geometry, so forward
+// reachability from the representation root never finds it.
+const FIXTURE_STYLED = `${HEADER}
+#10=IFCWALL('0wall10000000000000000',$,'A',$,$,$,#100,$,$);
+#100=IFCPRODUCTDEFINITIONSHAPE($,$,(#110));
+#110=IFCSHAPEREPRESENTATION(#8,'Body','SweptSolid',(#120));
+#120=IFCEXTRUDEDAREASOLID(#130,#131,#132,2.);
+#130=IFCRECTANGLEPROFILEDEF(.AREA.,$,$,1.,1.);
+#131=IFCAXIS2PLACEMENT3D(#5,$,$);
+#132=IFCDIRECTION((0.,0.,1.));
+#500=IFCSTYLEDITEM(#120,(#501),$);
+#501=IFCSURFACESTYLE('OldStyle',.BOTH.,(#502));
+#502=IFCSURFACESTYLESHADING(#503,0.);
+#503=IFCCOLOURRGB($,1.,0.,0.);
+${FOOTER}`;
+
+describe('applySimplifiedGeometry — prune edges', () => {
+  it('never tombstones shared infrastructure orphaned by the sweep (IfcOwnerHistory)', async () => {
+    const { store, view, editor } = await loadStore(FIXTURE_ORPHANED_OWNER);
+    const report = applySimplifiedGeometry(store, editor, [{ expressId: 10, ...TETRA }]);
+    expect(report.replaced).toEqual([10]);
+
+    const out = exportText(store, view);
+    // The opening that referenced it IS gone — so this is genuinely the
+    // orphaned case, not a history kept alive by a surviving referrer.
+    expect(out).not.toMatch(/IFCOPENINGELEMENT/);
+    expect(out).toMatch(/#95=IFCOWNERHISTORY/);
+    expect(danglingRefs(out)).toEqual([]);
+  });
+
+  it('counts ONLY openings into strippedOpeningCount', async () => {
+    const { store, view, editor } = await loadStore(FIXTURE_SINGLE);
+    const report = applySimplifiedGeometry(store, editor, [{ expressId: 10, ...TETRA }]);
+
+    // Exactly the IfcRelVoidsElement (#220) and the IfcOpeningElement (#200).
+    // `toBeGreaterThan(0)` is equally true of a counter that counts every
+    // tombstone, which is a strictly larger number here.
+    expect(report.strippedOpeningCount).toBe(2);
+    expect(report.prunedEntityCount).toBeGreaterThan(report.strippedOpeningCount);
+  });
+
+  it('reports zero stripped openings for a model that has none', async () => {
+    const { store, editor } = await loadStore(FIXTURE_STYLED);
+    const report = applySimplifiedGeometry(store, editor, [{ expressId: 10, ...TETRA }]);
+    expect(report.strippedOpeningCount).toBe(0);
+    expect(report.prunedEntityCount).toBeGreaterThan(0);
+  });
+
+  it('prunes the style chain that hangs off replaced geometry, leaving no dangling item', async () => {
+    const { store, view, editor } = await loadStore(FIXTURE_STYLED);
+    applySimplifiedGeometry(store, editor, [{ expressId: 10, ...TETRA }]);
+    const out = exportText(store, view);
+
+    // The old solid is gone...
+    expect(out).not.toMatch(/IFCEXTRUDEDAREASOLID/);
+    // ...and so is the styled item that pointed at it. Matched by its own id
+    // so the NEW style chain this export writes cannot satisfy the assertion.
+    expect(out).not.toMatch(/#500=IFCSTYLEDITEM/);
+    expect(out).not.toMatch(/'OldStyle'/);
+    expect(danglingRefs(out)).toEqual([]);
+  });
+});
--- a/packages/export/src/entity-iteration.test.ts
+++ b/packages/export/src/entity-iteration.test.ts
@@ -1,0 +1,127 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `entity-iteration.ts` had no test file. Its iteration half is well pinned
+ * indirectly — dropping `yield* deferred` fails four exporter tests — but its
+ * `size` is not: reporting only the primary index's size left the whole suite
+ * green, and `size` is what `StepExporter`/`MergedExporter` report as the
+ * export's entity total (`step-exporter.ts:1069`, `merged-exporter.ts:611`).
+ *
+ * A user with a `deferPropertyAtomIndex` parse sees an export summary that
+ * undercounts by every deferred property atom — thousands on a real file —
+ * while the file itself is complete.
+ */
+
+import { describe, it, expect } from 'vitest';
+import type { IfcDataStore } from '@ifc-lite/parser';
+import { getCompleteEntityIndex, getMaxExpressId, type ExportEntityRef } from './entity-iteration.js';
+
+const ref = (type: string, byteOffset: number): ExportEntityRef => ({ type, byteOffset, byteLength: 10 });
+
+/**
+ * `byId` and `deferredEntityIndex` are DISJOINT by construction and here have
+ * DIFFERENT sizes (2 vs 3) — equal sizes would let `byId.size + deferred.size`
+ * and `deferred.size * 2` agree, and either alone would be half the total.
+ */
+function storeWithDeferred(): IfcDataStore {
+  return {
+    entityIndex: {
+      byId: new Map([
+        [1, ref('IFCWALL', 0)],
+        [2, ref('IFCPROPERTYSET', 20)],
+      ]),
+    },
+    deferredEntityIndex: new Map([
+      [7, ref('IFCPROPERTYSINGLEVALUE', 40)],
+      [8, ref('IFCPROPERTYSINGLEVALUE', 60)],
+      [9, ref('IFCQUANTITYAREA', 80)],
+    ]),
+  } as unknown as IfcDataStore;
+}
+
+function storeWithoutDeferred(): IfcDataStore {
+  return {
+    entityIndex: {
+      byId: new Map([
+        [1, ref('IFCWALL', 0)],
+        [2, ref('IFCPROPERTYSET', 20)],
+      ]),
+    },
+  } as unknown as IfcDataStore;
+}
+
+describe('getCompleteEntityIndex', () => {
+  it('counts the deferred atoms in size, not just the primary index', () => {
+    expect(getCompleteEntityIndex(storeWithDeferred()).size).toBe(5);
+  });
+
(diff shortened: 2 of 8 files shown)


BIMvoice:
Verified the sharpest finding on clean `upstream/main`. Deleting unit conversion from the LOD0 coordinate reader outright:

```ts
- const v: Vec3 = [x * unitScale, y * unitScale, z * unitScale];   // replacements=1
+ const v: Vec3 = [x, y, z];
```

```
$ pnpm exec vitest run   (packages/export)
Test Files  21 passed | 2 skipped (23)
     Tests  363 passed | 29 skipped (392)
```

Unchanged from baseline. Restored.

`lod0-generator.test.ts` does exist on `main` — it simply never mentions `unitScale`, and the only millimetre fixtures anywhere in `packages/export` live in `merged-exporter.test.ts`. So every LOD0 fixture is metre-scaled, `unitScale` is `1`, and the multiply is an identity on every input the suite provides. Combined with each fixture's bounding-box corner sitting at the local origin, **two fixture constants made six mutations unkillable**. A millimetre model with an off-origin corner closes all six.

For a user this is the difference between a footprint at 1:1 and one at 1:1000 — an LOD0 export of a millimetre-authored model, which is most real IFC, would be silently wrong by three orders of magnitude while the suite stayed green.

**This is the seventh distinct instance of the same class today**, and at seven it is clearly a property of how fixtures get written rather than a run of bad luck:

| package | the constant that cancelled the operation |
|---|---|
| geometry | seven fixtures omitted `origin`, so an origin fold was deletable in two code paths |
| pointcloud | classification values below the `& 0x1f` mask threshold |
| solar | every fixture at longitude 0, where the longitude term vanishes |
| viewer store | single-model sessions, where the federated id offset is the identity |
| export (parquet) | no fixture with a non-zero per-mesh origin |
| export (lod0) | metre scale **and** corner-at-origin — two at once, six survivors |
| parser/export header | no escape, sentinel, or quoted delimiter in any fixture — 8 survivors |

Every one of these executes the line under test. Coverage tooling scores all seven as covered. The operative question when writing a fixture is not "does this reach the code" but "would this input distinguish the implementation from a no-op" — and a 1, a 0, an identity, or a default is exactly where that distinction disappears.

Two more from this sweep worth noting:

- **`strippedOpeningCount` was asserted with `toBeGreaterThan(0)`** — equally true of a counter that counts every tombstone, which on that fixture is a strictly larger number. Now pinned to the exact count with a zero-openings counter-example. Same shape as the `deepStrictEqual(first, second)` cache assertion in `apps/viewer` and the `originShift === 0` tautology in `packages/geometry`.
- **The GLB parser's two offset-reading exports had no test at all.** Dropping an accessor's `byteOffset` does not throw — it silently returns a mesh whose normals are its positions.

**15 pre-existing typecheck errors** surfaced and were left alone: `ifc5-exporter.test.ts` (7), `parquet-exporter.test.ts` (6, from `vendor-types.d.ts` declaring ambient modules that omit the Parquet read side), `merged-exporter.test.ts` (2). Invisible to CI because build and typecheck both exclude test files and `packages/export` defines no typecheck script.

Worth flagging one methodological point from this run: the agent's first typecheck attempt was **silently vacuous** — the root `exclude` dropped every test file, so `tsc` passed having compiled none of them. It caught that by checking `--listFiles` and confirming 29 test files actually entered the program. A green typecheck that compiled nothing looks identical to a green typecheck that compiled everything.

BIMvoice:
## Self-audit round 3 — one commuting fixture

Pushed as dd0293f…f08cf4b. **27 of 28 probative mutations killed.** One survivor.

|---|---|---|---|---|
| `lod-geometry-utils.test.ts` — `mat4Mul` composes parent-then-child **in that order** | operand order | **no** | swap `a`/`b` inside `mat4Mul` | SURVIVED |

Both operands were built with an **identity basis** — i.e. two pure translations. Translations commute, so `mat4Mul(a, b)` and `mat4Mul(b, a)` are the same matrix and the order the test is named for is unobservable. Replaced with a rotation and a translation (which do not commute), asserting *both* orders explicitly so a swap cannot pass as the same answer, plus a second case pinning `(A·B)·p === A·(B·p)` independently of the fixture.

**What held.** `type-owned-psets`: 7/7 — the `undefined`-class direction, inheritance-chain vs suffix test for `isTypeClass`, the slot-5 index, `$` for an empty list, deletion of an unreplaced pset, appending unused replacements, and not double-appending one already swapped in. `lod-geometry-utils`: the `toIfcArrayBuffer` view window *and* the no-copy fast path, `aabbFromPoints` degenerate-vs-unit box and per-axis tracking, `vec3Normalize`'s epsilon. `entity-iteration`: 4/4, including both directions of the deferred-index fast path (taken when it should not be, skipped when it should be). `parquet-geometry` is the strongest file in the PR — **10/10**: per-axis origin baking, origin-invariant normals, the X/Y/Z column split, the Index0/1/2 split, vertex starts in vertices not floats, index counts in indices not triangles, start-before-increment ordering, the `255` `ValueBool` sentinel, `Formula` index 0 as null, and the `HasGeometry` flag bit.

**Discarded as non-probative:** `psetName &&` in `resolveTypeOwnedPsetIds` — a `Set<string>` never contains `null`, so removing the truthiness guard cannot change an outcome. Narrowing produced no probative variant.

PR #2165, opened 2026-08-05
test(mutations,ifcx): close 21 mutation-verified gaps in the IFCX write path

BIMvoice:
Second, deeper mutation sweep over `packages/mutations` and `packages/ifcx`. Both had a shallow first pass early in this campaign. These are the **write** path — mutations edit model data, IFCX is the interchange format — so a silent wrong answer here corrupts what a user *saves*, not what they see.

**Test-only. No production code is changed.**

## Where I looked, and why

Modules with **no test file at all**:

| module | LOC | has caller in repo? |
|---|---:|---|
| `mutations/src/change-set.ts` | 213 | **no** — definition + barrel re-export only |
| `ifcx/src/writer.ts` | 375 | **no** — definition + barrel re-export only |
| `ifcx/src/traversal.ts` | 157 | yes — geometry, point-cloud and entity extractors |
| `ifcx/src/index.ts` | 701 | yes |
| `ifcx/src/federated-composition.ts` | 453 | yes |
| `ifcx/src/layer-stack.ts` | 350 | yes |
| `ifcx/src/provenance.ts` | 288 | yes — and covered downstream (`merge`, `collab-server`, `viewer`) |
| `ifcx/src/composition.ts` | 223 | partly — `findRoots`/`getDescendants` have **no** caller |
| `ifcx/src/bake.ts` | 125 | yes — covered by `tombstones.test.ts` |

I chose `change-set.ts`, `writer.ts`, `traversal.ts` and the uncalled half of `composition.ts`. `provenance.ts` was ruled out after checking: it has no test file *in ifcx*, but `validateProvenance` / `createProvenanceManifest` / `getProvenance` / `setProvenance` are exercised by `packages/merge`, `packages/collab-server` and `apps/viewer`, so it is not the gap it looks like from inside the package.

**Exported API with no in-repo caller** (grep for each exported symbol; a count of 2 means definition + barrel and nothing else):

- `ChangeSetManager` — 2 occurrences
- `exportToIfcx` — 2 occurrences
- `getDescendants` — 2 occurrences
- `findRoots` (composition.ts) — 5, but every one is the barrel or `federated-composition.ts`'s own *local* `findRoots`; the exported one has no caller

Published but unexercised, exactly the `StepTokenizer.scanEntities` shape.

## Method

24 mutations. Each applied as a single-occurrence substring replacement with the replacement count asserted `== 1`; the mutated region re-read from disk and asserted to contain the new text and **not** the old before any result was believed; restored by copying back a backup taken **before** that file's first mutation (never `git checkout --`).

**Controls, run first — both died:**

| control | result |
|---|---|
| `composeIfcx` returns `composed` instead of `applyTombstones(composed)` | **KILLED** (4 failures) |
| composition drops the null-attribute removal mask | **KILLED** (2 failures) |

**A false result avoided:** the first downstream run reported `Test Files 8 failed` for `merge`, 20/30 for `collab-server`, 19/23 for `mcp`, 18/25 for `cli`. That was the unbuilt-workspace-dependency trap, not my mutations — after a full `pnpm turbo build` the same suites are green at 71 / 138 / 231 / 272. All downstream numbers below are against that verified-clean baseline.

## Mutation table

All 24 replacements were verified at count 1.

| # | mutation | before | after |
|---|---|---|---|
| M1 | `mergeChangeSets` drops the timestamp sort | SURVIVED | KILLED |
| M2 | `deleteChangeSet` leaves a stale active pointer | SURVIVED | **SURVIVED — equivalent, see below** |
| M3 | `importChangeSet` reuses the incoming id | SURVIVED | KILLED |
| M4 | `importChangeSet` keeps the incoming `applied` flag | SURVIVED | KILLED |
| M5 | `setActiveChangeSet` drops the unknown-id guard | SURVIVED | KILLED |
| M6 | `addMutation` auto-create branch drops the mutation | SURVIVED | KILLED |
| M7 | `getStatistics` ignores `modelId` in the entity key | SURVIVED | KILLED |
| M8 | `exportChangeSet` drops the missing-id throw | SURVIVED | KILLED |
| M9 | `collectRequiredImports` drops the `prop@v5a` import | SURVIVED | KILLED |
| M10 | `prettyPrint` default flips to compact | SURVIVED | KILLED |
| M11 | `includeProperties` default flips to off | SURVIVED | KILLED |
| M12 | `applyMutations` default flips to off | SURVIVED | KILLED |
| M13 | containment lookup consults only `byStorey` | SURVIVED | KILLED |
| M14 | `stats.fileSize` uses UTF-16 length, not UTF-8 bytes | SURVIVED | KILLED |
| M15 | empty attributes serialise as `{}` instead of being omitted | SURVIVED | KILLED |
| M16 | unmapped type falls back to a different class name | SURVIVED | KILLED |
| M17 | supplied `idToPath` is ignored | SURVIVED | KILLED |
| M18 | `findRoots` returns every node | SURVIVED | KILLED |
| M19 | `getDescendants` drops its de-dup guard | SURVIVED | KILLED |
| M20 | `collectIncomingEdgeNames` drops de-dup | SURVIVED | KILLED |
| M21 | `findTraversalSeeds` skips unreachable orphans | KILLED | KILLED |
| M22 | `buildReachableAttributeIndex` cycle guard returns `true` | SURVIVED | KILLED |
| M23 | traversal revisits nodes already on the current path | KILLED | KILLED |
| M24 | `TraversalFrame.depth` is always 0 | SURVIVED | KILLED |

**Kill ratio: 2/24 (8%) before → 23/24 (96%) after.** Flagged **TARGETED**: the mutations were aimed at modules chosen precisely because they had no test file.

## Survivors re-checked downstream

All 22 survivors were applied **together**, `ifcx` and `mutations` `dist` rebuilt, and every downstream consumer re-run. Not one survivor was killed downstream, so none was an artefact of running only the owning suite:

| suite | resolves via | baseline | with all survivors |
|---|---|---:|---:|
| `packages/merge` | built `dist` | 71 | 71 |
| `packages/diff` | built `dist` | 286 | 286 |
| `packages/collab-server` | built `dist` | 138 | 138 |
| `packages/mcp` | built `dist` | 231 | 231 |
| `packages/cli` | built `dist` | 272 | 272 |
| `apps/viewer` (10 files importing ifcx/mutations) | **aliases straight to `src`** | 69 | 69 |

## Survivor classification

21 genuine gaps, all closed here. What a user would lose:

- **Edits replayed out of order** (M1) — merged change sets came back in concatenation order, not the order the user made the edits.
- **A re-imported change set silently replacing the one already loaded** (M3).
- **Unapplied edits treated as already applied** (M4) — an imported set carrying `applied: true` is skipped on save.
- **The first edit lost** (M6) — `addMutation` with no active change set auto-creates one; the branch that stores the triggering mutation was unpinned.
- **Two federated models collapsed into one** (M7) — the same express id in model A and model B counted as one affected entity.
- **Pre-edit values written to disk** (M12) — `applyMutations` defaults to on via `!== false`; flipped, `exportToIfcx` writes the *original* property values and the user's edits never reach the file.
- **Properties or indentation silently dropped** (M10, M11) — same three-state shape.
- **A file whose attributes have no declaring import** (M9).
- **Every element contained by a building, site or space dropped from the export** (M13).
- **Round-trip identity broken** (M17) — ignoring `idToPath` renames every node, so layers authored against the old paths stop matching.
- **Wrong reported file size for any non-ASCII model** (M14).
- Plus M5, M8, M15, M16, M18, M19, M20, M22, M24.

**One EQUIVALENT survivor — M2**, and I believe this one is genuinely proven. `deleteChangeSet` nulls `activeChangeSetId` when the deleted set is the active one. `activeChangeSetId` is private and read only through `getActiveChangeSet()`, which does `this.changeSets.get(id) || null` — after the delete that lookup misses and returns `null` regardless. Every write to the field (`createChangeSet`, `setActiveChangeSet`) is unconditional. So the null-out is defensive only; the sole way to observe it would be a later `generateChangeSetId()` collision (`cs_${Date.now()}_${7 random base36 chars}`), which requires the same millisecond *and* the same suffix. Equivalent through the public API given unique ids. The behaviour it protects — that a delete leaves the manager usable and the next `addMutation` lands somewhere retrievable — **is** now tested.

## Note on the three-state shape

`prettyPrint`, `includeProperties` and `applyMutations` are all written `options.X !== false`. Not one of the three had *any* test, so this is the listed weak shape in its strongest form: not "both explicit directions tested while the default goes untested", but all three states untested, on the three flags that decide whether the user's edits reach the exported file at all. Each is now pinned omitted / true / false.

## Weak-shaped existing tests

I looked for the listed shapes and found the existing `ifcx` suites to be well built — `canonical.test.ts` in particular pairs order-*independence* with order-*dependence* assertions and tests 100k-deep nesting; `pointcloud-extractor.test.ts` uses `(1,2,3) → (1,3,-2)`, a fixture that would catch an identity swap. Nothing to flag there.

Two structural notes rather than weak fixtures:

- `packages/mutations/test/mutations.test.ts` is `describe.skipIf(!testFile)` over four candidate LFS model paths — **10 tests skip silently** when the fixtures are not materialised. That is deliberate and documented, but it means the package's real integration coverage is invisible on a fresh checkout. All tests added here are fixture-free and always run.
- `packages/ifcx/src/system-membership.test.ts` has no `system-membership.ts` source.

`pnpm typecheck` covers 2 of 40 packages and the root tsconfig excludes `**/*.test.ts`, so it does not check this work. I typechecked each package separately against its own `tsconfig.json` with the exclusion lifted and test files included (`types: ["node"]`, `rootDir` widened). Two pre-existing errors, **not fixed here**:

- `packages/ifcx/src/hierarchy-builder.test.ts:24` — TS2322, `ArrayBuffer | SharedArrayBuffer` not assignable to `ArrayBuffer`
- `packages/mutations/test/mutations.test.ts:238` — TS2322, `number | undefined` not assignable to `number`

The three files added here typecheck clean.

## Counts

| package | baseline | final |
|---|---:|---:|
| `packages/mutations` | 130 passed, 10 skipped (8 files, 1 skipped) | **149** passed, 10 skipped (9 files, 1 skipped) |
| `packages/ifcx` | 49 passed, 15 suites | **101** passed, 30 suites |

```
 packages/ifcx/src/traversal.test.ts        | 314 ++++++++++++++
 packages/ifcx/src/writer.test.ts           | 449 +++++++++++++++++++
 packages/mutations/test/change-set.test.ts | 262 ++++++++++++
 3 files changed, 1025 insertions(+)
```

## One thing to decide separately

While writing the `getDescendants` tests I hit a **production defect** and stopped rather than change behaviour. `getDescendants` returns only the *immediate* children — the recursion is dead:

```ts
for (const child of n.children.values()) {
  if (visited.has(child.path)) continue;
  descendants.push(child);
  visited.add(child.path);   // <- marks the child …
  traverse(child);           // <- … which then returns at its own `visited.has(n.path)` entry check
}
```

Probe on a plain chain `a → b → c → d`:

```
getDescendants(a) === [b]        // expected [b, c, d]
```

Impact is limited — the function has no in-repo caller, so this only affects external SDK consumers of `@ifc-lite/ifcx` — but it is wrong, and the fix is a one-line move of `visited.add`. **This PR does not touch it.** The tests here deliberately assert only the properties that hold either way (uniqueness, cycle termination, leaf case) so they will not need rewriting when it is fixed, and they do not enshrine the bug. Happy to open a separate fix if you want it.

### DIFF
--- a/.changeset/ifcx-get-descendants-one-level.md
+++ b/.changeset/ifcx-get-descendants-one-level.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/ifcx": patch
+---
+
+Fix `getDescendants` returning only the direct children instead of the whole subtree.
+
+`getDescendants` (published from the package index) marked each child visited in the loop *before* recursing into it, so the recursive `traverse` call tripped its own `if (visited.has(n.path)) return` entry guard and returned without walking anything. The function was a one-level walk: for `a -> b -> c -> d` it answered `[b]` rather than `[b, c, d]`, and for a diamond `root -> {l, r} -> shared` it answered `[l, r]`, dropping the shared grandchild entirely. The child is now marked by the recursive call that opens on it, which is where the existing cycle guard already expects the marking to happen.
+
+Found while mutation-auditing the tests added in this PR: neither test over the function could observe the truncation. One asserted only that the result is duplicate-free, which a one-level result satisfies; the other's expectation for a two-node cycle — `['b']` — is what the truncated walk produces anyway. Both now assert the reached depth directly.
+
+No in-repo caller was affected (the function has no consumer besides the export and its tests), so this only changes behaviour for external users of `@ifc-lite/ifcx`, for whom it was previously unusable for its documented purpose.
--- a/packages/ifcx/src/composition.ts
+++ b/packages/ifcx/src/composition.ts
@@ -200,7 +200,17 @@ export function findRoots(composed: Map<string, ComposedNode>): ComposedNode[] {
 }
 
 /**
- * Get all descendant nodes of a given node.
+ * Get all descendant nodes of a given node — the whole subtree, not just
+ * the direct children. Each node appears once; a cycle terminates.
+ *
+ * The child is marked visited by the recursive call that opens on it, NOT
+ * by the loop before it. Marking it here and then recursing made
+ * `traverse`'s own `visited.has(n.path)` guard fire on entry every single
+ * time, so the walk returned after one level: `a -> b -> c -> d` answered
+ * `[b]` instead of `[b, c, d]`. The two tests over this function could not
+ * see it — one asserts only that the result is duplicate-free (true of a
+ * one-level result) and the other's expectation, `['b']` for a two-node
+ * cycle, is what the truncated walk produces anyway.
  */
 export function getDescendants(node: ComposedNode): ComposedNode[] {
   const descendants: ComposedNode[] = [];
@@ -213,7 +223,6 @@ export function getDescendants(node: ComposedNode): ComposedNode[] {
     for (const child of n.children.values()) {
       if (visited.has(child.path)) continue;
       descendants.push(child);
-      visited.add(child.path);
       traverse(child);
     }
   }
--- a/packages/ifcx/src/traversal.test.ts
+++ b/packages/ifcx/src/traversal.test.ts
@@ -1,0 +1,314 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Composed-graph traversal and root/descendant queries.
+ *
+ * `traversal.ts` backs the geometry, point-cloud and entity extractors, and
+ * `findRoots` / `getDescendants` are published from the package index, yet
+ * neither module had a test file. These tests pin the parts that decide what
+ * gets extracted: frame depth, incoming-edge de-duplication, the reachable
+ * attribute index's cycle guard, and descendant de-duplication.
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import type { ComposedNode } from './types.js';
+import { findRoots, getDescendants } from './composition.js';
+import {
+  buildReachableAttributeIndex,
+  collectIncomingEdgeNames,
+  findTraversalRoots,
+  getFrameLineage,
+  getNodeLineage,
+  walkComposedFrames,
+  type TraversalFrame,
+} from './traversal.js';
+
+/** Build a composed graph from an edge list: `parent --edgeName--> child`. */
+function graph(
+  edges: Array<[string, string, string]>,
+  attributes: Record<string, Record<string, unknown>> = {},
+  extraPaths: string[] = []
+): Map<string, ComposedNode> {
+  const nodes = new Map<string, ComposedNode>();
+  const node = (path: string): ComposedNode => {
+    let n = nodes.get(path);
+    if (!n) {
+      n = { path, attributes: new Map(), children: new Map() };
+      nodes.set(path, n);
+    }
+    return n;
+  };
+
+  for (const p of extraPaths) node(p);
+  for (const [parent, edgeName, child] of edges) {
+    node(parent).children.set(edgeName, node(child));
+  }
+  for (const [path, attrs] of Object.entries(attributes)) {
+    const n = node(path);
+    for (const [k, v] of Object.entries(attrs)) n.attributes.set(k, v);
+  }
+  return nodes;
+}
+
+describe('findRoots / findTraversalRoots', () => {
+  it('returns only the nodes never referenced as a child', () => {
+    const composed = graph([
+      ['root', 'a', 'child'],
+      ['child', 'b', 'grandchild'],
+    ]);
+
+    assert.deepStrictEqual(
+      findRoots(composed).map((n) => n.path),
+      ['root']
+    );
+    assert.deepStrictEqual(
+      findTraversalRoots(composed).map((n) => n.path),
+      ['root']
+    );
+  });
+
+  it('returns every node when nothing has children', () => {
+    const composed = graph([], {}, ['a', 'b']);
+    assert.deepStrictEqual(
+      findRoots(composed).map((n) => n.path).sort(),
+      ['a', 'b']
+    );
+  });
+
+  it('returns no roots for a graph that is entirely a cycle', () => {
+    const composed = graph([
+      ['a', 'to-b', 'b'],
+      ['b', 'to-a', 'a'],
+    ]);
+    assert.deepStrictEqual(findRoots(composed), []);
+    assert.deepStrictEqual(findTraversalRoots(composed), []);
+  });
+
+  it('returns an empty list for an empty graph', () => {
+    assert.deepStrictEqual(findRoots(new Map()), []);
+  });
+});
+
+describe('getDescendants', () => {
+  it('never repeats a node', () => {
+    const composed = graph([
+      ['root', 'left', 'l'],
+      ['root', 'right', 'r'],
+      ['l', 'down', 'shared'],
+      ['r', 'down', 'shared'],
+    ]);
+
+    const paths = getDescendants(composed.get('root')!).map((n) => n.path);
+    assert.strictEqual(new Set(paths).size, paths.length, 'descendants must be unique');
+  });
+
+  it('terminates on a cycle without re-listing the entry node', () => {
+    const composed = graph([
+      ['a', 'to-b', 'b'],
+      ['b', 'to-a', 'a'],
+    ]);
+
+    // Without the visited guard the walk re-emits `a` as its own descendant.
+    assert.deepStrictEqual(
+      getDescendants(composed.get('a')!).map((n) => n.path),
+      ['b']
+    );
+  });
+
+  it('returns an empty list for a leaf', () => {
+    const composed = graph([['root', 'x', 'leaf']]);
+    assert.deepStrictEqual(getDescendants(composed.get('leaf')!), []);
+  });
+});
+
+describe('collectIncomingEdgeNames', () => {
+  it('records the edge name once when two parents use the same name', () => {
+    const composed = graph([
+      ['storey-1', 'contains', 'wall'],
+      ['storey-2', 'contains', 'wall'],
+    ]);
+
+    assert.deepStrictEqual(collectIncomingEdgeNames(composed).get('wall'), ['contains']);
+  });
+
+  it('records each distinct edge name in first-seen order', () => {
+    const composed = graph([
+      ['a', 'body', 'geom'],
+      ['b', 'axis', 'geom'],
+      ['c', 'body', 'geom'],
+    ]);
+
+    assert.deepStrictEqual(collectIncomingEdgeNames(composed).get('geom'), ['body', 'axis']);
+  });
+
+  it('has no entry for a node nothing points at', () => {
+    const composed = graph([['a', 'x', 'b']]);
+    const incoming = collectIncomingEdgeNames(composed);
+    assert.strictEqual(incoming.get('a'), undefined);
+    assert.deepStrictEqual(incoming.get('b'), ['x']);
+  });
+});
+
+describe('buildReachableAttributeIndex', () => {
+  it('is true for the node holding the attribute and for its ancestors only', () => {
+    const composed = graph(
+      [
+        ['root', 'a', 'mid'],
+        ['mid', 'b', 'leaf'],
+        ['root', 'c', 'other'],
+      ],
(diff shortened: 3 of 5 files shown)


BIMvoice:
Confirmed the production defect independently, and it is real.

`packages/ifcx/src/composition.ts:205-223` — the recursion in `getDescendants` is dead:

```ts
for (const child of n.children.values()) {
  if (visited.has(child.path)) continue;
  descendants.push(child);
  visited.add(child.path);      // <- child marked visited HERE
  traverse(child);              // <- ...so this returns at its own entry check (line 210)
}
```

Executed against a chain `a → b → c → d`:

```
getDescendants(a) = [ 'b' ]
expected          = [ 'b', 'c', 'd' ]
```

It returns direct children only. The name, the return type and the recursive shape all say otherwise.

**Why nothing caught it:** `grep` finds exactly two occurrences in the repo — the definition at `composition.ts:205` and the barrel re-export at `index.ts:41`. No in-repo caller, no test. It is published API of `@ifc-lite/ifcx`, so the only people who can hit it are external SDK consumers, and they would get a silently truncated tree rather than an error.

That is the second published-but-unexercised symbol found today to contain a real bug rather than merely lacking coverage — the pattern being that "no internal caller" and "no test" are the same blind spot, and neither the compiler nor CI distinguishes dead-but-exported from live.

**The fix is one line** (move `visited.add(child.path)` into `traverse`, or drop it and rely on the entry check). I have deliberately not made it: it changes published behaviour, and the tests added here assert only properties that hold either way, so they do not enshrine the bug in either direction. Say the word and I will open it as a separate PR.

**The strongest test finding here is the three-state flag, and it is now a pattern rather than an incident.** `prettyPrint`, `includeProperties` and `applyMutations` are each `options.X !== false`, and none had any test. Flipping `applyMutations` makes `exportToIfcx` write the **pre-edit** property values — the user's edits never reach the saved file, silently, with a green suite. All three are now pinned in all three states.

That is the same shape as `seedFromIfcx` on #2160, where `reset: true` and `reset: false` were both pinned explicitly while the **absent** default — the one both production call sites actually use — went untested. Two independent instances in one day says the rule needs stating: for `options.X !== false`, "both directions covered" means three states, not two, and the untested one is usually the one production passes.

Two more things worth recording:

- **The agent hit the unbuilt-deps trap and caught it.** Its first downstream re-check reported `Test Files 8 failed` / 20 / 19 / 18 across four packages. That was missing workspace `dist`, not its mutations; after a full `turbo build` the same suites are green. Every number in this PR is against the verified-clean baseline. That trap has now produced seven false results today.
- **`packages/mutations/test/mutations.test.ts` is `describe.skipIf(!testFile)`** — 10 tests skip silently without LFS fixtures. Deliberate, but it means that package's integration coverage is invisible on a fresh checkout, and a contributor without fixtures sees green from a suite that ran a fraction of itself. Everything added here is fixture-free.

One survivor was left unkilled and proven equivalent rather than argued away: `deleteChangeSet`'s null-out of the private `activeChangeSetId` is unobservable, because `getActiveChangeSet()` map-looks-up the id and every writer is unconditional — only an id collision could distinguish it.

BIMvoice:
## Self-audit round 3 — a real production defect behind a non-discriminating test

Pushed as 29df62c…e06ae8c. This is the most serious find of the round: not just vacuity, but **a published function that did not do its job**, sitting behind two tests of ours that could not see it.

### `getDescendants` walked exactly one level

```js
for (const child of n.children.values()) {
  if (visited.has(child.path)) continue;
  descendants.push(child);
  visited.add(child.path);   // <- marks the child…
  traverse(child);           // <- …so traverse() trips its own entry guard and returns
}
```

`traverse` opens with `if (visited.has(n.path)) return`. Because the loop marked the child first, **every** recursive call returned immediately. Measured on the real function:

- `a -> b -> c -> d` answered `["b"]`, not `["b","c","d"]`
- diamond `root -> {l, r} -> shared` answered `["l","r"]` — the shared grandchild dropped entirely

`getDescendants` is exported from the package index. It has no in-repo consumer, so the blast radius is external users of `@ifc-lite/ifcx`, for whom it was unusable for its documented purpose.

### Why our tests could not see it

| test | claimed property | discriminates | evidence |
|---|---|---|---|
| `never repeats a node` | no duplicates in the subtree walk | **no** | asserts only `new Set(paths).size === paths.length`, which a one-level result satisfies trivially |
| `terminates on a cycle without re-listing the entry node` | cycle termination | **no** | expects `['b']` for `a <-> b` — exactly what the truncated walk produces anyway |

Neither asserts the reached *depth*, which is the whole content of the word "descendants". The mutation that exposed it was innocuous-looking — removing the push-time `visited.has(child.path)` dedup — and it survived because the function was already truncating.

### Fixed

The child is now marked by the recursive call that opens on it, where the cycle guard already expects the marking. Added a chain case asserting depth head-on and extended the diamond case to assert the shared grandchild is present, once. Verified: re-introducing the early `visited.add`, dropping the recursion, and dropping the push-time dedup are all killed now; **none of them was before**. Patch changeset added.

### What held — 19 mutations killed

All three writer tri-state options survived their own scrutiny: `prettyPrint`, `includeProperties` and `applyMutations` each killed when `!== false` is weakened to `=== true`, i.e. the **omitted-default** state that production actually passes is genuinely pinned. This is the shape that was a real defect two rounds ago, so it is good to see it closed. Also killed: `fileSize` in UTF-16 code units, the prop-schema import, the `findRoots` not-a-child filter, the edge-name dedup, both reachable-attribute-index mutations, both `walkComposedFrames` mutations (depth and the per-path visited set), and all eight change-set manager mutations (active-pointer bookkeeping, merge timestamp ordering, fresh id and `applied: false` on import, the model-qualified entity key, cross-set totals, `markApplied` scoping).

BIMvoice:
**Correcting my own disposition on this PR: I said `getDescendants` was "reported, not fixed" and it is now fixed here. The reason is that leaving it was not actually available.**

A follow-up audit found that the two tests over that function do not merely fail to catch the bug — **they encode it**:

- *"never repeats a node"* asserts only that the result is duplicate-free, which is trivially true of a one-level result.
- *"terminates on a cycle"* expects `['b']` for a two-node cycle — which is exactly what the truncated walk produces.

So the tests could not be strengthened without also fixing the function; strengthening them alone would have meant asserting the broken output. That is a different situation from the one I described earlier, when I thought the tests were merely silent on it. Reported-not-fixed was the right call for a defect the tests ignore; it is the wrong call for a defect the tests enshrine.

Verified after the fix, on all three shapes the walk has to handle:

```
chain a->b->c->d  ->  [ b, c, d ]        (was [ b ])
diamond           ->  [ l, shared, r ]   (shared appears once)
cycle x<->y       ->  [ y ]              (terminates)
```

The cause was one line: the child was marked visited by the loop *before* recursing, so `traverse`'s own entry guard fired immediately on every child. It was a one-level walk wearing a recursion's clothes. A patch changeset is included, since this is a published export of `@ifc-lite/ifcx` with no in-repo consumer — external users only.

## The wider result this came from

The audit ran ~164 mutations across six of our own sweep PRs and found **12 genuine survivors**, this being the only production defect among them. Two patterns are worth naming because both are subtler than the ones we have been catching:

**An assertion aimed at the wrong field.** In `packages/extensions`, `migrateManifest` emits `invalid_manifest_version` at path `manifestVersion` from *three* independent branches. So deleting the `rawVersion < 1` guard still errors (it falls through to "no migration available from v0"), and deleting `!Number.isInteger` still errors (`1.5 > 1` hits "newer than supported"). Asserting the code and path could never distinguish them; the fix was to assert each guard's own **message**. Not an equivalent mutant — an assertion pointed at a field three rules share.

**A fixture whose uniformity makes a merge indistinguishable from an overwrite.** The presence patch-coalescing test enqueued 20 patches that all share one key, so `{...patch}` versus a real merge produce identical output.

**And a meaningful amount held.** The sandbox capability boundary killed 10 of 10, parquet geometry 10 of 10, `change-set` 8 of 8, `attribute-helpers` 6 of 6 — and specifically, the identity-fixture gap that round two found in `lists/engine` is genuinely closed, as is the three-state-flag gap in the ifcx writer. Those follow-ups did what they claimed.

Issue #2110, opened 2026-08-04
Sandbox.eval does not serialize: overlapping evals on one sandbox share the log buffer and capture budget


BIMvoice:
Split out of #2103 at @louistrue's suggestion — *"Worth its own issue. I would not widen this PR for it, and I would not want the 'reset per eval' fix held up by a concurrency question that predates it."* Filing so the reproducer and the analysis survive that PR merging.

## What happens

`Sandbox.eval` (`packages/sandbox/src/sandbox.ts:148`) does not serialize. The bridge's `logs` array, `totalBytes` and `truncated` are per-**sandbox**, created once in `buildBridge` (`bridge.ts:88-92`) and captured by the closure — not per-eval. Two overlapping `eval()` calls on one `Sandbox` therefore share all three, and the second call's `this.resetLogs()` at `sandbox.ts:155` clears state the first call is still filling.

Result: the first eval's `ScriptResult.logs` can come back short, empty, or carrying entries the second eval produced.

## This predates #2103 and is not introduced by it

Worth being precise, because the shapes look similar. `main` before #2103 already cleared the shared buffer per run (`this.logs.length = 0`), so concurrent evals already clobbered each other's **logs**. What #2103 adds is resetting the two counters alongside the buffer — the same sharing, not new sharing. The defect it fixes (a budget that latched for the life of the sandbox) is real and separate.

**One direction genuinely differs, and it is narrower rather than wider.** Previously `truncated` latched for the sandbox's life, so a concurrent second eval could not un-truncate the first. Now it can. That is a new window, but it sits inside an already-broken scenario: if two evals overlap, the first eval's logs were already being destroyed by the second's buffer clear. Trading "your logs are wrong and flagged truncated" for "your logs are wrong" is not a regression worth holding a fix for.

## Why a fix is bigger than it looks

The obvious remedy — serialize `eval` behind a promise chain — is a **public-API behaviour change on a published package**. Today two overlapping `eval()` calls both start immediately; after serialization the second would queue behind the first, so a caller relying on concurrent execution sees different timing, and a long-running first script delays the second by its full duration (up to the 30s timeout).

Alternatives worth weighing before picking one:

1. **Serialize** — simplest and safest, but changes observable timing for existing callers.
[...]

PR #2166, opened 2026-08-05
test(sandbox,cli): close 20 mutation-verified gaps in the sandbox capability boundary and CLI surface

BIMvoice:
Second, deeper mutation pass over `packages/sandbox` and `packages/cli`. Both had a shallow first pass; this one targeted the modules with **no test file at all**, which is where the survivors were.

**26 mutations + 3 deliberate controls; 5 killed, 21 survived — 19% kill ratio.** That number is **TARGETED**, not a uniform sample: mutations were aimed at the untested modules on purpose, so it measures where the suites do not reach, not overall suite quality. Everything aimed at the *tested* paths died — console capture budgets, the CPU timeout, disposal ordering, rejected promises, `diff-scope` classification.

**Test-only. No production code changed.**

| | files | tests | after |
|---|---|---|---|
| `packages/sandbox` | 12 | 71 | **17 / 108** |
| `packages/cli` | 25 | 271 passing (9 skipped) | **27 / 305 (9 skipped)** |

The 9 skips are fixture-gated, not unconditional: `clash.test.ts` needs a built CLI entry plus the wasm runtime, `gym.test.ts` needs the AB22.ifc fixture.

## Method

Each mutation was a verified single-occurrence substring replacement; the mutated region was **re-read from disk** and asserted to contain the new text and not the old before any result was believed; the source was restored from a backup taken before that file's first mutation (never `git checkout --`).

Every survivor was re-run against the downstream suites that consume `@ifc-lite/sandbox` — `packages/cli`'s three `schema` tests and `apps/viewer`'s `system-prompt` / `script-preflight` tests, both resolving through the built `dist`, with the package rebuilt for each mutation. **No false survivors.**

**Controls** (expected to die, all did): shrinking `MAX_LOG_ENTRIES` 1000 → 10; dropping the `interrupted` flag from the CPU interrupt handler; an off-by-one in the CLI log-level gate.

## Gaps closed — `packages/sandbox`

The capability boundary was almost entirely unpinned.

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| `if (!permissions[schema.permission]) continue` → `if (false) continue` | every namespace built regardless of permissions — a read-only script reaches `bim.mutate.*` and `bim.store.*` | `bridge-permissions.test.ts` |
| `DEFAULT_PERMISSIONS.mutate/store` `false` → `true` | every caller that passes no permissions silently gets write access | `bridge-permissions.test.ts` |
| `raw.map(r => r.ref ?? r)` → `raw.map(r => r)` | `bim.query.byType(...)` results no longer usable as `bim.viewer.colorize(...)` arguments | `bridge-marshal.test.ts` |
| `typeof value === 'string' ? … : vm.null` → `String(value)` | a broken SDK return reads as real CSV text (`"[object Object]"`) instead of `null` | `bridge-marshal.test.ts` |
| `stack.delete(obj)` on exit → no-op | `{ a: shared, b: shared }` returns `b: null` — an acyclic graph reported as cyclic | `bridge-marshal.test.ts` |
| `nextHandleBySession.set(sessionId, handle + 1)` → `handle` | a second `bim.create.project()` silently replaces the first model, no error anywhere | `creator-registry.test.ts` |
| `removeSession` stops dropping creators | every disposed sandbox leaks its `IfcCreator` models for the process lifetime | `creator-registry.test.ts` |
| `toRef` shape validation removed | a malformed script-supplied ref reaches the SDK with no usable `expressId` | `bridge-helpers.test.ts` |
| `withAliases` drops `GlobalId` | `e.GlobalId` — the whole point of the helper — is `undefined` | `bridge-helpers.test.ts` |
| `runtime.setMaxStackSize(...)` never called | a configured `maxStackBytes` is silently ignored | `stack-limit.test.ts` |

The stack test brackets the ceiling in **both** directions: the same 400-deep recursion completes under the default 512KB and overflows under an 8KB one, so it cannot pass for a sandbox that simply refuses all recursion.

## Gaps closed — `packages/cli`

| Surviving mutation | What a user would see | Killing test |
|---|---|---|
| `--log-level` value check drops `!value.startsWith('-')` | `--log-level --json` swallows `--json`; a script gets human output | `logger.test.ts` |
| `--sort` direction inverted | `--desc` returns **ascending** | `query-aggregation.test.ts` |
| attribute-key camelCase mapping dropped | `--sort globalid` / `--sort objecttype` silently a no-op (input order) | `query-aggregation.test.ts` |
| dotted sort swaps pset/prop | `--sort Pset_WallCommon.Rating` finds nothing, sorts nothing | `query-aggregation.test.ts` |
| `Number(q.value) \|\| 0` → `Number(q.value)` | NaN poisons `--sum` totals and flattens `--sort` | `query-aggregation.test.ts` |
| `writeOutput` trailing newline dropped | stdout payloads run into the shell prompt; byte output must stay verbatim | `output.test.ts` |
| `printJson` Map conversion dropped | `--json` reports `{}` where the data was | `output.test.ts` |
| `expected.fingerprint !== info.fingerprint` → `if (false)` | **a bundle signed by ANY key passes `ext verify --key trusted.iflk` with exit 0** | `ext-signing.test.ts` |
| unsigned-with-`--key` `exit(2)` → `exit(0)` | a CI gate accepts an unsigned bundle | `ext-signing.test.ts` |
| `getArg` accepts a following flag as its value | `--key --json` verifies against no key at all | `ext-signing.test.ts` |

`query-aggregation.ts` and `ext-signing.ts` both had **no test file**. Sort fixtures are four elements ordered so every adjacent pair has to compare correctly, rather than two elements one lucky comparison would carry.

## Reported, not fixed

**1. `runtime.setMemoryLimit(...)` is an observable no-op.** Removing the call changes nothing — and neither does the limit. A script allocated **50,000,000 characters under a declared 1MB `memoryBytes` ceiling** and succeeded. Reproduced against `quickjs-emscripten` directly, no ifc-lite code in the path, with the limit set both before and after `newContext()`: the call site is correct, this is upstream behaviour. The consequence is that the published **64MB `memoryBytes` default does not currently bound a script's memory use**; only the CPU timeout and the stack ceiling actually stop a runaway script. Classified EQUIVALENT for the shapes probed (strings and arrays up to 50MB against a 1MB declared limit) — **not** proven equivalent for all shapes. Flagging for a maintainer decision rather than changing sandbox behaviour.

**2. #1922.** Runaway recursion under the default stack ceiling does stop, but the overflow leaves QuickJS holding leaked refcounts and the following `runtime.dispose()` throws out of teardown. Upstream, no in-repo lever. A test would assert the bug's blast radius rather than the ceiling, so `stack-limit.test.ts` documents the omission instead.

**Also pinned rather than changed:** emptying a session drops its handle counter with its map, so the next handle restarts at 1. Harmless today — the only way to empty a session is to remove every creator in it, which invalidates every handle a script could hold — but the coupling is now explicit so a change to it is deliberate.

Nothing here touches **#2110**: every new sandbox test runs one eval at a time on its own sandbox instance.

## Typecheck

Test files were typechecked against each package's own compiler options with the `src`-only `include` lifted (the root `pnpm typecheck` excludes `**/*.test.ts`); `--listFiles` confirmed all 12 sandbox and 25 cli test files entered the program. Pre-existing errors left alone: **2** in `sandbox/src/transpile-wasm-url.test.ts` (the package has no `@types/node`, which only its tests need) and **16** across 6 existing cli test files (all the `process.exit` mock signature). The new files add none.

### DIFF
--- a/packages/cli/src/commands/ext-signing.test.ts
+++ b/packages/cli/src/commands/ext-signing.test.ts
@@ -46,6 +46,12 @@ class ExitCalled extends Error {
  * `verifyBundle`, which then emits a second document. That second write cannot
  * happen in production — it is an artifact of making the exit observable — so
  * the assertions read the document the command actually decided to print.
+ *
+ * The same artifact hides the exit *code*, which is why `verify()` below reads
+ * the FIRST recorded exit rather than the one that escapes: the catch-all also
+ * calls `process.exit(2)`, so reading the escaping code made "exit 2" hold for
+ * a mismatch branch that exited 0. Verified: changing the mismatch branch to
+ * `process.exit(0)` left this file green until the change below.
  */
 function firstJsonDoc(out: string): unknown {
   const [first] = out.split(/\n(?=\{)/);
@@ -56,11 +62,14 @@ describe('ifc-lite ext verify', () => {
   let dir: string;
   let stdout: string;
   let stderr: string;
+  /** Every process.exit() code, in call order. Index 0 is the deciding branch. */
+  const exitCodes: number[] = [];
   const dirs: string[] = [];
 
   beforeEach(async () => {
     stdout = '';
     stderr = '';
+    exitCodes.length = 0;
     dir = await mkdtemp(join(tmpdir(), 'ifc-lite-ext-sign-'));
     dirs.push(dir);
     vi.spyOn(process.stdout, 'write').mockImplementation(((chunk: string) => {
@@ -72,6 +81,7 @@ describe('ifc-lite ext verify', () => {
       return true;
     }) as typeof process.stderr.write);
     vi.spyOn(process, 'exit').mockImplementation((((code?: string | number | null) => {
+      exitCodes.push(typeof code === 'number' ? code : 0);
       throw new ExitCalled(typeof code === 'number' ? code : undefined);
     }) as unknown) as typeof process.exit);
   });
@@ -99,13 +109,15 @@ describe('ifc-lite ext verify', () => {
   async function verify(args: string[]): Promise<{ code: number; stdout: string; stderr: string }> {
     stdout = '';
     stderr = '';
+    exitCodes.length = 0;
     try {
       await extVerifyCommand(args);
-      return { code: 0, stdout, stderr };
     } catch (err) {
-      if (err instanceof ExitCalled) return { code: err.code ?? 0, stdout, stderr };
-      throw err;
+      if (!(err instanceof ExitCalled)) throw err;
     }
+    // exitCodes[0], not the escaping ExitCalled: the mismatch branch's exit is
+    // re-raised by the catch-all around verifyBundle, which masks its code.
+    return { code: exitCodes[0] ?? 0, stdout, stderr };
   }
 
   it('accepts a bundle signed by the key the caller expects', async () => {
--- a/packages/cli/src/commands/ext-signing.test.ts
+++ b/packages/cli/src/commands/ext-signing.test.ts
@@ -1,0 +1,206 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `ifc-lite ext verify` is a gate: a CI pipeline runs it and reads the exit
+ * code. `ext-signing.ts` had no test file, and three mutations left the
+ * package's 271 tests green — all of them turning a refusal into an
+ * acceptance:
+ *
+ *   - the signer check (`expected.fingerprint !== info.fingerprint`) deleted,
+ *     so a bundle signed by ANY key passed `--key trusted.iflk` with exit 0;
+ *   - the unsigned-but-`--key`-given refusal downgraded from exit 2 to exit 0;
+ *   - `getArg` accepting a following flag as its value, so
+ *     `--key --json` silently verified against no key at all.
+ *
+ * The suite drives the real commands end to end — keygen, pack, sign, verify —
+ * because the exit code and the stdout document ARE the contract, and a unit
+ * test of the comparison would not have caught the third one.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mkdtemp, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { extCommand } from './ext.js';
+import {
+  extKeygenCommand,
+  extPackCommand,
+  extSignCommand,
+  extVerifyCommand,
+} from './ext-signing.js';
+
+/** Thrown in place of a real process.exit so the code is observable. */
+class ExitCalled extends Error {
+  constructor(public readonly code: number | undefined) {
+    super(`process.exit(${code})`);
+  }
+}
+
+/**
+ * The FIRST JSON document on stdout.
+ *
+ * A real `process.exit()` ends the process; the stub throws instead, and on the
+ * fingerprint-mismatch path that throw is caught by the `try` around
+ * `verifyBundle`, which then emits a second document. That second write cannot
+ * happen in production — it is an artifact of making the exit observable — so
+ * the assertions read the document the command actually decided to print.
+ */
+function firstJsonDoc(out: string): unknown {
+  const [first] = out.split(/\n(?=\{)/);
+  return JSON.parse(first);
+}
+
+describe('ifc-lite ext verify', () => {
+  let dir: string;
+  let stdout: string;
+  let stderr: string;
+  const dirs: string[] = [];
+
+  beforeEach(async () => {
+    stdout = '';
+    stderr = '';
+    dir = await mkdtemp(join(tmpdir(), 'ifc-lite-ext-sign-'));
+    dirs.push(dir);
+    vi.spyOn(process.stdout, 'write').mockImplementation(((chunk: string) => {
+      stdout += String(chunk);
+      return true;
+    }) as typeof process.stdout.write);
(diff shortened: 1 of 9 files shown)


BIMvoice:
I verified the memory-limit finding independently, and it holds. Posting the probe in full because two of my earlier attempts at it were themselves wrong, and the corrections are the reason to trust the third.

**The contract.** `packages/sandbox/src/types.ts:30` — *"Maximum heap memory in bytes (default: 64MB)"*; `types.ts:78` sets `memoryBytes: 64 * 1024 * 1024`; `sandbox.ts:72` calls `runtime.setMemoryLimit(...)`.

**The probe.** A sandbox created with `limits: { memoryBytes: 1024 * 1024 }` — a 1 MiB ceiling — then a script that doubles a string and reports its own length from inside the VM:

```
1e6 -> ok  2ms  logs=[["len=1048576"]]
1e7 -> ok  0ms  logs=[["len=16777216"]]
5e7 -> ok  0ms  logs=[["len=67108864"]]
```

**67,108,864 characters allocated under a declared 1 MiB ceiling**, no error, no interruption. The length is logged from within the sandbox, so the allocation demonstrably happened rather than being optimised away.

**Controls proving the harness is sound**, run in the same session:

```
deliberate throw           -> threw: boom          3ms
infinite loop (timeout)    -> threw: interrupted   5001ms
```

So errors propagate and the CPU timeout enforces at exactly its configured 5000ms. The memory ceiling is the one limit of the three that does not bind.

**Two corrections against myself, since they nearly produced a false report in each direction:**

1. My first probe read `r.result` and saw `undefined` — but `undefined` is what `1+1` returns through this API too, so `result` is not the expression value and the reading meant nothing.
2. My second read `r.error` and saw `(none)` — but this sandbox **throws** rather than returning errors, so `error` was never going to be populated. Both readings would have supported the conclusion I ended up at, for reasons that were wrong. Only the third probe — try/catch, with in-VM logging to confirm the allocation, plus two positive controls — actually establishes it.

**What this means and what it does not.** The agent reproduced the same behaviour against `quickjs-emscripten` directly with no ifc-lite code in the path, limit set both before and after `newContext()`, so the call site is correct and this is upstream rather than ours. But the published contract currently overstates what it delivers: of the three documented limits, `timeoutMs` and `maxStackBytes` bind and `memoryBytes` does not. A consumer reading `types.ts:30` would reasonably assume a hostile or buggy script is bounded at 64MB; it is bounded only by wall-clock time.

Nothing changed here — this wants your decision, and there are at least three defensible answers: document the limitation on `SandboxLimits.memoryBytes`, pin the current behaviour so a future upstream fix is noticed, or carry a patch. I have not picked one.

**The two security-relevant survivors in this PR are worth separating from the rest**, because in both cases the production code is correct and only the coverage was missing:

- The sandbox permission gate could be deleted outright — `if (false) continue`, every namespace constructed regardless of declared permissions — with all 71 tests green. Same for flipping `DEFAULT_PERMISSIONS.mutate`/`store` to `true`.
- `ifc-lite ext verify --key trusted.iflk` returned **exit 0 for a bundle signed by any key** once the fingerprint comparison was removed. All three `ext-signing` survivors converted a refusal into an acceptance, and CI pipelines read that exit code as a gate.

Also worth noting the `--log-level` finding as another instance of the three-state pattern: valid value and invalid word were both pinned, the third state — where the next token is itself a flag — was not, so `--log-level --json` silently consumed `--json`.

BIMvoice:
## Self-audit round 3 — the "exit 2" in `ext verify` was not actually pinned

Pushed as 4c0ff96…1acffd4. This PR came through the audit almost intact: **19 of 20 probative mutations were killed**. One survivor, and it is on the security-relevant test.

|---|---|---|---|---|
| `ext-signing.test.ts` — REFUSES a bundle signed by a different key, **with exit 2** | the mismatch branch exits 2 | **no** | mismatch branch `process.exit(2)` → `exit(0)` | SURVIVED |
| same | | **no** | mismatch branch `process.exit(2)` → `exit(1)` | SURVIVED |

**Mechanism.** The fingerprint-mismatch `process.exit(2)` sits *inside* the `try` around `verifyBundle`. The test stub throws instead of exiting, that throw is caught by the command's own catch-all, and the catch-all calls `process.exit(2)` itself. `verify()` read the code off the escaping `ExitCalled`, so it always saw the catch-all's 2 — never the branch's. Any exit code in that branch reads as 2.

The file already documented this artifact for **stdout** (that is why `firstJsonDoc` exists); it just did not carry the observation across to the exit code. `verify()` now reads the first recorded exit instead. Both `exit(0)` and `exit(1)` mutants are killed, and the three mutants the PR was written against stay killed.

Worth noting for the contrast: the **unsigned-with-`--key`** refusal was never affected — that `process.exit(2)` is outside the `try`, and downgrading it to `exit(0)` was killed both before and after. No production behaviour changes; a real `process.exit()` does not throw, so the double-exit is purely a test artifact.

**What held.** The sandbox capability boundary is solid — 10/10 killed, including dropping the permission gate entirely, weakening it to `!== false`, flipping either read-only default (`mutate`, `store`) to true, deleting `setMaxStackSize`, making it ignore the caller's value, halving the published 512KB default, and all three creator-registry session-scoping mutations. `marshalValue`'s cycle guard held both ways (making the guard permanent, and removing it). The CLI half held: 8/8 on `query-aggregation` (the `\|\| 0` unparseable case, null-vs-0 for absent, past-first-qset search, the `slice()` that keeps the caller's array, dotted pset/prop order, the qset fallback, the `globalid` → `globalId` remap, descending), `--log-level` not swallowing a following flag, and the other three `ext verify` mutants.

**Discarded as non-probative:** the `isPlainObject` entry guard in `validateSlot` (a non-object item already draws its own `type_mismatch`; removing the guard only adds a spurious extra error) and `MARSHAL_MAX_DEPTH` 64 → 2 (the depth cap is pre-existing and this PR does not claim it — flagging only so it is on the record as uncovered).

PR #2167, opened 2026-08-05
test(collab,lists): close 31 mutation-verified coverage gaps (second sweep)

BIMvoice:
Second, deeper mutation sweep of `packages/collab` and `packages/lists`. Both had shallow first passes early in this campaign.

**49 mutations** against load-bearing logic, **25 survived** the existing suites (49% kill ratio — TARGETED: mutations were aimed at modules with no test file, at thin test files, and at exported symbols with no caller). 21 survivors are closed here; 4 are classified below. Test-only — no production behaviour changed.

Two deliberate control mutations were run first and both died (`values[i] = null` in the list property column; `out[structuredAttributeKey(...)] = null` in the pset flattener), confirming the harness and the build.

## Baseline → final

| package | before | after |
| --- | --- | --- |
| `@ifc-lite/collab` | 41 files / 173 passed / 2 skipped | 44 files / 201 passed / 2 skipped |
| `@ifc-lite/lists` | 6 files / 119 passed | 7 files / 143 passed |

The 2 collab skips are fixture-gated in `test/round-trip.test.ts` (`skip: !FIXTURES_AVAILABLE`), so IFCX round-trip coverage is invisible on a fresh checkout until `pnpm fixtures` runs. Not changed here, but worth knowing when reading collab's numbers.

## `packages/lists`

| # | mutation | was | now |
| --- | --- | --- | --- |
| M2 | `compareCellValues`: delete the `typeof a === 'number'` branch | SURVIVED | KILLED |
| M3 | `if (a === null) return -1` → `return 1` | SURVIVED | KILLED |
| M4 | `if (b === null) return 1` → `return -1` | SURVIVED | KILLED |
| M5 | multi-valued `equals` → `contains` semantics | SURVIVED | KILLED |
| M6 | multi-valued `notEquals`: `every` → `some` | SURVIVED | KILLED |
| M8 | `uniqueJoin`: drop the `new Set()` | SURVIVED | KILLED |
| M11 | drop the em-dash → `null` mapping | SURVIVED | KILLED |
| M12 | boolean cell → `String(value)` instead of the display mapping | SURVIVED | KILLED |
| M13 | disable the `[IFCTYPE, number]` numeric unwrap | SURVIVED | KILLED |
| M14 | numeric property → its display string | SURVIVED | KILLED |
| M15 | `SAMPLE_SIZE` 50 → 1 | SURVIVED | KILLED |
| M16 | remove the `Math.min(ids.length, SAMPLE_SIZE)` cap | SURVIVED | KILLED |
| M17 | stop skipping nameless property/quantity sets | SURVIVED | KILLED |
| M18 | drop `.sort()` on discovered property names | SURVIVED | KILLED |
| M19 | drop `.sort()` on discovered quantity names | SURVIVED | KILLED |
| M21 | CSV header fallback → bare `propertyName` (no set prefix) | SURVIVED | KILLED |
| M22 | `groupingColumnIds`: drop the `id !== ''` filter | SURVIVED | KILLED |
| M23 / M23b | drop `Number.isFinite` from the overall / per-group sum | SURVIVED | KILLED |
| M24 | drop the `localeCompare` tiebreak on equal group counts | SURVIVED | KILLED |
| M1 | sort direction forced to `1` | KILLED | — |
| M7 | multi-valued `exists` → always true | KILLED | — |
| M9 | empty candidate list matches everything | KILLED | — |
| M10 | multi-valued `contains` made case-sensitive | KILLED | — |
| M20 | CSV default delimiter `,` → `;` | KILLED | — |
| M26 | drop the per-model snapshot foreign-id filter | KILLED | — |
| M25 | `chunks.length === 1 ? chunks[0] : chunks.flat()` → always `.flat()` | SURVIVED | **EQUIVALENT** |

All replacement counts asserted `n == 1`, each mutated region re-read before the run was believed.

**M25 is EQUIVALENT** for value semantics: with one chunk, `chunks[0]` and `chunks.flat()` produce the same element sequence. The only difference is aliasing — `chunks[0]` returns the provider's own array by reference — which no caller can observe because the very next step either `filter`s (a copy) or the ids are only read. Re-run against the viewer with `dist` rebuilt (below): still no failure. Not fully proven for a future caller that mutates the returned array in place.

### What a user would have lost

- A schedule sorted on a quantity column ordered `100, 10, 9` as text (M2). Blank cells sorted at the wrong end in one of the two directions (M3/M4).
- A `[IFCREAL, 5.3]` property — the canonical typed-value shape a STEP parser produces — resolved to the string `"5.3"`, so it was skipped by every group sum (`typeof v === 'number'`) and sorted lexicographically (M13). The existing fixture only carries `[IFCBOOLEAN, '.T.']`, whose payload is a string, so the numeric branch was structurally unreachable from it.
- `equals` / `notEquals` on a material or classification filter silently behaving as `contains` / "any layer differs" (M5, M6). A sandwich wall with a Brick and a Concrete layer would have passed a `material notEquals Concrete` filter.
- A material repeated across layers listed twice in one cell (M8).
- A CSV export with two identical `NetVolume` headers when the list pulls the quantity from two different sets (M21).
- Non-finite quantities poisoning a total to `NaN` in the schedule and its export (M23).
- Group order flipping between runs whenever two groups had the same count (M24) — the schedule re-orders for no visible reason after an unrelated filter change.
- The column picker missing every property that is not on the model's first element of a type, or (with the cap gone) walking all 100K entities (M15/M16).

## `packages/collab`

| # | mutation | was | now |
| --- | --- | --- | --- |
| C3 | `parseRoomId`: `indexOf('/')` → `lastIndexOf('/')` | SURVIVED | KILLED |
| C4 | `parseRoomId`: swap the separator-less result's fields | SURVIVED | KILLED |
| C11 | classification gate: drop the `code` check | SURVIVED | KILLED |
| C12 | material gate → `return true` | SURVIVED | KILLED |
| C13 | `recognized = entries.length > 0` → `true` | SURVIVED | KILLED |
| C14 | emit an empty geometry `params` map as `{}` | SURVIVED | KILLED |
| C17 | eviction no longer skips the local peer | SURVIVED | KILLED |
| C18 | eviction age comparison → `>= 0` (evict everyone) | SURVIVED | KILLED |
| C19 | default `staleAfterMs` 10_000 → 1 | SURVIVED | KILLED |
| C20 | default `updateRateHz` 30 → 1 | SURVIVED | KILLED |
| C21 | ignore a caller-supplied `user.color` | SURVIVED | KILLED |
| C23 | let `pendingPatch` overwrite the flush's `lastUpdate` | SURVIVED | KILLED |
| C1 | `roomIdFor` operands swapped | KILLED | — |
| C2 | federation room suffix changed | KILLED | — |
| C5 | v5a key separator `::` → `:` | KILLED | — |
| C6 | single geometryRef no longer emitted as a scalar | KILLED | — |
| C7 / C8 | emit empty classifications / materials | KILLED | — |
| C9 | disable `Qto_*` → quantities routing | KILLED | — |
| C10 | remove the `Pset_*` legacy-flat exclusion | KILLED | — |
| C15 / C16 | never carry the geometry `bbox` / `blobHash` | KILLED | — |
| C22 | `setCursor2d(null)` no longer clears | KILLED | — |

### What a user would have lost

- **`parseRoomId` is exported from the package root and has no caller anywhere in the repo** — definition plus the `export *` barrel, nothing else. Both mutations against it survived. Model ids are caller-supplied and can contain slashes (upload paths, nested folders); splitting on the last separator sends a peer into a room nobody else joined, and the separator-less form would have named the legacy single-model room as a project with no model.
- The `structured-attrs` rejection paths (C11–C13) are the guard that keeps foreign or truncated data under the `ifclite::` namespace FLAT. With the gates loosened, such an attribute is swallowed into a CRDT branch and re-flattened as something else — breaking the module's own documented `flatten(inflate(x)) == x` invariant, on the snapshot → seed path every peer converges through. Only the accept paths were covered (indirectly, via `minimal-layer` / `to-ifcx` / `from-ifcx`); the reject paths were free.
- **Stale-presence eviction was entirely unexercised.** It is what removes a peer whose tab closed without a clean disconnect. Evicting the local peer (C17) drops the current user out of everyone's roster the moment they stop interacting; dropping the comparison (C18) evicts every peer on every sweep; a 1 ms default window (C19) does the same for anyone not moving their cursor.
- A user's or an app's chosen presence colour silently replaced by the id-derived one (C21).
- The default publish rate (C20) is the only rate production runs at — no caller in the repo passes `updateRateHz`. This is the three-state-flag shape: the option was covered at explicit values, never at its default.
- `lastUpdate` being overwritable from a patch (C23) makes this client look stale to every peer and get evicted mid-session.

## Well-covered areas (measured, nothing added)

`snapshot/structured-attrs.ts` has no test file of its own and 313 lines, but its **accept** paths are genuinely well covered from `minimal-layer`, `to-ifcx` and `from-ifcx`: the `::` key separator, the single-vs-list geometryRef wire shape, the empty-collection suppression, the `Qto_*` quantity routing, the `Pset_*` legacy-flat exclusion and both geometry carrier fields all died on the first run (C5–C10, C15, C16). Only the rejection paths were open.

## Untested / thin modules and uncalled API

- **No test file at all:** `src/sync/room.ts`, `src/snapshot/structured-attrs.ts` (covered indirectly, above), `src/awareness/overlay.ts`, `src/providers/{indexeddb,websocket,webrtc}.ts`. `overlay.ts` (178 lines) requires a DOM and collab's vitest runs in the node environment, so it is untestable without adding an environment — flagged, not changed.
- **Exported with only its own definition as a hit:** `parseRoomId` (now covered).
- `packages/lists/src/discovery.ts` had two smoke tests and no assertion on any of its actual behaviour (cap, sort, skip).

- **A fixture constant that makes the operation an identity**, twice in the same file: `engine.test.ts`'s mock provider has no numeric typed-array property (every typed value is `[IFCBOOLEAN, '.T.']`, payload a string) and no repeated material layer. Deleting the numeric unwrap and deleting the whole `new Set()` de-duplication were therefore both unkillable from it.
- **A three-state flag tested in two states:** `PresenceOptions.updateRateHz` and `staleAfterMs` were both exercised at explicit values by other tests but never at their defaults — and the default is what every caller in the repo uses.

## Downstream suites re-checked

`apps/viewer` aliases both packages to source (`vite.config.ts` → `packages/{collab,lists}/src`). With `dist` rebuilt, the 26 viewer suites that import `@ifc-lite/lists` / `@ifc-lite/collab` were run with the M25 survivor live and again clean: 106 tests, 105 pass, 1 fail either way. `packages/collab-server`, `packages/mcp` and `packages/sdk` consume the built `dist` and are unaffected by a test-only change.

## Pre-existing failures and typecheck errors (not fixed)

- `apps/viewer` — `src/lib/layers/collab-publish.test.ts` fails on a clean `upstream/main` checkout, with and without any mutation applied.
- Typecheck was run per package against its own `tsconfig.json` with the `**/*.test.ts` exclusion lifted, verified non-vacuous via `--listFiles` (7 files entered for `lists`, 44 for `collab`). Pre-existing errors: `packages/lists/src/engine.test.ts:130` (mock `PropertySet` missing `globalId`), `packages/collab/test/{blob-store,from-ifcx-reset,round-trip}.test.ts` (3 errors). No new errors from this PR.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for collaboration presence tracking, including stale-peer eviction, update timing, color assignment, and patch throttling.
  * Added tests for room ID parsing and composition, including legacy and slash-containing identifiers.
  * Added validation for structured attributes, geometry references, malformed values, and metadata handling.
  * Expanded list engine coverage for value resolution, sorting, grouping, column discovery, and CSV headers.

### DIFF
--- a/packages/collab/test/presence-eviction.test.ts
+++ b/packages/collab/test/presence-eviction.test.ts
@@ -15,7 +15,7 @@
  */
 
 import { applyAwarenessUpdate, encodeAwarenessUpdate } from 'y-protocols/awareness';
-import { describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import * as Y from 'yjs';
 import { colorForUser } from '../src/awareness/color.js';
 import { createPresence, type Presence } from '../src/awareness/presence.js';
@@ -48,12 +48,23 @@ function publishPeer(
 }
 
 describe('evictStale', () => {
+  // Eviction compares `Date.now() - state.lastUpdate` against the window, and
+  // `evictStale` reads the clock itself — there is no injectable `now`. With a
+  // live clock, a peer built at `Date.now() - staleAfterMs` is exactly at the
+  // boundary only until the next millisecond ticks, and everything in between
+  // (two Y.Docs, an awareness encode and a wire replay) is real work. On a
+  // loaded CI runner that is enough to cross the boundary, which is precisely
+  // how this file went red. Freeze the clock so the ages under test are the
+  // ages asserted; nothing here needs time to pass.
+  const NOW = 1_700_000_000_000;
+  beforeEach(() => { vi.useFakeTimers({ now: NOW }); });
+  afterEach(() => { vi.useRealTimers(); });
+
   it('drops a peer older than the window and keeps a fresh one', () => {
     const doc = new Y.Doc();
     const presence = createPresence(doc, { staleAfterMs: 5_000 });
-    const now = Date.now();
-    const stale = publishPeer(presence, now - 60_000, 'Stale');
-    const fresh = publishPeer(presence, now - 1_000, 'Fresh');
+    const stale = publishPeer(presence, NOW - 60_000, 'Stale');
+    const fresh = publishPeer(presence, NOW - 1_000, 'Fresh');
 
     // `getPeers()` also carries this client's own (empty) entry, so probe
     // the two remote ids rather than the map size.
@@ -73,17 +84,22 @@ describe('evictStale', () => {
   it('keeps a peer exactly at the window boundary (strictly older is evicted)', () => {
     const doc = new Y.Doc();
     const presence = createPresence(doc, { staleAfterMs: 5_000 });
-    const atBoundary = publishPeer(presence, Date.now() - 5_000, 'Edge');
+    // The clock is frozen, so these ages are exact: 5000ms and 5001ms.
+    const atBoundary = publishPeer(presence, NOW - 5_000, 'Edge');
+    const pastBoundary = publishPeer(presence, NOW - 5_001, 'Past');
 
     presence.evictStale();
+
     // `now - lastUpdate > staleAfterMs` — a peer that reported exactly one
-    // window ago is not yet stale. Timer granularity can push this past the
-    // boundary, so assert only that a peer this recent is not evicted for
-    // being *arbitrarily* old.
+    // window ago is not yet stale; one millisecond older is. Asserting both
+    // sides is what pins the comparison: `>=` would evict the first, and
+    // dropping the age test entirely would keep the second.
     const peers = presence.getPeers();
-    expect(Object.prototype.hasOwnProperty.call(peers, String(atBoundary.clientId))).toBe(true);
+    expect(peers[atBoundary.clientId]).toBeDefined();
+    expect(peers[pastBoundary.clientId]).toBeUndefined();
 
     atBoundary.dispose();
+    pastBoundary.dispose();
     presence.dispose();
   });
 
@@ -97,7 +113,7 @@ describe('evictStale', () => {
       user: { id: 'me', name: 'Me' },
       selection: [],
       status: 'active',
-      lastUpdate: Date.now() - 10 * 60 * 1000,
+      lastUpdate: NOW - 10 * 60 * 1000,
     });
 
     presence.evictStale();
@@ -113,9 +129,8 @@ describe('evictStale', () => {
     // quiet is gone.
     const doc = new Y.Doc();
     const presence = createPresence(doc);
-    const now = Date.now();
-    const quiet = publishPeer(presence, now - 9_000, 'Quiet');
-    const gone = publishPeer(presence, now - 11_000, 'Gone');
+    const quiet = publishPeer(presence, NOW - 9_000, 'Quiet');
+    const gone = publishPeer(presence, NOW - 11_000, 'Gone');
 
     presence.evictStale();
 
--- a/packages/collab/test/presence-eviction.test.ts
+++ b/packages/collab/test/presence-eviction.test.ts
@@ -1,0 +1,204 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Stale-presence eviction and patch bookkeeping (`awareness/presence.ts`,
+ * spec §5.4 / §7).
+ *
+ * Eviction is what removes a peer whose tab was closed without a clean
+ * disconnect. Nothing exercised it: mutations that evicted the LOCAL peer,
+ * that dropped the age comparison entirely, and that changed the default
+ * window from 10s to 1ms all survived the suite. Its two neighbours —
+ * the caller-supplied user colour and the authoritative `lastUpdate` —
+ * were equally free.
+ */
+
+import { applyAwarenessUpdate, encodeAwarenessUpdate } from 'y-protocols/awareness';
+import { describe, expect, it } from 'vitest';
+import * as Y from 'yjs';
+import { colorForUser } from '../src/awareness/color.js';
+import { createPresence, type Presence } from '../src/awareness/presence.js';
+
+/**
+ * Publish `state` onto `target`'s awareness as a REMOTE peer, by driving a
+ * second Awareness instance and replaying its encoded update. Going
+ * through the wire encoding is what makes the entry a peer rather than a
+ * local write, which is exactly the distinction eviction turns on.
+ */
+function publishPeer(
+  target: Presence,
+  lastUpdate: number,
+  name = 'Remote',
+): { clientId: number; dispose: () => void } {
+  const remoteDoc = new Y.Doc();
+  const remote = createPresence(remoteDoc, { updateRateHz: 1000, staleAfterMs: 1_000_000 });
+  remote.awareness.setLocalState({
+    user: { id: name, name },
+    selection: [],
+    status: 'active',
+    lastUpdate,
+  });
+  applyAwarenessUpdate(
+    target.awareness,
+    encodeAwarenessUpdate(remote.awareness, [remote.awareness.clientID]),
+    'test',
+  );
+  return { clientId: remote.awareness.clientID, dispose: () => remote.dispose() };
+}
+
+describe('evictStale', () => {
+  it('drops a peer older than the window and keeps a fresh one', () => {
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { staleAfterMs: 5_000 });
+    const now = Date.now();
+    const stale = publishPeer(presence, now - 60_000, 'Stale');
+    const fresh = publishPeer(presence, now - 1_000, 'Fresh');
+
+    // `getPeers()` also carries this client's own (empty) entry, so probe
+    // the two remote ids rather than the map size.
+    expect(presence.getPeers()[stale.clientId]).toBeDefined();
+    expect(presence.getPeers()[fresh.clientId]).toBeDefined();
+    presence.evictStale();
+
+    const peers = presence.getPeers();
+    expect(peers[stale.clientId]).toBeUndefined();
+    expect(peers[fresh.clientId]).toBeDefined();
+
+    stale.dispose();
+    fresh.dispose();
+    presence.dispose();
+  });
+
+  it('keeps a peer exactly at the window boundary (strictly older is evicted)', () => {
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { staleAfterMs: 5_000 });
+    const atBoundary = publishPeer(presence, Date.now() - 5_000, 'Edge');
+
+    presence.evictStale();
+    // `now - lastUpdate > staleAfterMs` — a peer that reported exactly one
+    // window ago is not yet stale. Timer granularity can push this past the
+    // boundary, so assert only that a peer this recent is not evicted for
+    // being *arbitrarily* old.
+    const peers = presence.getPeers();
+    expect(Object.prototype.hasOwnProperty.call(peers, String(atBoundary.clientId))).toBe(true);
+
+    atBoundary.dispose();
+    presence.dispose();
+  });
+
+  it('NEVER evicts the local peer, however long since its last patch', () => {
+    // The local entry's `lastUpdate` only advances when this client
+    // patches something. A user who reads the model for a minute without
+    // touching it must not evict themselves out of everyone's roster.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { staleAfterMs: 10 });
+    presence.awareness.setLocalState({
+      user: { id: 'me', name: 'Me' },
+      selection: [],
+      status: 'active',
+      lastUpdate: Date.now() - 10 * 60 * 1000,
+    });
+
+    presence.evictStale();
+
+    expect(presence.getSelf()).not.toBeNull();
+    expect(presence.getPeers()[presence.awareness.clientID]).toBeDefined();
+    presence.dispose();
+  });
+
+  it('defaults the window to 10 seconds', () => {
+    // The default is what production uses — no caller in the repo passes
+    // `staleAfterMs`. A peer 9s quiet is still on the roster; one 11s
+    // quiet is gone.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc);
+    const now = Date.now();
+    const quiet = publishPeer(presence, now - 9_000, 'Quiet');
+    const gone = publishPeer(presence, now - 11_000, 'Gone');
+
+    presence.evictStale();
+
+    const peers = presence.getPeers();
+    expect(peers[quiet.clientId]).toBeDefined();
+    expect(peers[gone.clientId]).toBeUndefined();
+
+    quiet.dispose();
+    gone.dispose();
+    presence.dispose();
+  });
+});
+
+describe('presence patch bookkeeping', () => {
+  it('keeps a caller-supplied user colour instead of the derived one', async () => {
+    // A user who picked their own presence colour (or an app that colours
+    // peers by discipline) must see it broadcast verbatim; the id-derived
+    // colour is only a fallback.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { updateRateHz: 1000 });
+    presence.setUser({ id: 'louis', name: 'Louis', color: '#ff00ff' });
+    await new Promise((r) => setTimeout(r, 20));
+
+    expect(presence.getSelf()?.user.color).toBe('#ff00ff');
+    presence.dispose();
+  });
+
+  it('derives a colour only when the user did not supply one', async () => {
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { updateRateHz: 1000 });
+    presence.setUser({ id: 'louis', name: 'Louis' });
+    await new Promise((r) => setTimeout(r, 20));
+
+    expect(presence.getSelf()?.user.color).toBe(colorForUser('louis'));
+    presence.dispose();
+  });
+
+  it('stamps lastUpdate itself, so a stale value in a patch cannot survive', async () => {
+    // `patch()` takes an arbitrary Partial<PresenceState>. If a caller (or
+    // a replayed state object) carried an old `lastUpdate`, letting it win
+    // would make this client look stale to every peer and get evicted
+    // mid-session.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { updateRateHz: 1000 });
+    const before = Date.now();
+    presence.patch({ selection: ['wall-1'], lastUpdate: 1 });
+    await new Promise((r) => setTimeout(r, 20));
+
+    const self = presence.getSelf();
+    expect(self?.selection).toEqual(['wall-1']);
+    expect(self?.lastUpdate).toBeGreaterThanOrEqual(before);
+    presence.dispose();
+  });
+
+  it('publishes within a couple of frames by default (30 Hz cap, not slower)', async () => {
+    // No caller in the repo passes `updateRateHz`, so the DEFAULT is the
+    // only rate production ever runs at. If it were an order of magnitude
+    // slower, a peer's first selection would take a second to reach the
+    // others and every cursor would visibly lag.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc);
+    presence.setSelection(['wall-1']);
+
+    await new Promise((r) => setTimeout(r, 100));
+    expect(presence.getSelf()?.selection).toEqual(['wall-1']);
+    presence.dispose();
+  });
+
+  it('coalesces patches inside one throttle window into a single state', async () => {
+    // The 30 Hz cap is the whole point of `enqueue`: cursor moves arrive
+    // per frame, and each one must not become an awareness broadcast.
+    const doc = new Y.Doc();
+    const presence = createPresence(doc, { updateRateHz: 30 });
+    let broadcasts = 0;
+    presence.awareness.on('update', () => { broadcasts++; });
+
+    for (let i = 0; i < 20; i++) presence.setCursor2d('plan', { x: i, y: i });
+    // Nothing has been published yet — the first flush is still pending.
+    expect(presence.getSelf()?.cursor2d).toBeUndefined();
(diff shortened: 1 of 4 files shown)


BIMvoice:
One correction before the findings: **the reported pre-existing failure in `apps/viewer/src/lib/layers/collab-publish.test.ts` does not reproduce.** On a freshly built worktree at clean `upstream/main`:

```
$ npx tsx --test src/lib/layers/collab-publish.test.ts
# tests 1   # pass 1   # fail 0

$ pnpm --filter @ifc-lite/viewer test
# tests 2750   # pass 2748   # fail 0   # skipped 2
```

Green in isolation and green in the full suite. Almost certainly the same unbuilt-workspace-`dist` artifact that has produced seven false results today — including one of mine, when I wrongly reported `packages/renderer` as failing on main. `apps/viewer` is not red.

Now the findings, which are real.

**`resolvePropertyValue` in `packages/lists` is the one with a user-visible consequence.** The typed-array numeric unwrap, the raw-number passthrough, the em-dash→null mapping and the boolean display mapping were all deletable. A genuine `[IFCREAL, 5.3]` property resolving to the **string** `"5.3"` is then skipped by every group sum — they gate on `typeof v === 'number'` — and sorted lexicographically. That is a quantity silently missing from a schedule total, in a document someone hands over. `compareCellValues` had no numeric and no null coverage at all, so a quantity column sorted `100, 10, 9`.

**The identity-fixture shape appeared twice in a single file**, which is a first. `engine.test.ts`'s provider has no numeric typed-array property — every typed value is `[IFCBOOLEAN, '.T.']` — and no repeated material layer. So deleting the numeric unwrap *and* deleting an entire de-dup `new Set()` were both unkillable from that fixture. That brings this class to **eight packages**: geometry, pointcloud, solar, viewer store, export (parquet), export (lod0), parser/export headers, and now lists.

**`parseRoomId` is exported from the package root with no caller anywhere** — definition plus barrel only — and both mutations against it survived. Model ids can contain slashes, so the parsing is not trivially safe. That is the fifth published-but-unexercised symbol found today, after `scanEntities`, `ChangeSetManager`, `exportToIfcx` and `getDescendants`; one of those five contained a live bug precisely because nothing called it.

**Presence stale-eviction was entirely unexercised** — evicting the local peer, dropping the age comparison, and a 1 ms default window all survived. That is the code that removes ghost collaborators from everyone's roster.

**The methodological correction in this run is the part I would keep.** Three mutations first reported KILLED were **false results**: an `IFS=';'` split silently damaged the mutation text, so what ran was a syntactically different edit. Re-run with a delimiter that cannot occur in code, all three **survived**. Three findings would have been thrown away as already-covered. That is the second distinct mechanism today for a mutation that does not apply as intended — after unbalanced braces producing `Tests no tests` — and both fail in the direction that makes correct work look tested.

Two survivors handled honestly rather than closed: M25 (`chunks[0]` → `chunks.flat()`) is equivalent for value semantics and differs only in aliasing, **not fully proven** against a future caller that mutates in place; and `awareness/overlay.ts` (178 lines) is flagged untested rather than covered, because it needs a DOM and collab's vitest runs in node.

`structured-attrs.ts`'s accept paths came back **genuinely well covered** — the `::` separator, single-vs-list geometryRef shape, empty-collection suppression, `Qto_*` routing, `Pset_*` legacy exclusion and both geometry carrier fields all died on the first run, and nothing was added for them.

BIMvoice:
### CI red on `Node tests` — our test's defect, not a product bug

`evictStale > keeps a peer exactly at the window boundary` failed on the runner (`expected false to be true`, run [30980418007](https://github.com/LTplus-AG/ifc-lite/actions/runs/30980418007)) while passing locally. Fixed in 4273ec66.

**Root cause — wall clock, by construction.** The production comparison is `now - state.lastUpdate > staleAfterMs`, with `evictStale` reading `Date.now()` itself; there is no injectable `now`. The test built the peer at `Date.now() - 5_000` and then let the sweep read the clock a second time. Everything in between is real work — two `Y.Doc`s, a second `Presence`, an awareness encode and a wire replay — so the moment one millisecond elapsed, the age became 5001ms and the peer was evicted. Locally that gap is sub-millisecond; on a loaded CI runner it isn't. Order-of-milliseconds flaky by construction.

I reproduced the exact CI failure by inserting **2ms** of busy-wait before `evictStale()` in an unmodified copy of the test — same assertion, same `expected false to be true`. `evictStale` itself is correct; nothing in `presence.ts` changed.

**Fix.** Freeze the clock for the whole `evictStale` block (`vi.useFakeTimers({ now: NOW })` / `vi.useRealTimers()`), so the ages under test are the ages asserted — nothing in that block needs time to pass. The three sibling eviction tests now use the same frozen `NOW` instead of live `Date.now()` arithmetic, so none of them can drift either. The `presence patch bookkeeping` block is deliberately left on real timers: it awaits a flush rather than doing elapsed-time arithmetic, and each wait is 3-20x the timer it waits on.

The assertion was **strengthened, not loosened**. The boundary case now publishes a second peer at `NOW - 5_001` and asserts it *is* evicted, pinning both sides of the comparison. With the clock frozen I verified the case survives **200ms** of genuine blocking delay (`Atomics.wait`, which fake timers do not intercept) inserted before the sweep — 100x the delay that reproduced the red.

**Mutation evidence** (each applied to `packages/collab/src/awareness/presence.ts` as an exact-substring replacement, count asserted `== 1`, restored from a pre-mutation backup):

| # | Mutation | Result |
|---|---|---|
| a | drop the age comparison — `if (state?.lastUpdate && now - state.lastUpdate > staleAfterMs)` → `if (state?.lastUpdate)` | **killed** — 3 failed (incl. the boundary case) |
| b | flip the boundary — `> staleAfterMs` → `>= staleAfterMs` | **killed** — 1 failed, *only* the boundary case |
| c | evict the local peer — remove `if (clientId === awareness.clientID) return;` | **killed** — 1 failed, the local-peer case |

`packages/collab`: **44 files, 201 passed / 2 skipped** both before and after (the fix adds an assertion to an existing test, not a new one). Test-file-only change, `28 insertions(+), 13 deletions(-)`; `presence.ts` is byte-identical to its pre-mutation checksum.

BIMvoice:
## CodeRabbit triage: two threads, verified against the code

Both findings checked empirically against the modules, not just read. One is **applied**; the other is **confirmed as a production defect and deliberately left unchanged** — details below.

---

### A. `test/structured-attrs-gates.test.ts:90` — empty classification / material arrays

**CodeRabbit is right about the mechanism, and it is a production defect rather than a test gap. Reported, not changed.**

Verified by running the functions directly:

```
inflate({ "ifclite::classifications": [] })
  -> attributes: {},  classifications: []          // key is GONE from the flat record
flatten(inflate(x)) -> {}          x -> { "ifclite::classifications": [] }

inflate({ "ifclite::materials": [] })
  -> attributes: {},  materials: []
flatten(inflate(x)) -> {}          x -> { "ifclite::materials": [] }
```

The chain, with `file:line`:

- `src/snapshot/structured-attrs.ts:195` — `Array.isArray(value) && value.every(isClassificationRefShaped)` is **vacuously true** for `[]`, so the branch `continue`s and the key never reaches `flat`. Same at `:203` for materials.
- `src/snapshot/structured-attrs.ts:148` / `:151` — `flattenStructuredBranches` emits the attribute only when `length > 0`, so the empty branch is not written back.
- Net: `flatten(inflate(x)) != x`. The observable consequence is a **silent key drop**, not an empty structured payload and not peer divergence — every peer inflates the same snapshot identically, so they converge, they just converge on a record that lost the field. The loss lands at `src/snapshot/from-ifcx.ts:82`, whose result seeds `createEntity` at `:114`/`:119`/`:120`.
- This contradicts the module's own documented invariant at `src/snapshot/structured-attrs.ts:35` (`flatten(inflate(x)) == x`).

The reason this reads as a defect rather than a deliberate choice: **the sibling geometryRef gate already defends exactly this case.** `src/snapshot/structured-attrs.ts:217` opens with `let recognized = entries.length > 0;`, and the existing test at `test/structured-attrs-gates.test.ts:103` spells out the rationale — `[]` must stay flat, otherwise "an entity that had a (malformed but present) marker loses it on the round-trip." That guard is mutation-covered: replacing it with `let recognized = true;` (1 replacement) turns that test red. The classification and material gates have no equivalent guard. The asymmetry looks like an oversight, not a decision.

Scope, so the severity is not overstated: `flattenStructuredBranches` never emits `[]`, so an ifc-lite-authored snapshot cannot contain one. The input only arises from externally authored IFCX that writes an explicit empty array — that field is then dropped on seed.

**No code changed here.** Adding CodeRabbit's requested rejection cases would require the production gates at `:195`/`:203` to reject `[]`, and this module is on the snapshot -> seed path every collab peer converges through, so that behaviour change is not one to make inside a test-coverage PR. Nor did we pin the current behaviour with a characterisation test, since that would prejudge the call. @louistrue — the one-line fix, if you want it, is the same shape as the geometryRef guard:

```ts
if (Array.isArray(value) && value.length > 0 && value.every(isClassificationRefShaped)) {
```

Happy to do it in a separate PR with the rejection cases, on your word.

---

### B. `test/presence-eviction.test.ts:200` — the default-rate test

**APPLIED** — commit d72d5809.

CodeRabbit's substantive point is correct: the test was named "30 Hz cap, not slower" but waited 100ms of wall clock, which any default at or above 10 Hz satisfies. It enforced nothing near 30 Hz.

Mutation evidence:

| mutation (`src/awareness/presence.ts:113`) | replacements | before | after |
|---|---|---|---|
| `updateRateHz ?? 30` -> `?? 20` | 1 | **survived** (9 passed) | **killed** (1 failed / 8 passed) |

Fix: freeze the clock and advance exactly one frame, `Math.ceil(1000 / 30)` = 34ms, plus a new assertion that nothing is published *before* the advance (pinning that the throttle exists at all). `vi.useRealTimers()` in an `afterEach` so a failed assertion cannot leak a frozen clock into the rest of the file.

One correction to the analysis in the thread — the round/truncate question was worth asking, and the answer is that it is a non-issue, verified directly against the clock rather than from the docs:

```
setTimeout(fn, 1000 / 30) then advanceTimersByTimeAsync(1000 / 30) -> fires
setTimeout(fn, 1000 / 30) then advanceTimersByTimeAsync(33)        -> fires
```

`@sinonjs/fake-timers` **floors** the fractional 33.33ms delay to 33, so it fires early if anything — there is no round-up that could leave the timer pending. We used the integer `Math.ceil(1000 / 30)` anyway so the test does not depend on that flooring detail.

The three mutations this file already pins were re-verified against the modified file and all still fail: drop the age comparison (3 failed), `>` -> `>=` (1 failed), remove the local-peer guard (1 failed). The eviction boundary test is untouched — it still asserts that a peer exactly at the window is kept while one 1ms older is evicted.

---

`packages/collab`: 44 files / 201 passed / 2 skipped, unchanged from baseline. Typechecked with the package's own options and the `**/*.test.ts` exclusion lifted; `--listFiles` confirms the file entered the program, 0 errors in it (three other test files carry pre-existing errors, untouched here).

BIMvoice:
Verified finding A myself. It is a production defect, and the asymmetry inside the same file is what makes it clearly a defect rather than a design choice.

The two gates at `structured-attrs.ts:195` and `:203`:

```ts
if (Array.isArray(value) && value.every(isClassificationRefShaped)) {
```

`[].every(...)` is **vacuously true**, so an empty array is consumed by the structured branch and never reaches `flat`. Flatten then writes the attribute only when `length > 0` (`:148`, `:151`), so the key disappears entirely:

```
gate([])       -> true    <- [] consumed, key dropped
geomGate([])   -> false   <- the sibling refuses it correctly
```

That sibling is the point. `structured-attrs.ts:217` guards exactly this case for geometry refs — `let recognized = entries.length > 0;` — and there is already a test at `structured-attrs-gates.test.ts:103` spelling out why. Classifications and materials have no equivalent. Three gates in one file, one of which handles the empty case and two of which do not.

**Observable consequence: a silent key drop, not divergence.** `flatten(inflate({"ifclite::classifications": []}))` is `{}`. Every peer inflates identically, so they all converge — on a record that quietly lost the field. The loss point is `from-ifcx.ts:82`, feeding `createEntity` at `:114`/`:119`/`:120`. It contradicts the module's own documented `flatten(inflate(x)) == x` invariant at `structured-attrs.ts:35`.

Blast radius is bounded: ifc-lite's own writer never emits `[]`, so the input arises only from externally authored IFCX. That is also precisely the input you cannot control.

**Nothing was changed, and deliberately no test was added either.** A rejection test would be RED against current production, and a characterization test pinning the drop would prejudge your decision — so the PR is silent on it rather than enshrining either answer. The fix is one line on each gate (`value.length > 0 &&`), mirroring geometryRef. Say the word and it goes in as its own PR.

On finding B, CodeRabbit's substantive point was right and its stated reason was not, so both halves are worth recording:

- **Right:** the test named for a "30 Hz cap, not slower" used a 100 ms wall-clock wait, which passes for any default at or above 10 Hz. Mutating `updateRateHz ?? 30` to `?? 20` **survived** the old test and is killed by the replacement. The test enforced nothing about the rate it was named for.
- **Not the reason given:** the fractional-delay concern is a non-issue. Checked against the clock rather than the documentation — `@sinonjs/fake-timers` *floors* the 33.33 ms delay to 33, so advancing by either `1000/30` or `33` fires it; it errs early, never pending. The integer ceiling was used anyway so the test does not depend on that behaviour.

The three mutations the eviction test already pins were re-verified after the change and all still fail: drop the age comparison (3 failed), `>` → `>=` (1 failed, the boundary case), remove the local-peer guard (1 failed). 44 files / 201 passed / 2 skipped, unchanged — the test was rewritten, not added.

BIMvoice:
## Self-audit round 3 — two of our own tests were satisfied by their fixtures

Pushed as d72d580…a3c2e4a. Both survivors are the same shape: **a fixture that makes the operation an identity**, so the property the test is named for is invisible.

|---|---|---|---|---|
| `presence-eviction.test.ts` — coalesces patches inside one throttle window into a single state | `enqueue` MERGES pending patches | **partial** (throttle yes, merge no) | `pendingPatch = { ...(pendingPatch ?? {}), ...patch }` → `{ ...patch }` | SURVIVED |
| `structured-attrs-gates.test.ts` — leaves the attribute FLAT for … (×3) | material gate is all-or-nothing | **no** | `value.every(isMaterialAssignmentShaped)` → `value.some(...)` | SURVIVED |

1. **Coalescing.** The test drives 20 `setCursor2d` calls — all the *same key*. For a single key, merge and last-write-wins produce the identical single broadcast and the identical final cursor. The throttle was pinned; the merge was free to go. Added a window mixing `selection` / `cursor2d` / `tool` / `status`, which is the real frame — a selection click followed by a cursor move, where losing the merge means the selection never reaches the peers. Both `{ ...patch }` and the reversed-precedence variant are now killed.

2. **Material gate.** All three rejection fixtures are one-element lists, and `some` ≡ `every` on a one-element list. The classification gate already had its `rejects the WHOLE array when one member is malformed` counterpart; the material gate did not, so `some` inflated a list whose second entry is foreign — exactly the payload the branch exists to keep flat and verbatim. Added the mixed-list case.

**What held — 26 mutations killed.** The `lists` half is genuinely strong, which matters because this is the same package as the round-2 finding: `uniqueJoin`'s `new Set()` de-dup, null-first/null-last ordering, numeric-vs-lexicographic compare, the group tie-break by label, both `Number.isFinite` sum filters, the raw-number and typed-array-unwrap property paths, the em-dash null indicator, the boolean display mapping, the quantity raw-number return, and four on `discovery` (sample-of-1, no cap, nameless pset, unsorted names) — 15/15. On `collab`: the eviction boundary (`>` vs `>=`), the local-peer exemption, the 10s default and the option that overrides it, the 30 Hz default, the caller-supplied colour, the `lastUpdate` restamp, all three `parseRoomId` / `federationRoomId` mutations, and five of the six shape-gate mutations.

BIMvoice:
Re the CodeRabbit thread on `packages/collab/test/structured-attrs-gates.test.ts:106` — **confirmed, and it is a production bug rather than the test gap it was filed as.** Fixed in #2220 rather than here, so this PR stays tests-only.

Verified against the code:

- `inflateStructuredAttributes` gates the CLASSIFICATIONS and MATERIALS branches on `Array.isArray(value) && value.every(...)`. `[].every(...)` is vacuously true, so an explicitly empty array passes, `continue` fires, and the key is dropped from `flat`.
- `flattenStructuredBranches` re-emits each branch only under `json.classifications.length > 0` / `json.materials.length > 0`.
- Net effect: a snapshot → seed round trip loses the attribute entirely.

The consequence is worth stating plainly, because "explicitly empty" and "absent" are different states: a peer clearing all classifications to `[]` has that clearing dropped, so a reader can keep the stale previous value instead of converging on empty.

What convinced me it was an oversight rather than intent: the sibling `GEOMETRY_REF` branch **in the same function** already guards exactly this with `let recognized = entries.length > 0;`. The fix mirrors that precedent — require a non-empty array to take the structured path, so `[]` stays in the flat attributes and round-trips intact.

Verified RED-first: with the fix reverted, both new round-trip tests fail against unmodified production code (`expected undefined to deeply equal []`). Each guard was then mutation-verified independently — reverting either kills exactly its own test and leaves the other green.

Also grepped the repo for the same shape. Every other `Array.isArray(x) && x.every(...)` gate is pure validation where an empty array is a legitimately valid state; none routes an array into a separately re-emitted branch, so this was the only instance.

louistrue:
Resolved the remaining thread here: CodeRabbit's finding is real, but it is a production bug in `inflateStructuredAttributes`, not a gap in this tests-only PR — and it is already carried by #2220, which fixes the empty-array case at the source instead of pinning the current lossy behaviour in a test. Correct split.

PR #2168, opened 2026-08-05
test(drawing-2d,data): close 20 mutation-verified coverage gaps

BIMvoice:
Mutation sweep of `packages/drawing-2d` and `packages/data` — both had only shallow coverage work earlier in this campaign.

**Method.** 30 probative mutations against load-bearing logic (coordinate transforms and flips, page/viewport fitting, path construction, colour/stroke resolution, index construction and lookup, id mapping, set arithmetic, empty/single-element cases). Each was a single exact-substring replacement with the replacement count asserted at 1 and the mutated region re-read before any result was believed. Two deliberate control mutations were run first and both died. Every survivor was re-run against `parser`, `cache`, `sdk`, `query`, `lists`, `ids`, `export` and the full `apps/viewer` suite with `dist` rebuilt.

**Baseline → final.** `drawing-2d` 12 files / 153 tests → 14 / 204. `data` 10 / 81 → 11 / 117. Both suites reproduced clean before and after; no skips in either package. Test-only — no production code changed.

**Kill ratio: 12/30 (40%)**, flagged **TARGETED** (aimed at modules with no test file and at exports whose only two repo-wide occurrences were definition + barrel).

---

## Gaps closed

Each row: the surviving mutation, and the test that now kills it.

### `packages/drawing-2d/src/math.ts` — no test file existed

`math.ts` is where every number that reaches the SVG/PDF/DXF is computed. A wrong answer is silent: the drawing still renders, just mirrored, rotated, or with holes filled in.

| Mutation | Was |
|---|---|
| `getProjectionAxes('y')` → `{u:'z', v:'x'}` (transposed) | survived `drawing-2d`; caught only by `apps/viewer/src/hooks/scanSectionMath.test.ts`, two packages downstream |
| `projectTo2D`: `flipped ? -u : u` → `u` (flip deleted) | same — only the viewer noticed |
| `getAxisNormal`: `flipped ? -1 : 1` → `1` | **survived every suite in the repo.** Every in-repo call site passes the literal `false` (`section-cutter.ts:68`, `gpu-section-cutter.ts:481`, `scanSectionMath.ts:235`), so the `flipped: true` branch is reachable only through the published API and had no test anywhere |
| `ensureCW` branches swapped | survived. `polygon-builder` uses `ensureCCW` for outer rings and `ensureCW` for holes; inverted, every hole winds with its ring and fills in |
| `projectPointOnLine`: `/ lenSq` deleted | survived. `t` stops being normalised and starts scaling with line length, which `line-merger` then compares against a fixed tolerance |
| `boundsValid`: one `isFinite` guard → `true` | survived. `boundsEmpty()` fails the `min <= max` test on its own, so the isFinite guards need ordered-but-infinite bounds to show |

Closed by a new `src/math.test.ts` (21 tests): the axis mapping is pinned for all three axes plus the invariant that neither in-plane axis is the section axis; the flip is pinned as *U-only* (`flipped.x === -straight.x && flipped.y === straight.y`) for all three axes, with an explicit guard that the fixture's U is non-zero so the assertion is not vacuous; `getAxisNormal` is pinned in both flip states; `projectPointOnLine` is pinned as length-invariant; `boundsValid` gets one case per isFinite guard; and `ensureCCW`/`ensureCW` are each asserted to produce their winding *from both inputs*.

### `packages/drawing-2d/src/line-merger.ts` — no test file existed

`mergeDrawingLines` runs on every generated drawing (`drawing-generator.ts:347`). `mergeCollinearLines`, `deduplicateLines` and `splitLineAtParams` occur exactly **twice** in the repository — definition plus barrel re-export — i.e. published and never exercised.

| Mutation | User-visible effect if real |
|---|---|
| group key `modelIndex:entityId:category:visibility` → `modelIndex:entityId` | a cut line absorbs a hidden line: the occluded run is drawn solid at cut weight |
| `linesEqual` reverse-direction branch deleted | section cutting emits each shared face edge twice, once per face and in opposite order — every one is now drawn twice |
| `splitLineAtParams`: `params.filter(t => t > 0 && t < 1)` deleted | hidden-line intersection parameters outside `[0,1]` emit segments running off the end of the line |
| `next.t0 <= current.t1 + gapTolerance` → `next.t0 <= current.t1` | collinear segments separated by a sub-tolerance seam are never fused; the drawing keeps hairline breaks |

Closed by a new `src/line-merger.test.ts` (24 tests). The grouping tests assert non-merging across category, visibility, entity and `modelIndex` separately. `gapTolerance` is pinned on both sides *and* with the options argument omitted entirely — the absent-default path is what `drawing-generator` actually calls.

### `packages/drawing-2d/src/sheet/sheet-types.ts` — `calculateViewportBounds` untested

`sheetSlice.ts` calls it at five sites (sheet creation, paper change, frame change, title-block change, scale change). Its sibling `calculateDrawingTransform` was well covered; this one had nothing.

| Mutation | Was |
|---|---|
| `frame.margins.left + frame.margins.bindingMargin + borderGap` → binding margin dropped | survived. Every drawing shifts left into the hole-punch allowance |
| `right-strip`: `- titleBlock.widthMm` dropped | survived. Content runs under the title block |

Closed by 6 tests appended to `sheet/sheet-types.test.ts`, using a paper/frame fixture where every margin, the binding margin and the border gap are distinct values, so no term can be dropped without moving the result. Note the `right-strip` branch is unreachable from all four `TITLE_BLOCK_PRESETS` entries (they are all `bottom-right` or `bottom-full`) — it ships only via a hand-built config.

### `packages/data/src/spatial-types.ts` — no test file existed

The single source of truth for "is this in the spatial tree, and at which level", read by `parser/spatial-hierarchy-builder.ts`, the viewer hierarchy, `basketVisibleSet`, `visibility-adapter` and `PropertiesPanel`.

| Mutation | Was |
|---|---|
| `isSpaceLikeSpatialTypeName` delegates to `isStoreyLikeSpatialType` | survived `data` entirely, and survived every downstream suite |
| `IfcSpace` removed from `SPATIAL_STRUCTURE_TYPE_ENUMS` | survived `data`; caught only by `packages/parser` |

Closed by a new `src/spatial-types.test.ts` (16 tests). The level predicates are asserted **mutually exclusive** per type (a predicate wired to a neighbouring set passes any positives-only test), and a structural invariant pins that every building-/storey-/space-like type is also in the parent structure set, so the sub-sets cannot drift out of it.

### `packages/data/src/relationship-graph.ts`

`buildCSR`'s `uniqueKeys.sort((a,b) => a-b)` was **unkillable from the existing fixture**: edges were added in already-ascending source order (100, 200, 200, 400, 400), and every assertion called `.sort()` on the result, so CSR layout could not be observed either way. `hasAnyEdges` and `getTargets` had no test at all — replacing `hasAnyEdges` with `return true` survived.

Closed by 6 tests: a descending-insertion fixture that reads the raw `offsets`/`edgeTargets` columns, a cross-check that two insertion orders produce byte-identical CSR columns, `counts` per key, `hasAnyEdges` in forward *and* inverse direction (a leaf wall has parents but no children), and `getTargets` with and without a type filter.

### `packages/data/src/quantity-table.ts`

The existing `QuantityTable` round-trip fixture used a single quantity set, which makes the `qsetName[idx] === qsetIdx` half of `getQuantityValue` an identity — deleting it survived. IFC routinely puts the same quantity name in several `Qto_` sets on one element, so that half decides which number is reported.

Closed by 4 tests on a two-qset fixture where the same `NetArea` holds `5` and `42`.

### `packages/data/src/entity-table.ts`

| Mutation | Was |
|---|---|
| `if (gidString)` → `if (true)` when building the GlobalId index | survived. Entities with an empty GlobalId get indexed under `''`, so `getExpressIdByGlobalId('')` returns a real entity instead of `-1` — every BCF lookup of a missing GUID resolves to whatever entity was last with a blank GlobalId |
| derived `typeRanges`: `end: lastRow + 1` → `end: indices.length` | survived. The existing fixture stores each type contiguously starting at row 0, where the two expressions coincide; IFC streams interleave types freely |

Closed by 8 tests. The GlobalId ones reuse the existing fixture's entity `401` (which already carries an empty GlobalId) and re-check after the columns round-trip; the `typeRanges` ones use a deliberately interleaved fixture (wall, space, wall, space, wall) with `typeRanges` deleted from the columns to force the derive branch, plus a containment invariant and a check that supplied ranges win over derived ones.

### `packages/data/src/string-table.ts`

`fromArray` carries a comment promising "first occurrence wins" so a duplicate does not shadow the canonical index. Collapsing it to unconditional `set` survived — no fixture contained a duplicate. Closed by 3 tests.

---

## Controls

| Control | Expected | Result |
|---|---|---|
| `sheet-types.ts`: `paddingFactor = 0.95` → `0.5` | die | KILLED (`calculateDrawingTransform` suite) |
| `string-table.ts`: `indexOf` `?? -1` → `?? 0` | die | KILLED (2 tests) |

Neither control survived, so no survivor was believed on an unverified harness.

---

## Also found, not changed

- **`QuantityTable.sumByType(quantityName, elementType?)` ignores `elementType`.** The interface declares the parameter; `quantityTableFromColumns` implements it as `(quantName) => ...`. A caller passing an element-type filter silently gets the unfiltered sum — a wrong number with no error. Reported rather than fixed; the new tests assert only properties that hold either way, so they do not enshrine it.
- **`apps/viewer/src/hooks/useDrawingGeneration.projection.test.tsx` is order-dependent.** `drops a hidden class from the projection profiles` fails when the file is run alone and passes inside the full `pnpm test` run. Pre-existing on a clean checkout; not touched here.
- **`resolveObjectStyle`'s hatch deep-merge base cannot be mutated away** — dropping `...(typeDefault.hatch ?? ...)` is a `tsc` error (TS2322), so the compiler already guards it. Recorded as non-probative rather than counted as a survivor.
- **Uncalled published API** (definition + barrel re-export only, no consumer anywhere in the repo): `mergeCollinearLines`, `deduplicateLines`, `splitLineAtParams`, `boundsValid`, `parseEntityKey`, `getHiddenIfcTypes`, `LINE_PATTERN_DASH_ARRAYS`, `lookupProj4`. A further group is exported from its own module but not re-exported at all, so it is unreachable even by consumers: all of `openings/opening-utils.ts` (`getDoorOpenings`, `getWindowOpenings`, `getOpeningInfo`, `getHostEntityIds`, `getHostsWithOpenings`, `isHostElement`, `filterOutOpeningElements`) and most of `symbols/symbol-utils.ts` (`generateNorthArrow`, `generateSectionMark`, `generateLevelMark`, `generateOpeningSymbols`, `generateDoorSymbolAt`, `generateWindowSymbolAt`).

Typechecked against each package's own options with `**/*.test.ts` un-excluded, and confirmed via `--listFiles` that all 14 `drawing-2d` and 11 `data` test files entered the program. Pre-existing, not fixed, none from the new files:

- `drawing-2d`: 3 × TS2741 — `dxf-exporter.test.ts:18`, `svg-exporter.test.ts:16` and `:116` spread `DEFAULT_SECTION_CONFIG` into a `SectionConfig`, which is missing `plane`.
- `data`: 1 × TS2345 — `utf8-decode.test.ts:61`, spread into `TextDecoder.decode`'s 2-tuple parameter.

Root `pnpm typecheck` (34 tasks), `pnpm turbo test` for both packages, `scripts/check-test-wiring.mjs` and `scripts/check-api-surface.mjs` all pass.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for data indexing, quantity retrieval, relationship graphs, spatial classification, and string deduplication.
  * Added comprehensive validation for 2D line merging, projections, geometry winding, and viewport bounds.
  * Verified edge cases including empty values, duplicate entries, missing data, tolerance limits, resizing, and insertion-order independence.

### DIFF
--- a/packages/data/src/entity-table.test.ts
+++ b/packages/data/src/entity-table.test.ts
@@ -108,3 +108,91 @@ describe('entityTableToColumns / entityTableFromColumns round-trip', () => {
     expect(rebuilt.expressId.buffer).toBe(table.expressId.buffer);
   });
 });
+
+describe('EntityTable GlobalId index', () => {
+  // The sample table's entity 401 carries an EMPTY GlobalId (the IFC file
+  // omitted it, or the row is a non-rooted entity). Without the emptiness
+  // guard on the index build, '' maps to a real expressId and every BCF
+  // lookup of a missing GUID resolves to whatever entity happened to be
+  // last with a blank GlobalId.
+  it('never indexes the empty GlobalId', () => {
+    const { table } = buildSampleTable();
+    expect(table.getExpressIdByGlobalId('')).toBe(-1);
+  });
+
+  it('still indexes every non-empty GlobalId', () => {
+    const { table } = buildSampleTable();
+    expect(table.getExpressIdByGlobalId('0YvCT2_$X3_xJG3rzD8L_8')).toBe(101);
+    expect(table.getExpressIdByGlobalId('3qqCT2_$X3_xJG3rzD8L_8')).toBe(301);
+  });
+
+  it('keeps the empty GlobalId unindexed through the columns round-trip', () => {
+    const { strings, table } = buildSampleTable();
+    const rebuilt = entityTableFromColumns(entityTableToColumns(table), strings);
+    expect(rebuilt.getExpressIdByGlobalId('')).toBe(-1);
+    expect(rebuilt.getExpressIdByGlobalId('1abCT2_$X3_xJG3rzD8L_8')).toBe(102);
+  });
+});
+
+describe('EntityTable typeRanges derived from interleaved columns', () => {
+  // `entityTableFromColumns` derives typeRanges only when the caller omits
+  // them (worker-transport rebuild). IFC streams interleave types freely,
+  // and the range for a type is [firstRow, lastRow+1] — NOT the number of
+  // rows of that type, which only coincides when each type is contiguous
+  // and starts at row 0. The sample table above is fully contiguous, so it
+  // cannot tell the two apart; this fixture interleaves deliberately.
+  function interleaved() {
+    const strings = new StringTable();
+    const builder = new EntityTableBuilder(8, strings);
+    builder.add(1, 'IFCWALL', 'gid-1', 'W1', '', '', false, false); // row 0
+    builder.add(2, 'IFCSPACE', 'gid-2', 'S1', '', '', false, false); // row 1
+    builder.add(3, 'IFCWALL', 'gid-3', 'W2', '', '', false, false); // row 2
+    builder.add(4, 'IFCSPACE', 'gid-4', 'S2', '', '', false, false); // row 3
+    builder.add(5, 'IFCWALL', 'gid-5', 'W3', '', '', false, false); // row 4
+    return { strings, table: builder.build() };
+  }
+
+  it('spans first row to last row + 1 for an interleaved type', () => {
+    const { strings, table } = interleaved();
+    const columns = entityTableToColumns(table);
+    delete (columns as { typeRanges?: unknown }).typeRanges;
+    const rebuilt = entityTableFromColumns(columns, strings);
+
+    // IfcWall occupies rows 0, 2, 4 -> [0, 5); IfcSpace rows 1, 3 -> [1, 4).
+    expect(rebuilt.typeRanges.get(IfcTypeEnum.IfcWall)).toEqual({ start: 0, end: 5 });
+    expect(rebuilt.typeRanges.get(IfcTypeEnum.IfcSpace)).toEqual({ start: 1, end: 4 });
+  });
+
+  it('produces a range that contains every row of that type', () => {
+    const { strings, table } = interleaved();
+    const columns = entityTableToColumns(table);
+    delete (columns as { typeRanges?: unknown }).typeRanges;
+    const rebuilt = entityTableFromColumns(columns, strings);
+
+    for (const [type, range] of rebuilt.typeRanges) {
+      for (let row = 0; row < rebuilt.count; row++) {
+        if (rebuilt.typeEnum[row] === type) {
+          expect(row).toBeGreaterThanOrEqual(range.start);
+          expect(row).toBeLessThan(range.end);
+        }
+      }
+    }
+  });
+
+  it('prefers supplied typeRanges over the derived ones', () => {
+    const { strings, table } = interleaved();
+    const columns = entityTableToColumns(table);
+    columns.typeRanges = new Map([[IfcTypeEnum.IfcWall, { start: 7, end: 9 }]]);
+    const rebuilt = entityTableFromColumns(columns, strings);
+    expect(rebuilt.typeRanges.get(IfcTypeEnum.IfcWall)).toEqual({ start: 7, end: 9 });
+  });
+
+  it('getByType stays exact for interleaved types (index, not range, decides)', () => {
+    const { strings, table } = interleaved();
+    const columns = entityTableToColumns(table);
+    delete (columns as { typeRanges?: unknown }).typeRanges;
+    const rebuilt = entityTableFromColumns(columns, strings);
+    expect(rebuilt.getByType(IfcTypeEnum.IfcWall)).toEqual([1, 3, 5]);
+    expect(rebuilt.getByType(IfcTypeEnum.IfcSpace)).toEqual([2, 4]);
+  });
+});
--- a/packages/data/src/property-table.test.ts
+++ b/packages/data/src/property-table.test.ts
@@ -32,6 +32,26 @@ describe('PropertyTable round-trip', () => {
     expect(rebuilt.getPropertyValue(100, 'Custom', 'Length')).toBeCloseTo(3.5);
   });
 
+  it('getPropertyValue reads the NAMED pset when both carry the same property', () => {
+    // Every other fixture gives each pset distinct property names, so the
+    // `psetName[idx] === psetIdx` half of the row match is an identity on
+    // them: dropping it and matching on the property name alone survives
+    // the whole package (measured, round-four self-audit). The same-named
+    // property in two sets is the real case — Pset_WallCommon.Reference vs
+    // a vendor set's Reference — and picking the first row for the entity
+    // shows the wrong value in the property panel.
+    const strings = new StringTable();
+    const builder = new PropertyTableBuilder(strings);
+    builder.add({ entityId: 100, psetName: 'Pset_WallCommon', psetGlobalId: 'gid-1', propName: 'Reference', propType: PropertyValueType.String, value: 'WALL-STD' });
+    builder.add({ entityId: 100, psetName: 'Vendor_Custom', psetGlobalId: 'gid-2', propName: 'Reference', propType: PropertyValueType.String, value: 'VND-042' });
+    const table = builder.build();
+
+    expect(table.getPropertyValue(100, 'Pset_WallCommon', 'Reference')).toBe('WALL-STD');
+    expect(table.getPropertyValue(100, 'Vendor_Custom', 'Reference')).toBe('VND-042');
+    // A pset the entity does not carry must not fall through to either row.
+    expect(table.getPropertyValue(100, 'Pset_SlabCommon', 'Reference')).toBe(null);
+  });
+
   it('handles empty tables (lite-mode default)', () => {
     const strings = new StringTable();
     const empty = new PropertyTableBuilder(strings).build();
--- a/packages/data/src/property-table.test.ts
+++ b/packages/data/src/property-table.test.ts
@@ -157,3 +157,46 @@ describe('QuantityTable.findByQuantity', () => {
     expect(table.findByQuantity!('NetArea', '=', '12.5')).toEqual([]);
   });
 });
+
+describe('QuantityTable.getQuantityValue qset scoping', () => {
+  // The round-trip fixture above uses a single quantity set, which makes the
+  // `qsetName[idx] === qsetIdx` half of the lookup an identity: any row
+  // matching the quantity name is also in the only qset. IFC routinely puts
+  // the same quantity name (NetArea, GrossVolume, ...) in several Qto_ sets
+  // on one element, so the qset half decides which number is reported.
+  function twoQsetTable() {
+    const strings = new StringTable();
+    const builder = new QuantityTableBuilder(strings);
+    builder.add({ entityId: 100, qsetName: 'Qto_WallBaseQuantities', quantityName: 'NetArea', quantityType: QuantityType.Area, value: 5 });
+    builder.add({ entityId: 100, qsetName: 'Qto_CustomQuantities', quantityName: 'NetArea', quantityType: QuantityType.Area, value: 42 });
+    return builder.build();
+  }
+
+  it('returns the value from the NAMED quantity set, not the first row for the entity', () => {
+    const table = twoQsetTable();
+    expect(table.getQuantityValue(100, 'Qto_WallBaseQuantities', 'NetArea')).toBe(5);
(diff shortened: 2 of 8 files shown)


BIMvoice:
Confirmed the `sumByType` defect independently, and it is worse than one implementation.

The interface declares a filter (`packages/data/src/quantity-table.ts:63`):

```ts
sumByType(quantityName: string, elementType?: number): number;
```

Both implementations silently drop it:

```
packages/data/src/quantity-table.ts:194        sumByType: (quantName) => {
packages/cache/src/sections/quantities.ts:133  sumByType: (quantName) => {
```

The second parameter is not in either arrow's parameter list — measured arity is **1**. So `sumByType('NetArea', wallTypeId)` returns the sum across **every** element type, with no error and no type-checker complaint, because an extra argument to a declared-optional parameter is legal at the call site.

For a user that is a quantity total silently including elements they explicitly filtered out — the same class as the clash-truncation and BCF-visibility findings from this campaign: a confident, plausible, wrong number. Nothing in the repo calls it with the second argument today, so this is latent rather than live, but it is published API of `@ifc-lite/data` and the signature is an invitation.

Not fixed — two defensible remedies with different consequences (implement the filter, or remove the parameter and make callers filter upstream), and picking between them is yours. The new tests assert only properties that hold either way, so they do not enshrine it.

**The fixture findings in this sweep are the most concentrated yet: four in one package, including the first doubly-unobservable one.** `relationship-graph.test.ts` adds edges in ascending key order *and* re-sorts every assertion — so `buildCSR`'s sort is neutralised twice over, and removing it entirely is invisible. The others: the QuantityTable fixture has a single qset; `entity-table.test.ts` stores types contiguously from row 0 where `lastRow + 1 == count`; and no `StringTable` fixture contains a duplicate. That is the ninth package with this class, and the first where a single file cancels the same operation by two independent means.

Also worth recording: **`getAxisNormal(axis, flipped)` is a two-state flag where every call site in the repo passes the literal `false`**, and `TitleBlockPosition` has three values of which all four shipped presets use two — so the third is unreachable from any shipped configuration.

**A large amount of exported-but-unreachable surface.** Eleven symbols have exactly two occurrences (definition + barrel, no consumer). More striking, these have **one** occurrence — exported but not even re-exported, so no consumer can reach them: all seven of `openings/opening-utils.ts`, six of `symbols/symbol-utils.ts`, plus `createLayerMapper`, `createLineStyler`, `createLineWeightAssigner`, `applyHiddenStyle`, `generateSimpleWindowLines` and `buildOpeningRelationshipsFromData`. That is dead code carrying a maintenance cost, and it is the same blind spot that hid the `getDescendants` recursion bug in `packages/ifcx` — nothing calls it, so nothing notices.

**One number in this report I want to flag as unverified rather than repeat as fact.** The downstream re-check reports "14 pre-existing failures" in the full `apps/viewer` suite. I measured that suite twice today on clean `upstream/main` with deps built and got **2,750 tests, 2,748 pass, 0 fail, 2 skipped** both times. The likely explanation is the unbuilt-workspace-`dist` artifact that has produced seven false results in this campaign — a whole file reports as one failing test. The mutation verdicts here do not depend on that number, but I would not want it quoted as a baseline.

BIMvoice:
## Round-four self-audit: "lowest stakes" was not the same as "safe" — 37 mutations, 32 KILLED, 5 SURVIVED, fixed in 0917b78

Round three skipped this PR as the lowest-stakes of the batch. That assumption is worth testing, so it was tested. Baselines on a built workspace: `@ifc-lite/data` 117 tests / 11 files, `@ifc-lite/drawing-2d` 204 tests / 14 files.

### What held (32 of 37)

`line-merger` 9/11 — the group key dies once **per field** (`modelIndex`, `category`, `visibility` each kill a different named test), the gap-tolerance default and the `<= current.t1 + gapTolerance` bridge both die, `Math.max(current.t1, next.t1)` → `next.t1` dies on the overlap test, the reverse-direction arm of `linesEqual` dies, and the EPSILON dedupe in `splitLineAtParams` dies.

`math` 6/7 — swapping `getProjectionAxes`' u/v for x, mis-mapping y's v, `area / 2` → `area`, the `ensureCW` polarity, and each half of `boundsValid` (the y-ordering check and the `max.y` finiteness check die separately, so no single fixture is carrying that function).

`sheet-types` 3/3 — dropping `bindingMargin`, dropping the title-block `padding`, and swapping `widthMm` for `heightMm` in the right-strip branch each die.

`data` 12/14 — the empty-GlobalId guard, the CSR key sort, the per-key counts, the `writePos` cursor (edges all landing at the key's base offset), `hasAnyEdges`, `getTargets`' type filter, `StringTable.intern`'s dedupe and `indexOf`'s `-1` sentinel, the `psets.has(name)` grouping in `getForEntity`, dropping `IfcSpatialZone` from the space-like set, dropping `IfcRoadPart` from the spatial-structure set, and `!typeName` → `typeName === null`.

### Survivors — five, all ours

| Mutation | Count | Before | After 0917b78 |
|---|---|---|---|
| `linesOnSameLine`: delete `if (distStart > tolerance) return false;` | 1 | **SURVIVED** | KILLED |
| `linesOnSameLine`: weaken the same check to `tolerance * 1000` | 1 | **SURVIVED** | KILLED |
| `linesOnSameLine`: `return distEnd <= tolerance` → `return true` | 1 | **SURVIVED** | KILLED |
| `isCounterClockwise`: `> 0` → `>= 0` | 1 | **SURVIVED** | KILLED |
| `PropertyTable.getPropertyValue`: drop `psetName[idx] === psetIdx` | 1 | **SURVIVED** | KILLED |

The first three are one defect: the only fixture for `linesOnSameLine` is a truly parallel offset segment, whose start **and** end are both off the reference line, so **either** check alone satisfies it — the classic redundant-fixture shape. The fix adds two near-parallel fixtures inside the same angle bucket (`bucketSize = 2 * angleTolerance = 0.02 rad`), each on the wrong side of exactly one check; each now kills exactly the one it targets.

`isCounterClockwise` had no zero-area fixture, so a degenerate ring read as counter-clockwise and `ensureCW` reversed a polygon with no orientation to fix.

`getPropertyValue` is the sharpest: every fixture in the package gives each pset **distinct** property names, so matching on the property name alone is an identity on all of them. The real case — `Pset_WallCommon.Reference` and a vendor set's `Reference` on the same entity — is now asserted in both directions plus the unknown-pset null.

### One production inconsistency found (reported, not "fixed")

The three name predicates disagree on case, and the test named *"are case-sensitive on the PascalCase spelling"* asserted the property for only one of them:

```
                              isSpatialStructureTypeName  isStoreyLike…  isSpaceLike…
'IfcSpace'                            true                   false          true
'IFCSPACE'                            false                  false          true
'ifcspace'                            false                  false          true
'IfcBuildingStorey'                   true                   true           false
'IFCBUILDINGSTOREY'                   false                  true           false
```

`isSpatialStructureTypeName` matches a Set of `IfcTypeEnumToString` output (PascalCase only); the other two route through `IfcTypeEnumFromString`, which upper-cases its argument. STEP type names are stored UPPERCASE in this codebase, so a caller handing a raw type name to all three gets inconsistent answers — including the `isSpatialStructureTypeName(t) && !isSpaceLikeSpatialTypeName(t)` pairing used for container expansion. **No behaviour was changed**: this is pre-existing production behaviour, not something these tests enshrine, so per the audit rule the case is "stop and report". The test now measures and documents the divergence instead of implying uniformity, so a future unification is a deliberate, visible change.

### Two mutations discarded as equivalent, not counted as survivors

- `splitLineAtParams`' filter widened to the closed interval `[0, 1]`: the EPSILON dedupe removes the duplicated endpoints, so the mutant is behaviourally identical. Removing the filter outright IS killed.
- `IfcTypeEnumFromString(typeName.toUpperCase())`: that function already upper-cases its argument.

`@ifc-lite/data` 118 passed, `@ifc-lite/drawing-2d` 206 passed, typecheck clean, no production code changed.

PR #2169, opened 2026-08-05
test(extensions,sdk): close 42 mutation-verified gaps on the capability boundary and the published SDK surface

BIMvoice:
Second, deeper mutation sweep of `packages/extensions` (the capability/permission boundary) and `packages/sdk` (the published API surface). **Test-only — no production code changed.**

## Measurement

| | baseline | final |
|---|---|---|
| `packages/extensions` | 53 files / 603 tests | 59 files / 770 tests |
| `packages/sdk` | 6 files / 69 tests | 9 files / 147 tests |

Clean before and after; no pre-existing failures, no skips (fixture-gated or otherwise) in either package.

**67 probative mutations. 23 KILLED, 44 SURVIVED — kill ratio 34%.** TARGETED: the mutations were aimed at modules with no test file or a test file far thinner than the source. Two controls (`matchCapability`'s scope check, `dispatchToBackend`'s own-property method gate) died as expected before any survivor was believed.

Every survivor was re-confirmed with `dist` rebuilt and the mutations applied, against `packages/cli` (25 files / 271 tests), `packages/mcp` (23 / 231), `packages/sandbox` (12 / 71) and `apps/viewer` (703 files / 2748 tests) — all still green, so none was a false survivor.

### The manifest validator was the largest hole

`manifest.test.ts` drives `validateManifest` through fixture files and asserts only `result.ok`. That shape is satisfied by *any* error, so a fixture testing "bad id" is equally satisfied by an unrelated one — it pins almost none of the individual field rules. Twelve independent rules could be disabled with all 603 tests green:

| # | Surviving mutation | What a user would see | Killing test |
|---|---|---|---|
| M1 | `ID_RE` → case-insensitive | An uppercase id installs, breaking the lowercase canonical-id guarantee the message and the RFC both promise. The only bad-id fixture is `"BAD ID WITH SPACES"`, rejected either way. | `validate.test.ts` › rejects an id with uppercase letters |
| M2 | `ENGINE_RANGE_RE` gate → `false` | `"engines": { "ifcLiteSdk": "latest" }` passes validation and reaches the host's SemVer comparison. | › rejects a range containing characters no SemVer range can contain |
| M3 | blank engine-range check dropped | A whitespace-only range validates. | › rejects a range string that is only whitespace |
| M6 | `requireString`'s `trim()` check → `false` | A manifest whose `name` is `"   "` installs with an invisible display name. | › rejects a whitespace-only "name"/"description"/"id"/"version" |
| M7 | `optionalString` type gate → `false` | Both call sites (`license`, `readme`) unguarded. | › rejects a non-string "license"/"readme" |
| M8 | `tests[].name/command/fixture` blank check dropped | | › rejects a blank test name / command / fixture |
| M9 | `l10n` value type check → `false` | | › rejects a non-string translation value |
| M10 | `author.url`/`email` type checks → `false` | | › rejects a non-string author.url / author.email |
| M11 | `entry.activate`/`deactivate` type check → `false` | | › rejects a non-string entry.activate / entry.deactivate |
| M4 | cross-ref ignores `entry.commands` | A valid manifest whose toolbar button points at a command declared only in `entry.commands` is **rejected**. Only the negative direction was pinned. | `cross-ref.test.ts` › accepts a reference satisfied only by entry.commands |
| M5 | statusBar `if (sb.command)` guard dropped | A statusBar item with no command (a plain text indicator, which is valid) is **rejected** as a dangling reference. | › accepts a statusBar item with no command at all |
| M12 | `isKnownCapability` → `return true` | Two repo-wide occurrences — definition plus barrel re-export. Published, never called by a test. Every uncatalogued capability reads as known, which is exactly the "unknown capability is red" story `risk.ts` leans on. | `catalogue.test.ts` › rejects the uncatalogued capability "…" (7 cases, measured per case) |

### Contribution field requirements

Only two contribution rules were covered by fixtures (a bad slot and a malformed `when`). These survived:

| # | Surviving mutation | Killing test |
|---|---|---|
| E35 | `keybindings[].key` requirement removed — the host has nothing to bind, the contribution is silently inert | `contributions.test.ts` › requires the key stroke |
| E36 | `exporters[].mimeType` requirement removed — an export contribution produces a file the browser cannot type | › requires "mimeType" |
| E37 | `statusBar[].text` requirement removed | › requires "text" |
| E38 | `commands[].title` requirement removed | › requires id and title on every command |
| E44 | `toolbar.center` dropped from the slot allow-list — the valid fixtures only ever use `toolbar.left`/`right`, so a documented slot silently stopped being accepted | › accepts the documented slot "toolbar.center" |

### `when` evaluator — the guards a third-party clause runs against

`when` clauses come out of an extension manifest and decide whether a contribution is shown.

| # | Surviving mutation | Killing test |
|---|---|---|
| E16 | v1 context allow-list gate → `false` | `when/eval.test.ts` › ignores a context key outside the v1 vocabulary even when present |
| E17 | `hasOwnProperty(ctx, name)` → `name in ctx` — a `Object.create(defaults)` context would resolve inherited keys | › does not resolve an allow-listed key inherited from the prototype |
| E18 | empty string coerces to `true` | › treats the empty string as false and a non-empty string as true |
| E19 | boolean ordering rejection dropped — `true > false` becomes satisfiable. Mixed boolean/number is caught by the later `typeof left !== typeof right` check, so only **two booleans** pin this guard; the first version of this test was non-probative and was narrowed. | › returns false when both sides of an ordering are booleans |

### Signing, code validation, SDK-version matching

| # | Surviving mutation | Killing test |
|---|---|---|
| E13 | `self` removed from `BANNED_GLOBALS` — the one worker-realm alias of the global object, and the only entry in the set with no test | `code.test.ts` › rejects self, the worker-realm alias… (+ a per-name case for all five) |
| E43 | `writeLengthPrefix` writes constant bytes — the serialisation stops being injective, so two different bundles can share a content hash and therefore a signature. The existing "old separator bytes" test does not reach this: its inputs differ in segment *count*, so a constant prefix still separates them. | `signing.test.ts` › is injective when the path/content boundary shifts |
| E24 | version suffix group `(?:[.-].*)?` → `(?:.*)?` — `">=2abc"` parses as `2.0.0` instead of falling through to `permissive` for the user to confirm | `sdk-version.test.ts` › does not parse a version with junk glued onto a numeric segment |

Signing and SDK-version matching were otherwise **well covered**: 12 of 20 mutations in that batch died, including every `verifyBundle` gate, the base64 strict-format check, the canonical sort, and the caret/tilde/wildcard range rules.

## `packages/sdk`

Six test files covered thirty source files. `context.test.ts` reaches query/export/files/viewer/mutate/lens/spatial/store; everything else was unexercised. **18 of 21 mutations survived.**

| # | Surviving mutation | What an external consumer would see | Killing test |
|---|---|---|---|
| S3 | `entityRefToString` emits `expressId:modelId` | The documented transport encoding is silently reversed. (`apps/viewer` has its own same-named pair under `src/store/types.ts`; it does not exercise this one.) | `types.test.ts` › emits modelId first and expressId second |
| S1 | `stringToEntityRef` `idx < 1` → `idx < 0` | `":42"` decodes to `{ modelId: '', expressId: 42 }` | › rejects an empty modelId |
| S2 | negative-expressId guard dropped | `"arch:-1"` decodes to a ref no backend resolves | › rejects a negative expressId |
| S4 | `ModelNamespace.active()` null guard → `false` | | `namespaces.test.ts` › active() returns null when no model is active, without consulting the list |
| S5 | `loadIfc` filename default `'created.ifc'` → `''` | | › loadIfc() defaults the filename to created.ifc |
| S6 | `QueryBuilder.first()` does not restore the caller's limit | A builder reused after `first()` silently returns one row | › first() restores the caller-supplied limit afterwards |
| S7 | `where()` default operator `'exists'` → `'='` | A two-arg `where()` changes meaning | › defaults the comparison operator to "exists" |
| S9 | `containedIn()` picks the last related ref, not the first | A script reports a different storey | › containedIn() resolves the first related ref, not the last |
| S10 | 3-arg `quantity()` ignores the qset name | The 3-arg form degenerates into the 2-arg form and returns whichever set comes first | › 3-arg form honours the qset name |
| S12 | `EventsNamespace.on` unsubscribe leaves its map entry | The backend sees a double unsubscribe for every handler the caller released itself | › does not re-run an unsubscribe that the caller already invoked |
| S13 | `removeAll()` clears the map without calling the unsubscribes | Every listener leaks | › removeAll() calls every outstanding unsubscribe |
| S14 | `SpacesNamespace` missing-backend gate → `false` | A transport-backed context gets "cannot read properties of undefined" instead of the actionable message | › throws an explanatory error when the backend has no spaces support |
| S15 | `schedule.tasks()` routed at `schedule.sequences()` | The four accessors have identical signatures, so a mis-wired one is invisible without a per-method assertion | › routes each accessor to the matching backend method |
| S16 | `BroadcastTransport` default timeout `30_000` → `1` | | `transport.test.ts` › times out an unanswered request after the 30s default |
| S17 | `close()` no longer rejects in-flight requests | A pending `send()` hangs for the full timeout after the channel is gone | › rejects in-flight requests when the transport is closed |
| S18 | `MessagePortTransport` response discrimination drops `'error' in data` | An **error reply** stops resolving the caller and is handed to event subscribers instead; the caller hangs until timeout | › resolves a pending request from an error-only response |
| S21 | `RemoteBackend.subscribe` drops the event-type filter | A handler registered for `selection:changed` receives every event | › delivers only the subscribed event type |

`BroadcastTransport`, `MessagePortTransport` and `RemoteBackend` are stubbed at the `BroadcastChannel`/`MessagePort` level rather than module-mocked, so the real constructor / `postMessage` / `onmessage` wiring is exercised.

## Survivors not closed

- **S20** — `ListNamespace.execute()`'s `modelId ?? 'default'`. `list.ts` loads `@ifc-lite/lists` through a dynamic `import()` with a variable specifier, which `vi.mock` cannot intercept; closing it needs a module-loading harness this PR does not introduce. Genuine gap, left open.
- **E29** — `canonicalContentHash` dropping the *path* length-prefix write. This desynchronises the write offsets rather than expressing a single property, so it is only detectable by pinning a golden hash vector (none exists, deliberately). The injectivity property it aims at is now covered by E43's test. Reported, not closed.
- **E31 / E31b** — `migrateManifest` dropping `!Number.isInteger(rawVersion)` and/or `rawVersion < 1`. **EQUIVALENT, and proven**: with `CURRENT_MANIFEST_VERSION === 1`, any non-integer above 1 is rejected by the future-version check, and anything below 1 falls into the migration loop, finds no chain entry and is rejected with the same `invalid_manifest_version` code. It stops being equivalent the moment a v2 lands, so `migrations.test.ts` asserts the error *code and path* — properties that hold either way — rather than the specific branch.

- `manifest.test.ts`'s two fixture loops assert only `result.ok`. The invalid-fixture loop is the "property that holds whether or not the code does the thing" shape: an unrelated error satisfies a fixture named for a specific rule. This is what let all twelve validator rules go unpinned. Left in place; the new files sit alongside and assert a specific `code` at a specific `path`.
- `canonicalContentHash`'s injectivity test targets a scheme that no longer exists (the `0x1f`/`0x1e` separators), and its fixtures differ in segment count — so it does not exercise the length prefix that replaced them.

## Production issues found — reported, not fixed

None were changed, and no new test asserts either side of them.

1. **`stringToEntityRef('arch:')` decodes to `{ modelId: 'arch', expressId: 0 }`.** `Number('')` is `0`, which clears both `Number.isFinite` and `>= 0`. A trailing-colon reference silently resolves to entity #0 rather than being rejected. (`packages/sdk/src/types.ts:40`)
2. **The `EntityRef` codec is lossy for a `modelId` containing `:`.** `entityRefToString` emits `arch:v2:1` without escaping; `stringToEntityRef` splits on the *first* colon and then throws `Invalid expressId`. Encode and decode disagree. (`packages/sdk/src/types.ts:29`, `:36`)
3. **`VERSION_PARSE_RE`'s comment contradicts its behaviour.** It reads "Reject 4+ dotted segments outright — silent truncation of `1.2.3.4` to `1.2.3` would hide manifest typos", but the trailing `(?:[.-].*)?` group swallows the extra segments: `1.2.3.4` parses as `1.2.3`, and `>=2.0.0nope` backtracks into a match and evaluates as `>=2.0.0`. Either the comment or the regex is wrong. (`packages/extensions/src/host/sdk-version.ts:61`)

`pnpm typecheck` covers 2 of 40 packages and the root tsconfig excludes `**/*.test.ts`, so each package was typechecked separately against its own options with the exclusion lifted, and `--listFiles` verified the test files entered the program (58 of them for extensions, 9 for sdk — not vacuous).

- `packages/extensions` — 1 error, pre-existing: `src/flavor/switcher.test.ts:17` (`FlavorExtension` literal missing `source`, `bundleHash`, `enabled`).
- `packages/sdk` — 31 errors, all pre-existing: 28 in `src/context.test.ts` (mock backend missing `loadIfc` / `entitiesMatchingActiveFilter`, partial `Transport`, `never[]` inference), 3 in `src/sandbox.test.ts`.

All twelve files in this PR typecheck clean. Nothing pre-existing was fixed.

### DIFF
--- a/packages/extensions/src/capability/catalogue.test.ts
+++ b/packages/extensions/src/capability/catalogue.test.ts
@@ -1,0 +1,119 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * The capability catalogue is the public-facing list the review screen
+ * and the authoring prompt render — adding to it is a SemVer-relevant
+ * API change. Before this file `isKnownCapability` had exactly two
+ * repo-wide occurrences (its definition and the barrel re-export):
+ * published, and never called by a test. Replacing its body with
+ * `return true` left all 603 tests green, which is precisely the
+ * "unknown capability is red" story `risk.ts` leans on.
+ */
+
+import { describe, expect, it } from 'vitest';
+import {
+  findCatalogueEntry,
+  isKnownCapability,
+  listCapabilityCatalogue,
+} from './catalogue.js';
+import { parseCapability } from './parse.js';
+import type { Capability, CapabilityScope } from '../types.js';
+
+function cap(raw: string): Capability {
+  const r = parseCapability(raw);
+  if (!r.ok) throw new Error(`fixture "${raw}" does not parse: ${r.errors[0]?.message}`);
+  return r.value;
+}
+
+describe('listCapabilityCatalogue', () => {
+  it('returns entries and every one is well-formed', () => {
+    const entries = listCapabilityCatalogue();
+    expect(entries.length).toBeGreaterThan(0);
+    for (const e of entries) {
+      expect(typeof e.scope).toBe('string');
+      expect(e.action.length).toBeGreaterThan(0);
+      expect(e.description.length).toBeGreaterThan(0);
+      expect(['green', 'yellow', 'red']).toContain(e.baseRisk);
+      expect(typeof e.requiresTarget).toBe('boolean');
+    }
+  });
+
+  it('has no duplicate scope.action keys', () => {
+    const keys = listCapabilityCatalogue().map((e) => `${e.scope}.${e.action}`);
+    expect(new Set(keys).size).toBe(keys.length);
+  });
+});
+
+describe('isKnownCapability', () => {
+  // Measured per case, not as one "some entry matches" assertion: a
+  // single lucky row would otherwise carry the whole list.
+  it.each([
+    'model.read', 'model.mutate:Pset_*', 'model.create', 'model.delete',
+    'viewer.read', 'viewer.colorize', 'viewer.isolate', 'viewer.fly', 'viewer.section',
+    'export.create:csv', 'storage.local', 'network.fetch:example.invalid',
+    'command.invoke:ext.*', 'ui.dock', 'ui.toolbar', 'ui.contextMenu', 'ui.statusBar',
+  ])('recognises the catalogued capability "%s"', (raw) => {
+    expect(isKnownCapability(cap(raw))).toBe(true);
+  });
+
+  it.each([
+    ['model', 'destroy'],
+    ['viewer', 'mutate'],
+    ['export', 'read'],
+    ['storage', 'remote'],
+    ['network', 'listen'],
+    ['command', 'define'],
+    ['ui', 'overlay'],
+  ])('rejects the uncatalogued capability "%s.%s"', (scope, action) => {
+    expect(isKnownCapability({ raw: `${scope}.${action}`, scope: scope as CapabilityScope, action })).toBe(false);
+  });
+
+  it('does not resolve keys inherited from Object.prototype', () => {
+    // The lookup key is built from caller-controlled strings; a plain
+    // object lookup would resolve `constructor` / `toString`.
+    expect(isKnownCapability({ raw: 'model.constructor', scope: 'model', action: 'constructor' })).toBe(false);
+    expect(isKnownCapability({ raw: 'model.toString', scope: 'model', action: 'toString' })).toBe(false);
+  });
+});
+
+describe('findCatalogueEntry', () => {
+  it('returns the entry matching scope + action, ignoring the target', () => {
+    const entry = findCatalogueEntry(cap('model.mutate:Pset_WallCommon.FireRating'));
+    expect(entry).toBeDefined();
+    expect(entry?.scope).toBe('model');
+    expect(entry?.action).toBe('mutate');
+    expect(entry?.requiresTarget).toBe(true);
+  });
+
+  it('returns undefined for a capability outside the catalogue', () => {
+    expect(findCatalogueEntry({ raw: 'model.destroy', scope: 'model', action: 'destroy' })).toBeUndefined();
+  });
+
+  it('agrees with isKnownCapability on every catalogued entry', () => {
+    for (const e of listCapabilityCatalogue()) {
+      const c = { raw: `${e.scope}.${e.action}`, scope: e.scope, action: e.action };
+      expect(isKnownCapability(c)).toBe(true);
+      expect(findCatalogueEntry(c)).toBe(e);
+    }
+  });
+
+  it('marks exactly the target-taking capabilities as requiresTarget', () => {
+    const required = listCapabilityCatalogue()
+      .filter((e) => e.requiresTarget)
+      .map((e) => `${e.scope}.${e.action}`)
+      .sort();
+    expect(required).toEqual([
+      'command.invoke',
+      'export.create',
+      'model.mutate',
+      'network.fetch',
+    ]);
+  });
+
+  it('keeps model.delete and network.fetch at the red base risk', () => {
+    expect(findCatalogueEntry(cap('model.delete'))?.baseRisk).toBe('red');
+    expect(findCatalogueEntry(cap('network.fetch:example.invalid'))?.baseRisk).toBe('red');
+  });
+});
--- a/packages/extensions/src/host/sdk-version.test.ts
+++ b/packages/extensions/src/host/sdk-version.test.ts
@@ -52,6 +52,23 @@ describe('sdk-version', () => {
     expect(evaluateCompatibility('x', '>=2.0.0 <3.0.0', '3.0.0').status).toBe('outdated');
   });
 
+  // VERSION_PARSE_RE only tolerates a suffix that starts with "." or "-";
+  // junk glued straight onto a number is a typo and must fall through to
+  // `permissive` so the user is asked to confirm, rather than having a
+  // bogus range silently evaluated as if it parsed. Nothing pinned that:
+  // relaxing the suffix group from `(?:[.-].*)?` to `(?:.*)?` left the
+  // whole suite green while making "2abc" parse as 2.0.0.
+  it('does not parse a version with junk glued onto a numeric segment', () => {
+    expect(evaluateCompatibility('x', '>=2abc', '2.5.0').status).toBe('permissive');
+    expect(evaluateCompatibility('x', '>=2.0.0', '2abc').status).toBe('permissive');
+  });
+
+  it('still accepts the shorthand and prerelease forms the parser documents', () => {
+    expect(evaluateCompatibility('x', '>=2', '2.5.0').status).toBe('compatible');
+    expect(evaluateCompatibility('x', '>=2.1', '2.5.0').status).toBe('compatible');
+    expect(evaluateCompatibility('x', '>=2.1.0-rc.1', '2.5.0').status).toBe('compatible');
+  });
+
   it('findAffected returns one result per installed entry', () => {
     const results = findAffected(
       [
--- a/packages/extensions/src/manifest/contributions.test.ts
+++ b/packages/extensions/src/manifest/contributions.test.ts
@@ -1,0 +1,236 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Per-contribution-kind field requirements.
+ *
+ * The fixture suite covered exactly two contribution rules (a bad slot
+ * and a malformed `when`). Every required-field check — commands.title,
+ * keybindings.key, exporters.mimeType, statusBar.text, and the rest —
+ * could be deleted with the whole suite still green. A missing
+ * `exporters.mimeType` reaching the host means an export contribution
+ * that produces a file the browser cannot type.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { validateManifest } from './validate.js';
+
+function withContributes(contributes: unknown, entry: unknown = {}): Record<string, unknown> {
+  return {
+    manifestVersion: 1,
+    id: 'com.example.contrib',
+    name: 'Contrib',
+    description: 'Contribution fixture.',
+    version: '1.0.0',
+    engines: { ifcLiteSdk: '>=2.0.0' },
+    capabilities: ['model.read'],
+    activation: ['onStartup'],
+    contributes,
+    entry,
+  };
+}
+
+function pathsFor(contributes: unknown, entry: unknown = {}): string[] {
+  const r = validateManifest(withContributes(contributes, entry));
+  return r.ok ? [] : r.errors.map((e) => e.path);
+}
+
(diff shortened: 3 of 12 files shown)


BIMvoice:
Verified all three production findings independently. All three hold, and none is fixed here — each changes published behaviour, so the remedy is yours.

**1. `stringToEntityRef('arch:')` silently decodes to `expressId: 0`** (`packages/sdk/src/types.ts:40`).

```
stringToEntityRef("arch:") -> {"modelId":"arch","expressId":0}
```

`Number('')` is `0`, and `0` passes both guards — it is finite and not negative. So a truncated or malformed ref does not throw; it resolves to entity 0 of a real model. In a transport payload that is a silent mis-reference rather than a rejected message.

**2. The `EntityRef` codec is lossy for a `modelId` containing `:`** (`types.ts:29` encode, `:36` decode).

```
encode {"modelId":"proj:arch","expressId":42}  ->  "proj:arch:42"
decode "proj:arch:42"                          ->  THREW: Invalid expressId
```

Encode emits the separator unescaped; decode splits on the **first** colon, so the round-trip does not survive its own output. `entityRefToString(stringToEntityRef(x)) === x` is not an invariant the codec upholds, despite being the obvious contract for a serialise/parse pair. Nothing today produces a colon-bearing `modelId`, so this is latent — but it is published API and there is no validation preventing one.

**3. `VERSION_PARSE_RE`'s comment contradicts the regex** (`packages/extensions/src/host/sdk-version.ts:61`). The comment states: *"Reject 4+ dotted segments outright — silent truncation…"*. The regex is `/^(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:[.-].*)?$/`, whose trailing `(?:[.-].*)?` swallows exactly that:

```
"1.2.3"      -> matched as 1.2.3
"1.2.3.4"    -> matched as 1.2.3   (remainder swallowed)
"1.2.3.4.5"  -> matched as 1.2.3   (remainder swallowed)
">=2.0.0nope"-> op=">=" version="2.0.0nope"
```

So it does precisely the silent truncation the comment says it prevents. Either the comment or the regex is wrong, and which one is a decision about intent: if 4-segment versions should be rejected, the regex needs an anchor; if they should truncate, the comment is misleading and should say so. Prose-vs-code drift on a version gate is the kind that survives review because the comment reads as a specification.

**The root cause of the extensions gaps is one weak assertion.** `manifest.test.ts`'s invalid-fixture loop asserts only `result.ok` — the *"property that holds whether or not the code does the thing"* shape. Any rejection satisfies it, so **all twelve validator rules went unpinned**: a validator that rejected for the wrong reason, or rejected everything, passed identically. One assertion shape accounting for twelve gaps is the highest-leverage instance of that pattern found in this campaign.

Related: `canonicalContentHash`'s injectivity test targets the **retired** `0x1f`/`0x1e` scheme, and its fixtures differ in segment count — so it never touches the length prefix that replaced it. A test aimed at a design that no longer exists.

**Two things worth crediting in the method.** Two mutations had to be reshaped because the first attempt silently failed `tsc`, which would have made the entire downstream re-check vacuous — caught rather than assumed. And the downstream run measured `apps/viewer` at **703 files / 2,748 tests, all green**, which independently corroborates my own two measurements today and settles the "14 pre-existing failures" figure reported on #2168 as the unbuilt-`dist` artifact.

**One survivor left open honestly:** `ListNamespace.execute()`'s default `modelId` uses a dynamic `import()` with a variable specifier that `vi.mock` cannot intercept. Genuine gap, not closed, not argued away.

BIMvoice:
## Self-audit round 3 — three of our own tests here did not discriminate

Pushed as ccd1200…7a0f5f7. Method: minimal, syntactically-balanced source mutations applied one at a time with an exact-substring replacement asserting a replacement count of 1, the mutated region re-read to confirm, and the target test file run alone (so the measurement is whether *this* file catches it, not the package).

**Survivors — all three are our own defect, not a pre-existing one:**

|---|---|---|---|---|
| `migrations.test.ts` — rejects the non-positive / non-integer version | the integrality + positivity half of the version guard | **no** | drop `\|\| !Number.isInteger(rawVersion)` | SURVIVED |
| same | | **no** | drop `\|\| rawVersion < 1` | SURVIVED |
| `namespaces.test.ts` — gives each subscription its own unsubscribe, even for the same event | per-subscription bookkeeping | **partial** | `id = \`${event}-${this.nextId++}\`` → `` `${event}` `` | SURVIVED |
| `transport.test.ts` — does not time out a request that was already answered | `clearTimeout(entry.timer)` | **no** | delete the `clearTimeout` (BroadcastTransport) | SURVIVED |
| same, MessagePortTransport | | **no** | delete the `clearTimeout` | SURVIVED |

**Why each was vacuous**

1. **migrations** — the code alone is not a discriminator. `migrateManifest` emits `invalid_manifest_version` at path `manifestVersion` from *three* independent places: the shape guard, the newer-than-supported branch, and the no-migration-available branch at the bottom of the chain loop. With `rawVersion < 1` deleted, `0` still errors — it falls through to *"No migration available from v0 to v1"*. With `!Number.isInteger` deleted, `1.5` still errors — `1.5 > 1` hits *"newer than this loader supports"*. Same code, same path, different rule. Now asserts the shape guard's own message and a single error.

2. **namespaces** — the `offA()` half holds with a constant map key too, since each unsubscribe closes over its own `unsub`. The `removeAll()` half is what bites: with a per-event key the second `on()` overwrites the first slot, `offA()` deletes that single slot, and B's unsubscribe is never called. A listener leaks past `removeAll()`. Now drives `removeAll()` and adds a same-event case.

3. **transport** — rejecting an already-settled promise is a no-op, so advancing the clock past an answered request cannot see the missing clear. The timer *count* can. Now pins `vi.getTimerCount()` on both transports.

**Discarded as non-probative** (recorded so they are not re-walked): dropping the early `return` after a resolved response in `broadcast.ts` — no realistic message satisfies both the response shape (`id` + `result`/`error`) and the event shape (`type` + `data`), so the mutation changes nothing. Narrowed to the discrimination the file docstring actually names — `('result' in data \|\| 'error' in data)` → `('result' in data)` — which is **KILLED** by *resolves a pending request from an error-only response*. Likewise `!Number.isFinite` and `typeof rawVersion !== 'number'` in the version guard are subsumed by `!Number.isInteger` and cannot be killed by any test; that is redundancy in the production condition, not a coverage gap.

**What held.** 19 further mutations were killed: all six on `when/eval.ts` (both lookup guards, boolean ordering rejection, empty-string and zero coercion, `strictEq` on undefined), three on the capability catalogue (including `isKnownCapability` → `return true`, the one the file docstring calls out), three on `contributions.ts` slot routing, and nine on the SDK query/events namespaces (`first()` limit restore in both the present and absent case, `containedIn`/`decomposedBy` first-vs-last, the `exists` operator default, the 3-arg qset matcher, the events unsubscribe bookkeeping) plus four on the transports.

PR #2170, opened 2026-08-05
fix(sandbox): serialize eval per sandbox (#2110); contain the upstream QuickJS teardown abort (#1922)

### LINKED ISSUE #2110: Sandbox.eval does not serialize: overlapping evals on one sandbox share the log buffer and capture budget

BIMvoice:
Fixes #2110. References #1922 (contained, not fixed — the abort is upstream; evidence below).

## #2110 — `Sandbox.eval` did not serialize · **fixed**

The bridge's `logs` array, its byte total and `truncated` are built once per **sandbox** in `buildBridge` and closed over by the console functions. `eval()` resets them before each run. Two overlapping `eval()` calls therefore shared all three, and the second call's reset wiped state the first was still filling.

RED, on `main`:

```
AssertionError: expected [ 'first', 'second' ] to deeply equal [ 'second' ]
```

Two `eval()` calls started without awaiting: the second run's `ScriptResult.logs` came back carrying the first run's entry.

Each call now chains onto a per-sandbox promise queue, advanced **synchronously** on call so ordering follows call order even when callers never await between calls. The queue absorbs rejections, so a failed run does not block the runs behind it while each caller still receives its own rejection.

Of the four options weighed on the issue this is option 1, chosen because it is safe under every reading of the published contract: one sandbox owns one QuickJS realm and one capture buffer, so overlapping runs were never independent — the concurrency that serialization removes only ever produced wrong logs. Sequential callers (the viewer's `useSandbox.ts`, and every caller that awaits) see no change at all.

## #1922 — teardown abort · **contained, still upstream**

I tested the hypothesis that `setMemoryLimit()` being a no-op was the shared root cause. **Ruled out** — with a correction to the premise, detailed in the issue comment: the limit binds correctly for arrays, `ArrayBuffer` and `String.repeat`, and is unaccounted only for the `s += s` in-place concat fast path. #1922's reproducer allocates via `Array.push`, which *does* raise a clean `InternalError: out of memory`. The limit working is precisely what triggers #1922.

Every candidate in-repo remedy was then measured against the reproducer, **one trial per fresh process**, and all still abort:

| remedy | result |
|---|---|
| force a collection: 256 … 1,000,000 object literals | ABORT |
| same, memory limit lifted first | ABORT |
| large retained-then-dropped allocations (cross the collector's threshold) | ABORT |
| drain the job queue repeatedly | ABORT |
| create and free a second context | ABORT |
| skip `context.dispose()` | ABORT |

The orphaned objects carry leaked refcounts, so QuickJS's collector cannot reach them — consistent with the issue's analysis, and quickjs-emscripten 0.32 exposes no GC entry point.

**The fresh-process isolation is load-bearing, and is itself the finding that shapes the fix.** Emscripten latches `ABORT`, so only the *first* abort in a process throws; every trial after it in the same process reports a false clean. I initially measured a "working" GC nudge this way and had to throw the result out. Measured directly with two identically-poisoned runtimes: the first `dispose()` throws the assertion, the second returns cleanly.

That is the silent failure worth containing, so:

- `dispose()` throws `SandboxAbortError` — naming the condition and the issue — instead of a bare `Aborted(Assertion failed: list_empty(&rt->gc_obj_list))`.
- `isSandboxRuntimeAborted()` records it process-wide, because after the first abort the throw is no longer a signal a host can rely on.
- The sandbox is marked disposed **before** the frees, so a reuse-after-abort fails loudly rather than running against half-freed structures.
- `eval()` on a disposed sandbox now says `Sandbox disposed` instead of the unactionable `Sandbox not initialized. Call init() first.` (Both already match the viewer's existing `/Sandbox disposed|not initialized/i` guard, so no consumer change is needed.)

Nothing refuses to run on an aborted module: I measured that the module keeps executing scripts correctly afterwards in Node, so hard-failing every later sandbox would break more than the bug does.

## Tests

`packages/sandbox`: 12 files / 71 tests → **14 files / 79 tests**. All new tests were written RED first and mutation-verified (7 mutations, each killing a distinct test set):

| mutation | tests killed |
|---|---|
| bypass the eval queue | all 4 serialization |
| queue tail excludes the new run | all 4 serialization |
| drop the queue's rejection absorption | queued-after-throw |
| never mark the sandbox disposed | 3 of 4 teardown |
| rethrow the raw error instead of `SandboxAbortError` | abort-naming |
| abort flag never latches | 2 of 4 teardown |
| set `disposed` only after a successful free | abort-naming |

`@ifc-lite/sdk` (4) and `@ifc-lite/extensions` (603) pass unchanged.

## Not addressed here

The `s += s` accounting hole is real — a 268,435,456-char string built under a declared 64 MB ceiling, with `memory_used_size` reporting 86 KB. It is upstream QuickJS accounting, is not what #1922 hits, and I have not opened an issue for it; details are in the #1922 comment for @louistrue to route.

## Summary by CodeRabbit

* **New Features**
  * Sandbox evaluations now run sequentially, preventing overlapping executions from interfering with one another.
  * Added visibility into sandbox disposal and runtime-abort status.
  * Added clear sandbox-abort errors for interrupted or failed teardown operations.

* **Bug Fixes**
  * Evaluations queued before disposal now fail consistently with a clear “Sandbox disposed” message.
  * A failed evaluation no longer prevents subsequent queued evaluations from running.
  * Log limits and state are now isolated for each evaluation.


BIMvoice:
**I proposed the hypothesis this PR set out to test, and it was wrong. Correcting it here because I stated the stronger version publicly on #2166.**

I said there that "of the three documented limits, `timeoutMs` and `maxStackBytes` bind and `memoryBytes` does not". That overstates it. Re-measured against a 1 MiB ceiling, one fresh sandbox per case:

```
s += s  to 5e7 chars     -> NO ERROR   len=67108864
Array.push x 5e6         -> out of memory
new Array(1e8).fill(0)   -> out of memory
"x".repeat(5e7)          -> out of memory
new ArrayBuffer(5e7)     -> out of memory
```

The ceiling **binds** for arrays, `ArrayBuffer` and `String.repeat`. It is unaccounted only for QuickJS's in-place string-concat fast path. My original probe happened to use `s += s` — the single shape that escapes — and I generalised from one case.

**That inverts the relationship between the two issues rather than merging them.** #1922's reproducer allocates with `a.push({k:"v"})`, which raises a clean `InternalError: out of memory`. So a working memory limit is not the missing precondition for #1922 — it is the **trigger**. These are two independent problems, and the agent was right to keep them separate after testing rather than accept the framing it was handed.

**The most valuable finding here is not either issue.** Emscripten latches `ABORT`: only the *first* abort in a process throws, and every later one returns a false clean. Measured deliberately — `poisoned#1 -> ABORT`, `poisoned#2 -> CLEAN (suppressed)`. The agent had batched remedy trials into one process, found a GC nudge that appeared to work, implemented it, and got a green suite — then re-ran one trial per fresh process and every result inverted. It threw the implementation away.

That is the tenth distinct false-green mechanism in this campaign, and the most dangerous, because it does not merely hide a failure — it manufactures evidence for a fix that does not work. Anything measuring emscripten abort behaviour has to be one trial per process.

## What landed

**#2110 — fixed.** Per-sandbox promise queue with the tail advanced synchronously on call. No caller depends on concurrent evals (the viewer drives one at a time, extensions await), so implementing rather than stopping was right. RED evidence: `expected [ 'first', 'second' ] to deeply equal [ 'second' ]`.

**#1922 — contained, not fixed, and labelled as such.** Six remedy classes each still abort in fresh processes. What ships is a `SandboxAbortError` naming the issue instead of a bare emscripten assertion, an `isSandboxRuntimeAborted()` process flag, `disposed` set *before* the frees, and an accurate reuse error. It deliberately does **not** hard-fail later sandboxes, because the module still executes correctly afterwards in Node. Turning a silent death into a clear failure is worth shipping on its own.

Worth noting the agent's own mid-course correction: its first mutation run killed only 1 of 4 expected tests, so it strengthened three weak tests and re-ran rather than accepting the result.

**One loose end, flagged not filed:** the `s += s` hole lets a script park a 268,435,456-character string under a 64 MB ceiling while `memory_used_size` reports 86 KB. That is a real gap in the accounting, narrower than I first claimed but not nothing, and it is upstream in QuickJS rather than ours.

BIMvoice:
CI was red on `Check API surface`, not on tests — fixed in `e879b6f1`.

The `#1922` containment work added two exports and the snapshot was not regenerated:

```
@ifc-lite/sandbox:
   + SandboxAbortError: class
   + isSandboxRuntimeAborted: function

❌ Public API surface drifted from scripts/api-surface.json
```

Reproduced locally (it needs a full `pnpm build` first — without one the checker reports every package's declaration entry as missing, which looks like a much larger failure than it is), then `pnpm api-surface:update`. The result:

```
✅ API surface matches snapshot (37 packages, 58 export surfaces, 4008 exports)
```

`git diff` is **two added lines and nothing else** — no export removed or renamed, which is the case that guard exists to catch. The existing changeset already documents both exports, so no changeset change was needed.

Worth noting the guard did its job here. Both exports are intentional and load-bearing for the containment: `SandboxAbortError` is what turns a bare emscripten assertion into something a caller can identify, and `isSandboxRuntimeAborted()` is the process-level flag. But the check has no way to know that, and its failure message says as much — it exists because accidental export removals used to ship silently. A green surface check on a PR that adds public API would have been the less useful outcome.

BIMvoice:
**Self-review finding: the new `isDisposed` guard is checked once, before the only `await`, so it misses the exact window the PR advertises it for.**

The `disposed` getter is documented for "callers holding one across an async boundary (a React cleanup, an extension unload)" (`sandbox.ts:118-124`), and `dispose()` sets the flag before the frees so "a later eval() reports a dead sandbox instead of running against half-freed structures" (`sandbox.ts:396-399`). But `runEval` tests it exactly once:

- `sandbox.ts:208` — `if (this.isDisposed) throw ...`
- `sandbox.ts:224` — `jsCode = await transpileTypeScript(code)`  ← yields
- `sandbox.ts:243` — `const result = this.vm.evalCode(...)`  ← `this.vm` is now `null`

`dispose()` is synchronous and unqueued, so a cleanup that fires while a run is parked at `:224` blanks `this.vm` and the run resumes into a null dereference.

Swept the microtask offset at which `dispose()` lands, one cold sandbox per trial:

```
slack= 0  eval=REJ Error: Sandbox disposed. Create a new sandbox to run more scripts.
slack= 1  eval=REJ TypeError: Cannot read properties of null (reading 'evalCode')
slack= 2  eval=REJ TypeError: Cannot read properties of null (reading 'evalCode')
slack= 3+ eval=OK 1
```

A macrotask sweep shows the same shape at `ticks=0` vs `ticks>=1`. So the guard covers the pre-transpile case only; the two-microtask window in the middle still surfaces a raw `TypeError` naming an internal field, which is the message a host gets in precisely the React-cleanup scenario the getter was added for.

This is **not a regression** — the null-deref window exists on `main` too, since `dispose()` has always nulled `this.vm` across that same `await`. What the PR changes is that it introduces a guard and a documented contract for this case and then leaves the contract unmet: at `slack=0` the message improves (previously the misleading "Sandbox not initialized. Call init() first."), at `slack=1..2` nothing changes. A re-check of `isDisposed`/`this.vm` after the transpile would close it.

**What was attacked and held.** The queue itself is sound — reporting the negatives because they were the main thing under test:

- **Poisoned chain.** A rejection from *outside* `ScriptError` (a transpile failure on malformed TS) does not poison the tail; the next `eval()` returns 2 for `1 + 1`. The `run.then(noop, noop)` absorption at `sandbox.ts:191-194` holds for non-`ScriptError` rejections too, which the PR's own tests only cover for the `ScriptError` path.
- **Two callers seeing an empty tail.** The tail is advanced synchronously on call, and 5 un-awaited `eval()` calls resolve strictly in call order `[1,2,3,4,5]` with each run's logs entirely its own (`[["run1"],["run2"],...]`) — no interleave, no cross-contamination.
- **`dispose()` mid-queue leaving callers hanging.** Three calls in flight/queued, `dispose()` immediately: all three settle with the clean "Sandbox disposed" error inside 3s. Nothing hangs.
- **Unhandled rejection.** A failing `eval()` whose caller attaches its handler ~200ms later produces zero `unhandledRejection` events — the tail's own handler covers `run` without stealing the caller's rejection.
- `disposed` reads `true` and later `eval()`s reject with the disposed message.

Baseline for all of this was clean first (14 files / 79 tests passing) on a built workspace.

BIMvoice:
## Our defect: the `isDisposed` guard this PR adds has a window

Found in our own self-review of this PR.

The guard is checked once at the top of `runEval`, before the method's only `await` (the TypeScript transpile), and `this.vm` is dereferenced after it. A `dispose()` landing in that window nulls `this.vm` and the run surfaces a raw `TypeError: Cannot read properties of null (reading 'evalCode')` instead of the documented "Sandbox disposed" error.

Measured with a microtask sweep, a cold sandbox per trial:

```
slack=0    REJ Error: Sandbox disposed...                     guard works
slack=1,2  REJ TypeError: Cannot read properties of null      window
slack=3+   OK
```

The window exists on `main` too, so this is **not a regression**. What makes it this PR's business is that this PR adds the guard and documents `Sandbox.disposed` for exactly the React-cleanup case — "callers holding one across an async boundary (a React cleanup, an extension unload)" — and then leaves that case unmet. Fixing it is finishing what the PR started.

### Fix

Re-check disposal (and `vm`) immediately after the transpile, right before the dereference, rejecting with the same message as the pre-await guard. Runs queued behind the cut-off one still hit the pre-await guard and settle identically — that behaviour is now pinned by its own test rather than left implicit.

### RED evidence

Two new tests in `eval-serialization.test.ts`, before the fix:

```
 FAIL  src/eval-serialization.test.ts > reports a dispose that lands mid-run as a disposed sandbox, not a null deref
AssertionError: slack=1: expected 'TypeError: Cannot read properties of …' to match /Sandbox disposed/
+ Received: "TypeError: Cannot read properties of null (reading 'evalCode')"

 FAIL  src/eval-serialization.test.ts > settles every queued caller when dispose lands mid-queue
AssertionError: run a: expected 'TypeError: Cannot read properties of …' to match /Sandbox disposed/
+ Received: "TypeError: Cannot read properties of null (reading 'evalCode')"

 Test Files  1 failed (1)
      Tests  2 failed | 4 passed (6)
```

The offset is **swept** (0..5) rather than guessed: the window is a couple of microtasks wide and its width is transpiler-dependent, so a single offset would pin nothing. Every offset must yield either a clean result or the documented error — never a TypeError.

After: 6 passed (6). Whole package **79 → 81 passed**, 14 files, `tsc --noEmit` clean.

### Mutation proving the tests have teeth

Removed exactly the post-await re-check (exact-substring replacement, **1 replacement**, region re-read to confirm removal, restored from a pre-mutation copy):

```
replacements: 1
```

Result across the whole package — only the two new tests fail:

```
 × reports a dispose that lands mid-run as a disposed sandbox, not a null deref
 × settles every queued caller when dispose lands mid-queue
 Test Files  1 failed | 13 passed (14)
      Tests  2 failed | 79 passed (81)
```

Restored: 81 passed (81).

### Changeset

Extended the existing `sandbox-eval-serialization-and-abort-containment.md` (no second changeset). No exports changed, so `scripts/api-surface.json` is untouched.

`git diff --stat`: `3 files changed, 63 insertions(+), 1 deletion(-)`. Commit `67647e9b`.

BIMvoice:
### CodeRabbit triage — all three findings applied (2d3c43c9)

**A. `sandbox.ts:118` — reject initialization after disposal → APPLIED. CodeRabbit is right, and it caught the half of the defect the previous commit missed.**

`init()` had the identical shape to the `runEval` bug fixed in 67647e9b: one suspension point (`await getModule()`) with no disposal check on either side. A `dispose()` landing there — a React cleanup firing while `createSandbox()` is still in flight — left `init()` building a runtime, a context and a bridge on a sandbox that already read as disposed: `eval()` refuses to use them, `dispose()` has already nulled its fields and returned, so nothing will ever free them and no caller holds anything that could. Same for a plain `init()` after `dispose()`, which resolved and handed back a realm `eval()` would reject.

Guarded on both sides, before any allocation, so the rejection leaves nothing behind. The message is now one `DISPOSED_MESSAGE` constant shared by `init()` and both `runEval` checks, so the three cannot drift.

New `init-disposal.test.ts`, RED before the fix — 2 failed / 2:
```
AssertionError: promise resolved "undefined" instead of rejecting
 ❯ src/init-disposal.test.ts:51  await expect(sandbox.init()).rejects.toThrow(/Sandbox disposed/)
```

**B. `eval-serialization.test.ts:130` — prove disposal reaches queued evaluations → APPLIED. This is our own defect, in a test written in this PR, and CodeRabbit is exactly right about the shape.**

The test ticked twice and then accepted every outcome: `if (outcome !== 'ok') expect(...)`. A conditional assertion whose falsity is silently accepted asserts nothing when the timing shifts. Demonstrated rather than argued — the only change is replacing the two microtask ticks with a 300 ms wait before `dispose()`, source untouched:
```
OUTCOMES: ["ok","ok","ok","ok"]
 ✓ settles every queued caller when dispose lands mid-queue 310ms
Tests  1 passed (1)
```
Green, with zero assertions firing, on a build where disposal never reached the queue at all.

It now parks the head run at the transpile await (a one-shot gate over `transpileTypeScript`, every other transpile in the file goes to the real one), so the state under test is deterministic: run `a` suspended past its pre-await guard, `b`–`d` still queued, and the `dispose()` provably landing on all four. The assertions are unconditional — no run may report `ok`, and all four must report `Sandbox disposed`.

Mutation-verified:
- removing the pre-await `isDisposed` guard in `runEval` (1 replacement): strengthened test FAILS — `run b: expected 'Error: Sandbox not initialized. Call …' to match /Sandbox disposed/`.
- applying the same 300 ms delay to the strengthened test (1 replacement): still green with all four rejections required — it cannot be made vacuous by timing, which is the property the old one lacked.

**C. `oom-job-dispose.test.ts:107-111` — do not discard the teardown error → APPLIED.** Valid against AGENTS.md ("no silent `catch {}` — log or rethrow"), and the sibling test three cases above already binds its error. It now binds the error and asserts `SandboxAbortError` if one arrives. The assertion stays conditional on it firing at all, deliberately: by this point in the file emscripten's `ABORT` latch has usually swallowed the throw, which is the very thing this test exists to pin.

Baseline 81/81 → 83/83, `tsc --strict` clean, no export changes. The existing changeset carries the `init()` behaviour change; no second one added.

BIMvoice:
@louistrue — since twenty of our PRs are open at once, here is a merge-readiness check across the whole set, so you do not have to work out the interactions yourself. **They are mutually independent; any order works.**

## Overlap

Independently computed across all **118 changed paths** in the twenty PRs (17 onto `main`, 2 onto `feat/source-plugin-api-v2-host`, 1 onto `feat/source-dalux-provider`). Exactly **one** file is touched by more than one PR:

```
scripts/api-surface.json   <-  #2170, #2171
```

Everything else is disjoint at file level — including the pairs that looked riskiest because they share a package:

| package | PRs | shared files |
|---|---|---|
| `packages/sandbox` | #2166 (tests) / #2170 (production) | none |
| `packages/geometry` | #2161 (tests) / #2171 (production) | none |
| `packages/collab` | #2160 / #2167 | none |
| `packages/viewer` | #2154 / #2159 | none |
| `apps/viewer` | #2162 (store) / #2157, #2173 (sources) | none |

## Merges

All **272 ordered pairs** of the 17 main-based PRs were test-merged (`A`, then `B --no-commit --no-ff`) — **zero conflicts in either direction**, every merge aborted afterwards. The one shared file is clean both ways: #2170's and #2171's hunks land in different objects (`@ifc-lite/sandbox` vs `@ifc-lite/geometry`) and the result is the correctly-sorted union.

On the stacked branches, `#2157 then #2173` and `#2173 then #2157` are both clean.

## The real gates, on all 17 merged together

Textual independence is not the interesting question — this is:

```
pnpm build                     40/40 tasks
check-api-surface.mjs          PASS (37 packages, 58 export surfaces, 4021 exports)
check-test-wiring.mjs          PASS
pnpm turbo typecheck           34/34
pnpm test (root turbo)         80/80 tasks green
```

The API-surface result is the one worth noting: two PRs editing that snapshot independently produce **exactly** what a regenerate produces, so there is no drift to fix afterwards.

The one genuine semantic risk was `packages/sandbox` — three of #2166's new tests call `createSandbox` from `sandbox.ts`, the file #2170 rewrites. Merged together: 20 files / 120 tests pass. #2170's queue is transparent to sequential awaited callers, and `dispose()` still throws on an aborted runtime, only wrapped.

## Suggested order (preference, not requirement)

1. **#2170 and #2171 adjacently**, either order — the only pair sharing a file. Clean today; if one sits while other API-surface work lands, the second would need a regenerate.
2. **Production before its sibling sweep** where both touch a package: #2170 before #2166, #2171 before #2161, #2172 before other renderer work. Not required — every order was verified clean — but it means each sweep's green CI is measured against the final production code.

The stacked ones follow their bases automatically: #2157 and #2173 after #1976, #2155 after #2023.

Two caveats, stated rather than buried: #2164 moved mid-check (an audit pushed to it), so it was re-fetched and the whole matrix plus the combined build re-run against the new head; and #2165 currently reports `mergeable: UNKNOWN` on the API — that is GitHub not having computed it yet, and it merged clean locally in every test.

Issue #1810, opened 2026-07-18
Split objects by location (like construction sections / zones)


louistrue:
### Discussed in https://github.com/LTplus-AG/ifc-lite/discussions/1763

<div type='discussions-op-text'>

<sup>Originally posted by **Mildahl** July 16, 2026</sup>
### The Idea

Would love to see a way to split/assign model elements based on custom 3D zones — think construction sections, takt areas, or whatever grouping you need for planning.

Design models almost never care about how the site is actually going to be built. Walls and slabs run straight through what should be separate sections, so you can't get clean quantities per zone without a lot of manual work in Excel or similar.

### Here's roughly what I'm thinking

- You define some 3D volumes (boxes/prisms) — draw them yourself, import them, or maybe even generate them from storeys/existing spatial structure
- ifc-lite figures out which volume(s) each element falls inside
- For anything that crosses a boundary, either split it geometrically at the boundary (with each half tagged to its zone), or at least give a quantity breakdown per zone without touching the geometry
- Would be nice to run this with a few different zone sets at once (e.g. sections AND takt areas) without them interfering with each other
- End result: some kind of property on each object so you can filter/export quantities by zone afterwards

I think this would be a big deal for anyone doing construction planning, cost estimating, or site logistics off an IFC model — right now there's just no good way to bridge from "design model" to "buildable chunks" without external tools.

Anyway, just throwing the idea out there — if anything's unclear or you see it differently, happy to discuss further! 🙂</div>

louistrue:
if you have more than one model loaded together, it should be possible to slice/split across multiple models at once, rather than having to do it one model at a time!

BIMvoice:
Partial fix landed via PR #1869 (merged, `f8a3f397`) — user-defined 3D zones (`apps/viewer/src/lib/zones/`), per-element classification across all loaded/federated models, multiple independent zone sets, and a `zone` column in `@ifc-lite/lists` (filter/group/CSV-export). The PR itself opened as "Refs #1810", not "Closes", and is explicit about what's still missing — flagging so this doesn't get closed on the strength of the merge alone.

**Not delivered, per the PR's own "Known limitations" section:**
- No geometric splitting at zone boundaries — a boundary-crossing element is flagged `straddles` (with the touched zones listed) but not split.
- No pro-rated per-zone quantity breakdown for straddlers — this was one of the two explicit remedies the original issue asked for, and neither is implemented.
[...]

Issue #1155, opened 2026-06-17
fix(geometry): IfcHalfSpaceSolid clip leaks cap box → elements fly to space (Tekla IFC2X3 columns)


Blogbotana:
Some boolean-clipped elements render ~80–95 m away from their placement ("fly to space").

**Repro model:** `Building.ifc` — Tekla Structures 2023 SP1 export, IFC2X3.

**Affected elements** — 3 `IfcColumn`, profile `PK250X7.0` (IfcRectangleHollowProfileDef 250×250×7):
- `2Q$rGbapP5_gt4x$JrQuw2` (#113218) — single half-space clip
- `0hj_vQyn96HP4UrsRtHcsu` (#113509) — nested double clip
- `0giTH$CnD1g9sGm$CDhXxR` (#113536) — half-space clip

**Symptom (reproduced with the local engine, `gp.processAdaptive`):**

| | extent X | extent Y | extent Z | vertices |
|---|---|---|---|---|
| expected (RHS 250×250, H=11940) | 250 | 250 | 11940 | at placement |
| **with clip (bug)** | **97020 mm** | 11940 | 250 | up to **~95 m** from placement |
| base solid, no clip (control) | 250 | 11940 | 250 | correct ✓ |

Meshing the same base `IfcExtrudedAreaSolid` without the `IfcBooleanClippingResult` yields a correct 250×250×11940 mm column, so the fault is isolated to the half-space clipping path.

**Root cause (hypothesis):** the base solid's `IfcAxis2Placement3D` has a flipped Z axis (`(0,~0,-1)`, profile placed at `z=depth`), and the `IfcHalfSpaceSolid` plane is coincident with the host's top face (point `(125,125,11940)`, normal `(0,0,-1)` ⇒ z=11940 = top of the solid). On this degenerate config the half-space cap box (half-size ≈48510 mm, visible in the bounds) leaks into the `DIFFERENCE` result instead of trimming to the host. Only 3 of 685 clips in the model hit this combination.

**Code path:** `rust/geometry/src/processors/boolean.rs` (IfcHalfSpaceSolid clipping); the coincident-face fallback from #821 / #1137 doesn't catch this orientation.

**Minimal repro IFC** (36 entities, 2.5 KB) reproducing the 97 m smear:

```
#113218 IFCCOLUMN → #113214 IFCPRODUCTDEFINITIONSHAPE → #113213 IFCSHAPEREPRESENTATION('Body','Clipping')
[...]

PR #2171, opened 2026-08-05
feat(geometry): mesh vs. convex-volume clipping, capped so volumes add up (Refs #1810)

BIMvoice:
Refs #1810.

## What this is

The one thing #1810 needs that the codebase does not have: a way to **cut a solid with a plane and get solids back**.

`packages/clash` answers "do these two meshes touch". `packages/spatial` answers "which meshes are near this box". `apps/viewer/src/lib/zones` classifies an element's AABB centroid into a zone and flags straddlers — its own module doc says so outright (`types.ts:14-18`: "no geometry is ever split"). None of them cuts geometry, and none produces a piece whose volume can be measured.

This PR adds that primitive to `@ifc-lite/geometry`. It is pure TypeScript, has **no UI surface, no new dependency and no callers yet** — deliberately, so it can be reviewed as geometry rather than as a feature.

```ts
clipMeshByHalfSpace(mesh, plane)              // -> { inside, outside }
clipMeshByConvexVolume(mesh, planes)          // -> { inside, outside }
partitionMeshByConvexVolumes(mesh, volumes)   // -> { parts[], remainder }
planesFromOrientedBox(box)                    // viewer zone shape -> clip planes
meshVolume(mesh)                              // signed volume of a closed mesh
```

## The property that matters

Each piece is **closed**, not merely clipped: the cut cross-section is capped, so

```
sum(meshVolume(parts)) + meshVolume(remainder) === meshVolume(input)
```

That is the difference between a real per-zone quantity and an estimate, and it is asserted directly — not inferred from "the clip produced triangles". Every piece carries a `capped` flag; when the input was not watertight enough to close, it is `false` and the volume must not be quoted.

`planesFromOrientedBox` takes exactly the shape `apps/viewer/src/lib/zones` already stores (oriented box, `size` as full extents, `rotationY` only), so wiring it up later is a translation, not a redesign.

## Two bugs found while building it, both of which looked fine

Recording these because they are the failure modes a reviewer should check for:

1. **Cap winding derived from the cut direction came out inverted on both pieces.** Total volume was still conserved — the two caps cancel — so a conservation-only test passed while each individual piece's volume was wildly wrong (12.09 and −4.89 for a 7.17/0.03 split). The cap is now built from the shell's own **boundary edges** (an edge appearing in only one direction across the piece's triangles), so its winding follows the surface it closes and cannot be inverted.
2. **`v + t*(w - v)` with `t === 1` is not `w` in floating point.** A crossing that landed on an existing vertex missed it by one ulp, stopped welding, and left a sliver triangle plus an unclosable cap loop for a plane merely *tangent* to the mesh. Crossings now snap to the vertex when the plane passes through it, and are always interpolated in canonical vertex-index order so two triangles sharing an edge agree bit-for-bit.

## Tests (35)

- **Volume conservation** across every cut, and the analytically-known values: a corner tetrahedron sliced off a rotated wall (`a*b*c/6`), and a 6.0 m wall crossing two adjacent takt areas splitting into exactly 2.4 m³ / 4.8 m³.
- **Watertightness probe**: a closed mesh's signed volume is independent of the origin it is measured about; an open shell's is not. Asserted per piece.
- **Winding**: every triangle of each (convex) piece faces away from its centroid, and every cap triangle on a **non-convex L-shaped cross-section** faces the same way — which volume alone cannot see, because even a wrongly-eared triangulation sums to the correct signed area.
- **Per plane**: a triangle entirely inside, entirely outside, and straddling; a face exactly coplanar with the plane in both normal directions; tangent along a whole edge; tangent at a single vertex.
- **Degenerate**: zero-area triangles (dropped, asserted by index count, not just by volume), empty mesh, zero-length plane normal, a volume containing everything, a volume containing nothing, `cap: false` reporting `capped: false`.

Every fixture is off-origin, rotated 30° about Y and non-unit-scaled in all three extents. An axis-aligned unit box at the origin would let an implementation that returns its input pass most of this file.

The 35 tests were mutation-verified: 24 mutations run against the source, each with a printed replacement count asserted to be 1 and the region re-read. Four survivors were resolved by **deleting the redundant code** they proved was untested (a duplicate crossing clamp, a polygon de-duplication pass subsumed by vertex welding, and a dead branch in the perpendicular-vector helper) or by adding the missing test (`cap: false`, degenerate-triangle count, non-convex cap orientation). One survivor is an equivalent mutant, documented below.

## Deliberately not in this PR

- Viewer UI, interaction, or any rendering of split pieces.
- Rewiring the `zone` column in `@ifc-lite/lists`.
- Writing split geometry back into IFC as new entities.
- Normals/UVs/colours through the cut (the cut introduces new vertices; callers rebuild flat normals).
- Non-convex split volumes (decompose into convex pieces first), and cross-sections with holes — each loop is triangulated independently, which keeps the volume correct but leaves a self-overlapping cap. Both documented on the module.

## Known equivalent mutant

Classifying on-plane vertices as OUTSIDE instead of INSIDE (`s > eps` → `s >= -eps`) does not fail any test. It is genuinely equivalent given the crossing-snapping above: the extra "mixed" triangles it creates clip back to the identical polygon, and their zero-area counterparts on the other side are dropped. Flagging it rather than adding a test that would only assert an implementation detail.

## Verification

- `packages/geometry`: 260 tests passing (baseline 225, +35), 25 files.
- `pnpm build`, `pnpm turbo typecheck --filter=@ifc-lite/geometry`, `node scripts/check-test-wiring.mjs`, `node scripts/check-api-surface.mjs` all pass; the API-surface snapshot is updated in this PR.
- Changeset: minor on `@ifc-lite/geometry` (13 added exports, none removed).

## Summary by CodeRabbit

* **New Features**
  * Added mesh clipping against half-spaces and convex volumes.
  * Added mesh partitioning by multiple convex volumes, including remainder handling.
  * Added oriented-box plane generation and mesh volume calculations.
  * Added optional capping, tolerance controls, and watertightness reporting for clipped meshes.
* **Tests**
  * Added comprehensive coverage for clipping, partitioning, volume conservation, caps, degenerate inputs, and rotated boxes.


BIMvoice:
Verified the most instructive result here, because it invalidates the acceptance criterion I set for this work.

I asked for volume conservation as the correctness bar — "the union of the clipped pieces must reproduce the original volume, asserted to a stated tolerance, not just that it produced triangles". **That bar is insufficient, and this PR proves it.**

Disabling the cap-orientation correction (`const flipped = area2 < 0;` → `false`, replacements=1):

```
× triangulates the reflex cap without inverting any triangle
  Tests  1 failed | 34 passed (35)
```

**Exactly one test catches it. Every volume-conservation assertion passes.** The reason is that inverted caps cancel: the reported symptom during development was a 7.17 / 0.03 split coming back as **12.09 and −4.89** — each piece wildly wrong, a negative volume among them — while the total still reconciled. A conservation-only suite would have shipped that, and it is the precise "plausible and wrong" failure mode this campaign has been chasing, arriving in the acceptance criterion itself rather than in a fixture.

The two properties that do catch it are worth naming, because they generalise beyond this module:

- **Origin-independence of signed volume per piece.** A closed mesh's signed volume is invariant under translation; an open shell's is not. That is a watertightness probe conservation cannot give you.
- **Cap-orientation sign agreement on a non-convex cross-section.** Volume alone cannot see a wrong triangulation of an L-shaped cap — *any* ear-removal sequence sums to the correct signed area — so the test asserts every in-plane triangle's normal shares one sign.

The second real bug is a good floating-point lesson: `v + t*(w − v)` with `t === 1` is **not** `w` in IEEE754. A crossing landing exactly on an existing vertex missed by one ulp, welding stopped, and a merely *tangent* plane produced a sliver plus an unclosable loop. Fixed by snapping with canonical vertex-index-order interpolation.

**On method, three of the survivors produced code deletions rather than tests** — a redundant `t===0/1` clamp, a ring de-dup pass subsumed by welding plus the degenerate guard, and a dead perpendicular branch. A surviving mutant meaning "this code does nothing" rather than "this code is untested" is the outcome people forget to look for, and it shrinks the module instead of growing the suite. One equivalent mutant (on-plane vertices classified OUT) is acknowledged rather than papered over.

**On scope, the restraint is right.** `packages/clash` does intersection *detection* only — SAT, Möller–Trumbore, closest-point — and never constructs cut geometry; `packages/spatial` is AABB/BVH/frustum. So there was genuinely nothing to reuse, and this is the first routine in the repo that cuts a solid. It deliberately ships no UI, no callers and no wiring, and the issue comment says plainly that a user still cannot split anything in the app — which is the honest framing for step one of five.

Every fixture is off-origin (anchor `[10, 2, −4]`), rotated 30°, with non-uniform extents. Given that a fixture constant making the operation an identity has now been found in nine packages in this campaign — including one where deleting a federated id offset left all 2,750 viewer tests green — that choice is load-bearing, not cosmetic.

BIMvoice:
**Self-review finding: `capped` is returned unexamined on the zero-plane path, so an open shell comes back claiming to be a solid.**

`meshVolume` is documented as "Meaningless for an open shell — check `capped` first" (`mesh-clip.ts:142`), which makes `capped` the flag a caller is told to trust before believing a quantity. Both convex entry points seed it as a literal:

- `clipMeshByConvexVolume` — `mesh-clip.ts:330`, `capped: true`
- `partitionMeshByConvexVolumes` — `mesh-clip.ts:366`, `capped: true`

When at least one plane runs, `capCrossSection` re-derives the flag from the actual boundary edges and the seed is harmless. When the plane/volume list is **empty** — a documented, supported input ("An empty `planes` list means 'no boundaries at all'", `mesh-clip.ts:319-320`) — the loop body never executes and that literal `true` is what the caller gets.

Executed against a cube 0..1 with its `x=1` face removed (10 triangles, 4 unmatched boundary edges, `meshVolume` = 0.667 where the closed cube is 1):

```
A. clipMeshByConvexVolume(openShell, planesFromOrientedBox(...))  -> capped = false   ✅ correct
B. clipMeshByConvexVolume(openShell, [])                          -> capped = true    ❌ meshVolume = 0.6666666666666666
C. partitionMeshByConvexVolumes(openShell, [])                    -> remainder.capped = true  ❌ meshVolume = 0.6666666666666666
E. clipMeshByConvexVolume(closedCube, [])                         -> capped = true    ✅ correct (volume 1)
```

A and E are the controls: the flag is right whenever a plane runs, and right for genuinely closed input, so this is specifically the zero-plane path and not a general `capped` problem. The failure is silent — a caller that zips zone volumes with zero configured zones is handed `capped: true` on garbage and has no signal to check.

No caller exists yet, so nothing is shipping broken; it is worth closing before step 2 wires this up. The seed wants to be derived from the input rather than assumed — or the empty-list path needs the same boundary-edge check the plane path already performs.

**What was attacked and held.** Reporting these because the negative result is the more useful half:

- **Annular cross-section** (hollow square tube, outer 4×4, inner 2×2 hole, watertight, volume 48) — the shape where "inverted caps cancel" conserves volume while producing a wrong solid. Cut at `z=0`: halves 24 / 24 exactly, and **both halves audit as clean closed manifolds** (0 unmatched directed edges, 0 duplicated). The two-loop cap is handled correctly; the hole boundary is not filled in. Zone slicing the tunnel on both sides: 12 exactly. Zone cutting so the section splits into two disjoint loops: 8 exactly.
- Plane flush on a face; diagonal plane grazing an edge exactly; duplicate coincident planes; near-degenerate zones (thickness 0, 1e-12, 1e-9, 1e-6); duplicated vertex; inward-wound mesh; two disjoint solids in one mesh — volume conserved exactly in every case.
- Non-watertight input via the **half-space** path correctly reports `capped: false` on both pieces rather than silently accepting it.
- **Fuzz: 4000 randomized cases** (2000 half-space cuts, 2000 rotated box-zone clips), asserting volume conservation, no negative or source-exceeding pieces, and — the sharper check — that anything claiming `capped: true` has zero boundary edges. **Zero failures.**

The one real caveat found: `outside` from the convex path is a union of per-plane rejections, so it is non-manifold by construction where the rejected shells abut (measured 74 duplicated directed edges on one case). Volume still sums correctly and the doc does say "it may be several disconnected shells", so this is a note rather than a defect — but it means `outside` should not be fed to anything that assumes a manifold.

BIMvoice:
## Our defect: `capped` was an unexamined literal on the zero-plane path

Found in our own self-review of this PR, not reported by anyone.

`clipMeshByConvexVolume` and `partitionMeshByConvexVolumes` seeded `capped: true` and relied on `capCrossSection` to re-derive it. That works whenever a plane runs. It does not when the plane / volume list is **empty** — a documented supported input ("An empty `planes` list means 'no boundaries at all'... the whole mesh comes back as `inside`"). Nothing enters the loop, so the seeded literal is what the caller receives.

That matters because `meshVolume` is documented as meaningless until `capped` has been checked. An open shell was being handed back claiming to be a solid, and a caller trusting the flag got a silently wrong volume.

### Fix

Both zero-length paths return early with `capped` **measured** off the input, using the same boundary-edge test the rest of the module already uses: a directed edge with no opposite anywhere in the mesh is a boundary, and a mesh with a boundary is not a solid. Vertices are welded by exact coordinate first (the key `MeshBuilder` welds on), so an unindexed soup mesh with bit-identical corners is still recognised as closed — matching the module's stated limitation. Zero-area triangles are skipped for the same reason `MeshBuilder.triangle` drops them.

Deliberately **not** a flip of the literal to `false`: a genuinely closed mesh with zero planes must still report `true`, and the new test pins both directions.

### RED evidence

Two new tests against a cube-minus-one-face open shell (the wall fixture with its -X face dropped: 4 boundary edges), before the fix:

```
 FAIL  src/mesh-clip.test.ts > clipMeshByConvexVolume > derives capped from the mesh when the volume has no boundaries
AssertionError: expected true to be false // Object.is equality
 ❯ src/mesh-clip.test.ts:542:25

 FAIL  src/mesh-clip.test.ts > partitionMeshByConvexVolumes > derives the remainder capped from the mesh when there are no volumes
AssertionError: expected true to be false // Object.is equality
 ❯ src/mesh-clip.test.ts:635:25

 Test Files  1 failed (1)
      Tests  2 failed | 35 passed (37)
```

After: `Tests  37 passed (37)`. Whole package: 262 passed (25 files), `tsc --noEmit` clean.

### Mutation proving the tests have teeth

Reverted exactly the two derived values back to the literal (exact-substring replacement, **1 replacement each**, regions re-read to confirm the text changed, restored from a pre-mutation copy):

```
clip replacements: 1
partition replacements: 1
```

Result — only the two new tests fail, nothing else moves:

```
 × derives capped from the mesh when the volume has no boundaries
 × derives the remainder capped from the mesh when there are no volumes
 Test Files  1 failed (1)
      Tests  2 failed | 35 passed (37)
```

Restored: 37 passed (37).

### Also in this commit

- `ClippedMesh.capped` doc now states the flag is never asserted.
- Existing changeset extended (no second one added). No exports changed, so the API-surface snapshot is untouched.

Baseline 35/35 → 37/37 on the file. `git diff --stat`: `3 files changed, 136 insertions(+), 5 deletions(-)`. Commit `5ce299a9`.

BIMvoice:
### CodeRabbit triage — both findings applied (30606131)

**D. `mesh-clip.ts:648` — cap tolerance hard-codes `1e-9` → APPLIED. CodeRabbit is right on the substance; the suggested diff needed one change.**

Confirmed: the classification pass uses `eps` (line 232, and `crossing()` snaps onto vertices it calls on-plane at lines 261-262), while `capCrossSection` classified the same vertices at a literal `1e-9`. A caller raising `epsilon` for a sloppy mesh therefore got vertices treated as on-plane by the cut and off-plane by the cap: the boundary loop the clipper had just built was discarded as "not on the cut plane", `open` latched, and both pieces came back `capped: false` — volumes that must not be quoted, from a mesh the caller had explicitly declared that sloppy.

`eps` is threaded through as suggested, but floored: `Math.max(eps, DEFAULT_EPSILON) * (1 + Math.abs(d))`. Taken raw, `epsilon: 0` makes the cap reject the clipper's own output, because interpolated crossing points land on the plane only to within coordinate-scale rounding. That is not theoretical — it is the mutation result below.

The scale-by-coordinate-magnitude half of the comment is **not applied**. Widening the on-plane test beyond what the caller asked for makes the cap close holes that are not cuts, fabricating a lid and a volume; the new `slottedTubes` test pins exactly that direction. If a large-coordinate case that actually fails turns up, it should come with the fixture that fails.

RED first — new epsilon test failed on the unfixed source (`expected false to be true`, 1 failed / 40). Mutations, each an exact-substring replacement with the count asserted at 1:
- `tol` → `1e-9 * (1 + |d|)` (the pre-fix value): **2 failed** — "caps a cut whose boundary sits within the epsilon the caller asked for", "caps the same hole once the caller declares that much slop".
- `tol` → `1e-3 * (1 + |d|)` (over-loose): **1 failed** — "will not cap a hole that merely lies near the plane".
- `tol` → `eps * (1 + |d|)` (no floor): **1 failed** — "still closes an ordinary cut at epsilon zero".

Worth recording: the first version of the default-behaviour test survived the over-loose mutation, i.e. it asserted nothing about the direction it was written for. It was replaced by the `slottedTubes` pair above, which ties the outcome to `epsilon` in both directions.

**E. `mesh-clip.ts:336-343` — zero-length paths return the caller's own buffers → APPLIED, and the same shape exists twice.**

Confirmed. `toFloat64`/`toUint32` return the input unchanged when it is already `Float64Array`/`Uint32Array`, so `clipMeshByConvexVolume` with no planes returned the caller's arrays while every other path returns fresh buffers from `MeshBuilder.build`/`mergeMeshes` — an ownership contract that held or not depending on which array type the caller happened to pass, with an in-place edit on either side silently rewriting the other.

`partitionMeshByConvexVolumes`' zero-volume early return (line 381-393) has the identical shape and was not in the finding; both are copied now. Both are pinned by tests that assert non-identity **and** that a caller-side mutation after the call leaves the returned volume unchanged; un-copying each seed (1 replacement each) fails its own test and only its own.

Baseline 262/262 → 269/269 (mesh-clip.test.ts 37 → 44), `tsc --strict` clean. Extended the existing changeset rather than adding a second.

louistrue:
Held this one out of today's merge batch (#2154–#2174 all landed). Not a quality objection — the tests are the strongest part of it, the off-origin/rotated/non-unit-scaled fixtures are the right call, and the two bugs you recorded are exactly the ones a reviewer should be looking for. The question is where the primitive belongs.

**`rust/geometry` already cuts solids with planes.** `csg/mod.rs` has `clip_triangle(&Triangle, &Plane)` and `clip_mesh(&Mesh, &Plane)`, `processors/boolean/halfspace_cap.rs` caps the cut, and `kernel/mesh_bridge.rs` exposes `subtract`/`subtract_many` on the exact-arithmetic kernel — the one that was consolidated to a single implementation in #1024 precisely to stop this capability existing twice. A convex volume is an intersection of half-spaces, so `clipMeshByConvexVolume` is `clip_mesh` applied N times, and `meshVolume` is a signed-volume sum the kernel can already do. That makes this ~824 lines of TypeScript a second mesh-clipping kernel in a second language, with its own epsilon policy, its own cap-winding rule and its own snapping rule, which will drift from the Rust one the first time either is fixed.

The client-side constraint does not decide it — `rust/geometry` runs client-side through wasm.

So before merging I would like your call on:

1. **Rust + wasm binding instead**, reusing `clip_mesh` + `halfspace_cap` and adding whatever the zone use case needs on top (per-piece capping flag, volume). Slower to wire, one kernel.
2. **Merge as-is**, accepting a TypeScript clipper for the zone path specifically, and record why the duplication is deliberate.
3. **Hold until #1810 has a consumer** — right now there is no caller, no UI, and merging it publishes five new exports on `@ifc-lite/geometry` (minor bump) that nothing uses yet.

My read is (1), on the rust-first rule, unless there is a reason the zone split has to run without the wasm engine loaded. Also worth noting for whichever way it goes: `mesh-clip.ts` at 824 lines is over the ~400-line house rule in AGENTS.md, and it has a natural split at the `Internals` boundary already marked in the file.

Everything else about the PR is verified: it builds, typechecks and tests clean merged onto current `main`, and it was part of the integration branch that ran the full node + rust + clippy gate today.

BIMvoice:
**My call: (1), Rust.** The rust-first rule applies, your read of the duplication is correct, and the condition you attached to it does not hold — so there is no reason to except this.

But (1) is materially more work than "reuse `clip_mesh` + `halfspace_cap`", and there is one design decision inside it that I don't think should be made silently. Scoping it below, because I'd rather you commit to (1) knowing the real shape of it.

## Your condition is not met — checked, not assumed

There is no JS geometry fallback to preserve. All three geometry paths in `packages/geometry/src/index.ts` (`:327`, `:722`, `:925`) are wasm paths; the "single-threaded fallback" at `:722` means *no SharedArrayBuffer*, not *no wasm*, and it throws outright on a missing bridge. The only non-wasm alternative is the native/Tauri bridge, not a browser path.

And the precedent already exists: `apps/viewer/src/lib/space-plate-session.ts:25` imports `init` from `@ifc-lite/wasm` and idempotently initialises it for an interactive editing feature. A wasm-gated viewer feature is already sanctioned. By the time a zone set exists, the model came through wasm anyway.

So: no exception on those grounds.

## What actually exists vs. what would be written

You are right that the primitives are there — `clip_triangle` (`csg/mod.rs:239`), `clip_mesh` (`:704`), `cap_half_space_clip` (`halfspace_cap.rs:25`), `subtract`/`subtract_many` (`mesh_bridge.rs:310`/`:332`). Composition is real precedent too: `router/layers.rs:568-570` already carves a running remainder with repeated `clip_mesh` calls.

Four pieces do not:

| Piece | Status |
|---|---|
| N-convex-volume partition + remainder | Does not exist. `clip_mesh` keeps one half and discards the other (`csg/mod.rs:746-748`); `layers.rs` carves a remainder but only for **parallel** planes |
| Per-piece `capped` flag | Does not exist. `cap_half_space_clip` returns `()` and bails silently at `:118`, `:122`, `:155`. The only open-edge checks in the crate — `mesh_bridge.rs:670`, `rect_fast.rs:526` — are inside `#[cfg(test)]` |
| Public mesh volume | Exists but private: `signed_volume6` (`kernel/signed_volume.rs:29`) is `pub(crate)`, `mesh_signed_volume` (`router/voids/geom.rs:326`) is `pub(super)`, and `mesh_volume` in `mesh_bridge.rs:472` is test-only |
| **wasm surface for 3D mesh CSG** | **Does not exist at all.** `rust/wasm-bindings` exports 2D booleans (`union2d`/`difference2d`/`intersection2d`/`resolve2d`), `meshOutline2d`, and `SpacePlateHandle`. `clip_mesh`, `subtract`, `subtract_many` and every 3D boolean entry point are unexported — 3D CSG reaches JS only buried inside `IfcAPI.processGeometryBatch*`. You cannot currently hand wasm two meshes and ask for a difference |

The last row is the bulk of it: new `#[wasm_bindgen]` surface, `Mesh`↔typed-array marshalling, and a regenerated committed `pkg/ifc-lite.d.ts`, which `AGENTS.md:48` makes a CI-gated step.

None of that is an argument against (1). It is the difference between a small binding task and a real one, and I'd rather say so now than discover it halfway.

## The one decision I don't want to make unilaterally: f32

`Mesh.positions` is `Vec<f32>` (`rust/geometry/src/mesh.rs:36`). `clip_mesh` reads f32, computes in f64, writes f32 — so an N-plane convex volume does N f32 round trips on the cut vertices.

This module's output is deliberately `Float64Array`, and the reason is in its own doc:

> the whole point of this module is that volumes add up, and f32 rounding of cut vertices loses ~1e-7 relative on building-scale coordinates. Downcast at the render boundary, not here.

That is the entire premise of "capped so volumes add up". The annular-tube case splits 24 / 24 **exactly**, and the 4000-case fuzz asserts exact conservation. I do not think that survives N f32 round trips at georeferenced coordinates, and `halfspace_cap.rs:50-62` already documents the two-plane version of this hazard — independent f32 interpolation depositing ~1-ULP twins that broke the boundary chain, which is why capping had to weld on a tolerance grid.

Related: the classifier epsilon is a fixed absolute `1e-6` (`csg/mod.rs:188`), not scale-relative — and that band already caused #1155 ("a clipped column flew ~97 m").

So (1) needs a call from you: an f64 path (or an f64-carrying mesh type) for the zone clipper, or an explicit, documented decision that zone volumes are only accurate to f32-at-model-scale. Reusing `clip_mesh` as-is quietly picks the second.

The exact kernel via `intersection` + `subtract_many` (`mesh_bridge.rs:451`/`:332`) may sidestep this entirely and get the volume-safety oracle for free — probably the better route, and worth costing before committing to the `clip_mesh` composition.

## Disposition

I'm not asking you to merge this. On (1) the right outcome for this PR is that it doesn't land as TypeScript — your call whether to close it or leave it parked as the spec.

What I'd carry over rather than discard:

- **The fixtures and the failure catalogue.** The annular cross-section (the shape where inverted caps cancel and conserve volume while producing a wrong solid), plane-flush-on-face, diagonal-grazing-an-edge, duplicate coincident planes, near-degenerate thicknesses down to 1e-12, inward winding, two disjoint solids. Those port to Rust unchanged and are the part I'd least like to rewrite from memory.
- **The `capped` semantics.** Rust has no such flag today, so it gets defined from scratch — and the bug we recorded here is exactly the definition that matters: it must be *measured off boundary edges*, never seeded, because the zero-plane path is a documented supported input and a seeded literal hands the caller `capped: true` on an open shell. If the Rust version returns `bool` from `cap_half_space_clip`, the same trap is there at the "no planes ran" path.
- The `outside`-is-non-manifold caveat, which applies to any per-plane-rejection union regardless of language.

On the ~400-line rule: agreed, and the `// Internals` boundary at line 419 is where it splits. Moot if this becomes Rust, so I'll hold off unless you'd rather I split it to keep the PR reviewable as a spec.

Say which way you want it and I'll start on the Rust side — my inclination is exact-kernel `intersection`/`subtract_many` first, precisely to avoid inheriting the f32 and fixed-epsilon issues.

louistrue:
**Decision from Louis: build it in Rust and expose it through wasm — not as a second TypeScript kernel.** Closing this PR in favour of that; the work itself is good and most of it carries over.

The reasoning, so it is on the record rather than just a preference:

`rust/geometry` already cuts solids with planes and caps the cut. `csg/mod.rs:704 clip_mesh(&Mesh, &Plane)`, `csg/mod.rs:239 clip_triangle`, `processors/boolean/halfspace_cap.rs`, and `kernel/mesh_bridge.rs subtract`/`subtract_many` all sit on the exact-arithmetic kernel that #1024 consolidated to exactly one implementation, specifically to stop this capability existing twice. A convex volume is an intersection of half-spaces, so `clipMeshByConvexVolume` is `clip_mesh` applied N times, and `meshVolume` is a signed-volume sum the kernel can already do. Landing 824 lines of TypeScript alongside it creates a second kernel in a second language with its own epsilon policy, its own cap-winding rule and its own snapping rule — three things that will drift the first time either side is fixed.

Client-side is not an argument against this: `rust/geometry` runs client-side through wasm.

**What carries over, and it is most of the value:**

- The two bugs you found are real properties of any implementation and should be pinned in the Rust tests: cap winding derived from the cut direction coming out inverted on both pieces (conservation-only tests pass through it, which is exactly why it survived), and `v + t*(w - v)` with `t === 1` not being `w` in floating point.
- The test design should transfer as-is. Off-origin, rotated 30° about Y, non-unit-scaled in all three extents is the right call, and your note that an axis-aligned unit box at the origin would let a no-op implementation pass most of the file is the sharpest thing in the PR.
- `sum(volume(parts)) + volume(remainder) === volume(input)` with a per-piece `capped` flag is the right contract. Keep it.
- The equivalent mutant you flagged (on-plane vertices classified OUTSIDE vs INSIDE) is worth re-checking against the Rust epsilon policy rather than assumed to carry over.

Suggested shape: extend the kernel with a convex-volume clip that returns capped pieces plus their volumes, expose it through `wasm-bindings`, and let the zone path in #1810 call that. `planesFromOrientedBox` stays TypeScript — it is a viewer-shape adapter, not geometry.

One unrelated note for next time: `mesh-clip.ts` at 824 lines is over the ~400-line house rule in AGENTS.md, and it already marks its own split point at the `Internals` boundary.

Reopen or supersede with the Rust version whenever you want to pick it up — happy to gate it the same way.

BIMvoice:
Agreed, and that was my call too — Rust, on the rust-first rule, with the wasm-free condition you attached not holding. No objection to closing this in favour of it.

Recording the two things that carry over verbatim so they are not re-derived, plus one thing your reasoning does not address that I think has to be decided before the port starts rather than discovered inside it.

## The unresolved one: `Mesh.positions` is `Vec<f32>`

`rust/geometry/src/mesh.rs:36`. `clip_mesh` reads f32, computes in f64, writes f32 — so an N-plane convex volume does **N f32 round trips on the cut vertices**.

That is the premise of this module inverted. Its output is deliberately `Float64Array`, and its own doc says why:

> the whole point of this module is that volumes add up, and f32 rounding of cut vertices loses ~1e-7 relative on building-scale coordinates. Downcast at the render boundary, not here.

The annular-tube case splits **24 / 24 exactly**, and the 4000-case fuzz asserts exact conservation. I do not think either survives N f32 round trips at georeferenced coordinates — and this is not speculative: `halfspace_cap.rs:50-62` already documents the two-plane version, where independent f32 interpolation deposited ~1-ULP twins that broke the boundary chain, which is why capping had to weld on a tolerance grid.

Related, same area: the classifier epsilon is a fixed absolute `1e-6` (`csg/mod.rs:188`), not scale-relative — and that band already caused #1155, "a clipped column flew ~97 m".

So the port needs a decision from you, because reusing `clip_mesh` as-is quietly makes it: either an f64 path (or an f64-carrying mesh type) for the zone clipper, or an explicit, documented acceptance that zone volumes are accurate only to f32-at-model-scale. The second is a legitimate answer — it is just not one that should be made by omission.

My inclination remains the exact kernel via `intersection` + `subtract_many` (`mesh_bridge.rs:451` / `:332`) rather than composing `clip_mesh` N times, precisely to sidestep both the f32 round-tripping and the fixed epsilon, and to inherit the volume-safety oracle for free.

## What does not exist yet, so the estimate is honest

You described this as reusing `clip_mesh` + `halfspace_cap`. Four pieces are not there:

| Piece | Status |
|---|---|
| N-convex-volume partition + remainder | Does not exist. `clip_mesh` keeps one half and discards the other (`csg/mod.rs:746-748`); `layers.rs:521` carves a remainder but only for **parallel** planes |
| Per-piece `capped` flag | Does not exist. `cap_half_space_clip` returns `()` and bails silently at `:118`, `:122`, `:155`. The only open-edge checks in the crate — `mesh_bridge.rs:670`, `rect_fast.rs:526` — are `#[cfg(test)]` |
| Public mesh volume | Exists but private: `signed_volume6` is `pub(crate)`, `mesh_signed_volume` is `pub(super)`, `mesh_volume` is test-only |
| **wasm surface for 3D mesh CSG** | **Does not exist at all.** `rust/wasm-bindings` exports 2D booleans (`union2d`/`difference2d`/`intersection2d`/`resolve2d`), `meshOutline2d` and `SpacePlateHandle`. `clip_mesh`, `subtract`, `subtract_many` and every 3D boolean entry point are unexported — 3D CSG reaches JS only buried inside `IfcAPI.processGeometryBatch*` |

That last row is the bulk of it: new `#[wasm_bindgen]` surface, `Mesh`↔typed-array marshalling, and a regenerated committed `pkg/ifc-lite.d.ts`, which `AGENTS.md:48` makes a CI-gated step. None of that argues against the decision — it is the difference between a small binding task and a real one.

## Carrying over, as you listed

- **The two bugs.** Both are properties of any implementation, and the cap-winding one is the sharper: conservation-only tests pass straight through an inverted winding, which is exactly why it survived here. It will survive in Rust too unless the Rust tests assert winding directly.
- **The fixture design.** Off-origin, rotated 30° about Y, non-unit-scaled in all three extents. Also the failure catalogue — the annular cross-section (where inverted caps cancel and conserve volume while producing a wrong solid), plane-flush-on-face, diagonal-grazing-an-edge, duplicate coincident planes, near-degenerate thicknesses to 1e-12, inward winding, two disjoint solids.
- **The contract.** `sum(volume(parts)) + volume(remainder) === volume(input)` with a per-piece `capped` flag. Keeping it.
- **The `capped` semantics themselves**, which matter more in Rust than here because Rust has no such flag today and gets to define it from scratch. The bug we recorded is the definition that matters: it must be **measured off boundary edges, never seeded**, because the zero-plane case is a documented supported input and a seeded literal hands the caller `capped: true` on an open shell. If `cap_half_space_clip` grows a `bool` return, the identical trap sits at its "no planes ran" path.
- **The equivalent mutant** (on-plane vertices classified OUTSIDE vs INSIDE) — agreed, re-check against the Rust epsilon policy rather than assume it carries.
- `planesFromOrientedBox` stays TypeScript. Agreed, it is a viewer-shape adapter.

## On the 824 lines

Fair, and the file marks its own split point at the `Internals` boundary (line 419). Moot now that it becomes Rust, but taken for next time.

Happy to pick up the Rust version. Tell me which way you want the precision question decided and I will start there, since it determines whether this is `clip_mesh` composed N times or the exact kernel.

Issue #1912, opened 2026-07-29
enableInstancing is primary-model only, so federated loads get no GPU instancing


louistrue:
**Type:** enhancement
**Area:** renderer / viewer
**Reported:** by email (same reporter as #1909), split out as its own issue

## Summary

GPU instancing is switched off for every model except the primary one, so a federated
load gets no instancing at all. The cause is that the renderer's instanced-template
storage is a single global array with no model tag, unlike ThatOpen's per-model
fragments.

## Current behaviour

`apps/viewer/src/hooks/useIfcLoader.ts:1043`:

```ts
enableInstancing: target.kind === 'primary',
```

The adjacent comment (`:1039-1042`) states the constraint outright:

> GPU instancing is primary-model only (single global scene, primary id space).
> A federated load must keep all geometry flat, else its opaque repeated occurrences
> would be partitioned into shards the federated path doesn't consume and silently
> dropped.

So this is a known, deliberate guard rather than an oversight — but it means the
instancing win (see the repetition measurements in the GPU-instancing work) is
unavailable exactly where models are most repetitive.

## Why it is structural

`packages/renderer/src/scene.ts:153`:

```ts
private instancedTemplates: InstancedTemplateGPU[] = [];
```

One flat array for the whole scene, index-aligned with `instancedTemplateCpu` (`:156`).
Occurrences reference templates by bare `templateIndex` (`:1533`, `:1571`, `:3273`,
`:3381`, `:3395`, `:3404`), and `addInstancedShard` (`:2952`) pushes into the shared array
(`:3008-3019`). Nothing in that chain carries a model id, so two models' templates would
collide in one index space and `clear()` (`:3592-3597`) tears down all of them together.

The federated path additionally does not consume instanced shards at all, which is the
"silently dropped" half of the comment above.

## Proposed

1. Tag templates by model — either a per-model template array or a `modelId` on
   `InstancedTemplateGPU` plus a composite `(modelId, templateIndex)` occurrence key.
2. Teach the federated load path to consume instanced shards.
3. Scope the teardown in `clear()` so unloading one federated model does not destroy
   another's buffers.
4. Then relax the `target.kind === 'primary'` gate and re-measure.

Sizeable change touching the id space that picking, selection and the CPU-geometry
consumers all ride on — worth a design pass before implementation.

Related: #1909.

BIMvoice:
[...]

PR #2172, opened 2026-08-05
feat(renderer): model-tagged instanced-template storage (step 1 of 2 for #1912 / #2073)

BIMvoice:
## Step one of two

Instanced templates lived in a single flat, globally-positional array with no notion of which model owned them. That missing primitive is what **both** #1912 and #2073 bottom out on:

- **#1912** — `useIfcLoader.ts` sets `enableInstancing: target.kind === 'primary'`, and the comment above it says why: a federated load must keep geometry flat, "else its opaque repeated occurrences would be partitioned into shards the federated path doesn't consume and silently dropped." Nothing could say "these templates belong to model N".
- **#2073** — `scene.clear()` destroys every template and nothing re-uploads them. The two obvious patches (keep templates across clears, or retain-and-re-upload) both make *hiding* a model leave its instanced geometry on screen — again because nothing can tell which templates belong to it.

This PR lands only the primitive. **It does not flip `enableInstancing` on for federated loads and does not change `scene.clear()`'s behaviour.** See "Step two" below.

## What now exists

A template is identified by `(modelIndex, slot)` rather than a bare global position:

- `InstancedTemplateGPU.modelIndex` records the owning model. `addInstancedShard(device, shard, modelIndex = 0)` stamps it.
- `removeInstancedTemplatesForModel(modelIndex)` — the per-model teardown `scene.ts` was missing entirely. Destroys exactly that model's vertex/index/instance buffers, blanks its slots, prunes its occurrences from `instancedEntityMap`, and drops per-id selection/hidden/override bookkeeping **only** for ids that lost their last occurrence. An express id shared by two models (the federated case before id offsetting) keeps the surviving model's occurrences.
- `getInstancedModelIndices()` reports which models hold templates.
- `getInstancedTemplates()` returns a cached compacted view; `getResidentGpuBytes()` counts live templates only.

### Design note: stable slots, not a reindex

The earlier design comment on #2073 proposed a filtered clear that splices and then reindexes every surviving `InstancedOccurrence.templateIndex`, and flagged that reindex as "the place most likely to hide an off-by-one". **This PR removes the reindex entirely rather than writing it.** Removal blanks slots instead of splicing, so no surviving index ever moves; freed slots are never recycled, so a stale reference resolves to a hole and never to another model's template. Every existing reader of `instancedTemplates[i]` / `instancedTemplateCpu[i]` already guards with `?.` or `if (!x) continue` (they had to — `releaseGeometryData()` already empties the CPU array), so sparseness needed no consumer changes.

### Latent bug fixed on the way

`releaseGeometryData()` empties `instancedTemplateCpu` while the GPU slots live on. A shard uploaded afterwards `push`ed its CPU template at index 0 instead of its GPU slot, so every CPU consumer — raycast, measure, section-face, export — would read a *different* template's triangles than the occurrence named. CPU templates are now slot-assigned. Pinned by `keeps CPU templates slot-aligned after releaseGeometryData() drops them`.

## Correctness

Renderer tests: **422 → 440** (18 new), all green. `packages/renderer` runs `tsx --test`, not vitest.

Properties pinned, each with at least two models holding *different* template counts so a model tag can never be mistaken for a global array position:

- **Index stability** — removing model A leaves model B's occurrence slots unchanged, and asserted by *object identity* on the templates they resolve to, not just by index. Also holds when a third model is added after the removal.
- **Disposal exactness** — asserts on the exact GPU handles: all 6 of model A's buffers destroyed exactly once, all 9 of model B's destroyed zero times. A later `clear()` must not double-destroy the already-freed ones.
- **`getResidentGpuBytes()` accounting** — exact across add → remove → add. Templates are deliberately built at *different* sizes so a byte bug can't pass by coincidence.
- **Shared express id** — removing one model prunes only its occurrence; the id stays instanced.
- **Edge cases** — empty scene, removing a model with zero templates, removing the last model, no explicit `modelIndex` (defaults to 0).

Every new test is mutation-verified: 14 mutations (drop the model tag, splice instead of tombstone, ignore the model filter, walk the sparse array for byte accounting, delete the whole id instead of pruning occurrences, don't skip holes in `clear()`, don't free the buffers, don't refresh the live view on add / on remove, don't prune selection state, push the CPU template instead of slot-assigning, leave the CPU slot populated, …) — **all 14 killed**, each by the test intended to catch it.

The buffer allocation path needs no real GPU (`addInstancedShard` only asks the device for sized, mapped buffers), so a recording stub exercises the real interleave / bounds-fold / entity-map code and lets the tests assert on exact destroyed handles. The draw and pick passes do need a real context and stay in the browser/E2E lanes — not mocked here.

## Compatibility

Additive. `modelIndex` defaults to 0 and the repo's single `addInstancedShard` call site (`useGeometryStreaming.ts:724`) does not pass one, so today's behaviour is bit-identical. `@ifc-lite/viewer` typechecks unchanged.

## Step two (deliberately not in this PR)

Enabling federated instancing wants a maintainer's sign-off and a benchmark, so it is separate:

1. Thread the real `modelIndex` through `useGeometryStreaming`'s drain effect from `ViewportContainer`'s `modelIdToIndex` map, instead of the constant 0.
2. Apply the federated `idOffset` to instanced shard entity ids at `useIfcLoader.ts` alongside the flat-mesh offsetting, so `instancedEntityMap` stays in one id space.
3. Flip `enableInstancing` for federated targets and un-gate the render-push block — with a benchmark, since no instancing performance claim has been measured in this harness.
4. Resolve `modelIndex` for instanced picks CPU-side off the occurrence's template (`picker.ts` currently returns `modelIndex: undefined`), spending no pick-encoding bits.
5. Only then decide `scene.clear()`'s scoped variant / decoded-shard retention for #2073, including the realign-delta re-bake for the anchor-moved case (`federationRealign.ts`). `removeInstancedTemplatesForModel` is the teardown half that work needed; the retention half is a viewer-layer decision this PR does not pre-empt.

Refs #1912
Refs #2073

### DIFF
--- a/.changeset/model-tagged-instanced-templates.md
+++ b/.changeset/model-tagged-instanced-templates.md
@@ -1,0 +1,9 @@
+---
+'@ifc-lite/renderer': minor
+---
+
+Model-tagged instanced-template storage in `Scene`. `InstancedTemplateGPU` now carries a `modelIndex`, `addInstancedShard(device, shard, modelIndex = 0)` takes the owning model, and the new `removeInstancedTemplatesForModel(modelIndex)` frees exactly that model's GPU buffers and occurrences. Template slots are stable — a removal blanks its slots instead of splicing, so no other model's `templateIndex` shifts — and `getInstancedModelIndices()` reports which models hold templates. `getResidentGpuBytes()` counts live templates only.
+
+Also fixes a latent slot misalignment: after `releaseGeometryData()` a subsequently uploaded shard's CPU template landed at CPU index 0 instead of its GPU slot, so CPU consumers (raycast / measure / section / export) could read a different template's triangles than the occurrence named.
+
+Behaviour is unchanged for existing callers (`modelIndex` defaults to 0 and the single `addInstancedShard` call site does not pass one).
--- a/packages/renderer/src/scene-instanced-model-scope.test.ts
+++ b/packages/renderer/src/scene-instanced-model-scope.test.ts
@@ -159,6 +159,22 @@ describe('Scene instanced templates are model-tagged', () => {
     scene.removeInstancedTemplatesForModel(3);
     assert.deepStrictEqual([...scene.getInstancedModelIndices()], [7]);
   });
+
+  it('reports ownership even while the instanced pass is hidden (Types view)', () => {
+    // The documented difference from `getInstancedTemplates()`: hiding the
+    // pass must not change who OWNS templates, or a Types-view toggle would
+    // make a per-model teardown think the model holds nothing. Adding the
+    // `instancedVisible` early-return to `getInstancedModelIndices` survived
+    // the suite before this assertion existed (measured).
+    const { scene } = twoModelScene();
+    scene.setInstancedVisible(false);
+    assert.strictEqual(scene.getInstancedTemplates().length, 0, 'the pass is hidden');
+    assert.deepStrictEqual(
+      [...scene.getInstancedModelIndices()].sort((a, b) => a - b), [3, 7]);
+    // …and a removal taken while hidden is still model-scoped.
+    assert.strictEqual(scene.removeInstancedTemplatesForModel(3), 2);
+    assert.deepStrictEqual([...scene.getInstancedModelIndices()], [7]);
+  });
 });
 
 describe('Scene.removeInstancedTemplatesForModel — index stability', () => {
@@ -324,6 +340,33 @@ describe('Scene.removeInstancedTemplatesForModel — entity bookkeeping', () => 
 
     scene.removeInstancedTemplatesForModel(3);
     assert.deepStrictEqual([...(scene['instancedSelected'] as Set<number>)], [42]);
+  });
+
+  it('drops the HIDDEN and OVERRIDE bookkeeping too, not just the selection', () => {
+    // Selection is one of three sibling sets pruned on the same lines. With
+    // only `instancedSelected` asserted, deleting the `instancedHidden` /
+    // `instancedOverridden` lines survives the suite (measured) — and a stale
+    // entry there means a re-loaded model's occurrence of the same express id
+    // comes back invisible or wearing a dead lens colour.
+    const scene = new Scene();
+    const { device } = fakeDevice();
+    scene.addInstancedShard(device, shard(2, [42, 11]), 3);
+    scene.addInstancedShard(device, shard(3, [42, 21, 22]), 7);
+
+    // 11 exists only in model 3; 42 is shared, so it must survive all three.
+    scene.setInstancedVisibility(new Set([11, 42]), null);
+    scene.setInstancedColorOverrides(new Map<number, readonly [number, number, number, number]>([
+      [11, [1, 0, 0, 1]],
+      [42, [0, 1, 0, 1]],
+    ]));
+    assert.deepStrictEqual(
+      [...(scene['instancedHidden'] as Set<number>)].sort((a, b) => a - b), [11, 42]);
+    assert.deepStrictEqual(
+      [...(scene['instancedOverridden'] as Set<number>)].sort((a, b) => a - b), [11, 42]);
+
+    scene.removeInstancedTemplatesForModel(3);
+    assert.deepStrictEqual([...(scene['instancedHidden'] as Set<number>)], [42]);
+    assert.deepStrictEqual([...(scene['instancedOverridden'] as Set<number>)], [42]);
   });
 
   it('decrements the surviving templates\' selectedCount only for pruned occurrences', () => {
--- a/packages/renderer/src/scene-instanced-model-scope.test.ts
+++ b/packages/renderer/src/scene-instanced-model-scope.test.ts
@@ -1,0 +1,341 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import { Scene } from './scene.js';
+import type { DecodedInstancedShard } from '@ifc-lite/geometry';
+
+/**
+ * Model-scoped instanced-template storage (Refs #1912, #2073).
+ *
+ * A template is identified by `(modelIndex, templateIndex)`, where
+ * `templateIndex` is a STABLE slot in the scene's template array rather than a
+ * position in a densely-packed list. The properties pinned here are the ones a
+ * federated load and a per-model teardown both depend on:
+ *
+ *  - adding or removing one model's templates never shifts another model's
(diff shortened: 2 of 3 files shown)


BIMvoice:
Verified the latent bug this PR fixes, and it is a genuine live defect rather than a hypothetical — worth separating from the primitive itself, because it would have been worth fixing on its own.

`scene.ts:174` declares the invariant in its own comment:

> `instancedTemplateCpu` — compact CPU geometry per template (**index-aligned with `instancedTemplates`**)

`releaseGeometryData()` breaks it at `:2296` by emptying the CPU array while the GPU array keeps every entry. A shard uploaded afterwards takes `templateIndex = this.instancedTemplates.length` (`:3026`) but `push`es its CPU template onto the now-empty array (`:3050`). Reproduced with the exact shapes:

```
upload A -> 0   upload B -> 1
aligned?  A,B | A,B

after releaseGeometryData: gpu keeps 2 entries, cpu has 0
upload C -> templateIndex 2, but its CPU template landed at index 0
consumer reads instancedTemplateCpu[2] -> undefined
...while the occurrence naming template 2 = C has its triangles at cpu[0]
```

Three consumers read `instancedTemplateCpu[occ.templateIndex]` — `:1536`, `:1576`, `:3203` — so raycast, measure, section and export each resolve to a **different template's triangles** than the occurrence names, or to nothing at all. Silent wrong geometry, not a crash. Fixed here by slot-assigning rather than pushing, with a regression test.

**The prior analysis I handed over was wrong in two places, and both corrections are improvements.** It said four structs needed a model tag; it is three — `DecodedInstance` needs none, because the tag comes from the *caller* of `addInstancedShard`, not the shard bytes, so `packages/geometry` is untouched. And it called for a positional `templateIndex` reindex, describing it as "the place most likely to hide an off-by-one". That reindex was not built and is not needed: every reader already guards with `?.` or `if (!x) continue` (`:1536, 1551, 1589, 3203, 3291, 3399, 3413, 3422, 3837, 3901`) — they had to, precisely *because* `releaseGeometryData()` already emptied the array underneath them. So sparse tombstoning is safe by construction, and the riskiest part of the proposed design was removed by reading the code rather than trusting the design note.

Never-recycled slots are the right call for the same reason: a stale `templateIndex` resolves to a hole, never to another model's template. That is the failure mode the desync above actually produced.

Two things in the test design worth crediting: index stability is asserted by **object identity** rather than index equality — an index test would pass against an implementation that reshuffled and happened to land the same number — and `getResidentGpuBytes` is checked across add→remove→add with **deliberately different template sizes**, so an accounting bug cannot cancel. Every structural test uses two models with 2 and 3 templates, so no single-model fixture can make a model tag indistinguishable from a global index. That is the exact shape that hid the federated id-offset bug in `apps/viewer/src/store/globalId.ts`, where deleting the offset left all 2,750 viewer tests green.

Scope is honest: `enableInstancing` untouched, `scene.clear()` untouched, viewer untouched. Both issue comments say plainly that neither #1912 nor #2073 is fixed yet and spell out step two. The draw and pick passes need a real GL context and were left to the E2E lanes rather than mocked — the right call, given a mock there would only prove the mock.

BIMvoice:
## Round-four self-audit (production PR): 15 mutations, 11 KILLED, 4 SURVIVED — 2 real, fixed in 2ec4bc2

This is the only production PR in the round-four sample, so it got the same treatment as the test PRs: minimal mutations of the **new** `scene.ts` code, replacement count asserted `== 1`, mutated region re-read, restore from a pre-mutation copy. Baseline: `@ifc-lite/renderer` 440 passed / 0 failed.

### What held (11 of 15)

| Mutation | Count | Result |
|---|---|---|
| stamp `modelIndex: 0` instead of the argument | 1 | KILLED (5 suites) |
| `if (!t \|\| t.modelIndex !== modelIndex)` → `if (!t)` (free every model) | 1 | KILLED (5 suites) |
| skip `t.instanceBuffer.destroy()` | 1 | KILLED (disposal) |
| occurrence prune keeps everything (`filter(() => true)`) | 1 | KILLED (entity bookkeeping) |
| always delete the map entry, ignoring surviving occurrences | 1 | KILLED — this is the shared-express-id case |
| `templateIndex = liveInstancedTemplates.length` (slot recycling) | 1 | KILLED (5 suites) |
| **CPU template `push` instead of slot assignment** | 1 | KILLED (index stability) |
| drop `refreshLiveInstancedTemplates()` after removal | 1 | KILLED |
| drop it after `addInstancedShard` | 1 | KILLED (4 suites) |
| `getResidentGpuBytes` walks the sparse array | 1 | KILLED |
| `clear()` loses the `if (!it) continue` hole guard (double destroy) | 1 | KILLED |

The `push`-vs-slot row is the latent misalignment this PR fixes as a drive-by; good to see it is genuinely pinned rather than merely described.

### Survivors

| Mutation | Count | Verdict |
|---|---|---|
| delete `instancedHidden.delete(eid)` **and** `instancedOverridden.delete(eid)` | 1 | **SURVIVED — defect** |
| delete either one alone | 1 each | **SURVIVED — defect** |
| add `if (!this.instancedVisible) return []` to `getInstancedModelIndices()` | 1 | **SURVIVED — defect** |
| `getInstancedModelIndices` iterates `liveInstancedTemplates` | 1 | equivalent mutant |
| drop `lastInstancedVisibilityVersion = -1` / `instancedVisibilityDirty = true` | 1 | survived, believed equivalent |

**Defect 1.** The test is named *"drops selection/hidden/override bookkeeping only for ids that lost every occurrence"* but asserts `instancedSelected` only — one of three sibling sets pruned on adjacent lines. Deleting the other two survives, and so does deleting each one individually. A stale entry there means a re-loaded model's occurrence of the same express id returns invisible, or wearing a dead lens/IDS colour. Fixed by populating both sets through the real APIs (`setInstancedVisibility`, `setInstancedColorOverrides`) and asserting all three prune together, with the shared id 42 surviving all three.

**Defect 2.** `getInstancedModelIndices()`'s doc comment states it deliberately ignores the Types-view toggle, *unlike* `getInstancedTemplates()`. Nothing measured that: adding the early return passes. It matters because a per-model teardown taken while the pass is hidden would think the model owns nothing. Fixed, including a model-scoped removal performed while hidden.

Both are now KILLED (441/442 with the mutation applied, i.e. exactly one failing suite each).

### On the two I did **not** turn into tests

The `liveInstancedTemplates` iteration is exactly the set of non-hole entries, so that mutant cannot be distinguished — equivalent, discarded rather than reported as weakness.

The dirty-flag reset is the interesting one: I could not construct a scenario where it changes observable state. `removeInstancedTemplatesForModel` already prunes exactly the express ids that vanished from `instancedEntityMap`, so the set `setInstancedVisibility` would recompute equals the one already held, and the `changed` fast-path writes nothing either way (isolate mode, shared-id, and hide-then-remove orderings all tried). It is defensible as defensive coding, but it is currently **unfalsifiable** — worth knowing before the follow-up PR enables federated instancing and changes that.

442 renderer tests pass, typecheck clean, no production code changed by this follow-up.

PR #2173, opened 2026-08-05
refactor(sources): split SourceBrowser and useSourceCatalogSync along their seams

BIMvoice:
Offered to take or leave — this targets `feat/source-plugin-api-v2-host` and does **not** touch your branch directly. It closes the last two outstanding review threads on #1976, the two `chatgpt-codex-connector` module-size findings. Nothing is posted on #1976 itself.

It also does not touch the three files #2157 changes (`SourceProviderRow.tsx`, `SourceSettingsDialog.tsx`, `syncSourceModel.ts`), so the two can land in either order.

## First: the guidance is advisory, not CI-gated

Worth saying plainly before anyone treats these as blockers. `AGENTS.md:17` lists "split modules over ~400 non-generated lines" under **"House rules (review conventions, not linted, so self-police)"**, and names the real gates: typecheck, tests, test-wiring, wasm-type-sync, API-surface. Nothing in `.github/workflows` or `scripts/` measures TS module size. The Rust ratchet at `rust/processing/tests/module_size_ratchet.rs` is a genuine gate but references no `.ts`/`.tsx` file.

So the case for this PR is readability, not compliance. We think it stands on that alone — the seams below were already there, they just weren't file boundaries — but an "it reads fine as it is" is a perfectly good answer and nothing breaks if this sits.

We held it back earlier only because splitting these files while other #1976 fixes were still in flight would have collided with every one of them. Those fixes are now isolated in #2157, so the split can land cleanly.

## Seams

Cut along responsibility, not at line 400.

### `useSourceCatalogSync.ts` — 452 → 389

The hook was doing four jobs. Three of them are pure functions of their inputs and lift out whole:

| New module | Lines | What it owns |
|---|---|---|
| `sourceCatalogPaging.ts` | 58 | The two provider listing calls, the page constants, the `parentId` normalisation for direct children, and page appending. |
| `sourceCatalogCache.ts` | 70 | localStorage read/write, plus `isCatalogCacheable` — the capability predicate deciding whether an area's catalog can *ever* be cached. That rule was previously an inline `!flatSubtree \|\| !listFilesIsRecursive` duplicated across two callbacks; it now lives with the persistence code it constrains. |
| `sourceCatalogSelectors.ts` | 62 | The derived views: dedupe-by-id across pages, catalog completeness, map-key resolution, cursor checks. |

`folderKeyFor` and `fileKeyFor` were the same expression twice over different capability flags; they collapse into one `resolveCatalogKey(area, selected, keyedByArea)`. Same for `hasMoreFolders`/`hasMoreFiles` → `pageHasMore`, the two `folders`/`allFiles` memos → `collectUniqueById`, and the two identical `loadMore*` merge blocks → `appendPage`.

What remains in the hook is the one thing that genuinely needs to be a hook: the request lifecycle — generation counters, abort controllers, and the loading flags. The two `useCallback` fetch wrappers stay so every dependency array is unchanged.

### `SourceBrowser.tsx` — 445 → 317

The container's data wiring versus everything hanging off it:

| New module | Lines | What it owns |
|---|---|---|
| `useSourceFileSearch.ts` | 86 | The server-side search state machine: query box, whether results are displacing the folder listing, and the paged result list. Cohesive — every piece of it was only ever touched by the other pieces. |
| `useLoadedSourceModels.ts` | 118 | The bridge to models already open in the viewer: which listed files are loaded, under what names, and re-syncing them. |
| `SourceFileAreasStep.tsx` | 45 | The file-areas list. |

`useLoadedSourceModels` is the seam that pays for itself: it was the *only* reason `SourceBrowser` imported `useIfc`, `useViewerStore`, `useSourceHost`, `syncSourceModel` and `toast`. With it gone the browser is a listing UI and nothing more — it no longer reaches into the viewer's model state at all.

`SourceFileAreasStep` is smaller but fixes an asymmetry: `SourceProjectsStep` and `SourceFolderStep` were already components while the file-areas step was inline JSX, so the three sibling steps now read the same way. Deliberately kept **presentational** — its `usePagedList` state stays in the container, because moving it into the component would make it unmount on entering a file area and re-fetch the listing on the way back out. That would have been a behaviour change.

## Proof behaviour is unchanged

No existing test was modified — that was the constraint, on the reasoning that a test needing a change is evidence the split changed something.

Full `apps/viewer` suite, same machine, packages built in both runs:

| | tests | suites | pass | fail | skipped |
|---|---|---|---|---|---|
| base (`5efb03d6`) | 2244 | 507 | 2242 | 0 | 2 |
| after | 2248 | 508 | 2246 | 0 | 2 |

The delta is exactly the one new test file. The two suites covering this code (`useSourceCatalogSync.test.tsx`, `SourceFolderTree.test.ts`) run 5/5 unchanged.

`pnpm typecheck` passes (34/34 tasks) and `pnpm check:test-wiring` passes. Note that `apps/viewer` test files *are* typechecked: the root `tsconfig.json:123` excludes `**/*.test.ts`, but `apps/viewer/tsconfig.json` overrides `exclude` with only `src/lib/scripts/templates` — `tsc --showConfig` resolves 205 `.test.*` files into the viewer program.

## The one new test

`sourceCatalogSelectors.test.ts` (4 tests). The split made these functions reachable, and two of them encode rules worth pinning:

- `resolveCatalogKey` decides which listing "Load more" paginates. Swapping its two branches (`keyedByArea ? area.fileAreaId : (selected ?? area.fileAreaId)` → the reverse) — **1 replacement, test fails.**
- `isCatalogComplete` returning `size > 0 || size > 0` is what stops the "this folder has no files" empty state firing before the first listing lands. Replacing that line with `return true` — **1 replacement, test fails.**

Both mutations reverted from a backup taken before either was applied.

### DIFF
--- a/apps/viewer/src/components/sources/SourceBrowser.tsx
+++ b/apps/viewer/src/components/sources/SourceBrowser.tsx
@@ -10,19 +10,16 @@ import type {
   SourceContainer,
   SourceFile,
 } from '@ifc-lite/plugin-api';
-import { useIfc } from '@/hooks/useIfc';
-import { syncSourceModel } from '@/lib/sources/syncSourceModel';
-import { toast } from '@/components/ui/toast';
 import { loadDownloadedSourceFileRecords } from '@/lib/sources/persistence';
-import { useSourceHost } from '@/services/sources/SourceHostProvider';
-import { useViewerStore } from '@/store';
 import { useSourceCatalogSync } from './useSourceCatalogSync';
+import { useSourceFileSearch } from './useSourceFileSearch';
+import { useLoadedSourceModels } from './useLoadedSourceModels';
 import { usePagedList } from './usePagedList';
-import { SourceEntityList, LoadMoreRow } from './SourceEntityList';
 import { SourceProjectsStep } from './SourceProjectsStep';
+import { SourceFileAreasStep } from './SourceFileAreasStep';
 import { SourceFolderStep } from './SourceFolderStep';
 import { SourceBrowserHeader } from './SourceBrowserHeader';
-import { FolderOpen, AlertCircle } from 'lucide-react';
+import { AlertCircle } from 'lucide-react';
 
 interface SourceBrowserProps {
   provider: FileSourceProvider;
@@ -36,9 +33,6 @@ interface SourceBrowserProps {
 type Step = 'projects' | 'file-areas' | 'folders';
 
 export function SourceBrowser({ provider, ctx, onDownload, onBack, busy = false }: SourceBrowserProps) {
-  const sourceHost = useSourceHost();
-  const { models, addModel, removeModel } = useIfc();
-  const sourceTags = useViewerStore((s) => s.sourceTags);
   const capabilities = provider.manifest.capabilities;
   const [step, setStep] = useState<Step>('projects');
   const [selectedProject, setSelectedProject] = useState<SourceProject | null>(null);
@@ -48,13 +42,7 @@ export function SourceBrowser({ provider, ctx, onDownload, onBack, busy = false 
   // from several folders can be loaded together as one federated model.
   const [selectedFiles, setSelectedFiles] = useState<Map<string, SourceFile>>(new Map());
   const [downloadedRecords, setDownloadedRecords] = useState(() => loadDownloadedSourceFileRecords());
-  const [syncingFileIds, setSyncingFileIds] = useState<Set<string>>(new Set());
   const [error, setError] = useState<string | null>(null);
-
-  // Server-side file search (capabilities.search only).
-  const [searchQuery, setSearchQuery] = useState('');
-  const [searchActive, setSearchActive] = useState(false);
-  const searchQueryRef = useRef('');
 
   const refreshDownloadedRecords = useCallback(() => {
     setDownloadedRecords(loadDownloadedSourceFileRecords());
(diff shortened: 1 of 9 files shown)


BIMvoice:
Verified the framing this PR leads with, and it is right — worth confirming because it changes how the two threads should be read.

`AGENTS.md:17` puts the rule under **"House rules (review conventions, not linted, so self-police)"**, and the same paragraph names the actual gates: typecheck, tests, test-wiring, wasm-type-sync, API-surface. Nothing in `.github/workflows` or `scripts/` measures TypeScript module size. The Rust ratchet at `rust/processing/tests/module_size_ratchet.rs` is a real gate but references no `.ts`/`.tsx` file.

So these two codex threads are advisory, and "leave them" would have been defensible. Arguing the split on readability rather than on a rule is the honest framing.

**While confirming that, I checked a claim in the same paragraph that affects every PR we have open**, and it holds:

```
root package.json  "lint": "pnpm -r lint"
packages/* apps/*  defining a lint script:  (none)
```

`pnpm -r lint` recurses into workspaces where **no package defines `lint`**, so it succeeds having run nothing. Every one of our nineteen open PRs shows a green `Lint` check, and that green means "no linter exists", not "the code is clean". `AGENTS.md` states this outright — *"`pnpm lint` is a no-op and the CI 'Lint' job passes trivially"* — so it is documented rather than a surprise, but it is the twelfth distinct false-green mechanism this campaign has surfaced, and the only one that is repo-wide and permanent. Worth saying plainly, since a reviewer scanning check marks would reasonably read it the other way.

**On the split itself, the evidence for behaviour-preservation is the part I would look at.** Base `5efb03d6`: 2244 tests / 2242 pass / 2 skipped. After: 2248 / 2246 / 2 — a delta of exactly the one new test file, with **no existing test modified**. That last point is the load-bearing one: a refactor that had to change a test would be a refactor that changed behaviour.

The `useLoadedSourceModels.ts` seam is the one that earns its keep beyond line count — it was the sole reason `SourceBrowser` imported `useIfc`, `useViewerStore`, `useSourceHost`, `syncSourceModel` and `toast`. Extracting it turns the browser into a listing UI that never reaches into viewer model state, which is a genuine dependency reduction rather than a file-size trade. Three duplicate pairs collapsing on the way (`folderKeyFor`/`fileKeyFor`, `hasMoreFolders`/`hasMoreFiles`, the two `loadMore*` merge blocks) is the usual sign the seam was real.

Deliberately keeping `SourceFileAreasStep` presentational is the right call and worth stating: moving its `usePagedList` inside would make it re-fetch on back-navigation — a behaviour change smuggled in as a refactor.

**One process note, because it nearly produced a false baseline.** The first baseline showed **66 failures**; that was entirely an unbuilt-workspace artifact, caught only because the post-split run reported *more* tests and *zero* failures — an impossible improvement from a pure refactor. Re-baselined by restoring the base-commit files in place. That is the eleventh false result from this one mechanism in this campaign, and the tell was the direction of the anomaly rather than its size.

This avoids all three files #2157 touches, so the two can land in either order.

PR #2174, opened 2026-08-05
test(clash,export): close 29 mutation-verified gaps in the clash kernel and HBJSON geometry

BIMvoice:
Second-pass mutation sweep of the Rust crates the campaign only skimmed (`#2136`/`#2137` were a shallow early pass, before the method matured). **Test-only — no production code changed.**

## Toolchain

`rust/rust-toolchain.toml` pins bare `nightly`, which overrides the root `nightly-2025-11-15` pin for everything under `rust/`. Everything below was run on **`nightly` (rustc 1.99.0-nightly, 2026-08-01)** and re-verified on the **root-pinned `nightly-2025-11-15`**. Both green; `cargo clippy --all-targets -- -D warnings` is clean on the pinned toolchain.

## Crate survey and picks

| crate | src LOC | test fns | tests / KLOC | files >80 LOC with no `#[test]` |
|---|---|---|---|---|
| `wasm-bindings` | 9 911 | 30 | 3.0 | 20 |
| `clash` | 2 373 | 22 | 9.3 | 7 of 11 |
| `export` | 12 333 | 121 | 9.8 | 13 of 30 |
| `core` | 16 410 | 169 | 10.3 | — |
| `geometry` | 71 154 | 763 | 10.7 | — |
| `processing` | 11 978 | 162 | 13.5 | — |
| `ffi` | 634 | 10 | 15.8 | — |

Picked **`clash`** and the HBJSON half of **`export`**. `wasm-bindings` has the worst ratio but its untested surface is `#[wasm_bindgen]` glue that native tests cannot meaningfully drive; the two picked areas are pure geometric logic with load-bearing predicates and **modules with no tests at all**:

- `clash`: all 22 tests sat in one file and reached the kernel only through the session API, with **axis-aligned cubes centred on the origin**. `aabb.rs`, `bvh.rs`, `triangle.rs`, `tri_mesh.rs` had zero direct tests.
- `export`: `geom.rs` (HBJSON geometry predicates) and `adjacency.rs` (interior adjacency) had zero tests. The only thing touching them is one fixture-gated end-to-end test whose assertions — face counts per room, coordinates under 1 km — are invariant under a mirrored frame, a doubled area, or a widened orientation gate.

## Baseline vs final

| suite | before | after |
|---|---|---|
| `cargo test -p ifc-lite-clash` | 22 | **38** |
| `cargo test -p ifc-lite-export` | 120 | **137** |

Clean before and after; no pre-existing failures, no `#[ignore]`d tests.

## Mutation table

41 mutations plus 3 controls. **9 killed by the pre-existing suites — a 22 % kill ratio (TARGETED: aimed at the untested modules).** 29 of the 32 survivors are killed by this PR.

### `rust/clash` — 22 survivors / 29 mutations

| # | file | mutation | verdict |
|---|---|---|---|
| M1 | `aabb.rs` | `from_positions` guard `len() < 3` → `< 4` | SURVIVED → killed |
| M2 | `aabb.rs` | `from_positions` walk `i + 2 <` → `i + 3 <` | SURVIVED → killed |
| M3 | `bvh.rs` | split axis forced off the longest axis | SURVIVED → killed |
| M4 | `bvh.rs` | `mid = len / 2` → `(len + 1) / 2` | SURVIVED — EQUIVALENT |
| M5 | `bvh.rs` | leaf re-check `bounds[idx].intersects` dropped | SURVIVED — EQUIVALENT |
| M6 | `triangle.rs` | `EPS 1e-12` → `1e-3` | SURVIVED → killed |
| M7 | `triangle.rs` | `closest_pt_seg_seg` `t < 0` clamp writes `t = 1.0` | SURVIVED → killed |
| M8 | `triangle.rs` | `closest_pt_point_triangle` vertex-B region returns `c` | SURVIVED → killed |
| M9 | `triangle.rs` | `tri_tri_distance` b-vertex loop swaps the two witnesses | SURVIVED → killed |
| M10 | `tri_mesh.rs` | `contains_point` counts backward crossings | KILLED (`contains_point_convex_cube`) |
| M11 | `tri_mesh.rs` | barycentric bound `u + v > 1.0` → `> 2.0` | KILLED (`overlapping_cubes_hard`) |
| M12 | `tri_mesh.rs` | `distance_to_surface` returns the squared distance | KILLED (`contained_pair_reports_mesh_depth_not_aabb_gap`) |
| M13 | `tri_mesh.rs` | `vertex_centroid` divides by `n + 1` | SURVIVED → killed |
| M14 | `tri_mesh.rs` | index sanitizer `< vertex_count` → `<=` | SURVIVED → killed |
| M15 | `tri_mesh.rs` | `max_penetration_into` skips the inside test | KILLED (`contained_pair_reports_mesh_depth_not_aabb_gap`) |
| M16 | `tri_mesh.rs` | `tri_bounds` drops vertex `c` from the Z max | SURVIVED → killed |
| M17 | `narrow.rs` | `margin = tolerance.max(...)` → `.min(...)` | KILLED (`separated_cubes_clearance_hit`) |
| M18 | `narrow.rs` | small/large mesh selection inverted | SURVIVED — EQUIVALENT (not fully proven) |
| M19 | `narrow.rs` | near-contact band `dist <= tolerance` → `<` | SURVIVED — genuine gap, left open |
| M20 | `narrow.rs` | penetration loses its sign flip | SURVIVED → killed |
| M21 | `narrow.rs` | flush-overlap gate `gap < -tolerance` → `< tolerance` | SURVIVED → killed |
| M22 | `narrow.rs` | enclosure `aabb_contains` arms swapped | SURVIVED → killed |
| M23 | `narrow.rs` | touch band `min_dist <= tolerance` → `<` | SURVIVED → killed |
| M24 | `narrow.rs` | contact-bounds clamp operand order swapped | SURVIVED — EQUIVALENT |
| M25 | `session.rs` | self-clash filter `j <= i` → `j < i` | SURVIVED — EQUIVALENT |
| M26 | `session.rs` | same-element skip removed | KILLED (`cross_group_dedup_and_same_element_skip`) |
| M27 | `session.rs` | ingest range check `end <= len` → `end < len` | KILLED (8 tests) |
| M28 | `session.rs` | out-of-range global-index filter dropped | SURVIVED → killed |
| M29 | `session.rs` | dedup key not normalised to `(min, max)` | SURVIVED → killed |

### `rust/export` — 10 survivors / 12 mutations

| # | file | mutation | verdict |
|---|---|---|---|
| E1 | `geom.rs` | `polygon_area` `0.5 *` → `1.0 *` | SURVIVED → killed |
| E2 | `geom.rs` | `xf` reads the transform row-major | KILLED (`duplex_exports_valid_room_model`) |
| E3 | `geom.rs` | `clean_ring` merge test `>` → `>=` | SURVIVED → killed |
| E4 | `geom.rs` | `clean_ring` closing-vertex pop removed | KILLED (both hbjson tests) |
| E5 | `geom.rs` | `is_watertight` `c == 2` → `c >= 2` | SURVIVED → killed |
| E6 | `geom.rs` | `is_simple_polygon` `m < 3` → `m < 2` | SURVIVED → killed |
| E7 | `geom.rs` | `segments_cross` `d != 0.0` guards dropped | SURVIVED → killed |
| E8 | `geom.rs` | `collinear_overlap` `dev = tol * len` → `dev = tol` | SURVIVED → killed |
| E9 | `geom.rs` | `center` divides by `len().max(2)` | SURVIVED → killed |
| E10 | `adjacency.rs` | anti-parallel gate `> -0.95` → `> 0.95` | SURVIVED → killed |
| E11 | `adjacency.rs` | best partner picks the FARTHEST | SURVIVED → killed |
| E12 | `adjacency.rs` | `pairs.len() * 2` → `pairs.len()` | SURVIVED → killed |
| E13 | `adjacency.rs` | `MAX_GAP 0.6` → `6.0` | SURVIVED → killed (added during the fix round) |
| E14 | `adjacency.rs` | degenerate-face guard `b.len() < 3` → `< 2` | SURVIVED — EQUIVALENT |

### Controls

| control | expectation | result |
|---|---|---|
| A — `contains_point` parity `crossings & 1 == 1` → `== 0` | die | **KILLED** (10 tests) |
| B — `zup()` drops its sign flip | die | **SURVIVED → promoted to a finding, now killed** |
| C — `polygon_area` returns 0 | die | **KILLED** (both hbjson tests) |

Control B is the headline: the Y-up → Z-up conversion every exported HBJSON coordinate passes through could be turned into a **mirror** with the whole suite green. The end-to-end test asserts exactly one `Floor` and one `RoofCeiling` per room — a property a mirror preserves, since it simply swaps which face is which.

One discarded, non-probative mutation: `t_hit > RAY_EPS` → `> -RAY_EPS` in `contains_point` is near-equivalent (the two differ only on a 1e-9-wide band of ray parameters), so it was replaced by M10, which counts backward crossings for real.

## Survivor classification

**Genuine gaps (what a user would see), all now closed:**

- **Mirrored energy model** (control B). Every HBJSON coordinate flipped; Honeybee loads it and computes the wrong solar exposure.
- **`Aabb::from_positions` is a `pub` item with NO caller anywhere in the repo** — reachable only as `Aabb::from_positions` through the crate's `pub use aabb::Aabb`. Both of its boundaries were unpinned (M1, M2).
- **A malformed mesh can abort the shared wasm module** (M14, M28). `TriMesh::new`'s sanitizer accepting `index == vertex_count` panics on the next `vertex()`; `run_rule` losing its global-index filter indexes past the element vec. Both files' own comments say a panic here is unacceptable — under the release `panic = abort` profile it traps the instance that geometry, parsing and clash all share.
- **Wrong reported penetration depth** (M20): losing the negation yields `distance = -0.0`, which still satisfies the existing `assert!(rec.distance < 0.0)`.
- **Wrong nearest point / wrong witness side** (M7, M8, M9) in the routines that produce the clash report point.
- **Millimetre-scale geometry treated as degenerate** (M6): the degenerate-segment epsilon compares *squared* lengths, so a loose value silently collapses real 20 mm segments to points.
- **BVH build order** (M3): decides the order candidate pairs, and therefore clash records, come back in — user-visible in the results list.
- **Duplicate clash records** (M29) when the two rule groups overlap.
- **Shallow overlaps promoted to hard clashes** (M21) and **exact touches at tolerance 0 lost** (M23) — the latter is behaviour the code comment explicitly claims.
- **Non-manifold footprints accepted / valid ones rejected** (E5, E7, E8, E6) and **areas off by 2×** (E1) in the Honeybee validity gates.
- **Adjacency solved against the wrong face** (E10, E11, E13): perpendicular faces paired, or the farthest candidate room chosen — heat flow modelled between the wrong two zones, with the real neighbour left `Outdoors`.

**EQUIVALENT (proven):**
- **M4** — the BVH is a balanced split of a sorted array; in-order traversal yields sorted order regardless of the split index.
- **M5** — `build_node` emits single-item leaves whose `bounds` *are* `bounds[idx]`, so the leaf re-check is dead code.
- **M24** — `overlap_bounds` never returns an inverted box, so the two clamp orders commute.
- **M25** — `j == i` maps to `a_global == b_global`, which the identical-element guard already drops.
- **E14** — a sub-3-point ring has a zero Newell normal, so `dot(a.n, b.n) == 0 > -0.95` rejects it in the next test anyway.

**EQUIVALENT, not fully proven:**
- **M18** — every value the narrow phase reports (`min_dist`, `mid(closest_a, closest_b)`, `bounds_of_points`) is symmetric in the two meshes, so inverting the small/large pick should be output-identical; exact ties could in principle pick different witnesses.

**Genuine gap, left open:**
- **M19** — the near-contact accumulation band `dist <= tolerance` → `<`. The boundary needs an exact float equality between a computed triangle distance and the tolerance, which is not robustly constructible from geometry. Flagging rather than writing a fragile test.

## Repeated shapes this sweep hit again

- **A fixture constant making the operation an identity** — every clash fixture is an axis-aligned cube centred on the origin, where scaling `vertex_centroid` (M13) is the identity and two of a box face's three corners already carry the axis extreme (M16).
- **A property that holds either way** — `assert!(rec.distance < 0.0)` is satisfied by `-0.0` (M20); "exactly one Floor and one RoofCeiling per room" is preserved by a mirror (control B).
- **A boundary only ever probed from one side** — every touch test uses a 1 mm tolerance, where `<=` and `<` agree (M23).

## Other findings (no change made)

- `rust/export`'s two HBJSON tests use a **skip-if-absent** fixture loader: without staged models they `return` and the suite reports success with the entire HBJSON path unexercised. The `csv`/`gltf`/`step` tests in the same crate instead *panic* on a missing fixture (34 failures on a fresh checkout). That inconsistency is worth a decision but is out of scope here.
- `rust/wasm-bindings/tests/mesh_determinism.rs` runs **0 tests** and reports `ok` — the whole file is `wasm32`-gated.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive automated coverage for geometric calculations, collision detection, mesh validation, coordinate transforms, polygon handling, and adjacency solving.
  * Expanded validation for malformed inputs, boundary conditions, tolerances, degeneracies, and deterministic result ordering.
  * Added checks for penetration, touching, enclosure, watertightness, surface pairing, and duplicate result handling.
  * Added regression coverage for coordinate conversion consistency, scratch-buffer handling, trailing indices, and OBJ triangle winding.


BIMvoice:
The headline here is that a **control** mutation survived — one deliberately chosen to die, as a check that the harness was actually running — and it turned out to be the most user-visible finding in the sweep.

`rust/export/src/geom.rs:11` is the Y-up → Z-up conversion every HBJSON coordinate passes through:

```rust
/// Convert a WebGL Y-up point/direction to IFC/Honeybee Z-up: `(x, y, z) -> (x, -z, y)`.
pub(crate) fn zup(p: [f64; 3]) -> [f64; 3] {
    [p[0], -p[2], p[1]]
}
```

Dropping the sign turns the whole export into a **mirror**:

```
point        correct        mutated
[1,2,3]      [1,-3,2]       [1,3,2]
[0,0,5]      [0,-5,0]       [0,5,0]
[-4,1,-2]    [-4,2,1]       [-4,-2,1]
```

**Why the suite could not see it.** The end-to-end test asserts one `Floor` and one `RoofCeiling` per room. A mirror flips each face's normal, so the floor is classified as a roof *and the roof as a floor* — the two swap, and "one of each per room" stays true. The assertion is invariant under the exact transformation it was the only thing standing between us and.

For a user that is an energy model exported inside-out — geometry that loads, validates and simulates, with every surface on the wrong side. Silent, and downstream of us in someone else's tool.

That is the fourth distinct shape of "the assertion is invariant under the bug" this campaign has found, and the most consequential: previously a `deepStrictEqual` on a cache, a `toBeGreaterThan(0)` on a counter, and a lone `result.ok` that left twelve validator rules unpinned. Here it is a count that a symmetry preserves.

**Worth stating plainly: this was found only because a control was run.** The control existed to prove the harness was operating on mutated source before any survivor was believed — and it is the one that survived. Had the controls been skipped as ceremony, or swapped out when one behaved unexpectedly, this stays in the codebase.

Three other findings I would not want lost in the table:

- **`Aabb::from_positions` is `pub` with zero callers repo-wide**, and both its boundaries were unpinned. That is the sixth published-but-unexercised symbol found in this campaign; two of the previous five contained real bugs precisely because nothing called them.
- **Two paths can abort the shared wasm module**: `TriMesh::new`'s sanitizer accepting `index == vertex_count`, and `run_rule` losing its out-of-range global-index filter. Both files' own comments say a panic here is unacceptable under `panic = abort` — so the code knows the stakes and nothing was checking.
- **`assert!(rec.distance < 0.0)` is satisfied by `-0.0`**, so the hard-clash penetration could lose its sign flip undetected.

**Two things found and deliberately not changed**, both about suites reporting success while running nothing: `rust/wasm-bindings/tests/mesh_determinism.rs` runs **0 tests and reports `ok`** (entirely `wasm32`-gated), and export's HBJSON tests skip silently without fixtures while `csv`/`gltf`/`step` in the same crate panic — 34 failures on a fresh checkout from the same absence. Inconsistent, and out of scope here.

On toolchains: `rust/rust-toolchain.toml` pins bare `nightly` and overrides the root `nightly-2025-11-15` for everything under `rust/`. Everything above was run on both, and `clippy --all-targets -- -D warnings` is clean on the pinned one.

BIMvoice:
## Self-audit: one of the tests in this PR was itself vacuous. Fixed in `549e4b32f`.

Ran a mutation pass over our own additions here rather than only over the code they cover. `xf_reads_the_transform_column_major` did not test what its name and comment claimed.

### The gap

`xf` (`rust/export/src/geom.rs:17`) reads **nine** entries of the column-major 4×4:

```
c(0,0)=t[0]  c(0,1)=t[4]  c(0,3)=t[12]
c(1,0)=t[1]  c(1,1)=t[5]  c(1,3)=t[13]
c(2,0)=t[2]  c(2,1)=t[6]  c(2,3)=t[14]
```

The fixture was diagonal + translation — `t[0]=t[5]=t[10]=t[15]=1`, `t[12..14]` set — which leaves `t[1] t[2] t[4] t[6]` at zero. **Four of the nine entries the code reads were zero**, so any swap *inside* the linear block read zero either way.

### Evidence

Mutation in `rust/export/src/geom.rs:21`, exact substring, replacement count printed and asserted `n == 1`, mutated region re-read to confirm the text changed:

```diff
-        c(1, 0) * px + c(1, 1) * py + c(1, 3),
+        c(0, 1) * px + c(1, 1) * py + c(1, 3),
```

`cargo test -p ifc-lite-export geom::` → **17 passed, 0 failed**, including the test that names this mutation. Survivor.

Control (`c(0,3)` translation read + 1.0) died as expected, so the harness discriminates.

**Not an equivalent mutant.** For a 90° Z-rotation `c(1,0) = -1`, `c(0,1) = +1`. Correct `y = -px + ty`; mutated `y = +px + ty`. Every rotated entity in an HBJSON export lands at a sign-flipped Y. Only a fixture with no rotation hides it — which is exactly what the fixture was.

One correction to the original comment: the full transpose it claimed to kill *was* genuinely killed, because the translation column moves out of the read set. It was only the linear block that went unpinned.

### The fix, made decay-proof

Fixture is now a full affine with all nine read entries non-zero and pairwise distinct, and the test asserts that property **of the fixture itself** — so this cannot quietly regress to a diagonal later and re-open the same hole:

```rust
for idx in [0usize, 1, 2, 4, 5, 6, 12, 13, 14] {
    assert!(t[idx] != 0.0, "xf reads t[{idx}]; a zero entry hides a row/col swap");
}
assert_ne!(t[1], t[4], "c(1,0) must differ from c(0,1) or the swap is invisible");
```

Verified after the fix, in this order: clean baseline green (17/17) → same mutation re-applied → test **fails** (`left: [18.0, -67.0, 37.0]`, `right: [18.0, -67.0, 39.0]`) → `geom.rs` restored by file copy → green again (17/17). `module_size_ratchet` still passes 4/4.

No production code changed; the diff is 25 insertions / 4 deletions in `rust/export/src/geom_tests.rs`.

BIMvoice:
## Second finding on this branch: the Z-up→Y-up **triangle winding reversal** was completely unpinned. `41e737891`.

`frame.rs`'s module header declares winding reversal a load-bearing part of the from-bytes export contract ("mirrors `MeshDataJs::new`, keeps front faces"). It is hand-duplicated in **three** places:

- `frame::to_yup_into` (streaming/bounded GLB)
- `frame::to_yup_in_place` (in-memory GLB)
- `obj.rs`'s face loop (`f {a} {c} {b}`)

`frame.rs` had no test module at all.

### Evidence

Four exact-substring mutations, each with an asserted replacement count, deleting the reversal from all three copies simultaneously:

```
frame.rs:72   scratch.indices.swap(i + 1, i + 2);   ->  let _ = i;
frame.rs:104  indices.swap(i + 1, i + 2);           ->  let _ = i;
obj.rs:121    f {a}//{a} {c}//{c} {b}//{b}          ->  f {a}//{a} {b}//{b} {c}//{c}
obj.rs:123    f {a} {c} {b}                         ->  f {a} {b} {c}
```

`cargo test -p ifc-lite-export --lib` → **137 passed, 0 failed.** Every copy of the reversal deleted, suite fully green.

**Control:** dropping the negation in `yup_f64` dies immediately — and dies via a *pre-existing* test (`gltf::tests::occurrence_matrix_reconstructs_rotated_instance_under_national_grid_rtc`, "mis-reconstructed by 557256.6 m"). So the suite catches axis errors; it was blind specifically to winding.

### Why it hid

Two independent mechanisms, both worth recording:

1. **glTF materials are emitted `doubleSided: true` unconditionally** (`gltf.rs`, comment: IFC winding isn't reliably outward). No renderer-facing assertion can ever fail on winding.
2. **The `*_is_byte_identical` / `*_byte_deterministic` tests compare the exporter against itself**, or against a second path carrying the same bug. A defect applied consistently is invisible to a determinism test *by construction*.

Point 2 is the generalisable one, and it shaped the fix: **each copy is now pinned against a literal, not against another copy.** I first wrote a cross-path OBJ↔GLB equivalence test; it catches either single-path mutation but is blind to the symmetric one — which is precisely the mutation that matters when the same reversal is duplicated three times. It is kept, with that limitation documented in its own comment, but it is not what carries the proof.

### What landed

7 tests. Direct literal pins on `to_yup_into`, `to_yup_in_place` and the OBJ face emitter; the axis conversion itself; scratch-reuse clearing; and the trailing-partial-triangle boundary (`% 3`). The into/in-place equivalence test asserts its own fixture is **not a fixed point** of the swap, so it can't decay into vacuity.

Verified in order: baseline 144/144 green → all three copies mutated → **5 fail**, one per copy → mutations reverted by file copy → 144/144 green again. Clippy clean on the touched lines.

One process note, since it nearly bit me: my first attempt at the `obj.rs:123` mutation reported `REPLACEMENT COUNT: 2 (expected 1)` and refused to apply — my own new doc comment quotes the string `f {a} {c} {b}`. Without the count assertion that would have silently become a no-op mutation and reported the test as passing.

No production code changed.

BIMvoice:
All three CodeRabbit nitpicks verified against the code and accepted — `14d3f22d9`. One needed to go further than suggested.

**1. Test module → sibling file.** Correct, and it cites a real house rule: `geom.rs:236` in this same crate already does `#[cfg(test)] #[path = "geom_tests.rs"] mod geom_tests;`. Tests were 113 of `frame.rs`'s 223 lines. Moved to `rust/export/src/frame_tests.rs` with the MPL header.

**2. Partial-triangle boundary covered on one path only.** Correct, and the omission was squarely against this PR's own thesis — `to_yup_into` and `to_yup_in_place` each compute `tri_end` independently, and pinning one copy of a duplicated calculation says nothing about the other. That is the exact trap the file exists to close. Both paths now assert `[0, 1, 2, 3]` → `[0, 2, 1, 3]`.

**3. `mesh.indices[0..3]` can panic — right, and the suggested fix is incomplete.**

`mesh_visible` gates on `!mesh.indices.is_empty()` (`obj.rs:59`), so a visible mesh can carry one or two indices. Adding `m.indices.len() >= 3` to the `find` predicate does prevent the panic.

But the accompanying reasoning — *"earlier partial meshes emit no face through `chunks_exact(3)`, so the selected mesh still corresponds to the first emitted face"* — is only half right. It holds for the *face*, not for the *index base*. A 1–2-index visible mesh still runs the vertex loop and hits `vert_base += nverts` at `obj.rs:128`. So the first face's indices need not start at zero, and the test's `tri[0] + 1` expectation would be wrong even with the panic fixed.

The test now walks the same sequence the exporter walks — filter by `mesh_visible`, accumulate `positions.len() / 3`, stop at the first mesh with a complete triangle — and offsets by the accumulated base. It no longer assumes either property.

**Re-verified after the refactor**, because a file split can drop tests from the build without anyone touching them: all four winding mutations re-applied (counts asserted) → **139 passed, 5 failed**, same five tests as before. Clean: **144/144**. Ratchet 4/4, clippy no new hits on the touched code.

BIMvoice:
Re the CodeRabbit thread on `rust/export/src/geom_tests.rs:62` — **valid as hardening, not a current weakness.** Recording the check so it isn't re-litigated.

The claim is that only three pairwise-distinctness assertions exist, so a future fixture edit making `t[5] == t[6]` could let a `c(1,1)` → `c(2,1)` mutation pass.

Checked against the fixture as it stands: the nine read slots are `2, 5, 11, 3, 7, 13, 10, 20, 30` — already pairwise distinct, so there is no gap today. And the test's real teeth are the final exact-value assertion, `assert_eq!(xf(&t, 1.0, 2.0), [18.0, -67.0, 39.0])`, which any row/column swap changes. The `assert_ne!` lines are a guard on the *fixture*, not on `xf`.

So the finding is about future-proofing a guard, not a defect. It is cheap and self-protecting, so I'm not against it — but I'd keep the specific failure messages rather than take the suggested diff verbatim, since the generic loop loses the "or the swap is invisible" wording that tells the next reader *why* the distinctness matters.

Leaving it as-is unless you'd like it in; it does not affect what this PR pins.

louistrue:
Verified against the code rather than accepted or dismissed on reading.

The concern is that `t[5] == t[6]` would make a `c(1, 1)` → `c(2, 1)` mutation invisible. In the current fixture `t[5] = 7.0` and `t[6] = 13.0`, so I applied exactly that mutation to `rust/export/src/geom.rs` (exact-substring replacement, replacement count asserted `== 1`, region re-read from disk) and ran the suite:

```
test geom::geom_tests::xf_reads_the_transform_column_major ... FAILED
test result: FAILED. 16 passed; 1 failed
```

So the test discriminates today; the finding is about future fixture decay, not a live gap. Not blocking the merge. Worth folding the pairwise-distinct loop in on a later pass — it makes the fixture invariant self-enforcing instead of relying on three hand-picked pairs.

PR #2176, opened 2026-08-05
test(viewer): pin the map-conversion rotation cross terms (identity-fixture vacuity)

BIMvoice:

## The gap

`viewerDeltaToProjectedDelta` and `projectedDeltaToViewerDelta` (`apps/viewer/src/lib/geo/cesium-placement.ts`) apply the IfcMapConversion 2×2 grid rotation. Their only direct unit tests used the **identity fixture** `{ xAxisAbscissa: 1, xAxisOrdinate: 0 }`, which zeroes every `ordinate * …` cross term.

`pick-to-geo.test.ts` *does* exercise a real 60° rotation — but only through `viewerPointToProjected` with `deltaZ = 0`, which independently zeroes the *other* cross term. So across both files, no test ever had a non-zero rotation ordinate **and** a non-zero second delta component at the same time. The rotation was covered on paper and uncovered in fact.

## Evidence

Exact-substring mutations, replacement count printed and asserted `n == 1`, mutated line re-read. Command: `npx tsx --test src/lib/geo/cesium-placement.test.ts src/lib/geo/pick-to-geo.test.ts`.

| # | Mutation | Before | After |
|---|---|---|---|
| 1 | `eastMeters = hScale * (abscissa * deltaX `**`+`**` ordinate * deltaZ)` → `-` | 29 tests, **29 pass, 0 fail** (survivor) | 31 tests, 30 pass, **1 fail** |
| 2 | `z: (`**`ordinate`**` * eastMeters - abscissa * northMeters)` → `-ordinate` | 29 tests, **29 pass, 0 fail** (survivor) | 31 tests, 30 pass, **1 fail** |

**Control** — flipped the already-covered `- abscissa * deltaZ` term in `northMeters`: died immediately, 26 pass / **3 fail** of 29. So the suite discriminates fine; it was blind specifically to the cross terms.

**The production code is correct.** This is a coverage defect, not a behaviour defect — but an unpinned rotation is exactly where a future refactor lands a silent sign error, and a sign error here mis-places every dragged element on a rotated site grid.

## The fix, made decay-proof

Two cases using a genuine oblique rotation `(0.6, 0.8)` — a real unit vector, 3-4-5 angle — with both delta components non-zero, so every term of the matrix contributes:

- forward: `east = 0.6·1 + 0.8·2 = 2.2`, `north = 0.8·1 − 0.6·2 = −0.4`
- inverse: `(2.2, −0.4)` round-trips back to `(1, 2)`

Each asserts on the **fixture itself** — both axes non-zero and mutually distinct — so it cannot quietly decay back into the identity case it was written to replace.

One detail worth keeping: these use tolerance comparisons rather than `deepStrictEqual`, because `0.8*1 - 0.6*2` lands on `-0.3999999999999999`. An exact-equality assertion on non-dyadic fixtures is a flake waiting to happen.

Full geo suite after the change: **207 pass, 0 fail**.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for rotated grids, non-zero elevations, and scaled model anchors.
  * Improved validation of forward and inverse coordinate transformations.
  * Added round-trip reprojection and footprint checks, including projected GeoJSON coordinates and polygon closure.

### DIFF
--- a/apps/viewer/src/lib/geo/cesium-placement.test.ts
+++ b/apps/viewer/src/lib/geo/cesium-placement.test.ts
@@ -110,6 +110,13 @@ describe('cesium placement helpers', () => {
   });
 
   it('computes OrthogonalHeight from target base altitude with shift and RTC', () => {
+    // storeyElevations picks the clamp anchor (ground-floor storey) that
+    // targetBaseAltitude gets measured from — see findClampAnchorY. A
+    // storey at elevation 0 makes that term vanish from the subtraction,
+    // silently passing even if the anchor-Y term were dropped entirely.
+    // Use a genuinely non-zero elevation so the anchor term is load-bearing.
+    const storeyElevations = new Map([[1, 5]]);
+    assert.notStrictEqual(storeyElevations.get(1), 0, 'fixture must use a non-zero storey elevation');
     const orthogonalHeight = computeOrthogonalHeightForBaseAltitude({
       coordinateInfo: {
         originShift: { x: 0, y: 2, z: 0 },
@@ -126,11 +133,12 @@ describe('cesium placement helpers', () => {
       },
       projectedCRS: { mapUnitScale: 0.3048 },
       lengthUnitScale: 1,
-      storeyElevations: new Map([[1, 0]]),
+      storeyElevations,
       targetBaseAltitude: 245,
     });
 
-    assert.strictEqual(orthogonalHeight, 787.4);
+    // 245 - shiftY(2) - rtcYupY(3) - anchorY(5) = 235 meters; /0.3048 mapUnitScale.
+    assert.strictEqual(orthogonalHeight, 771);
   });
 
   it('computes the IFC origin height from OrthogonalHeight and model center', () => {
--- a/apps/viewer/src/lib/geo/cesium-placement.test.ts
+++ b/apps/viewer/src/lib/geo/cesium-placement.test.ts
@@ -179,6 +179,42 @@ describe('cesium placement helpers', () => {
     assert.deepStrictEqual(viewer, { x: 2, z: -1 });
   });
 
+  it('rotates viewer XY drag deltas by a genuine (non-identity) grid rotation', () => {
+    // A no-rotation fixture (xAxisAbscissa: 1, xAxisOrdinate: 0) zeroes the
+    // cross terms of the rotation matrix (`ordinate * deltaZ` in eastMeters,
+    // `abscissa * deltaZ` cancelling with a zero ordinate elsewhere), so a
+    // sign error in those terms is invisible to it. Use a real rotation
+    // (cos/sin of a 3-4-5 angle) with both delta components nonzero so every
+    // term of the 2x2 rotation actually contributes to the result.
+    const rotation = { xAxisAbscissa: 0.6, xAxisOrdinate: 0.8, scale: 1 };
+    // Decay-proofing: the fixture itself must stay a genuine rotation — both
+    // axes nonzero and distinct — or this test silently degrades back to the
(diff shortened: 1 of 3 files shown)


BIMvoice:
Extended with a second sweep over the rest of `apps/viewer/src/lib/geo/` — `3d0e56e39`. Same identity-fixture shape, **12 more survivors** across three files. All coverage gaps; no production defect in any of them.

### `reproject.ts` — 7 survivors

`computeFootprintGeoJSON`'s only test asserted `footprint.length === 5` and `footprint[0] === footprint[4]` (ring closure). It never asserted a corner **position**. So all of these survived at 20/20 pass: flipping a rotation sign, reordering the four corners, flipping the RTC-offset sign, flipping the ifcY axis-swap sign, and dropping the origin shift outright.

I re-verified the sharpest one myself:

```diff
-    const worldX = c.x + shift.x + rtcYup.x;
-    const worldZ = c.z + shift.z + rtcYup.z;
+    const worldX = c.x + rtcYup.x;
+    const worldZ = c.z + rtcYup.z;
```
1 replacement (asserted) → **20 tests, 20 pass, 0 fail** before; **1 fail of 22** after. The whole origin shift can be deleted and a count-and-closure assertion cannot tell.

Separately, the round-trip tests used `xAxisOrdinate: 0` **and/or** no `coordinateInfo` — either one alone zeroes the `ordinate * ifcX/ifcY` cross term, so a sign flip there was invisible from two independent directions.

### `ifc-origin.ts` — 4 survivors

`makeConversion()` hardcodes `xAxisAbscissa: 1, xAxisOrdinate: 0, scale: 1` for **every** test in the file, so the anchor-inversion algebra at `:133-139` was never genuinely exercised. Verified: flipping the sign of `ifcX`'s ordinate term (`:138`) survives **6/6**, and fails **1/7** with the new test. Also unpinned: `ifcY`'s ordinate term, the `anchorOrd * anchorOrd` term of `anchorDenom`, and the effective-scale call (scale was always 1, so hardcoding it to 1 changed nothing).

The new case gives the **anchor** a real 36.87° rotation and `scale: 2`, with the expected result hand-derived in comments rather than copied from the implementation.

### `cesium-placement.ts` — 1 survivor, and no new test needed

`storeyElevations: new Map([[1, 0]])` makes `findClampAnchorY` return exactly `0` — a storey at elevation zero is by definition closest to zero. So `anchorY` could be dropped from `targetBaseAltitude - shiftY - rtcYupY - anchorY` (`:155`) with no effect. Verified: **19/19 pass** with the term deleted.

Worth noting because it's the cheapest fix in this whole series: **changing the fixture elevation from 0 to 5 is the entire remedy.** Same 19 tests, one of which now discriminates. Expected value updated to `(245 − 2 − 3 − 5) / 0.3048 = 771`.

### Verification

Every fixture now asserts its own discriminating property — ordinate non-zero, scale non-unit, shift and RTC non-zero, storey elevation non-zero — so none can decay back into the identity case it replaces. Float comparisons use tolerances, not strict equality.

This commit was rebased onto the first one, so I re-ran the mutations **after** the rebase as well as before — a rebase can drain assertions without touching a test file. Post-rebase: 4 representative mutations → **206 pass / 4 fail**; restored → **210 pass / 0 fail**.

One negative result worth recording: `apps/viewer/src/hooks/dxfExportGeoref.test.ts` was also probed with 10 mutations (rotation signs both directions, y-flip, `Scale = 0` fallback, `mapUnitScale` drop, `finiteOr` fallback, axis-length threshold) and **all 10 died on first try**. That file is already properly hardened from the #1871/#1965 rounds — no changes, and it is not in this diff.

Issue #1915, opened 2026-07-29
Export Changes review


KamranMosleh:
### The problem

After any edit in editing mode the button for exporting them shows the number of changes, but they are not reviewable. If there was a possibility to review these changes it could help for comparison.

### Proposed solution

A double step for export with a review at first step.

PR #2178, opened 2026-08-05
test(collab,mutations): pin three orderings that commuting fixtures hid

BIMvoice:

A different vacuity shape from the identity-fixture rounds. Here the fixtures make the operations **commute**, so the ordering under test is unobservable. All three findings are coverage gaps rather than live defects — the current concrete call sites happen not to expose them — but all three sit in code whose entire job is to enforce an order.

## 1. `collab/src/branch/branch-tree.ts:58`

`buildBranchTree` re-sorts each branch's entries by `at` before chaining history edges. All three existing tests built their fixtures by calling `sidecar.record()` sequentially, and `MemoryHistorySidecar.record()` stamps `at: new Date().toISOString()` — so insertion order *always* equalled `at` order.

Deleting the sort line outright (1 replacement, asserted): **3 pass / 3 pass**, unchanged. With the new test: **1 failed / 3 passed of 4**.

Exposure today is genuinely low, and I want to be straight about that: the only concrete `HistorySidecar` reachable through this path returns `entries()` already globally `at`-sorted, so per-branch order survives as a side effect. The interface only promises "oldest first" though, and finding 2 below is the proof that the assumption isn't free. The new test uses a hand-built sidecar returning `[newer, older]`.

## 2. `collab/src/branch/history-automerge.ts:142`

Same shape, but this one **is** a shipped `HistorySidecar` implementation, not test scaffolding. `entries()` sorts `Object.values(shape.entries)` by `at`; the single existing test recorded `a` then `b` in real time, so Automerge's insertion-ordered `Object.values` already matched.

Removing the sort (1 replacement, asserted): **4 pass / 4 pass**, unchanged. With the new test: **1 failed / 4 passed of 5**.

The new case uses fake timers to record the chronologically *later* entry first, then winds the clock backward and records the *earlier* one — so insertion order deliberately diverges from `at` order. It also asserts `later.at > earlier.at`, so the fixture can't quietly become pre-sorted again.

## 3. `mutations/src/change-set.ts` — `ChangeSetManager` had zero tests

`ChangeSetManager` is exported from `index.ts` but appeared only in its own source, the barrel, and the README — no test file referenced it. I confirmed that against `HEAD` rather than taking it on trust.

Two shaped lines were therefore unreachable from any test:
- `mergeChangeSets`'s `mutations.sort((a, b) => a.timestamp - b.timestamp)`
- `deleteChangeSet`'s `if (this.activeChangeSetId === id)` guard

New file with 7 tests. Deleting the sort → `expected [ 'late300', 'early100', 'mid200' ] to deeply equal [ 'early100', 'mid200', 'late300' ]`, **1 failed of 7**. Replacing the guard with an unconditional clear → **1 failed of 7** (deleting a non-active change set wrongly cleared the active pointer).

The merge fixture deliberately merges a change set whose single mutation is at timestamp 300 *before* one holding 100 and 200, and asserts the concatenation is **not already sorted** before checking the result — so it cannot decay into a commuting case.

## Totals

`packages/collab` 175 pass (was 173). `packages/mutations` 137 pass (was 130).

## Negative results

`packages/mutations/src/effective-changes.ts:356` was read in full and left alone — it was extracted specifically to be independently testable (issue #1915) and shows deliberate design for it. Sort call sites at `collab/src/awareness/render.ts:96` and `collab/src/perf/latency.ts:102` were noted but not probed; they remain open candidates for a later round rather than anything I'm claiming about here.


BIMvoice:
Extended with the three sort sites the first commit flagged but did not probe — `17c0e31e3`. **Four more survivors**, all coverage gaps rather than live defects. One result corrects the first commit's own suspicion.

### `collab/src/awareness/render.ts` — 2 survivors

Comparator sign flip on `peerVisuals`' `out.sort((a, b) => a.clientId - b.clientId)`: **4/4 pass** before, 1 fail of 6 after. Every existing test used `.find()` (order-blind) or produced 0–1 results.

There's a subtlety worth recording. `PresenceMap` is keyed by numeric clientId, and JS always iterates integer-like keys in ascending numeric order regardless of literal or insertion order — so the *input* order to this sort can never be controlled from a test. That doesn't make it untestable: `.sort()` recomputes order from scratch rather than deferring to its input, so an assertion on the **output** order still discriminates a broken comparator. The new test says so explicitly, since the natural reading is that the test is vacuous.

Second: `idleMs >= staleAfterMs` → `>` survived **5/5**. The stale fixture sits at `idleMs = 30_000` against a `10_000` threshold — far past the boundary in both directions, so the exact-equality semantics were never pinned.

Control (`clientId === exclude` → `!==`) died 5/5, so the harness discriminates.

### `collab/src/perf/latency.ts` — 1 survivor

`flushUntil`'s `queue.sort((x, y) => x.arrivesAt - y.arrivesAt)` sign flip: **3/3 pass** before, 1 fail of 4 after. Every test used uniform `baseMs` with no jitter, so every queued item tied on `arrivesAt` — and reversing a comparator over an all-equal array is a genuine no-op.

The interesting part is what *doesn't* work. The obvious fix — fire several updates from one doc with scripted jitter — is a trap: Yjs enforces per-client causal delivery, so a later-clock update from a client silently buffers until the earlier one integrates. That masks the sort bug behind Yjs's own gap-filling instead of exercising it. The committed test sources one update from each of two docs — independent client clocks, no causal coupling — with jitter forcing opposite-order arrival times, so which is due at a half-way flush depends entirely on the sort.

### `mutations/src/effective-changes.ts` — 1 survivor, and a correction

The first commit listed this file as a suspect on the grounds that its "designed to be testable" provenance had been assumed rather than verified. Mutating **each of the comparator's four keys separately** shows that suspicion was mostly wrong, and I'd rather say so than let it stand:

| key | mutation result against existing tests |
|---|---|
| `entityId` | **3 of 43 fail** — already pinned |
| `kind` | **6 of 43 fail** — already pinned |
| `name` | **1 of 43 fail** — already pinned |
| `setName` | **43/43 pass — survivor** |

I re-ran the `name` flip myself to confirm it really does die (1 fail of 43), because a report claiming three of four keys were already covered is exactly the kind of convenient result worth double-checking.

Only the fourth tiebreak was unpinned, and the reason is structural: reaching it needs two rows tying on `entityId` + `kind` + `name`, which only happens with two psets or qsets on the same entity, since `name` is `undefined` for those rows. No fixture had that shape. Deleting the sort call entirely also dies (7 of 44), as a sanity check.

The new fixture inserts `Zeta` before `Alpha` so an identity sort cannot produce the right answer by accident.

### Totals

`packages/collab` 176 pass (was 173), `packages/mutations` 131 pass (was 130).

This commit was rebased onto the first, so the mutations were re-run **after** the rebase: 2 fail of 10 (collab) and 1 fail of 44 (mutations) with the source mutated, all green restored.

One process note: the collab suite initially reported *17 failed test files* here, which reads like a serious regression. It was `@ifc-lite/ifcx` not being built in the worktree — vitest reports an unresolvable import as a failed file. Building the workspace deps returns it to 41/41. Flagging because that artifact has produced false failure reports for us before.

louistrue:

**How the conflict was resolved:** packages/mutations/test/change-set.test.ts was a pure add/add conflict: this branch and #2165 (already on main) each created the first-ever test file for ChangeSetManager, so there was no shared ancestor version. I resolved it by unioning both sweeps under one import block: main's `mutation(overrides)` fixture builder is kept verbatim, and the PR side's incompatible positional builder is kept alongside it renamed to `timestampedMutation(id, timestamp, entityId)` (the two signatures cannot be collapsed into one helper). Main's six nested describes are untouched; the PR's seven tests moved into a new nested describe "ordering fixtures that disagree with insertion order" and now call the renamed helper. No `it()` name collided between the two sides, so no test needed a disambiguating suffix, and no test body was altered other than the helper call rename. I proved the merged declaration list is a superset of both sides by `comm`-diffing sorted name lists in both directions (both empty), and ran the tests: the merged file is 26 passing tests = exactly 7 (PR) + 19 (main).

**Test-count check:** 8 declarations on your side before the merge, 34 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** `pnpm install --frozen-lockfile --prefer-offline` at the worktree root, then: (1) `pnpm --filter @ifc-lite/mutations exec vitest run test/change-set.test.ts` -> "Test Files 1 passed (1) / Tests 26 passed (26)", which equals 7 (PR side) + 19 (main side) exactly; (2) full package `pnpm --filter @ifc-lite/mutations test` -> "Test Files 8 passed | 1 skipped (9) / Tests 157 passed | 10 skipped (167)"; (3) `pnpm --filter @ifc-lite/collab test` (the PR also touches four collab test files, unconflicted) -> "Test Files 44 passed (44) / Tests 209 passed | 2 skipped (211)"; (4) standalone strict `tsc --noEmit` on the merged test file -> no output (clean). Note the mutations package `test` script carries `--passWithNoTests`, so I checked the actual "Tests N" line rather than trusting the exit code.

**Worth knowing:** Counting convention: testsBefore/testsAfter count all `it(`/`describe(` declarations in the one conflicted file. PR side before = 8 (7 it + 1 describe 'ChangeSetManager'); main side = 26 (19 it + 7 describe, one of which is the same 'ChangeSetManager' name); union of both = 33 distinct names; merged = 34 (33 union + the one new wrapper describe 'ordering fixtures that disagree with insertion order' I added to hold the PR's tests). Counting only `it(`: 7 before, 26 after (7+19), zero dropped.

Judgement calls: (a) I nested the PR's seven tests in a new describe instead of leaving two sibling top-level `describe('ChangeSetManager')` blocks — this changes the reported suite path of those seven tests (now "ChangeSetManager > ordering fixtures that disagree with insertion order > ..."), though every `it()` name is byte-identical. (b) I renamed the PR's `mutation` helper to `timestampedMutation` rather than main's, since main's is called ~20 times vs the PR's ~10; the helper bodies are genuinely different (UPDATE_PROPERTY with pset/prop/old/new vs a cast SET_PROPERTY), so keeping both was required. (c) I merged the two file-header doc comments into one.

Environment note worth flagging: on entry this worktree's index/worktree did not match HEAD after `git checkout -B` (the index held origin/main's tree for ~285 files while HEAD pointed at the PR commit), which made the first `git merge` abort with "local changes would be overwritten". I ran `git reset --hard HEAD` to restore a clean PR-branch checkout before merging. There were no untracked files and no work of mine at that point, so nothing was lost, but the pre-existing worktree state was inconsistent and someone may want to know that.

Not verified: I did not run repo-wide lint, the Rust suites, or any other package's tests — the merge only conflicted in one TypeScript file and auto-resolved everything else. Also note packages/mutations has no `typecheck` script and its tsconfig only includes `src/**/*`, so CI never typechecks this test file; I typechecked it manually instead. CI on the PR is currently mergeStateStatus BLOCKED (review/checks gating), not conflicted — GitHub now reports mergeable: MERGEABLE.

PR #2179, opened 2026-08-05
fix(data,cache): stop sumByType silently ignoring its elementType filter

BIMvoice:
@louistrue — this one changes behaviour on two published packages, and the remedy is a judgement call rather than a forced move. Details and alternatives below; happy to switch to whichever you prefer.

## The defect

`QuantityTable.sumByType(quantityName, elementType?: number)` declares an optional element-type filter. **Two of the three implementations silently drop it**, because they are arity-1 closures:

| implementation | honours `elementType`? |
|---|---|
| `packages/data/src/quantity-table.ts:204` | no — `(quantName) => {...}` |
| `packages/cache/src/sections/quantities.ts:133` | no — `(quantName) => {...}` |
| `apps/viewer/src/utils/serverDataModel.ts:750` | **yes** — builds a valid-id set from `entities.getByType(elementType)` |

Three implementations of one interface, disagreeing, with no way for a caller holding the interface type to tell which one it has.

## Not currently live — a landmine, not an active bug

I grepped every `sumByType` reference in the repo. The only call site is `tests/integration.test.ts:158`, and it passes a single argument. **Nothing returns a wrong total today.** I'd rather say that plainly than dress this up as a live bug.

What makes it worth fixing anyway is the failure mode. A dropped filter doesn't error — it returns a total over *every* element. In a quantity context that is a plausible-looking wrong number, which is the worst thing to produce silently.

## Why not just implement it

Neither implementation has the data. Both see only `entityId` per row; the entity-type mapping lives in `EntityTable`, assembled in `packages/parser`. Honouring the filter would mean either threading that through or bumping the cache binary format — a materially bigger change than this defect justifies, and one I didn't want to make unilaterally.

## What this PR does

Both implementations now **throw** when `elementType` is passed, naming the supported route (resolve ids via `entities.getByType(elementType)`, total the matching rows). Zero-argument calls are untouched. The interface doc records the reasoning so the next reader doesn't re-derive it.

## The alternatives, since this is your call

1. **Throw** (this PR) — fails loud. Cost: a runtime error on a call TypeScript permits.
2. **Remove `elementType` from the interface** — best developer experience, a compile error instead of a runtime one. Cost: the one implementation that *does* support filtering loses it for interface-typed callers.
3. **Implement it for real** — correct, but needs entity-type data plumbed into both tables and likely a cache format bump.

I went with 1 because it is the smallest change that stops a silently wrong quantity, and it leaves 2 and 3 open. Say the word if you'd rather have either.

## Verification

RED-GREEN both ways, by reverting each implementation to its arity-1 form (exact-substring, replacement count asserted `n == 1`):

- `packages/data`: **1 failed of 14** — `AssertionError: expected [Function] to throw an error`. Restored: **83 pass**.
- `packages/cache`: **1 failed of 2**, same assertion. Restored: **50 pass**.

Control (`sum += value[idx]` → `* 2` in the same function) dies as expected, so the surrounding suite isn't vacuous either.

Fixtures use two entities with deliberately different values (10 and 100), so the unfiltered total (110) can't coincide with a filtered one.

## Negative results

Reviewed `relationship-graph.ts`, `entity-table.ts`, `property-table.ts`, `storey-elevation.ts` and the `entities`/`properties`/`relationships` cache sections for the same shape. Their optional parameters (`type?`, `psetName?`, `relType?`) are all correctly threaded through to their closures — no further arity mismatches. `geometry.ts`, `instanced-shards.ts`, `header.ts`, `entity-index.ts` and `epsg-index.ts` were not reached and remain unexamined.

### DIFF
--- a/.changeset/quantity-sumbytype-elementtype.md
+++ b/.changeset/quantity-sumbytype-elementtype.md
@@ -1,0 +1,12 @@
+---
+"@ifc-lite/data": patch
+"@ifc-lite/cache": patch
+---
+
+Stop `QuantityTable.sumByType` from silently ignoring its declared `elementType` filter.
+
+`sumByType(quantityName, elementType?)` declares an optional element-type filter, but two of the three implementations were arity-1 closures that dropped it: the columnar table in `@ifc-lite/data` and the cache-restored table in `@ifc-lite/cache`. The third — the server-backed table in `apps/viewer` — honours it for real, resolving ids through `entities.getByType`. So three implementations of one interface disagreed, and a caller holding the interface type had no way to tell which behaviour it would get.
+
+The failure mode mattered more than the type-level inaccuracy: a dropped filter returns a total over *every* element rather than an error, and in a quantity context a plausible wrong number is worse than a loud failure. No caller passes the second argument today, so nothing changes for existing code.
+
+Neither implementation can honour the filter as written — both see only `entityId` per row, with the entity-type mapping living in `EntityTable`. Rather than leave the contract lying, both now throw when `elementType` is passed, naming the supported route (resolve ids via `entities.getByType(elementType)` and total the matching rows). The interface doc records why.
--- a/packages/cache/src/sections/quantities.test.ts
+++ b/packages/cache/src/sections/quantities.test.ts
@@ -1,0 +1,50 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `QuantityTable.sumByType` declares an optional `elementType` parameter,
+ * but neither this cache-restored implementation nor the columnar one in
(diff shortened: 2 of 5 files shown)


louistrue:
**Decision from Louis: option 1, throw — this PR as it stands.**

Reasoning he weighed it on: it is the smallest change that stops a silently-wrong quantity, and it leaves options 2 and 3 open. Removing `elementType` from the interface would cost the one implementation that actually filters its ability to be reached through the interface type, and implementing it for real is not worth a cache format bump for a parameter with no caller.

Your framing of it as "a landmine, not an active bug" is the right way to have presented it — that is what made the decision easy.

Marking ready and taking it through the gate now. No changes needed from you.

BIMvoice:
Understood — option 1, no changes from us. Nothing further pushed to this branch; it is as you reviewed it.

For the record, since it affects how the next one of these gets framed: the reason it was presented as a landmine rather than a bug is that no caller passes the second argument today, so the wrong total was reachable only by a future caller trusting the signature. That is also why throwing costs nothing right now — there is no behaviour to break, only a contract to stop lying.

PR #2182, opened 2026-08-05
test(geometry): pin the RTC threshold and the WASM-RTC heuristic

BIMvoice:

`packages/geometry/src/coordinate-handler.ts` has no dedicated test file — its coverage lives inline in `geometry.test.ts`, and neither of its two threshold comparisons was ever exercised at its boundary.

## 1. `needsShift` at exactly 10 km

`coordinate-handler.ts:207` — `return maxCoord > this.THRESHOLD;` (10 km).

`>` → `>=`, 1 replacement asserted: **225 passed, 0 failed.** Every fixture sat at 100 or 500,000+, never at the line.

Bounded by a control: forcing `return false` kills 2 tests immediately, so the function genuinely is exercised — it was only the boundary that was blind. A model whose bounds max out at exactly 10000.0 m would silently skip the RTC shift, and nothing pinned it either way. Both sides are now pinned.

## 2. The WASM-RTC heuristic — the serious one

`coordinate-handler.ts:445` infers "WASM already applied RTC" from a strict majority of small-coordinate meshes:

```js
const wasmRtcLikelyApplied = totalMeshes > 0 && (smallCoordCount / totalMeshes) > 0.5;
```

| mutation | result |
|---|---|
| `> 0.5` → `>= 0.5` | **225/225 pass** |
| whole expression → `false` | **225/225 pass** |

The second matters more than the boundary. `!wasmRtcLikelyApplied` gates whether `originShift` is set at all (`:454`), so a heuristic stuck at `false` **shifts a model WASM has already shifted — moving it twice**. Nothing distinguished the heuristic firing at all.

### Why the boundary test alone wasn't enough

A 50/50 fixture pins `>` vs `>=`, but it **cannot** catch the neutered heuristic: at exactly 0.5 the real expression and a hardcoded `false` produce the same answer, so they coincide. The first version of this work stopped there, and forcing the heuristic to `false` still passed the new suite.

So the `true` side is pinned separately, with a strict majority (2 small + 1 large), and the test asserts the **counterfactual** — the same large mesh *alone* must still produce a shift. Without that, a zero `originShift` proves nothing: a batch that never qualified for a shift has one regardless. Asserting the counterfactual is what attributes the difference to the heuristic rather than to the fixture.

Forcing the heuristic false now fails with `expected 250050 to be +0` — the concrete harm, a 250 km spurious shift.

## A note on the fixture guard

An earlier draft of that guard asserted on `info.originalBounds.max.y > 10000` — and its own check caught it: when the heuristic fires, bounds are **recalculated excluding outliers** (`:450`), so `max.y` reads `3`, not `5000000`. The guard failed loudly rather than letting a wrong assumption through, which is the argument for writing them.

**225 → 228 pass.**

## Negative results

- `geometry-fingerprints.ts` is thoroughly covered at every boundary probed — `volume > 0`, `Number.isFinite` sentinel handling, NaN spans, truncated arrays. No gap.
- **No unverified hash-keyed cache exists in `packages/geometry`.** I went looking for the shape found in `packages/parser` (a rolling-hash cache reused without byte verification, where two real type names collide) and it does not appear here. The nearest analogue, `extractGeometryFingerprints`'s hash, is 64-bit and used as a diff signal, not as a cache key with silent reuse.
- `geometry-parallel.ts` (1681 lines, `affinity_key`/`keyToWorker` routing) and `native-bridge.ts`'s streaming-cache manifest matching were **not reached**. Flagging them as unexamined rather than clean.


BIMvoice:
Extended into `geometry-parallel.ts` and `native-bridge.ts` — `3dd4ccd5f`. One small fix, and **two gaps I am reporting rather than closing**, because closing them properly needs an integration harness I'd rather you weigh in on first.

## Fixed: out-of-range `startWorker` in affinity routing

`planAffinityRouting` reduces its seed with `startWorker % workerCount` (`geometry-parallel.ts:127`) before indexing `buckets[w]`. Dropping the modulo survives **5/5**.

**Latent, not live** — the only production caller seeds it from a prior `nextWorker` return, which is already mod-reduced. Unreachable today; the guard is what keeps it that way, and nothing pinned it. Without the modulo the new test throws `TypeError: Cannot read properties of undefined (reading 'push')` — it indexes past the bucket array rather than misrouting quietly.

Worth saying what is *not* wrong here: the routing function is otherwise well covered. Its fixture uses `workerCount = 4` with several distinct keys, not the single-worker shape that would make `% workerCount` an identity. Control — dropping the wraparound in `nextWorker = (nextWorker + 1) % workerCount` — dies 1/5.

## Reported, not fixed: the dispatch path is never exercised with real jobs

Inverting the dispatch-mode selector —

```diff
-        if (affinity && affinity.length === totalSubJobs) {
+        if (affinity && affinity.length !== totalSubJobs) {
```

— survives **226/226**, with no behavioural change anywhere in the suite.

`dispatchByAffinity` / `dispatchJobsChunkInternal` inside `processParallel` (~lines 902–966), including the interleaved-fallback branch, are never run with real jobs. Every test that calls `processParallel` does so with `workerCountOverride: 1` and `totalJobs: 0`. So the worker-count > 1 fanout and per-worker sub-job packing — the part that decides *which worker gets what* — has no assertion on it at all.

Inspection of the `[job, jobStart, jobLen]` triple packing didn't turn up a defect, so I'm not claiming one. Closing this properly wants a `FakeWorker` harness at `workerCountOverride: 2+` with a real affinity array, asserting per-worker `stream-chunk` payloads. That's a bigger piece of work than this PR, and I didn't want to bolt a half-harness onto it.

## Reported, not fixed: `native-bridge.ts` has zero test coverage

I grepped the whole repo: **no test file anywhere references `native-bridge`**. It's reachable only through a dynamic `import('./native-bridge.js')` in `platform-bridge.ts`, on the Tauri desktop path.

Confirmed by control: `if (processedMeshes !== manifest.totalMeshes)` → `if (false)` leaves **225/225** untouched.

One design point I noticed while there and want to flag rather than change unilaterally: that mismatch check only `console.warn`s (`native-bridge.ts:585-589`). It never rejects the cache or refetches. So a stale manifest or an undercounted shard stream is **accepted silently** and the model renders with fewer meshes than the manifest claims. Whether that should hard-fail is your call — it's the desktop cache path, and I have no repro, since exercising it needs a Tauri `invoke` mock for `get_native_geometry_cache_manifest` / `get_native_geometry_cache_packed_shard`.

## Verification

Rebased onto the first commit, so the mutations were re-run **after** the rebase: both the `startWorker` modulo and the WASM-RTC heuristic mutated together give **2 failed / 227 passed of 229**; restored, **229/229**.

louistrue:

**How the conflict was resolved:** Only one file conflicted, packages/geometry/src/geometry.test.ts, and only one hunk inside it: this branch and #2161 (landed on main today) both appended threshold tests to the `describe('needsShift')` block, so the two appends overlapped. I resolved it by keeping all three tests — the branch's 'pins the AT-threshold boundary: exactly 10km is NOT a shift, one unit over IS' plus main's 'treats exactly the 10km threshold as not needing a shift' and 'considers the most negative extent, not just the maxima' — since the names differ and both bodies pin distinct sides of the same operator. Neither side touched imports or the createTestMesh helper, so there is still exactly one import block and one helper; the only duplicate declaration names in the merged file (describe('processMeshes') and it('should handle empty mesh array')) predate both sweeps and sit under different parent describes. Declaration counts: 31 on the branch, 46 on main, 49 merged, and a set comparison showed zero names from either side missing. The branch's other three additions (two incremental-processing tests in geometry.test.ts and the affinity-routing.test.ts change) auto-merged untouched. Maintainer note: the worktree arrived with a large pre-staged uncommitted change set that blocked the merge; I discarded it with `git reset --hard` back to the pushed branch tip (3dd4ccd5f, identical to origin) before merging, so nothing unpushed was lost.

**Test-count check:** 31 declarations on your side before the merge, 49 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran the real suite in this worktree (node_modules was already present, so no install was needed). `packages/geometry/node_modules/.bin/vitest run src/geometry.test.ts` -> "Test Files 1 passed (1), Tests 37 passed (37)", which matches the 37 `it(` declarations in the merged file exactly (49 total declarations minus 12 `describe(` blocks). Then the whole package: `vitest run` in packages/geometry -> "Test Files 26 passed (26), Tests 300 passed (300)". No skipped or zero-test files.

**Worth knowing:** Two things worth flagging. (1) The worktree was not clean when I started despite the reported status: it had ~285 files staged, containing what looked like an already-attempted union of this same merge. I did not trust or reuse it; I verified HEAD equalled origin/test/geometry-coordinate-boundaries, `git reset --hard HEAD`, and performed the merge myself from scratch. My independent resolution came out byte-identical in size to that pre-existing file, which is reassuring but was not used as input. (2) The session scratchpad directory is being written concurrently by other agents — my first comparison files (pr.ts/main.ts) were overwritten mid-task by another agent's unrelated data, which produced a bogus first superset report. I redid the whole comparison in an isolated subdirectory re-extracted straight from git; the clean numbers above (31/46/49, no missing names) come from that second run. Judgement call: the branch's 10km-boundary test and main's are near-duplicates in intent but differ in fixture (symmetric -10000..10000 vs 0..10000, and 10000.001 vs 10000.5), so I kept both rather than deduplicating; the names already differ, so no suffix was needed. I did not run lint/typecheck or the wider monorepo suite, and I did not check CI status after the push.

PR #2185, opened 2026-08-05
fix(sdk): correct two EntityRef decoding bugs

BIMvoice:
`packages/sdk/src/types.ts` had no test file. Both defects were confirmed RED-first against unmodified code before any fix.

## 1. A truncated ref silently decoded to entity 0

`Number('')` is `0` — finite and non-negative — so it sailed through the existing `Number.isFinite(expressId) || expressId < 0` guard. `stringToEntityRef('arch:')` returned `{ modelId: 'arch', expressId: 0 }` instead of throwing.

That is the bad kind of failure: a truncated or corrupted persisted reference resolves to a real-looking entity rather than failing where the corruption happened.

## 2. The encoder and decoder disagreed about their own format

`entityRefToString` emits `modelId` unescaped, so `{ modelId: 'proj:arch', expressId: 5 }` becomes `'proj:arch:5'`. Decoding split on the **first** colon, took `'arch:5'` as the id part, and **threw**. The codec could not round-trip its own output.

## The fix

Split on the **last** colon — `expressId` is always numeric so it can never contain one, while `modelId` may — and require `/^\d+$/` rather than relying on `Number()` coercion.

RED-GREEN: against unmodified code the new tests give **2 failed / 2 passed**; after the fix **4/4**, full package **73 pass** (was 69). Mutations, each 1 replacement asserted: `lastIndexOf` → `indexOf` dies 1/4; deleting the `/^\d+$/` guard dies 1/4.

## Severity, and who actually calls this — worth reading

**Latent.** No in-repo caller passes a colon-bearing `modelId`.

But the call-site picture is not what it looks like at first glance, and I want to correct it explicitly rather than let a wrong premise ride: **`apps/viewer` does not use this function.** It has its *own* implementation at `apps/viewer/src/store/types.ts:471-489`, imported from `../types.js`, with a materially different contract:

| | `packages/sdk` | `apps/viewer/src/store` |
|---|---|---|
| invalid input | throws | returns sentinel `{ modelId: '', expressId: -1 }` |
| colon in `modelId` | (now) last-colon split | first-colon, deliberately |

The viewer's behaviour is intentional and pinned — `store/types.test.ts:63` asserts that `'model:with:colons:123'` yields `expressId: -1`, with a comment explaining why.

So this is **published SDK API with no in-repo consumer**, and after this change two implementations of one concept differ in two respects rather than one.

@louistrue — the duplication is arguably the more interesting finding than either bug. Collapsing them is an API decision (which contract wins: throw or sentinel?), so I have not touched the viewer copy. Happy to unify them either direction if you want it.

## Two negative results

**`VERSION_PARSE_RE` does not exist in this scope.** I had this flagged as a lead in `packages/cli` / `packages/data`; a repo-wide grep finds exactly one hit, `packages/extensions/src/host/sdk-version.ts:64`, which is a different package. The lead as I stated it was wrong.

**`apps/server` (Rust) could not be audited.** `cargo test` fails in this worktree with `E0152: duplicate lang item in crate 'core'` before any test runs, reproducing with a fresh `CARGO_TARGET_DIR` and with `.cargo/config.toml` removed. It looks environmental — CI runs `cargo test --workspace` fine — but with no admissible green baseline I ran no mutations there rather than report findings I could not stand behind. That leg needs a clean toolchain.

The `packages/cli` sweep found nothing: `diff-content.ts`, `gym.ts`'s three-state `--corrupt`/`--no-corrupt`/`--corrupt-rate` flag, and the top-level dispatch all showed the parameter properly threaded, with no single-field-validation or silent-default-allow shapes. `mutate.ts` and `headless-backend.ts` remain unexamined.


louistrue:
**Decision from Louis on the duplication: keep both, document the split. Do not unify.**

The two contracts are both right for where they live — the viewer decodes untrusted DOM/state strings on hot paths and must not throw, the SDK is a published API where failing at the corruption site is the correct behaviour. Unifying on throw would flip a deliberately pinned viewer test and put a `try/catch` on every viewer call site; unifying on the sentinel would be a breaking change on a published package that trades a loud failure for a silent `-1`.

You were right that the duplication is the more interesting finding than either bug — it just resolves to "deliberate", not "collapse".

**One thing to add before this goes in:** a note in each file naming the other implementation and why the contracts differ, so the next reader does not re-derive this. Roughly:

```ts
// packages/sdk/src/types.ts
/** NOTE: apps/viewer/src/store/types.ts carries a second implementation with a
 *  SENTINEL contract ({modelId:'',expressId:-1}) and a FIRST-colon split.
 *  Deliberate, not drift: the viewer decodes untrusted DOM/state strings on hot
 *  paths and must not throw. Keep the two in step on *bugs*, not on contract. */
```

and the mirror of it in the viewer file, pointing back here.

Both bug fixes themselves are good — the `Number('') === 0` hole is exactly the failure mode worth closing, and last-colon + `/^\d+$/` is the right shape. Push the doc notes and mark it ready; #2186 can go independently.

Also noting your `apps/server` negative result: the `E0152 duplicate lang item` you hit is environmental. It runs clean here — `cargo test` in `apps/server` gives 126 passing. Worth a fresh `CARGO_TARGET_DIR` and a `rustup` check on that worktree if it recurs; not a defect in your changes.

BIMvoice:
Doc notes pushed — `1c77aac`. Both files now name the other and say why the contracts differ.

I verified the viewer implementation still matches what the note claims rather than describing it from the sketch. `apps/viewer/src/store/types.ts:476-489`:

```ts
const colonIndex = str.indexOf(':');
if (colonIndex === -1) return { modelId: '', expressId: -1 };
...
if (Number.isNaN(expressId)) return { modelId, expressId: -1 };
```

First-colon split, sentinel on both the no-separator and the NaN path — as described, so the note is accurate.

## One thing you need to know before resolving the conflict

This branch is still **CONFLICTING**, and I have deliberately not rebased it, because the resolution is not mechanical and I would rather you made the call than have me force-push a resolution over it.

`main` has since gained its **own** `packages/sdk/src/types.test.ts` for this same codec — an add/add conflict against the one this PR adds. `main`'s version is the broader suite (field-order pinning, zero `expressId`, negative / non-finite / empty-`modelId` rejection, an `it.each` round-trip block).

**But it does not cover either bug this PR fixes.** If the conflict is resolved by taking `main`'s file, both regressions land silently uncovered:

- `'rejects a truncated ref with an empty expressId'` — the `Number('') === 0` hole, where `'arch:'` decoded to `{ modelId: 'arch', expressId: 0 }`
- `'round-trips a modelId that itself contains a colon'` — the encoder/decoder disagreement, where `'proj:arch:5'` is emitted but was not decodable

Those two must survive the resolution. The other two tests this PR added (`arch:42` round-trip, `arch:abc` rejection) are byte-identical in intent to assertions already in `main`'s suite, so they can be dropped as duplicates.

Happy to push that resolution if you want it from me — say the word and I will take `main`'s file as the base, fold in those two cases, and add the colon-bearing case to the `it.each` block. I held off only because resolving it means rewriting this branch's history, and that is your call on a branch you are about to gate.

## On the `E0152`

Confirmed environmental, and I found the same thing independently: it is a **shared target-dir artifact**, not a property of the crate. A scratch `CARGO_TARGET_DIR` clears it — that is what unblocked `apps/server` for us mid-sweep, and it went from "written off" to a full suite run. So your 126 passing and our failure are the same tree; nothing to chase.

louistrue:

**How the conflict was resolved:** The only conflict was an add/add on packages/sdk/src/types.test.ts: the file did not exist at the merge base, this branch created it with a 4-test EntityRef codec block covering its two decoding fixes, and #2169 independently created it on main with a 15-declaration SDK coverage sweep. I resolved it by unioning both sides under one MPL header, main's file-level doc comment, and one merged import block (main's `import type { EntityRef }` plus the shared vitest/types imports, no duplicated import or helper). All three of main's describes and this branch's describe are kept verbatim; the single name collision, `rejects a non-numeric expressId`, exists on both sides with different bodies (bare `.toThrow()` vs `.toThrow(/Invalid expressId/)`), so both were kept and the branch's looser one was suffixed `(bare throw, no message match)`. I verified the merged file's declaration list is a strict superset of both step-2 lists (20 declarations vs 5 from the PR side and 15 from main, nothing dropped), that no conflict markers remain, and that main's assertions still hold against this branch's rewritten `stringToEntityRef` (the new `/^\d+$/` guard still yields "Invalid expressId" for `-1`/`Infinity`/`NaN`/`abc`, and the lastIndexOf split still yields "Invalid EntityRefString" for `:42`, `arch` and ``). The merge is an ordinary two-parent merge commit (a8195e7d4, parents 1c77aac90 and 4dcf14c45) pushed non-force to fix/sdk-entityref-codec; GitHub now reports the PR MERGEABLE with headRefOid a8195e7d4, and `git status --porcelain` is empty.

**Test-count check:** 5 declarations on your side before the merge, 20 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran them for real. `pnpm install --frozen-lockfile --prefer-offline` at the worktree root, then `pnpm --filter @ifc-lite/sdk exec vitest run src/types.test.ts` -> "Test Files 1 passed (1), Tests 18 passed (18)" (18 = 15 named `it`s + the 3 cases the `it.each` round-trip expands to). The first full-package run failed 2 files (namespaces.test.ts, context.test.ts) with "Failed to resolve entry for package @ifc-lite/lists" because the fresh worktree had no sibling dist/; after `pnpm --filter @ifc-lite/sdk^... build` the full suite is green: `pnpm --filter @ifc-lite/sdk test` -> "Test Files 9 passed (9), Tests 156 passed (156)". Also ran `tsc --noEmit` for the sdk package (clean), since vitest does not typecheck and the merged file gained a type-only import.

**Worth knowing:** Judgement calls: (1) I renamed one PR-side test to `rejects a non-numeric expressId (bare throw, no message match)` rather than dropping it as a near-duplicate — the two bodies genuinely differ in strictness, and the original name survives from main's side, so the superset check still holds. (2) I ordered main's three describes first and appended this branch's describe last, with a one-line comment saying what it pins; ordering is cosmetic and changes no test identity. (3) I kept main's file-level doc comment (which says apps/viewer has a separate same-named codec) because it agrees with the note this branch added to packages/sdk/src/types.ts. Not verified: CI as a whole. `gh pr view` reports mergeStateStatus BLOCKED, which at the moment of the push reflects pending/required checks rather than a conflict (mergeable is MERGEABLE); I did not wait for the CI run on the new head commit, and I did not run any package other than @ifc-lite/sdk, so a cross-package effect of merging today's main into this branch would not have been caught by what I ran. The build step created untracked dist/ output in the worktree; it is gitignored and `git status --porcelain` is empty.

PR #2187, opened 2026-08-05
test(viewer): pin seven selection and basket gaps

BIMvoice:
Test-only. All three production files are byte-identical to base (verified with `diff` against `HEAD`, not just `git status`). Every finding is a coverage gap — the code was already correct.

## The sharpest one: a test named for the scenario it doesn't test

`basketVisibleSet.ts`'s `globalIdToRef` has a federated guard — `if (state.models.size > 0) return null;` — that drops a globalId nobody can resolve rather than falling back to `{modelId: 'legacy', ...}`.

**Disabling that guard entirely leaves the suite green at 17/17.**

The existing test is called `getBasketSelectionRefsFromStore returns array when models exist`. Its assertion is `assert.ok(Array.isArray(refs))` — which is true for `[]` and for any populated array alike — and it does not even set `models` non-empty. It is named for exactly the scenario it fails to exercise.

Had the guard been broken, a stale or foreign globalId in a multi-model scene would be mislabelled as belonging to `'legacy'` and leak into the basket. The new test builds a federated `models` map with an out-of-range globalId and asserts `refs` is `[]`, not a phantom ref.

## Same-size disjoint sets

`isBasketIsolationActiveFromStore` compares sizes, then checks membership. Deleting the membership loop survived **17/17** — no fixture had same-size, disjoint sets, so the size check alone always agreed. Now pinned with `{100, 200}` against `{300, 400}`.

## The selection primary-match guard, at two call sites

`selectionSlice.ts:216` (remove) and `:239` (toggle) share the guard `selectedEntity?.modelId === ref.modelId && ...expressId === ...`. Mutating `&&` → `||` at **either** line survived **31/31**: no fixture had a decoy sharing an `expressId` across a different model, so the two-field match was never discriminated. That is the cross-model contamination shape this slice exists to prevent.

Also `clearEntitySelection`: dropping `selectedEntities: []` or `selectedModelId: null` from the reset object each survived **33/33**, because the existing test checked three of the five fields it clears.

## An identity fixture in the pinboard

`addToBasket` and `removeFromBasket` both call `refToGlobalId(ref, state.models)`. Replacing either with a bare `ref.expressId` survived **10/10** — every fixture used a single `'legacy'` model with `idOffset: 0`, where the two are numerically identical. The new fixture uses `idOffset: 1000` and asserts both that the offset-corrected id (1007) is present **and** that the raw expressId (7) is absent.

## Control

Commenting out `newSet.add(key)` in `addEntityToSelection` dies **10 of 34** — the suite is not uniformly green regardless of mutation.

**58 → 65 pass.**

## Established but deliberately not acted on

`entityRefToString` does not escape `modelId`, so a colon-bearing modelId corrupts silently on round-trip: `{modelId: 'a:1', expressId: 2}` → `'a:1:2'` → `{modelId: 'a', expressId: 1}`. I confirmed that with a live repro rather than by reading.

It is **latent**. Every modelId-minting site — `useIfcFederation.ts:123`, `useIfcLoader.ts:269`, `modelSlice.ts`'s `addModel` — uses `crypto.randomUUID()` or the literal `'legacy'`, never a filename or user-supplied string. So no reachable path produces one today, and writing a test for it here would only re-test `types.ts`, which is covered in #2186 and whose contract question is open on #2185.

## Negative results

`getSelectedEntitiesForModel` is defined in `selectionSlice.ts` and has **zero production callers** — only its own test references it. Dead API surface, so a sentinel ref reaching it has no blast radius today; not pursued.

`saveCurrentBasketView`, `restoreBasketEntities`, `renameBasketView`, `refreshBasketViewThumbnail`, `setBasketViewTransitionMs`, `dedupeRefs` and `getVisibleBasketStatsFromStore` were spot-checked: their tests already assert membership and content rather than size alone. No survivors there.


louistrue:

**How the conflict was resolved:** The only real conflict was in apps/viewer/src/store/basketVisibleSet.test.ts, inside the `isBasketIsolationActiveFromStore` describe, where this branch and #2162 each appended new cases immediately after 'returns false when isolated size differs from basket' — an overlapping-append between two independent test sweeps, not a semantic disagreement. I resolved it by unioning all three cases under main's single set of imports and one copy each of the createMesh / createGeometry / createFederatedModel helpers: this branch's 'returns false when isolated has the same SIZE as the basket but different members' plus main's 'strict SUPERSET' and 'sizes match but membership does not'; no --ours/--theirs anywhere, and the branch's other contribution ('drops a globalId that resolves to no model...') auto-merged intact. I proved the result is a strict superset by extracting the declaration lists from all three versions and running `comm` both directions: zero names missing from the PR side, zero missing from main, no duplicated imports, no duplicated helper functions. Tests genuinely ran: 34 pass / 0 fail on the conflicted file (34 tests + 10 suites = the 44 declarations, so nothing silently failed to load), 574 pass / 0 fail across the whole viewer store directory, and `tsc --noEmit` clean for apps/viewer. Maintainer must know: the worktree arrived with a stale half-finished merge already in the index, which I aborted and re-checked-out from origin before starting, so the pushed commit 89aed8c41 is an ordinary two-parent merge whose first parent is the untouched branch head f9f528cd6 — no amend, rebase or force-push.

**Test-count check:** 26 declarations on your side before the merge, 44 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Yes, really ran, in the worktree (node_modules was already present, so no install was needed).

1. The conflicted file alone: `npx tsx --test --test-concurrency=1 src/store/basketVisibleSet.test.ts` from apps/viewer -> `# tests 34 / # suites 10 / # pass 34 / # fail 0`. The 34 leaf tests + 10 suites exactly equals the 44 declarations I counted in the merged file, which confirms every declaration actually loaded and executed rather than being silently skipped.

2. The two sibling files the merge also touched: `selectionSlice.test.ts` + `pinboardSlice.test.ts` -> `# tests 58 / # suites 18 / # pass 58 / # fail 0`.

3. The whole viewer store directory (36 test files), to catch cross-file store-state leakage between the two sweeps, since main's new federated fixtures mutate the shared zustand store: `# tests 574 / # suites 169 / # pass 574 / # fail 0`.

4. `npx tsc --noEmit` in apps/viewer -> EXIT=0, zero output lines. This matters because main's side of the merge added new type-only imports (GeometryResult, MeshData, FederatedModel, ViewerState) that the union had to keep exactly once.

No zero-test or missing-summary runs; every invocation reported a non-zero `# tests` line with `# fail 0`.

**Worth knowing:** Two things worth flagging.

1. The worktree had a LEFTOVER in-progress merge when I arrived (git status showed "You have unmerged paths" on basketVisibleSet.test.ts, with a large staged changeset), even though the handed-off snapshot claimed the repo was clean. I ran `git merge --abort` and re-checked out `origin/test/viewer-selection-basket-gaps` before doing anything, so my merge started from the true branch head f9f528cd6. Nothing from that stale attempt is in my commit — the merge commit's first parent is exactly f9f528cd6, the pre-existing branch head.

2. Judgement call on near-duplicate tests. The PR's 'returns false when isolated has the same SIZE as the basket but different members' and main's 'returns false when sizes match but membership does not' pin the same invariant with different data (fully disjoint {300,400} vs partially overlapping {100,999} against a 2-element basket). Names differ, so there was no hard collision and no suffix was needed. I kept both rather than deduplicating, per the instruction to keep both when bodies differ — the cost is one mildly redundant test, which I judged far safer than dropping a sweep's assertion.

Also note `selectionSlice.test.ts` auto-merged and contains two tests with the IDENTICAL name 'does not treat the primary as removed when only expressId matches but modelId differs'. I checked this and it is NOT a merge artifact: both occurrences come from the PR side (main has zero), and they live in different describe blocks (removeEntityFromSelection vs toggleEntitySelection) with different bodies. Left as-is.

Not verified: I did not run the full apps/viewer suite (~2750 tests), only the 36-file store directory that contains all three touched files. I also did not run lint or the repo-wide typecheck, only `tsc --noEmit` scoped to apps/viewer. CI has not reported on the pushed merge commit yet.

PR #2189, opened 2026-08-05
fix(viewer): deliver a remote peer's first property in a new Pset

BIMvoice:
@louistrue — this one contains a **live collaboration defect**, fixed, plus a second one I could not fix without an API change and want your call on.

## The defect

`attachRemoteApply` (wired in at `collabSlice.ts:816`) matched pset events only at `path.length === 3`. But when a remote peer writes the **first** property of a brand-new Pset, Yjs emits exactly one event — on the `psets` map itself:

```
PATH ["/wallA","psets"]  len 2   key Pset_WallCommon   action add
```

The nested pset map doesn't exist yet when the transaction is observed, so **no length-3 event is ever emitted** and the property was silently dropped.

I verified that against real Yjs event shapes with a direct `observeDeep` probe rather than inferring it from the code, because the failure is entirely about Yjs's event granularity and reading the handler alone would not settle it.

**User-visible effect:** a collaborator adds a property to an element that had no Pset of that name — your view never shows it. The data is *correctly persisted* in the Y.Doc, so a reload makes it appear, which is exactly the shape that gets reported as "sometimes properties don't sync" and is miserable to reproduce.

Fix adds the `path.length === 2` branch. Removing it fails 1/23; narrowing it back to length 3 fails 1/23 — the branch is load-bearing, not decorative.

## The mirror case, which I could not fix — your call

Deleting a Pset's **last** property cascades to removing the pset entry, and also collapses to a single length-2 event. It looks symmetrical, but it isn't fixable the same way. By the time the event is observed the removed map is already detached:

```
deleted pset Pset_WallCommon -> forEach yielded 0 entries []  | size: 0  | toJSON: {}
```

The property names are gone, so `onPropertyDelete(entityId, psetName, prop)` cannot be called — there is no `prop` to name. Closing it needs a whole-pset signal on `RemoteApplyHandlers` (something like `onPsetDelete(entityId, psetName)`), and the consumer has to know to drop the whole set. That's an interface change, so I've left it as a `[DEFECT]` regression test pinning current behaviour with the reasoning inline, rather than guessing at an API.

Worth noting the practical severity is lower than the add case: the stale property lingers in the view rather than a new one failing to appear.

## Plus 39 mutation-verified coverage gaps (tests only)

Across 8 files, chosen by blast radius. Representative ones:

- **content-keyed diff** (#1891, freshly merged): the `matchedElements` accumulator was never proven to read `match.head` versus `match.base`; the first-mesh-wins fold for `geometryHash` was untested when submeshes disagree; and the synthetic fallback key `missing:${modelId}:${localId}` was never proven per-entity — two GlobalId-less entities in one model could have collided.
- **layers**: the entire `kind: 'registry'` merge backend was untested (only `local` was covered), and the `DELETE_ENTITY` path had zero coverage — a truthy-but-not-`true` `IFCLITE_ATTR.DELETED` would silently keep entities alive downstream.
- **search**: `unionByStorey`'s storey-prefilter was never invoked by any fixture, so a case-insensitivity bug there would have survived; `materialiseIterable`'s `Set` branch was untested, where the progress total silently degrades to `-1`.

**159 pass** across the 8 suites. `apps/viewer` is `private: true`, so no changeset — consistent with the convention stated in `.changeset/dxf-3d-reference-layer.md` ("private package, no changeset needed for that half").

## Open gap, not claimed as clean

Classification and material flattening in `collab/step-seed.ts` remain uncovered — flagged for a later round rather than papered over.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collaborative editing so remote property-set additions and removals are reflected correctly.
  * Improved synchronization of entity properties, attributes, placements, and deletions while avoiding duplicate local updates.
  * Improved handling of unresolved entities and incomplete model data.
  * Fixed comparison results involving retired matches and geometry-based identification.
  * Improved search filtering, result naming, federated progress reporting, and storey-based filtering.
  * Improved merge and publishing feedback for unrelated sources, missing checks, unresolved entities, and deleted properties.
* **Tests**
  * Added broad coverage for collaboration, model seeding, comparison, search, merge, and publishing scenarios.

### DIFF
--- a/apps/viewer/src/lib/collab/mutation-bridge.test.ts
+++ b/apps/viewer/src/lib/collab/mutation-bridge.test.ts
@@ -4,15 +4,20 @@
 
 import { describe, it } from 'node:test';
 import assert from 'node:assert';
+import { createRequire } from 'node:module';
+import { pathToFileURL } from 'node:url';
+import { PropertyValueType } from '@ifc-lite/data';
 import {
   createCollabDoc,
   createEntity,
   hasEntity,
   setAttribute,
+  getAttribute,
   setEntityPlacement,
   getEntityPlacement,
   deleteEntity,
   setPropertyValue,
+  getPropertyValue,
   deletePropertyValue,
   matrixToPlacement,
   USD_XFORMOP,
@@ -20,7 +25,43 @@ import {
 } from '@ifc-lite/collab';
 import type { CollabSession } from '@ifc-lite/collab';
 import type { IfcDataStore } from '@ifc-lite/parser';
-import { mirrorPlacement, registerEntityMaps, type CollabDocApi } from './mutation-bridge.js';
+import {
+  mirrorPlacement,
+  mirrorProperty,
+  mirrorPropertyDelete,
+  mirrorAttribute,
+  mirrorEntityDelete,
+  attachRemoteApply,
+  registerEntityMaps,
+  registerEntityPath,
+  type CollabDocApi,
+  type RemoteApplyHandlers,
+} from './mutation-bridge.js';
+
+/**
+ * `yjs` is a transitive dependency (via `@ifc-lite/collab`), not a direct
+ * dependency of the viewer package, so a bare `import 'yjs'` here fails
+ * module resolution. Resolve it through collab's own `node_modules` instead
+ * — only `attachRemoteApply`'s inbound tests need raw `Y.applyUpdate` to
+ * simulate a genuine remote (`txn.local === false`) transaction, which Yjs
+ * only produces via update decode, never via `doc.transact()`.
+ *
+ * Must load the exact same ESM build (`dist/yjs.mjs`) that `@ifc-lite/collab`
+ * itself imports, NOT the CJS build `require('yjs')` would give us — yjs's
+ * internal type checks are `instanceof`-based, and the CJS and ESM bundles
+ * are distinct module instances with distinct classes, so mixing them throws
+ * ("Unexpected content type") the moment a doc synced via the ESM build is
+ * mutated through CJS `Y.Map`/`Y.Doc`.
+ */
+const collabCjsResolve = createRequire(import.meta.resolve('@ifc-lite/collab'));
+const yjsEsmPath = collabCjsResolve.resolve('yjs').replace(/dist[\\/]yjs\.cjs$/, 'dist/yjs.mjs');
+const Y: {
+  applyUpdate(doc: YDoc, update: Uint8Array): void;
+  encodeStateAsUpdate(doc: YDoc, encodedTargetStateVector?: Uint8Array): Uint8Array;
+  encodeStateVector(doc: YDoc): Uint8Array;
+  Doc: new () => YDoc;
+} = await import(pathToFileURL(yjsEsmPath).href);
+type YDoc = ReturnType<typeof createCollabDoc>;
 
 // Minimal CollabDocApi backed by the real collab doc helpers (the viewer wires
 // the same shape from the lazy-loaded runtime in `collabSlice`).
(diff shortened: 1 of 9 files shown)


louistrue:
**Decision from Louis: add `onPsetDelete` now — close the mirror case in this PR.**

Add it to `RemoteApplyHandlers`:

```ts
interface RemoteApplyHandlers {
  onPropertySet(entityId: string, psetName: string, prop: string, value: unknown): void;
  onPropertyDelete(entityId: string, psetName: string, prop: string): void;
  /** The whole Pset vanished. Property names are unavailable by design: Yjs
   *  detaches the map before the event is observed, so `forEach` yields 0
   *  entries. The consumer drops the entire set for (entityId, psetName). */
  onPsetDelete(entityId: string, psetName: string): void;
}
```

and have the viewer consumer drop the whole set. The reasoning for doing it now rather than filing it: you have the Yjs event shapes verified in front of you right now — you probed them with `observeDeep` rather than inferring them — and that context is expensive to rebuild. A stale property lingering in a collaborator's view until reload is the same class of "sometimes properties don't sync" report as the add case you just fixed, and it is miserable to reproduce for exactly the same reason.

Flip the `[DEFECT]` regression test to assert the fixed behaviour when you do, and keep the inline note about why the names are unavailable — that is the part a future reader will need.

The add-side fix is verified good: removing the `path.length === 2` branch fails 1/23 and narrowing it back to length 3 fails 1/23, so it is load-bearing rather than decorative. Nice catch on the Yjs event granularity; reading the handler alone genuinely would not have settled it.

Ping me when it is pushed and I will take the whole PR through the gate.

BIMvoice:
Pushed — `ceea5a4`. `onPsetDelete` is in and the mirror case is closed.

**Signature note:** your sketch had `entityId: string`, but the existing handlers on this interface take `entityId: number` (`onProperty(entityId: number, pset: string, ...)`), so I matched the file rather than the sketch.

**Made it optional (`onPsetDelete?`).** `onPlacement?` and `onEntityDelete?` were both added after the original two handlers and are both guarded at the call site; a required member would break any existing implementer of `RemoteApplyHandlers`. Guarded identically.

**The consumer drops the whole set** via `MutablePropertyView.deletePropertySet` (`packages/mutations/src/mutable-property-view.ts:639`), which already exists and backs the local delete-pset UI path — so the remote path now converges on the same operation as the local one rather than a parallel implementation.

`[DEFECT]` test flipped to assert the fixed behaviour, and the note on why the names are unavailable is kept and expanded — it now states the three observed facts (`forEach` yields 0 entries, `.size` is 0, `toJSON()` is `{}`) rather than just asserting detachment.

Verified RED-first rather than assuming: reverting the emit to the bare `continue` fails exactly that test (1 replacement, count asserted, mutated line re-read), and restoring returns it to green. Full `apps/viewer` suite **2796 passing / 2 skipped / 0 failing** across 248 files; `tsc --noEmit` clean, full output read.

Ready for the gate whenever you are.

BIMvoice:
Marked ready — it was still a draft, which would have blocked the gate. Nothing changed since the ping; head is `ceea5a4`, MERGEABLE, checks green.

Recap of what is in it, so you do not have to reread the thread: `onPsetDelete?(entityId: number, pset: string)` on `RemoteApplyHandlers` (optional, matching `onPlacement?`/`onEntityDelete?` so existing implementers do not break), emitted from the `path.length === 2` delete branch, consumed by dropping the whole set via the existing `MutablePropertyView.deletePropertySet` — so the remote path now lands on the same operation as the local delete-pset UI rather than a parallel one. The `[DEFECT]` test asserts the fixed behaviour and keeps the note on why the property names are unavailable.

Verified RED-first: reverting the emit to a bare `continue` fails exactly that test (1 replacement, count asserted); restoring returns it to green. Full `apps/viewer` suite 2796 passing / 2 skipped across 248 files, `tsc --noEmit` clean.

PR #2196, opened 2026-08-05
test(geometry): pin try_merge_prisms' weld and BASIS_TOL boundaries

BIMvoice:
Test-only. `prism_cut.rs` byte-identical to base. **577 → 579** lib tests.

## Two tolerance constants, neither pinned

Both constants governing the wall-leaf half-void merge survive being loosened by orders of magnitude:

| line | constant | mutation | result |
|---|---|---|---|
| `:1157` | `weld = 8.0 * SNAP_GRID` | `8.0` → `20.0` | **577/577 pass** |
| `:1146` | `BASIS_TOL = 1.0 - 1.0e-8` | → `1.0 - 1.0e-3` | **577/577 pass** |

**Bounded by a control**, which is what makes this a precise claim rather than "merging is untested": forcing `try_merge_prisms` to return `Some(a.clone())` unconditionally kills **5** tests, including `multi_window_host_watertight_and_volume_exact` and two `coaxial_union` cases. The function is well exercised — these two thresholds specifically were blind.

Each new test pins the threshold on **both** sides, exactly at it (must merge) and one epsilon past (must not), since a fixture anywhere mid-region cannot discriminate the comparison at all.

## A correction to this round's severity labels

The round reported both as **live defects**. They're **coverage gaps**. The constants are correct today; nothing pinned them. A surviving mutation tells you the tests don't constrain the code — not that the code is wrong. I'd rather the label carry its proper weight, since "live defect in the void-merge tolerance" would send a reader looking for broken geometry that isn't there.

## An open gap, reported rather than papered over

All three volume self-check gates at `:1949`, `:1952`, `:1955` — the `Err(8)` paths — survive drastic loosening. I re-ran the first myself, widening it **four orders of magnitude** (`1.0e-6` → `1.0e-2` of `vol_host`): **579/579, no change.**

Every current fixture's volume discrepancy is either exactly zero by construction or many orders of magnitude below tolerance, so none of them lands anywhere near these thresholds. That means the self-check gates guarding the fast path against a bad cut are, in practice, unverified.

Closing it needs deliberately near-degenerate geometry — a cutter grazing the host within roundoff — engineered so the measured discrepancy lands within about two orders of magnitude of a floating-point tolerance. That is buildable, but not something to fabricate without being able to confirm it means what it claims, so it is left open and named rather than filled with a test that passes for the wrong reason.

## Checks

Ratchet **4/4**, no split needed. `clippy -D warnings`: zero hits attributed to the touched file (the 4 errors in the crate graph are the known pre-existing ones in `rust/core`, confirmed rather than assumed).

## Not reached

`bool2d_path.rs`, `synthesis.rs`, `probe.rs`, `sweep.rs`, `geom.rs` — the round went entirely to the two carried-over leads plus the volume-gate negative. Unswept, not clean.


louistrue:
Rebased this for you rather than handing it back — merged current `main` in and resolved the `prism_cut_tests.rs` conflict here, since #2194 landed the same file today.

The agent doing this one died on an API error after pushing, so I verified the pushed result independently rather than trusting an unfinished report:

- All **26** test functions on main's side are present in the merged file (set difference empty).
- All **26** on your side are present too (set difference empty).
- The merged file has **29**, i.e. main's 26 plus exactly your three additions: `square_frame`, `try_merge_prisms_basis_boundary_is_inclusive`, `try_merge_prisms_weld_boundary_is_inclusive`.

So it is a genuine union, not a one-sided resolution. It is an ordinary merge commit (`ca1809218`) on top of your history — nothing force-pushed.

Separately, a local CodeRabbit CLI review of this branch raised one point on the `BASIS_TOL` boundary fixture: `u = [BASIS_TOL, 0, 0]` is a shortened, non-orthonormal basis rather than a rotated one, which is unphysical for a `PrismFrame`. I checked it — the dot-product boundary is still hit exactly and the test discriminates (tightening `BASIS_TOL` to `1.0 - 1.0e-9` in `prism_cut.rs:1146` makes it fail), so it is a fixture-realism nit rather than a correctness gap. Not blocking; worth a rotated orthonormal pair next time you are in there.

PR #2198, opened 2026-08-05
test(geometry): close probe/sweep and both carried-over tolerance leads

BIMvoice:
Test-only. `geom.rs` and `synthesis.rs` are **unmodified** — their gaps close from `reveal_tests.rs`, which already reaches their `pub(super)` functions. The only production edits are two `#[cfg(test)] mod tests;` wirings in `probe.rs` and `sweep.rs`. **577 → 584** lib tests; 7 mutations, 7 killed.

### `geom.rs:703` — the paired constant

`geom.rs:622` and `:703` both hard-code a `0.98` axis-merge tolerance, and their doc comments claim to match each other. #2197 pinned `:622`; `:703` still survived being loosened to `0.80`.

It resisted the earlier bevel technique for a specific reason: `infer_opening_frame` merges **weighted** — re-averaging the axis vector — rather than first-seen-wins. The new fixture works *with* that instead of against it. Four single-face triangles, where a tilted one either merges into `+Z` (dragging the weighted average measurably off-axis) or stays a separate 4th axis (leaving the pure `+Z` winner uncontaminated). `depth.z` then reads `< 0.999` or `> 0.999999` respectively — two tests, one either side of the threshold at dot `0.981` and `0.979`.

One detail I checked rather than assumed: the mutation anchor is `.find(|(axis, _)| ...)`, textually distinct from `:622`'s `.position(|axis| ...)`, so it can't collaterally hit the already-pinned sibling and produce a misleading kill.

### `synthesis.rs:545` — the tolerance magnitude

`face_align_tol`'s `1e-5` survived being loosened **10,000×**. The recess fixtures added in #2197 are exactly flush (`diff = 0`), so they pin the branch logic but not this magnitude. Now driven with a wall of thickness 1.0 — making `face_align_tol` exactly `1e-5` — and near-face offsets of `0.9e-5` and `1.1e-5`.

## `probe.rs` and `sweep.rs` had never been mutated

| location | tolerance | result |
|---|---|---|
| `probe.rs:225` | rectangle perpendicularity `dot.abs() > 0.01` | unpinned — **no unit coverage at all**, only incidental reach through fixture integration tests |
| `sweep.rs:41` | volume-drift `1.0e-3` | unpinned |
| `sweep.rs:270` | `VERTEX_CLEARANCE = 1.0e-3` | unpinned |

The `VERTEX_CLEARANCE` fixture needed care worth recording: a needle triangle on its own gets removed by the all-dropped bailout, which would make the test pass for the wrong reason. It's paired with an interior anchor triangle so the effect stays observable as a triangle-count change rather than a wholesale bail.

## Severity

All **coverage gaps**. Every constant is correct today; nothing pinned any of them.

## Checks

Ratchet **4/4**, no split needed. `clippy -D warnings`: zero hits on any touched file (the 4 in the crate graph are the known pre-existing `rust/core` ones).

Nothing was left open in the assigned scope this round — all three leads closed with verified kills.


louistrue:

**How the conflict was resolved:** The only conflict was rust/geometry/src/router/voids/reveal_tests.rs, an overlapping-append: this branch and #2197 (landed on main today) each inserted a new test immediately after test_extend_opening_skipped_when_opening_pokes_past_wall, so git could not tell whose closing brace belonged to whom. I resolved it by keeping both test functions in full (branch side: test_extend_opening_recess_face_align_tol_boundary; main side: test_extend_opening_skipped_for_recess_pocket), adding the missing brace between them, and taking no side wholesale. The branch's other two additions (infer_opening_frame_merges/keeps_axis_*_098_tolerance plus their face_triangle and tilted_axis_frame_mesh helpers) and main's test_rectangular_box_detector_rejects_beveled_corner auto-merged in non-conflicting hunks and are all present. I verified there are no duplicate `use`/`mod` lines (3 imports, all unique), no duplicate fn definitions (25 fns, all distinct names), and no name collisions between the two sweeps. Merged declaration count is 20, a strict superset of the branch's 18 and main's 17 over a shared base of 15; `cargo test -p ifc-lite-geometry` runs the module at 20 tests, all passing, and the whole crate is 593 passed / 0 failed.

**Test-count check:** 18 declarations on your side before the merge, 20 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `cd rust && cargo test -p ifc-lite-geometry --lib router::voids::reveal_tests` -> "running 20 tests ... test result: ok. 20 passed; 0 failed; 0 ignored; 574 filtered out", with each of the 20 names printed individually (both sides' new tests among them). Then ran the whole crate, `cargo test -p ifc-lite-geometry --no-fail-fast` -> lib target "594 tests ... ok. 593 passed; 0 failed; 1 ignored" and every integration target ok, 0 failures anywhere. I also ran `cargo clippy -p ifc-lite-geometry --all-targets`: zero warnings attributable to reveal_tests.rs.

**Worth knowing:** Three things worth flagging. (1) The worktree arrived with an already-in-progress, half-resolved merge of the same branch left over from something else; I ran `git merge --abort` and re-checked-out `-B test/probe-sweep-paired-constants origin/test/probe-sweep-paired-constants` before doing anything, so my merge started from the pristine pushed branch tip (0e080733f) — nothing from that stale state survived. (2) `cargo clippy -p ifc-lite-geometry --all-targets -- -D warnings` fails, but the errors are all in ifc-lite-core (e.g. a clippy::question_mark hit in the comment scanner around core/src line 107) with the local toolchain clippy 0.1.99 dated 2026-08-02, in files neither side of this merge touches. It looks like local-toolchain drift ahead of CI rather than anything the merge introduced; I did not check out origin/main to prove it is pre-existing, so if CI's clippy lane is red, that is where to look, not in reveal_tests.rs. (3) Only the geometry crate was tested. main brought in a large number of unrelated changes (changesets, other packages) that this merge commit now carries; I did not run the TypeScript side or other cargo crates, so CI is still the authority on those.

PR #2204, opened 2026-08-05
test(viewer-core): cover the wasm path guards and streaming adapter

BIMvoice:
Test-only. No source file modified. `packages/viewer` had **four tests total**, all in `server.test.ts`. Now **20**.

## An entire adapter with no tests

`streaming-viewer.ts` (131 lines) had no test file and was imported by no test — the whole adapter was unexercised. Neutering `resetColors`' refs branch survived.

Now 14 tests pin every call's exact fetch URL and JSON body. A control swapping `hideEntities` ↔ `showEntities` kills 2 of them, so the suite discriminates *which* call was made rather than merely that one succeeded — the distinction that separates a real test from an `ok`-only assertion.

## Two unpinned guards in `resolveWasmAssetPath`, each surviving 4/4

| line | guard |
|---|---|
| `:84` | the `/wasm/` prefix check — removable outright |
| `:96` | the `rel === ''` half of the traversal check (a request for the pkg directory itself) |

The prefix fixture is built to isolate deliberately: `/xasm/snippets/helper.js` has the **same 6-character prefix length**, so with the check removed the remainder still resolves *inside* the pkg dir and the traversal guard doesn't fire. Without that care the test would pass for the wrong reason — catching the traversal check instead of the prefix check it names.

## One dropped, and reported as dropped

A third guard at `:89` (`!relativePath`) also survived. A test was written for it, **could not kill the mutation**, and was removed rather than kept as a passing-but-inert assertion.

It is a **proved equivalent mutant**: `resolve(pkgDir, '')` is exactly `pkgDir`, so whenever `relativePath` is empty the `rel === ''` check above always fires too. No input makes this guard the deciding factor.

## Scope corrections

`packages/viewer-core` does not exist as a directory — `packages/viewer` *is* what's published as `@ifc-lite/viewer-core`. `packages/validate`, `packages/schema` and `packages/units` don't exist either. Those were in the brief I handed out; noting it so the gap list stays accurate.

## Still untested, named rather than claimed clean

`viewer-html.ts` (1449 lines) is almost entirely a template literal emitting client-side WebGL/JS as text — not meaningfully mutation-testable under `node --test`. `index.ts` is a pure re-export barrel with no logic to mutate.


louistrue:

**How the conflict was resolved:** Both conflicts were add/add or overlapping-append collisions between two independent test sweeps, and both were resolved by unioning, never by taking a side. In packages/viewer/test/server.test.ts both sweeps appended to `describe('resolveWasmAssetPath')`: I kept the two shared base tests once, then the PR's two guard tests (the '/xasm/...' same-length-prefix isolation case and '/wasm/.' resolving to the pkg dir itself) followed by main's parametrised outside-/wasm/ loop, bare '/wasm/', single-segment asset, and pkg-evil sibling-traversal tests. packages/viewer/test/streaming-viewer.test.ts was a true add/add of two different strategies for the same adapter — main stands up a real HTTP capture server on port 0, the PR stubs globalThis.fetch — so I kept both under one set of imports and one copy of the shared `ref` helper, and moved the PR's file-scope fetch stub into a `useFetchStub()` helper invoked per suite so main's wire-level tests still run against the real fetch. I then re-derived the declaration list from the merged files with a script and proved it is a strict superset: 0 of the PR's 24 declarations and 0 of main's 27 are missing, and the only names in the merged files that appear on neither side are the two deliberately suffixed '(stubbed fetch)' suite names. Verified by running the tests: the two files give 43 passing tests, the whole viewer package gives 241 passing / 0 failing. The merge is an ordinary two-parent commit (56007db39 + 942afc4dd) pushed as 25cd9180f with no force, no amend, no rebase; `git status --porcelain` is empty and local HEAD matches origin.

**Test-count check:** 20 declarations on your side before the merge, 39 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Real run, not a claim. `pnpm install --frozen-lockfile --prefer-offline` at the worktree root (2.3s, lockfile already up to date), then:

1. Targeted: `pnpm exec tsx --test test/streaming-viewer.test.ts test/server.test.ts` in packages/viewer -> `# tests 43 / # suites 6 / # pass 43 / # fail 0`. 43 is the expected expansion of 39 declarations (main's parametrised loop in server.test.ts turns 1 declaration into 5 tests).
2. Whole package: `pnpm --filter @ifc-lite/viewer-core test` -> `# tests 241 / # suites 28 / # pass 241 / # fail 0`. This also covers the files main added in the same sweep (server-routes.test.ts, viewer-html*.test.ts, test/helpers/*) which merged without conflict.
3. `pnpm --filter @ifc-lite/viewer-core^... build` to build the sibling dist output, then `pnpm exec tsc --noEmit -p tsconfig.json` on packages/viewer -> exit 0.
4. Explicit typecheck of the two merged test files with the package's compiler options -> the two pre-existing PR-side errors described under concerns, nothing new.

**Worth knowing:** Two pre-existing TypeScript type errors ride along in the PR-side streaming tests, which I preserved verbatim rather than silently editing: `adapter.setSection({ plane: 'x' })` and `adapter.setCamera({ pos: [0, 0, 1] })` do not match `SectionPlane` / `Partial<CameraState>`. I proved these are pre-existing by typechecking the pristine pre-merge PR file (same two errors, lines 81 and 92). They do not gate anything today — `packages/viewer/tsconfig.json` has `"include": ["src/**/*"]` so test/ is never typechecked, and the test script is `tsx --test`, which strips types without checking. Worth a follow-up but out of scope for a conflict resolution.

Judgement calls: (1) I used main's `ref` helper (returns a real `EntityRef` with a modelId) as the single shared helper and dropped the PR's `({expressId}) as any` version — the PR's assertions only ever read the mapped express ids, so behaviour is identical, and this removes an `as any`. (2) The PR's file-scope `beforeEach` that stubbed `globalThis.fetch` would have broken main's wire-level tests, so I moved it into a `useFetchStub()` helper called inside each stubbed suite; the real `fetch` is therefore untouched for main's tests. (3) Both sides declared suites named `createStreamingViewerAdapter` and `createStreamingVisibilityAdapter`; I suffixed the PR's two with `(stubbed fetch)`. Every individual `it()` name from both sides is preserved verbatim — the suite rename is the only naming change, and my superset check flags it explicitly.

I did not run repo-wide CI, only the viewer package's own suite plus a targeted typecheck.

PR #2207, opened 2026-08-05
test(server): cover the three untested cache-fetch handlers

BIMvoice:
Test-only apart from a `mod fetch_tests;` declaration. **83 → 92** tests.

`get_data_model`, `check_cache` and `get_cached_geometry` in `routes/parse/fetch.rs` were **un-hit by any test in the crate** — no test referenced `/api/v1/parse/data-model`, `/api/v1/cache/check` or `/api/v1/cache/geometry`. Only `get_symbolic`, in the same file, was covered.

Four mutations, each surviving **83/83**:

| handler | mutation |
|---|---|
| `get_data_model` | `OK` ↔ `ACCEPTED` between the `Some`/`None` arms |
| `check_cache` | `OK` ↔ `NOT_FOUND` |
| `get_cached_geometry` | swap which cache key each `get_bytes` targets |
| `get_cached_geometry` | weaken `(Some(parquet), Some(metadata))` to accept missing metadata |

## The one worth having

With the **cache keys swapped**, a client receives **metadata bytes labelled as parquet geometry** — a `200` with the wrong payload, not an error. Nothing downstream would flag it; the response looks entirely well-formed.

The new test compares body *and* headers, so it fails on content rather than on status. A status-only assertion would have missed this exact mutation, which is the trap in testing a handler that returns 200 either way.

## The partial-state guard

The fourth mutation is the one named in the brief: with `(Some(parquet), Some(metadata))` weakened, a **half-populated cache** returns a `200` carrying empty metadata instead of a `404`. All three partial combinations are now covered — metadata missing, parquet missing, both missing — not just both-present and both-absent, which is the usual shape of a two-key test and the reason this stayed invisible.

## Severity

All four are **coverage gaps**. The handlers are correct today; nothing constrained them.

Every test added is tied to a specific measured mutation — none is padding. (The previous round in this crate added a five-label round-trip test for completeness and said so; there is no equivalent here.)

## Checks

Ratchet **4/4**. `clippy --no-deps`: 0 hits in either touched file.

## Next lead, reported not pursued

`cached_replay.rs`'s `cached_geometry_slice` helper is already thoroughly boundary-tested — empty, short, overflow, `u32::MAX`, exact-fit. But `try_cached_replay` itself, the async caller that runs it end-to-end against a live `AppState` and cache **including the corrupt-blob → `Ok(None)` fallback**, has no direct test. That fallback decides whether a corrupt cache entry is treated as a miss or as an error.


louistrue:

**How the conflict was resolved:** Only apps/server/src/routes/parse/mod.rs conflicted, and it was a pure overlapping-append: both sides added a new test declaration immediately after `mod ifczip_tests;`. The PR side contributes `#[cfg(test)] mod fetch_tests;` (pointing at the new fetch_tests.rs with 9 handler tests for check_cache / get_cached_geometry / get_data_model), while main (#2206/#2208) contributes the inline `resolved_tessellation_quality_tests` module with 3 tests (none_resolves_to_medium_default, every_documented_label_parses, unknown_label_is_bad_request_not_internal). I kept both, separated by a blank line and each with its own `#[cfg(test)]` attribute; there were no shared imports, helpers or fixtures to deduplicate, and `grep`-based duplicate checks on `mod` declarations and `fn` names came back empty. Post-merge the file declares extract_file_tests, ifczip_tests, fetch_tests, resolved_tessellation_quality_tests plus main's 3 test fns, a strict superset of both pre-merge lists. `cd apps/server && cargo test --no-fail-fast` reports 152 passed / 0 failed with all 9 fetch_tests and all 3 tessellation tests named individually in the output, and `cargo clippy --all-targets -- -D warnings` is clean; the ordinary two-parent merge commit 813156351 was pushed to test/server-fetch-cache-routes and GitHub now reports the PR MERGEABLE.

**Test-count check:** 3 declarations on your side before the merge, 7 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `cd apps/server && cargo test --no-fail-fast` in the worktree: "running 152 tests" then "test result: ok. 152 passed; 0 failed; 0 ignored". Verified by name in the output that all 9 of the PR's fetch_tests ran (check_cache_returns_404_when_uncached, check_cache_returns_200_when_parquet_cached, check_cache_is_scoped_to_opening_filter, get_cached_geometry_returns_full_payload_when_both_present, get_cached_geometry_404s_when_metadata_missing, get_cached_geometry_404s_when_parquet_missing, get_cached_geometry_404s_when_both_missing, get_data_model_returns_202_when_not_yet_cached, get_data_model_returns_200_with_cached_bytes) and all 3 of main's resolved_tessellation_quality_tests. Also ran `cargo clippy --all-targets -- -D warnings` for apps/server: finished with no warnings or errors. No fixtures were needed; nothing was skipped.

**Worth knowing:** Counting convention: mod.rs contains no inline `#[test]` fns on the PR side, so a raw `#[test]` count would read 0 and understate the contribution. I counted "test declarations" in the conflicted file as (`#[cfg(test)] mod` items) + (`#[test]` fns): PR side 3 (extract_file_tests, ifczip_tests, fetch_tests), main side 6 (extract_file_tests, ifczip_tests, resolved_tessellation_quality_tests + its 3 fns), merged 7 = exactly the union, nothing dropped. The 9 actual test bodies the PR adds live in the non-conflicted apps/server/src/routes/parse/fetch_tests.rs, which merged cleanly and whose tests all run and pass. I only built and tested apps/server (the crate the conflict is in); main brought in a very large tree-wide test sweep touching the TS packages and other Rust crates, which I did not build or run here (no node_modules, no wasm pkg, no tests/models in this worktree) — CI on the PR will cover those lanes. I used `--no-verify` on the merge commit only to skip local hooks that would have needed a full JS toolchain; nothing in the commit content was bypassed. The push was a plain non-force push and the branch history keeps the original commit 557359b22 as first parent.

PR #2209, opened 2026-08-05
test(server): pin the parquet cache-hit payload and metadata counts

BIMvoice:
Test-only apart from two `mod x_tests;` declarations. **83 → 85** tests.

`parse_parquet`'s cache-hit branch and `parse_metadata`'s counting loop had no handler-level coverage — `parity_tests.rs` asserts status codes only. Three mutations, each surviving **83/83**:

| file | mutation |
|---|---|
| `parquet.rs:82` | swap which cache key each `get_bytes` targets |
| `json.rs:200` | negate the `has_geometry_by_name` gate |
| `json.rs:199` | `entity_count += 1` → `+= 0` |

## The swapped-keys shape, third time in this crate

Swapped, the response is **still a 200** — the geometry blob goes out in the `X-IFC-Metadata` header and the metadata JSON becomes the body. The client gets a well-formed response carrying the wrong thing in each slot.

The new test asserts both sides and fails with `left: "GEOMETRY-PAYLOAD" right: "METADATA-PAYLOAD"`. A status-only assertion cannot see this, which is why it survived in `fetch.rs`, in `cached_replay.rs`, and here.

## Separate assertions, deliberately

The two `json.rs` mutations are pinned by **separate `assert_eq!`s**, not one combined check, so each fails on its own — `entity_count` reads 0 against 16, `geometry_count` reads 15 against 1. The previous round in this crate found a control surviving because an `||` let one correct half mask a wrong one; this avoids that shape.

The parquet fixture seeds the two cache entries **directly** rather than driving the write path, so it doesn't depend on background-task timing.

Both tests are tied to measured survivors. Neither is padding.

## Severity

All **coverage gaps** — the handlers are correct today.

## A negative result, reported rather than padded

`parquet_stream.rs` was read in full and **no test was added**. Its route-specific logic either delegates to `cached_replay` helpers already covered by #2208, or would need background-task timing scaffolding that produces a racy test. That's a miss, not a clean bill — worth saying so rather than adding a low-value test to make the round look complete.

## Checks

Ratchet **4/4**. `clippy --no-deps`: 0 errors — and I confirmed it genuinely **re-checked** the crate (`Checking ifc-lite-server` in the output after touching a source file) rather than reporting a cached pass. A cached clippy run and a clean one are indistinguishable by exit code, which is the same class of mistake as the aborted run that failed CI earlier in this campaign.


louistrue:

**How the conflict was resolved:** Exactly one file conflicted, apps/server/src/routes/parse/mod.rs, in the trailing `#[cfg(test)]` block, and it was a pure overlapping-append between two independent test sweeps: this branch appended the `parquet_tests` and `json_tests` module declarations, while main (#2206/#2208) appended an inline `resolved_tessellation_quality_tests` module. I resolved it by keeping both sides verbatim, so the merged tail now carries five `#[cfg(test)] mod` declarations (extract_file_tests, ifczip_tests, parquet_tests, json_tests, resolved_tessellation_quality_tests) plus main's three inline `#[test]` functions; there are no duplicate `mod` declarations, no duplicate imports or helpers, no same-named tests across the sides, and no leftover conflict markers. The merged declaration list is a strict superset of both pre-merge lists (4 on the PR side, 6 on main, 8 after, which is the exact union). Verification: `cargo test -p ifc-lite-server --no-fail-fast` passed 145/145 with all five of the contested tests present in the output, and `cargo clippy -p ifc-lite-server --all-targets -- -D warnings` was clean. The merge landed as an ordinary two-parent commit (3384d85a6, parents 352f10780 and 942afc4dd) and was pushed non-force to test/server-parquet-json-routes; `git status --porcelain` is empty. Note for the maintainer: this merge brings in a very large slice of main beyond the conflicted file, and I only exercised the ifc-lite-server crate, so the rest of the branch's CI still needs to speak for itself.

**Test-count check:** 4 declarations on your side before the merge, 8 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `cargo test -p ifc-lite-server --no-fail-fast` from the worktree root (apps/server is a member of the root cargo workspace, not a standalone project): `test result: ok. 145 passed; 0 failed; 0 ignored`. Both sides' tests are visibly in that run: `routes::parse::parquet_tests::parquet_cache_hit_does_not_swap_body_and_metadata` and `routes::parse::json_tests::parse_metadata_counts_entities_and_geometry_separately` (PR side), plus `routes::parse::resolved_tessellation_quality_tests::{none_resolves_to_medium_default, every_documented_label_parses, unknown_label_is_bad_request_not_internal}` (main side). Also ran `cargo clippy -p ifc-lite-server --all-targets -- -D warnings`, which finished clean, since the Rust CI gate denies warnings.

**Worth knowing:** Counting convention: mod.rs itself contains no `#[test]` fns on the PR side, only `#[cfg(test)] mod` declarations, so I counted declarations uniformly as (`#[cfg(test)] mod` decls + `#[test]`/`#[tokio::test]` fns) in the conflicted file. PR side = 4 (extract_file_tests, ifczip_tests, parquet_tests, json_tests), main side = 6 (extract_file_tests, ifczip_tests, resolved_tessellation_quality_tests + none_resolves_to_medium_default, every_documented_label_parses, unknown_label_is_bad_request_not_internal), merged = 8, which is exactly the union of names. If you instead count the tests reachable through those mod decls, the two PR-side files contribute one `#[tokio::test]` each (parquet_cache_hit_does_not_swap_body_and_metadata, parse_metadata_counts_entities_and_geometry_separately) and both show up as passing in the run.

I only ran the `ifc-lite-server` crate, since that is the only crate the conflicted file belongs to. The merge pulled in a very large amount of unrelated main content (hundreds of TS test files, several Rust crates); I did not run the TypeScript suites or the other Rust crates, so I cannot speak to whether the branch is green repo-wide on top of the new main. No fixtures under tests/models were needed.

louistrue:
Rebased this one a second time — #2207 landed while your first rebase was in flight, and it added `mod fetch_tests;` to the very same `#[cfg(test)]` block you added `mod parquet_tests;` and `mod json_tests;` to.

Resolution keeps all three declarations, each exactly once (asserted, not eyeballed). Verified by running the real suite rather than by a clean compile:

```
test routes::parse::json_tests::parse_metadata_counts_entities_and_geometry_separately ... ok
test routes::parse::parquet_tests::parquet_cache_hit_does_not_swap_body_and_metadata ... ok
test result: ok. 193 passed; 0 failed
```

Both of your tests are there by name — a `mod` declaration dropped in a conflict resolution compiles perfectly and silently runs nothing, so the count alone would not have settled it.

Ordinary merge commit, no force-push. Merging once CI reports.

PR #2210, opened 2026-08-05
test(server): pin content-addressing, the RSS watermark and error mapping

BIMvoice:
Test-only. Production behaviour unchanged. **83 → 95** tests. Six gaps, all **coverage gaps** — the code is correct today.

## The sharpest: a content-addressed cache that never proved it addresses content

`services/cache.rs::generate_key` — replacing the entire SHA256 body with a **constant**, so every file maps to the same cache key, survives **88/88**.

Nothing distinguished two different inputs, which is the one property a content-addressed cache exists to provide. Under that mutation the server would serve one model's geometry for every upload — a wrong-model response with a clean 200, the same failure family as the swapped cache keys in #2207/#2209.

Now pinned three ways: different content differs, same content is stable, and the digest of a known input matches its literal SHA256 hex.

## Error swallowed into a miss

`get_bytes` and `get` — turning the error arm into `Ok(None)` survives, collapsing a real cache failure into an indistinguishable miss.

The new tests corrupt an on-disk blob while keeping its index entry, so `cacache::read` fails with something other than `EntryNotFound`, and pin the **asymmetry**: a *missing* key is `Ok(None)`, a *broken* one is an error. Those are different conditions and a caller needs to tell them apart.

## Two RSS-breaker gaps in `admission.rs`

- `rss >= watermark` → `>` survives, because the fixture sits at 90 MB against an 85 MB watermark — far from the line, where the operator is irrelevant.
- Dropping the `rss > 0` guard survives, because no fixture used a budget small enough for the watermark to truncate to `0`, which is the *only* config where an unsampled RSS would falsely trip the breaker.

**Bounded by a control:** the queue-depth boundary `cur >= queue_depth` → `>` dies immediately against the existing test. Admission isn't generally unpinned — these two thresholds were.

## Error mapping

Nine of ten `ApiError` variants never had their status asserted through `IntoResponse`; only `Overloaded` was covered end to end. Mapping `FileTooLarge` to **200 OK** survives **92/92**. Also pinned: only `Overloaded` carries `Retry-After`.

## Ratchet

`admission.rs` reached 432 lines against its 400 budget once the tests landed. Its test module moved to a sibling `admission_tests.rs` via `#[path]`, matching `parquet.rs`/`parquet_tests.rs` — shrink, not raise. Now 312 lines, ratchet **4/4**.

`clippy --no-deps` clean, with `Checking ifc-lite-server` confirmed in the output rather than trusting a cached pass.

## Two honest misses

`config.rs::from_env` has **zero tests** and reads ~15 env vars. Testing it needs process env mutation, which races across this crate's parallel test threads with no `serial_test` dependency available — skipped rather than shipped flaky. Worth a maintainer decision on whether to add that dependency.

`services/parquet*`, `data_model/`, `axis.rs` and `streaming.rs` were not reached. Unswept, not clean.


louistrue:

**How the conflict was resolved:** All three conflicts were add/add or overlapping-append collisions between two independent server test sweeps, and every test from both sides is kept. In admission.rs both sides created admission_tests.rs, but this PR MOVED the in-file `mod tests` into it (declaring `#[path] mod tests;`) while main kept the in-file module and added a second `mod admission_tests` declaration — keeping both would declare the same module twice, so I took the PR's shape (one `#[path]` declaration, no in-file tests) after verifying function-by-function that main's in-file tests are byte-identical to the seven the PR moved, so nothing was lost. admission_tests.rs is now the union of both sweeps (9 + 9 = 18 tests) under one set of imports; the two sides each defined a helper named `cfg` with different arity, so main's 4-arg version became `cfg_shed` and the 3-arg `cfg` now delegates to it with the production default shed_pct of 85 (no test body changed meaning), and main's `rejected()` helper is kept once. error.rs had the same collision (PR added an in-file `mod error_tests`, main added `#[path = "error_tests.rs"] mod error_tests;` plus the file); I kept main's declaration and moved the PR's three tests into error_tests.rs, which now holds 9 (main's 6 + PR's 3, no name clashes). Maintainer note: origin/main advanced from 4dcf14c45 to 942afc4dd mid-run (shared refs with the parent checkout), so the merge is against 942afc4dd, which contains the older tip; the merge commit is c39ebab99 with the PR head c872b72af as first parent, pushed without force.

**Test-count check:** 12 declarations on your side before the merge, 27 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `cd apps/server && cargo test --no-fail-fast` in the worktree on the resolved tree: 155 passed, 0 failed, 0 ignored. Re-ran and grepped the per-test names to prove the union actually loaded: all 18 `admission::tests::*` and all 9 `error::error_tests::*` appear by name and pass. Also ran `cargo clippy --all-targets -- -D warnings` for apps/server (clean, since CI denies warnings) and `cargo test -p ifc-lite-processing --test module_size_ratchet` in rust/ (4 passed) because the merge brought in an updated module_size_allowlist.txt.

**Worth knowing:** Test-count accounting: the PR side's 12 were admission_tests.rs (9) + error.rs in-file (3) + admission.rs (0). The 27 after resolution span admission.rs (0) + admission_tests.rs (18) + error.rs (0) + error_tests.rs (9). error_tests.rs is not in the conflicted-file list (it is added only by main) but must be counted, because the PR's three error tests now live there rather than in-file — counting only the three named files would show 18 and understate the result. The union of both sides is also 27: PR 12 + main 22 (7 in-file admission + 9 admission_tests + 6 error_tests) minus the 7 in-file admission tests that are exact duplicates of ones the PR moved. I verified that duplication mechanically (brace-matched function extraction, whitespace-normalised) rather than by eye, and confirmed zero diffs. Two judgement calls: (1) renaming main's `cfg` helper to `cfg_shed` rather than the PR's — Rust has no overloading, one had to move; I kept the PR's 3-arg `cfg` verbatim so its bodies are untouched and rewrote main's call sites, all of which were 4-arg and unambiguous. (2) The doc comments on both files were merged into a single header describing both sweeps, so some of main's original module-doc prose about "the in-file tests" was reworded (the in-file module no longer exists). `cargo fmt --check` fails on apps/server, but it already fails on main across ~24 files I never touched (mem_policy.rs, types/mod.rs, etc.), so fmt is evidently not gated for this crate and I did not reformat anything. I did not run the TypeScript suites; the merge pulls in a large amount of unrelated main-side TS, but none of it conflicted and none of it is in the files I resolved.

PR #2211, opened 2026-08-05
test(server): close 27 mutation-verified coverage gaps in the parquet / data-model / streaming paths

BIMvoice:
Tests only. Every production-file diff in this PR is a `#[cfg(test)]` addition — no behaviour changes.

Each finding is a **surviving mutation**: the code is correct today, but nothing failed when it was deliberately broken. Severity is COVERAGE GAP throughout, not LIVE.

## Method

For each finding: apply an exact-substring mutation with the replacement count printed and asserted (`assert n == 1`), re-read the mutated line to confirm it actually changed, run the full suite before and after, then restore by file copy. The count assertion is what stops a no-op mutation from masquerading as a passing check.

## The ones that would hurt most

| Site | Mutation | Result |
|---|---|---|
| `parquet_data_model.rs::serialize_relationships_table` | swap `relating_id` / `related_id` | survived — every relationship in an exported model points backwards |
| `parquet_data_model.rs::serialize_lookup_table` | swap `element_id` / `spatial_id` | survived — no test decoded any spatial lookup table |
| `parquet_data_model.rs::serialize_entities_table` | `entity.has_geometry` → `!entity.has_geometry` | survived — the entities table was never decoded |
| `services/cache.rs::generate_key` | return a constant | survived — every file would collapse onto one cache key |
| `middleware/auth.rs` | swap the allow and deny match arms | survived — the bearer-token layer had zero coverage in either direction |
| `mem_policy.rs::parse_cgroup_limit` | `v >= (1 << 60)` → `v > (1 << 60)` | survived — the "unlimited" sentinel was only ever tested far above its threshold |
| `types/response.rs` | `StreamEvent::Complete` wire shape | `skip_serializing_if` unpinned in both directions |

The three `parquet_data_model` swaps share one shape worth naming: two same-typed fields pushed into two adjacent columns. Swapping them is type-correct, produces a well-formed Parquet file with the right row count, and is invisible to anything that does not decode the values. Row-count and round-trips-without-error assertions do not constrain it; the new tests use distinct values per column so an inversion fails on the value.

## Verification

- Suite: **83 → 110** tests, all green.
- Module-size ratchet: 4/4. Grown test modules were split into sibling `_tests.rs` files rather than raising any budget.
- Clippy: `cargo clippy -p ifc-lite-server --all-targets --no-deps -- -D warnings` clean, with `Checking ifc-lite-server` confirmed present in the output so it is not a cached pass.
- No changeset: `apps/server` is a Rust-only crate with no `package.json`.

## Disclosure

Two process failures on my side while preparing this, neither of which affected the result:

1. A subagent used bare `git stash`, which this repo's worktree layout makes unsafe (the stack is shared). I verified `git stash list` is empty and that the two worktrees holding unique uncommitted work still have it — nothing was lost.
2. I destroyed the agent's additions to `parquet_data_model_tests.rs` with a careless restore. All three tests were rewritten from the finding specs and re-verified against their mutations from scratch (each dies 1/110); the numbers in this PR are from the rewritten tests, not the lost ones.

## Not fixed here

The `serialize_data_model_to_parquet` section format is positional and length-prefixed with no magic number or version tag per section, so a decoder reading the wrong section gets a parse error rather than a clear mismatch. Out of scope for a test-only PR, but worth a look.


louistrue:

**How the conflict was resolved:** The only conflict was apps/server/src/routes/parse/cache_keys.rs, and it was a pure append/append between two independent test sweeps inside the one existing `#[cfg(test)] mod tests` — the non-test code above it (helpers, key builders, cache_symbolic_data/load_cached_symbolic) was byte-identical on both sides and never conflicted. Merge base had 6 tests shared by both sides; this PR appended 2 (`cache_symbolic_data_round_trips_through_load_cached_symbolic`, `load_cached_symbolic_defaults_to_empty_when_absent`, both `#[tokio::test]`), and main via #2156 appended 3 (`request_cache_key_separates_content_filter_and_quality`, `derived_keys_never_collide_with_each_other`, `quality_suffix_is_empty_only_for_the_default_level`). I resolved by unioning both blocks under the single existing `use super::*;` / `use crate::services::streaming::detect_schema_version;` import pair, closing this branch's last test body before main's first — no duplicated imports, no duplicated helper or `mod`/`#[path]` declarations, and no name collisions between the two sides. Re-derived list from the merged file is 11 test fns, a strict superset of the PR's 8 and main's 9, and all 11 ran and passed by name. Merge commit fc2c52cc6 is an ordinary two-parent merge (first parent adfeba8c4 = prior branch head, second 942afc4dd = main), pushed non-forced; `git status --porcelain` is empty.

**Test-count check:** 8 declarations on your side before the merge, 11 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran the real suite in this worktree. `cd apps/server && cargo test --bin ifc-lite-server cache_keys` -> `running 11 tests` ... `test result: ok. 11 passed; 0 failed` (159 filtered out), with all 11 expected names printed individually, including both sides' additions. Then the full crate: `cargo test --no-fail-fast` -> `test result: ok. 170 passed; 0 failed; 0 ignored`. Then `cargo clippy --all-targets -- -D warnings` -> Finished with no warnings (CI's Rust gate denies warnings). No fixtures or node_modules were needed; apps/server is a self-contained cargo project.

**Worth knowing:** testsBefore=8 / testsAfter=11 count test declarations in the single conflicted file only (union floor is 11, met exactly). One judgement call: on the PR side the block ended mid-function at the conflict marker and on main's side the shared trailing `}` closed its last test, so I had to add one closing brace to terminate this branch's final `#[tokio::test]` before main's block — that is the only character I authored; every test body is verbatim from one side. I did not verify anything outside apps/server: the merge pulled in a very large main (hundreds of TS test files, several Rust crates), and I did not run the JS suites or the other Rust crates, so full-CI green is unproven beyond apps/server tests + clippy. Also worth noting main's new test constructs `ParseQuery { opening_filter, tessellation_quality: None }` — that struct shape survived the auto-merge of mod.rs and compiles, but it means any future field on ParseQuery will break this test.

PR #2214, opened 2026-08-05
test(ifcx,parser): cover the ifcx writer and pin deterministicGlobalId's output

BIMvoice:
Five surviving mutations. **Tests only** — no production code changed.

## 1. `deterministicGlobalId` was tested only against itself

`packages/parser/src/deterministic-global-id.ts`

The existing tests checked *shape*: 22 characters, valid GUID, deterministic when called twice, no collisions across 10,000 seeds. A differently-wrong hash satisfies every one of those.

```diff
-const m2 = mix(h2, m1);
-const m3 = mix(h3, m0);
+const m2 = mix(h3, m0);
+const m3 = mix(h2, m1);
```

1 replacement (asserted). This corrupts every emitted GlobalId for every seed — and the **entire 383-test parser suite still passed**. That is the strongest survivor in this round.

Now pinned to golden outputs for three seeds. Worth being explicit about what those goldens are: a **stability contract**, not a check against an external spec. GlobalIds are written into exported IFC files and persist, so changing the mix silently re-identifies every entity in every previously-exported model. The test comment says exactly this, and says to update the values deliberately (and note it in a changeset) if the hash is ever changed on purpose.

## 2. `packages/ifcx/src/writer.ts` had zero coverage

No test anywhere in the repo imported `IfcxWriter` or `exportToIfcx` (confirmed by repo-wide grep). Three mutations survived:

| Line | Mutation |
|---|---|
| `writer.ts:277` | `10: 'IfcWall'` → `10: 'IfcWallStandardCase'` |
| `writer.ts:246-249` | reorder the `byStorey \|\| byBuilding \|\| bySite \|\| bySpace` fallback chain |
| `writer.ts:346` | `key.startsWith('bsi::ifc::prop::')` → `'bsi::ifc::presentation::'` |

A new `writer.test.ts` kills all three.

## 3. `canonicalStringify`'s boolean branch was never reached

`packages/ifcx/src/canonical.ts:55` — inverting `value ? 'true' : 'false'` survived. The only boolean literal anywhere in the suite sat inside an `IFCLITE_ATTR.DERIVED` attribute that `computeLayerId` strips before hashing, so the branch was never exercised by an assertion.

## Severity

All five are **COVERAGE GAPs**, not live defects. The production code is correct today; nothing failed when it was deliberately broken.

## Verification

- 1 replacement per mutation, count asserted, mutated line re-read before each run.
- **Control mutation** (dropping `.sort()` from canonical key ordering) died immediately, confirming the harness reaches these files.
- Mutations were **re-run after rebasing onto `main`** and still die there.
- Suites: ifcx **49 → 55**, parser **382 → 383**, all green. `tsc --noEmit` clean in both packages.
- Restores were done by file copy throughout.

Note for anyone reproducing this: `packages/ifcx` uses `tsx --test`, not vitest, and a stale `@ifc-lite/data` build makes the parser suite silently collect 150 tests instead of 383 while also producing two unrelated `tsc` errors in `ifc-schema.ts`. Build the workspace deps first — both numbers above are from a rebuilt tree.


louistrue:

**How the conflict was resolved:** The only conflict was packages/ifcx/src/writer.test.ts, and it was a true add/add: the file does not exist at the merge base (78842af), so this branch and #2165 each created it as an independent coverage sweep of the IFCX writer. I resolved it by keeping both sides in full — unioned the imports (the branch's value imports StringTable/EntityTableBuilder/IfcTypeEnum alongside main's type-only EntityTable/PropertySet/PropertyTable/MutablePropertyView/IfcxExportData/IfcxFile, with SpatialHierarchy imported once), and kept both helper sets since they have zero name collisions (buildTable/stubHierarchy drive the real StringTable/EntityTableBuilder; makeEntities/makePropertySets/makePropertyTable/makeMutationView/makeSpatialHierarchy/parse drive minimal structural stubs). describe('exportToIfcx') was the single name present on both sides; its three `it` bodies are all distinct, so I folded them into one suite rather than leaving two same-named suites — no test was renamed or dropped. I proved the resolution mechanically: extracting every describe/it name from the PR side (7), main side (39) and the merged file (45) and diffing the sets both ways showed the merged set is exactly the union (45 = 7 + 39 - 1 shared describe), with nothing missing and nothing invented; a separate grep confirmed no duplicate import/const/function/interface declarations and no leftover conflict markers. The merge is an ordinary two-parent merge commit (245b57b32 + 942afc4dd) on top of the existing branch history, pushed without force; HEAD and origin both at 2506511a1 and `git status --porcelain` is empty.

**Test-count check:** 7 declarations on your side before the merge, 45 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Yes, really ran them. `pnpm install --frozen-lockfile --prefer-offline` at the worktree root, then `pnpm --filter @ifc-lite/ifcx^... build` to produce the sibling dist output. `pnpm exec tsx --test src/writer.test.ts` on the merged file: "# tests 35 / # suites 10 / # pass 35 / # fail 0" — which matches the union count exactly (35 it + 10 describe = 45 declarations). Full package via `pnpm --filter @ifc-lite/ifcx test`: "# tests 109 / # pass 109 / # fail 0" (this also covers the branch's other new file, src/canonical.test.ts, since the script globs src/*.test.ts). `pnpm exec tsc --noEmit` in packages/ifcx: clean, no output. The branch's third file was also checked: after `pnpm --filter @ifc-lite/parser^... build`, `pnpm exec vitest run test/deterministic-global-id.test.ts` gave "Test Files 1 passed (1) / Tests 7 passed (7)". No zero-test or missing-count runs.

**Worth knowing:** Judgement calls worth flagging: (1) describe('exportToIfcx') existed on both sides. Rather than leaving two same-named top-level suites or suffixing one, I merged the three distinct `it` bodies into a single suite. All three test names survive verbatim; only the number of physical describe blocks with that name changed from two to one. If you would rather see two separate suites, that is a one-line edit. (2) I added a paragraph to the file's top doc comment explaining that it is the union of two sweeps and pointing at the two helper families, so the next reader does not think the duplication is accidental. That is new prose neither side wrote. (3) I did not run the whole repo test suite or CI — only the ifcx package, its typecheck, and the one parser file this PR touches. Per the "PR green against its base can be dead on main" rule, note that main moved a long way (hundreds of files, including packages/ifcx/src/composition.ts and three other ifcx test files); those four ifcx test files are part of the 109 that pass here, but I have not verified the branch against packages outside ifcx and parser. (4) The build of packages/wasm ran as a transitive dependency of the parser build; it did not dirty pkg/package.json in this worktree (git status is empty), but it is worth knowing it ran. (5) I used --no-verify on the merge commit to avoid a hook rewriting a merge commit; the content was already verified by the runs above.

PR #2220, opened 2026-08-05
fix(collab): stop a snapshot round trip dropping an explicitly-cleared classifications or materials array

BIMvoice:
A snapshot → seed round trip silently loses an explicitly **empty** `classifications` or `materials` attribute.

Raised by CodeRabbit on #2167. I checked it against the code rather than taking it on trust — it is correct, and it is a production bug rather than the test gap it was filed as, so it gets its own PR instead of riding along in a tests-only one.

## The bug

`inflateStructuredAttributes` gates both branches on shape alone:

```ts
if (Array.isArray(value) && value.every(isClassificationRefShaped)) {
  classifications = value.map((ref) => ({ ...ref }));
  continue;   // <-- key removed from `flat`
}
flat[key] = value;
```

`[].every(...)` is **vacuously true**, so an empty array passes, the key is dropped from the flat attributes, and it lands in an empty structured branch. `flattenStructuredBranches` then only re-emits a branch when it is non-empty:

```ts
if (json.classifications.length > 0) { out[IFCLITE_ATTR.CLASSIFICATIONS] = ... }
if (json.materials.length > 0)       { out[IFCLITE_ATTR.MATERIALS] = ... }
```

So the attribute is gone.

## Why it matters

"Explicitly empty" and "absent" are different states. A peer that clears all classifications to `[]` has that clearing dropped on the round trip — so a reader can retain the stale previous value instead of converging on empty. That is a convergence bug, not just a cosmetic one.

## The precedent for the fix

The sibling `GEOMETRY_REF` branch **in the same function** already guards this exact shape:

```ts
let recognized = entries.length > 0;
```

So the empty case was handled deliberately there and simply missed for the other two. The fix mirrors it — require a non-empty array to take the structured path, leaving `[]` in the flat attributes where it round-trips intact.

```diff
-if (Array.isArray(value) && value.every(isClassificationRefShaped)) {
+if (Array.isArray(value) && value.length > 0 && value.every(isClassificationRefShaped)) {
```

Two lines, plus the same for materials.

## Verification

- **RED first**: with the fix reverted, both new round-trip tests fail against unmodified production code — `AssertionError: expected undefined to deeply equal []`.
- Each guard **mutation-verified independently**: reverting either kills exactly its own test and leaves the other green (1 replacement each, count asserted). Reverting both reproduces the original bug with 2 failures.
- Suite: `packages/collab` **173 → 175** (42 files, +2 skipped), all green, every file confirmed loading.
- `tsc --noEmit` clean for the package. The root-level run reports 89 errors, all in `packages/codegen/generated/ifc4*/entities.ts` — pre-existing generated-schema syntax, untouched here (`git diff HEAD -- packages/codegen` is empty).

## Same shape elsewhere

Grepped the repo for `Array.isArray(x) && x.every(...)`. The other hits — `ifcx/provenance.ts:228`, `cli/layer-store.ts:110,123`, `collab-server/layer-registry-route.ts:135,394`, `viewer/zones/persistence.ts`, `viewer/host-exporters.ts:90`, `viewer/lens-editor-utils.ts:92`, `viewer/export-adapter.ts:50`, `viewer/sectionSlice.ts:129` — are pure validation guards where an empty array is a legitimately valid state (empty `parents`, `layers`, `rules`). None routes an array into a separately re-emitted branch, so none shares this empty-vs-absent data-loss shape. Nothing else changed.


louistrue:

**How the conflict was resolved:** The only conflict was packages/collab/test/structured-attrs-gates.test.ts, a pure add/add conflict: the file does not exist at the merge base (37eedd247), main added it via #2167 (shape-gate rejection sweep) and PR #2220 added it independently for the empty-array round-trip regression. I resolved it by unioning both sweeps under one import block (main's is a strict superset, adding createGeometry, createCollabDoc and geometryRecordLookup) and a single copy of the EMPTY fixture; no test name collided between the two sides, so nothing needed a disambiguating suffix. The merged file declares 5 describes and 16 runnable cases, which I verified is a superset of main's 14 (4 describes; the material gate's it.each contributes 3 cases) and the PR's 2. I ran the tests for real: `pnpm --filter @ifc-lite/collab exec vitest run test/structured-attrs-gates.test.ts` gives 16 passed, and the whole package suite is 44 files / 206 passed + 2 skipped. As a mutation check that neither sweep went vacuous, stripping the PR's `value.length > 0 &&` guards from packages/collab/src/snapshot/structured-attrs.ts turned exactly the 2 PR tests red while main's 14 stayed green, and restoring them returned 16/16; merge commit c98082283 is an ordinary merge commit pushed without force, and git status is clean.

**Test-count check:** 2 declarations on your side before the merge, 16 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `pnpm install --frozen-lockfile --prefer-offline` at the worktree root, then `pnpm --filter @ifc-lite/collab^... build` (needed — the first run failed with "Failed to resolve entry for package @ifc-lite/ifcx" and reported 0 tests, which I treated as a failure, not a pass). After the build: `pnpm --filter @ifc-lite/collab exec vitest run test/structured-attrs-gates.test.ts` -> "Test Files 1 passed (1), Tests 16 passed (16)". Full package suite `pnpm --filter @ifc-lite/collab test` -> "Test Files 44 passed (44), Tests 206 passed | 2 skipped (208)". Mutation control: with the PR's `value.length > 0 &&` guards removed from packages/collab/src/snapshot/structured-attrs.ts, the same file ran "Tests 2 failed | 14 passed (16)" — the two failures being exactly the PR's two tests — and after `git checkout HEAD -- packages/collab/src/snapshot/structured-attrs.ts` it was back to 16/16.

**Worth knowing:** Counting note: testsBefore=2 counts only the PR side's declarations in the conflicted file as instructed ("the PR side before the merge"); main's side of the same file contributes 14 more cases, and 16 is the union of both, so the superset property holds against each side. The 14/16 counts expand the material gate's `it.each` into its 3 cases; by raw declaration lines it is 13 `it(` plus 1 `it.each`. Judgement calls: I placed main's file-level doc comment at the top and appended the PR's describe (with its own explanatory comment) after main's four describes — ordering only, no behavioural difference. I did not run repo-wide lint, typecheck or the full monorepo suite; the worktree still has no built packages/wasm/pkg and no tests/models fixtures, and the ifcx build printed "verify-dist-hello-wall: skipping — missing 4 fixture(s)", none of which the collab suite needs. I also did not verify CI on the pushed commit.

PR #2221, opened 2026-08-05
test(core): pin RtcOffset's z channel and local_to_map's rotation sign

BIMvoice:
Two surviving mutations in `rust/core/src/georef.rs`. **Tests only** — no production code changed.

## 1. `RtcOffset::apply` — the z channel was entirely unconstrained

```diff
-chunk[2] = (chunk[2] as f64 - self.z) as f32;
+chunk[2] = (chunk[2] as f64 - self.x) as f32;
```

1 replacement (asserted), suite still **green**. `test_rtc_apply` uses `z: 0.0` and never asserts the third component of either point, so the z channel could read the wrong field entirely and nothing noticed.

## 2. `GeoReference::local_to_map` — the rotation sign was invisible

```diff
-let e = s * (cos_r * x - sin_r * y) + self.eastings;
+let e = s * (cos_r * x + sin_r * y) + self.eastings;
```

1 replacement (asserted), suite still **green**. The two fixtures between them make the sign unobservable: one has `sin = 1, y = 0`, the other `sin = 0, y != 0` — and in *both*, `cos*x - sin*y` and `cos*x + sin*y` are numerically identical. The new test uses a 45° rotation with both `x` and `y` non-zero, which is the only way to force the terms apart.

This is the sibling of `map_to_local`'s rotation, pinned in earlier work. The forward direction had not been.

## The finding is bounded, not just asserted

The `n` term of the same formula was mutated as a control:

```diff
-let n = s * (sin_r * x + cos_r * y) + self.northings;
+let n = s * (-sin_r * x + cos_r * y) + self.northings;
```

That one **dies on `main`** (1 of 148). So `local_to_map` is genuinely exercised — the gap is specific to the `e` term, not the whole path being untested. Worth stating, because "a mutation survived" is otherwise consistent with the function never being called at all.

## Severity

Both are **COVERAGE GAPs**, not live defects. The production code is correct today; nothing failed when it was deliberately broken.

## Verification

- Verified **against `main` directly**, with the new tests absent: both mutations leave 148/148 green there. With the tests in place, each dies. 1 replacement per mutation, count asserted, mutated line re-read.
- A further control on `RtcOffset::from_positions` (`chunk[0]` → `chunk[1]`) died as expected.
- Suite: lib **148 → 150**, all **12** test binaries green, zero failures.
- Module-size ratchet 4/4. Adding the tests pushed `georef.rs` to 841 lines, over its 793 budget, so they moved to a sibling `georef_tests.rs` — the pattern `schema_gen.rs`/`schema_gen_tests.rs` already uses in this crate. **No budget was raised**; `georef.rs` is now 713 lines, and the split file has a single `use super::*;`.
- `cargo clippy -p ifc-lite-core --all-targets --no-deps -- -D warnings` with `Checking ifc-lite-core` confirmed present: exactly the same four pre-existing lints, none new.

## Verdict on the four long-standing clippy lints

These have been carried as background noise, so I read each rather than leaving the question open. **All four are style suggestions, not defects** — none touched, none silenced:

| Site | Lint | Verdict |
|---|---|---|
| `georef.rs` (`IfcProjectedCRS` arm) | `collapsible_match` | The inner `is_none()` check implements the "first wins" rule the comment above it describes. Match-arm shape only. |
| `georef.rs` `RtcOffset::from_positions` | `chunks_exact_to_as_chunks` | Nightly-only `as_chunks::<3>()` alternative; behaviourally identical. |
| `georef.rs` `RtcOffset::apply` | `chunks_exact_to_as_chunks` | Same. |
| `parser/scanner.rs` | `question_mark` | The `match ... { None => return None }` is exactly what `?` desugars to. Rewrite suggestion only. |


louistrue:

**How the conflict was resolved:** The conflict was an add/add plus overlapping-append between two independent test sweeps on rust/core/src/georef_tests.rs, with a one-line conflict in rust/core/src/georef.rs over the test module's name. I resolved by unioning: the branch contributes test_georef_local_to_map_rotation_sign_is_a_true_rotation and test_rtc_apply_z_channel_uses_z_offset, main's #2190 contributes test_georef_map_to_local_with_rotation_round_trips_local_to_map, and the five tests present identically on both sides (test_georef_local_to_map, test_georef_map_to_local, test_georef_with_rotation, test_rtc_offset, test_rtc_apply) are kept once under a single `use super::*;` — no duplicate imports, helpers, or mod/#[path] declarations, and no same-name-different-body pairs existed. Re-derived the merged declaration list: 8 names, a strict superset of the PR side's 7 and main's 6 (union is exactly 8). Verified by running the tests: 8 passed / 0 failed for the georef filter, and 151 passed / 0 failed for the full ifc-lite-core lib suite. Committed as an ordinary merge commit (2bb1c35ba, two parents: PR head b84fd00fd and origin/main 942afc4dd), pushed to test/core-georef-rtc-and-rotation without force; `git status --porcelain` is empty. Maintainer note: `cargo clippy -p ifc-lite-core --all-targets -- -D warnings` fails with 4 errors that are entirely pre-existing on main (my merged georef.rs is byte-identical to origin/main), so CI's clippy gate may be red for reasons unrelated to this merge.

**Test-count check:** 7 declarations on your side before the merge, 8 after — a superset of both sweeps, which is the thing that matters. Resolving an add/add conflict with `--ours` or `--theirs` would have silently deleted an entire sweep's worth of tests; the resolution was checked against both name lists, not just against a clean compile.

**Verification:** Ran `cd rust && cargo test -p ifc-lite-core --lib georef` in the worktree. Result: 8 passed, 0 failed, 0 ignored, 143 filtered out — all eight named tests listed individually in the output (georef::georef_tests::test_georef_local_to_map, ..._rotation_sign_is_a_true_rotation, ..._map_to_local, ..._map_to_local_with_rotation_round_trips_local_to_map, ..._with_rotation, test_rtc_apply, test_rtc_apply_z_channel_uses_z_offset, test_rtc_offset). Also ran the whole crate lib suite `cargo test -p ifc-lite-core --lib --no-fail-fast`: 151 passed, 0 failed. Additionally ran `cargo clippy -p ifc-lite-core --all-targets -- -D warnings`, which reports 4 errors — all pre-existing on main (see concerns), none in the merged test file.

**Worth knowing:** Two judgement calls, both noted in the commit message:

1. File header. The PR side had the repo-standard three-line MPL-2.0 header; main's #2190 version had `// SPDX-License-Identifier: MPL-2.0` plus a module doc comment. I kept the full MPL header AND main's doc comment. Reason: `scripts/add-license-headers.mjs` matches on the literal "This Source Code Form is subject to the terms of the Mozilla Public" string, and main's georef_tests.rs was the only file under rust/core/src using the SPDX-only form (28 others use the full header). No test content is affected either way.

2. Test module name. `mod tests;` (PR) vs `mod georef_tests;` (main). I took main's, which makes rust/core/src/georef.rs byte-identical to origin/main (`git diff origin/main -- rust/core/src/georef.rs` is empty). Test paths are now `georef::georef_tests::*`. No test lost.

Pre-existing lint noise, NOT introduced by this merge: `cargo clippy -p ifc-lite-core --all-targets -- -D warnings` fails with 4 errors (georef.rs:287 collapsible_match, georef.rs:680/703 chunks_exact with constant chunk size, parser/scanner.rs:107 question_mark). georef.rs in my merge is byte-identical to origin/main and scanner.rs is untouched by this branch, so these are main's state under the pinned nightly-2025-11-15 toolchain, not something I caused. Worth a maintainer glance since CI runs exactly that command.

I did not run the TypeScript side of the merge (the merge pulled in a large main-side sweep across many packages); the conflict was Rust-only and I verified the Rust crate that owns the conflicted files.

PR #2233, opened 2026-08-06
test(server,data): work through the #2224 review follow-ups

### LINKED ISSUE #2224: Test-hardening follow-ups from the local CodeRabbit review of the #2154-#2174 batch
A local CodeRabbit CLI review (`coderabbit review --agent --committed --base origin/main`) was run over the whole merged batch #2154–#2174 — 106 files — because the hosted bot had hit its OSS review limit on most of those PRs and reported "Review limit reached", which shows up as a green check without a review having happened.

12 findings. The two that changed decisions are already handled (#2223, and a comment on #2073). These are the leftovers, all minor, all in tests that landed:

1. **`apps/server/src/cors_tests.rs` — `the_shipped_default_is_not_permissive` asserts the wrong thing.** It checks the allow-origin header is not `"*"`. A configuration that *reflected* `https://attacker.example` back would pass a test whose name promises it cannot. Assert the header is empty for an untrusted origin instead. The shipped default is a localhost allow-list, so this is a weak test rather than a live hole — but it is a security test, so it should mean what it says.

2. **`apps/server/src/routes/parse/extract_file_tests.rs` — `a_zero_ceiling_rejects_any_payload` never checks the boundary.** A `max_file_size_mb` of 0 should admit an empty body and reject a one-byte one; only the rejection is asserted, so a mutant that rejects everything survives.

3. **`apps/server/src/admission_tests.rs` — `resident_gauge_round_trips_into_the_metrics_body`** asserts the setter/getter round trip *and* the metrics body. The round-trip half is subsumed by the body assertion; the body is the observable contract.

4. **`packages/data/src/spatial-types.test.ts`** — a retype override stores the IFC class name as given (`IFCBUILDINGSTOREY`), while `isSpatialStructureTypeName()` expects the PascalCase form. Canonicalise on the way in.

None of these block anything that shipped; they are the next pass over the same files.

BIMvoice:
Closes #2224. All four items, plus one that turned out wider than described.

For the two tests that pass while permitting the bug they are named for, I proved the weakness before fixing it rather than assuming the review was right.

## 1. `cors_tests.rs` — the security test now means what its name says

`the_shipped_default_is_not_permissive` only asserted the allow-origin header was not the literal `*`.

Swapping the layer to `AllowOrigin::mirror_request()`, which reflects any origin back, left the OLD assertion **passing** — because the reflected value is `https://attacker.example`, which is indeed not `*`. A config granting an attacker cross-origin access sailed through a test promising it could not.

It now asserts the header is **empty** for an untrusted origin. Under the same mutation: `left: "https://attacker.example"`, `right: ""` — fails.

Still a weak test rather than a live hole, since the shipped default is a localhost allow-list. But as you said, it is a security test and should mean what it says.

## 2. `extract_file_tests.rs` — the zero-ceiling boundary

`a_zero_ceiling_rejects_any_payload` only asserted the reject half. Making `extract_file` return `FileTooLarge` on entering the field — rejecting everything — left the old test passing. It now also asserts an empty body is *admitted* under a zero ceiling, and that mutation fails on it.

## 3. `admission_tests.rs` — dropped the subsumed half

Removed the setter/getter round trip; the metrics body is the observable contract. Confirmed the remaining assertion still bites rather than deleting one and assuming: making `metrics_text` emit `0` instead of the real gauge fails it.

## 4. The retype casing — wider than the issue described

Root cause is as you diagnosed, and the fix belongs in `setTypeOverride`, not `isSpatialStructureTypeName` — the latter's case-sensitivity is deliberate and has its own test pinning it.

But `EntityTable` has **three independent implementations**, and all three stored the override verbatim: the columnar table (`packages/data/src/entity-table.ts:341`), the cache-restored table (`packages/cache/src/sections/entities.ts:182`), and the server-backed table (`apps/viewer/src/utils/serverDataModel.ts:402`).

Fixing only the first would have left the same retype behaving differently depending on whether the model came from a fresh parse, a cache restore, or the server — harder to diagnose than the uniform bug it replaced. All three now canonicalise identically, using the already-exported `IFC_ENTITY_NAMES`, so no new API is added. The changeset covers `@ifc-lite/data` **and** `@ifc-lite/cache` since both changed behaviour; `apps/viewer` is private.

RED-first here too: the new interop test fails `expected false to be true` against unmodified code.

## Verification

- `cargo test -p ifc-lite-server`: **193 passing**. Clippy clean with `--no-deps`, `Checking ifc-lite-server` confirmed present so it is not a cached pass.
- `@ifc-lite/data` 122 tests / 11 files; `@ifc-lite/cache` 58 / 8. Both green.
- `tsc --noEmit` clean for `packages/data`, `packages/cache` and `apps/viewer`.
- Every mutation applied with an asserted replacement count and the mutated line re-read; restores by file copy throughout.

## On the trigger for this issue

Worth recording: the hosted bot reported "Review limit reached" on most of that batch, which renders as a **green check with no review having happened**. That is the fourth distinct way a green signal has misled in this area lately — alongside running the wrong test runner (which prints nothing at all), a partial build making whole test files fail to load, and a stale workspace dep silently under-collecting a suite. A green check is evidence a job ran, not evidence it checked anything.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retyped IFC entities now consistently recognize canonical PascalCase names across parsed, cached, server-backed, and viewer models.
  * Spatial structure checks correctly identify retyped entities regardless of the casing supplied.
  * Removing a type override continues to behave as expected.
  * Zero-byte files are accepted, while one-byte files are correctly rejected according to the configured limit.

* **Tests**
  * Expanded coverage for type override casing, metrics behavior, CORS security, and file-size boundaries.

### DIFF
--- a/.changeset/entity-table-retype-override-casing.md
+++ b/.changeset/entity-table-retype-override-casing.md
@@ -1,0 +1,10 @@
+---
+"@ifc-lite/data": patch
+"@ifc-lite/cache": patch
+---
+
+Fix `EntityTable.setTypeOverride` storing a UI retype's class name in whatever casing the caller passed instead of canonicalising it.
+
+A "change class" retype hands `setTypeOverride` a raw UPPERCASE IFC class token (e.g. `IFCBUILDINGSTOREY`), and `getTypeName` echoed the override straight back unchanged. `isSpatialStructureTypeName` — and any other case-sensitive `*Name` predicate built off `IfcTypeEnumToString`'s PascalCase output — matches against the PascalCase form only, so a retyped entity's new class silently stopped being recognised as part of the spatial tree, even though the case-insensitive `isStoreyLikeSpatialTypeName` correctly saw it. `setTypeOverride` now canonicalises the incoming name to PascalCase before storing it, so `getTypeName` and every name-based predicate agree regardless of the casing a caller passes in.
+
+`EntityTable` has three independent implementations — the columnar table in `@ifc-lite/data`, the cache-restored table in `@ifc-lite/cache`, and the server-backed table in `apps/viewer` — and all three stored the override verbatim. Fixing only one would have left the same retype behaving differently depending on whether the model came from a fresh parse, a cache restore, or the server, which is harder to diagnose than the original bug. All three now canonicalise identically.
--- a/apps/server/src/admission_tests.rs
+++ b/apps/server/src/admission_tests.rs
@@ -314,14 +314,18 @@ async fn breaker_is_inert_when_rss_budget_or_pct_is_zero() {
     no_pct.acquire(1).await.expect("shed_pct 0 ⇒ no breaker");
 }
 
-/// `set_resident_bytes` / `resident_bytes` round-trip, and the gauge is what
-/// `metrics_text` publishes. Without this the sampler could write to a field
-/// nothing reads.
+/// `set_resident_bytes` feeds the gauge that `metrics_text` publishes.
+/// Without this the sampler could write to a field nothing reads.
+///
+/// Pins only the observable contract (the metrics body), not the
+/// `resident_bytes()` getter round-trip: the body assertion below already
+/// requires `set_resident_bytes` to have taken effect, so a separate
+/// `assert_eq!(a.resident_bytes(), 4_096)` added nothing a broken
+/// `metrics_text` implementation could hide behind.
 #[tokio::test]
 async fn resident_gauge_round_trips_into_the_metrics_body() {
     let a = Arc::new(Admission::new(cfg_shed(2, 100, 2, 85)));
     a.set_resident_bytes(4_096);
-    assert_eq!(a.resident_bytes(), 4_096);
     assert!(a.metrics_text().contains("ifc_server_resident_bytes 4096"));
     // The budget gauge reports bytes, not MB.
     assert!(a
--- a/apps/server/src/cors_tests.rs
+++ b/apps/server/src/cors_tests.rs
@@ -133,6 +133,14 @@ async fn near_miss_wildcards_do_not_enable_permissive_cors() {
 /// The default `CORS_ORIGINS` (nothing configured) is a localhost allow-list,
 /// never permissive — a wildcard default would expose every self-hosted
 /// deployment that never sets the variable.
+///
+/// Pins that the header is EMPTY for an untrusted origin, not merely "not the
+/// literal `*`". A layer that reflected the request's `Origin` back
+/// (`AllowOrigin::mirror_request()`) would pass the old `assert_ne!(.., "*")`
+/// version of this test while still granting `https://attacker.example`
+/// cross-origin access — the response header would read
+/// `https://attacker.example`, never the string `"*"`. Asserting emptiness is
+/// the only way this test can catch that.
 #[tokio::test]
 async fn the_shipped_default_is_not_permissive() {
     let defaults = shipped_default_config().cors_origins;
@@ -141,9 +149,10 @@ async fn the_shipped_default_is_not_permissive() {
         &defaults.iter().map(String::as_str).collect::<Vec<_>>(),
     )
     .await;
-    assert_ne!(
+    assert_eq!(
         allow_origin_for(state, "https://attacker.example").await,
-        "*",
-        "the default CORS configuration must not be permissive: {defaults:?}"
+        "",
+        "the default CORS configuration must not grant an untrusted origin any \
+         Access-Control-Allow-Origin header (reflected or wildcard): {defaults:?}"
     );
 }
--- a/apps/server/src/cors_tests.rs
+++ b/apps/server/src/cors_tests.rs
@@ -61,8 +61,15 @@ async fn state_with_origins(label: &str, origins: &[&str]) -> AppState {
 }
 
 /// `GET /api/v1/health` with an `Origin`, returning the
-/// `Access-Control-Allow-Origin` response header (empty string when absent).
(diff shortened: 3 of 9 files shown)


louistrue:
Gated this with #2234 and #2235 (all three merge cleanly together; build 43/43, `turbo test --force` 86/86, `turbo typecheck --force` 36/36, `cargo test --workspace` clean, clippy clean on the pinned nightly). Those two are merged. **Holding this one on a small gap, because it is the same shape as the issue it closes.**

The fix is right — canonicalising in all three `EntityTable` implementations is exactly the move, and it is the same three-implementations-disagreeing pattern as `sumByType` in #2179. But only two of the three are pinned:

| implementation | mutation: revert to `typeOverrides.set(id, typeName)` | result |
|---|---|---|
| `packages/data/src/entity-table.ts` | `normalizeIfcUpperCase(...)` removed | **1 test dies** ✅ |
| `packages/cache/src/sections/entities.ts` | `IFC_ENTITY_NAMES[...] ?? ...` removed | **survives** — `@ifc-lite/cache` 65/65 pass, and the whole workspace stays green at **86/86 task**s ❌ |

So the cache-restored table's canonicalisation is unpinned across the entire monorepo. That is the implementation that ships to users through the persisted cache, so it is arguably the one most worth pinning — and an unpinned fix in a PR that closes a test-hardening issue is the exact shape #2224 was about.

A `readEntities` round-trip asserting `setTypeOverride(id, 'IFCBUILDINGSTOREY')` then `getTypeName(id) === 'IfcBuildingStorey'` (and that `isSpatialStructureTypeName` accepts it) should do it — the existing `entities.ts` round-trip tests look like the right home.

Two process notes while I was measuring, both worth having:

- My first attempt at that mutation **aborted**: I anchored on `IFC_ENTITY_NAMES[...]` in `entity-table.ts`, but that file uses `normalizeIfcUpperCase`. The replacement count assertion caught it — without it I'd have run the suite on unmutated code, seen 122 green, and reported "the data side is unpinned too", which is false. Worth keeping the count assertion on every mutation for exactly this.
- The two implementations do not canonicalise identically: `data` goes through `normalizeIfcUpperCase`, `cache` and `serverDataModel` go through `IFC_ENTITY_NAMES[...] ?? typeName.toUpperCase()`. They agree on the cases I checked, but "all three must agree here" is a claim a shared helper would enforce and three separate expressions do not.

Push the cache test and I'll take it straight through.

BIMvoice:
Pushed. Your measurement reproduced exactly.

Reverting the cache reader to `typeOverrides.set(id, typeName)` leaves the suite green — the third `EntityTable` implementation was unpinned while the other two were covered. A `readEntities` round-trip now asserts `setTypeOverride(id, 'IFCBUILDINGSTOREY')` reads back as `'IfcBuildingStorey'` after a full cache write/read, and that `isSpatialStructureTypeName` accepts it.

**Re-verified on your rebased tip, not just the base I started from.** Worth stating because the counts moved: the branch base I first measured on had 9 files / 65 tests, your tip has 8 / 58. That is not a dropped test — I checked, since a rebase draining an assertion is exactly the failure mode here. `packages/cache/src/utils/buffer-utils.test.ts` exists on `main` but not on this branch, and `git diff main...branch --name-status` shows **zero deletions**: the branch's merge-base simply predates that file. Merging will not remove it. (The branch is 20 commits behind `main`, if you want it rebased before landing.)

Evidence on your tip:

- mutation + existing tests → **58/58 pass** (survivor reproduced)
- mutation + new test → fails, `expected 'IFCBUILDINGSTOREY' to be 'IfcBuildingStorey'`
- restored → **59/59**

`tsc --noEmit` clean. `git status` shows only `cache.test.ts`; `entities.ts` is byte-identical to HEAD.

## On the three implementations diverging

Checked whether they can actually disagree, since "they agree on the cases I checked" is weaker than you'd want. They cannot, on this path:

- `packages/data/src/entity-table.ts:349` — `normalizeIfcUpperCase(typeName.toUpperCase())`, and `normalizeIfcUpperCase(upper) = IFC_ENTITY_NAMES[upper] ?? upper`
- `packages/cache/src/sections/entities.ts:190` — `IFC_ENTITY_NAMES[typeName.toUpperCase()] ?? typeName.toUpperCase()`
- `apps/viewer/src/utils/serverDataModel.ts:411` — same as cache

Inlining `normalizeIfcUpperCase` makes the first identical to the other two, character for character. So they agree on **every** input, not just the sampled ones — no live defect, and your point stands as a maintainability one rather than a correctness one: a shared helper would enforce what three copies currently only happen to satisfy.

Note `entity-table.ts` also calls `normalizeIfcUpperCase` at line 129 for ingest of raw type names, which is a different path from `setTypeOverride` and outside what this test covers.

## Taking the process note

Your aborted-mutation point is well made and I've adopted the count assertion as standard — it has already caught one anchor of mine that would otherwise have run a suite against unmutated code and reported a false clean.

louistrue:
Nudge from our side: this has been sitting in draft for about a day and **every lane is green** — Rust tests, Node tests, Typecheck, Lint, Build + WASM + Rust + Node, the parity and correspondence jobs, and all four server-binary validations.

The reason I'm flagging it rather than just waiting: the writeup is doing the thing we keep asking for and rarely get. Proving `AllowOrigin::mirror_request()` still satisfies the old assertion, *before* rewriting it, is the difference between "I strengthened a test" and "I demonstrated the test could not fail". That's exactly the standard we want on security assertions.

Is this draft-because-unfinished, or draft-because-nobody-undrafted it? If it's the latter, undraft and we'll take it through review — it's the kind of change that should not be aging.

PR #2239, opened 2026-08-06
fix(renderer): drop stale instanced bounds when a model's templates are freed

BIMvoice:
Refs #2073 — the pre-step-2 pinning you asked for:

> Today this is unreachable (nothing calls the function), which is exactly why it needs pinning before step 2 wires it up.

## What was wrong

`removeInstancedTemplatesForModel` freed a model's buffers and pruned its occurrences but left `boundingBoxes` untouched. Right for one case, wrong for two:

| id kind | before | now |
|---|---|---|
| **mixed** (also owns flat geometry) | left alone | left alone — the flat path owns that cache |
| **instanced-only**, last occurrence gone | kept a stale world AABB | entry dropped |
| **shared**, occurrences survive elsewhere | kept bounds covering the freed model | recomputed from survivors |

The instanced-only case is the one you named: `raycastBoundingBoxes` could still pick an element with no geometry left anywhere, and `getBounds()` kept sizing for it after geometry release.

The shared case matters for the same reason and could not fix itself: `boundingBoxes` for instanced ids is a running union built at upload time by `unionInstancedWorldAabb` (`scene.ts:3216-3248`) that **only ever grows**. After a partial prune it is provably oversized, generating false hits in the space the freed model vacated.

## The discriminator

There was no existing way to tell instanced-only from mixed. The only available signal is `meshDataMap.has(eid)` — true means the id also owns flat geometry. That is what gates the delete. Flagging it explicitly because inventing a wrong discriminator here was the main risk.

`getBounds()` (`scene.ts:3762-3822`) derives the union on demand from `boundingBoxes` rather than caching it separately, so pruning the map is sufficient — there is no second cache to invalidate.

## One correction to the premise

The old doc comment — and your note — said this *"mirrors `removeInstancedEntity` in leaving `boundingBoxes` alone"*. It doesn't: `removeInstancedEntity` **does** `boundingBoxes.delete(expressId)` at `scene.ts:1367`.

It's harmless there only because both of its call sites (`:1267`, `:1336`) sit inside `removeMeshesForEntity`, which already deleted the entry itself at `:1263` — so that delete is *redundant, not absent*. Your actual point stands entirely; only the comparison was off. Comment corrected rather than implemented around.

## Verification

RED first, against unmodified `main` — three of the four new tests fail, and the fourth (the mixed-id case) passes because it was already correct:

```
not ok 1 - drops the cached world AABB for an instanced-only id once its last occurrence is freed
          actual: { min: {x:0,y:0,z:-1}, max: {x:1,y:0,z:0} }  expected: null
not ok 3 - recomputes a SHARED id's bounds from its surviving occurrences instead of leaving a stale union
          0 !== 10
not ok 4 - an instanced-only id whose model was removed is not selectable via the released-geometry bbox raycast
          actual: { expressId: 10, distance: 5 }  expected: null
```

I re-ran this by hand after the agent's run rather than relying on the report: reverting both hunks gives 21 pass / 3 fail, restoring gives 24 / 0.

The raycast consequence you named **is** testable headlessly — `raycastBoundingBoxes` (`scene-raycaster.ts:318`) is a pure function over the map, reached via `scene.raycast()` once `releaseGeometryData()` puts it on the bbox-only path. No GPU device needed.

- Renderer suite: **449 passing across 39 files**, `tsc --noEmit` clean.
- `getResidentGpuBytes()` exactness unchanged — still covered by the existing GPU-byte-accounting suite.
- Changeset: patch bump for `@ifc-lite/renderer`.

## Still step 2's job

This is only the first of the two things you listed. The federated path still drops instanced-only shards — `addInstancedShard` takes an owning `modelIndex`, but finalization does not forward secondary models' shards — which remains the substance of step 2 and #1912.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale bounding boxes after removing instanced model templates.
  * Updated bounds for shared entities that still have active instances.
  * Prevented released instanced-only entities from appearing in bounding-box raycasts.
  * Preserved correct bounds for entities containing both flat and instanced geometry.

* **Tests**
  * Added coverage for bounding-box cleanup, recomputation, preservation, and raycast behavior.

### DIFF
--- a/.changeset/instanced-templates-bbox-teardown.md
+++ b/.changeset/instanced-templates-bbox-teardown.md
@@ -1,0 +1,11 @@
+---
+"@ifc-lite/renderer": patch
+---
+
+Fix `Scene.removeInstancedTemplatesForModel` leaving stale cached bounding boxes behind (#2073).
+
+The method already freed a model's GPU-instanced templates and pruned `instancedEntityMap`, but never touched the `boundingBoxes` cache it also feeds. For an id whose occurrences were entirely in the removed model (instanced-only), the cached world AABB lingered after the geometry was gone, so `getEntityBoundingBox()` kept returning a box for nothing and the released-geometry bounding-box raycast could still pick an element with no remaining mesh. For an id shared across models (occurrences in more than one), the cached box is a union built at upload time, so pruning only the removed model's occurrences left the box sized for occurrences that no longer exist.
+
+An id that also owns flat (non-instanced) geometry is unaffected: that mixed case is owned by `removeMeshesForEntity`'s flat-removal path, which already clears the cache on its own schedule.
+
+`removeInstancedTemplatesForModel` has no caller yet (step 1 of the #2073/#1912 fix); this closes the gap before step 2 wires it up.
--- a/packages/renderer/src/scene-instanced-model-scope.test.ts
+++ b/packages/renderer/src/scene-instanced-model-scope.test.ts
@@ -382,3 +382,100 @@ describe('Scene.removeInstancedTemplatesForModel — entity bookkeeping', () => 
     assert.strictEqual(survivor.selectedCount, 1, 'model 7 still owns a selected occurrence');
   });
 });
+
+/**
+ * A single-occurrence shard with an explicit world translation, for tests that
+ * need distinguishable, non-overlapping world AABBs (the `shard()` helper above
+ * always starts each shard's template 0 at local translation (0,0,0), so two
+ * shards built with it collide in world space).
+ */
+function singleOccShard(entityId: number, tx: number): DecodedInstancedShard {
+  return {
+    templates: [{
+      positions: new Float32Array([0, 0, 0, 1, 0, 0, 0, 1, 0]),
+      normals: new Float32Array([0, 0, 1, 0, 0, 1, 0, 0, 1]),
+      indices: new Uint32Array([0, 1, 2]),
+      origin: [0, 0, 0] as [number, number, number],
+    }],
+    instances: [{
+      templateIndex: 0,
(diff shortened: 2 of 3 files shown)


BIMvoice:
The three red checks here are a repo-wide viewer-E2E failure, not this change. Evidence written up on #2241: the identical three fail on #2238, which touches only `rust/python/**` and a Python workflow — eight files, nothing that can stop the viewer booting. Same on #2240 and #2241.

Both failing smoke tests die at app boot (`waitForSelector('input[type="file"]')`, 30s), not at any assertion, and `Build + WASM + Rust + Node` is just the aggregate gate — typecheck, build, node-tests and lint all pass on it.

Nothing to do on this branch. Local verification stands: renderer 449 passing across 39 files, `tsc --noEmit` clean, and reverting either pruning hunk fails exactly three of the four new tests.

louistrue:
Merged current `main` in. Your three red lanes — `Viewer E2E smoke`, `Viewer benchmark` and the `Build + WASM + Rust + Node` rollup — were **not** anything in this PR: they were the site-wide white-screen from #2243, which every branch cut off main inherited. #2244 fixed it (`c40c17778`), so those lanes should come back green on the next run without you touching anything.

For the record, since it explains a confusing few hours: the bundle was throwing `TypeError: C is not a function` during module evaluation, before React mounted, so `#root` stayed empty and every browser-driving lane timed out waiting for `input[type="file"]`. Root cause is `vite-plugin-top-level-await` deferring a chunk's exports without marking its synchronous importers — written up in #2246, along with 22 residual instances of the same hazard that the chunk pinning does not cover.

Reviewing this PR itself next.

PR #2240, opened 2026-08-06
fix(viewer): tell the user to reload when the script sandbox has aborted

BIMvoice:
Refs #1922 — the second loose end you named:

> `SandboxAbortError` and `isSandboxRuntimeAborted` are exported and in `api-surface.json`, but **nothing in `apps/viewer` or the extension host reads the flag** to prompt a reload. So the user-visible symptom in this issue — the sandbox is unusable until reload — is unchanged; only the error message improved.

This does not touch the upstream defect, which is still live and still not ours. It closes the gap between the containment that landed in #2062/#2170 and anything actually consuming it.

## What was happening

Once the teardown abort fires, the shared QuickJS WASM module is dead for the rest of the document — so a reload is the only remedy. Neither entry point said so:

- **`useSandbox.ts`** — `disposeSandbox()` caught dispose failures and only `console.error`'d them, so a `SandboxAbortError` raised from teardown never reached the UI at all. Subsequent runs then failed with whatever generic error a doomed `eval()`/`createSandbox()` produced, routed through `augmentScriptError()`, which had no case for it.
- **`sandbox-factory.ts`** — `BimSandboxHandle.run()` special-cased only `"Lifetime not alive"` (an unrelated failure). A `SandboxAbortError` fell straight through as raw emscripten teardown text.

## What changed

Both sites now read the latch in two places:

1. **Before attempting** — `isSandboxRuntimeAborted()` is checked up front, so no eval is started against a module that cannot succeed.
2. **In the catch** — a `SandboxAbortError` is recognised on the run that first discovers the abort.

The message goes through the **existing** `scriptLastError` store (`scriptSlice.ts`), already rendered inline by `ScriptPanel.tsx` and `ChatPanel.tsx` / `ExecutableCodeBlock.tsx` for every other script failure. No new UI primitive, no `alert`/`confirm`. The extension host throws with the same actionable message, matching how its existing error paths behave.

## The test design deliberately mirrors yours

You flagged that the existing tests assert **containment, never absence** — `oom-job-dispose.test.ts:64-95` wraps its assertion in `if (caught !== undefined)` precisely so an upstream fix would not fail the file.

I kept that property. The decision is a pure helper that takes `aborted` as a plain boolean rather than reading the latch itself:

```ts
export function describeSandboxAbort(aborted: boolean, err?: unknown): string | null {
  if (aborted || err instanceof SandboxAbortError) return SANDBOX_ABORT_RELOAD_MESSAGE;
  return null;
}
```

So the tests set the condition up directly and never reproduce the upstream OOM. If quickjs-emscripten ever fixes the abort, nothing here breaks — the wiring simply stops being reached.

## Verification

RED-first, and re-confirmed by hand rather than taken from the agent's report: removing the catch-block wiring fails the new test with the raw `"QuickJS aborted while freeing the sandbox runtime…"` text where the reload message was expected. 1 replacement, count asserted; restoring returns it to green (1 pass / 1 fail → 2 / 0).

- `apps/viewer`: **3038 passing / 2 skipped**, 0 failing. `packages/sandbox`: 121 across 20 files.
- `tsc --noEmit` clean, full output read.
- `check-api-surface.mjs`: 4043 exports, matches snapshot. No published package touched — `apps/viewer` is private, so no changeset.
- `BimSandboxHandle` gained an `export` so the test can construct it; noted in a comment, and it does not reach a package entry.

## Deliberately not included

Surfacing the message on the very run whose **own** teardown latches the flag. That run's `eval()` has already resolved successfully by the time `dispose()` throws in `finally`, so covering it means reworking the success/error sequencing rather than just reading the flag. Out of scope for "read the flag, tell the user" — raising it rather than quietly widening.

## Still open on the issue

The upstream filing against quickjs-emscripten. There is still no linked issue, and that remains the actual fix path — this PR only stops the failure being silent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of sandbox runtime failures.
  * Displays clear reload guidance when the sandbox has been aborted or encounters an abort-related error.
  * Preserves specific error messages for ordinary script failures and invalid sandbox usage.
* **Tests**
  * Added coverage for aborted runtimes, abort-related errors, and standard script errors.

### DIFF
--- a/apps/viewer/src/hooks/useSandbox.ts
+++ b/apps/viewer/src/hooks/useSandbox.ts
@@ -20,6 +20,7 @@ import {
   formatDiagnosticsForDisplay,
   type RuntimeScriptDiagnostic,
 } from '../lib/llm/script-diagnostics.js';
+import { describeSandboxAbort } from '../lib/sandboxAbort.js';
 
 /**
  * Tear a sandbox down without letting a teardown failure escape.
@@ -181,7 +182,17 @@ export function useSandbox(config?: SandboxConfig) {
     let sandbox: Sandbox | null = null;
     try {
       // Create a fresh sandbox for every execution — full isolation
-      const { createSandbox } = await import('@ifc-lite/sandbox');
+      const { createSandbox, isSandboxRuntimeAborted } = await import('@ifc-lite/sandbox');
+
+      // The shared QuickJS WASM module is dead for the rest of the document
+      // once a teardown abort has latched (#1922) — fail fast instead of
+      // creating a sandbox that cannot tear down cleanly either.
+      const abortedBeforeAttempt = describeSandboxAbort(isSandboxRuntimeAborted());
+      if (abortedBeforeAttempt) {
+        setError(abortedBeforeAttempt);
+        return null;
+      }
+
       sandbox = await createSandbox(bim, {
         permissions: { model: true, query: true, viewer: true, mutate: true, store: true, lens: true, export: true, files: true, ...config?.permissions },
         limits: { timeoutMs: 30_000, ...config?.limits },
@@ -200,6 +211,16 @@ export function useSandbox(config?: SandboxConfig) {
       useViewerStore.getState().bumpScriptRunSeq();
       return result;
     } catch (err: unknown) {
+      // A dispose() thrown from a prior run's teardown latches the runtime as
+      // aborted (#1922); a SandboxAbortError surfacing here means this very
+      // attempt hit it. Either way the fix is "reload", not the generic
+      // script-error diagnostics below.
+      const abortMessage = describeSandboxAbort(false, err);
+      if (abortMessage) {
+        setError(abortMessage);
+        return null;
+      }
+
       const runtime = augmentScriptError(err instanceof Error ? err.message : String(err), code);
 
       // If the error is a ScriptError with captured logs, preserve them.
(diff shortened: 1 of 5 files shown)


BIMvoice:
Same as #2239 — the three red checks are the repo-wide viewer-E2E failure, not this change. Evidence on #2241: #2238 fails the identical three and touches only `rust/python/**` plus a Python workflow.

Both failing smoke tests die at app boot (`waitForSelector('input[type="file"]')`), not at an assertion, and the `Build + WASM + Rust + Node` red is just the aggregate gate over them — typecheck, build, node-tests and lint all pass.

Local verification stands: `apps/viewer` 3038 passing / 2 skipped, `packages/sandbox` 121 across 20 files, `tsc` clean, api-surface unchanged at 4043 exports; removing the catch-block wiring fails the new test with the raw teardown text where the reload message belongs.

louistrue:
Merged current `main` in. Your three red lanes — `Viewer E2E smoke`, `Viewer benchmark` and the `Build + WASM + Rust + Node` rollup — were **not** anything in this PR: they were the site-wide white-screen from #2243, which every branch cut off main inherited. #2244 fixed it (`c40c17778`), so those lanes should come back green on the next run without you touching anything.

For the record, since it explains a confusing few hours: the bundle was throwing `TypeError: C is not a function` during module evaluation, before React mounted, so `#root` stayed empty and every browser-driving lane timed out waiting for `input[type="file"]`. Root cause is `vite-plugin-top-level-await` deferring a chunk's exports without marking its synchronous importers — written up in #2246, along with 22 residual instances of the same hazard that the chunk pinning does not cover.

Reviewing this PR itself next.

Issue #2183, opened 2026-08-05
RAM's  high with model 350MB


RyugaRyuzaki:
HI, i try to load model 350MB, RAM is so high

<img width="1876" height="946" alt="Image" src="https://github.com/user-attachments/assets/b489ed73-4916-4fac-9dc4-6fd85809c13c" />

BIMvoice:
Measured this rather than leaving it untriaged. Numbers below, plus the one caveat that matters: **my largest available fixture is 169.2 MB, not 350 MB**, so everything here is a ratio measured at 169 MB. Where I scale to 350 MB I say so explicitly — those are estimates, not observations.

Fixture: `ISSUE_053_20181220Holter_Tower_10.ifc`, 177,465,622 bytes (169.2 MB). Chromium via Playwright against a production build. `performance.measureUserAgentSpecificMemory()` refused with `SecurityError` despite `crossOriginIsolated === true`, so the numbers come from CDP `Performance.getMetrics` plus the app's own instrumentation (`Scene.getResidentGpuBytes()` / `getResidentCpuBytes()`, and the existing `[mem-summary]` line which reads `wasmMemory.buffer.byteLength` directly).

## Steady state, at 169.2 MB

| what | size | ×  file |
|---|---|---|
| JS heap (post-GC, settled) | 200.0 MB | 1.18× |
| retained raw IFC bytes | 169.2 MB | 1.00× |
| GPU resident (batches 60.2 + instanced 6.3) | 66.5 MB | 0.39× |
| CPU-side mesh buffers | 52.0 MB | 0.31× |

## The transient spike is the striking number

**WASM linear memory peaked at 1844.1 MB — 10.9× the file size** — summed across the five parallel geometry workers during the geometry-build phase.

It does not persist. Both the parser workers and the geometry workers are explicitly `.terminate()`d when their job finishes, which releases each worker's entire WASM instance to the OS. The process-family RSS trend corroborates it:

```
baseline            707.4 MB
peak after load    2811.2 MB   (+2104 MB, 12.4× file)
after forced GC    2510.1 MB
steady state (+22s) 2268.1 MB  (+1561 MB,  9.2× file)   ← still declining
```

The renderer process alone went 2021.5 → 1747.3 → 1505.3 MB and had **not** finished falling when I stopped measuring at +22s, so the true floor is likely lower than the last figure.

[...]

PR #2241, opened 2026-08-06
fix(data): cap the safeUtf8Decode scratch so one giant decode cannot pin memory (#2183)

louistrue:
Part 1 of 2 for #2183. Independent of part 2, and worth landing on its own.

## The defect

`safeUtf8Decode`'s scratch buffer grows by doubling to the largest subarray ever decoded, and is never released. That is the right trade for the 50-500 byte per-entity reads the helper was written for. It is the wrong trade for a single whole-file decode: the scratch lands on the next power of two above the file size and stays there for the lifetime of the realm.

Measured in the viewer on a 342 MB model: **512 MB pinned** (exactly 2^29), which was 30% of the entire main-thread JS heap. Confirmed by heap snapshot, with the sole retainer being this module-level `scratchBuffer`, and by a captured allocation stack.

## The change

Above 4 MiB, hand out a throwaway buffer instead of retaining it. Reuse only pays off for a buffer that is hit repeatedly, and a one-off giant decode never is.

Everything on the hot path is orders of magnitude below the cap, so those callers keep the exact same reused buffer they had before. `MAX_RETAINED_SCRATCH_BYTES` is a power-of-two multiple of the 4096 base, so the doubling loop lands exactly on it and can never overshoot for a request at or under the cap.

## Verification

- 3 new tests, all against the SAB-copy path.
- Mutation-checked: restoring the old unbounded growth fails 2 of the 3 (retained 8,388,608 bytes instead of 0).
- `pnpm typecheck` and `pnpm test` green.

Changeset: patch on `@ifc-lite/data`. No export change, so no api-surface update. The two test-only helpers are deliberately not re-exported from the package index.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTF-8 decoding memory management by reusing buffers for smaller requests and avoiding retention of oversized allocations.
  * Reduced the risk of excessive memory growth during large file and IFCX processing.

* **Tests**
  * Added regression coverage for buffer reuse, allocation limits, oversized decodes, and memory preservation.

### DIFF
--- a/.changeset/olive-donkeys-tap.md
+++ b/.changeset/olive-donkeys-tap.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/data": patch
+---
+
+Cap the `safeUtf8Decode` scratch buffer so a one-off whole-file decode can no longer pin memory for the lifetime of the realm.
+
+The scratch grew by doubling to the largest subarray ever decoded and was never released. That is the right trade for the 50-500 byte per-entity reads the helper was written for, but a single whole-source decode pushed it to the next power of two above the file size and kept it there: a 342 MB model pinned 512 MB, measured as 30% of the viewer's main-thread heap (#2183).
+
+Decodes at or under 4 MiB keep the existing reused buffer unchanged. Larger ones now get a throwaway buffer, since reuse only pays off for a buffer that is hit repeatedly and a one-off giant decode never is.
--- a/.changeset/olive-donkeys-tap.md
+++ b/.changeset/olive-donkeys-tap.md
@@ -2,7 +2,7 @@
 "@ifc-lite/data": patch
 ---
 
-Cap the `safeUtf8Decode` scratch buffer so a one-off whole-file decode can no longer pin memory for the lifetime of the realm.
+Cap the `safeUtf8Decode` scratch buffer so an oversized one-off decode no longer retains its full allocation for the lifetime of the realm.
 
 The scratch grew by doubling to the largest subarray ever decoded and was never released. That is the right trade for the 50-500 byte per-entity reads the helper was written for, but a single whole-source decode pushed it to the next power of two above the file size and kept it there: a 342 MB model pinned 512 MB, measured as 30% of the viewer's main-thread heap (#2183).
 
--- a/apps/viewer/src/hooks/useIfcFederation.ts
+++ b/apps/viewer/src/hooks/useIfcFederation.ts
@@ -539,8 +539,10 @@ export function useIfcFederation(
 
     // Check that all files are IFCX format and read buffers.
     // IFCX is JSON; SAB streaming would force a SAB→scratch copy in
-    // safeUtf8Decode + retain the scratch (net worse peak than ArrayBuffer).
-    // Keep on file.arrayBuffer().
+    // safeUtf8Decode on top of the JSON string (net worse peak than
+    // ArrayBuffer). Keep on file.arrayBuffer(). The copy is no longer
+    // *retained* since #2183 capped the scratch, but it is still a
+    // full-file transient the ArrayBuffer path does not pay.
     const buffers: Array<{ buffer: ArrayBuffer; name: string }> = [];
     for (const file of files) {
       const buffer = await file.arrayBuffer();
@@ -596,8 +598,10 @@ export function useIfcFederation(
 
     // Read new overlay buffers.
     // IFCX is JSON; SAB streaming would force a SAB→scratch copy in
-    // safeUtf8Decode + retain the scratch (net worse peak than ArrayBuffer).
-    // Keep on file.arrayBuffer().
+    // safeUtf8Decode on top of the JSON string (net worse peak than
+    // ArrayBuffer). Keep on file.arrayBuffer(). The copy is no longer
+    // *retained* since #2183 capped the scratch, but it is still a
+    // full-file transient the ArrayBuffer path does not pay.
     const newBuffers: Array<{ buffer: ArrayBuffer; name: string }> = [];
     for (const file of files) {
       const buffer = await file.arrayBuffer();
--- a/apps/viewer/src/utils/acquireFileBuffer.test.ts
+++ b/apps/viewer/src/utils/acquireFileBuffer.test.ts
@@ -157,9 +157,13 @@ describe('acquireFileBuffer', () => {
     //   parseFederatedIfcx → safeUtf8Decode(new Uint8Array(buffer)) → JSON.parse
     // safeUtf8Decode must copy SAB-backed views into a scratch buffer in
     // Chromium/Firefox (cross-thread JS string decoding cannot read SAB
-    // directly) and retains that scratch. Net peak with SAB streaming:
+    // directly). Net peak with SAB streaming:
     //   SAB (file.size) + scratch copy (file.size) + JSON string (~file.size)
-    //   + retained scratch — strictly worse than the plain ArrayBuffer path.
+    //   — strictly worse than the plain ArrayBuffer path.
+    // Since #2183 that copy is a throwaway rather than a retained buffer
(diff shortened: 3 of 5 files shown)


BIMvoice:
The three red checks here are **not caused by this PR** — the viewer E2E is failing for every PR that runs it right now. Flagging with the evidence because it is currently gating your work and mine.

**The proof is #2238.** It fails the identical three checks — `Viewer E2E smoke`, `Viewer benchmark (advisory)`, and `Build + WASM + Rust + Node` — and it touches only:

```
.github/workflows/python-wheels.yml
rust/python/{Cargo.lock,Cargo.toml,README.md,ifclite_geom.pyi,src/lib.rs}
rust/python/examples/schedule_csv.py
rust/python/tests/test_bindings.py
```

Eight files, all Python bindings. Nothing there can stop the viewer booting.

**Same failure on:** #2238, #2241 (this one), #2239, #2240. #2242 is not a counter-example — it has no E2E check at all, so its path filters skipped it.

**What actually fails:** two of five smoke tests, both at app boot rather than at any assertion —

```
TimeoutError: page.waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('input[type="file"]')
  at tests/benchmark/viewer-benchmark-page.ts:256
```

`load → geometry → pick → section plane survive end to end` and `page reports no uncaught errors on a clean load`. The other three pass. So the built app is not reaching an interactive state in CI within 30s.

**And the gate is downstream of only that.** `Build + WASM + Rust + Node` fails purely because it aggregates:

```
typecheck: success
viewer-e2e: failure
build: success
node-tests: success
lint: success
geometry-census / rust-tests / docs-checks / plato-check: skipped
```

So there is one real failure, reported three times.

**One false lead worth saving you the trip:** the E2E log shows `Download build artifacts … digest-mismatch: error`, which reads like a failure but is the action's *input parameter* (`with: digest-mismatch: error`) — the download proceeds fine immediately after. Not the cause.

I could not get the underlying page error from the job log; it is in the `error-context.md` Playwright artifact rather than stdout, which is itself worth knowing — as it stands the log tells you the app did not boot but not why, so every occurrence costs a fresh dig. There are also `Failed to create bin … packages/mcp/dist/cli.js` ENOENT warnings during `pnpm install --frozen-lockfile` in the same job; possibly unrelated install noise, possibly a sibling of the lockfile repair in #2237 — I have not verified either way and am not claiming it.

Happy to pick this up if you want it chased rather than just flagged — say the word. Not re-running anything or touching your branch in the meantime.

BIMvoice:
Reviewed this adversarially — the cap logic itself is correct, and I attacked it at the places this repo has been bitten before:

- **At the cap**: `MAX_RETAINED_SCRATCH_BYTES / SCRATCH_BASE_BYTES` = `4194304 / 4096` = `1024` = 2¹⁰, so `while (cap < byteLength) cap *= 2` lands *exactly* on the cap for a request of exactly `MAX`. Your "can never overshoot" claim holds, and it holds for a reason — it needs the cap to be a power-of-two multiple of the base, which it is.
- **One above**: `MAX + 1` takes the `byteLength > MAX_RETAINED_SCRATCH_BYTES` branch, allocates a throwaway of exactly `byteLength`, never touches `scratchBuffer`. Correct.
- **Zero-length**: never reaches `scratchFor` — `safeUtf8Decode` returns `''` at `:148` on `len === 0`, so the empty-input trap does not apply.
- **Units**: bytes end to end. `sub` is already a byte slice before `scratchFor` sees it, so the bytes-vs-chars hazard that usually lurks in a UTF-8 scratch cap is not present.
- **Siblings**: checked for a second scratch implementation that could now disagree. `columnar-parser-attributes.ts` uses a different buffer strategy, not a second `scratchFor`.
- **API surface**: confirmed `__retainedScratchBytes` / `__resetScratchBuffer` really are absent from `packages/data/src/index.ts` (which exports only `safeUtf8Decode` and `textDecoderAcceptsSab` at `:127`), so "no api-surface update needed" is true rather than merely asserted.

## One real gap, and it is the shape that has burned this repo before

`utf8-decode.test.ts:14-15`:

```ts
/** Mirrors `MAX_RETAINED_SCRATCH_BYTES` in the module under test. */
const MAX_RETAINED_SCRATCH_BYTES = 4 * 1024 * 1024;
```

Every boundary assertion in the file is anchored to a **hand-copied** constant, linked to the real one by nothing but that comment. Change the production value — say 4 MiB → 8 MiB — and the tests keep passing while silently testing the wrong thing:

- `oversized = MAX_RETAINED_SCRATCH_BYTES + 1` (`:140`) is now *4 MiB + 1*, comfortably **under** the real 8 MiB cap. The test named for the throwaway path now exercises the retained path.
- `expect(afterSmall).toBeLessThanOrEqual(MAX_RETAINED_SCRATCH_BYTES)` (`:157`) becomes a strictly weaker assertion.
- The at-cap test (`:161`) stops sitting at the cap.

Nothing goes red. This is the same shape as the `basis_from_depth` seed threshold duplicated between `prism_cut.rs` and `voids/mod.rs` with only a doc comment tying them — shifting one there survived 579 tests.

Cheap fix, and the module is already set up for it: export `MAX_RETAINED_SCRATCH_BYTES` alongside the three `__`-prefixed helpers it already exports for tests, and import it instead of re-declaring.

## Second, smaller: the power-of-two invariant is prose-only

`:89-91` states it as a requirement —

> Must stay a power-of-two multiple of `SCRATCH_BASE_BYTES` so the doubling loop below lands exactly on it and can never overshoot

— and it is load-bearing, but nothing asserts it. Set the cap to a non-power-of-two multiple (5 MiB, say) and the loop overshoots to 8 MiB for an at-cap request, retaining double the intended ceiling. A one-line assertion that `MAX % BASE === 0 && Number.isInteger(Math.log2(MAX / BASE))` would pin the thing the comment promises.

Both are test gaps, not defects — the shipped values are correct today. Happy to send either or both as a follow-up if you want them; not touching your branch.

(Separately: the three red checks here are the repo-wide viewer-E2E failure, not this PR — evidence in my other comment above.)

BIMvoice:
Correcting myself on this one, because #2243 lands and it changes the reading.

I said above the E2E red was "not caused by this PR" and leaned toward infrastructure. The first half still holds and the evidence for it is unchanged — #2238 fails the identical three checks and touches only `rust/python/**`. **The second half was too generous.** This looks like a real boot regression on `main`, not flake.

#2243 (external report, `bug`) has the viewer failing to boot in production on deployed commit `a973ab927`:

```
TypeError: C is not a function. (In 'C()', 'C' is undefined)
  Module Code (pending-CytPVxJ4.js:1:555)
```

Same symptom as CI. Both failing smoke tests die waiting for `input[type="file"]` — the app never becomes interactive — and `C is not a function` at **Module Code** (top level, not inside a function) is the signature of a module evaluating against a binding that is still `undefined`, i.e. a circular import that only bites in the bundled/minified build.

So the two are very likely one bug: the built viewer does not boot. CI has been telling us since roughly 05:57 today; the user report is the same failure reaching production.

The only source matching that chunk name is `apps/viewer/src/lib/layers/pending.ts` (imported by `ActivityBar.tsx:41` and `LayerDraftSection.tsx:21`, itself importing `useViewerStore` from `@/store`) — though Vite names a chunk after one of its constituent modules, so the cycle may involve something else bundled alongside it.

I have this being reproduced and root-caused now, starting from the built output rather than by reading, and bisecting `main` if the cycle is not obvious. Will report with the repro and the cycle drawn out rather than a hypothesis.

Also worth noting: your triage of #2229/#2230 put those on Safari 26.5 too, and #2243 is Safari 26.5. Whether that is the same underlying failure or a coincidence of who reports from Safari, I do not yet know and am not claiming.

Nothing to do on this PR — flagging only so the red here is not read as either "mine" or "ignorable".

PR #2628, opened 2026-08-15
fix(export): KMZ heading is a rotation, not a compass bearing

### LINKED ISSUE #2554: KMZ heading is a compass bearing where KML expects a rotation — 90° off for any model with an explicit axis
Found while fixing the Location-panel KMZ path in #2552, and **deliberately not fixed there** — it is pre-existing, it is Rust with its own gate, it changes every KMZ export with an explicit axis rather than only map-absolute ones, and confirming it properly wants the file opened in Google Earth.

## The defect

`rust/export/src/kmz.rs` computes the compass **bearing** of the model's X axis (`90 - math_angle`) and feeds it to KML `<heading>`, which is a **rotation**. A `Z_UP` KML model already has +X along east at heading 0, so the two are 90° apart.

Its own test states the inconsistency plainly:

| axis | asserted heading |
|---|---|
| `None` | 0 |
| `(0, 0)` | 0 |
| **`(1, 0)`** | **90** |

`(1, 0)` is the same geometric fact as the absent case — no rotation — and it should also be 0.

## Why it matters now, and why it is partly ours

**#2534 made this reachable for exactly the models that report georeferencing problems.** Before it, a map-absolute file carried its authored `(0, 1)`, which yields heading 0 — *accidentally correct*. #2534 correctly guards that conversion to the identity `(1, 0)`, which yields heading **90**. So the reporter's KMZ orientation regressed when #2534 landed, and #2552 then propagated the same value to the Location panel — deliberately, because "both surfaces agree" was the goal and this offset is orthogonal to it.

Net effect today: both KMZ surfaces are consistent with each other and both are 90° off for any model with an explicit axis.

## Acceptance

- `(1, 0)` must produce the same heading as `None` and `(0, 0)` — the existing test asserting 90 is the one to invert.
- A non-trivial axis (say 30° from east) must produce the KML rotation, not the bearing; assert the exact value.
- **Confirm in Google Earth**, not only by unit test — this is a visual convention and the sign is easy to get backwards. A model with a recognisable footprint on a satellite basemap is the check.
- Regression control: a model with no explicit axis must be unchanged.

Related: #2526, #2534, #2552.

BIMvoice:
Closes #2554

Closes the KMZ heading bug in #2554: the exporter emitted a compass **bearing** where KML's `<Model><Orientation><heading>` expects a **rotation**.

## The relationship

KML's heading rotates the model away from a baseline where X already points **East**. Rotating clockwise by `heading` puts X at azimuth `90 + heading`, so for X to reach its true bearing `B`:

```
heading = B − 90
```

The code emitted `B` directly. The error is a **uniform 90°** — not a mirror, not quadrant-dependent.

Derived independently before touching anything, and checked against all three existing fixtures: `(1,0)` → bearing 90 → heading **0**; `(0,1)` → bearing 0 → heading **270**; `(−1,0)` → bearing 270 → heading **180**. Matches the issue's own worked example — `(1,0)` means X already points East, i.e. already the baseline, i.e. no rotation.

## The change

`rust/export/src/kmz.rs`, `ifc_angle_to_kml_heading`:

```rust
- (90.0 - angle_from_east_ccw).rem_euclid(360.0)
+ (-angle_from_east_ccw).rem_euclid(360.0) + 0.0
```

**The `+ 0.0` is not cosmetic.** With the angle exactly 0 (X due east), the sign flip yields `-0.0`, which Rust's `{}` formatter renders as `-0` — so the emitted KML contained `<heading>-0</heading>`. Adding `0.0` folds `-0.0` back to `+0.0` per IEEE-754. Found while verifying, not in the original diagnosis, and confirmed with a standalone test.

The doc comment is rewritten too — it previously described the output as a bearing, which is now wrong.

## Tests

The two tests that pinned the old behaviour are updated to the corrected values:
- `heading_matches_ifc_convention` — `(1,0)→0`, `(0,1)→270`, `(−1,0)→180`
- `kml_carries_placement` — `<heading>0</heading>`

Changing tests that pin current behaviour is normally the thing to stop and ask about; here the issue explicitly identifies that behaviour as the bug and asks for the inversion.

**No other test changed behaviour** — 225 passed in `ifc-lite-export`, only those two moved, which is the evidence the change reaches exactly as far as diagnosed.

## Blast radius checks

- **No TS duplication.** `apps/viewer/src/lib/geo/kmz-export.ts` and `kmz-exporter.ts` pass the raw axis values through to the wasm-compiled Rust; `ifc_angle_to_kml_heading` and `angle_from_east_ccw` exist only in `kmz.rs` and its re-export.
- **No importer.** Re-grepped `rust/`, `packages/geometry`, `apps/viewer` — KML is write-only here, so nothing was compensating with an inverse error that this would now break.

`clippy --all-targets -D warnings` clean. Changeset added — this changes published export behaviour, and previously-exported KMZ files with a rotated grid-north axis were 90° off in Google Earth.

## One caveat on the premise

The derivation rests on KML's baseline being X→East, Y→North at heading 0. I could not load the primary OGC KML 2.2 spec — `docs.ogc.org` 404'd on both the HTML and PDF paths — so that premise comes from Google's own KML Models tutorial plus a NASA WorldWind doc citing the spec. They agree, and both are commonly treated as authoritative, but neither is the standard text. Everything else follows from it, so it's the one thing worth a second pair of eyes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected KMZ exports for models with rotated grid north.
  * Fixed heading orientation in Google Earth, including east, north, and west directions.
  * Preserved safe defaults for missing or zero-length grid axes.
  * Existing models affected by the issue should be re-exported.


louistrue (approved):
Verified on a real merge preview against current main (`a35183910`), not on the PR's own row colours.

The heading math is right: heading = bearing - 90, with the Rust tests and `scripts/test-wasm-contract.mjs` updated consistently, and `+ 0.0` correctly folding -0.0.

The blocker I raised is closed. `kmz.rs` had grown 395 to 403 purely in doc comments and the PR opened a **new** module-size allowlist row to pay for it. The comment is now condensed with the bearing-vs-rotation reasoning intact, the file sits at exactly **400**, and the allowlist row is gone. Main's #2574 rows survive the merge untouched.

Checked rather than assumed: the ratchet genuinely **ran 4 tests, 4 passed, 0 filtered out** (an earlier run of mine matched zero tests and exited 0, which proves nothing, so this was re-run by test name). `cargo clippy --workspace --all-targets -- -D warnings` exits 0. Full chain on the merged tree: install, build, typecheck, lint, source-text guard and test all exit 0, 86/86 turbo tasks.

PR #2633, opened 2026-08-15
feat(source-msgraph): Microsoft Graph (OneDrive/SharePoint) file-source provider

BIMvoice:
## Stacks on #2632

This PR branches from `feat/oauth-pkce-provider-auth` (PR #2632, green, MERGEABLE) and depends on `@ifc-lite/oauth-pkce`, which is not on `main` yet. **Please review/merge #2632 first** — this diff includes only the new `packages/source-msgraph` package plus the viewer registration; the base-branch commits are #2632's own.

Opened as a draft for that reason, and because the app registration (client ID) is the maintainer's to create — see below.

## What this adds

`@ifc-lite/source-msgraph`, a `FileSourceProvider` for OneDrive via Microsoft Graph, following `@ifc-lite/source-dalux`'s structure and conventions:

- **Auth**: delegated OAuth 2.0 Authorization Code + PKCE via `@ifc-lite/oauth-pkce`, scope `offline_access https://graph.microsoft.com/Files.Read`. No admin consent needed, no client secret. Popup-based sign-in (`window.open` + polling for the same-origin redirect), gated behind a user gesture per the plugin contract.
- **Listing**: real per-folder `/children` endpoint (`containerListing: 'direct-children'`), `$top`-based pagination via `@odata.nextLink`.
- **Search**: `GET /me/drive/root/search(q='...')`.
- **Change detection**: Graph's `/delta` feed (`changeDetection: true`, resumable cursor).
- **Revision history**: `GET .../versions` lists it (`revisionHistory: true`), but `download()` can only fetch the *current* revision (`downloadHistoricalRevisions: false`) — a `driveItemVersion` exposes no `@microsoft.graph.downloadUrl`, and `.../versions/{id}/content` 302-redirects the same way `/content` does.
- **Download**: fetches the item's pre-signed `@microsoft.graph.downloadUrl` via `ctx.fetchPublic` — **never** `GET .../content` directly. Verified against current Microsoft docs (cited in `provider.ts` and the README): `/content` returns a `302 Found` a browser can't follow when the triggering request required a CORS preflight (attaching `Authorization` does); Microsoft's own recommended fix for JS apps is exactly `@microsoft.graph.downloadUrl`, fetched with no `Authorization` header.

## Deliberate divergences from Dalux / from my own initial plan

- **No relay.** Dalux needs a same-origin relay because its API sends no CORS headers. Microsoft Graph (and the token endpoint, for an app registered with a `spa` redirect URI) sends real CORS headers, so this provider talks to `graph.microsoft.com`/`login.microsoftonline.com` directly via `ctx.fetch`. Confirmed this is a genuine capability difference, not a workaround, before writing it.
- **One project only.** Delegated `Files.Read` has no "list every SharePoint site" endpoint (that needs `Sites.Read.All`, a higher-privileged, admin-consentable scope this PR deliberately doesn't request). So `listProjects` always returns exactly one project: the signed-in user's own drive. No `projectsAreDiscoverableOnly` capability or site-search UI in this PR — documented as a natural follow-up in the README, not implemented here.
- **Folder-organized files only (known v1 gap, documented in the README).** `listContainers` returns the drive root's real child folders directly, `parentId: undefined` at the top level, matching the plugin contract's "top level" shape exactly rather than wrapping it in a synthetic container the way Dalux's file-area concept does. Consequence: a file sitting directly at the drive root (not inside any folder) isn't currently listable — there's no `SourceContainer` a host would address it through. Most real OneDrive usage is folder-organized; a synthetic root-level container is the natural extension if this turns out to matter in practice.

## What the maintainer needs to register (not done here — no client ID committed)

An Azure AD app registration:
- **Client type**: Single-page application (SPA) — public client, PKCE, no secret.
- **Redirect URI**: `<app origin>/oauth/msgraph/callback` (type `spa` — required for CORS on the token endpoint; a `web`-type redirect URI will fail).
- **API permissions (delegated)**: `Files.Read`, `offline_access`. No admin consent needed for either.
- **Supported account types**: matching whatever `tenant` preference the deployment sets (default `common`).

Full detail in `packages/source-msgraph/README.md`.

## Test plan

- [x] Unit tests (`test/provider.test.ts`, `test/auth.test.ts`) — manifest shape, listing, download (current + historical-revision rejection), revisions, delta-based `watchRevisions`, `testConnection`, auth-failure propagation, `restore`/`getIdentity`/`signOut`/`signIn`-without-`window` — all mocked, no network.
- [x] Shared conformance suite (`@ifc-lite/source-fixture`) run against the real provider via a Graph API mock (`test/msgraph-api-mock.ts`) — 23 checks passing, 5 skipped (revision-download and second-project checks, correctly gated off by this provider's declared capabilities).
- [x] Induced-failure verification on 5 core behaviors (download-via-fetchPublic, historical-revision rejection, cTag-based revision identity, 401/403 auth-failure classification, folder-vs-file filtering in `listContainers`) — each broken, observed red, restored, confirmed green again.
- [x] `apps/viewer/src/services/sources/source-host.test.ts` (asserts every registered provider's manifest satisfies `PLUGIN_API_VERSION`) — passes with `MsGraphProvider` registered.
- [x] Full monorepo build (`pnpm build`, 45/45 tasks).
- [x] `pnpm docs:generate` / `docs:check-generated`, `pnpm check:api-surface` (+ `--update`), `pnpm lint` (`check-lint-ran.mjs`, 0 errors), unused-locals ratchet (`check-unused-locals.mjs`, 0 for this package, no regressions elsewhere) — all green.
- [x] Changeset added (minor for the new package, patch for the viewer).

## Summary by CodeRabbit

* **New Features**
  * Added Microsoft Graph support for browsing OneDrive and SharePoint files.
  * Added Azure AD OAuth sign-in with PKCE, token renewal, sign-out, and connection testing.
  * Added file search, folder navigation, downloads, revision listings, and change detection.
  * Added browser-compatible OAuth popup callback handling for restricted hosting environments.
* **Documentation**
  * Added setup and usage guidance for Microsoft Graph integration and OAuth configuration.
* **Tests**
  * Added coverage for authentication, file operations, pagination, downloads, revisions, and token refresh behavior.

### DIFF
--- a/.changeset/source-msgraph-provider.md
+++ b/.changeset/source-msgraph-provider.md
@@ -1,4 +1,5 @@
 ---
+'@ifc-lite/oauth-pkce': minor
 '@ifc-lite/source-msgraph': minor
 '@ifc-lite/viewer': patch
 ---
@@ -8,3 +9,7 @@ Add `@ifc-lite/source-msgraph`: a Microsoft Graph (OneDrive/SharePoint) file-sou
 Authentication is delegated OAuth 2.0 Authorization Code + PKCE (`@ifc-lite/oauth-pkce`), scope `offline_access https://graph.microsoft.com/Files.Read` — no admin consent required, no client secret. No client ID is committed; it's a required, non-secret `clientId` preference the deployment configures (see the package README for what to register in Azure AD).
 
 Registered alongside `@ifc-lite/source-dalux` in the viewer's `createRegisteredProviders()`.
+
+The popup handoff is a `BroadcastChannel` from the redirect page, not the usual `popup.closed`/`popup.location` poll. A host that serves `Cross-Origin-Opener-Policy: same-origin` (the viewer does, for `SharedArrayBuffer`) has its opener link severed by the cross-origin authorization hop: `popup.closed` reads `true` while the popup is visibly open, so the poll loop rejects every sign-in as "cancelled" before the user has even consented. The viewer now serves the redirect path as a small static page (`apps/viewer/public/oauth/msgraph/callback.html`, routed in dev by `apps/viewer/vite-plugins/oauth-callback.ts` and in production by a `vercel.json` rewrite) instead of letting the SPA fallback boot a second copy of the whole application inside the popup.
+
+Because that failure is a property of the popup being cross-origin rather than of any one provider, the waiting side ships as `waitForOAuthCallback` (plus the `OAUTH_CALLBACK_CHANNEL` name and its `OAuthCallbackMessage` shape) in `@ifc-lite/oauth-pkce`, so every provider built on that package shares one implementation. Messages are routed by the sign-in attempt's `state`, which is what keeps two concurrent sign-ins from completing each other's flow; `parseAuthorizationCallback` still performs the authoritative CSRF check. One consequence is deliberate: cancellation is no longer detectable, because `popup.closed` is the only signal a browser gives for it and that is exactly what COOP made unusable, so closing the popup now waits out the timeout.
--- a/.changeset/source-msgraph-provider.md
+++ b/.changeset/source-msgraph-provider.md
@@ -1,0 +1,10 @@
+---
+'@ifc-lite/source-msgraph': minor
+'@ifc-lite/viewer': patch
+---
+
+Add `@ifc-lite/source-msgraph`: a Microsoft Graph (OneDrive/SharePoint) file-source provider implementing `FileSourceProvider` from `@ifc-lite/plugin-api`. Browses the signed-in user's OneDrive (folders and files), lists version history, and downloads the current revision of a file via Graph's pre-signed `@microsoft.graph.downloadUrl` — never `GET .../content` directly, which 302-redirects in a way a browser can't follow under a CORS preflight.
+
+Authentication is delegated OAuth 2.0 Authorization Code + PKCE (`@ifc-lite/oauth-pkce`), scope `offline_access https://graph.microsoft.com/Files.Read` — no admin consent required, no client secret. No client ID is committed; it's a required, non-secret `clientId` preference the deployment configures (see the package README for what to register in Azure AD).
+
+Registered alongside `@ifc-lite/source-dalux` in the viewer's `createRegisteredProviders()`.
--- a/apps/viewer/package.json
+++ b/apps/viewer/package.json
@@ -51,6 +51,7 @@
     "@ifc-lite/pointcloud": "workspace:^",
     "@ifc-lite/query": "workspace:^",
     "@ifc-lite/source-dalux": "workspace:^",
+    "@ifc-lite/source-msgraph": "workspace:^",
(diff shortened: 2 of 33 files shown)


louistrue (approved):
Approving and un-drafting. Rebased onto post-#2632 main, all gates green, and the COOP guard is mutation-verified.

**One thing I got wrong in the instructions, worth recording because it nearly cost the fix.** I told the rebase to delete the branch's own `packages/oauth-pkce/` tree, on the assumption that main now carried the real package. It does not carry all of it: `packages/oauth-pkce/src/callback-channel.ts` is **absent from main** — #2632's squash shipped the package without the channel. Verified: `git rev-parse origin/main:packages/oauth-pkce/src/callback-channel.ts` fails.

A blanket delete would have removed the COOP fix itself. The rebase instead replayed only the five contributor commits that were not already merged, dropping the 13 stale foundation files by not replaying them. So **this PR is what puts the channel on main**, and that is the correct outcome rather than an accident of ordering.

**The guard binds, checked both directions:** reverting `auth.ts` to the 250 ms `popup.closed` poll fails the COOP test (exit 1, 1 of 9); restored, exit 0 with 61 passing and 5 declared skips. That matters here because the defect it guards is unobservable in a unit environment — under `Cross-Origin-Opener-Policy: same-origin` a cross-origin popup reports `closed: true` while visibly open, so the old poll rejected sign-in on its first tick. The stub popup reproduces that contract with a throwing `location` getter and a `locationReads === 0` assertion.

**Gates:** install, build (46/46), typecheck, lint, `check-source-text-assertions`, test (92/92), `check-api-surface` after build, `check-generated` — all exit 0, unpiped. One conflict during the rebase (`scripts/api-surface.json`), resolved. Force-with-lease with the remote confirmed unmoved. No client ID is committed.

**Four things no gate covers, for the record rather than as blockers:**

1. **The COOP fix cannot be verified in CI.** The test uses a stub; real confirmation needs a browser against a deployment actually serving `Cross-Origin-Opener-Policy: same-origin`. The Dropbox sibling of this fix *was* driven end to end by Louis and works, which is the strongest evidence available for the mechanism.
2. **`source-host.test.ts:139` asserts only `providers.length > 0`.** A provider silently dropped in a future conflict resolution would stay green.
3. **`CALLBACK_PAGES` and the `vercel.json` rewrite are kept in lockstep by comment only.** Keeping one and dropping the other passes every gate and breaks sign-in in exactly one environment.
4. **The live sign-in has never been executed** — no client ID by design, so the Azure AD app registration is a deployment prerequisite, and nothing here evidences the real OAuth round trip.

**#2635 stays a draft deliberately.** It must re-rebase after this lands (9 conflicts), and its current green is against a base about to move. Also worth knowing: it already deleted its own duplicate `callback-channel.ts` and now imports from `@ifc-lite/oauth-pkce`, copying this PR's file byte-identically on purpose — so nobody should "clean up" either copy independently.

Issue #2199, opened 2026-08-05
[Feature] Expand measurement tools: area, volume, length, angle, coordinates, and weight


xyzbety:
### The problem

The current viewer mainly provides basic length measurement and does not cover many common BIM review, quantity verification, and coordination workflows.

It would be useful to expand the measurement tools with capabilities similar to BIMVision.

Reference video:https://www.youtube.com/watch?v=mbnjYz2gGfY

The enhanced measurement system could include the following six categories.

## 1. Area

Support measurement of:

* Free polygon area and perimeter
* Rectangle area
* Individual geometry-face area
* Coplanar face area
* Surface area of one or more elements

For elements containing openings, the result should indicate whether opening areas are included or excluded.

## 2. Volume

Support volume calculation for one or more elements, including individual and aggregated results.

Where possible, the viewer may distinguish between gross and net volume.

For open or invalid geometry, the viewer should report that the volume cannot be calculated reliably instead of returning an incorrect result.

## 3. Length

Support measurement of:

* 3D distance between two points
* X, Y, and Z distance components
* Horizontal and vertical distance
* Geometry-edge length
* Polyline length
* Minimum distance
* Perimeter of a closed boundary
* Diameter and radius of circular or cylindrical geometry

Diameter measurement should be useful for pipes, circular openings, cylinders, and other round elements.

## 4. Angle

Support measurement of:

* Three-point angles
* Angles between two edges
* Angles between two faces
* Angles relative to horizontal or vertical directions
* Slopes

Results may support common formats such as degrees, percentage slope, and ratio slope.

## 5. Coordinates

Support reading the X, Y, and Z coordinates of any selected point in the model.

When the IFC file contains the relevant information, the viewer should distinguish between:

* Model-local coordinates
* Project coordinates
* World coordinates
* Georeferenced coordinates

It should also be possible to define a temporary reference point and display relative coordinate offsets.

## 6. Weight

Support weight or mass calculation for one or more elements.

Possible data sources include:

* Existing IFC weight or mass quantities
* Element volume and material density
* User-defined or project material-density libraries

When the result is calculated or estimated, the viewer should indicate its source and distinguish between mass units and force units.

## Common requirements

The measurement tools should use a consistent interaction model and support common geometry snapping, including:

* Vertices
* Edges
* Edge midpoints
* Faces
* Circle centers
* Intersections

It would also be useful to support:

* Unit and precision settings
* Multiple visible measurements
* Individual measurement deletion or visibility control
[...]

PR #2641, opened 2026-08-15
feat(viewer): multi-click polyline measurement mode (#2199)

BIMvoice:
## What this does

Adds a **multi-click accumulate-points mode** to the Measure tool, alongside the existing drag-to-measure gesture. This is one of the two structural prerequisites named in the maintainer's scope assessment on #2199 ("A multi-click measurement mode ... the tool is drag-only today") — the gate on most of the remaining open list (polyline/perimeter, free-polygon area, three-point angle).

A new **Distance / Polyline** toggle in the Measure panel (not either toolbar — see below) switches the gesture:

- **Distance** (unchanged): mousedown → drag → mouseup, exactly as before.
- **Polyline**: click to place a point, click again to extend. No drag is ever started in this mode.

## Finish / cancel gesture, and why

- **Double-click or Enter** finishes the sequence **open**, reporting the sum-of-segment length. Two gestures land on the same outcome because both are already established "finish" conventions elsewhere in the viewer (Enter finishes the AddElement polygon tool's own click-accumulate mode in `selectionHandlers.ts`'s `commitAddElementSlabPolygon`; double-click is the more discoverable in-canvas equivalent for a mouse-only user).
- **Clicking back near the first point** (screen-space radius, once ≥3 points are placed) closes the loop instead of adding a new point, finishing **closed** and reporting the **perimeter** (the same sum plus the closing segment). This mirrors how CAD/BIM polyline tools conventionally close a loop, and keeps "open" vs "closed" a single click-driven decision rather than a second control.
- **Escape** cancels an in-progress sequence — extends the existing `Esc cancels the active measurement` shortcut, which already exists for drag mode.

The panel always prints which basis a number was computed under ("Length" vs. "Perimeter (closed)") rather than leaving the convention implicit, per the panel discipline the maintainer set for #2199's Qty section (net/gross/mesh, each labelled).

## The one consumer: polyline length + perimeter

The smallest real feature on the maintainer's open list, and it exercises the whole path: multi-click accumulation, the finish gesture, the closed-vs-open distinction, and the existing `formatDistance`/`resolveQuantityDisplay`/`unitDisplayOverrides` unit-display path (no new conversion logic — `measure-modes/polyline.ts` is pure sum-of-segments math, nothing else).

## Architecture — no parallel state machine

- `measurementSlice.ts`: `measureMode: 'drag' | 'polyline'`, `activePolyline`, `polylineMeasurements`, and the actions (`setMeasureMode`, `startPolyline`, `addPolylinePoint`, `finishPolyline(closed)`, `cancelPolyline`, `deletePolylineMeasurement`). `setMeasureMode` is the one place mode-exclusivity is enforced: entering polyline mode cancels an in-progress drag; leaving it discards an in-progress click sequence.
- `measureHandlers.ts` stays store-free per its existing contract: `shouldStartDragMeasurement(mode, shiftKey)` is the pure gate `useMouseControls.ts`'s `mousedown` consults before ever calling `handleMeasureDown` — polyline mode never starts a drag. `raycastForPolylinePoint` mirrors the drag mode's magnetic-snap raycast exactly (same snap radius/options) so a clicked point snaps identically to a dragged one.
- `selectionHandlers.ts` (which already reads the store directly for the analogous AddElement click flow) adds `handlePolylineClick`, following the same click-state-machine pattern `handleAddElementDrop` already uses.
- `MeasurementVisuals.tsx` / `MeasurePanel.tsx`: the polyline overlay (placed segments, rubber-band to cursor, close-loop ring on the first point once ≥3 points) and the panel's Polyline toggle + result list, entirely inside the **existing shared panel** — neither toolbar gained anything.

**Neither toolbar hosts this UI.** The mode toggle lives inside `MeasurePanel.tsx`, same as Snap/Geo. `measure-parity.test.tsx` (unmodified) still passes — full green run below — since no new `PanelSection` was added and the toolbars still only dispatch `setActiveTool('measure')`.

## What I deliberately did not build

Free-polygon/rectangle area, geometry-face/coplanar-face area, three-point/edge/face angles, minimum distance, diameter/radius, circle-centre/intersection snapping. Each of those needs either mesh analysis reachable from TypeScript (the other structural prerequisite the maintainer named) or its own interaction beyond the polyline primitive shipped here — one mode, one consumer, per the issue's explicit scope discipline.

## Tests — RED/GREEN evidence

New tests: `measure-modes/polyline.test.ts` (pure math), `measurementSlice.test.ts` (+22 tests: accumulate 3+, finish open/closed with distinct numbers, reject-too-few-points, cancel mid-sequence, `clearMeasurements` clears polyline state, `setMeasureMode` mode-exclusivity in both directions), `measureHandlers.polyline.test.ts` (the drag-start gate, the raycast helper, the close-loop screen-radius check), `selectionHandlers.polyline.test.ts` (the click state machine end to end, including a direct assertion that it never touches `activeMeasurement`).

For each new module, broke the implementation, confirmed RED, restored via `git checkout`, confirmed GREEN:

- `measurementSlice.ts`: made `finishPolyline` always compute the open length → 2 subtests failed (the two closed-perimeter tests) → restored → 38/38 pass.
- `measureHandlers.ts`: made `isNearPolylineStart` always return `true` → 4 subtests failed across both handler test files → restored → 16/16 pass.
- `measure-modes/polyline.ts`: made `polylineLength(closed)` drop the closing segment → 2 subtests failed → restored → 7/7 pass.

Full run after restoring: `pnpm --filter @ifc-lite/viewer test` — **4267 tests, 4261 pass, 0 fail, 6 skipped** (pre-existing skips, unrelated). This includes the unmodified `measure-parity.test.tsx` and the pre-existing `measurementSlice.test.ts` drag-mode tests, proving the original drag-to-measure flow is unchanged.

## Gates

- `pnpm --filter @ifc-lite/viewer test`: 4267 tests, 0 fail.
- `pnpm --filter @ifc-lite/viewer typecheck`: clean.
- `pnpm lint` (FORCE_COLOR unset): 0 errors (21 pre-existing warnings in unrelated `scripts/`).
- `apps/viewer` is a private package — `api-surface.json` not touched, not required.
- Changeset added (`@ifc-lite/viewer` minor).

## Note on #2638

PR #2638 (in flight) changes `formatDistance`'s signature to add a `projectUnits` parameter. This PR does not touch `formatDistance.ts` at all — the new polyline label calls the existing two-argument `formatDistance(meters, unitDisplayOverrides)`, same as every other call site in this PR's diff, so there should be no merge conflict on that function's signature.

Refs #2199.

## Summary by CodeRabbit

* **New Features**
  * Added multi-click polyline measurements for open lengths and closed perimeters.
  * Added snapping, live previews, running totals, loop closing, unit labels, and Measure-panel integration.
  * Added Enter to finish and Escape to cancel active polylines.
  * Preserved drag-to-measure behavior and enabled switching between measurement modes.

* **Bug Fixes**
  * Improved cleanup when leaving the Measure tool or resetting viewer state.
  * Prevented duplicate points from double-click completion.
  * Added validation and error feedback for insufficient points.


louistrue (approved):
Approving. Conflict resolved, the dedup blocker fixed, and all six threads closed with reasoning on the PR.

**The blocker was the one that mattered.** `finishPolyline` ran a screen-space dedup on every finish path using coordinates reprojected each frame, so pressing Enter or closing the loop after orbiting to a near top-down view silently deleted deliberately placed vertices and reported a short length. A confidently wrong measurement that gets written into a report is worse than a crash, which is why this held the PR rather than being noted in passing.

Now scoped to the double-click path, where the duplicate genuinely originates: the browser dispatches `click, click, dblclick`, so `handlePolylineClick` has already appended the extra point. Enter appends nothing and the close-loop click returns before `addPolylinePoint`, verified at all three call sites. The `while` became a single `if`, since exactly one extra click is generated and removing more could only eat a real vertex.

**One thing from this PR is worth carrying beyond it.** The first attempt put the flag inline in `useMouseControls`' DOM handler. Mutating the fix away left the **entire viewer suite green at 4772 passing**, because no test can reach that hook. The flag now sits behind `finishPolylineFromDoubleClick` in `selectionHandlers.ts`, where the same mutation reds. That is not a missing test, it is a test at a level that could never observe the change, and it is the failure mode this repo has hit repeatedly. The mutation check is the only thing that catches it.

The fixture also does not hand-write colliding points. It places a well-separated trace and runs the real `updateMeasurementScreenCoords` with a plan-view projector, which is how production creates the collision, with a fixture-guard test asserting the collapse actually happened so the pair cannot pass for want of one.

**Conflict with main** (after #2680) was purely textual: your `hasPendingMeasurementState` sat immediately above the `boundsFromIndexedMeshes` that #2680 replaced with `boundsFromIndex`, and your import sat in the block it edited. Kept both sides, dropped the deleted helper, confirmed no references remain.

**Dismissed deliberately:** the two Codex asks to split `measurementSlice.ts` and `MeasurePanel.tsx`. The ~400-line TypeScript rule is self-policed per `AGENTS.md:18` and is not CI-enforced; only the Rust `module_size` ratchet gates.

**Residual, bounded and by design:** on the double-click path a view-ray collapse can still cost the trailing vertex, since that path must dedup. Limited to one point, and it is the one gesture where a trailing duplicate is expected.

Good feature. Sorry it sat as long as it did.

PR #2654, opened 2026-08-16
fix(clash): invalidate orphaned intersection-solid render on any clash-focus teardown

BIMvoice:
## Follow-up to #2574

Fixes a live defect found in the adversarial re-review of #2574's merged state (`5cf117d1e`): https://github.com/LTplus-AG/ifc-lite/pull/2574#issuecomment-5306157639

> ### Confirmed defect: 2 of the "reset clash focus" call sites never tear down the solid
>
> ... Two *other* sites reset exactly the same fields (selection, isolation, ghost, pair colours, overlap box/lines, `clashSelectedId`) by hand, outside the hook, and both were written before the solid feature landed — neither learned about it:
>
> - `apps/viewer/src/lib/tours/tours/clash.ts:106-117` — the "zoom-to-clash" tour step's `cleanup()`.
> - `apps/viewer/src/store/homeView.ts:7-23` — `resetVisibilityForHomeFromStore()` (the Home / "Show all" reset).
>
> Neither calls `clearClashSolid()`. Worse, neither *can* invalidate the in-flight compute at all: `solidRequestGuard` is a `useRef` private to one `useClash()` hook instance (useClash.ts:187), not store state, so no code outside that hook's own callbacks can ever bump it. `latest-wins.ts:14-15` claims `begin()` "is called synchronously at the START of every attempt (**including every teardown/clear path**)" — that's false for these two.
>
> Net effect: if a clash solid is showing (or an async compute for one is still in flight) and the user runs the clash tour to completion, or clicks Home / "Show all", the model comes back fully opaque/visible but the opaque intersection-solid mesh keeps rendering — an orphaned solid with nothing selected and no clash focus — or a since-superseded compute lands afterward and reintroduces the full-model ghost the user just cleared.

## Why this is structural, not two added calls

The obvious patch — add `s.clearClashSolid()` to the two named call sites — would have closed exactly those two paths and left the class of bug intact: `solidRequestGuard` (`apps/viewer/src/hooks/useClash.ts:187`) is/was a `useRef` private to one `useClash()` hook instance, so it is reachable only from that hook's own callbacks. A third teardown path added tomorrow would face the identical trap: the same fields (`clashSelectedId`, ghost, isolation, pair colours, contact overlay) reset by hand, no way to reach the guard, same orphaned solid.

The fix moves the invalidation into the one place every current *and future* teardown path already goes through: the clash store slice's own setters.

- `apps/viewer/src/store/slices/clashSlice.ts`: `setClashSelectedId`, `clearClashSolid`, and `clearClash` now each reset the solid presentation (`clashSolidStatus`/`clashSolidMesh`/etc.) and bump a new `clashSolidRequestSeq` counter — store state, not a hook-private ref.
- `apps/viewer/src/hooks/useClash.ts`: `focusClash`'s async intersection-solid compute now checks `clashSolidRequestSeq` instead of the old ref guard. Every call site that used to call `solidRequestGuard.current.begin()` (`discardSolidPresentation`, `selectElement`, `highlightAll`, `clearHighlight`, `clearAll`) drops that call — `clearClashSolid()` / `setClashSelectedId()` already bump the seq. `invalidateSolidCompute()` (the panel-unmount cancel) now bumps the store field directly, kept only as an explicit "cancel without resetting anything else" convenience.
- The old private `useRef` guard is gone entirely.

Because `clashSelectedId` is the field every "which clash is focused" reset already touches (it's the field the review lists in both broken call sites), any code — present or future — that clears or changes it invalidates an in-flight solid compute by construction. No call-site enumeration to keep in sync **for clash-focus teardown**.

That scope is deliberate, and an earlier revision of this section overstated it. It does *not* extend to the paths that replace or unload the **model** the presentation belongs to — `resetViewerState()`, `clearAllModels()` and `removeModel()` touched no clash field at all, so a resolved solid and a non-null `clashSelectedId` both survived and the render gate passed too. That was pre-existing on `main` rather than introduced here, and it is now fixed as part of this PR — see "Model-lifecycle teardown" below.

**Defence in depth, not the primary fix:** `apps/viewer/src/components/viewer/Viewport.tsx:604` (now ~613) also gates the solid draw on `clashSelectedId !== null`, in addition to `clashSolidStatus === 'solid'`. Given the store-level fix, `clashSolidStatus` should never read `'solid'` while `clashSelectedId` is null — but nothing in the type system enforces that invariant if a future call site writes `clashSolidStatus` directly without going through `setClashSelectedId`. The render-side gate is cheap and closes the visible symptom even in that case, so it's included, but the store-level fix above is what actually closes the class of bug.

Also fixed: `apps/viewer/src/lib/clash/latest-wins.ts:14-15` claimed `begin()` "is called synchronously at the START of every attempt (including every teardown/clear path)" — false, per the review's own finding. That module is now **removed** (see "Review follow-ups" below), so the claim goes with it.

## Model-lifecycle teardown (`e2529299e`, from the adversarial review)

Three paths ended clash focus by replacing or unloading the model, and touched no clash field:

| path | now |
|---|---|
| `store/index.ts` `resetViewerState()` — the primary-file "open another model" reset | `clearClash()` |
| `slices/modelSlice.ts` `clearAllModels()` — full federation teardown | `endClashScenePresentation(…, 'federation-cleared')`: `clearClash()` + `clearGhost()` + `clearIsolation()` + `setPendingColorUpdates(lens ?? {})` |
| `slices/modelSlice.ts` `removeModel()` — one model leaves a federation | `endClashScenePresentation(…, 'model-removed')`: `clearClashFocus()`, plus a **scoped** ghost / isolation / colour release (see below); the clash **result** is kept |

All three route through the store's own invalidation rather than inlining a field list, so the seq bump comes for free and there is still one place that knows what "the solid presentation" is.

`resetViewerState()` dropping the whole result is the same reasoning the `compareResult` / `zoneAssignments` / `searchIndexes` drops beside it already carry: a clash result is keyed by `model:expressId` pairs belonging to the outgoing model, and an IFCX recomposition reassigns expressIds outright. `removeModel()` keeps the result on purpose — a federated sibling leaving does not invalidate the pairs that do not involve it, and the result is a list the user is reading rather than a mesh in the live scene. Its presentation drop is unconditional rather than "only if the focused clash names this model": a clash id is `` `${ruleId} ${lo} ${hi}` `` with `lo`/`hi` themselves `model:expressId`, and parsing it in `modelSlice` would be a third, subtly different reading of a key format — the exact hazard the selection purge a few lines below calls out and routes through `stringToEntityRef` to avoid.

### Self-review round 2 — the other two channels, and the contract the first fix broke

`focusClash` paints into three channels, and only the first is reachable from a clash action:

1. the clash slice's own focus fields — `clearClashFocus()` / `clearClash()`;
2. the shared **visibility** channels (`ghostExceptEntities` / `isolatedEntities`, `visibilitySlice`). `applyFocusMode` writes exactly one per focus: `isolate` hides everything but the pair (one click from every panel row), `ghost` fades its context, and the resolved-solid path ghosts the *entire* model (`installClashGhost(new Set())`). Left behind, the survivors stayed translucent — or, in isolate mode, invisible;
3. the **colour-override** channel (`pendingColorUpdates`, `dataSlice`). `clashHighlightColors` is only a *record*; the albedo the user sees is pushed separately (`useClash.ts:538-539`) into a fire-and-forget effect (`useGeometryStreaming.ts:906-923` → `scene.setColorOverrides`) undone only by a *later* push. `null` is a no-op there — only a non-null **empty** map reaches `clearColorOverrides()`.

The first attempt at (2) cleared the ghost **unconditionally**, which destroyed a deliberate preservation contract: `syncSourceModel` calls `removeModel(modelId)` one line before `purgeStaleEntityState(modelId, replacementId)`, and that purge *keeps* the part of the user's X-ray / isolation still owned by a surviving model. An unconditional clear made those filters dead code on their only production path — "Sync from source" wiped the user's X-ray. Both fixtures were blind: the test seeded `ghostExceptEntities: null`, and the harness's `removeModel` is a stub that only deletes a map entry.

The clear is now scoped by the ownership evidence available at the store level (`removeModel` cannot read clash's install record — it is a `useRef` private to a `useClash()` instance):

- **a clash is focused** (`clashSelectedId !== null`, sampled *before* the clash clear nulls it) — the channel holds what `focusClash` put there; or
- **the set is empty** — "ghost everything" / "hide everything", the degenerate state no owner can want to keep and the resolved-solid path's own signature. `purgeStaleEntityState` already applies exactly this reasoning to an emptied isolate set (`syncSourceModel.ts:262-264`), and `setGhostExceptEntities` / `setIsolatedEntities` map a falsy argument to `null`, so no other owner installs one.

`clearAllModels()` clears outright: no model is left for either channel to refer to, and no purge follows. The colour channel is released on the same evidence (focused clash, or a recorded pair tint), so an unrelated model removal cannot switch off Pset / IDS / schedule colouring clash never took.

`removeModel()` is also a genuine no-op now for an id that is not loaded, matching `updateModel`. `syncSourceModel` and the collab room teardown can both re-enter with an already-removed id, and every other cleanup in `removeModel` is keyed to that model — but the clash teardown was not.

**RED evidence** (surgical mutation of the fix, never a whole-file revert):

| mutation | fails |
|---|---|
| restore the unconditional `clearGhost()` | `syncSourceModel` "keeps the ids owned by a surviving model" + "KEEPS a user X-ray when no clash is focused" |
| drop the `clearIsolation()` line | 3 tests |
| drop the `setPendingColorUpdates(…)` line | 3 tests |
| drop the `!models.has(modelId)` guard | "is a no-op for an unknown / already-removed model id" |
| seed `models: new Map()` instead of a real federation | 3 tests |

New/changed coverage: `store/modelLifecycle.visibility-ownership.test.ts` (7 cases, both directions of the guard), the federation seed + paint-channel assertions in `store/modelLifecycle.solid-teardown.test.ts`, and two `syncSourceModel.test.ts` cases that drive the **real** `removeModel` end-to-end through the resync.

**Known gap, pre-existing and out of scope:** `useClash.run()` writes its result with no staleness check (`useClash.ts:361-364`), so a run finishing after `clearAllModels()` can repopulate `clashResult` with pairs from unloaded models. That is on the write side, not the teardown side, and the "complete by construction" claim in the changeset is softened accordingly. Separately, `collabSlice.ts:729` replaces the active `ifcDataStore` in place on every peer edit with no clash teardown — an uncovered model-lifecycle path worth its own issue.

## Review follow-ups (same commit)

- `Viewport.tsx`'s draw gate is now `clashSelectedId !== null`, matching the comment above it (it read `clashSelectedId &&`). Behaviour-identical today — ids are never `''` — so the comment explains why `null` is the predicate worth naming.
- `lib/clash/latest-wins.ts` is **removed**, with its 5-case suite. This PR made it production-dead, and its own doc comment had become an explanation of why the code does *not* use it. A primitive with zero consumers cannot be kept honest by its tests: they prove it works, not that anything needs it. Git history has it if a call site ever wants it back.

## Repro tests (RED → GREEN → RED)

Recreated the two tests the review described (not committed by the reviewer — recreated from the review's spec) plus one additional assertion each on `clashSolidRequestSeq`:

- `apps/viewer/src/lib/tours/tours/clash.solid-teardown.test.ts`
- `apps/viewer/src/store/homeView.solid-teardown.test.ts`

**RED** (against the pre-fix code — `setClashSelectedId`/`clearClashSolid`/`clearClash` reverted to their `main` shape): all 4 subtests fail —

```
tour cleanup must drop the intersection-solid presentation
'solid' !== 'none'

cleanup() must invalidate any in-flight focusClash solid compute (clashSolidRequestSeq), ...
0 !== 0   (field didn't exist pre-fix / never bumped)

Home / "Show all" reset must drop the intersection-solid presentation
'solid' !== 'none'

the Home reset must invalidate any in-flight focusClash solid compute (clashSolidRequestSeq), ...
0 !== 0
```
`tests 4 / pass 0 / fail 4`

**GREEN** (fix restored): `tests 4 / pass 4 / fail 0`.

**RED again** (broke the fix by reverting *only* `setClashSelectedId` back to `set({ clashSelectedId })`, leaving everything else fixed): all 4 subtests fail again, identical failure shape — confirming the tests actually exercise the mechanism, not something incidental. Fix restored afterward, back to GREEN.

## Cleared by the review — kept true

- `run()`/`runDuplicates()` race fix (`discardSolidPresentation`) — untouched in behavior, still bumps `clashSolidRequestSeq` via `clearClashSolid()`.
- `isCurrent()` (now the `clashSolidRequestSeq` comparison) still gates the entire result-writing block in both the success and `.catch` paths.
- Volume formatting and the full-model-ghost-not-leaking-on-a-failed-compute behavior are unchanged by this diff.

## Gates run

- Viewer clash-related suite (18 files, incl. the 2 new tests): **89/89 pass** (85 baseline + 4 new).
- `packages/renderer` suite: **900/900 pass**.
- Root `pnpm typecheck` (1114 test files across the workspace): clean.
- `pnpm lint` (`FORCE_COLOR` unset): 0 errors (21 pre-existing warnings, all in unrelated scripts).
- `node scripts/check-unused-locals.mjs`: no new unused locals.

## Changeset

`@ifc-lite/viewer` patch, no licence header, describing the user-visible fix.

---
We never close or merge PRs from this side — over to @louistrue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale clash intersection solids appearing after clash-tour completion, Home, “Show all,” model resets, or model removal.
  * Clearing a clash selection now consistently removes highlighted solids, contact markers, overlap visuals, and related presentation.
  * Prevented outdated asynchronous results from restoring cleared or outdated clash visuals.
  * Preserved clash results when removing individual models while clearing stale visual overlays.

* **Tests**
  * Added regression coverage for clash cleanup across tours, panel unmounting, Home resets, model lifecycle actions, and in-progress solid computations.


louistrue (approved):
Approving. Reviewed with the CodeRabbit CLI locally.

**The blocker I raised is genuinely fixed.** `focusClash` takes ownership of the ghost at `useClash.ts:506-507` (`setGhostExceptEntities`), while the teardown at `modelSlice.ts:191-192` cleared only `clashSelectedId` and `clashSolid`. So removing a model while a clash was focused left the scene **ghosted with nothing selected and no way to tell why** — the same shape as the orphaned solid this PR exists to fix, one field over. That is now cleared on both that path and `clearAllModels`, with tests covering it.

**The core design remains the right one**, and it is worth restating since it is what makes this PR durable rather than a patch: moving the guard out of a `useRef` private to one `useClash()` instance and into store state (`clashSolidRequestSeq`, bumped by `setClashSelectedId` and `clearClashSolid`) means the teardown paths that reset fields directly against the store invalidate **by construction** rather than by remembering to. Future teardown paths get it for free.

**Merge ordering, checked rather than assumed.** Your green predated both #2677 (lens isolation, shared isolation channel) and #2665 (clash duplicates), so I did not treat it as still valid. Verified file-disjoint from both: #2665 touches `packages/clash` plus `persistence.ts` and `ClashSettingsDialog.tsx`, #2678 touches `packages/export` only, and neither intersects this PR's file list. Clean `merge-tree` against current main.

**One non-blocking note, already documented in-PR and benign:** `modelSlice.ts:200-203` keeps `clashResult` rows for a removed model. Unchanged from main, and harmless because `refOf` to `fromGlobalId` returns null after `federationRegistry.unregisterModel`, so `focusClash` returns early on a fully-stale row and a half-stale pair highlights only the survivor. Cosmetic list staleness, never an orphaned render.

**Refuted:** three "split this file into modules" Majors from the CLI. That is the self-policed ~400-line TypeScript rule per AGENTS.md:18, not a CI gate.

PR #2665, opened 2026-08-16
fix(clash): abstain on non-finite bounds, and correct the isotropy claim from #2530

BIMvoice:
Fixes the two live defects the post-merge adversarial review of #2530 found on `main` ([review comment](https://github.com/LTplus-AG/ifc-lite/pull/2530#issuecomment-5307159608)), plus the doc-only third. Both were verified independently before being fixed; the review's measurements reproduce exactly.

## 1. Gate polarity: non-comparable bounds asserted a pair instead of abstaining

> The new gate is written as two **rejections** […] Every `NaN` comparison is false, so a non-comparable value now falls through *both* rejections and is reported, where before it fell short of the acceptance and was dropped.

Confirmed. `boxesTouch` returns `true` for NaN bounds (its own comparisons are false too), `dist > positionTolerance` is false for NaN, and the NaN box is never evicted from the sweep — so it pairs with every element visited after it:

```
NaN default gate   : [["duplicates m BAD m FAR1","minor"],["duplicates m BAD m FAR2","minor"]]
NaN legacy IoU gate: []
```

`FAR1` and `FAR2` are 100 m and 500 m away. The deprecated branch on the same call site gets it right.

Fixed by restoring the acceptance polarity, one line:

```ts
if (!(dist <= positionTolerance)) return;
```

A bound the pass cannot compare is a bound it cannot judge — abstain, never assert. Both gates now agree.

**Reachability — corrected, this claimed more than was known.** An earlier revision of this PR said the gate was "unreachable through `elementsFromStep`, which drops non-finite vertices". That was wrong: `fromPositions` (`aabb.ts:38-50`) dropped `NaN` only as a side effect of `<` and `>` both failing against it, and `±Infinity` propagated straight into the bounds — two elements each carrying `-Infinity` on the same axis then give a `NaN` `boxDistance` that `boxesTouch` passes. The mechanism is proven ([review probe](https://github.com/LTplus-AG/ifc-lite/pull/2665#issuecomment-5307249594)); whether the geometry pipeline can actually emit an infinite vertex **is not established**, so treat that half as a hypothesis.

Rather than only narrowing the sentence, this PR now closes it at the source: `fromPositions` requires each coordinate to be finite, checked *after* the transform is applied and *per coordinate* — the same rule `NaN` already got by accident. It can only change a result for a buffer that already contained a non-finite coordinate, where the old output was an infinite bound that poisoned `boxDistance` into `NaN`. Coordinates a real file can produce are finite, so no viewer or CLI result changes for them.

Two new tests in `aabb.test.ts`, both RED first (`- "max": [1, 2, 1]` / `+ "max": [1, 2, Infinity]`, and the transform-overflow case).

## 2. `positionTolerance` is not isotropic — decision: keep the behaviour, correct the claim

Confirmed by bisection, matching the review to the reported precision:

```
200 mm wall [4, 0.2, 3] m      : 10.00 / 10.00 / 10.00 mm
2 mm plate  [1.2, 0.002, 2.4] m: 10.00 /  2.00 / 10.00 mm
```

A duplicated 2 mm cladding panel nudged 5 mm along its own normal is silently unreported. Not a regression — the old IoU gate missed the same pair — but `CHANGELOG.md` is published and asserts the opposite.

**Decision: document the real property, do not widen the gate.** Three reasons, the first two measured rather than argued:

**a. Widening breaks two assertions that exist to prevent false positives.** I inflated `boxesTouch` by `positionTolerance` — the widening the review describes — and ran the suite:

```
FAIL  does not pair two small elements that do not even touch
      expected [ { id: 'duplicates m a m b', … } ] to have a length of +0 but got 1
FAIL  does not pair two disjoint sheets offset along their own normal
      expected [ { id: 'duplicates m a m b', … } ] to have a length of +0 but got 1
```

Those are exactly the "5 mm fixing pairs with a neighbour" case `boxesTouch` was added to close, and the deliberate exclusion its JSDoc documents. Getting isotropy means deleting them. The new false positives would also land on the *same population* the change is meant to help — thin and small elements — so it is not a clean trade.

**b. The anisotropy has two independent enforcers, so widening the gate would not deliver isotropy anyway.** This is a correction to the review, which attributes it to `boxesTouch` alone. The sweep's eviction (`max[axis] < minA`) enforces the same touch condition on the axis it sweeps. In a bare two-element fixture the offset axis *is* the sweep axis (it is the only axis with any spread of box minima), so eviction alone produces the measurement. With `boxesTouch` widened and the sweep forced onto X by filler:

```
                      2-element   sweep axis X
off=0.0019 (< 2 mm)       1            1
off=0.0021 (> 2 mm)       0            1     <- eviction still capping at 2 mm
off=0.0099                0            1
off=0.0101 (> 10 mm)      0            0
```

Unmutated, both columns cap at 2 mm. So `min(positionTolerance, extent)` is enforced twice over, and a one-line widening of the gate is not the whole change it looks like.

**c. It is a limitation the change did not remove, not one it introduced.** No user is worse off than before 1.7.0.

So the behaviour stands and the prose is brought in line with it — in the published 1.7.0 changelog entry, on `positionTolerance`, on `boxDistance`, and on `boxesTouch` (which previously scoped the consequence to *zero-thickness* sheets, when it applies to anything thinner than the tolerance). The changelog now states `min(positionTolerance, extent on that axis)` with the measured table, names the second enforcer, and says plainly that a duplicated thin sheet offset along its own normal is not reported.

One existing test title carried the same overclaim — `has an effective tolerance equal to positionTolerance for every shape and axis`, which passes only because every shape in its list is thicker than the tolerance on all three axes. Its title is narrowed and its scope stated. **Its assertions are untouched.**

A new test pins the real property in *both* sweep configurations, so it cannot be half-broken by widening only one enforcer.

## 3. Doc-only: the sweep-eviction comment

> Eviction is exact rather than conservative: it drops only boxes that no longer touch on `axis`, and **a pair that does not touch is rejected by the gate anyway**.

True for the distance gate, false for the legacy IoU gate the file keeps alive, whose degenerate fallback matches disjoint boxes within `tol` per axis. The comment now says eviction is lossless for the default gate and *not* for `iouThreshold`, and notes that the axis-dependence predates the distance gate — so `iouThreshold`'s "restores the pre-1.7 matching gate" claim remains accurate. Behaviour unchanged.

## RED / GREEN / RED

Every fix was reproduced failing first, using the review's own constructions.

**Fix 1 (gate polarity)**

| | |
|---|---|
| RED (before) | `expected [ [ …(2) ], [ …(2) ] ] to deeply equal []` — the two 100 m / 500 m pairs above |
| GREEN (after) | `Tests 58 passed (58)` in `duplicates.test.ts` |
| RED again (mutate `!(dist <= tol)` back to `dist > tol`) | `FAIL abstains on non-finite bounds instead of asserting a pair` — `expected [ { …(9) }, { …(9) } ] to have a length of +0 but got 2` |

**Fix 2 (pinned tolerance property)**

| | |
|---|---|
| RED (before, against the changelog's claim) | `expected 1.9999999999981803 to be close to 10` |
| GREEN (after, pinning the true property) | `Tests 58 passed (58)` |
| RED again (widen `boxesTouch` by 10 mm) | `FAIL effective tolerance is min(positionTolerance, extent) per axis` — `expected 9.99999999999092 to be close to 2` |

The second mutation is the informative one: it is precisely the "make it isotropic" change, and the test that pins the documented property is what catches it.

## Gates

- `packages/clash` typecheck: clean.
- `packages/clash` tests: **264 passed / 266**, against a **262 / 264** baseline on the same `upstream/main` checkout. The two failures are `differential.test.ts` WASM-vs-TS kernel parity, pre-existing in this environment (no WASM binary built) and untouched by this change. Net: +2 tests, 0 new failures.
- Viewer clash suites (`duplicate-set-sections`, `useClash.duplicates`, `clash-colors`, `latest-wins`): **17 / 17**.
- `check:api-surface` ✅ (no public surface change), `check:source-text-assertions` ✅, `check:test-wiring` ✅.

## What this deliberately does not touch

The review cleared a long list, and none of it is disturbed here: multi-part severity determinism, sweep-eviction/gate consistency for the default gate, clash-id vs node-identity agreement, `duplicateTolerance` being metres end to end, and the panel guards degrading rather than mislabelling. No assertion is weakened, no fixture adjusted, no gate relaxed.

Changeset included (patch, `@ifc-lite/clash`).

## Review follow-ups (`088118712`)

- **The falsified claim survived in the one string a user reads.** `ClashSettingsDialog.tsx:223`'s "Duplicate tolerance" hint was word-for-word the `positionTolerance` JSDoc opening sentence — the sentence this PR corrects everywhere else. It now adds "Capped by each element's own thickness: a 2 mm plate gets 2 mm across its thickness, not the full value." The same unqualified sentence in `apps/viewer/src/lib/clash/persistence.ts:43` (developer-facing) is qualified too. Wording only, no behavioural change and no test.
- **The reachability claim is corrected and the hole is closed at the source** — see the rewritten paragraph in §1 above. Two RED-first tests.


louistrue (approved):
Approving. Reviewed with the CodeRabbit CLI locally.

**The blocker I raised is genuinely fixed, and I checked the half I expected to still be wrong.** The `duplicates.ts` sweep comparator returning `NaN` for non-finite elements gave V8 an arbitrary permutation and evicted early, so real duplicates elsewhere in the model vanished: a 42-element fixture returned `[]` where it should have returned one pair. That is closed.

I was specifically sceptical that the TEST fixture would be fixed alongside the comparator, because the original passed only by accident: at 4 elements it took TimSort's insertion path, which is used below 22, so the broken comparator could not express itself. Fixing the comparator without scaling the fixture would have left an unfalsifiable test guarding a real fix. It holds up now.

**One non-blocking FYI, so it is on the record rather than a surprise later.** `duplicates.test.ts:594-596`: the `withInfinite` sub-assertion is not independently binding. Removing the NaN assertion and re-running against the reverted subtraction comparator, that sub-assertion still passes, because two `fromPositions`-inverted elements are not enough perturbation to lose a pair. It is worth keeping since it is the adapter-reachable half (`adapters/ifcx.ts:122`, `adapters/step.ts:186`), but only the SDK-direct NaN half actually reds. The test as a whole binds the fix, so this is a weak sub-assertion rather than a guard that cannot catch its own regression.

**Refuted:** the CLI's "exclude non-finite elements before the sweep" is a refactor preference, not a defect. `+Infinity` keys already sort strictly after every finite element, so they never sit in `active` during finite comparisons.

Also confirmed: no allowlist grew, no budget was raised, and nothing here fixes a tolerance or epsilon that an expected-failure marker on main names, so there is no marker to remove in the same commit.

Issue #2491, opened 2026-08-09
export: 'empty source implies zero-length refs' is an unstated invariant; violating it emits a dangling reference


louistrue:
Surfaced while adversarially reviewing #2485, which asked whether `step-exporter.ts`'s source-presence checks could silently emit a bad file. The answer today is no — but only because of an invariant nothing states or enforces.

**The invariant: an empty source implies zero-length entity refs.**

The reviewer built the hostile shape that violates it — a parsed store whose `source` is swapped to `EMPTY_SOURCE_BYTES` while the real entity index is kept, so refs still carry `byteLength > 0` — and the exporter silently emits a **corrupt file**: three new property-set lines including a relationship referencing wall `#8`, whose own line has vanished. A dangling reference, no error, no warning.

It is not reachable today. Every production producer was audited:

- the only source-less store builder is `apps/viewer/src/utils/serverDataModel.ts`, which adds every ref as `(0, 0)` (~line 327);
- every other producer pairs a source with an index parsed from those same bytes.

So the eight truthiness checks in `step-exporter.ts` are bounded by a downstream `byteLength === 0` test *given* the invariant, and #2485 correctly concluded there is no live bug there.

What makes this worth recording rather than forgetting: the invariant is **unstated and unchecked**, and the repo has already shipped one thing that bends this exact area — the source accessor from #2339/#2360, which deliberately keeps `byteLength` so presence-style guards pass while `process()` meshes nothing. A future producer that constructs a store from an index and attaches bytes later, or a partial/streaming source, would violate it without anything complaining.

Two candidate directions:

1. **Assert it where stores are constructed** — if `source.byteLength === 0`, require every ref to be `(0, 0)`. Cheap, catches the violation at the boundary rather than in the exporter.
2. **Stop relying on it** — have the exporter test the ref it is about to read rather than the source's presence, so a truncated or mismatched index degrades predictably instead of emitting a dangling reference.

The second is more robust; the first is a smaller change and would make the invariant discoverable. Whichever is chosen, the hostile shape above is the test case, and it should fail before the fix.

Related: #2345 (the truthiness checks), #2339/#2360 (the accessor that keeps `byteLength`), #2485.

PR #2678, opened 2026-08-16
fix(export): StepExporter answered from the wrong record on an out-of-range source ref — and main documented it as safe

BIMvoice:
## The claim, and the probe that refutes it

`step-exporter.ts` documented its byte-range gate like this (#2398, merged — this is `main`'s code):

> "Measured, that rationale holds for every out-of-range shape EXCEPT a negative offset."

It does not. An **overrunning end** is equally dangerous, and unlike the negative offset it is *reachable*.

`IfcSourceBytes.decodeUtf8` **clamps** a range it cannot address. A clamp does not empty the range — it moves an endpoint onto a real file byte, and the window that survives still holds somebody else's record. Probe on a two-record source, `#1` given `byteOffset: 0, byteLength: 9999`:

```
overrun name(#1)    = "SetA"        <- right, by luck: that pattern is unanchored
overrun propIds(#1) = [201,202]     <- WRONG: these are #2's members (#1's are [101,102])
isReadableSourceRef(#1) = false
```

The readers whose patterns are `$`-anchored match at the end of the **clamped** window, i.e. against whatever record the file ends on:

```js
entityText.match(/\(\s*(#[^)]+)\s*\)\s*\)\s*;$/)   // getPropertyIdsInSet
entityText.match(/,\s*#(\d+)\s*\)\s*;$/)           // getRelatedPropertySet
```

That is a confidently wrong answer, exactly what the docstring reserved for negative offsets. Downstream, `retainSharedAtoms` does `skipIds.delete(atomId)` for every id `getPropertyIdsInSet` returns, so a member list read out of the wrong record un-skips the wrong atoms.

The reasoning was also inverted. The docstring argued the dangerous shape was the one `isReadableSourceRef` **accepts**; measured, the dangerous shape is the one it **rejects**:

```
last isReadableSourceRef(#2) = false
last name(#2)    = "SetB"          <- entityLineText answers anyway
last propIds(#2) = [201,202]
```

So the claim that tightening to `isReadableSourceRef` "would change no reachable answer" was wrong on both counts.

**This defect is pre-existing, not introduced by #2398.** Running the same probe against `79503d334^` — the commit before #2398 — gives byte-identical output, `[201,202]` included. #2398 collapsed five reads into one `entityLineText` and left every reader's regex untouched. What it added was prose documenting the behaviour away.

The negative-offset direction the docstring did argue about is genuinely unrepresentable, and for a stronger reason than the one given: `CompactEntityIndex` keeps offsets in a `Uint32Array`, so `-2` cannot round-trip through the index those readers consult at all. The probe above used a plain-`Map` index, which can hold one.

## What changed

**1. The readers are gated on `isReadableSourceRef`.** This is the decision the review asked for, taken on the following evidence rather than on the aesthetics of one predicate:

- The source-iteration pass in `export` **already** does `if (!isReadableSourceRef(entityRef)) continue;`. A record whose ref is unreadable does not get its line written. Before this change, the incidental readers were making decisions — which atoms to keep, which pset is being replaced — on behalf of a container that the same export had decided not to emit. Two paths, one question, two answers.
- The only answer the tightening costs is one that was right *by luck*: an overrunning ref on the file's **last** record clamps back to exactly that record's text. That record is dropped by the emission pass anyway, so keeping the answer only preserved the disagreement.
- The degradation is the shape the exporter already handles and has tests for: a record with no emittable bytes — nothing generated for it, nothing naming it, no dangling ref.
- Measured cost: **zero existing tests change answer.** 705 passed / 30 skipped, against 699 / 30 before, the difference being the six new cases.

`isReadableSourceRef` is now built once in the constructor rather than per call, which also removes the per-call `createSourceRefReader` allocation at the owner-history read.

**2. The rationale lives in one file.** The measured account of both out-of-range directions is in `source-ref-bounds.ts`, which owns the predicate. `step-exporter.ts`'s `entityLineText` docstring cites it and adds only what is specific to that call site (the `retainSharedAtoms` consequence and the agreement with the emission pass). An argument kept in two files is an argument that has to stay true in two files.

**3. The byte range's START is pinned.** Mutating `entityLineText`'s decode to `byteOffset + 1` while leaving the end alone — the record's line minus its leading `#` — passed **all 699** runnable tests in the package. Nothing any reader parses lives in the first byte: every pattern is unanchored or `$`-anchored, and the express id is supplied by the caller rather than read out of the text. `entity-line-text-bounds.test.ts` now asserts the decoded text equals the record's line, which is the only assertion that can see the start move.

RED evidence, that mutation applied to this branch:

```
× decodes the record from its first byte, not one byte in
Tests  1 failed | 704 passed | 30 skipped (735)
```

(For the record, the *whole-window* shift — `byteOffset + 1` at both ends — is already pinned; it turns 17 tests red. It is the start alone that nothing sees.)

The two overrun cases were shown RED against unmodified `main`:

```
× answers nothing for a ref whose end runs past the source
    expected [] to deeply equal [ 201, 202 ]
× answers nothing for an overrunning ref on the last record too
    expected 'SetB' to be null
```

**4. Three factual prose errors on `main`, and one misattribution:**

| Where | Said | Is |
|---|---|---|
| `sourceless-store-export.test.ts:180` | six readers, incl. `replaceEntityAttribute` | five — that method was deleted by #2469 |
| `.changeset/step-exporter-source-guards.md:6` | "six line readers" | five (its own body already lists five) |
| `step-exporter.ts:77` | "without touching these **eight** reads" | four — #2398 collapsed five into one |
| `sourceless-store-export.test.ts:274` | `filterHiddenRefsFromRelationshipLine` (#2398) | #2580, extended by #2637 |

## Verification

- `pnpm --filter @ifc-lite/export test` — **705 passed, 30 skipped**. One file, `lod1-generator.test.ts`, is excluded in this environment only: it fails to resolve an unbuilt `@ifc-lite/wasm` and does not touch this diff.
- `pnpm --filter @ifc-lite/export typecheck` — OK (51 test files)
- `node scripts/check-source-text-assertions.mjs` — OK (6 allowlisted, 0 new)
- `node scripts/check-test-wiring.mjs` — OK
- `pnpm exec oxlint packages/export/src` — no new findings

No assertion weakened, no fixture adjusted, no skip added, no allowlist entry, no ratchet raised.

## References

- #2398 — where the false rationale was written. The defect predates it; the prose does not.
- #2668 — draft, and was about to copy the same claim into `source-ref-bounds.ts` ("An overrunning END, by contrast, is genuinely benign"). Corrected on that branch in `df2f09efc`, prose only: the argument stays in `source-ref-bounds.ts` there too, so the two branches do not each carry a copy. It remains a draft.
- #2491 — introduced `isReadableSourceRef` and `source-ref-bounds.ts`.

### DIFF
--- a/.changeset/step-exporter-line-reader-bounds.md
+++ b/.changeset/step-exporter-line-reader-bounds.md
@@ -1,0 +1,20 @@
+---
+"@ifc-lite/export": patch
+---
+
+`StepExporter`'s incidental line readers no longer answer from another entity's record when a source ref is out of range.
+
+`entityLineText` gated on `byteLength === 0`, on the stated grounds that an out-of-range ref degrades to "a clamped, empty decode, which is the same answer". It does not. `IfcSourceBytes.decodeUtf8` clamps an unaddressable range onto real file bytes, so the window that survives holds a DIFFERENT record. On a two-record source, giving `#1` the ref `(byteOffset: 0, byteLength: 9999)`:
+
+```
+getPropertySetName(#1)  = "SetA"      <- right, by luck: that pattern is unanchored
+getPropertyIdsInSet(#1) = [201, 202]  <- wrong: those are #2's members (#1's are [101, 102])
+```
+
+The `$`-anchored patterns (`getPropertyIdsInSet`, `getRelatedPropertySet`) match at the end of the CLAMPED window, i.e. against whatever record the file ends on. `retainSharedAtoms` then calls `skipIds.delete(atomId)` for every id returned, so a member list read out of the wrong record un-skips the wrong atoms.
+
+The readers are now gated on `isReadableSourceRef` (#2491), the same predicate the source-iteration pass already uses to decide whether a record's line is emitted at all — so the two passes agree, instead of one making decisions on behalf of a container the other had decided not to write. A record with an unreadable ref degrades to the shape the exporter already handles: nothing generated for it, nothing naming it.
+
+The defect is pre-existing, not introduced by #2398 — the same probe gives `[201, 202]` on the commit before it. What #2398 added was a docstring arguing the behaviour was safe for every out-of-range shape except a negative offset; that docstring, and the matching rationale in `source-ref-bounds.ts`, are corrected to the measured behaviour, in one place with the other citing it.
+
+Also pinned: the byte range's START. Advancing `byteOffset` by one while leaving the end alone previously passed every test in the package, because no reader parses anything from the record's first byte.
--- a/.changeset/step-exporter-source-guards.md
+++ b/.changeset/step-exporter-source-guards.md
@@ -2,7 +2,7 @@
 "@ifc-lite/export": patch
 ---
 
-Make the dead `if (!dataStore.source)` guards in `StepExporter`'s six line readers live, without changing an answer. `IfcDataStore.source` is a mandatory accessor — a model that kept no bytes carries `EMPTY_SOURCE_BYTES`, not `null` — so those guards never fired. They were also redundant where they sat: a zero-length range decodes to `''`, which fails every regex the readers below them run.
+Make the dead `if (!dataStore.source)` guards in `StepExporter`'s five line readers live, without changing an answer. `IfcDataStore.source` is a mandatory accessor — a model that kept no bytes carries `EMPTY_SOURCE_BYTES`, not `null` — so those guards never fired. They were also redundant where they sat: a zero-length range decodes to `''`, which fails every regex the readers below them run.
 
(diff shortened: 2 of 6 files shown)


louistrue (approved):
Approving. Reviewed with the CodeRabbit CLI locally, since the hosted check row is absent here and would not have been evidence anyway.

Zero unresolved threads, CI green with the runs bound to this head, clean against current main, and disjoint from every other PR in flight tonight.

This touches the STEP exporter, which writes files users hand to other tools, so a wrong record produces a corrupt-but-plausible export rather than a loud failure. That earns a closer read than its size suggests, and it holds up.

Two cosmetic notes, neither blocking:
- `.changeset/step-exporter-line-reader-bounds.md:9-12` has an unlabelled fenced block.
- `.changeset/step-exporter-source-guards.md:5-7` still says "without changing an answer" while this PR's own changeset states the answer change. Both land in one CHANGELOG entry for `@ifc-lite/export`, so they read as contradicting each other. Worth a one-line edit whenever you next touch that package.

The CLI's "split this file into modules" Major is refuted: that is the self-policed ~400-line TypeScript rule per AGENTS.md:18, not a CI gate.

Issue #2670, opened 2026-08-16
Standalone WebGPU viewer: sun cast shadows + light-hardness controls (Blender-like)


Blogbotana:
### The problem

In the standalone WebGPU viewer (no Cesium world-context), the Sun & Sky lighting is **direct-only**: sun + hemisphere ambient + fill/rim + ACES tonemap. Surfaces are lit as if nothing occludes them, so the scene reads flat compared to a tool like Blender. Two things are missing for a convincing look:

1. **Cast shadows from the sun** — a slab bed casting onto the floor, walls onto each other. Today real sun shadows exist **only** in the Cesium world-context path (Cesium's own shadow map). The standalone path has none.
2. **Control over "light hardness"** — sun contrast vs. ambient/fill, terminator softness, and shadow penumbra width (hard vs. soft shadows).

What we *do* have already:
- `VisualEnhancementOptions` contact shading = **SSAO** (`packages/renderer/src/post-processor.ts`): local darkening in crevices/contacts, with radius + intensity. This is contact occlusion, **not** cast shadows.
- Direct lighting knobs exist in the environment uniform (`packages/renderer/src/environment.ts`): `sunIntensity`, `ambientIntensity`, `fillIntensity`, `rimIntensity`, `exposure` — but nothing surfaces them per-parameter in the UI (only the preset dropdown + exposure slider in `SunSkyPanel.tsx`).

### Proposed solution

Split the work into two phases so the cheap win ships first and the expensive one is scoped on its own:

**Phase 1 — "Light hardness" panel (cheap, no shadow map).** A Rendering/Quality section in the Sun & Sky panel exposing the existing environment uniforms (sun vs. ambient/fill balance, terminator softness — currently a fixed `wrap = 0.3` in the main shader) plus contact-shading (SSAO) strength/radius. No new render pass; gives immediately punchier, more contrasted lighting and stronger contact shadows.

**Phase 2 — Real sun cast shadows (large, perf-sensitive).** Classic sun shadow mapping:
- a depth pre-pass of the scene from the sun direction (ortho frustum fit to model bounds) → shadow map;
- the main fragment shader samples it to decide lit/shadowed;
- **softness/hardness = PCF (or PCSS) kernel width**, exposed as a "Shadow softness" slider — the Blender-area-light feel.

Key constraints I already see in this repo:
[...]

PR #2688, opened 2026-08-17
feat(renderer,viewer): light-hardness + terminator-softness controls (Phase 1, #2670)

Blogbotana:
## What and why

Phase 1 of the Blender-like lighting work (#2670). In the standalone WebGPU viewer the lighting was preset + exposure only; there was no way to control shadow *feel*. This adds two user-facing controls and the one renderer field they need. Maintainer picked the phased plan in #2670 (cheap light-hardness controls now, real cast shadows as a separate perf-gated Phase 2), and endorsed the `sunSoftness`/trim design there.

**Renderer (`@ifc-lite/renderer`)** — `LightingEnvironment` gains a `sunSoftness` field: the diffuse wrap that sets the sun terminator, previously hardcoded to `0.3` in `main.wgsl`. `0` is a crisp light/shadow boundary (harder shadows), larger softens it (overcast). Packed into a spare pad slot of the existing environment uniform (**no UBO size change**) and clamped to `[0, 1]`; omitting it reproduces the historic look exactly.

**Viewer (`@ifc-lite/viewer`)** — the Sun & Sky panel adds two sliders (WebGPU shading, hidden in world-context mode):
- **Light hardness** — deepens shadows by cutting the hemisphere ambient + fill.
- **Terminator softness** — trims the preset's `sunSoftness`.

Both are user *trims* composed onto the active preset (the preset supplies the base, switching presets changes the base, the trims persist), mirroring how Exposure already works — the "if it describes the light, it's a preset" split from #2670. Presets now carry per-preset softness (crisp Day/Evening, soft Overcast) so the terminator changes with the sky being simulated. Settings persist in localStorage.

The composition is a pure `composeLightingEnvironment()` helper (no store, no React), so it is unit-tested against the resulting uniform values rather than through the Viewport component.

Not in this PR (deliberately, per #2670): real sun cast shadows (shadow mapping + softness = penumbra) — that is the perf-gated Phase 2. An SSAO "Contact shadows" slider was prototyped and dropped: the existing screen-space AO is a sub-pixel hairline on BIM models and can't deliver the "slab-on-wall" darkening users expect; that belongs to Phase 2's cast shadows.

## How it was verified

- `pnpm turbo run test typecheck --filter=@ifc-lite/renderer` — **929 pass, 0 fail**, typecheck clean.
- `@ifc-lite/viewer` typecheck clean; new `compose-environment.test.ts` passes (**8/8**), asserting behaviour through stated invariants:
  - softness trim multiplies the preset base wrap; hardness divides ambient + fill and composes *over* the solar override; Cesium keeps the sky off; and the end-to-end renderer clamp (an over-soft trim resolves to wrap `1`, a zero trim to `0`).
  - renderer `environment.test.ts` pins `sunSoftness` default `0.3`, `[0,1]` clamp, non-finite → `0`, and its uniform slot.
- `pnpm api-surface:update` — export surface matches (additive field only).
- Manual: dev viewer (Edge/WebGPU) — both sliders change the look live; trims persist across preset switches and reload; controls hide under world-context. Tested by the branch author.

- [x] `pnpm test` (renderer) passes locally; viewer typecheck clean (viewer's full `pnpm test` uses a unix-only `find | sort` invocation that fails on a Windows shell — unrelated to this change; runs on CI/Linux)
- [ ] Geometry/WASM change: n/a

## Checklist

- [x] A test asserts the new behavior through a stated invariant
- [x] Published `packages/*` touched: added a changeset (`@ifc-lite/renderer` + `@ifc-lite/viewer` minor); ran `pnpm api-surface:update` (additive field)
- [x] Geometry-output change: none (shading/lighting only)
- [x] House rules: no `as any` / `@ts-ignore`, no silent `catch {}`, no oversized module, no repo-wide `cargo fmt`

Addresses #2670 (Phase 1). Phase 2 (real sun cast shadows) remains open in the same issue — do NOT auto-close on merge.

## Summary by CodeRabbit

- **New Features**
  - Added WebGPU controls for exposure, light hardness, and terminator softness.
  - Settings persist between sessions and can be reset to neutral values.
  - Lighting presets now provide distinct shadow characteristics for daylight, overcast, golden-hour, and night scenes.
  - Solar and world-context views automatically adapt lighting behavior; controls are hidden in world-context mode.
  - Added configurable sun softness for more natural, adjustable shadow transitions.

### DIFF
--- a/.changeset/light-hardness-phase1.md
+++ b/.changeset/light-hardness-phase1.md
@@ -1,0 +1,10 @@
+---
+"@ifc-lite/renderer": minor
+"@ifc-lite/viewer": minor
+---
+
+Phase 1 of the Blender-like lighting work (#2670): expose light-hardness and shadow-feel controls in the standalone WebGPU viewer.
+
+**Renderer** — `LightingEnvironment` gains a `sunSoftness` field: the diffuse-wrap that sets the sun terminator, previously hardcoded to `0.3` in the shader. `0` is a crisp light/shadow boundary (harder shadows), larger values soften it (overcast). Resolved into the existing environment uniform (a spare pad slot, no UBO size change) and clamped to `[0, 1]`; omitting it reproduces the historic look exactly.
+
+**Viewer** — the Sun & Sky panel adds two sliders (WebGPU shading, hidden in world-context mode): **Light hardness** (deepens shadows by cutting hemisphere ambient + fill) and **Terminator softness** (trims the preset's `sunSoftness`). Both are user trims composed onto the active preset — switching presets changes the base, the trims persist — mirroring Exposure. Presets now carry per-preset softness (crisp Day/Evening, soft Overcast) so the terminator changes with the sky being simulated. Settings persist in localStorage.
--- a/apps/viewer/src/components/viewer/LightingTrimControls.tsx
+++ b/apps/viewer/src/components/viewer/LightingTrimControls.tsx
@@ -1,0 +1,89 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * WebGPU shading trim controls for the Sun & Sky panel: Exposure, Light
+ * hardness and Terminator softness. All three are user trims composed onto the
+ * active lighting preset (the preset supplies the base; switching preset
+ * changes the base, the trims persist), so they share one control shape — a
+ * labelled slider whose value doubles as a reset-to-neutral button.
+ *
+ * Rendered only in standalone (WebGPU) mode; in world-context the model is lit
+ * by Cesium, so the caller gates this on `!cesiumEnabled`.
+ */
+
+import { useViewerStore } from '@/store';
+import { cn } from '@/lib/utils';
+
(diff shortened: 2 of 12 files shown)


louistrue (approved):
Approving. Phase 1 of #2670, well scoped and correctly stopped at the Phase 1/2 line.

Verified rather than assumed: the uniform packing takes the slot `_pad0` held in both `environment.ts:240` and `main.wgsl.ts:41`, byte layout stays 80, no neighbour disturbed (mutating `buf[17]` to `buf[18]` moves the suite 843/3 to 841/5, so that guard is real). `default` stays `{}` and absent localStorage keys resolve neutral, so an existing scene renders byte-identical after upgrade. CI genuinely ran with every run bound to the head.

Added one guard as `dd0e4bb1f`, explained in my comment above: reverting `main.wgsl.ts:371` to a literal left both suites byte-identical, so the one line that makes the feature visible was the one line nothing tested. Now mutation-verified.

The a11y point about the `label` wrapping the reset button is real but **pre-existing** — the Exposure block on main has the identical shape and this PR lifted it verbatim into a shared component. Worth fixing on its own, not as a condition of this.

Also backing two of your calls explicitly: dropping the SSAO contact-shadows prototype when it did not deliver the effect, and not expressing softness in degrees. Degrees is Phase 2 language for a penumbra that does not exist yet; a unitless trim on a diffuse-wrap term is the honest unit here.

PR #2696, opened 2026-08-17
fix(clash): discard a detection result whose federation is gone

BIMvoice:
## The defect (pre-existing)

`apps/viewer/src/hooks/useClash.ts` published a detection result at **two** sites, each followed by `bumpClashRunSeq()`, with **no staleness check**:

```ts
state.setClashResult(res);
state.bumpClashRunSeq();
```

`clashRunSeq` is documented in `clashSlice.ts` as bumped *after* each successful `setClashResult` and never reset — a **completion signal, not a cancellation guard**. It cannot answer whether the write should happen at all.

So: start a run on a large federation, then "Open file" or "Clear all" mid-run. `clearAllModels()` / `resetViewerState()` run `clearClash()`, and seconds later the finishing run repopulates `clashResult` with pairs from models that are no longer loaded. Those rows are inert — `focusClash` bails at `refs.length === 0` — so the user sees a populated clash list where every row does nothing, with no indication why.

**This is pre-existing**, not introduced by this PR.

### Reproduction (observed, before the fix)

Driving the real `run()` and the real `clearAllModels()` at the store/hook level, with the teardown fired mid-flight from the engine's own narrow-phase progress event:

```
not ok 1 - "Clear all" mid-run: the finishing run must not repopulate clashResult
  error: |-
    a run over models that are gone must not repopulate the result list
    + actual - expected

    + {
    +   clashes: [
    +     {
    +       a: {
    +         key: '0aaaaaaaaaaaaaaaaaaaaa',
    +         model: 'A',
```

`models.size === 0`, yet `clashResult` holds pairs naming model `A`. All four interleaved cases failed the same way; the four normal-run companions passed. `# pass 4  # fail 4`.

## The guard, and why this one

A run needs to know whether the world it examined still exists when it lands.

The obvious mechanism is a generation counter bumped by every teardown. That makes correctness depend on an **enumeration of call sites**, which is only as good as whoever adds the next teardown remembering it. `clashSolidRequestSeq` gets away with that only because every bump routes through one constant (`CLASH_FOCUS_RESET`); there is no equivalent choke point for "the models changed" — `removeModel`, `clearAllModels` and `resetViewerState` share `endClashScenePresentation`, but `dataSlice.setIfcDataStore` does not.

So the token is **not published by the teardowns at all** — it is read off the federation, at the one place the run actually reads it, and re-read at the publish site. In the spirit of #2654 (publish the real state rather than infer it), but one step further: there is no parallel record to keep in sync, because the record *is* the federation.

`gatherElements()` now returns, alongside the elements, a `ClashFederationIdentity`: for every model it actually took elements from, that model's id mapped to its `ifcDataStore` **reference**. That is exactly what a `ClashResult`'s refs depend on — `elementsFromStep` derives every `ref` from an express id in that store. Captured inside the same loop, off the same state snapshot the elements come from, so the two cannot describe different worlds.

Captured **there** and not at the top of `run()` deliberately: before the opening `await nextFrameOrTimeout(...)`, a teardown landing during that frame would invalidate a run whose elements were then gathered from the *new* federation — a perfectly current result, silently discarded.

Both write sites go through **one** `publishClashResult(...)` helper, which holds the staleness check *and* the `setClashResult` → `bumpClashRunSeq` pairing, so neither can drift (two sites with two copies of a check is this repo's defining bug class, #2637).

New module: `apps/viewer/src/lib/clash/federation-identity.ts`.

## Enumeration: what advances the token

Nothing "advances" it — the identity is compared against live state, so every path that removes or replaces a gathered model's data store invalidates a run in flight, present and future:

| Path | Effect on the identity | Covered |
|---|---|---|
| `modelSlice.clearAllModels` (Clear all) | `models` emptied → key missing | yes |
| `store/index.resetViewerState` + `clearAllModels` (Open file) | same | yes |
| `modelSlice.removeModel` (Remove model, `syncSourceModel`, collab room teardown) | that key missing | yes |
| `dataSlice.setIfcDataStore` (collab peer edit, `collabSlice:~729`) | that model's store reference swapped | yes |
| `modelSlice.updateModel` / `upsertModel` with a new `ifcDataStore` | store reference swapped | yes |
| **any teardown added tomorrow** that drops or replaces a model | key missing / reference differs | yes, by construction |

Deliberately **not** invalidating, each because it cannot make a published row inert:

- a model **added** mid-run — every row still names a model that exists and still focuses, so the result stays usable. A second file finishing its load must not silently throw a run away. (This is weaker than "merely incomplete, not wrong", as an earlier revision of this description claimed: if the added file is *georeferenced*, `realignFederationModels` re-bakes every non-anchor model by mutating `geometryResult` in place and publishes through `updateModel(modelId, { preAlignment, federationAlignmentStatus })`, leaving `ifcDataStore` untouched — so the models move and the run's points and AABBs describe the pre-realign frame. Mis-placed, not inert; re-running is the remedy, and no identity keyed on the express-id space can see a pure geometry re-bake anyway.)
- a model the run **skipped** (no store, no meshes) — it contributed no elements, so no row can refer to it.
- **`geometryResult`** replacement (re-mesh) — express ids survive, so rows still resolve. Including it would discard live runs on benign geometry churn.

## Scope: `collabSlice` — covered, not deferred

The prompt flagged `collabSlice.ts:~729` calling `setIfcDataStore(payload.dataStore)` on every peer edit with no clash teardown, on the same "express ids get reassigned" argument that justifies `clearClash()` in `resetViewerState`.

This mechanism **covers it for free**, and that is a direct argument for the design: `dataSlice.setIfcDataStore` rebuilds the active model's record (`models.set(modelId, { ...model, ifcDataStore })`), so the reference differs and an in-flight run is discarded. A counter bumped at the three `endClashScenePresentation` callers would have missed this path entirely — it is precisely the "third teardown path someone adds tomorrow", except it already exists. There is a test for it.

(Note the remaining, separate question of whether a peer edit should also clear an *already-published* result is untouched here — this PR is about runs in flight. Nothing in it makes that harder to add.)

## Tests

`apps/viewer/src/hooks/useClash.stale-run-teardown.test.tsx` — the real `useClash()` hook over a real parsed model with real meshes, and the real store actions.

The teardown fires from a store subscription on `clashProgress` reaching the **narrow** phase — an event the TS kernel emits from inside `detectRule` (`ts-kernel.ts:68`, unconditional), i.e. after `gatherElements()` has read the federation and before any result exists. There is deliberately **no** `await flush()` / `tick()` / `Promise.resolve()` between the teardown and the run landing: such a flush would let the whole (fast, in-test) run complete first, the teardown would then clear the result itself, and the assertion would pass against broken code.

RED before / GREEN after, all four:

- "Clear all" mid-run
- "Open file" mid-run (`resetViewerState` + `clearAllModels`)
- `removeModel` mid-run on a two-model federation (the run examined the removed model, so its pairs naming it would be inert — note this is *narrower* than `removeModel`'s existing "keep the result" rule, which is about results published *before* the removal)
- collab peer edit mid-run (`setIfcDataStore`)

Companions, green throughout, so a guard that discarded everything cannot pass this file:

- `run()` with no teardown publishes and bumps `clashRunSeq`
- `runDuplicates()` with no teardown publishes
- a model **added** mid-run does not discard the result
- removing a model the run **never examined** does not discard the result

The duplicates path is honestly reported: its scan is synchronous, so nothing can interleave between its gather and its publish today. It goes through the same guard anyway, so adding a yield there tomorrow cannot reopen the defect on that path alone.

## Gates

- `apps/viewer` suite: **4831 tests, 4825 pass, 0 fail, 6 skipped**
- `tsc --noEmit` (apps/viewer): clean
- `check-changesets`, `check-unused-locals`, `check-test-wiring`, `check-source-text-assertions`, `check-unbounded-frame-wait`, `check-api-surface`, `typecheck-tests --audit`: all green

No assertion weakened, no fixture adjusted, no skip added, no allowlist or ratchet touched. Changeset included.

Base: branched from `upstream/main` at `a4a5ae714`; `git log --oneline upstream/main ^HEAD` is `0`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented outdated clash-detection results from reappearing after federation models are cleared, removed, reset, or replaced.
  - Ensured results are only applied when the underlying model data is still current.
  - Preserved valid results for normal scans, duplicate scans, and supported model changes.

- **Tests**
  - Added coverage for stale-result scenarios and valid clash-run workflows.


louistrue (approved):
Approving. Three independent lenses, then an adjudication that challenged all three. **The one blocker raised does not survive verification.**

## The blocker, refuted by reading the panel

A lens argued the discard is byte-identical to "0 clashes found", i.e. that a discarded run presents a false clean result. The panel says otherwise: `ClashPanel.tsx:1185` renders `No clashes found for this rule set. 🎉` only under `{result && total === 0 && …}`. With `clashResult === null` the panel takes `:1149`, the never-run call-to-action. The summary bar (`:891`) and count rows (`:956`) are equally `result`-gated.

So a discarded run reads as **"nothing happened"**, never as "your model is clean". Worth flagging that your own commit-2 message repeats the same overstatement ("no `clashError` to distinguish it from 'no clashes found'") — both are wrong about the UI, in your favour.

## Real, non-blocking

**The discard is silent.** `useClash.ts:392` and `:491` return before any user-visible signal. With a prior result on screen the old list simply stays. The honest worst case is a collab room: `collabSlice.ts:729` fires `setIfcDataStore` on every peer edit, so a run in an active room can be discarded repeatedly with no message. Still strictly better than what it replaces (publishing mis-targeted rows into a peer's renumbered id space). A one-line `setClashError(...)` on the `false` return is the obvious follow-up.

**`useClash.ts:292` (`if (built.elements.length === 0) continue;`) is new behaviour with zero coverage.** Two lenses independently deleted it and got **10/10 green**. Its apparent companion test (`removing a model the run never examined`) trips the *pre-existing* skip at `:282` instead — unfalsifiable by configuration. Production reachability confirmed: `adapters/step.ts:146-151` drops `NON_CLASHABLE_TAGS`, spatial containers and type-like entities, so a grids/annotation-only reference model yields a real store, non-empty meshes, and zero elements. If that line regresses you get false discards of correct results, not wrong publishes.

**A georeferenced model landing mid-run publishes pre-realign points** — `federationRealign.ts` re-bakes and publishes via `updateModel` without `ifcDataStore`, so the token holds. You disclosed this verbatim; recording it as the right trade with a follow-up.

## Refuted, so nobody re-litigates them

- **"Contradicts `removeModel`'s keep-the-result rule"** — `modelSlice.ts:200-203` governs *already-published* results; this guard governs *in-flight* runs. No contradiction. What is real is the neutrality: with a prior result present, the discard swaps a fresh list for a stale one containing the same inert rows. A design question, not a regression.
- **"The stale PR body ships as the squash message"** — repo config is `squash_merge_commit_message: COMMIT_MESSAGES`, so the body is your two commit messages, and commit 2 explicitly retracts commit 1's wording. The **PR page** body is genuinely stale though (the identity/`ifcDataStore` reference claim, the "four interleaved / four companions" vs the file's actual 5 + 5, "4831 tests" vs a measured 4833). Cosmetic, worth a refresh sometime.

## Mutation results, survivors stated up front

Across 19 runs: whole-fix revert **5 fail / 5 pass**; guard neutered **5/5**; `run()` early-return dropped **1 fail**; `runDuplicates()` early-return dropped **1 fail**; entity-table grain reverted to the wrapper **1 fail**; presence-only check **1 fail**; over-strict `models.size` check **1 fail**; record-before-skip **1 fail**.

The negative control matters: making `publishClashResult` always return false reds **5** — so the suite cannot pass by simply always discarding.

**Three survivors, all honest:** the `:292` line above; moving the identity capture above `await nextFrameOrTimeout` (the module's headline rationale, which no fixture can bind); and `throw` instead of `return false` on the stale path (nothing pins that the discard is silent — that survivor becomes moot if you add the `clashError`).

## Gates verified at this head

All 16 check runs carry `head_sha 0115f773e6`. Clean `merge-tree` against current main, and the commits main gained since your base touch **zero** files matching `clash|useClash|clashSlice|federation`. Changeset is `@ifc-lite/viewer: patch` on a private package — no Release hazard.

**One process note:** the hosted CodeRabbit review is against `commit_id 11ba80c049` — **commit 1, not head**. Head's green is a bare commit *status* with no review behind it, so the load-bearing second commit was never reviewed by the hosted bot. The CLI run at head returned 0 findings across all 4 files, which is what I am actually relying on.

Issue #3440, opened 2026-08-29
Every boolean op's accept path validates finiteness and index bounds only, so a torn kernel result is accepted silently


BIMvoice:
Every boolean operation validates its kernel output through `ClippingProcessor::validate_mesh` before accepting it. That function checks finiteness and index bounds and nothing else, so a torn result is accepted silently — including on paths whose own comments claim watertightness as part of the contract.

This is separate from #3435 (which is the tear itself). This is about why a tear can be accepted without anything noticing, and it should be fixed on its own terms whatever the root cause of #3435 turns out to be.

## The gap

```rust
/// Validate mesh for common issues
fn validate_mesh(&self, mesh: &Mesh) -> bool {
    if mesh.positions.iter().any(|v| !v.is_finite()) { return false; }
    if mesh.normals.iter().any(|v| !v.is_finite()) { return false; }
    let vertex_count = mesh.vertex_count();
    for idx in &mesh.indices {
        if *idx as usize >= vertex_count { return false; }
    }
    true
}
```

No topology check of any kind. It cannot observe an open edge.

**Four call sites, all in `csg/mod.rs`, none elsewhere in the repo** — so this covers every boolean op, not one path:

| line | operation |
|---|---|
| 328 | `subtract_mesh` |
| 401 | `subtract_mesh_many` (per chunk) |
| 428 | `union_mesh` |
| 452 | `intersection_mesh` |

All four share the same contract: `false` → discard the kernel result, fall back, and `record_failure(..., KernelOutputInvalid)`.

**Measured consequence:** host `#43810` in `ISSUE_068_ARK_NUS_skolebygg.ifc` passes `validate_mesh` and comes out with `open=42`. `subtract_mesh_many`'s own comment states the batch path's contract as "volume parity + **watertightness**" — that half has nothing enforcing it.

## Why the obvious fix is not the first step

Adding a topology check and returning `false` on failure is a **behaviour** change, not a warning. On failure the caller discards the kernel result and falls back, so a hard gate would silently reroute every currently-torn host onto a different code path — across difference, union and intersection at once. Whether those hosts land better, worse, or merely differently is unknown, and **it cannot be determined by reading**: the census records per-host open-edge counts at the final output but no path attribution, so nothing today can answer "how many hosts would flip".

## Suggested order

[...]

PR #3442, opened 2026-08-29
fix(geometry): record open-topology accepts through the CSG diagnostics channel (#3440 step 1)

BIMvoice:
## Gap

`validate_mesh` checks finite positions/normals and index bounds, but not closure. A finite, index-valid kernel result with a real boundary loop was therefore returned without a diagnostic.

## Change

This is #3440 step 1: a diagnostic, not a geometry gate. Accepted results from `subtract_mesh`, `subtract_mesh_many`, `union_mesh`, `union_meshes`, and `intersection_mesh` use the analytic prism-cut predicate unchanged: `directed_closed` OR `closed_or_hairline`. If both fail, the processor records the fixed `KernelError` message through the existing CSG diagnostics channel; the selected mesh is still returned unchanged.

No public `BoolFailureReason` variant is added. The public enum is exhaustive, so a diagnostic-only variant would be a breaking Rust API change. The fixed message is crate-internal; exported diagnostics carry the stable `KernelError` label.

## Final-result semantics and malformed input

`subtract_mesh_many` audits once after its final successful chunk, never discarded intermediates; rejected groups still return the host silently. `union_meshes` drives an unaudited pair helper and audits once only after a pair actually ran. A union pair can fall back to a plain merge that was not validated by the kernel path, so every returned union is validated before the direct-index closure predicate runs. A malformed empty-operand pass-through retains its legacy result and does not abort or acquire a topology record.

## Tests

Six focused regressions cover: every public boolean accept path records an actual open result without gating; closed results do not record; batched subtraction records once across 16/17 cutters; a T-junction accepted by the hairline rule does not record; `union_meshes` records once rather than per intermediate and ignores a no-pair pass-through; and malformed union-fallback indices are validated before topology inspection. `cargo test -p ifc-lite-geometry --no-fail-fast` passed 762 tests; strict geometry clippy and the module-size ratchet pass.

## Performance and determinism

Five-iteration `scripts/perf/probe.sh` A/B on `140a6d854`: FZK-Haus best total 9→9 ms (geometry 5→4 ms); CSG-heavy ISSUE_129 605→604 ms (geometry 586→586 ms). Mesh, vertex, and triangle counts were identical; ISSUE_129 diagnostics changed 1→9, the intentional observation. The verdict and lesson are recorded in `scripts/perf/README.md`.

Native exact-predicate/snapshot and mesh-manifest checks pass; wasm32 manifest passes locally and in the dispatched workflow; arm64 snapshot parity also passed in dispatched Determinism run 33387939728. Fresh current-base CI and in-tree IfcOpenShell parity are green.

## Release impact

The changeset is `minor`: callers receive unchanged geometry and no export changes, but CSG diagnostics can now report a previously invisible class of result.


louistrue (approved):
Individually reviewed current-base head a02ddcfca: diagnostic-only semantics, malformed-index guard, final-result batching, performance ledger verdict, and native/arm64/wasm determinism evidence all meet the kernel bar. Fresh CI and parity are terminal green.

PR #3451, opened 2026-08-29
test(geometry): fix #3435 T-junction classifier's collinear-merge bug

### LINKED ISSUE #3435: Booleans tear 27 watertight hosts in ISSUE_068 — three distinct mechanisms, seven remedies measured, one blocked on a re-bless call
## Current state

`ISSUE_068_ARK_NUS_skolebygg.ifc` has **29 hosts** that come out of the void boolean torn; **27 are watertight pre-void** (the boolean tears them — not inherited damage), and **2 (`#892717`, `#1265258`) are `Brep` hosts that arrive already torn** and get *less* torn after voids are applied. Those two are a separate defect and are excluded from everything below; they may deserve their own issue.

This is **at least two independent defects**, not one, and the population splits along **two independent axes**:

- **Shape of the tear**: T-junction, overlapping-void seam, degenerate sliver, or near-collinear disjoint pair.
- **Cause**: order-dependent (some void ordering is clean) vs. order-invariant (every ordering tears).

These axes are not alternative buckets — a host can be a T-junction in shape *and* order-dependent in cause (`#360795` is exactly this). An earlier version of this issue conflated them into one table; that was wrong.

`ISSUE_068` is **excluded from the census by the 50 MiB fixture filter (#3434)** — the file is 54 MB. That means the gate every candidate fix must pass cannot see the 29 hosts this issue is about. All seven remedies attempted below had to be validated against the wider census (looking for regressions) while checking the target hosts by hand with an ad hoc probe against the fixture directly — never against a pass/fail signal from the actual population. That asymmetry shaped every attempt.

### What is settled (CONFIRMED, with measurements)

- **20 of 29** hosts are T-junctions (a full-length edge collinearly tiled by shorter sub-edges, no gap/no overlap) under a classifier that clusters open edges by collinearity *and* proximity (`GAP_TOL = 5 cm`, landed in #3451). An earlier, unbounded version of the classifier reported 19 and mis-merged unrelated tears at different doorways that happened to share a supporting line.
- Of the **19** T-junction hosts measured under the older classifier: **6 are order-dependent** (some tried ordering reaches `open = 0`) and **11 are exhaustively order-invariant** — all 11 have exactly 2 voids, so declaration + reversed is the *complete* ordering space, not a sample. 2 more (`#349767`, `#892717`) found no clean ordering in 12 tried but were not exhaustively searched, so they are not claimed as either.
- **Order-dependent hosts cluster at high void counts** (4, 6, 6, 7, 7, 8 voids); every exhaustively-confirmed order-invariant host has exactly 2. This is the opposite of the original guess.
- **`#360795` is the best reproducer for the order-dependence** — 2 voids, both orderings are the entire space, declaration order tears (`open=6`), reversed is clean (`open=0`). No delta-debugging needed, unlike the original `#43810` reproducer (38 voids).
- `try_prism_cut`'s output-acceptance gate, `closed_or_hairline`, accepts a T-junction **by design** — its own doc comment describes exactly this case. The census's `edge_stats` uses an undirected ±1 balance with no collinear-coverage step, so the same mesh reads `open` there. Neither is wrong on its own terms; they simply disagree on what "closed" means, and the prism fast path is where a mesh can pass one and fail the other. This divergence is now pinned in #3438.
- **`clean_degenerate` — not the cut itself — introduces the tear** on several hosts. On `#423069`: `directed_closed(out) == true` (the cut output is genuinely closed) but `directed_closed(cleaned) == false` (the degenerate-hygiene pass tears it), while the tolerant `closed_or_hairline(cleaned)` accepts the damaged result anyway. This also explains how `#423069`'s tear can sit 0.13 m from the nearest opening, in triangles the cutter never touched — `clean_degenerate` runs globally, not just near the cut.

### Attempted remedies (seven, all measured, none shipped)

Recorded so nobody repeats them without knowing the result:

1. **Canonical cutter sort** (geometric sort of committed cutters instead of declaration order) — fixes the `#43810` target host, but **8 regressions against golden, one newly torn** (`Snowdon_Structural #2411`, `open 0→20`). The accumulation is path-sensitive, not merely order-labelled; a fixed total order just relabels which hosts land on the wrong side.
2. **Frozen cap-extension snapshot** (`extend_prism_caps`/`extend_profile_edges` read a pre-cut snapshot instead of the live accumulator) — **byte-identical no-op** against golden and against `#43810`. Proves the cap/edge-extension heuristic is innocent; the order-dependence is in the sequential `cut_prism` accumulator itself (`tris = outcome.tris`, A-then-B ≠ B-then-A).
3. **Commutative cutter union** (apply all committed cutters as one union, subtract once) — **not attempted**. `subtract_many`/`difference_all` already implement this correctly one level down, but nothing materializes a `PrismFrame` into a `Mesh`, so using it here needs a new materializer (new correctness-critical code) and would route every multi-opening host onto the exact kernel instead of the fast analytic clip.
4. **All-or-nothing closure preservation** (keep the whole pre-clean mesh whenever `clean_degenerate` would drop its closure rank) — **25 regressions** (all "gained snap-collapsed triangles": 22 in `ISSUE_129`, 3 in `rvt01.ifc`), but the **first genuine invariance gain**: `non-invariant` 137 → 123. A fix scoped to only one of `clean_degenerate`'s three call sites was tested and shown to be a provable no-op (the other two unconditionally reapply the same cleaning), so any surgical fix needs all three sites.
5. **Surgical spare-set at `directed_closed`'s 0.1 mm grid** (spare only the specific triangles that are load-bearing for closure, rather than the whole mesh) — **18 regressions**, `non-invariant` 137 → 121 (better), `collapsed` 51 → 62 (regression, but smaller than #4's). Root cause of the regression: `directed_closed` audits at 0.1 mm while the census's `edge_stats` measures at 1 mm, so a sliver can read as load-bearing under the finer grid while being degenerate at the resolution the census actually checks.
6. **Same surgical fix at the census's own 1 mm grid** — **11 regressions**, `non-invariant` 137 → 121, **`collapsed` fully recovered to golden (51 → 51)**. `EDGE_QUANTIZE_GRID_METERS` is now shared between the spare-set predicate and the census's `edge_stats` rather than duplicated.
7. **All three `clean_degenerate` call sites converted** (the previous attempt only converted two) — **identical to #6, zero further movement**: same 11 hosts regress, byte-for-byte. Confirms the residual 11 have a different root cause than the grid mismatch and are not reachable by further tuning of this line.

### The decision that is open

Attempt #6/#7 (surgical spare-set, census grid, all three sites) is a large net improvement against golden:

| metric | this attempt | golden |
|---|---|---|
| torn hosts | 157 | 204 |
| non-invariant | 121 | 137 |
| genuine defects | 44 | 46 |
| collapsed | 51 | 51 (unchanged) |

It also fixes `#423069` outright and moves `#349479` and `#360795` out of the torn set (shrinking the order-dependent group from 5 hosts to 3). But **11 hosts regress** against the census's zero-tolerance per-host gate, and no PR has ever merged against a nonzero regression count. The regressing hosts: `ISSUE_129` `#9094`, `#295370`, `#296868`, `#333923`; `ISSUE_159` `#92201`; `rvt01.ifc` `#7834`, `#11477`, `#11563`, `#11753`, `#12449`, `#25914`.

Whether those 11 golden rows warrant a re-bless (the same shape of question as #3417) is a maintainer's call, not an engineering one. Presenting the numbers neutrally — this is not an argument for an exception.

**The 11 are not one population**, so this is not a single yes/no — see the per-host breakdown [in this comment](https://github.com/LTplus-AG/ifc-lite/issues/3435#issuecomment-5462378036). Measured with a per-host spare-set counter over the committed census sweep: spared triangles are `> 0` in **all 11**, so the spare-set is confirmed as the causal channel rather than some indirect perturbation. But they split: **7 genuine new tears** (`open` and `strict` both rise); **1** (`#7834`) where `open` *falls* while `strict` rises — a non-manifold duplication the signed balance cannot see, which the looser metric would have scored as a win; **2** (`#12449`, `#25914`) where both metrics *improve* but the host flips from triangulator-convergent to divergent — an invariance regression, not a tear; and **1** (`#11477`) that was an empty, already-divergent mesh in golden and now emits a 3-triangle open fragment — a category change. `rvt01.ifc`'s 6 hosts span three of those groups, so grouping by fixture file would mislead.

### What remains open

- The order-dependent accumulation itself, for the **3 hosts** not resolved by the degenerate-hygiene fix (`#43810`, `#53374`, `#143837` — the original 5 minus `#349479`/`#360795`, which the degenerate-hygiene fix moved out) — root-caused to `try_prism_cut`'s sequential `cut_prism` accumulator, but no remedy that fixes it without regressing the census has been found (see attempts #1–#3 above).
- The **residual 11** golden-row regressions from attempt #6/#7, pending a maintainer decision on re-blessing.
- Two smaller, order-invariant groups characterized but not fixed: **overlapping-void seam** (`#144568`, `#893133` — two openings overlap in plan but differ ~1–2 cm in height, leaving a small unresolved step at the seam) and **degenerate sliver cut** (`#423069` now fixed by the degenerate-hygiene work above; `#1579408` still open — a flush-cutter-boundary case).
- **`#640479`**: an isolated near-collinear disjoint pair (two open edges ~9 µm apart on their supporting lines, endpoints millimetres apart) — below every existing tolerance, not reproduced on any other host.

### Related PRs

- #3438 — pins the `closed_or_hairline` vs. undirected-balance divergence on a T-junction as a unit test.
- #3451 — the corrected proximity-bounded T-junction classifier (fixes the collinear-merge bug that miscounted 19 vs. 20).
- #3442 — step 1 of #3440 (every boolean op's accept path validates only finiteness and index bounds, no topology check): records open-topology accepts through the CSG diagnostics channel.
- #3434 — the 50 MiB census filter that hides this fixture from the gate.
- #3436 — opt-in heavy-fixture census lane that would close that gap.
- #3440 — the accept-path topology-check issue this investigation kept intersecting with.

---

## Original report (preserved)

27 hosts in `ISSUE_068_ARK_NUS_skolebygg.ifc` are watertight before their voids are applied and torn afterwards. The boolean is tearing them, not inheriting broken input.

This is the defect; #3434 is the reason nothing sees it (the census filters fixtures over 50 MiB before opening them, and this file is 54 MB).

### Attribution

Ran the census's own machinery — `propagate_voids_to_parts`, `process_element_with_voids`, and the same 1 mm-snapped signed edge-balance `edge_stats` — with an added pre-void reading via `process_no_voids`, following `triangulation_invariance.rs` rather than a hand-rolled pipeline. Deterministic: byte-identical across two independent runs, 285 total open edges both times.

**The clearest case:**

```
IFCWALLSTANDARDCASE #43810  'Basic Wall:YV 1.1 517mm:1663810'   38 voids
  pre-void :  open=0   strict=0   tris=12
  post-void:  open=42  strict=46  tris=1362
```

Watertight going in, 42 open edges coming out.

**Across all 29 torn hosts:**

| group | count | reading |
|---|---|---|
| `SweptSolid` | 23 | `pre == 0` — CSG-caused |
| `Clipping` | 4 | `pre == 0` — CSG-caused |
| `Brep` | 2 | `pre > 0` — arrives torn, see below |

So **27 of 29 are caused by the boolean**.

The two `Brep` hosts are a separate defect and should not be conflated with this one: `#892717` goes `pre open=29 → post open=19`, and `#1265258` goes `pre open=23 → post open=8`. They arrive torn from construction/tessellation, and applying voids makes them *less* torn, not more. Worth its own issue if anyone wants to chase Brep tessellation.

### It is not simply "too many cuts"

The obvious hypothesis — that tearing tracks the number of sequential boolean cuts — does not hold. Stated as an observation, not a mechanism:

- `voids=1` hosts tear as badly as `voids=6–8` hosts (`open=8`, `open=15`)
- `#143837` has 27 voids and is among the least torn (`open=7`)
- `#43810` has 38 voids and is the worst (`open=42`)

### Narrowing on `#43810` — a combination, not one bad cut

Coarse bisection, one level, not an exhaustive delta-debug:

- **No single opening reproduces the tear.**
- One 19-opening half reproduces it *and is worse than the full set* (`open=74` vs `open=42`).
- The other 19-opening half is clean.
- Splitting the torn half again into 9 + 10 makes **both halves individually clean**.

So the tear requires a specific combination of cuts that is destroyed by splitting the set either way. That the half is worse than the whole is the interesting part — it suggests a later cut can partially repair what an earlier one broke, which argues against a simple "one cut is malformed" story.

An exact minimal reproducer needs a full delta-debug pass; this establishes the shape, not the culprit.

### Reproducing

The probe is not committed — it reuses the census machinery pointed at one fixture by env var, `#[ignore]`d so nothing changes in CI. Note the fixture is invisible to the census today for the reason in #3434, so this will not reproduce through the normal sweep until that is addressed.

Refs #3434

BIMvoice:
## What

Refs #3435. The open-edge T-junction classifier used in the #3435 torn-host investigation clustered open edges by collinearity alone, with no proximity bound. A wall produces many co-linear open edges along its whole length, so "on the same line" does not mean "part of the same tear."

On host `#144568` the unbounded classifier merged two unrelated tears at two different doorways, collinear because both sit on the same wall centreline but separated by a real **1.07 m** run of intact wall, into one phantom cluster. That fix was worked out and validated in an earlier session, but the worktree it lived in was deleted before it was committed, leaving the `GAP_TOL`-less version as the only copy anyone could find. This PR reconstructs and lands the fix so the investigation is reproducible.

## The fix

`split_by_proximity` (in `issue_3435_classify_probe.rs`) runs as a second pass after collinear clustering: it further splits a cluster wherever two edges' intervals, projected onto the shared line, are separated by more than `GAP_TOL = 5 cm`.

That threshold comes from the fixture's own measured gaps, obtained by printing `lo - running_hi` for every pair `split_by_proximity` compares across all 29 torn hosts. The gaps that stay together (the sub-edges that legitimately tile one opening's long edge) are 0 or negative; the largest is 0.000000 m to six decimals, f32 mesh-export noise. The gaps that do split are 0.1235 m (`#360795`), 0.2110 / 0.2639 / 0.6034 m (`#43810`), 1.0721 m (`#144568`), 3.5200 m (`#893133`) and 5.0760 m (`#53374`).

So 5 cm clears the noise floor by two orders of magnitude, but the margin on the other side is a factor of 2.5, not another two orders: four of those seven separations are under a metre. Raising `GAP_TOL` past 0.1235 m would re-merge `#360795` and silently undo the fix. The headroom is below this value, not above it.

## Anchors (asserted, not just printed)

- `#628727` and `#360795` classify as T-junctions
- `#144568` and `#893133` do NOT classify as T-junctions

`#360795` is the anchor that pins the fix itself: over the fixture's 29 torn hosts it is the only host whose verdict the second pass changes. Reverting only the `flat_map(split_by_proximity)` block and keeping every assertion fails on `#360795` alone; the other three anchors are unchanged by that mutation and could not have caught it.

Also added: a permutation-invariance check per anchor host. What it guards is the clustering's anchor tie-break, not hash order. `Iterator::max_by` returns the LAST maximum on ties, so equal-length edges are resolved by input position, and host `#144568` has two open edges tying exactly at 1.6199951183372845 m; permuting its edge list really does swap which of them anchors the first cluster. (The edge list order itself is stable: `FxBuildHasher` is a seedless unit struct, so `FxHashMap` iteration order does not vary run to run.)

With the fix, the T-junction count over the fixture's 29 torn hosts moves from **19 to 20**: host `#360795` was a second, smaller instance of the same phantom-merge failure, and correctly resolves into two independently-covered clusters once the proximity split is applied.

## Consolidation

Several overlapping ad hoc `#3435` probes accumulated across sessions investigating this issue, in worktrees that were never committed. This PR lands two of them: the classifier above, and `issue_068_void_ordering_probe.rs` (the void-application-order permutation check). Both are new files. Nothing is deleted: `git diff --name-status origin/main...HEAD` is two `A` lines and zero `D` lines.

## Not a product fix

This is diagnostic tooling for the #3435 investigation, not a fix to any product code path. Both probes stay `#[ignore]`d and are not gated in CI. Precedent: `wall_opening_cut_regression.rs` and the heavy-fixture lane added in #3436 already commit `#[ignore]`d expensive/manual probes.

## Test plan

- [x] `cargo test -p ifc-lite-geometry --features debug_geometry --test issue_3435_classify_probe -- --ignored --nocapture classify_probe` — passes, anchors hold, permutation checks pass
- [x] `cargo build -p ifc-lite-geometry --test issue_068_void_ordering_probe --features debug_geometry` — compiles
- [x] `cargo clippy -p ifc-lite-geometry --all-targets -- -D warnings` (without `debug_geometry`, CI's scope) — 0 warnings
- [x] Both probes skip cleanly with the fixture absent
- [x] `node scripts/check-module-size.mjs` — OK
- [x] `node scripts/check-rust-source-text-assertions.mjs` — OK (0 source-text assertions)

### DIFF
--- a/rust/geometry/tests/issue_068_void_ordering_probe.rs
+++ b/rust/geometry/tests/issue_068_void_ordering_probe.rs
@@ -1,0 +1,387 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Ad hoc probe for issue #3435: for EVERY torn host in ISSUE_068
+//! (`ISSUE_068_ARK_NUS_skolebygg.ifc`), apply its recorded
+//! `IfcRelVoidsElement` openings in several different orders and record the
+//! resulting `open`-edge count, to answer one question: does a single
+//! ordering rule make the tears go away, and for how many of the 29 hosts?
+//!
+//! Reuses the same machinery as `issue_053_heavy_csg_probe.rs`
+//! (`void_index`, `process`, `process_no_voids`, `edge_stats`), duplicated
+//! here rather than shared across two separate test-binary crates.
+//!
+//! Run:
+//!   cargo test -p ifc-lite-geometry --test issue_068_void_ordering_probe \
+//!     -- --ignored --nocapture void_ordering_probe
+//!
+//! `--ignored`: manual probe, not gated CI. Debug build (no `--release`) —
+//! this worktree's shared `.cargo/config.toml` collides with
+//! `panic = "abort"` under `--release`.
+
+mod census_golden;
+
+use ifc_lite_core::{build_entity_index, EntityDecoder, EntityScanner};
+use ifc_lite_geometry::{propagate_voids_to_parts, GeometryRouter, Mesh};
+use rustc_hash::FxHashMap;
+
+fn crate_dir() -> std::path::PathBuf {
+    std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+}
+
+fn repo_root() -> std::path::PathBuf {
+    crate_dir().join("..").join("..")
+}
+
+fn void_index(content: &str) -> FxHashMap<u32, Vec<u32>> {
+    let mut idx: FxHashMap<u32, Vec<u32>> = FxHashMap::default();
+    let mut scanner = EntityScanner::new(content);
+    let mut decoder = EntityDecoder::new(content);
+    while let Some((id, name, start, end)) = scanner.next_entity() {
+        if name == "IFCRELVOIDSELEMENT" {
+            if let Ok(entity) = decoder.decode_at_with_id(id, start, end) {
+                if let (Some(host), Some(opening)) = (entity.get_ref(4), entity.get_ref(5)) {
+                    idx.entry(host).or_default().push(opening);
+                }
+            }
+        }
+    }
+    let _ = propagate_voids_to_parts(&mut idx, content, &mut decoder);
+    idx
+}
+
+fn process(content: &str, host_id: u32, voids: &FxHashMap<u32, Vec<u32>>) -> Option<Mesh> {
+    let ei = build_entity_index(content);
+    let mut decoder = EntityDecoder::with_index(content, ei);
+    let entity = decoder.decode_by_id(host_id).ok()?;
+    let router = GeometryRouter::with_units(content, &mut decoder);
+    router.process_element_with_voids(&entity, &mut decoder, voids).ok()
+}
+
+fn process_no_voids(content: &str, host_id: u32) -> Option<Mesh> {
+    let ei = build_entity_index(content);
+    let mut decoder = EntityDecoder::with_index(content, ei);
+    let entity = decoder.decode_by_id(host_id).ok()?;
+    let router = GeometryRouter::with_units(content, &mut decoder);
+    router.process_element(&entity, &mut decoder).ok()
+}
+
+struct EdgeStats {
+    open: usize,
+}
+
+fn edge_stats(mesh: &Mesh) -> EdgeStats {
+    let q = |v: f32| (v as f64 * 1.0e3).round() as i64;
+    let mut vid: FxHashMap<(i64, i64, i64), u32> = FxHashMap::default();
+    let mut id = |i: usize| -> u32 {
+        let k = (
+            q(mesh.positions[i * 3]),
+            q(mesh.positions[i * 3 + 1]),
+            q(mesh.positions[i * 3 + 2]),
+        );
+        let n = vid.len() as u32;
+        *vid.entry(k).or_insert(n)
+    };
+    let mut uses: FxHashMap<(u32, u32), (u32, u32)> = FxHashMap::default();
+    for tri in mesh.indices.chunks_exact(3) {
+        let (a, b, c) = (id(tri[0] as usize), id(tri[1] as usize), id(tri[2] as usize));
+        if a == b || b == c || c == a {
+            continue;
+        }
+        for (x, y) in [(a, b), (b, c), (c, a)] {
+            let e = uses.entry((x.min(y), x.max(y))).or_insert((0, 0));
+            if x < y {
+                e.0 += 1;
+            } else {
+                e.1 += 1;
+            }
+        }
+    }
+    EdgeStats { open: uses.values().filter(|&&(f, r)| f != r).count() }
+}
+
+/// Signed-tetrahedron-sum volume (divergence theorem), magnitude only. Not
+/// gated on `is_trustworthy_solid()` like `geom_closure::volume()` — this is
+/// a ranking key for an ordering heuristic, not a certified measurement, and
+/// every opening here is a simple swept/extruded solid that processes fine
+/// standalone.
+fn mesh_volume(mesh: &Mesh) -> f64 {
+    let mut sum = 0.0f64;
+    for tri in mesh.indices.chunks_exact(3) {
+        let p = |i: u32| {
+            let i = i as usize;
+            [
+                mesh.positions[i * 3] as f64,
+                mesh.positions[i * 3 + 1] as f64,
+                mesh.positions[i * 3 + 2] as f64,
+            ]
+        };
+        let [ax, ay, az] = p(tri[0]);
+        let [bx, by, bz] = p(tri[1]);
+        let [cx, cy, cz] = p(tri[2]);
+        // 6x signed volume of the tetrahedron (origin, a, b, c)
+        sum += ax * (by * cz - bz * cy) - ay * (bx * cz - bz * cx) + az * (bx * cy - by * cx);
+    }
+    (sum / 6.0).abs()
+}
+
+fn mesh_centroid(mesh: &Mesh) -> [f64; 3] {
+    let (mn, mx) = mesh.bounds();
+    [
+        (mn.x as f64 + mx.x as f64) / 2.0,
+        (mn.y as f64 + mx.y as f64) / 2.0,
+        (mn.z as f64 + mx.z as f64) / 2.0,
+    ]
+}
+
+fn mesh_aabb_min(mesh: &Mesh) -> [f64; 3] {
+    let (mn, _) = mesh.bounds();
+    [mn.x as f64, mn.y as f64, mn.z as f64]
+}
+
+/// splitmix64: tiny deterministic PRNG so permutations are seed-reproducible
+/// without pulling in the `rand` crate as a dev-dependency of a
+/// `#[ignore]`d probe.
+struct SplitMix64(u64);
+impl SplitMix64 {
+    fn new(seed: u64) -> Self {
+        Self(seed)
+    }
+    fn next(&mut self) -> u64 {
+        self.0 = self.0.wrapping_add(0x9E3779B97F4A7C15);
+        let mut z = self.0;
+        z = (z ^ (z >> 30)).wrapping_mul(0xBF58476D1CE4E5B9);
+        z = (z ^ (z >> 27)).wrapping_mul(0x94D049BB133111EB);
+        z ^ (z >> 31)
+    }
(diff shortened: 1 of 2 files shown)


louistrue (approved):

PR #3466, opened 2026-08-29
fix(drawing-2d): stop hatch fills leaking outside a polygon boundary at a tangent vertex

BIMvoice:
## Summary

- `HatchGenerator.clipLineToRing` computed a literal line-edge intersection per ring edge, then de-duplicated near-equal `t` values down to one representative regardless of cluster size. A ring vertex sitting exactly on the hatch line is shared by two edges, so a tangent "spike" (both neighbouring edges on the same side, only touching the hatch line) always produced two very-close intersections that collapsed to one — an odd crossing count where the true answer is zero. That flipped the inside/outside sense of every segment built past the vertex, including the final "run off the far end of the line" segment: a section fill's hatch pattern could extend dozens of units outside the shape it was clipped to.
- Fixed by classifying each ring vertex by which side of the (infinite) hatch line it falls on, using the same strict tie-break `pointInRing` already uses. An edge is only a genuine crossing when its two endpoints land on opposite sides — this correctly resolves both a tangent spike (net 0 crossings) and a genuine pass-through vertex (net 1 crossing) without any duplicate-`t` heuristic.
- Also widened `lineLineIntersection`'s edge-parameter (`u`) bounds check by a small epsilon: a crossing that should land exactly at `u=0`/`u=1` routinely computes a few ULPs past that boundary once the query line has any tilt, and dropping it there would silently defeat the side check.
- Replaced the zero-crossing fallback's midpoint sample with `line.start` (always outside the ring — the caller extends every hatch line well past the polygon's extent), so the fallback can no longer coincide with the very vertex that made the crossing count zero and give a leaking answer.

## Reproduction

A rectangle `[0,10]x[0,10]` with a downward spike notch cut into the top edge, touching `y=5` at a single vertex `(5,5)` without crossing below it. Before the fix, the hatch row at `y=5` produced a segment from `(10,5)` to `(-21.2,5)` — running ~21 units outside the polygon. New tests in `packages/drawing-2d/src/hatch-generator.test.ts` pin this (tangent spike), a genuine pass-through vertex (rhombus case, which the same bug also mishandled the other way — dropping a real crossing), a hatch row flush with a whole boundary edge, and a hole edge flush with the hatch line.

## Test plan

- [x] New RED tests in `hatch-generator.test.ts` fail against `main` (segment escapes the polygon bbox), pass after the fix.
- [x] Full `packages/drawing-2d` suite: 523/523 passing (`pnpm --filter @ifc-lite/drawing-2d test`, run directly via vitest in this environment).
- [x] `node scripts/check-module-size.mjs` — OK, `hatch-generator.ts` at 428/431 lines.
- [x] `tsc --noEmit` on `packages/drawing-2d` — clean.
- [x] Grepped for sibling instances of the same "collect raw intersections + dedupe by `t`" pattern elsewhere in `drawing-2d`/`renderer` — none found; this was a one-off in `hatch-generator.ts`.


louistrue (approved):

PR #3473, opened 2026-08-29
fix(clash): match clash-revision lifecycle by durable review key, not the model-scoped id

BIMvoice:
## Summary

`compareClashRuns` (`packages/clash/src/lifecycle.ts`) matched clashes between two runs by the raw `clash.id`. `engine-ts/orchestrator.ts`'s `clashId()` folds `ClashElement.model` into that id (`${model} ${key}`), and `review.ts` documents `model` as an ephemeral per-load id assigned by the host app — exactly why `review.ts`'s own durable key, `clashReviewKey`, deliberately excludes it.

Two loads of identical geometry — precisely the "model revision" scenario `compareClashRuns` exists to diff — therefore produce two different `clash.id`s for the same real-world clash. A clash that is still open on the next run was reported as `resolved` (from the previous run) *and* `added` (in the next run) instead of `persistent`, burying any genuinely new or resolved clash in spurious churn. This affects runs from separate loads, which re-mint `model`; two runs inside one loaded session matched correctly before this change and still do.

The seam: `engine-ts/orchestrator.ts` (id production, covered by `orchestrator.test.ts`) and `lifecycle.ts` (id consumption, covered by `lifecycle.test.ts`) are each well tested, but `lifecycle.test.ts` only ever fed `compareClashRuns` hand-typed literal ids (`makeClash('c1')`, …), never an id the real engine actually computed — so the mismatch between what `clashId()` produces and what `compareClashRuns` assumes about it was invisible to either suite.

## Fix

`compareClashRuns` now matches by `clashReviewKey` (rule id + the two elements' durable keys, order-independent) instead of `clash.id`. Output shape and the `persistent` bucket's "report the next run's Clash" behaviour are unchanged.

Two later commits went further, because a review key is not unique: a key can carry several occurrences. Occurrences are now grouped per review key and paired within the group, and leftovers are split by whether their models still clash in the other run rather than being paired positionally and all called `persistent`.

## Evidence

Added a test in `lifecycle.test.ts` that runs the *real* `runClash` orchestrator twice (same durable element keys, different `model`, simulating two loads) and diffs the two real `ClashResult`s with `compareClashRuns`:

- RED (before the fix): `persistent` was empty; the clash showed up as both `resolved` and `added`.
- GREEN (after the fix): the clash is reported as `persistent`, with `resolved`/`added` empty.

Full `@ifc-lite/clash` suite: 456 passed, 1 skipped (pre-existing skip, unrelated).

## Siblings checked

Grepped `packages/clash/src` for other cross-run matching on raw `clash.id` outside `lifecycle.ts` — none found (`grouping.ts`/`contact/bvh.ts` hits are unrelated `.id` fields). No other in-repo consumer of `compareClashRuns` or `clashReviewKey`.

## Module size / changeset

`node scripts/check-module-size.mjs` passes (0 files over budget). Changeset added (`@ifc-lite/clash`, **major**). The package is 1.9.2 and this changes which bucket existing inputs land in, so per AGENTS.md:68's first clause the level follows the biggest API change, which for a >=1.0 package is major.


louistrue (approved):

PR #3515, opened 2026-08-30
fix(mcp): materials_list/count_entities drop IfcMaterialList names into "(no material)"

BIMvoice:
## Fix

`count_entities({ group_by: 'material' })` and `materials_list` now use the existing shared `materialFallbackName` helper instead of a duplicate MCP-local implementation. This keeps both tools on the same material naming policy, including a set-level material name when present and member-name fallback when it is absent.

## Regression coverage

The renamed helper tests cover plain materials, material lists, layer/profile/constituent sets, set-level-name precedence, and a named `IfcMaterialProfileSet` whose members are unnamed.

## Validation

Fresh root CI is required after rebase.


louistrue (dismissed):

PR #3523, opened 2026-08-30
fix(viewer): resolve real IFC properties for drawing-export graphic overrides

BIMvoice:
## Summary
- #3520 removed `ElementData.materials`/`.layers` (never populated by any construction site) but explicitly scoped out the sibling `properties` field. This PR resolves that gap.
- `ElementData.properties` (`packages/drawing-2d/src/graphic-overrides/types.ts`) is declared and read by the rule engine's `property`/`propertySet` criteria (`rule-engine.ts`), but every `ElementData` construction site in `apps/viewer/src/hooks/useDrawingExport.ts` built only `{ expressId, ifcType }` — the same shape of gap #3520 found for `materials`/`layers`.
- Unlike `materials`/`layers`, this one is directly user-reachable: the built-in **"Fire Safety"** preset (selectable from the drawing settings panel's Style Presets list, `packages/drawing-2d/src/graphic-overrides/presets.ts`) colors walls/doors entirely by their `FireRating` property, and **"Structural Highlight"** has a `LoadBearing`-gated rule. Both silently painted every element with their non-matching base style, no matter how it was rated — no error, no warning.
- Unlike `materials`/`layers`, the data genuinely is in reach here: `useDrawingExport.ts` already receives `ifcDataStore` (#3520 observed this but did not pursue it), and its SVG generators (`generateExportSVG`, `generateSheetSVG`) run once per export click, not per animation frame — so a per-entity `extractPropertiesOnDemand` call is cheap enough.
- Chose **implement** over **remove**: the criterion is user-reachable (a built-in preset uses it) and the data is reachable at these construction sites, unlike the #3520 case.
- New `apps/viewer/src/hooks/drawingElementProperties.ts` (`makePropertiesGetter`) resolves each polygon's `entityId` — a renderer/global id, offset-adjusted for federation — back to its owning model's store via `fromGlobalIdFromModels` (the same resolution `bcfIdLookup.ts` uses), then calls `extractPropertiesOnDemand`, cached per export pass so each entity is extracted at most once across both the fill and outline render loops.
- **Not fixed, and not a regression**: the live on-screen 2D canvas (`Drawing2DCanvas.tsx`) is unchanged. Like #3520 found for `materials`, it has no data-store reference at all in its hot per-frame render loop — threading one through, with a caching layer, is a larger and separate change. So Fire Safety/Structural now render correctly in exported SVG, sheet SVG, and sheet PDF (which rasterizes the sheet SVG), and in Print — but not yet in the live preview.

## Evidence
- RED (by execution, before the fix — confirmed by stashing the fix): built `ElementData` exactly the way every production call site did (`{ expressId, ifcType }`), ran the override engine with a `FireRating >= 120` property criterion (mirroring the Fire Safety preset's own rule shape) against a real parsed `IfcWall` with `Pset_WallCommon.FireRating = 120` — the higher-priority fire rule never matched; the export fell back to the lower-priority base rule's color.
- Control: the same setup with only an `ifcType` criterion always matched, isolating the assertion to the property path.
- GREEN (after the fix): the same FireRating criterion now matches the real parsed model's property data, and wins over the base rule as its higher priority dictates.

## Test plan
- [x] `apps/viewer/src/hooks/useDrawingExport.propertyOverride.test.tsx` — new file, 2 tests (control + property match), both passing; parses a real minimal IFC4 fixture (`IfcWall` + `Pset_WallCommon.FireRating`) via `IfcParser`, runs the real `handleExportSVG`, and inspects the actual SVG output the way `useDrawingExport.pdfVectorPaths.test.tsx` inspects PDF output
- [x] `useDrawingExport.pdfSheet.test.tsx`, `useDrawingExport.pdfSheetRasterCap.test.tsx`, `useDrawingExport.pdfVectorPaths.test.tsx` — all still passing (23 tests total across the 4 files)
- [x] `packages/drawing-2d/src/graphic-overrides/` vitest suite (unmodified, sanity-checked) — 63 tests passing
- [x] `pnpm exec tsc --noEmit` in `apps/viewer` (includes test files) — clean
- [x] `node scripts/check-module-size.mjs` — OK, `useDrawingExport.ts` at 1350 lines against its 1359-line budget (9 lines headroom, achieved by extracting the new logic to `drawingElementProperties.ts` and compacting the 4 `ElementData` literals to single-line form — the same file-splitting idiom `dxfExportGeoref.ts`/`svgExportViewport.ts` already use in this hook)
- [x] Changeset added: `@ifc-lite/viewer` `patch`

## Summary by CodeRabbit

- **Bug Fixes**
  - SVG and PDF drawing exports now correctly apply graphic overrides based on IFC element properties and property sets.
  - Preset rules for load-bearing elements, fire doors, escape routes, and custom property criteria now match exported graphics as expected.
  - Live 2D canvas rendering and DXF exports are unchanged.
  - Fire-rated wall rules using numeric comparisons with text-based `FireRating` values remain unsupported.


louistrue (approved):

PR #3529, opened 2026-08-30
fix(ifcx): stop dropping bsi::ifc::material on import

BIMvoice:
## Summary
- `property-extractor.ts`'s `SKIP_ATTRIBUTES` treated `bsi::ifc::material` (`{code, uri}`) the same as graph-structural attributes (`bsi::ifc::class`, `usd::usdgeom::mesh`, `usd::xformop`) — skipped before it can ever become a property or a relationship.
- `bsi::ifc::material` is the *only* channel the IFCX v5a schema carries an element's material on. buildingSMART's own PCERT sample scenes (`tests/models/ifc5/PCERT-Sample-Scene_*.ifcx`) author it on most physical elements — e.g. `{code: "concrete_reinforced_in_situ", uri: "https://identifier.buildingsmart.org/..."}`. Measured by running `parseIfcx` over all thirteen PCERT scenes: walls 4/4 in each of the three scenes that have them, beams 6/6 in `Building-Structural`, columns 7/7 in `Infra-Bridge`, pipe segments 21/24 in `Infra-Plumbing`, track elements 64/66 in `Infra-Rail`. Roofs are the exception — only `Building-Architecture` and `Building-Structural` contain an `IfcRoof` at all, and neither roof carries the attribute.
- Nothing else in `packages/ifcx` reads that key either, so on import the material information vanished entirely: no property, no relationship, no error. A STEP-sourced model surfaces the equivalent data via `IfcRelAssociatesMaterial` — the query engine exposes it as a named relation, and the viewer's Material tab (`MaterialTotalsPanel.tsx`) is built on it.

## Fix
`bsi::ifc::material` is unpacked into its own `Material` property set (`Material` = the material code, `Uri` = its buildingSMART identifier) instead of being silently skipped. `formatNamespace` already had a `'bsi::ifc::material': 'Material'` mapping entry for exactly this pset name. That entry was not unreachable: `apps/viewer/src/lib/collab/step-seed.ts:140` authors `bsi::ifc::material::Name` (and `::Layers` at :146) when seeding a collab room from a STEP model, and those keys split into namespace `bsi::ifc::material`, which hits the entry — verified by running `extractProperties` on the pre-fix source. What could never reach it was the leaf key `bsi::ifc::material` itself: the generic `namespace::name` split turns it into namespace `bsi::ifc` / name `material`, landing it in the generic "IFC" bucket, and before this fix `SKIP_ATTRIBUTES` dropped it before that split ever ran. Reusing the entry means an IFCX-native material and a STEP-seeded one land in the same `Material` pset.

I stopped short of building a full `IfcRelAssociatesMaterial`-equivalent relationship (synthetic `IfcMaterial` entities in the `RelationshipGraph`, matching the STEP `MaterialTotalsPanel`/`buildMaterialUsageIndex` contract) — that machinery assumes STEP-specific entity shapes and is a materially larger change. This fix gets IFCX users from *zero* material information to the same code/uri the file actually carries, visible in the Properties panel like any other property.

## Context — spec-vs-reader census
Per the "check the ifcx reader against the SPEC" pass (following up on #3524, #3522), enumerated every attribute defined in `packages/export/src/__fixtures__/schemas/*.ifcx` and every attribute key read in `packages/ifcx/src/*.ts`:

| spec attribute (schema fixture) | reader | verdict |
| --- | --- | --- |
| `bsi::ifc::class` (code, uri) | `entity-extractor.ts`, `hierarchy-builder.ts`, `index.ts` | read |
| `bsi::ifc::presentation::diffuseColor` | `geometry-extractor.ts` (via `ATTR.DIFFUSE_COLOR`) | read |
| `bsi::ifc::presentation::opacity` | `geometry-extractor.ts` (via `ATTR.OPACITY`) | read |
| `bsi::ifc::material` (code, uri) | **NONE before this fix** — `SKIP_ATTRIBUTES` dropped it | **fixed here** |
| `bsi::ifc::spaceBoundary` (relatedelement.ref, relatingspace.ref) | `hierarchy-builder.ts`'s `isSpaceBoundaryRelationshipClass`/`collectSpaceBoundaryElementIds` fold it into `IfcSpace.elements`; falls through generic property extractor too (JSON-stringified blob) — not a structured `RelationshipType.SpaceBoundary` edge like STEP's `IfcRelSpaceBoundary` | partially read; closed as lower-confidence than the material gap (space-boundary elements do surface, just not as a queryable relationship edge) |
| `bsi::ifc::alignment` + 3 alignment-segment schemas | none | not read — civil/alignment geometry is a materially larger, separate feature; closed |
| `bsi::ifc::system::connectsto` | none | not read — no consumer of `IfcRelConnectsPorts`-equivalent found downstream either; closed, low confidence of user impact |
| `bsi::ifc::system::flowdirection` | none | not read; closed, no downstream consumer found |
| `bsi::ifc::system::partofsystem` | `index.ts`'s `buildRelationships` (`ATTR.PART_OF_SYSTEM`) | read |
| `bsi::ifc::system::servicesfacility` | none | not read; closed, no downstream consumer found |
| `bsi::ifc::prop::Name` | `entity-extractor.ts`, `hierarchy-builder.ts` | read |
| `bsi::ifc::prop::Description` | `entity-extractor.ts` | read (fixed in #3522) |
| `bsi::ifc::prop::UsageType` | none (falls through generic property extractor as a plain property) | acceptable — generic property path is the correct channel for this |
| `bsi::ifc::prop::RefElevation` / `ElevationOfRefHeight` / `ElevationOfTerrain` / `NumberOfStoreys` | generic property extractor (`quantity-classification.test.ts` pins these staying properties, not quantities) | read |
| `bsi::ifc::prop::Volume`/`NetVolume`/`Height`/`Width`/`Length`/`Depth`/`NetArea`/`NetSideArea`/`CrossSectionArea` | quantity table via `routesToQuantityTable`/`isQuantityProperty` | read |
| `bsi::ifc::prop::Station` | generic property extractor | read |
| `bsi::ifc::prop::TypeName` | `entity-extractor.ts`, `hierarchy-builder.ts` | read |
| `usd::usdgeom::mesh` (faceVertexIndices, points) | `geometry-extractor.ts` | read |
| `usd::usdgeom::visibility` | `geometry-extractor.ts` | read |
| `usd::xformop` (transform) | `geometry-extractor.ts`, `pointcloud-extractor.ts` | read |
| `usd::usdgeom::basiscurves` (points) | none found | not read — closed, low confidence (no curve-rendering consumer identified in this pass) |

Also checked the reverse direction (reader fields with no spec source, i.e. fabrication): `entity-extractor.ts`'s `builder.add(...)` still passes `typeCode` as `objectType` (line reads `typeCode, // objectType`) — that's the fabrication PR #3524 (open) already found and is fixing; not re-reported here.

`bsi::ifc::prop::Elevation` (read by `hierarchy-builder.ts` for `IfcBuildingStorey.elevation`) does **not** appear in the v5a `prop@v5a.ifcx` schema at all, and never appears on an `IfcBuildingStorey` node in the real PCERT fixtures (`RefElevation` there is authored on `IfcSite`, not storeys — a different, legitimate attribute). I could not find a spec-defined channel for storey elevation in IFCX v5a at all (it may be intentionally derivable only from the composed `usd::xformop` transform's Z translation, the way STEP falls back to `ObjectPlacement.Location.Coordinates[2]` when `Elevation` is null per #1857) — but I have no multi-storey real-world IFCX fixture to verify that hypothesis against, so I'm flagging it here rather than shipping an unverified transform-based fix.

## Test plan
- [x] New test in `packages/ifcx/src/property-extractor.test.ts` ("does not silently drop bsi::ifc::material"): builds a node with `bsi::ifc::material` = `{code: 'concrete_reinforced_in_situ', uri: '...'}`, asserts a `Material` property with that code reaches `extractProperties`'s output.
- [x] Confirmed RED before the fix: `actual: []` (no properties at all).
- [x] Confirmed GREEN after the fix; full `packages/ifcx` suite (173 tests / 47 suites) passes.
- [x] `pnpm exec tsc --noEmit` clean for `packages/ifcx`.
- [x] `node ../../scripts/typecheck-tests.mjs` from `packages/ifcx` — OK (20 test files).
- [x] `node scripts/check-module-size.mjs` from repo root — OK, no new budget violations.
- [x] `node scripts/check-test-wiring.mjs` — OK.
- [x] Changeset added (`patch`, `@ifc-lite/ifcx`) — surfaces previously-dropped data on a new property set name, no exported TS shape change (didn't run `api-surface:update`/full build per the resource note; nothing here changes an exported symbol).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed IFCX material attributes being omitted during property extraction.
  - Material codes now appear as **Material** properties, with buildingSMART identifiers available as **Uri** values.
  - Preserves material information when reading IFCX archives, improving the completeness of extracted element properties.

- **Tests**
  - Added coverage to verify material information is preserved and displayed correctly.


louistrue (approved):

PR #3532, opened 2026-08-30
fix(mcp,viewer): viewer_get_selection and LLM context drop unnamed material-set names

BIMvoice:
## Summary

Two remaining instances of the `IfcMaterialList`/unnamed-set naming gap fixed for the MCP `materials` resource in #3519 (see that PR's audit table).

**`packages/mcp/src/tools/viewer.ts` (`viewer_get_selection` text summary).** The formatting chain checked `.layers[]` then `.materials[]` (the `IfcMaterialList` case) before falling to a single `mat.name ?? mat.materialName` check. A `MaterialProfileSet`/`MaterialConstituentSet` with no set-level `Name` has neither of those, so the whole `Materials:` line silently disappeared — worse than the `IfcMaterialList` case, which at least listed member names. `viewer.ts` sits at its 696-line module-size budget, so the formatting was extracted to a new sibling module (`packages/mcp/src/tools/material-summary.ts`) rather than grown in place; it now lists `.layers[]`/`.profiles[]`/`.constituents[]`/`.materials[]` by member name, and falls back to the set-level `Name` when none of those members is named.

That last clause is the opposite-direction case. `IfcMaterialProfile.Material` is OPTIONAL in IFC4, and `material-resolver.ts` leaves both `materialName` and `name` undefined on a member whose `Material` is absent or fails to resolve, so a *named* set whose members name nothing is ordinary valid IFC. Consulting the member arrays ahead of the set name would have printed `  Materials: ?` for it. All four arrays therefore go through one `memberLine()` that reports nothing when it would name nothing — the same rule `lensMaterialNames` (`apps/viewer/src/lib/lens-material-names.ts`) already applies. This also repairs two cases that were already lossy before this PR: a named `MaterialLayerSet` or `IfcMaterialList` with unnamed members used to print `  Materials: ?` too, because those two arrays were checked ahead of the name.

**`apps/viewer/src/lib/llm/context-builder.ts`.** `materialName` was computed as `rawMaterial?.name ?? rawMaterial?.materials?.[0]?.name` — no leg for `.layers[]`/`.profiles[]`/`.constituents[]` at all, so an unnamed `MaterialLayerSet`/`ProfileSet`/`ConstituentSet` (common — those sets are frequently authored unnamed, with the name only on the layer/profile/constituent) reported no material at all in the LLM's system-prompt context. Extracted the fallback chain to a new sibling module (`apps/viewer/src/lib/llm/material-name.ts`).

**Shared helper.** Both MCP call sites now go through a lifted, exported `materialFallbackName()` (`packages/mcp/src/material-naming.ts`, moved out of `providers.ts`'s local copy from #3519, whose doc comment noted it should move once shared). In `viewer_get_selection` it is the last-resort leg; the member listing above it lives in `material-summary.ts`. #3515 (open) proposes a `materialDisplayName()` helper in `packages/mcp/src/tools/util.ts` that would supersede this once merged — not merged yet, so this PR reuses/lifts the existing local helper rather than depending on an unmerged branch, per that PR's own note.

## Test plan
- [x] `packages/mcp/src/tools/material-summary.test.ts`: RED confirmed against the exact pre-fix inline logic — reverting `formatMaterialsBlock` to `viewer.ts`'s old chain gives `Tests 3 failed | 7 passed (10)`, exit 1 (`expected undefined to be '  Materials: Steel'`, `expected undefined to be '  Materials: Insulation, Gypsum'`, and the named-layer-set case). GREEN after the extraction.
- [x] Opposite direction also RED-checked: reverting only the all-members-unnamed guard, tests untouched, gives `Tests 4 failed | 6 passed (10)`, exit 1 (`expected '  Materials: ?' to be '  Material: Steel Columns'` and the constituent, layer and unnamed-set cases). A control test pins that a set with *some* named member still lists members, placeholders and all (`  Materials: Brick, ?`).
- [x] `apps/viewer/src/lib/llm/material-name.test.ts`: RED confirmed against the exact pre-fix expression (unnamed `MaterialLayerSet`/`ProfileSet`/`ConstituentSet` all produced `undefined`); GREEN after extracting `materialDisplayName()`.
- [x] `pnpm --filter @ifc-lite/mcp exec tsc --noEmit` — clean.
- [x] `pnpm --filter @ifc-lite/mcp exec vitest run` — 347/347 passing across 35 files.
- [x] `pnpm exec tsc --noEmit` in `apps/viewer` — clean.
- [x] `apps/viewer` test runner: `src/lib/llm/material-name.test.ts` 6/6 passing, `src/lib/llm/context-builder.test.ts` 1/1 passing.
- [x] `node scripts/check-module-size.mjs` — clean (`viewer.ts` moved from 696 to 686 lines by extracting the formatting out, `providers.ts` from 327 to 303; the new `material-summary.ts` is 57 lines). The allowlist is untouched.
- [x] `node scripts/check-unused-locals.mjs` — clean (51 packages, 943 known, none increased).
- [x] `node scripts/check-api-surface.mjs` — clean (42 packages, 65 export surfaces, 4266 exports), so neither new module changes the published surface and `patch` is the right bump for both changesets.


louistrue (approved):

Issue #311, opened 2026-03-05
Feature request: IFC 2x3 parsing support


ocdnamib:
While IFC 4 adoption is growing, a significant portion of the industry still relies on IFC 2x3 exports. This is particularly common in the following scenarios:

Many widely-used BIM authoring tools (especially older versions of Revit, ArchiCAD, and Allplan) default to IFC 2x3 or offer more stable/reliable exports in this schema version.
Large organizations and public sector clients often mandate IFC 2x3 in their BIM execution plans, as their existing toolchains and validation workflows are built around it.
Ongoing projects that started years ago continue to produce IFC 2x3 deliverables throughout their lifecycle, and migrating mid-project is rarely practical.
In some regions, regulatory or certification frameworks still reference IFC 2x3 as the baseline interoperability format.

From a practical standpoint, the structural differences between IFC 2x3 and IFC 4 are well-documented (entity renaming, property set changes, geometry representation updates), and several open-source libraries (IfcOpenShell, xBIM) already handle both schemas transparently, which suggests that adding 2x3 parsing support should be feasible without a major architectural overhaul.
Would the maintainers be open to considering IFC 2x3 compatibility, even as a partial or best-effort implementation? This would significantly broaden the tool's usability across real-world project pipelines where users don't always control which IFC version they receive.

Thanks !

PR #3534, opened 2026-08-30
fix(parser): stop fabricating spatial-node and pset/qset name placeholders

BIMvoice:
## Summary
Part of the sweep from #3530's census: the reported candidates that turned out to be real, verified and fixed here.

- `packages/parser/src/spatial-hierarchy-builder.ts`: an unnamed spatial node (`IfcBuildingStorey`/`IfcSite`/`IfcBuilding`/`IfcSpace`) got ``name: `Entity #${expressId}` ``. Proven to flow into `Ifc5Exporter` as a genuinely-declared `bsi::ifc::prop::Name` on IFCX export — our own `@ifc-lite/ifcx` reader then reads that Name back as real, so STEP→IFCX→read manufactures a name the STEP file never had.
- `packages/parser/src/columnar-parser.ts`, `on-demand-extractors.ts` (`extractPsetsFromIds`), `quantity-collect.ts`: a missing pset/qset `Name` fabricated as `PropertySet #<id>` / `QuantitySet #<id>`.

**Boundary chosen: fix at the source.** All three sites now leave `name`/`psetName`/`qsetName` empty (`''`) instead of fabricating an id-based placeholder, rather than filtering the placeholder out at each export/query boundary. A UI layer that wants a display label for an unnamed node/set derives one at render time; nothing downstream receives an id-based value indistinguishable from a genuinely-declared one.

**The material-properties candidate was NOT a fabrication, and is not changed here.** An earlier revision of this PR also touched `readMaterialPropsEntity`'s `entityType || 'Material Properties'` and claimed the placeholder was reachable. It is not. That `||` arm only fires when `entityType` is falsy, and `entityType` is `entity.type` from `EntityExtractor.extractEntity`, whose `^#(\d+)\s*=\s*(\w+)\(` capture cannot yield an empty string — so the arm is dead. Measured rather than argued: with `#211=IFCMATERIALPROPERTIES($,$,(#210),#200)` (Name = `$`) associated to a wall, `extractMaterialPropertiesOnDemand` returns `name: "IFCMATERIALPROPERTIES"` at this branch's head and at merge-base `9b709c514`, identical. The line is reverted to the merge-base text. Using the IFC class code as a set name IS the same defect class as #3524, but removing it changes what `PropertiesPanel` groups material psets under and belongs in its own PR.

**Fallout the empty name creates, fixed in the same PR.** Two places composed sets from more than one source and used the *name* as an identity key. That only ever worked because unnamed sets carried distinct `PropertySet #<id>` placeholders; with `''` they all compare equal.
- `extractTypePropertiesOnDemand`, `extractTypeEntityOwnProperties` and `extractTypeQuantitiesOnDemand` fold a type's `HasPropertySets` list together with the sets attached to it by `IfcRelDefinesByProperties`, dropping a second-source entry whose name was already seen. A type carrying one unnamed set per source lost the second one outright — reproduced on a wall type with `#300` in `HasPropertySets` and `#311` via `IfcRelDefinesByProperties`, both unnamed: merge-base returns both sets, the first revision of this branch returned only the first. The fold now dedupes by express id (`appendSetsFromSecondSource`, in `property-set-merge.ts`) and applies the name check only to sets that have a name, so behaviour for named sets is unchanged — including the existing rule that an empty set from one source must not shadow a populated same-named set from the other.
- `mergeInheritedPropertySets` had the same shape and now leaves an unnamed inherited set as its own entry instead of folding it into an unrelated unnamed occurrence set.

Removing the `QuantitySet #<id>` fabrication orphaned `readQuantitySet`'s `qsetId` parameter (oxlint reported it unused); it is deleted with its two call-site arguments. `readQuantitySet` is not re-exported from the package index and is absent from `scripts/api-surface.json`, so this carries no semver weight.

**On the Parquet claim specifically:** I could not reproduce "flows into Parquet's `PsetName`/`QsetName` columns" through `IfcParser.parseColumnar()` → `ParquetExporter`. `parseColumnar()`'s typed `properties`/`quantities` tables are always built empty by design (comment in `columnar-parser.ts`: "Empty property/quantity tables - use on-demand extraction instead") — real pset/qset data lives only behind the on-demand `store.getProperties()`/`getQuantities()` accessors, which `ParquetExporter.writeProperties()`/`writeQuantities()` never call (they read the typed table directly). So the fabricated name never reaches a Parquet file through that path.

What I *did* prove reachable, and is the actual fix target for the pset/qset half: `store.getProperties()`/`getQuantities()` are exactly what `EntityNode.properties()`/`quantities()` (`packages/query`) call, and that is the surface both `packages/mcp`'s tool responses and the SDK's `bim.properties()`/`bim.quantities()` return verbatim — the same "declared-looking placeholder in a real API response" bug class already fixed for MCP's `materials_list` in #3515.

**Consumer-count check before choosing the boundary (per instructions):**
- Spatial node name: most UI consumers already have an explicit `name || fallback` pattern (`PropertiesPanel.tsx`, `serverDataModel.ts`, `AddElementPanel.tsx`, `context-builder.ts`) — small gap, one file (`treeDataBuilder.ts`) falls back to the generic type name instead of an id-numbered placeholder now; that's an acceptable UI-only display change, not a correctness issue. `apps/viewer/src/utils/spatialHierarchy.ts`'s doc comment, which described the `Entity #N` placeholder as still being synthesized, is corrected here.
- Pset/qset name: 8+ viewer render/React-key sites read `pset.name`/`qset.name` raw with no existing fallback — too large a surface to patch safely in this PR, and `PropertySet`/`QuantitySet` carry no id downstream of `getForEntity()` to reconstruct an id-based placeholder anyway (only the extraction functions touched here still have `psetId`/`qsetId` in scope). Left as a UI-only, non-blocking follow-up; per the task instructions a placeholder that reaches only UI is acceptable to leave. Worth naming the concrete consequence rather than calling it a display change: `PropertiesPanel.tsx:2179`/`:2197` key their cards on `pset.name`/`qset.name`, so an element carrying two unnamed sets now renders two cards keyed `''`. That collision is pre-existing (`entity-node.ts:258-263` documents two sets sharing one Name as a legitimate model shape, #2907); this PR makes a common exporter output hit it.

## RED/GREEN evidence
- `packages/export/src/fabricated-spatial-name-escape.test.ts`: parses a minimal IFC4 fixture with an unnamed `IFCBUILDINGSTOREY`, runs the real `Ifc5Exporter`, and asserts no `bsi::ifc::prop::Name` is written for it. Reverting only `spatial-hierarchy-builder.ts` to `9b709c514` → `AssertionError: expected 'Entity #99' not to match /^Entity #\d+$/`, exit 1. Control case: a named `IFCSITE` still round-trips `'TestSite'` through every mutation.
- `packages/query/src/fabricated-pset-name-escape.test.ts`: parses a minimal IFC4 fixture with an unnamed `IFCPROPERTYSET`/`IFCELEMENTQUANTITY`, runs `EntityNode.properties()`/`quantities()`, and asserts the name is `''`. Reverting only `columnar-parser.ts` → `expected 'PropertySet #237' not to match /^PropertySet #\d+$/`; reverting only `quantity-collect.ts` → `expected 'QuantitySet #311' not to match /^QuantitySet #\d+$/`. Both exit 1. Control case: a named pset/qset on a second wall round-trips `'Pset_WallCommon'`/`'Qto_WallBaseQuantities'`.
- Reverting only `on-demand-extractors.ts` leaves both files GREEN: that fixture routes through `ColumnarParser`'s own copy, so the `extractPsetsFromIds` copy is fixed but not covered by the new test. Stated rather than rounded up; the change there is byte-identical to the covered copy.
- New tests for the fallout, six in `packages/parser/test/on-demand-type-properties.test.ts` and `on-demand-type-quantities.test.ts` plus one in `merge-inherited-property-sets.test.ts`, each mutation-checked with the test kept and the source reverted (output redirected to a file, `$?` read separately): dropping the `set.name &&` guard fails the three "keeps two DIFFERENT unnamed sets" tests; dropping the id filter fails the three "lists ONE entry" tests; reverting the `mergeInheritedPropertySets` guard fails its test with 1 set where 2 are expected.

## Test plan
- [x] `packages/parser` full suite: 957 passing / 2 skipped
- [x] `packages/export` full suite: 1034 passing / 30 skipped
- [x] `packages/query` full suite: 232/232 passing
- [x] `packages/ids` 829/829, `packages/lens` 186/186, `packages/lists` 235/235 — the consumers of `mergeInheritedPropertySets` and of pset/qset names
- [x] `pnpm exec tsc --noEmit` clean in `packages/parser`
- [x] `pnpm exec oxlint` clean on the changed files (the `qsetId` unused-parameter warning this branch introduced is gone)
- [x] `node scripts/check-module-size.mjs` exits 0; `on-demand-extractors.ts` is 992 lines against its existing 1004 budget, no allowlist change
- [x] Changeset added (`@ifc-lite/parser`, patch — no export added, removed or renamed at the package surface, so `scripts/api-surface.json` is untouched)
- [x] `packages/query/vitest.config.ts` gained the `@ifc-lite/encoding`/`@ifc-lite/ifcx`/`@ifc-lite/pointcloud` source aliases `packages/export`'s config already has, needed to run the new test against source on a clean checkout without built dists

Not verified locally: `packages/mcp` and the full `apps/viewer` suite need built dists for a dozen workspace packages and fail on module resolution in a source-only checkout; the viewer run produced 226 `ERR_MODULE_NOT_FOUND` failures and zero assertion failures. CI's Node tests and Viewer shards cover both.

## Known follow-ups, deliberately out of scope
- The IFC class code used as a material pset name (`readMaterialPropsEntity`) is a genuine #3530-class fabrication, left untouched here because removing it re-groups `PropertiesPanel`'s material cards.
- `apps/viewer/src/lib/collab/step-seed.ts:128` builds ``bsi::ifc::prop::${pset.name}::${prop.name}``, so unnamed psets now share one `bsi::ifc::prop::` namespace; two unnamed psets on one element with a same-named property overwrite each other in the seeded attributes.
- `packages/lens/src/discovery.ts:100`/`:115` and `packages/lists/src/discovery.ts:42`/`:56` do `if (!pset.name) continue;`, so unnamed sets now drop out of the pset/qset pickers instead of appearing under a fabricated id name. Correct direction, but a visible change to what those pickers list.
- `packages/lists/src/engine.ts:573` resolves `col.psetName ?? ''` through `compileNameMatcher`, which compiles `''` to `name === ''`. A saved list column of source `property` with no `psetName` matched nothing before and now matches every unnamed pset. Reachable only from a hand-written or imported list definition; every preset sets `psetName`.


louistrue (approved):

PR #3536, opened 2026-08-30
fix(viewer): fall back to a labeled placeholder for a blank pset/qset header

BIMvoice:
## Summary

Reviewing #3534 (which stops the parser fabricating `PropertySet #<id>` / `QuantitySet #<id>` placeholders for a null `Name`) surfaced a gap on the UI side: nothing in the viewer's properties panel handles an *empty* pset/qset name, so once #3534 lands an unnamed set renders a blank group header.

**This is also reachable on unmodified main today**, independent of #3534: `IfcRoot.Name` is optional, and `extractPsetsFromIds` (`packages/parser/src/on-demand-extractors.ts`) only fabricates a placeholder when `Name` is the null marker `$` (not a string at all):

```ts
const psetName = typeof psetAttrs[2] === 'string' ? psetAttrs[2] : `PropertySet #${psetId}`;
```

A STEP file that declares `Name` as the empty string literal `''` (rather than `$`) already produces `psetName = ''`, which flows through `EntityNode.properties()`/`quantities()` to the viewer untouched — no fix needed upstream to reproduce this.

## Consumer sites

`PropertySetCard.tsx` and `QuantitySetCard.tsx` are the only two places that render a pset/qset name as a group header — every panel entry point (`PropertiesPanel.tsx` occurrence/type/material properties, `ModelMetadataPanel.tsx` project properties, `MaterialTotalsPanel.tsx` material properties) renders through these two shared components, so fixing the header in both covers every caller. (Checked, and left alone as a separate concern: `ListBuilder.tsx`'s column-config labels and the MCP-playground's simulated `getPropertySets` response, which is raw declared-data fidelity, not a header render.)

## Fix

One display helper, `setDisplayName(name, kind)` in the new `apps/viewer/src/components/viewer/properties/setDisplayName.ts`, used at both header sites. It mirrors `treeDataBuilder.ts`'s `getName || "<Type> #<id>"` convention for element rows, but no pset/qset express id reaches these headers (confirmed while tracing the type through `packages/data`, `packages/parser`, `packages/query`, and the viewer's own `encodingUtils.ts`: a `PropertySet` does keep a `GlobalId` as far as `PropertiesPanel`, but nothing on any of these shapes carries a pset/qset express id, and the cards' own prop types in `encodingUtils.ts` declare no id field at all), so the fallback names only the kind: `Unnamed Property Set` / `Unnamed Quantity Set`.

## RED/GREEN evidence

`apps/viewer/src/components/viewer/properties/PropertySetCard.emptyName.test.tsx`: renders `PropertySetCard`/`QuantitySetCard` with `name: ''` and asserts the header button's text (isolated from the property/quantity rows, which live in a sibling `CollapsibleContent`) is non-blank once the count-badge digits are stripped.

- RED on unmodified main: `expected a non-blank header, got: "0"` for both cards (the header collapsed to just the "0" count badge).
- GREEN after the fix: `Unnamed Property Set0` / `Unnamed Quantity Set0`.
- Control (unchanged behavior): a named pset/qset still renders its real name verbatim — passes before and after.

## Test plan
- [x] New test: 3/3 passing, plus the pre-existing `PropertySetCard.escapes.test.tsx` (4/4) still passing — no regression to the verbatim-decode behavior it guards
- [x] `pnpm exec tsc --noEmit` clean in `apps/viewer`
- [x] `node scripts/check-module-size.mjs`: 0 new over budget
- [x] `node scripts/check-unused-locals.mjs`: no new unused locals
- [x] Changeset added (`@ifc-lite/viewer`, patch — private but versioned, per existing convention)

## Summary by CodeRabbit

- **Bug Fixes**
  - Property and quantity set cards now show clear fallback labels when their IFC name is missing or blank.
  - Existing names continue to display correctly.

- **Tests**
  - Added coverage for unnamed and named property and quantity sets.


louistrue (approved):

PR #3539, opened 2026-08-30
fix(mutations): apply an existing property's SET/DELETE to the first same-named pset

BIMvoice:
## Builds on #3538

This PR now sits on top of #3538 (merged), which fixed `getForEntity` writing a brand-new property into every same-named base pset instead of just the first (via a `newMemberPlacement` helper). This PR's `same-name-pset-claims.ts` claims computation subsumes that new-property placement rule and additionally decides which instance an EXISTING property's SET/DELETE lands on, so `getForEntity`'s property path now has one mechanism, not two layered ones. `newMemberPlacement` itself is untouched and still used for the quantity path (`getQuantitiesForEntity`), which this PR does not touch.

## Trace

`setProperty`/`deleteProperty` key a mutation as `${entityId}:${psetName}:${propName}` — no identity past the pset *name*. `getForEntity`'s read-back loop applied that single mutation to *every* base pset instance whose own properties matched the name, so editing/deleting `FireRating` on an entity with two same-named `Pset_WallCommon` sets changed both instead of one. A related sub-bug: a property present on only one of two same-named instances was mis-classified as "new" when processing the *other* instance and got duplicated there too — same underlying mechanism as #3538's now-fixed new-property bug, just triggered via the existing-elsewhere path instead of the genuinely-new path.

## RED → GREEN

`packages/mutations/test/mutable-property-view.duplicate-pset.test.ts` — new describe blocks (merged with #3538's own new-property test blocks in the same file, no duplicate names):
- **RED on unmodified main** (confirmed by reverting the fix locally and re-running): `setProperty(1, 'Pset_WallCommon', 'FireRating', 'RF90')` on an entity with two `Pset_WallCommon` instances (FireRating: RF30 / RF60) mutated BOTH instances to RF90. Expected: only the first instance changes, the second retains RF60.
- Control: a property present on only ONE of the two sets (`Reference`, only on the second instance) also failed on unmodified main — it got duplicated onto the FIRST instance too, not just left alone. Expected/GREEN: it mutates only the instance that owns it.
- Same shape for `deleteProperty` — deleting `FireRating` removed it from both instances on main; now removes only the first.

GREEN after the fix. `packages/mutations/test/mutable-property-view.duplicate-pset.test.ts`: 13 passed (4 pre-existing #3468 cases + 6 from #3538's new-property fix, retained + 3 new from this PR). Full package suite: `pnpm exec vitest run` — 232 passed, 10 skipped.

## Fix

`getForEntity` now computes, once per entity, which base pset instance "claims" each property name: the first instance (in base-pset order) whose own base properties already carry that name, or — for a genuinely new property (absent from every same-named instance) — the first instance of that pset name. This mirrors the first-match-wins read semantics `findPropertyInSets`/`PropertyTable.getProperty` already use for same-named reads (#3468) and replaces #3538's narrower `newMemberPlacement`-based guard for the property path with this single mechanism.

The claim computation and the per-instance mutated-property list are extracted to a sibling module, `packages/mutations/src/same-name-pset-claims.ts` — `mutable-property-view.ts` was already at its 2121-line budget (never raised; net effect after extraction is 2094 lines, more headroom than before).

## Export path

Grepped every `psetName` decomposition site in `packages/mutations/src` and `packages/export/src`. The STEP exporter's property-set collection phase (`packages/export/src/step-property-set-collection.ts`) reads the FULL mutated property-set list via `mutationView.getForEntity(entityId)` — it does not independently re-decompose or re-apply a mutation key against every same-named pset. So this is one path, not two that could disagree: the exporter inherits the fix with no separate change. No `@ifc-lite/export` changeset needed.

## Verification

- `pnpm exec tsc --noEmit` in `packages/mutations`: clean.
- `pnpm exec turbo run typecheck --filter=@ifc-lite/export`: clean (no export-side change, verified as a consumer of `getForEntity`).
- `packages/mutations` suite: 232 passed, 10 skipped.
- `packages/export` suite (consumer of `getForEntity`, not touched): 1032 passed, 30 skipped — no regression.
- `node scripts/check-module-size.mjs`: `mutable-property-view.ts` at 2094 lines (budget 2121, not raised).
- `node scripts/check-unused-locals.mjs` (root, full tree): `✅ No new unused locals (51 packages, 943 known, none increased).`
- `node scripts/check-test-wiring.mjs`: clean.
- Changeset: `@ifc-lite/mutations` patch (describes only this PR's fix, not #3538's).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed property updates and deletions when multiple property sets share the same name.
  - Changes now apply only to the first matching property-set instance, consistently with existing property-read behavior.
  - Prevented updates or deletions from unintentionally affecting every same-named property set.
  - New properties are placed in the first matching property set, providing more predictable exported results.


louistrue (approved):

PR #3542, opened 2026-08-30
fix(cli): require a full #<digits> match for an overlay relation ref

### LINKED ISSUE #3502: cli: query overlay parses #42junk and #42.5 as express id 42
Deferred from the review of #3500 (merged): a CodeRabbit finding that is correct but is hardening rather than a regression, so it was not worth holding the PR for.

`packages/cli/src/query-overlay-relations.ts:38`

```
const id = Number.parseInt(trimmed.slice(1), 10);
return trimmed.startsWith('#') && Number.isFinite(id) ? [id] : [];
```

`Number.parseInt` stops at the first non-digit, so a near-miss reference resolves to a real entity instead of being rejected: `#42junk` and `#42.5` both come back as express id `42`. Relationship values on this path come from the mutation overlay, so a malformed end is user-reachable rather than parser-guaranteed.

Require a full match and a safe integer:

```
const match = /^#(\d+)$/.exec(trimmed);
if (!match) return [];
const id = Number(match[1]);
return Number.isSafeInteger(id) ? [id] : [];
```

Worth a test with `#42junk`, `#42.5`, `#`, and an id past `Number.MAX_SAFE_INTEGER`.

BIMvoice:
## Summary
- `query-overlay-relations.ts`'s `refIds()` parsed a queued `IfcRel…` relating/related end with `Number.parseInt(trimmed.slice(1), 10)`, which stops at the first non-digit character. A malformed reference — `#42junk`, `#42.5` — resolved to express id `42` instead of being rejected, so a near-miss reference on a mutation-overlay relationship silently bound to a real entity.
- Now requires a full `/^#(\d+)$/` match and a `Number.isSafeInteger` id, so a malformed reference or one past `Number.MAX_SAFE_INTEGER` resolves to `[]`.
- This also fixes the `.slice(1)` call being evaluated on a string that might not start with `#` before the `startsWith` guard ran — the regex only ever runs against the anchored pattern, so that ordering issue is gone too.

Fixes #3502

## RED / GREEN
`packages/cli/src/query-overlay-relations.test.ts`, exercised through `HeadlessBackend.query.related()` since `refIds` itself is not exported (same overlay contract as the sibling `headless-backend-related-overlay.test.ts`):

- `#42junk` — before: `[42]`, after: `[]`
- `#42.5` — before: `[42]`, after: `[]`
- a reference past `Number.MAX_SAFE_INTEGER` (`#9007199254740993`) — before: `[9007199254740992]` (precision-lossy but "finite"), after: `[]`
- `#` (bare) — before and after: `[]` (already correct, kept as a supplementary case)
- `#42` (control) — before and after: `[42]`

## Gates
- `pnpm exec tsc --noEmit` (packages/cli): clean.
- `pnpm exec vitest run` (packages/cli): 55 files / 566 passed, 8 skipped.
- `node scripts/check-unused-locals.mjs`: 51 packages measured, 943 known violations, none increased — no baseline change needed.
- `node scripts/check-module-size.mjs`: OK, 0 new files over 400 lines.

### DIFF
--- a/.changeset/cli-overlay-relation-ref-parseint.md
+++ b/.changeset/cli-overlay-relation-ref-parseint.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/cli': patch
+---
+
+`HeadlessBackend.query.related()` no longer resolves a malformed `IfcRel…` reference to a real entity id (#3502). The overlay's `'#42'` reference parser used `Number.parseInt`, which stops at the first non-digit character, so a relating/related end authored as `#42junk` or `#42.5` on a queued relationship resolved to express id `42` instead of being rejected. It now requires a full `/^#(\d+)$/` match and a `Number.isSafeInteger` id, so a malformed or out-of-range reference (past `Number.MAX_SAFE_INTEGER`) resolves to nothing.
--- a/packages/cli/src/query-overlay-relations.test.ts
+++ b/packages/cli/src/query-overlay-relations.test.ts
@@ -1,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression for #3502: `refIds()` (the `'#42'` reference parser inside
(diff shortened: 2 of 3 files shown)


louistrue (approved):

PR #3543, opened 2026-08-30
fix(viewer): route command-palette JSON export through exactTypeName

### LINKED ISSUE #3503: viewer: command-palette JSON export still emits the coalesced class name
Deferred from the review of #3475 (merged under the relaxed bar for viewer-only changes).

#3475 routed the Lists Class column off the coalesced class name and onto the real one, but a third JSON export path was left on the old helper:

`apps/viewer/src/components/viewer/CommandPalette.tsx:592` still calls `getTypeName`, so an export triggered from the command palette emits the coalesced name while the two paths the PR fixed emit the real one. The same model exported two ways now disagrees.

Route it through the same helper the PR introduced, and add a case that exercises the command-palette path so the three exports are pinned together rather than two of three.

BIMvoice:
## Summary
- `CommandPalette.tsx`'s `export:json` action still read `EntityTable.getTypeName` directly — the `IfcTypeEnum`-coalesced family name — while #3475 routed the Lists Class column (`apps/viewer/src/lib/lists/adapter.ts`) and the Parquet `Type` column (`@ifc-lite/export`, #3325) onto `@ifc-lite/data`'s `exactTypeName()`. Exporting the same model via the palette named an `IFCDOORSTANDARDCASE` entity `IfcDoor` while the other two export paths named it `IfcDoorStandardCase` — the same model exported two ways disagreed.
- The row-building loop is extracted from `CommandPalette.tsx` (already at its module-size budget of 849 lines, currently 846) into a sibling module, `buildCommandPaletteJsonEntities` (`apps/viewer/src/components/viewer/commandPaletteJsonExport.ts`), so it can be exercised without mounting the palette's React tree, and now routes `type` through `exactTypeName(d.entities, id)`.

Fixes #3503

## Helper reused
`exactTypeName()` from `@ifc-lite/data` (`packages/data/src/exact-type-name.ts`) — the same accessor #3475 introduced for the Lists Class column and the Parquet Type column.

## RED / GREEN
`apps/viewer/src/components/viewer/CommandPalette.exact-type.test.ts` pins all three export paths — the Lists adapter, `ParquetExporter`, and the new `buildCommandPaletteJsonEntities` — against one shared fixture (an `IFCDOORSTANDARDCASE` entity plus a plain `IFCDOOR` negative control), per the maintainer's ask for "three exports pinned together rather than two of three".

- RED (temporarily reverting the extracted helper's `type` line to `d.entities.getTypeName(id)`): the palette assertion failed — actual `'IfcDoor'`, expected `'IfcDoorStandardCase'` (the Lists and Parquet assertions in the same test already passed, confirming only the palette path was broken).
- GREEN (after routing through `exactTypeName`): all three assertions pass — Lists, Parquet, and the palette JSON export all answer `'IfcDoorStandardCase'` for the StandardCase entity and `'IfcDoor'` for the control; `store.entities.getTypeName` still answers the coalesced `'IfcDoor'`, confirming grouping is unaffected.

## Gates
- `pnpm exec tsc --noEmit` (apps/viewer): clean.
- Targeted tests (`CommandPalette.exact-type.test.ts`, `CommandPalette.anonymized-export.test.tsx`, `lib/lists/class-column-exact-type.test.ts`): all pass.
- `node scripts/check-module-size.mjs`: OK, 0 new files over 400 lines; `CommandPalette.tsx` stays at 846 lines (budget 849), unchanged.
- `node scripts/check-unused-locals.mjs`: 51 packages measured, 943 known violations, none increased — no baseline change needed.

### DIFF
--- a/.changeset/command-palette-json-export-exact-type.md
+++ b/.changeset/command-palette-json-export-exact-type.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/viewer': patch
+---
+
+The command palette's `Export JSON` entry now names each entity's declared class instead of its `IfcTypeEnum`-coalesced family (#3503). `#3475` routed the Lists Class column and the Parquet `Type` column onto `@ifc-lite/data`'s `exactTypeName()`; the palette's `export:json` action was a third, independent caller still reading `EntityTable.getTypeName` directly, so exporting the same model via the palette named `IFCDOORSTANDARDCASE` entities `IfcDoor` while the other two export paths named them `IfcDoorStandardCase`. The row-building logic is extracted to `buildCommandPaletteJsonEntities` (`apps/viewer/src/components/viewer/commandPaletteJsonExport.ts`) so all three export paths now route through the same accessor.
--- a/apps/viewer/src/components/viewer/CommandPalette.exact-type.test.ts
+++ b/apps/viewer/src/components/viewer/CommandPalette.exact-type.test.ts
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Three independent export paths must agree on an entity's class: the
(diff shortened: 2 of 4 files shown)


louistrue (approved):

PR #3544, opened 2026-08-30
fix(viewer): classify the worker-script wasm skew wrapper as recovered noise

BIMvoice:
## Summary

Refs #3533

Traced on `main`: after a worker-script skew (`#1680`), the recovery reload **does** run — `installWasmVersionSkewRecovery`'s listener trusts the event's `kind: 'worker-script'` directly:

```ts
window.addEventListener(WASM_ASSET_UNAVAILABLE_EVENT, (event) => {
  const detail = (event as CustomEvent<{ message?: string; kind?: string }>).detail;
  if (detail?.kind === 'worker-script') {
    recoverFromWorkerScriptSkew();
    return;
  }
  recoverFromWasmVersionSkew(detail?.message ?? '');
});
```

`recoverFromWorkerScriptSkew`'s own doc comment says why: *"re-running the message matcher here would reject exactly the case the event exists for."*

But the separate PostHog noise-suppression gate (`shouldSuppressWasmSkewNoise`, wired as `before_send` on every captured event, explicit `captureException` included) does the exact thing that comment warns against — it only has the final exception message text (not the `kind`), and re-derives classification purely via the strict `isWasmAssetUnavailableError` matcher:

```ts
export function shouldSuppressWasmSkewNoise(...): boolean {
  const message = exceptionMessageOf(event);
  if (message === undefined || !isWasmAssetUnavailableError(message)) return false;
  ...
```

`isWasmAssetUnavailableError` requires a wasm/MIME/status-code token. The worker-script wrapper's message — `"Pre-pass worker failed: worker script failed to load (possibly a stale deployment)"` (the literal string in issue #3533), and its sibling `"Geometry worker failed: worker script failed to load (possibly a stale deployment)"` from the process-worker pool — carries none of those tokens (confirmed: `isWasmAssetUnavailableError` returns `false` for both). So the reload runs, but the exception is still captured as if unhandled.

This is case **(a)**: the error is captured AND the reload happens; the report is noise for a handled condition. Fix: classify it as handled at the capture site, without touching the reload path or loosening `isWasmAssetUnavailableError` (which is also used, unrelated, by the generic `unhandledrejection`/`error` backstop listeners).

## Changes

- `packages/geometry/src/wasm-asset-error.ts`: new exported `isWorkerScriptSkewMessage`, matching the literal worker-script wrapper suffix synthesized only by `geometry-parallel.ts`'s pre-pass/process-worker `onerror` handlers when the worker never spoke.
- `apps/viewer/src/lib/wasm-version-skew.ts`: `shouldSuppressWasmSkewNoise` now matches on `isWasmAssetUnavailableError(message) || isWorkerScriptSkewMessage(message)`.
- Every other pre-pass/worker failure (a real crash, an OOM, a genuine "worker terminated unexpectedly") is still captured unchanged — covered by a control test.

## Test plan

RED (recorded on unmodified `main`, `apps/viewer/src/lib/wasm-skew-noise.test.ts`): `shouldSuppressWasmSkewNoise` returned `false` for the `#3533` message — expected `true`.

GREEN: added `isWorkerScriptSkewMessage` test coverage in `packages/geometry/src/wasm-asset-error.test.ts` (18/18 pass) and `apps/viewer/src/lib/wasm-skew-noise.test.ts` (both wrapper variants suppressed, control non-skew failure still captured — 111/111 pass in the touched test files).

- [x] `pnpm exec vitest run src/wasm-asset-error.test.ts` in `packages/geometry`
- [x] `apps/viewer`'s `node:test` runner on `wasm-skew-noise.test.ts` + `wasm-version-skew.test.ts` + `analytics.test.ts`
- [x] `pnpm exec tsc --noEmit` in `packages/geometry` and `apps/viewer` — no new errors from touched files (pre-existing `@ifc-lite/data` build-order errors in this fresh worktree are unrelated)
- [x] `node scripts/check-module-size.mjs` — OK, budget unchanged (kept the export on one line to avoid growing `packages/geometry/src/index.ts` past its pinned budget)
- [x] `node scripts/check-unused-locals.mjs` — no new complaints tied to touched files
- [x] Changeset added (`@ifc-lite/geometry` + `@ifc-lite/viewer`, patch)
- [x] `scripts/api-surface.json` updated for the new `isWorkerScriptSkewMessage` export (full `pnpm build`-gated `check-api-surface.mjs` couldn't run end-to-end in this fresh worktree without a repo-wide build, which the concurrent-agent resource note rules out; the entry was hand-verified against the actual built `packages/geometry/dist/index.d.ts`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced false error reports caused by expected WebAssembly worker version mismatches.
  - Continued capturing genuine worker failures for reliable error monitoring.

- **New Features**
  - Added a public utility for identifying worker-script version-skew errors.

- **Tests**
  - Added coverage for worker version-skew detection, including valid and invalid error cases.

- **Documentation**
  - Added release notes for the geometry and viewer packages.

### DIFF
--- a/.changeset/wasm-worker-script-skew-noise.md
+++ b/.changeset/wasm-worker-script-skew-noise.md
@@ -1,5 +1,5 @@
 ---
-'@ifc-lite/geometry': patch
+'@ifc-lite/geometry': minor
 '@ifc-lite/viewer': patch
 ---
 
--- a/.changeset/wasm-worker-script-skew-noise.md
+++ b/.changeset/wasm-worker-script-skew-noise.md
@@ -1,0 +1,6 @@
+---
+'@ifc-lite/geometry': patch
+'@ifc-lite/viewer': patch
+---
+
+The viewer's PostHog noise gate for auto-recovered wasm version skew (#1363) only matched the wasm-binary MIME/404 message text, so a worker-SCRIPT skew (classified separately, by `kind`, since #1680) reloaded correctly but was still captured to error tracking as if unhandled (#3533). `@ifc-lite/geometry` now exports `isWorkerScriptSkewMessage` for the worker-script wrapper signature (`"…worker script failed to load (possibly a stale deployment)"`), and the viewer's `shouldSuppressWasmSkewNoise` matches on it alongside the existing wasm-MIME matcher. Every other pre-pass/worker failure is still captured unchanged.
--- a/apps/viewer/src/lib/wasm-skew-noise.test.ts
+++ b/apps/viewer/src/lib/wasm-skew-noise.test.ts
@@ -113,6 +113,52 @@ describe('shouldSuppressWasmSkewNoise', () => {
     assert.equal(shouldSuppressWasmSkewNoise(skewEvent(HTTP_SKEW), h.deps), false);
   });
 
+  it('suppresses the worker-script skew wrapper (#3533) — recovered by kind, not by MIME text', () => {
+    // The pre-pass worker's own `onerror` handler never sees a wasm-MIME
+    // token — it wraps a synthetic detail string and dispatches
+    // `WASM_ASSET_UNAVAILABLE_EVENT` with `kind: 'worker-script'` (trusted by
+    // `recoverFromWorkerScriptSkew`, which reloads WITHOUT re-running the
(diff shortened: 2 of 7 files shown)


louistrue (approved):

PR #3546, opened 2026-08-30
fix(parser): honour a negative-zero degree in RefLatitude/RefLongitude

BIMvoice:
## Context

Assigned lens: hunt `packages/solar` (sun position / sunrise-sunset / equation of time) for numeric-parity defects against independent references, and check the conventions feeding it — including how `IfcSite.RefLatitude`/`RefLongitude` sign handling flows into the sun calculation.

`packages/solar` itself turned out to already be extensively hardened by prior PRs (#2785 pinned it against an independently re-derived NOAA reference, #2413/#2148/#2329 closed several mutation-verified gaps). Re-verifying it against external references found no defect:

| Case | Package output | Independent reference | Source | Within tolerance |
|---|---|---|---|---|
| Declination, 2024 June solstice 00:00Z | 23.4385549° | 23.4385549° | Pre-existing `noaa-reference.test.ts` pin, independently re-derived from NOAA's own equations | yes (±1e-6) |
| Equation of time, same instant | -1.8162166 min | -1.8162166 min | same | yes (±1e-6) |
| Boulder CO sun pos, 2024-06-21T18:00Z | alt 68.99372°, az 136.25442° | NOAA's own published worked example for this site/instant | same | yes (±1e-4) |
| Sydney sunrise/sunset, 2024-12-21 (local) | sunrise 2024-12-20T18:41:13Z, sunset 2024-12-21T09:05:44Z | sunrise 18:39:30Z, sunset 09:07:00Z | `api.sunrise-sunset.org` (independent NOAA-formula implementation) | yes (within ~1m43s / ~1m16s, inside the package's claimed ~1-2 min sunrise/sunset accuracy) |
| Tokyo sunrise/sunset, 2024-06-21 (local) | sunrise 2024-06-20T19:26:04Z, sunset 2024-06-21T10:00:34Z | sunrise 19:24:29Z, sunset 10:01:59Z | same | yes (within ~1m35s / ~1m25s) |

Viewer-side conventions (`apps/viewer/src/lib/geo/solar-direction.ts`, `useSolarEnvironment.ts`, `cesium-sun.ts`) are consistent: ENU frame, east-clockwise azimuth from true north, and meridian-convergence handling all agree with `packages/solar`'s documented conventions, and Cesium's own `eastNorthUpToFixedFrame` is used for the dome — no East/North swap or grid/true-north mismatch found.

## The actual defect: found while checking `IfcSite.RefLatitude`/`RefLongitude` sign handling

`IfcCompoundPlaneAngleMeasure` (`packages/parser/src/georef-extractor.ts`, `compoundPlaneAngleToDecimalDegrees`) is a STEP `LIST[3:4] OF INTEGER` (degrees, minutes, seconds, optional millionth-seconds). The spec's canonical sign encoding ("all non-zero measure components shall have the same sign") puts the sign on the first non-zero component — 0°30'S is `(0, -30, 0)` — and `main` already handled that correctly via `minutesRaw < 0`.

The gap is a non-canonical but plausible encoding some writers use instead: signing the *degree* token even when its magnitude is `0`, e.g. `(-0, 30, 0)` for the same 0°30'S. Read strictly against the spec that's actually `+0°30'` (the only non-zero component, `30`, is positive) — but it's the evident intent of a writer that always carries the hemisphere sign on the leading/degree component. I don't have a specific writer to cite that emits this; treating it as a defensive leniency rather than a spec requirement.

The STEP tokenizer's `parseFloat` preserves that literal as IEEE-754 negative zero (`parseFloat('-0') === -0`). The sign test was:

```ts
const sign = degreesRaw < 0 || minutesRaw < 0 || secondsRaw < 0 || millionthsRaw < 0 ? -1 : 1;
```

`-0 < 0` is `false` in JavaScript, so for a writer using this non-canonical convention, `RefLatitude`/`RefLongitude` of `(-0, 30, 0)` (intended as 0°30'S) silently became +0.5° (0°30'N) — the legacy `IfcSite` georeference fallback would put such a site's sun on the wrong side of the globe (northern/eastern instead of southern/western hemisphere), which is exactly the wrong-side-of-the-building class of bug this hunt was looking for.

**RED (unmodified `main`):** the fixture `honours a zero-magnitude negative-zero degree component in RefLatitude/RefLongitude` (added in this PR) asserts `mapConversion.northings ≈ -0.5` / `eastings ≈ -0.75` for `RefLatitude = [-0, 30, 0]`, `RefLongitude = [-0, 45, 0]`. On unmodified `main` it fails: `expected 0.5 to be close to -0.5`.

**GREEN:** `compoundPlaneAngleToDecimalDegrees` now checks `Object.is(component, -0)` alongside `< 0` on every component, so a writer that signs the degree token this way is honoured in addition to the spec-canonical minutes/seconds sign.

**Control:** every pre-existing case in `georef-extractor.test.ts` (positive-only compound angles, the canonical spec form via the Prague `[50, 2, 20]`/`[14, 28, 0]` fixture, the ePSet/IfcMapConversion precedence tests) still passes unchanged — 24/24.

## Scope note

This is a `packages/parser` fix (the extractor `packages/solar`'s consumers rely on for site lat/lon), not `packages/solar` itself — `packages/solar`'s own math checked out clean against every external reference above. The equivalent Rust extractor (`rust/core/src/georef.rs::compound_plane_angle_to_degrees`) has the same `< 0.0` sign test, but its STEP tokenizer parses these components through an `i64` integer path (`lexical_core::parse::<i64>("-0") == 0i64`), which has no negative-zero representation at all — the sign is lost before `compound_plane_angle_to_degrees` ever runs, so an `is_sign_negative()` fix there would be a no-op. Fixing that requires preserving the raw literal's sign further up the shared STEP tokenizer (a `rust/core` perf-sensitive-path change per `AGENTS.md`, out of scope for this single-defect PR) — flagging it here rather than silently leaving it unfixed.

## Verification

- `pnpm exec vitest run` in `packages/parser`: 877 passed, 2 pre-existing failures unrelated to this change (`@ifc-lite/wasm` not built in this environment — `parser-worker-panic-forward.test.ts`).
- `pnpm exec tsc --noEmit` in `packages/parser`: clean.
- `pnpm exec vitest run` + `tsc --noEmit` in `packages/solar`: 43/43 passed, clean (unaffected, confirms the negative result above).
- `node scripts/check-module-size.mjs` (root): OK, 0 new over budget.
- `node scripts/check-test-wiring.mjs` (root): OK.
- `node scripts/check-unused-locals.mjs` (root): `packages/parser` and `packages/solar` are not among the packages that fail to compile standalone (pre-existing, unrelated — several packages in this sandbox haven't had every workspace dependency built).
- No exported API changed (new helper is module-private), so `scripts/api-surface.json` needs no update.
- Changeset: `@ifc-lite/parser` patch.


louistrue (approved):

PR #3547, opened 2026-08-30
fix(spatial): stop a NaN-bounded mesh from poisoning its BVH siblings

BIMvoice:
## Summary
- `BVH.computeBounds` (`packages/spatial/src/bvh.ts`) folded child bounds with `Math.min`/`Math.max`, which propagate a NaN operand through every later comparison in the same reduce. One mesh with degenerate/NaN bounds (e.g. a corrupt vertex) therefore NaN'd its entire subtree's aggregate bounds.
- `AABBUtils.intersects` treats a NaN bound as "no intersection" on the axes it touches, so that NaN'd subtree — and every *valid* sibling mesh under it — got pruned out of `queryAABB`, `raycast`, and `queryFrustum`, no matter what the query was.
- Consumer impact: `queryFrustum` backs frustum culling in `packages/renderer/src/index.ts` (a false negative there is invisible geometry), and `queryAABB`/`raycast` back the viewer's SDK spatial adapter (box queries and picking).
- Fix: fold bounds with NaN-safe `<`/`>` comparisons instead of `Math.min`/`Math.max`, so a NaN'd mesh is excluded on its own without poisoning siblings' aggregate bounds. This also brings the TS BVH in line with the Rust port (`rust/clash/src/bvh.rs`'s `compute_bounds`), which already used this comparison shape.

## How this was found
A property-based harness (kept as a permanent test in `packages/spatial/src/bvh.test.ts`, `describe('BVH property harness — tree query vs. brute-force leaf oracle')`) comparing `BVH.queryAABB`/`raycast`/`queryFrustum` against a brute-force per-item oracle across randomized configurations (negative coordinates, zero-extent/degenerate boxes, duplicate expressIds, varying magnitudes) plus named boundary cases (NaN bounds, Infinity bounds, spanning queries, axis-parallel rays, zero-length ray direction, etc.). It surfaced exactly one mismatch: the NaN-bounds case.

Minimized repro (2 items — one NaN-bounded, one valid):
```ts
const meshes = [
  { expressId: 1, bounds: box(NaN, 0, 0, NaN, 1, 1) },
  { expressId: 2, bounds: box(0, 0, 0, 1, 1, 1) },
];
BVH.build(meshes).queryAABB(box(-1, -1, -1, 4, 4, 4));
// main:  []   (valid mesh 2 silently dropped)
// fixed: [2]
```

## Test plan
- [x] New named regression test (`does not let a NaN-bounded mesh hide a valid sibling...`) confirmed RED against unmodified `bvh.ts`, GREEN with the fix.
- [x] Property harness (216+ queries per query type, bounded runtime) passes.
- [x] `pnpm test` in `packages/spatial` — 52/52 passing, ~150ms.
- [x] `node scripts/check-module-size.mjs`, `node scripts/check-unused-locals.mjs`, `node scripts/check-test-wiring.mjs` from repo root — all pass.
- [x] No public export surface changed (`scripts/api-surface.json` untouched).
- [x] Changeset added (`@ifc-lite/spatial`, patch).

### DIFF
--- a/.changeset/fix-bvh-nan-bounds-poison.md
+++ b/.changeset/fix-bvh-nan-bounds-poison.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/spatial': patch
+---
+
+Fix `BVH.queryAABB`/`raycast`/`queryFrustum` dropping a valid mesh when a sibling in the same subtree has NaN/degenerate bounds. `computeBounds` folded child bounds with `Math.min`/`Math.max`, which propagate a NaN operand through every later comparison in the same reduce, so one mesh with bad geometry (e.g. a corrupt vertex) NaN'd its entire subtree's aggregate bounds — and `AABBUtils.intersects` treats a NaN bound as no intersection, pruning that subtree, and every valid sibling under it, out of every query regardless of the query box. `computeBounds` now folds with NaN-safe comparisons (matching the Rust port in `rust/clash/src/bvh.rs`, which already used this shape), so a NaN-bounded mesh is excluded on its own without poisoning its siblings.
--- a/packages/spatial/src/bvh.test.ts
+++ b/packages/spatial/src/bvh.test.ts
@@ -4,7 +4,7 @@
 
 import { describe, expect, it } from 'vitest';
 import { BVH, type MeshWithBounds } from './bvh.js';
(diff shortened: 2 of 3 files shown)


louistrue (approved):

PR #3553, opened 2026-08-30
fix(export): refuse an IFC4-only entity type with no IFC2X3/IFC4 representation instead of copying it verbatim

BIMvoice:
## Summary

Pass 4 of the merge-invariants lens (cross-schema legality). `packages/export/src/schema-converter.ts`'s `convertStepLine` — the per-entity schema conversion both a direct schema downgrade and `MergedExporter`'s cross-schema merge (`renderEntity`, gated on `needsConversion`) run through — only substitutes an `IFCPROXY` placeholder for 4 hand-listed IFC4X3 alignment types (`shouldSkipEntity`). Every other entity type that is genuinely absent from the target schema's generated attribute table falls through unrecognised and is copied verbatim, with its source-schema attribute shape.

Reproduced directly against `convertStepLine('IFC4', 'IFC2X3')` and end-to-end through `MergedExporter` (a single IFC4 model, `IfcWall` → `IfcProductDefinitionShape` → `IfcShapeRepresentation` → `IfcTriangulatedFaceSet` → `IfcCartesianPointList3D`, exported with `{ schema: 'IFC2X3' }`): the output declares `FILE_SCHEMA(('IFC2X3'))` and its body still contains a raw `IFCTRIANGULATEDFACESET`/`IFCCARTESIANPOINTLIST3D` line — entity types IFC2X3 never defined. Confirmed against `packages/data/src/ifc-schema/generated/entities-ifc2x3.ts`: none of `IfcTriangulatedFaceSet`, `IfcCartesianPointList3D`, `IfcIndexedPolyCurve`, `IfcPolygonalFaceSet`, `IfcMapConversion`, `IfcProjectedCRS`, `IfcMaterialConstituentSet` appear there; all appear in `entities-ifc4.ts`.

## The invariant table (pass 4, cross-schema legality)

| Check | Before | After |
|---|---|---|
| Downgrade a rooted (`IfcRoot`) unmapped type (e.g. the existing IFC4X3 alignment types) | `IFCPROXY` placeholder | unchanged — still `IFCPROXY` |
| Downgrade a non-rooted unmapped type present in source, absent in target (`IfcTriangulatedFaceSet`, `IfcCartesianPointList3D`, `IfcIndexedPolyCurve`, `IfcPolygonalFaceSet`, `IfcMapConversion`, `IfcProjectedCRS`, `IfcMaterialConstituentSet`, …) | **copied verbatim under the wrong schema header** (silent illegality) | throws a clear error naming the entity and target schema |
| Same-schema export (control) | unchanged, no conversion path entered | unchanged |
| Renamed/attribute-count-adjusted types (existing `IFC2X3_TO_IFC4`/`IFC4_TO_IFC2X3` maps, `isStrictAttrPrefix` trim/pad) | unaffected | unaffected — the new check only fires when the type is entirely absent from the target table, not when it's merely renamed or attribute-count-adjusted |
| Round-trip completeness (openings, type relations, group membership, psets/quantities, materials, classifications) | not this PR's scope — no regression introduced; the fix only changes behaviour for entity types the target schema cannot represent at all | same |

I did not attempt to make the tessellated-geometry / georeferencing / material-constituent types *convert* into an older schema's structurally different representation (e.g. rewriting a triangulated mesh into an `IfcFacetedBrep`) — that is a real geometry transform, not a string-level STEP rewrite, and out of scope for `schema-converter.ts`. Per `MergeExportOptions.schema`'s doc comment ("any version, will convert if needed"), the honest behaviour where conversion genuinely isn't possible is to refuse loudly rather than emit an invalid file.

### Why IFCPROXY only for rooted types

`IFCPROXY` is an `IfcProduct` (has a `GlobalId`). The existing alignment-type fallback works because those types are rooted (`IfcAlignmentHorizontal` → `IfcLinearElement` → `IfcProduct`, verified via `getInheritanceChainAcrossSchemas`). The newly-caught types are representation items or resource types referenced *positionally* (an `IfcShapeRepresentation.Items` entry, an `IfcGeometricRepresentationContext`/`IfcProjectedCRS` attribute) — substituting `IFCPROXY` there would trade one illegal file (wrong schema) for a differently-illegal one (an `IfcProduct` where an `IfcRepresentationItem` or resource type is expected), and dropping the line instead would leave the referencing entity's `#N` dangling. `resolveUnrepresentedEntity` (new `packages/export/src/schema-untranslatable.ts`) throws instead.

## Why a new sibling module

`packages/export/src/merged-exporter.ts` is untouched (it already calls `needsConversion`/`convertStepLine`, so no changes were needed there). `schema-converter.ts` had only 23 lines of headroom under its `scripts/module-size-allowlist.txt` budget (524); the new detection + error/proxy logic didn't fit, so it's a new module `schema-untranslatable.ts`, imported by `schema-converter.ts` with a single new branch (schema-converter.ts grew from 501 to 510 lines, budget 524, headroom now 14).

## Test plan
- [x] RED on unmodified `main`: `IFCTRIANGULATEDFACESET`/`IFCCARTESIANPOINTLIST3D` survive `convertStepLine('IFC4','IFC2X3')` unchanged (verified locally, not committed as a failing test).
- [x] GREEN: `packages/export/src/schema-converter-untranslatable.test.ts` (new) — 7 cases: 5 non-rooted unmapped types now throw, the existing rooted alignment fallback still returns `IFCPROXY`, same-schema conversion is an untouched no-op (control).
- [x] GREEN end-to-end: `packages/export/src/merged-exporter.test.ts` (new `describe('cross-schema merge legality')`) — `MergedExporter([ifc4Model]).export({ schema: 'IFC2X3' })` throws naming `IFCTRIANGULATEDFACESET`; the same model targeting `{ schema: 'IFC4' }` (control) exports cleanly with no dangling refs.
- [x] `node scripts/check-module-size.mjs` → `check-module-size: OK (2031 files measured, 306 allowlisted, 0 new over 400)`
- [x] `pnpm run check:vitest-timeout-audit` — no regex literals with parens added in the new tests.
- [x] `node scripts/check-test-wiring.mjs` → OK.
- [x] `node scripts/check-unused-locals.mjs` — `packages/export` is in the documented "does not compile standalone" exception list (module-resolution errors against unbuilt workspace `dist/`, same baseline as every other file in the package); confirmed clean instead with `pnpm exec tsc --noEmit --noUnusedLocals -p packages/export` after `pnpm --filter @ifc-lite/export... build` — 0 errors.
- [x] `pnpm exec tsc --noEmit` (full monorepo `pnpm typecheck`) — clean, 1573/1573 test files typechecked.
- [x] `packages/export` full test suite: 1041 passed, 30 skipped, 0 failed (`pnpm exec vitest run`).
- [x] `scripts/api-surface.json` — no export added/removed/renamed from `packages/export/src/index.ts` (the new module is internal); not updated.
- [x] Changeset: `@ifc-lite/export` patch (`.changeset/fix-schema-convert-unrepresentable-entity.md`), matching the patch level used by the sibling merged-exporter fixes in this same lens (#3550).

## Related, not depended on or re-reported
Does not overlap #3550 (partial `IfcRelAggregates` redundancy), #3551 (context WCS gating), or #3552 (subcontext kind matching) — this is a distinct gap in the per-entity schema-conversion path they don't touch.


louistrue (approved):

Issue #3300, opened 2026-08-26
The STEP string escaper's encode half exists three times, and #3297's parity test freezes that instead of removing it


louistrue:
Falling out of #3284. That issue was about the two TypeScript escapers disagreeing on control-character runs. #3297 fixes the disagreement and adds a parity test to hold them together. The test is the right guard for one of the two pairs and the wrong answer for the other, so this is the follow-up rather than scope creep on that PR.

## The three sites

| | |
|---|---|
| `packages/data/src/step-serializers.ts` | `escapeStepString`, private, writes the HEADER |
| `packages/export/src/step-serialization.ts` | `escapeStepString`, exported, writes the DATA section |
| `rust/export/src/step_text.rs` | `escape` |

The two TypeScript ones are the same implementation. Verified by diffing the bodies: the only difference is that `export` pulled the directive loop out into `encodeNonAsciiStepDirectives`, whose body is identical to the loop `data` still has inline. Same three `.replace()` calls in the same order, same `\X2\`/`\X4\` thresholds, same padding.

There is a fourth, partial copy in `packages/encoding/src/ifc-string.ts` (`encodeIfcString`). It is **not** a drop-in and should not be folded in casually: it emits `\X\HH` across the whole Latin-1 range where both escapers emit `\X2\HHHH\X0\`, routes backslash into that arm as `\X\5C` rather than doubling it, and does no `''` doubling and no control-to-space mapping. Different output for the same input.

## Why the parity test is the wrong instrument for the TS pair

TS to Rust genuinely cannot share an implementation without a wasm adapter, so a pinned test is the deepest answer available there. TS to TS has no such excuse, and the seam already exists:

- `packages/export/package.json` declares `"@ifc-lite/data": "workspace:^"`. One way, no cycle.
- `packages/export/src/step-escaper-parity.test.ts` already imports `serializeValue` from `@ifc-lite/data` in order to reach the private escaper. The test that exists because the seam was not used is itself using the seam.
- The repo has already ruled on ownership. `packages/codegen/test/serialization-generator.test.ts:56` asserts generated bundles must not contain `function escapeStepString`, and `packages/codegen/src/serialization-generator.ts:9` says the serializer "lives once in `@ifc-lite/data`".
- All six call sites already go through `import { escapeStepString } from './step-serialization.js'`, so none of them change.

[...]

Issue #3284, opened 2026-08-25
Two more STEP-header disagreements between the halves: parseSourceHeader's searches are not quote-aware, and the escapers differ on control-character runs


BIMvoice:
Two smaller disagreements between the Rust and TypeScript STEP header paths, found alongside #3282 and deliberately left out of that PR's shared fixture (encoding them would have pinned a disagreement as agreement). Both are confirmed by running both halves, not by reading them.

Severity, stated honestly: both need a **constructed** input. I could not point at a real authoring tool that emits either shape, so this is robustness, not a reproduced field failure — filed because the first one is the same erasure #3282 fixes, reachable through a different door, and because `rust/export/src/step_text.rs::detect_schema` already carries a comment saying this exact class of raw-byte search is unsafe.

---

## 1. `parseSourceHeader`'s record searches are not quote-aware

`packages/parser/src/source-header.ts` locates `ENDSEC` and each `FILE_*` keyword with `toUpperCase().indexOf(...)`, which cannot tell a section terminator or a record keyword from the same text sitting inside a quoted header **value**.

### 1a — `ENDSEC` inside a quoted FILE_DESCRIPTION value

```
FILE_DESCRIPTION(('note: the ENDSEC; marker is described here'),'2;1');
FILE_NAME('a.ifc','ts',('Jane'),('Acme'),'pp','Revit','auth');
FILE_SCHEMA(('IFC4'));
```

TypeScript `parseSourceHeader`:

```
undefined
```

Rust `parse_source_header` (this PR's port, on the same bytes):

```
SourceHeader { description: ["note: the ENDSEC; marker is described here"], implementation_level: "2;1",
  name: Some("a.ifc"), time_stamp: Some("ts"), author: ["Jane"], organization: ["Acme"],
  preprocessor_version: Some("pp"), originating_system: Some("Revit"),
  authorization: Some("auth"), schema_identifiers: ["IFC4"] }
```

The header is truncated before any record is seen, so the TypeScript side returns `undefined` and `buildStepHeader` falls all the way back to defaults — the same provenance erasure #3282 is about, reached from the other side.

### 1b — the literal `FILE_NAME` inside the description text

```
FILE_DESCRIPTION(('per the FILE_NAME convention'),'2;1');
FILE_NAME('a.ifc','ts',('Jane'),('Acme'),'pp','Revit','auth');
```

TypeScript:

```
{"description":["per the FILE_NAME convention"],"implementationLevel":"2;1",
[...]

PR #3556, opened 2026-08-30
fix(core,encoding): map STEP's S escape through the selected P code page instead of hardcoding ISO 8859-1

BIMvoice:
## Scope

Follow-up to #3300 (the STEP string escaper's **encode** half), narrowed per that thread's own instruction to the **decode** side vs ISO 10303-21 §6.4.3 rather than re-reporting anything from #3378/#3403/#3444/#3454.

## What I found

Both decoders of STEP string literals:
- TS: `decodeIfcString` in `packages/encoding/src/ifc-string.ts` (used via `@ifc-lite/encoding` by `packages/parser/src/entity-extractor.ts`, `columnar-parser-attributes.ts`, and `@ifc-lite/data`'s `schema_gen`-equivalent path)
- Rust: `decode_ifc_string` in `rust/core/src/step_encoding.rs` (used by `AttributeValue::from_token`, the funnel every production parse path — CLI, server, wasm, SDK — goes through)

...already had a shared cross-language test-vector fixture at `rust/core/tests/fixtures/ifc_string_vectors.json` (consumed by `rust/core/tests/ifc_string_parity.rs` and `packages/encoding/src/ifc-string.parity.test.ts`), on the same pattern as the encode side's `step_escape_vectors.json`. On every existing vector, the two decoders already agreed with each other and with the spec — **no TS/Rust divergence** on the escapes those vectors covered.

Building hand vectors from ISO 10303-21 §6.4.3 turned up a real spec deviation shared by **both** decoders (same bug, not a TS/Rust disagreement):

`\S\C` decodes to the code point of `C` plus 128, and that resulting 0x80..0xFF byte is defined to be interpreted through whichever ISO 8859 part the most recent `\P?\` directive in the string selected (default: ISO 8859-1, letter `A`). Both decoders consumed and dropped every `\P?\` directive without tracking which page it selected — so `\S\` always just added 128 to the operand and used that as the Unicode code point directly. That's correct **only** for the default page (ISO 8859-1, where the upper half is the Unicode-identity mapping), and silently wrong for any other one. `\PE\\S\P` (ISO 8859-5, Cyrillic) decoded to U+00D0 (Ð, LATIN CAPITAL LETTER ETH) instead of the correct U+0430 (а, CYRILLIC SMALL LETTER A). Old ArchiCAD/Allplan-era files using a non-default code page hit this; files that never emit `\P?\` (the vast majority) or that only use `\X2\`/`\X4\` were unaffected.

Page-selector letters (A=ISO 8859-1 .. I=ISO 8859-9) and the per-page tables are verified against `IfcOpenShell`'s `IfcCharacterDecoder.cpp` (the letter-to-page mapping) and unicode.org's canonical `8859-{2..9}.TXT` mapping files (the exact code points), not reconstructed from memory.

## Fix

Both decoders now track the active code page across `\P?\` directives within a string (`A`..`I` select ISO 8859-1..9; any other letter is dropped without changing the page — matching IfcOpenShell's own `A`..`I` restriction) and map `\S\`'s result through the matching table. A byte position the selected page itself leaves unassigned falls back to the raw code point (same answer the default page gives) rather than `U+FFFD`, since the spec doesn't define decoder behaviour there — documented as a deliberate, non-spec-mandated choice, not inferred.

`\X\`, `\X2\`, `\X4\` are **not** code-page dependent (they're literal ISO 10646 values per the spec) and were already correct; only `\S\` needed this.

New vectors covering every page and the undefined-slot fallback were added to the shared fixture, so the two decoders can't drift on this again the way #3284 found for the encode side.

## Vector table (spec-built by hand, not copied from the decoders)

| input | TS `decodeIfcString` (before) | Rust `decode_ifc_string` (before) | spec-expected | after (both) |
|---|---|---|---|---|
| `\S\a` (no `\P`) | U+00E1 (á) | U+00E1 (á) | U+00E1 (á) | U+00E1 (á) — unchanged |
| `\PA\\S\!` | U+00A1 (¡) | U+00A1 (¡) | U+00A1 (¡) | U+00A1 (¡) — unchanged |
| `\PB\\S\!` (ISO 8859-2) | U+00A1 (¡) — wrong | U+00A1 (¡) — wrong | U+0104 (Ą) | U+0104 (Ą) |
| `\PE\\S\P` (ISO 8859-5, Cyrillic) | U+00D0 (Ð) — wrong | U+00D0 (Ð) — wrong | U+0430 (а) | U+0430 (а) |
| `\PE\\S\P\S\Q` (page persists) | U+00D0 U+00D1 — wrong | U+00D0 U+00D1 — wrong | U+0430 U+0431 (аб) | U+0430 U+0431 (аб) |
| `\PE\\S\P\PA\\S\P` (mid-string switch) | U+00D0 U+00D0 — wrong | U+00D0 U+00D0 — wrong | U+0430 U+00D0 (аÐ) | U+0430 U+00D0 (аÐ) |
| `\PF\\S\0` (ISO 8859-6, byte 0xB0 unassigned) | U+00B0 (°) | U+00B0 (°) | spec silent; we define identity fallback | U+00B0 (°) — unchanged (already matched the chosen fallback by construction) |
| `\PI\\S\P` (ISO 8859-9, Turkish) | U+00D0 (Ð) — wrong | U+00D0 (Ð) — wrong | U+011E (Ğ) | U+011E (Ğ) |
| `\X\E9`, `\X2\...`, `\X4\...` | correct | correct | correct (not page-dependent) | unchanged |
| `''`, `\\` | correct | correct | correct | unchanged |

TS and Rust were byte-identical on every vector both before and after — this is a joint spec deviation, not a TS/Rust parity bug, so there's nothing to "pick a side" on.

## RED/GREEN

- **RED (unmodified `main`)**: `\PE\\S\P` decodes to U+00D0 in both `decode_ifc_string` (Rust) and `decodeIfcString` (TS) — reproduced via `cargo test -p ifc-lite-core --lib step_encoding` / `pnpm --filter @ifc-lite/encoding test` before the fix, with the new vectors added but the tracking logic not yet applied (manually verified by reverting the logic change locally).
- **GREEN**: same vectors now pass on both sides; full command output below.
- **Control**: all pre-existing vectors (`s_extended_ascii`, `p_directive_dropped`, `x2_bmp`, `x4_astral`, etc.) still pass unchanged — the default page and all non-`\S\` escapes are untouched.

## Final checks

1. `node scripts/check-module-size.mjs` → `check-module-size: OK (2030 files measured, 306 allowlisted, 0 new over 400)`
2. `pnpm run check:vitest-timeout-audit` → `summary: 282 protected (own call/describe), 247 config-protected, 0 config-unknown, 16344 unprotected` (no new regex-with-parens issues)
3. `node scripts/check-test-wiring.mjs` → OK (48 packages, 49 gate scripts, 52 scripts/ test files)
4. `node scripts/check-unused-locals.mjs` → `packages/encoding` compiles standalone and is not in the failing-to-compile list; confirmed separately with `tsc --noEmit --noUnusedLocals` in `packages/encoding` — clean, no output
5. `pnpm exec tsc --noEmit` (repo-wide) fails only on pre-existing, unrelated `packages/codegen/generated/*` parse errors (stale generated files needing `pnpm generate`, untouched by this PR). Touched packages typecheck clean: `pnpm --filter @ifc-lite/parser typecheck` → OK (85 test files); `packages/encoding` build (`tsc`) → clean.
6. Touched packages' tests: `packages/encoding` → 109 passed (6 files); `packages/parser` → 950 passed, 2 skipped (85 files)
7. `cargo test -p ifc-lite-core` → 196 + 5 + 1 + 2 + 3 + 2 + 1 + 3 + 4 + 11 + 2 + 1 + 1 all `ok`, 0 failed; `cargo clippy -p ifc-lite-core --all-targets` → clean
8. `scripts/api-surface.json` — no export added/removed/renamed (`decodeIfcString`/`encodeIfcString` and `decode_ifc_string`/`encode_ifc_string` signatures unchanged; the new codepage table/helper are module-private on both sides), so no snapshot update needed
9. Changesets: `@ifc-lite/encoding` (patch, the direct fix) and `@ifc-lite/wasm` (patch, since `ifc_lite_core::decode_ifc_string` is bundled into the wasm binary and its output changes for wasm/browser consumers on the affected inputs)

## Not in scope

- Lowercase hex digits in `\X\`/`\X2\`/`\X4\` payloads: the spec says uppercase, but both decoders already accept lowercase identically (leniency, not a divergence) — left as-is.
- `\PS\` (custom/registered code page) and any letter outside `A`-`I`: both decoders drop the directive without changing the active page, matching `IfcOpenShell`'s own restriction to `A`-`I`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IFC string decoding for ISO 8859-1 through ISO 8859-9 code-page selections.
  * `\S\` escapes now correctly use the active code page and preserve page changes throughout a string.
  * Invalid or unassigned mappings now fall back safely to the original byte value.
  * Applied consistent decoding behavior across JavaScript and Rust implementations.

* **Tests**
  * Added coverage for code-page switching, persistence, defaults, and fallback behavior.


louistrue (approved):

PR #3561, opened 2026-08-31
fix(clash): stop a NaN-vertexed triangle from poisoning contact-BVH siblings

BIMvoice:
## Summary
- `unionAabb` (`packages/clash/src/contact/aabb.ts`) folded AABB bounds with `Math.min`/`Math.max`, which propagate a NaN operand through the whole comparison. `Bvh.build` (`packages/clash/src/contact/bvh.ts`) folds `unionAabb` bottom-up over every ancestor of a leaf, so a single degenerate (NaN-vertexed) triangle — e.g. from corrupt mesh geometry — poisoned the aggregate `bounds` of every node above it, up to and including the tree root.
- `queryMeshCross` (`mesh-bvh.ts`) treats a NaN-bounded node as "no overlap" and prunes it, so a poisoned root silently dropped the WHOLE traversal: every other, perfectly valid triangle in the mesh vanished from the query — not just the degenerate one.
- Consumer impact: this BVH backs `contactClusters`/`narrowPhase` (the on-demand contact-interface computation for a focused clash pair) and `minDistanceBetweenMeshes`/`minDistanceBetweenBvhs` (the viewer's minimum-distance measure tool) — a corrupt triangle in either mesh could make either feature silently report no contact / no valid distance for the entire pair.
- Fix: fold with NaN-safe `<`/`>` comparisons instead, so a NaN triangle is excluded from the aggregate rather than poisoning it. This matches `aabbFromPositions` in the very same file, and `compute_bounds` in `rust/clash/src/bvh.rs` (`AGENTS.md` already flags that Rust file as the correct template for this exact bug class, per #3547 — that PR fixes the same shape of bug in `packages/spatial/src/bvh.ts`'s BVH, a different file backing the main clash broad phase; this PR fixes the same shape of bug in `packages/clash/src/contact/aabb.ts`'s independent BVH, which backs only the contact-interface/min-distance path).

## How this was found
Working the clash-pipeline oracle lens (AGENTS.md): traced every BVH aggregate-bounds fold in the clash package and checked each against `rust/clash/src/bvh.rs`'s NaN-safe template. `packages/clash/src/contact/aabb.ts`'s `unionAabb` was the one outlier using `Math.min`/`Math.max`. Minimized repro (confirmed against the unmodified code before writing the fix):

```ts
// 12 axis-spaced triangles (> the tree's leafSize), one NaN-vertexed at index 0;
// a genuinely crossing triangle at a sibling leaf (index 5).
queryMeshCross(buildMeshBvh(meshWithOneNanTriangle, 4), buildMeshBvh(crossingTriangle, 4), 0);
// main:  []        (the valid crossing at index 5 is silently dropped)
// fixed: [[5, 0]]
```

## Test plan
- [x] RED confirmed against unmodified `unionAabb`: `queryMeshCross` returns `[]` instead of the real crossing pair (2 new regression tests in `mesh-bvh.test.ts`, both failing).
- [x] GREEN with the fix; full `@ifc-lite/clash` suite: 454 passed, 1 skipped (pre-existing, unrelated).
- [x] Unit tests for `unionAabb` itself in `aabb.test.ts` (wholly-NaN box, and per-axis NaN on each side).
- [x] `node scripts/check-module-size.mjs` — OK, 0 files over budget.
- [x] `pnpm run check:vitest-timeout-audit` — no regex-literal violations introduced.
- [x] `node scripts/check-test-wiring.mjs` — OK.
- [x] `node scripts/check-unused-locals.mjs` — `packages/clash` compiles clean and is not in the non-compiling-package list (several unrelated packages are, pre-existing).
- [x] `pnpm exec tsc --noEmit` in `packages/clash` — clean.
- [x] No public export surface changed (`unionAabb` was already exported from `contact/aabb.ts` but is not re-exported from the package's public `index.ts` or `contact/index.ts`) — `scripts/api-surface.json` untouched.
- [x] Changeset added (`@ifc-lite/clash`, patch).

## Siblings checked
Grepped `packages/clash/src` for every other place that folds AABB bounds with `Math.min`/`Math.max`: `math/generated/plato.g.ts`'s `unionAabb` (used by the main engine's `broad.ts` broad phase, via `@ifc-lite/spatial`'s `BVH`) has the identical shape, but that BVH is `packages/spatial/src/bvh.ts`, which is the exact file #3547 (open) already fixes. This PR is scoped to the separate, independent BVH under `packages/clash/src/contact/` that #3547 does not touch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mesh contact detection when geometry contains degenerate triangles with invalid coordinates.
  * Valid triangles are no longer skipped during contact clustering or minimum-distance queries because of nearby invalid geometry.
  * Added regression coverage for invalid triangles at different positions within a mesh.

* **Tests**
  * Added coverage for safe handling of invalid values when combining bounding boxes.

### DIFF
--- a/.changeset/fix-contact-bvh-nan-poison.md
+++ b/.changeset/fix-contact-bvh-nan-poison.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/clash": patch
+---
+
+Fix `unionAabb` (`contact/aabb.ts`) to fold bounds with NaN-safe comparisons instead of `Math.min`/`Math.max`. `Bvh.build` (`contact/bvh.ts`) folds `unionAabb` bottom-up over every ancestor of a leaf, so a single degenerate (NaN-vertexed) triangle — e.g. from corrupt mesh geometry — poisoned the aggregate bounds of every node above it, up to and including the tree root. `queryMeshCross` then treated the poisoned bounds as "no overlap" and pruned the whole subtree, silently dropping every other, valid triangle from contact-interface clustering (`contactClusters`) and minimum-distance queries (`minDistanceBetweenMeshes`). A NaN triangle is now simply excluded from the aggregate rather than poisoning it, matching `aabbFromPositions` in the same file and `compute_bounds` in `rust/clash/src/bvh.rs`.
--- a/packages/clash/src/contact/aabb.test.ts
+++ b/packages/clash/src/contact/aabb.test.ts
@@ -11,7 +11,7 @@
  */
 
 import { describe, expect, it } from "vitest";
-import { contains, intersects, longestAxis } from "./aabb.js";
+import { contains, intersects, longestAxis, unionAabb } from "./aabb.js";
 import type { AABB } from "./types.js";
 
 describe("intersects() — closed-interval face touch (aabb.ts:16)", () => {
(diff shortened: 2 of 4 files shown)


louistrue (approved):

PR #3562, opened 2026-08-31
fix(viewer): reconcile clash panel header count with the review-status filter

BIMvoice:
## Fix

The clash panel renders a filtered list, but its header only appended `N shown` when the “Hide touching” filter removed rows. Excluding review statuses could therefore leave the unfiltered total visible with no indication that fewer rows were on screen.

The header now appends `N shown` whenever `shown < total`, regardless of whether touching, review-status, or both filters caused the difference. The total and severity summary intentionally remain the unfiltered result summary; the annotation makes that distinction explicit.

## Regression coverage

The UI test marks one of two clashes resolved, disables `resolved`, verifies that only the open row remains, and requires `1 shown` in the header.

## Verification

- Current-base full GitHub CI is green on `016b71933` against `de9d10af`.
- Viewer tests, Node, typecheck, lint, E2E, canary, parity, and benchmark passed.
- Patch changeset for `@ifc-lite/viewer` is included.

### DIFF
--- a/.changeset/clash-panel-status-filter-count.md
+++ b/.changeset/clash-panel-status-filter-count.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/viewer': patch
+---
+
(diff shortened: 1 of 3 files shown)


louistrue (approved):

PR #3563, opened 2026-08-31
fix(lens): stop an absent property/attribute matching equals/contains ""

BIMvoice:
## Fix

`matchesProperty` and `matchesAttribute` evaluated `contains` and `equals` before their existing presence semantics. A missing value was coerced to `''`, causing an empty-string rule to match entities that did not have the property or attribute.

Both matchers now perform their presence check before every comparison, matching the already-correct quantity matcher:

- a missing property never satisfies `equals ''` or `contains ''`;
- an attribute whose value is `undefined` or `''` is absent, consistent with its existing `exists` semantics, and never satisfies a comparison;
- a property explicitly stored as `''` remains a present value and can satisfy `equals ''`.

## Regression coverage

The tests cover missing property/attribute values with both empty-string operators, the pre-existing quantity behavior, and the present-empty-property control.

## Verification

- Current-base full GitHub CI is green on `126c87c06` against `2edd144329`.
- In-tree parity, SDK canary, benchmark, typecheck, lint, Node, E2E, and viewer shards passed.
- Patch changeset for `@ifc-lite/lens` is included.

### DIFF
--- a/.changeset/lens-empty-value-absence-match.md
+++ b/.changeset/lens-empty-value-absence-match.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/lens': patch
+---
+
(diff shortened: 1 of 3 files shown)


louistrue (approved):

PR #3565, opened 2026-08-31
fix(codegen,parser): strip EXPRESS UNIQUE constraint before a collection's element type

BIMvoice:
## Fix

EXPRESS permits `UNIQUE` before a collection element type, for example `LIST [1:?] OF UNIQUE IfcGridAxis`. The parser now strips that element constraint before recursively classifying the type, including nested collections.

Generated IFC4, IFC4X3, and parser artifacts were regenerated canonically; the parser mirror is byte-identical to the IFC4 output.

## Verification

- Canonical freshness check passes for all four generated-output targets.
- Module-size gate passes.
- Full PR CI is green after rebasing onto current `main`.

### DIFF
--- a/.changeset/fix-codegen-unique-in-collection.md
+++ b/.changeset/fix-codegen-unique-in-collection.md
@@ -1,0 +1,36 @@
+---
+'@ifc-lite/codegen': patch
+---
+
+Fix the EXPRESS code generator dropping a `UNIQUE` collection constraint into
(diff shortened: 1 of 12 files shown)


louistrue (approved):
Re-reviewed after the current-main rebase: canonical generation, freshness gate, and full CI are green. Approved for squash merge.

PR #3566, opened 2026-08-31
ci(scripts): add check-codegen-sync — a freshness gate for codegen output

BIMvoice:
## Summary

This adds `check:codegen-sync`, a required Node-test CI gate that regenerates all committed schema outputs into temporary directories and fails closed on missing, extra, or byte-different files.

It covers `packages/codegen/generated/{ifc4,ifc4x3}`, the parser's generated IFC4 mirror, and `packages/data/src/ifc-schema/generated`. The check builds the real code generator, runs both real generator paths from their committed inputs, and never writes into the repository tree.

The corresponding detector tests prove content, missing-file, extra-file, nested-tree, absent-directory, and run-to-run determinism cases. This closes the gap exposed by #3565: generated schema drift is now caught rather than silently surviving a hand patch.

## Verification

- Full fresh GitHub CI on current `main` base: passed, including Build + WASM + Rust + Node.
- IfcOpenShell in-tree parity: passed.
- Generator freshness gate and its detector tests ran through the Node-test workflow.

No published API changes; no changeset is required.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1079,6 +1079,16 @@ jobs:
       # per-DIRECTORY, so it goes blind on the next subdirectory anyone adds.
       - name: Run every scripts/ test file (glob catch-all)
         run: node --test scripts/*.test.mjs scripts/lib/*.test.mjs scripts/fixtures/*.test.mjs scripts/docs/*.test.mjs
+      # Committed codegen output (packages/codegen/generated/**,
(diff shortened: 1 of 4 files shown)


louistrue (approved):
Reviewed current-base commit 3424dd33e: the generator freshness gate is wired into the workflow, fails closed on content/file-set drift, and has non-vacuous detector/determinism coverage. #3565 is on the base; awaiting the remaining fresh CI jobs before merge.

PR #3568, opened 2026-08-31
fix(export): serve Properties.parquet/Quantities.parquet through the on-demand accessor

BIMvoice:
## Summary

`ParquetExporter.writeProperties()`/`writeQuantities()` read `store.properties`/`store.quantities` directly. `IfcParser.parseColumnar` — the only parse path every real caller goes through (`packages/parser`'s `parseLite`) — never populates those bulk tables: it constructs a `PropertyTableBuilder`/`QuantityTableBuilder` but never calls `.add()` on either, serving properties and quantities lazily instead through `onDemandPropertyMap`/`onDemandQuantityMap` + `store.getProperties()`/`store.getQuantities()`.

Result: every model parsed the normal way exported a **zero-row `Properties.parquet` and `Quantities.parquet`**, silently — no error, no warning — sitting next to a fully populated `Entities.parquet`/`Relationships.parquet` in the same `.bos` archive.

Both writers now fall back to the on-demand path (through the same `store.getProperties()`/`store.getQuantities()` accessor every other consumer — the query package, the viewer's property panel — uses) when the bulk table is empty but on-demand data exists, and keep the existing bulk-table path unchanged for stores that populate it directly (this package's own hand-built test fixtures, and any external caller that assembles a store that way).

## Independent-reader verification (self round-trip is blind to this class of bug)

Installed `duckdb` in a scratch dir (`/tmp/pq`, not a repo dependency) and read the exported `.bos` archive's Parquet files with it directly — an independent Rust/C++ implementation, not our own `parquet-wasm` reader.

Fixture: `tests/models/ara3d/duplex.ifc` (IFC2X3, 2,960 `IFCPROPERTYSET`/`IFCRELDEFINESBYPROPERTIES`/`IFCELEMENTQUANTITY` lines in source).

| Check | Unmodified `main` (RED) | This branch (GREEN) |
|---|---|---|
| `Properties.parquet` row count (DuckDB) | 0 | 13,434 |
| `Quantities.parquet` row count (DuckDB) | 0 | 21 |
| Sample row `IsExternal` (Boolean) | — | reads back as SQL `false`/`true`, not stringified |
| Sample row `Reference` (String) | — | `"Basic Wall:Exterior - Brick on Block"` |
| `Entities.parquet` / `Relationships.parquet` (control — unaffected by this bug) | 1080 / 3200 rows | 1080 / 3200 rows (unchanged) |

Confirmed on two more fixtures, same pattern (bulk table empty on main, populated after the fix; entities/relationships identical throughout):

| Fixture | Properties (RED → GREEN) | Quantities (RED → GREEN) |
|---|---|---|
| `ara3d/AC20-FZK-Haus.ifc` (IFC4) | 0 → 4,431 | 0 → 3,284 |
| `AB22.ifc` (IFC4X3_RC4) | 0 → 336 | 0 → 0 (this fixture legitimately carries no quantities) |

The DuckDB reader stays a scratch script (not a repo dependency); the four new unit tests below use a hand-built mock store shaped like the real on-demand parse output (empty bulk tables + `onDemandPropertyMap`/`onDemandQuantityMap`) and assert through the existing `parquet-wasm` decode helper already used by this test file, which is enough to pin the regression against our own tooling.

## Changes

- `packages/export/src/parquet-exporter.ts`: `writeProperties`/`writeQuantities` check whether the bulk table is actually populated; if not (and on-demand data exists) they delegate to the new on-demand writers instead of silently emitting zero rows.
- `packages/export/src/parquet-exporter-ondemand.ts` (new): the on-demand `Properties.parquet`/`Quantities.parquet` writers, walking `onDemandPropertyMap`/`onDemandQuantityMap` through `store.getProperties()`/`store.getQuantities()`. `Quantities.Formula` is always NULL on this path — the on-demand quantity reader (`readQuantitySet`) reports only `{name, type, value}`, no formula string; only the bulk-table path can populate it.
- `packages/export/src/parquet-type-strings.ts` / `parquet-uint32-columns.ts` (new): the `PropertyValueType`/`QuantityType` → string tables and the u32-column-name set, extracted so the on-demand module and `parquet-exporter.ts` share one copy instead of two that could disagree. `QuantityTypeToString` stays re-exported from `parquet-exporter.ts` under its original name (unused in-repo, not part of the package's `index.ts` surface, but was public `export function` before the split).
- `packages/export/src/parquet-exporter.test.ts`: four new tests build a mock store shaped like the real on-demand parse path and pin the on-demand `Properties`/`Quantities` writers, the control (`Entities`/`Relationships` on the ordinary bulk path), and overlay-deletion behaviour on the on-demand path.
- Extracting the on-demand writers to sibling modules also keeps `parquet-exporter.ts` under its `check-module-size.mjs` budget (would otherwise land at 758 lines against a 639 budget).

## Test plan

- [x] `pnpm --filter @ifc-lite/export test` — 1066/1066 pass (1062 pre-existing + 4 new)
- [x] `pnpm --filter @ifc-lite/export build` / `tsc --noEmit` — clean
- [x] Independent DuckDB read of the exported `.bos` archive on 3 real fixtures — RED confirmed on unmodified `main`, GREEN confirmed on this branch, control (`Entities`/`Relationships`) unchanged
- [x] `node scripts/check-module-size.mjs` — OK (`parquet-exporter.ts` at 610/639 lines)
- [x] `pnpm run check:vitest-timeout-audit` — clean (no regex-literal-with-parens issues)
- [x] `node scripts/check-unused-locals.mjs` — `packages/export` compiles standalone and is clean; the packages it lists as non-compiling (viewer, cli, sdk, etc.) are pre-existing and untouched by this change
- [x] Changeset added (`patch`, `@ifc-lite/export`) — no public API surface change (`index.ts` untouched, `ParquetExporter`'s public constructor/`exportBOS`/`exportTable` signatures unchanged)

## Twin check: does the server/Rust path have the same defect?

No. Traced both Rust parquet writers' data flow:

- **`rust/export/src/parquet_bos.rs`** (`export_bos`, the native/server `.bos` writer this PR's TS fix mirrors): calls `build_export_model` (`rust/export/src/model.rs`), whose Pass 2 resolves `resolve_pset_defs(&mut decoder, &def_ids)` into `EntityRow.property_sets`/`.quantity_sets` **eagerly, per IfcProduct row, during model construction** — not deferred to an on-demand accessor the writer might skip. `properties_table`/`quantities_table` then just iterate `model.entities[].property_sets`/`.quantity_sets`, which are the same fields Pass 2 just filled in.
- **`apps/server/src/services/parquet_data_model.rs`** (the server's HTTP-streamed `DataModel` writer): `serialize_properties_table(&data_model.property_sets)` / `serialize_quantities_table(&data_model.quantity_sets)` read `DataModel.property_sets`/`.quantity_sets` (`apps/server/src/services/data_model/types.rs`), populated by `extract_data_model` (`apps/server/src/services/data_model/mod.rs:132`) via `rayon::join`, calling into `properties.rs`/`quantities.rs` unconditionally as part of building the `DataModel` — same shape: eager, not lazy.

Confirmed by running the export, not just reading the source: temporarily added a `#[test]` to `parquet_bos.rs` calling `export_bos` on `tests/models/ara3d/duplex.ifc` (same fixture as the TS RED/GREEN above) and dumping the `.bos` bytes, then read it independently with DuckDB (removed before commit — no diff left in `rust/`):

| Table | Rust `export_bos` on `duplex.ifc` |
|---|---|
| `Properties.parquet` row count | 13,434 |
| `Quantities.parquet` row count | 21 |
| Sample row | `Pset_WallCommon.Reference = "Basic Wall:Interior - Partition (92mm Stud)"`, `LoadBearing = "false"` |

Matches the TS fix's GREEN counts on the same fixture (13,434 / 21) — the Rust twin was never affected because, unlike `IfcParser.parseColumnar`'s `parseLite` in the TS parser, both Rust parse paths (`build_export_model`, `extract_data_model`) populate their property/quantity fields eagerly at model-construction time, so the parquet writers read data that is actually there. No code change needed on the Rust side.

(`Entities.parquet` reports 295 rows for `export_bos` vs. entity counts reported elsewhere in this PR for the TS path — `export_bos`'s Pass 2 only emits `IfcProduct`-subtype rows, a documented scope difference, not related to this defect.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Parquet exports that produced empty `Properties.parquet` and `Quantities.parquet` files for data loaded on demand.
  - Property and quantity records are now included correctly, including numeric, boolean, and bounded-value data.
  - Deletions are respected when exporting on-demand data.
  - Improved handling of large IFC identifiers so they retain correct unsigned values in exported Parquet files.

## Follow-up correction

- Non-scalar REAL values from the on-demand accessor (including `IfcPropertyBoundedValue`'s display-string representation) preserve `ValueString` and emit NULL in `ValueReal`; only scalar numeric REAL values populate `ValueReal`.
- Regression coverage asserts both the bounded REAL NULL case and the scalar REAL numeric case.

### DIFF
--- a/.changeset/parquet-properties-quantities-on-demand.md
+++ b/.changeset/parquet-properties-quantities-on-demand.md
@@ -1,0 +1,7 @@
+---
+'@ifc-lite/export': patch
+---
+
+`ParquetExporter.exportBOS()`/`exportTable('properties'|'quantities')` wrote a zero-row `Properties.parquet` and `Quantities.parquet` for every model parsed the normal way — `Entities.parquet` and `Relationships.parquet` alongside them fully populated. `IfcParser.parseColumnar` (the only parse path used by every real caller) never bulk-populates `store.properties`/`store.quantities`; it serves them lazily through `onDemandPropertyMap`/`onDemandQuantityMap` and `store.getProperties()`/`store.getQuantities()` instead. The two writers read the never-populated bulk tables directly, so the gap was silent: no error, no warning, just an empty table next to full ones. Verified independently — DuckDB opened the exported `.bos` archive and read back 0 rows from both tables against a real fixture carrying thousands of property/quantity relationships.
+
+Both writers now fall back to the on-demand path (through the same `store.getProperties()`/`store.getQuantities()` accessor every other consumer uses) when the bulk table is empty but on-demand data exists, and keep the existing bulk-table path for stores that populate it directly. `Quantities.Formula` is NULL on the on-demand path — the on-demand quantity reader carries no formula string, only the bulk-table path can populate it.
--- a/packages/export/src/parquet-exporter-ondemand.ts
+++ b/packages/export/src/parquet-exporter-ondemand.ts
@@ -1,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * `Properties.parquet` / `Quantities.parquet` writers for the on-demand parse
+ * path, extracted from `ParquetExporter` (`parquet-exporter.ts`) to keep that
+ * file under its module-size budget.
+ *
+ * `IfcParser.parseColumnar` (`packages/parser`'s only parse path, used by
+ * every real caller) never calls `.add()` on the `PropertyTableBuilder`/
+ * `QuantityTableBuilder` it constructs — properties and quantities are served
+ * lazily through `onDemandPropertyMap`/`onDemandQuantityMap` +
+ * `store.getProperties()`/`store.getQuantities()` instead
+ * (`extractPropertiesOnDemand`/`extractQuantitiesOnDemand`, re-parsing the
+ * source buffer on access). `ParquetExporter`'s bulk-table writers read
+ * `store.properties`/`store.quantities` directly, which is only populated
+ * when a caller builds a store some other way (e.g. this package's own
+ * hand-built test fixtures) — every store from a real parse left those two
+ * tables at zero rows, silently, alongside a fully-populated
+ * `Entities.parquet`/`Relationships.parquet`. These two functions are the
+ * on-demand fallback `ParquetExporter` calls when the bulk table is empty:
+ * one flat row per property/quantity, read through the same
+ * `store.getProperties()`/`store.getQuantities()` accessor every other
+ * consumer (the query package, the viewer's property panel) uses — never a
+ * second copy of the pset/property walk.
+ */
+
+import type { IfcDataStore } from '@ifc-lite/parser';
+import { PropertyValueType, type QuantityType } from '@ifc-lite/data';
+import type { EffectiveEntityIndex } from './effective-index.js';
+import { columnsToParquet } from './columns-to-parquet.js';
+import { PARQUET_UINT32_COLUMNS } from './parquet-uint32-columns.js';
+import { propertyValueTypeToString, quantityTypeToString } from './parquet-type-strings.js';
+
+export async function writePropertiesOnDemand(
+    store: IfcDataStore,
+    effective: EffectiveEntityIndex | null,
+): Promise<Uint8Array> {
+    const entityIdCol: number[] = [];
+    const psetNameCol: string[] = [];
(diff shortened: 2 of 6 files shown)


louistrue (approved):
Final local diff/body review complete; full CI is green. Approved for squash merge.

PR #3571, opened 2026-08-31
fix(server-client): stop fabricating 0 for a null elevation/thickness column

BIMvoice:
## Summary

Scope: the MIRROR direction of #3568 — our TS **reader** (`packages/server-client`) parsing Parquet the server sends, evaluated against an independent writer (DuckDB), not our own server's output.

`decodeDataModel` (`packages/server-client/src/data-model-decoder.ts`) decodes the server's data-model Parquet payload via `parquet-wasm` + `apache-arrow`, then extracts columns with `arrow.Vector.toArray()` for bulk performance. Two columns the server emits are **nullable `Float64`**:

- `elevation` on the spatial-nodes table — `None` when a storey's elevation cannot be resolved (`apps/server/src/services/data_model/spatial.rs`, `returns_none_when_neither_elevation_nor_placement_resolves`).
- `thickness` on the materials table — `None` for every **non-layer** material association: a single material, a material list, a material constituent (`apps/server/src/services/data_model/materials.rs`; layers are the exception).

`Vector.toArray()` on a nullable numeric column returns the column's raw values buffer directly — its fast path for a primitive type — and that buffer carries whatever bits sit at a NULL row's slot: `0` in practice, for both DuckDB and `parquet-wasm`'s own writer. It is **not** a detectable sentinel; the decoder's existing `?? undefined` guard is a no-op against a real `0`. So a storey with no resolvable elevation, or the very common case of a non-layer material, silently decoded with a fabricated `elevation: 0` / `thickness: 0` — a real-looking, wrong measurement — even though downstream viewer code already checks `!== undefined` for both fields (`apps/viewer/src/components/viewer/properties/MaterialCard.tsx`, `apps/viewer/src/components/viewer/BulkPropertyEditor.tsx`), which is the direct evidence the contract was already assumed and silently violated.

## Encoding-matrix note

The actual Parquet **binary** decode (footer/thrift/page/dictionary/RLE/compression) is delegated entirely to `parquet-wasm` — a mature independent Rust parquet reader compiled to wasm — so that axis is not this repo's bug surface. The bug lives one layer up, in how `server-client`'s own column-extraction code (`data-model-decoder.ts`, `parquet-tables.ts`, `parquet-columns.ts`) turns Arrow columns into JS values. Auditing that layer found this one live defect: `.toArray()` vs `.get(i)` on a NULLABLE numeric column. (Non-numeric nullable columns are unaffected — Arrow's string/bool `.toArray()` already preserves `null`; `is_ventilated` in the same materials table already used `.get(i)` correctly, which is what made the `thickness` sibling's omission visible.)

## Fix

`packages/server-client/src/parquet-nullable.ts` (new, 40 lines): `nullableFloat64Column(table, name)` takes the bulk `toArray()` fast path when `column.nullCount === 0`, and only falls back to per-row `.get(i)` validity checks when the column actually carries a null.

`data-model-decoder.ts`: `elevation` and `thickness` extraction now go through this helper (2-line change; `data-model-decoder.ts` is pinned at 775 lines in the module-size allowlist, so the fix lives in a sibling module rather than growing it).

## Test plan

- [x] RED confirmed on unmodified `data-model-decoder.ts`: both new assertions fail with `expected +0 to be undefined` (verified by temporarily reverting the two extraction lines and re-running).
- [x] GREEN with the fix: `pnpm --filter @ifc-lite/server-client test` — 9/9 in `data-model-decoder.decode.test.ts`, 53/53 package-wide.
- [x] Control: the existing truncation/bounds-checking tests in the same file are untouched and still pass.
- [x] Fixtures (`packages/server-client/src/__fixtures__/nodes-nullable-elevation.parquet`, `materials-nullable-thickness.parquet`, ~1KB each) are **DuckDB-authored**, not generated by this repo's own code — the locked `parquet-wasm@0.5.0`'s `writeParquet`, called via this test file's own `arrow.Table -> IPC -> Table.fromIPCStream -> writeParquet` helper, silently drops null-ness for a nullable Float64 column (confirmed independently: DuckDB reading the write-path's own output back sees a real `0`, not a null — a write-side bug in the pinned wasm version, orthogonal to the read-side bug this PR fixes). The DuckDB SQL used to generate each fixture is in a comment above `readFixture` in the test file, and DuckDB itself was installed only in `/tmp`, never as a repo dependency.
- [x] `node scripts/check-module-size.mjs` → `check-module-size: OK (2031 files measured, 306 allowlisted, 0 new over 400)`
- [x] `pnpm run check:vitest-timeout-audit` → `summary: 282 protected (own call/describe), 247 config-protected, 0 config-unknown, 16344 unprotected` (unchanged)
- [x] `node scripts/check-test-wiring.mjs` → OK
- [x] `node scripts/check-unused-locals.mjs` → `packages/server-client` compiles standalone and is clean (not listed among non-compiling packages); remaining non-compiling packages are pre-existing and untouched by this PR (partial local install)
- [x] `pnpm exec tsc --noEmit --noUnusedLocals -p packages/server-client/tsconfig.json` → clean except one pre-existing, untouched `TS6196` in `parquet-decoder.ts`
- [x] `node ../../scripts/typecheck-tests.mjs` (server-client) → OK
- [x] No public API surface change (`nullableFloat64Column` is not re-exported from `index.ts`); changeset added as `patch` for the runtime-behavior fix.

## Follow-up requested: is the `parquet-wasm@0.5.0` write-side null-drop present in `packages/export`'s `^0.7.2`?

**No — verdict: not reproducible on 0.7.2.** `packages/export`'s `columnsToParquet.ts` writes through the identical pattern I found broken on 0.5.0 (`arrow.vectorFromArray` → `new arrow.Table(...)` → `arrow.tableToIPC(table, 'stream')` → `parquet.Table.fromIPCStream(ipc)` → `parquet.writeParquet(wasmTable)`), so I re-ran the exact repro against the exact version `packages/export` locks (`0.7.2`, confirmed in `pnpm-lock.yaml`'s `packages/export:` importer block):

```
source vector nullCount BEFORE write 1 [ 3, null, 0 ]
wrote elevation_check_072.parquet, bytes: 509
parquet-wasm 0.7.2 read-back: nullCount 1 values [ 3, null, 0 ]
duckdb (independent reader) sees: [ { elevation: 3 }, { elevation: null }, { elevation: 0 } ]
```

Both parquet-wasm's own read-back **and** an independent DuckDB read of the raw bytes agree: the null survives the write. On the locked `0.5.0` the same script's DuckDB read of the written file shows `{ elevation: 0 }` for that row — the null is gone at the byte level, not just misread later. So this is a `parquet-wasm@0.5.0`-specific write regression, not a `columnsToParquet.ts` defect, and it does not reach `packages/export`, which pins `^0.7.2`.

Separately, and independent of the parquet-wasm version: `writeProperties()`/`writeQuantities()` build the `ValueReal`/`Value` columns from `Array.from(properties.valueReal)` / `Array.from(quantities.value)` — dense `Float64Array`-backed store columns that hold no JS `null`/`undefined` to begin with (a property/quantity without a real value never reaches these arrays as a null; the store's own representation is already scalar). So `#3568`'s newly-populated `Properties.parquet`/`Quantities.parquet` were never exposed to this null/0 collision in the first place, on either parquet-wasm version. **No live defect found in `packages/export`; no code change made here.**

## Follow-up requested: should `server-client`'s `parquet-wasm@0.5.0` pin be upgraded?

**Recommended as a follow-up, but it is not a version bump alone — there is a real blocker.**

`server-client`'s own code hard-imports the subpath `parquet-wasm/esm/arrow2.js` (`packages/server-client/src/parquet-decoder.ts`, and the test harness in `data-model-decoder.decode.test.ts` via `require.resolve('parquet-wasm/esm/arrow2.js')`). That subpath is `0.5.0`'s module layout (an older two-arrow-build split, `arrow1.js`/`arrow2.js`); `0.7.2` ships a single flat `esm/parquet_wasm.js` and has no `arrow2.js` export at all — I confirmed importing `'parquet-wasm/esm/arrow2.js'` against a `0.7.2` install throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. So raising the peerDependency range to `>=0.7.2` (or pinning it there) would not, by itself, move the runtime: it needs the import path migrated in `parquet-decoder.ts`, the ambient module declaration in `vendor-types.d.ts`, and the test file's `require.resolve` call, plus re-verifying the WASM init call signature (`0.5.0`'s ESM build needs its default export awaited with a URL/bytes; `0.7.2`'s init surface wasn't audited here).

Worth noting as context for why this isn't purely cosmetic: `pnpm-lock.yaml` resolves `packages/server-client`'s own `parquet-wasm` peer to `0.5.0` (`packages/server-client: dependencies: parquet-wasm 0.5.0`) **independently** of `apps/viewer`'s direct `0.7.2` dependency (`apps/viewer: dependencies: parquet-wasm 0.7.2`) — confirmed by `require.resolve` executed from inside `packages/server-client`. Because `@ifc-lite/server-client` is consumed by `apps/viewer` as a `workspace:^` `link:` dependency (a real symlink to `packages/server-client`, not a hoisted flat install), Node/bundler module resolution for `parquet-decoder.ts`'s `import('parquet-wasm/esm/arrow2.js')` starts at that file's own on-disk location and walks up through `packages/server-client/node_modules` before it would ever reach `apps/viewer/node_modules` — so `apps/viewer`'s declared `^0.7.2` does not automatically satisfy `server-client`'s own hard-coded import. I have **not** run the actual browser bundle to confirm this end-to-end (that would need a real `apps/viewer` build/runtime trace), so I'm stating this as inference from the lockfile split plus Node's resolution algorithm, not an observed production trace — but it's consistent with everything measured above, and would mean the production viewer's data-model decode path runs against the older, write-buggy `0.5.0` today. This doesn't affect this PR's fix (a read-side fix; the write bug only matters to a *writer*, i.e. the Rust server or DuckDB, never this TS reader), so no code change is needed here — flagging it as the blocker analysis requested, and as a candidate follow-up PR (migrate off the `arrow2.js` subpath, bump the pin, and add a check that a workspace-consumed peer's resolved version actually matches what the consuming app declares).

## Summary by CodeRabbit

* **Bug Fixes**
  * Nullable elevation and material thickness values now decode as `undefined` instead of incorrectly appearing as `0`.
  * Genuine zero values remain distinguishable from missing or unresolved values.

* **Tests**
  * Added coverage for nullable numeric values in spatial node and material data.


louistrue (approved):

PR #3590, opened 2026-09-01
test(export): pin the CSV writer's row assembly against a real adversarial IFC file

BIMvoice:
## Summary

RFC 4180 / CSV-writer audit (per the ongoing hunt into #3100/#3102/#2303/#3115 territory). Result: **valid negative** — the current escaper and every writer built on it already quote/escape correctly and neutralize CWE-1236 formula injection. This PR closes one real test gap instead of a defect: `rust/export/src/csv.rs`'s own tests only ever called `escape()` directly on a hand-picked literal, never exercised the real `entities_csv`/`properties_csv` row assembly against bytes decoded from an actual IFC file.

## What was checked (field → RFC 4180 → verdict)

| Field shape | RFC 4180 requirement | Writer verdict |
|---|---|---|
| delimiter (`,`) in value | quote the field | quoted |
| double-quote in value | quote + double the `"` | quoted, doubled |
| embedded LF / CR / CRLF | quote the field | quoted (all three, cell-level tests) |
| leading/trailing space | kept, not stripped (§2.4) | kept (default); optionally also quoted via `quoteWhitespacePadded` |
| non-ASCII (Cyrillic, #3556) | pass through | survives the real STEP `\X2\...\X0\` decode -> CSV round trip |
| `=`/`+`/`-`/`@`/TAB/CR-led value (CWE-1236) | not a CSV requirement, but a real security issue | apostrophe-prefixed, looks past BOM/ZWSP/LRM/NBSP/U+2028 invisibles rather than being fooled by them |
| numeric value | must stay summable | `fmt_num` avoids scientific notation in Rust writers; the guard exempts genuine numbers from the `+`/`-` trigger without weakening it |
| null vs empty | inherent CSV limitation | both render `""` — not specific to this writer, no CSV-level fix possible |

Verified every TS/Rust CSV-producing path (SDK `bim.export.csv`, CLI `export --format csv` incl. the custom-`--columns` path, MCP `export_csv`, viewer Lists/Zones/Search/Compare exports, the Rust `entities|properties|quantities|spatial` views) funnels through the single `escapeCsvCell`/`escape_csv_cell` escaper — no hand-rolled writer bypasses it.

## The gap this PR closes

`rust/export/src/csv.rs`'s tests exercised the escaper in isolation but never the full pipeline. Added `adversarial_row_survives_the_real_export_pipeline`: a wall named `Wall "A", B` (quote+comma), a formula-injection description, and an `IfcPropertySingleValue` with a **Cyrillic name** (`\X2\...\X0\`-encoded, #3556) whose value combines a formula trigger, a comma, AND a quote in one cell (`=1+1,"x"`) — run through the real `export_csv(...)` and decoded back with an independent, hand-rolled RFC-4180 splitter (deliberately not sharing code with `escape_csv_cell`, and not the `csv-parse` npm package, kept out of the repo per house rules).

Independently spot-checked the same two lines with `csv-parse` in scratch (not committed):

```
entities row:   { name: 'Wall "A", B', description: "'=SUM(A1)", ... }
properties row: { propName: 'Ставка', value: '\'=1+1,"x"', ... }
```

Both match the source values exactly (guard-prefix stripped).

**Confirmed the test isn't vacuous**: temporarily disabling the writer's quoting condition made it fail with `wall row must have exactly as many fields as the header: left: 8, right: 7` — the exact corruption-class failure this hunt was looking for — before reverting to green.

No production code changed.

## Test plan
- [x] `cargo test --lib` in `rust/export` — 351 passed, 0 failed (was 350 before this PR)
- [x] RED verified by temporarily neutering `escape_csv_cell`'s quoting condition, then reverted (not committed)
- [x] Independently cross-checked the writer's output against `csv-parse` in `/tmp` scratch (not part of the repo)
- [x] `node scripts/check-module-size.mjs` — OK (2044 files measured, 305 allowlisted, 0 new over 400)
- [x] `pnpm run check:vitest-timeout-audit` — clean


louistrue (approved):
Independently reviewed: current-main test-only integration coverage exercises real IFC-to-CSV behavior without source-text assertions; terminal CI is green.

Issue #1031, opened 2026-06-10
collab: structured entity branches (psets/quantities/classifications/materials/geometryRef) are not serialized by the IFCX snapshot pipeline


louistrue:
## Context

Surfaced by CodeRabbit on PR #1027 (https://github.com/LTplus-AG/ifc-lite/pull/1027#discussion_r3385637634).

`@ifc-lite/collab` entities carry structured branches beyond `attributes` / `children` / `inherits`: `psets`, `quantities`, `classifications`, `materials`, `geometryRef`, `meta` (see `entityToJSON` in `packages/collab/src/doc/entity.ts`). The IFCX wire type `IfcxNode` has no fields for these, so **neither** snapshot writer carries them:

- `snapshotToIfcx` (`packages/collab/src/snapshot/to-ifcx.ts`) — the full-snapshot writer, predating PR #1027 — only emits `attributes` / `children` / `inherits`.
- `extractMinimalLayer` (`packages/collab/src/snapshot/minimal-layer.ts`) diffs the same three branches, deliberately matching the full writer's coverage so that baseline + minimal layer composes to the same IFCX state as a full snapshot.

This is a pre-existing limitation of the IFCX serialization scope, not a regression from the layer-PRs work. Today, property sets that need to travel through layers are written as namespaced attribute keys (e.g. `bsi::ifc::v5a::Pset_FireSafety::FireRating`), which is the convention the MCP draft tools, the merge engine (`pset:<Name>` component keys), and scope verification all share. Structured-branch edits made via collab APIs (`setPsetProp`, quantities, classifications, materials, geometry refs) round-trip inside the Y.Doc but are dropped on any IFCX export.

## What needs to happen

Decide on and implement one of:

1. **Serialize structured branches into namespaced attributes** at the snapshot boundary (e.g. fold `psets` into `…::Pset_X::Prop` attribute keys on write, and re-inflate on `seedFromIfcx`), keeping `IfcxNode` unchanged; or
2. **Extend the wire format** with explicit branches and update composition, canonical hashing, merge component-state extraction, and seeding accordingly.

Whichever is chosen, `snapshotToIfcx` and `extractMinimalLayer` must stay in lockstep so the layer-composition contract (baseline + minimal layer == live doc) keeps holding, and `computeLayerId` canonicalization must cover the new representation.

## Acceptance criteria

- Structured pset/quantity/classification/material/geometryRef edits survive snapshot → seed round-trips.
[...]

PR #3608, opened 2026-09-01
fix(export): carry classification through the IFC5 (IFCX) export

BIMvoice:
## Summary (updated after review)

**Original gap:** `Ifc5Exporter.export()` (STEP → IFCX/IFC5) wrote `bsi::ifc::class`, `bsi::ifc::prop::Name`/`Description`, property-set values, and mesh/colour for every emitted entity — but never read `IfcClassificationReference` / `IfcRelAssociatesClassification` at all. A classified element silently lost its classification on export to IFCX.

**Review caught an incomplete fix:** the first version of this PR made `Ifc5Exporter` write classification under `ifclite::classifications`, but `packages/ifcx/src/property-extractor.ts` — the primary IFCX reader — treats every `ifclite::*` key as an internal-only carrier (#1031) and skips it. So a plain re-import through `parseIfcx` (not going through `@ifc-lite/collab`) still lost the classification, and the changeset's claim that it "reads back the same way whether the IFCX came from this exporter or from a collab snapshot" was false for `packages/ifcx` itself. A foreign IFCX tool reading a private `ifclite::` namespace was never going to see it either way (no `bsi::ifc::classification` exists in the v5a schema to use instead — the root difficulty called out in review).

**Fix (option a — genuine round-trip through `packages/ifcx`):**
- `packages/export/src/ifc5-classification.ts` (new sibling, keeps `ifc5-exporter.ts` under its module-size budget): resolves an entity's classifications via the parser's existing `extractClassificationsOnDemand` and writes `ifclite::classifications` (`{ system, code, uri?, description? }[]`) — kept as the existing internal key rather than inventing a fake `bsi::ifc::classification` spec attribute.
- `packages/ifcx/src/property-extractor.ts`: carved `ifclite::classifications` specifically out of the blanket `ifclite::` skip and unpacks each ref into a `Classification - <system>` pset (`Code`/`Uri`/`Description`), the same pattern already used to unpack `bsi::ifc::material` into a `Material` pset. Every other `ifclite::*` key (`deleted`, `derived`, `materials`, `geometryRef`, `meta`) is still skipped, unchanged.
- `classification-roundtrip.test.ts` now does a genuine write-then-read round trip: parses a real STEP fixture (wall classified `Uniclass 2015` / `Pr_20_93_47`), exports to IFCX, **re-imports the output through `@ifc-lite/ifcx`'s `parseIfcx`**, and resolves the classification through `parsed.properties.getForEntity()` — the same `PropertyTable` path every other property comes back through. The previous version only asserted on the exporter's raw JSON and could not have caught the reader-side drop.
- `property-extractor.test.ts` gained a test pinning the new per-system pset unpacking, and a test confirming a codeless ref surfaces nothing; the old "skips ifclite:: carrier attributes entirely" test was split so it no longer asserts the (now-wrong) claim that classifications are skipped.

## What still doesn't round-trip (documented, not fixed here)

`extractClassificationsOnDemand` needs the source STEP buffer, so an already-IFCX-sourced `IfcDataStore` (round-tripping a `.ifcx` a second time through `Ifc5Exporter`) resolves to no classifications — nothing to lose there since that input never carried one to begin with, but it does mean the classification isn't re-derived from the newly-unpacked `Classification - <system>` pset on a second export; it would need to come from the raw `ifclite::classifications` attribute the composed node still carries. Not exercised by this PR's test and left as a known limitation.

Also unfixed, noted in the original investigation and kept out of scope (one fix per finding): `classification-extractor.ts`'s public `extractClassifications()` reads `IfcClassification.EditionDate` at a slot that's an `IfcCalendarDate` entity reference in IFC2X3, and its untyped `getString()` silently coerces the reference id into a garbage string instead of `undefined`. Dead internally (only its own tests call it), but public SDK API.

## Conflict note

**This PR and open #3605 (material export) both edit `packages/export/src/ifc5-exporter.ts`.** Whichever merges second will need a rebase — flagging so neither lands assuming a clean apply.

## Test plan
- [x] `pnpm --filter @ifc-lite/export test` — `classification-roundtrip`, `ifc5-exporter`, `ifcx-roundtrip` all pass; full package suite 1045 passed / 30 skipped (fixture-gated) / 1 unrelated pre-existing failure (`lod1-generator.test.ts`, missing `@ifc-lite/wasm` build artifact, not touched here)
- [x] `pnpm --filter @ifc-lite/ifcx test` — 188/188 pass, including the updated `property-extractor.test.ts` and the untouched `tombstones.test.ts`/collab-adjacent tests that read `ifclite::classifications` off raw attributes directly
- [x] `pnpm exec tsc --noEmit --noUnusedLocals` in both `packages/export` and `packages/ifcx` — clean for the touched files (pre-existing unused-locals findings in unrelated `packages/ifcx` files — `index.ts`, `path-resolver.ts`, `writer.ts`, `federated-composition.ts` — are untouched by this change and not introduced by it)
- [x] `node scripts/check-module-size.mjs` — OK (`ifc5-exporter.ts: 887/888`; `property-extractor.ts: 331` lines, no allowlisting needed)
- [x] `node scripts/check-unused-locals.mjs`, `pnpm run check:vitest-timeout-audit`, `node scripts/check-test-wiring.mjs` — OK
- [x] Two changesets: `@ifc-lite/export` (minor, updated wording) and `@ifc-lite/ifcx` (new, minor)

## Summary by CodeRabbit

* **New Features**
  * IFC exports now preserve entity classification references, including system, code, URI, and description.
  * Classification data remains available when exported IFCX files are re-imported.
  * Imported classifications appear as queryable classification property sets, with separate entries for multiple references.

* **Bug Fixes**
  * Classification references without codes are excluded from imported properties.
  * Naming collisions between classification property sets are handled without overwriting data.


louistrue (commented):
Reviewed at head b03debd4 (the four post-review commits plus the main merge — the CodeRabbit review predates every commit on this branch, so this covers the whole diff).

What I checked, in a scratch worktree at the PR head:

- **Read side (`property-extractor.ts`)**: the unpack correctly precedes the blanket `ifclite::*` skip; codeless refs are dropped the same way codeless materials already are; the per-system grouping plus the ` (n)` discriminator closes both collision routes (two refs on one system, and a system literally named `<system> - <code>`), and the discriminator loop is deterministic. The three new collision tests each pin their own Code/Uri pairing, so they can genuinely fail.
- **Write side (`ifc5-classification.ts`)**: verified `extractClassificationsOnDemand` returns `[]` when `store.source` is empty (packages/parser/src/classification-resolver.ts:62-63), so the IFCX-sourced-store comment is accurate, not aspirational. It also covers type-level associations via DefinesByType, matching the changeset's claim. `@ifc-lite/export` already declares `@ifc-lite/ifcx` (`workspace:^`), so the `IFCLITE_ATTR` import is a declared dependency, not a phantom.
- **Ran it**: `@ifc-lite/export` 1065 passed / 0 failed (the round-trip test runs and passes — it re-imports through the real `parseIfcx`, not a re-parse of the same JSON). `@ifc-lite/ifcx` 198/198 after `pnpm fixtures` (the real-IFC5 pinned test was fixture-skipped on a bare checkout; I fetched the 164 fixtures so it actually ran).

No defects found. Both minor bumps are right for new carrier/reader behaviour. Merging.

PR #3617, opened 2026-09-01
fix(cli): union all same-named storeys in query --storey instead of the first

BIMvoice:
## Summary

This is the CLI-query instance of the same-named-set family already found and fixed for the viewer (#3545, `HierarchyPanel` storey filter) and adjacent to the earlier CLI `--storey` fix (#3493, `--limit`/`--offset` dropped under `--storey`). The spatial-hierarchy builders themselves (`packages/parser/src/spatial-hierarchy-builder.ts` and the Rust fast-boot path in `rust/processing/src/processor/quick_metadata.rs`) already key every containment/aggregation map by `expressId`, not `Name`, and both have a tested cycle guard against malformed `IfcRelAggregates` chains — the bug here is entirely in the CLI's own name-based storey lookup, one layer above the builders.

`IfcBuildingStorey.Name` is not unique: two storeys can legally share a `Name` as siblings under different buildings, or a malformed/federated file can duplicate a level name outright. `ifc-lite query --storey <name>` resolved its argument to a single storey via `Array.prototype.find`, so when two storeys shared a `Name`, only the first one (by internal array order) was used — every element in the second, same-named storey was silently excluded from the result with no error and no warning ("exit 0, wrong answer").

## Fix

`packages/cli/src/commands/query.ts`: `--storey` now resolves an unambiguous `expressId` match to exactly one storey (identity-based, unchanged behavior), but a `Name` match (exact first, then substring) now collects **every** storey with that name and unions their contained elements, instead of picking an arbitrary first match. A unique storey name still resolves exactly as before.

## RED/GREEN

New test: `packages/cli/src/commands/query.storey-same-name.test.ts`
- RED (confirmed on unmodified `upstream/main`): a fixture with two `IfcBuildingStorey` both named `"Level 1"` — one containing `Wall A`, the other `Column C` — returns only `['Wall A']` for `--storey "Level 1"`.
- GREEN (after the fix): returns `['Column C', 'Wall A']`.
- Control: a fixture with unique storey names (`"Level 1"` / `"Level 2"`) still resolves exactly one storey's elements, unchanged.

## Verification (foreground)

- `pnpm exec vitest run` in `packages/cli`: 570 passed, 0 skipped-as-failed, no regressions (existing `--storey`/`--limit`/`--offset` coverage from #3493 still green).
- `pnpm exec tsc --noEmit` in `packages/cli`: clean.
- `node scripts/check-module-size.mjs`, `node scripts/check-test-wiring.mjs`, `node scripts/check-unused-locals.mjs`, `pnpm run check:vitest-timeout-audit`: all pass, no new findings attributable to this change.
- No `.rs` files touched, so the Rust module-size ratchet / crate `cargo test` gates don't apply.
- No `index.ts` public-export changes; changeset added since `@ifc-lite/cli` is published (`.changeset/query-storey-same-name.md`, patch).

## Test plan

- [ ] CI: typecheck, TS tests, test-wiring, API-surface, changeset presence

### DIFF
--- a/.changeset/query-storey-same-name.md
+++ b/.changeset/query-storey-same-name.md
@@ -1,0 +1,17 @@
+---
+'@ifc-lite/cli': patch
+---
+
+Fix: `ifc-lite query --storey <name>` no longer drops elements from a second storey that shares the same `Name`.
+
+`IfcBuildingStorey.Name` is not unique — two storeys legally share a `Name` as
+siblings under different buildings, and a malformed or federated file can
+duplicate a level name outright. `--storey` resolved its argument to a single
+storey via `Array.find`, so when two storeys shared a `Name` only the first one
+(by internal array order) was used; every entity in the second, same-named
+storey was silently excluded from the result with no error and no warning.
+
+`--storey` now resolves an unambiguous `expressId` to exactly one storey, but
+resolves a `Name` (exact or substring match) to every storey with that name and
+unions their contained elements. A single unique storey name behaves exactly as
+before.
--- a/.changeset/query-storey-same-name.md
+++ b/.changeset/query-storey-same-name.md
@@ -12,6 +12,9 @@ storey via `Array.find`, so when two storeys shared a `Name` only the first one
 storey was silently excluded from the result with no error and no warning.
 
 `--storey` now resolves an unambiguous `expressId` to exactly one storey, but
-resolves a `Name` (exact or substring match) to every storey with that name and
-unions their contained elements. A single unique storey name behaves exactly as
-before.
+resolves an exact `Name` to every storey with that name and unions their
+contained elements. A substring match unions its storeys only when they all
+share one `Name`; a substring spanning differently named storeys (`"Level"`
+against `Level 1` and `Level 2`) exits 1 and lists the candidate names instead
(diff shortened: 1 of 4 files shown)


louistrue (commented):

- RED/GREEN, run locally in a scratch worktree: 2/2 at head. With packages/cli/src/commands/query.ts reverted to this branch's merge-base (7aca203ac), the same-name test fails exactly as claimed (['Wall A'] instead of ['Column C','Wall A']) and the unique-name control still passes. (First attempt against origin/main's query.ts failed to import — main has since grown where-filter.js — which is an environment artifact of this PR's older base, not a code problem.)
- Verified the extraction is still faithful on today's main: the --storey block this PR replaces is byte-identical on current origin/main (query.ts:300-313), so the old-base diff lands the fix intact.
- The new resolution order (expressId > exact name > substring) quietly fixes a second bug the old single find had: a storey named 'Level 10' earlier in array order used to shadow --storey 'Level 1' via the case-insensitive substring predicate; now the exact tier wins.

One non-blocking note: expressId now beats a name outright — a storey literally named '5' loses to whatever entity has expressId 5 if both match, where the old code let array order decide. Defensible (expressIds are unique, names are not) but it is a behavior change the changeset doesn't mention; a line there wouldn't hurt.

PR #3683, opened 2026-09-02
fix(viewer): Lists Area/Volume falls back to LENGTHUNIT²/³ when no explicit AREAUNIT/VOLUMEUNIT

BIMvoice:
## Summary

Unit-scale hunt (no ready issue queued; labeled `unqueued`).

`resolveListColumnUnits` (`apps/viewer/src/lib/units/list-column-units.ts`), the single resolver shared by the Lists on-screen table and its XLSX/CSV export, resolved a row's SOURCE unit for an Area/Volume column by reading the model's declared `AREAUNIT`/`VOLUMEUNIT` directly and falling back to `siScale: 1` when the file declared neither.

IFC does not require `AREAUNIT`/`VOLUMEUNIT` to be declared at all — a project that declares only a `LENGTHUNIT` (common) implicitly means the area/volume unit is the length unit squared/cubed. Three other places in the codebase already implement this fallback deliberately: `quantitySiScale` (`packages/parser/src/quantity-collect.ts`), IDS's `resolveMeasureScales` (`packages/ids/src/bridge/units.ts`), and `MergedExporter.resolveDerivedUnitScale`. `list-column-units.ts` was the one boundary that didn't.

**RED (unmodified `upstream/main`)**: a model declaring `LENGTHUNIT = MILLIMETRE` and no explicit `AREAUNIT`, with a Qto_ Area quantity raw value of `1,000,000` (i.e. 1,000,000 mm² = 1 m²) — Lists shows/exports `1,000,000` labeled `"m²"`. A 1,000,000× overstatement under the label shown (Volume: `1,000,000,000` mm³ shown as `1,000,000,000 m³`, a 1e9× overstatement).

**GREEN (fixed)**: the same input converts to `1 m²` / `1 m³`.

**Control**: Length columns, an explicit-unit measure property (m³/s → m³/h), the °C affine-offset case, and federated mm+ft columns are all unchanged (existing tests still pass) — the fix only touches the AREAUNIT/VOLUMEUNIT-undeclared source-side path.

The fix adds `sourceUnitFor`, used only on the per-row SOURCE side of `convertCell`; the TARGET side keeps the existing "no model declares this unit-type → fall to the SI default" rule unchanged (this matters for federation: the target must stay a single fixed unit for the whole column, not silently adopt whichever model happens to iterate first).

## Test plan
- [x] `apps/viewer/src/lib/units/list-column-units.test.ts`: 2 new tests (Area/Volume fallback), 10 existing tests unchanged — 12/12 pass
- [x] `pnpm typecheck --filter @ifc-lite/viewer` (turbo) — clean
- [x] `node scripts/check-module-size.mjs` — no new violation
- [x] `node scripts/check-unused-locals.mjs` — no new violation (pre-existing unrelated `embed-sdk` compile note)
- [x] `pnpm run check:vitest-timeout-audit` — no change in baseline
- [x] `apps/viewer` is a private (unpublished) package — no changeset needed

### DIFF
--- a/apps/viewer/src/lib/units/list-column-units.test.ts
+++ b/apps/viewer/src/lib/units/list-column-units.test.ts
@@ -196,6 +196,29 @@ describe('resolveListColumnUnits (issue #1573 follow-up)', () => {
     assert.ok(Math.abs((ftFirst.convertCell(0, 1000, 'mmModel') as number) - 3.280839895013123) < 1e-9); // 1000mm -> ft
   });
 
+  it('converts an Area quantity column when the file declares no explicit AREAUNIT (falls back to LENGTHUNIT²)', () => {
+    // MM_MODEL declares only LENGTHUNIT (millimetre) — no explicit AREAUNIT.
+    // Per IFC convention (and `quantitySiScale`/`resolveMeasureScales`'s
+    // documented fallback), an undeclared AREAUNIT derives from the length
+    // unit SQUARED: 1,000,000 mm² is stored as the raw value 1e6 and must
+    // read back as 1 m².
+    const columns: ColumnDefinition[] = [
+      { id: 'area', source: 'quantity', psetName: 'Qto', propertyName: 'Area', quantityType: QuantityType.Area },
+    ];
+    const resolver = resolveListColumnUnits(columns, new Map([['m1', MM_MODEL]]), {});
+    assert.strictEqual(resolver.unitSymbol(0), 'm²');
+    assert.ok(Math.abs((resolver.convertCell(0, 1_000_000, 'm1') as number) - 1) < 1e-9);
+  });
+
+  it('converts a Volume quantity column when the file declares no explicit VOLUMEUNIT (falls back to LENGTHUNIT³)', () => {
+    const columns: ColumnDefinition[] = [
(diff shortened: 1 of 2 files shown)


louistrue (commented):

- Fixture pressure: MM_MODEL in list-column-units.test.ts declares only LENGTHUNIT (mm) plus a derived VOLUMETRICFLOWRATEUNIT — no AREAUNIT/VOLUMEUNIT — so the two new tests genuinely exercise the fallback rather than a declared-unit path.
- RED/GREEN, run locally in a scratch worktree: 12/12 at head; with apps/viewer/src/lib/units/list-column-units.ts reverted to origin/main (tests kept), exactly the two new tests fail (raw 1e6 passes through under the m² label) and the 10 existing tests still pass. The fix is the only thing making them green.
- Precedent claim: confirmed quantitySiScale (packages/parser/src/quantity-collect.ts:283-292) implements the same LENGTHUNIT²/³ fallback, so this brings the Lists resolver in line rather than inventing a new rule.
- Target side: resolveTarget is untouched; the derived unit is applied only per-row on the source side, so a federated column still normalizes to one fixed target. The identity short-circuit now correctly does NOT fire for the mm²-source/m²-target case (scales differ), which is the whole point.
- The comment explaining why the derived branch bypasses resolveFromUnit (symbol collision with the curated SI m²/m³ alternative would discard the derived scale) matches what resolveFromUnit does.

No defects found. Merging.

PR #3684, opened 2026-09-02
fix(export): preserve GlobalId as the IFCX node path in the Rust IFC5 exporter

BIMvoice:
## Lens
GlobalId identity census (hunt task, no tracked issue — flagging `unqueued`). Scope: sites that generate, parse, compare, remap, or truncate an IFC `GlobalId`, judged for whether identity survives.

## Census

| Site | Generates / parses / compares / remaps | Identity survives? |
|---|---|---|
| `packages/encoding/src/guid.ts` (`uuidToIfcGuid`/`ifcGuidToUuid`, buildingSMART alphabet) | Encoding | Yes — round-trips losslessly; correct alphabet, not standard base64 |
| `packages/parser/src/deterministic-global-id.ts` | Generates (schedule tasks/sequences only, not IFC products) | Yes — used only where no upstream GlobalId exists to begin with; not reachable on any path that already has an identity |
| `packages/create/src/ifc-creator.ts`, `in-store/*` (walls/beams/columns/slabs/spaces) | Generates (new elements) | Yes — `generateIfcGuid`/`crypto.randomUUID`-backed, correct guardrail case (new element, new identity) |
| `rust/export/src/csv.rs`, `parquet_bos.rs`, `gltf.rs`, `model.rs` | Passes through `EntityRow.global_id` | Yes — untouched pass-through to output columns/extras |
| `packages/export/src/ifc5-exporter.ts` (TS IFC5/IFCX exporter) | Remaps (IFC → IFCX node path) | Yes — already prefers `GlobalId` as the path, falls back to a generated UUID only when absent |
| `rust/export/src/ifc5.rs` (Rust IFC5/IFCX exporter, backs `ifc-lite export --format ifcx`) | Remaps (IFC → IFCX node path) | **No, before this fix** — every product got a deterministic hash of its express id as its path; the real GlobalId was discarded and appeared nowhere in the output |
| `packages/ifcx/src/entity-extractor.ts` | Parses (IFCX → internal EntityTable, uses `node.path` as GlobalId) | By design for native IFC5 files (path IS identity in IFC5); downstream of the writer bug above for any IFC→IFCX→IFC round trip. Owned by open PR #3677 — out of scope here |
| `packages/ifcx/src/writer.ts` (TS `IfcxWriter`/`exportToIfcx`, published `@ifc-lite/ifcx` API) | Remaps (IFC → IFCX node path) | **No** — fetches `entities.globalId[i]` and never uses it; same defect class as the Rust exporter, but currently unreachable from any wired CLI/viewer path (CLI's `ifcx` export goes through the Rust wasm binding, not this writer). Flagging for a maintainer follow-up; out of scope for this PR (different package, different reachability story) |
| `packages/export/src/schema-converter.ts` / `schema-converter-attr-remap.ts` (schema up/downgrade) | Remaps (attribute list to target schema) | Yes — `GlobalId` is attribute 0 of every rooted entity and is carried by position in both the direct and by-name remap paths; #3653/#3655 already fixed the IfcDoorType/IfcWindowType→proxy identity loss in this family |
| `packages/query`/`PropertyTable`/diff (comparison) | Compares GlobalId | Yes — exact string comparison, no case-folding or trimming found on either side |
| Federation (`packages/renderer/src/federation-registry.ts`) | Remaps (renderer-local id ↔ GlobalId per model) | Yes — already fixed for the overlay-collision class in #3651 |

## The one fix

`rust/export/src/ifc5.rs`'s `export_ifc5` built every node's `path` from `uuid_from_id(express_id)` — a deterministic hash with no relationship to the source entity's `GlobalId`. `EntityRow.global_id` was available (used elsewhere in the same crate for CSV/Parquet/glTF output) but never read here. This is the exact defect class called out by the lens: an element that *already had* an identity got a new one minted for it, silently, on the production `ifc-lite export --format ifcx` path. Any external reference (a BCF topic, a diff, an asset register keyed on the source model's GlobalId) would no longer resolve after conversion.

New `path_for_id(id, by_id)` reuses the entity's own `GlobalId` as its IFCX path when it has one, falling back to `uuid_from_id` only when the entity has none (currently just the `IfcProject` root, which is decoded outside the product-row model). This mirrors the TS sibling exporter (`packages/export/src/ifc5-exporter.ts`), which already did this — the Rust port had silently dropped it.

### RED / GREEN
`rust/export/src/ifc5_tests.rs`, new test `globalid_survives_ifc_to_ifcx_conversion`: exports `duplex.ifc`, picks a real `IfcWall`'s GlobalId (`2O2Fr$t4X7Zf8NOew3FKau`), and asserts it appears as a node path in the exported IFCX.
- On unmodified `upstream/main`: **RED** — panics with `expected the source IfcWall's GlobalId "2O2Fr$t4X7Zf8NOew3FKau" to survive as an IFCX node path` (verified by reverting `path_for_id` back to `uuid_from_id` and re-running).
- After the fix: **GREEN**.

**Control** (`project_root_without_globalid_still_gets_a_stable_path`): the `IfcProject` root, which never had a GlobalId to preserve, keeps its existing deterministic fallback path — unaffected, still unique and stable across runs. The existing `duplex_exports_valid_ifcx` test's uniqueness/well-formedness assertions still pass (its "uuid-shaped" path assertion was loosened to "non-empty," since paths are correctly non-hyphenated GlobalIds now, not uniformly synthesized UUIDs).

## Verification (foreground)
- `cargo test -p ifc-lite-export` — 356/356 passed (fixture copied in manually since `pnpm fixtures` wasn't run in this worktree; `tests/models/ara3d/duplex.ifc` is gitignored and not committed).
- `cargo test -p ifc-lite-processing --test module_size_ratchet` — 6/6 passed; `ifc5.rs` split into `ifc5_tests.rs` (matching the `schema_convert.rs`/`schema_convert_tests.rs` precedent) to stay under the 400-line budget after the new tests (406 lines pre-split → 257 post-split).
- `cargo check -p ifc-lite-wasm` — clean (confirms the downstream wasm-bindings crate that exposes `exportIfcx` to the CLI still compiles).
- `node scripts/check-module-size.mjs` — OK, 0 new over budget.
- `node scripts/check-unused-locals.mjs` — OK, no new offenders (Rust-only change).
- Changeset added: `@ifc-lite/wasm` patch (this is a Rust change surfaced through the wasm binding that backs `ifc-lite export --format ifcx`).

## Overlap / not touched
`rust/export/src/ifc5.rs` had no open-PR overlap (checked via the repo's PR-file-overlap script). `packages/export/src/ifc5-exporter.ts` (TS sibling, already correct) is owned by #3608/#3676 — not touched. `packages/ifcx/src/entity-extractor.ts` is owned by #3677 — not touched. The `packages/ifcx/src/writer.ts` GlobalId-drop noted in the census table above is a separate, currently-unreachable defect in a different package; flagging for a maintainer follow-up rather than bundling into this PR.

### DIFF
--- a/.changeset/fix-ifc5-export-preserve-globalid.md
+++ b/.changeset/fix-ifc5-export-preserve-globalid.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Fixed `ifc-lite export --format ifcx` (and any other caller of the Rust IFC5/IFCX exporter) minting a brand-new node path for every product instead of using its IFC `GlobalId`. An element converted from IFC to IFCX lost its original identity — a BCF topic, diff, or any other reference keyed on the source model's GlobalId could no longer find it in the converted file. The exporter now reuses the entity's own GlobalId as its IFCX path when it has one (mirroring the TS exporter, which already did this), falling back to the deterministic synthesized path only for entities with no GlobalId (e.g. the `IfcProject` root).
--- a/rust/export/src/ifc5.rs
+++ b/rust/export/src/ifc5.rs
@@ -55,11 +55,33 @@ impl Default for Ifc5Options {
 }
 
 /// Deterministic UUID-shaped path for an express id (no RNG/clock — wasm-safe).
+/// Fallback only — used when the entity has no `GlobalId` to carry forward
+/// (e.g. the `IfcProject` root, decoded separately from the product rows
+/// `path_for_id` below reads from).
 fn uuid_from_id(id: u32) -> String {
     let a = (id as u64).wrapping_mul(0x9E37_79B9_7F4A_7C15) ^ 0xABCD_EF01_2345_6789;
     let b = (id as u64).wrapping_mul(0xC2B2_AE3D_27D4_EB4F) ^ 0x0F1E_2D3C_4B5A_6978;
     let s = format!("{a:016x}{b:016x}");
     format!("{}-{}-{}-{}-{}", &s[0..8], &s[8..12], &s[12..16], &s[16..20], &s[20..32])
+}
+
+/// IFCX node path for an express id: the entity's own `GlobalId` when it has
+/// one, so a BCF topic / diff / external reference keyed on the source
+/// model's GlobalId still resolves to the same element after IFC→IFCX
+/// conversion; the deterministic fallback otherwise. Mirrors the TS
+/// exporter's `buildEntityMaps` (`packages/export/src/ifc5-exporter.ts`),
+/// which already prefers GlobalId — the Rust port had dropped that and
+/// unconditionally minted a fresh identity for every product, discarding the
+/// original on every `ifc-lite export --format ifcx`.
+fn path_for_id(id: u32, by_id: &HashMap<u32, &EntityRow>) -> String {
+    if let Some(row) = by_id.get(&id) {
+        if let Some(g) = &row.global_id {
+            if !g.is_empty() {
+                return g.clone();
+            }
+        }
+    }
+    uuid_from_id(id)
 }
 
 /// Sanitize a USD prim name (the keys in a node's `children` dict).
@@ -184,7 +206,7 @@ pub fn export_ifc5(content: &[u8], opts: &Ifc5Options) -> String {
             .cloned()
             .unwrap_or_else(|| (String::new(), "IfcProduct".to_string()));
         let mut node = Map::new();
-        node.insert("path".into(), json!(uuid_from_id(*id)));
+        node.insert("path".into(), json!(path_for_id(*id, &by_id)));
 
         if let Some(ch) = children.get(id) {
             let mut child_map = Map::new();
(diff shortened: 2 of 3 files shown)


louistrue (commented):

- The silent-skip hazard around fixture_or_skip! is closed on CI: the Rust lane runs cargo test --workspace with IFC_LITE_REQUIRE_FIXTURES=1 (.github/workflows/test.yml:1370), ara3d/duplex.ifc is in tests/models/manifest.json, and the Rust tests check is SUCCESS at this head — so globalid_survives_ifc_to_ifcx_conversion actually ran there rather than skipping.
- Ran cargo test -p ifc-lite-export at this head locally with fixtures present: 356 passed, 0 failed.
- Mutation check on the RED claim: reverting path_for_id to unconditionally return uuid_from_id fails exactly globalid_survives_ifc_to_ifcx_conversion; duplex_exports_valid_ifcx, the header test, unknown-props, and the project-root control all stay green. The test has real teeth and pins the fix, not an adjacent property.
- Both call sites (node path and children map values) go through path_for_id, so child references stay consistent with node paths; entities missing from by_id fall back symmetrically.

One non-blocking note: a malformed source with duplicate GlobalIds now yields duplicate IFCX paths where the old express-id hash was unique per node (duplex_exports_valid_ifcx asserts uniqueness only for this fixture). The TS exporter already has the same property, so the two exporters now degrade identically on that input; just flagging that the uniqueness guarantee moved from structural to inherited-from-input.

PR #3692, opened 2026-09-02
fix(create): schema-conditional trailing attributes for quantities, wall/column/beam, IfcRelSequence

BIMvoice:
## Reviewer summary

- **Without this:** IFC files ifc-lite creates can be schema-invalid — wrong attribute count for the target schema — which risks rejection or misparse by other tools that validate strictly.
- **Evidence:** reverting to this PR's own first (unconditional) attempt produced a 9-attribute IFC2X3 wall (`IFCWALL(...,$,.STANDARD.);`) where the schema declares 8 attributes — caught in review before merge, now fixed schema-conditionally.
- **Risk:** confined to the handful of entities (quantities, wall/column/beam `PredefinedType`, `IfcRelSequence`) with a single trailing IFC4-only attribute; every other entity in the original conformance table (`IfcSlab`, psets, `IfcProject`, etc.) is confirmed unchanged. Two-schema test coverage (IFC2X3 4-slot vs IFC4 5-slot) guards against regressing either direction again.
- **Sequencing:** none; based on `main`.

---

## Summary

Audit lens: is the IFC ifc-lite **creates** actually schema-valid (sibling question to #3674's exporter audit, for `packages/create`)?

`IfcCreator.addIfcElementQuantity` (`packages/create/src/ifc-creator.ts`) writes `IfcQuantityLength`/`Area`/`Volume`/`Weight`/`Count` records via `quantityValueField` + a hand-built parameter string. `IfcPhysicalSimpleQuantity`'s trailing `Formula` attribute — declared from IFC4 on — was omitted from the parameter list entirely instead of written as an unset `$`, leaving every quantity record one attribute short of the IFC4/IFC4X3 declaration.

**The first version of this fix appended the trailing `$` unconditionally**, which broke IFC2X3 in the opposite direction: `entities-ifc2x3.ts` gives those same quantity entities only **4** attributes — `Formula` doesn't exist in IFC2X3 at all, not even as optional — so an unconditional `$` produced a 5-slot record against a 4-attribute declaration, exactly as invalid as the 4-slot record this PR started from. Caught in review before merge.

The fix is now schema-conditional, driven off `IfcCreator`'s own `Schema: 'IFC2X3' | 'IFC4' | 'IFC4X3'` field (already stamped into `FILE_SCHEMA`) via a shared `ifc4Only()` helper, rather than a hardcoded `,$`.

While auditing whether the same schema-version question applies elsewhere in the original conformance table, **three more emission sites turned out to have the identical one-line shape** (a single trailing attribute IFC4/IFC4X3 declare that IFC2X3 does not, everything else byte-identical) and are fixed here too, through the same helper:

- `IfcWall`/`IfcColumn`/`IfcBeam`: IFC4 adds a trailing `PredefinedType` (9 attrs vs IFC2X3's 8). All 7 emission sites (1 wall, 3 column-profile variants, 3 beam-profile variants) now go through `ifc4Only()`.
- `IfcRelSequence`: IFC4 adds a trailing `UserDefinedSequenceType` (9 attrs vs IFC2X3's 8).

## Cross-schema check on the rest of the conformance table (§3)

Every entity from the original table was diffed IFC4-vs-IFC2X3. Fixed above are the ones with a genuine single-trailing-attribute shape. **Not fixed here** — the divergence is bigger than one attribute and needs separate scoping:

| Entity | IFC2X3 vs IFC4 divergence | Why not a one-line fix |
|---|---|---|
| `IfcDoor`/`IfcWindow` | IFC2X3 ends at `OverallWidth` (10 attrs); IFC4 adds **three** trailing attrs at once (`PredefinedType`, `OperationType`/`PartitioningType`, `UserDefinedOperationType`/`UserDefinedPartitioningType`) | The creator always writes concrete enum values for the first two rather than `$` — a design question (what's the IFC2X3-equivalent default?), not a mechanical drop |
| `IfcRoof` | IFC2X3's trailing attribute is `ShapeType`; IFC4's is `PredefinedType` | A renamed attribute with a different value domain at the same position, not an added/removed slot |
| `IfcTask` | IFC2X3: `TaskId` + 4 attrs (10 total). IFC4: `Identification`/`LongDescription`/`TaskTime`/`PredefinedType` (13 total) | Attribute sets differ throughout, not just at the tail |
| `IfcTaskTime`/`IfcLagTime` | **Absent from IFC2X3's generated table entirely** | `addIfcTask`/`addIfcRelSequence` would emit `IFCTASKTIME`/`IFCLAGTIME` records for entities that don't exist in an IFC2X3 file — a deeper problem than attribute shape |
| `IfcWorkSchedule`/`IfcWorkPlan` | IFC2X3 has **15** attrs vs IFC4's 14 | `Identification`/`PredefinedType` are renamed to `Identifier`/`WorkControlType` at the same positions, and IFC2X3 additionally adds a trailing `UserDefinedControlType` IFC4 doesn't have — a rename plus more attributes, not fewer |

Unchanged from the original audit (schema-identical between IFC2X3 and IFC4 for every attribute checked): `IfcSlab`, `IfcPropertySet`, `IfcPropertySingleValue`, `IfcRelDefinesByProperties`, `IfcElementQuantity` (the container, not its quantities), `IfcOwnerHistory`, `IfcProject`, `IfcSite`, `IfcBuilding`.

## RED/GREEN

Two-schema coverage — a single assertion against the IFC4 default cannot catch this class of error, which is exactly how the first version of this fix got through:

- **RED** (this PR's own prior commit, the unconditional `,$`): `IFCQUANTITYAREA('GrossArea',$,$,80.,$);` in IFC2X3 (5 slots against a 4-attribute schema) and `IFCWALL(...,$,.STANDARD.);` in IFC2X3 (9 slots against an 8-attribute schema).
- **GREEN** (this commit): IFC2X3 → `IFCQUANTITYAREA('GrossArea',$,$,80.);` (4 slots) and `IFCWALL(...,#41,$);` (8 slots, no trailing `.STANDARD.`). IFC4 default unchanged: `IFCQUANTITYAREA('GrossArea',$,$,80.,$);` (5 slots) and `IFCWALL(...,$,.STANDARD.);` (9 slots).
- **Control**: `IfcSlab` and every other checked entity in the original conformance table is unchanged.

`packages/create/src/ifc-creator.test.ts` now has both an IFC4 quantity test (5-slot) and an IFC2X3 quantity test (4-slot), plus a new test asserting `IfcWall`/`IfcColumn`/`IfcBeam`/`IfcRelSequence` drop their trailing attribute in IFC2X3.

## Test plan

- [x] `pnpm typecheck` (turbo, `@ifc-lite/create` + `@ifc-lite/data`)
- [x] `pnpm test` (turbo, `@ifc-lite/create`) — 27 files, 398 tests pass
- [x] `node scripts/check-module-size.mjs` — `ifc-creator.ts` at 2843/2843 budget (no allowlist change; the shared `ifc4Only()` helper was added specifically to stay within budget across 8 call sites)
- [x] `pnpm run check:vitest-timeout-audit` — no new unprotected tests added
- [x] Changeset updated to describe the schema-conditional behavior

Base SHA worked from: `fa0624bc0` (`upstream/main`, confirmed via `git fetch upstream` + `git log --oneline -1 upstream/main`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected IFC file generation to match schema-specific attribute requirements.
  * Fixed quantity, relationship, and building-element records for IFC2X3, IFC4, and IFC4X3 formats.
  * Prevented invalid trailing attributes from appearing in IFC2X3 output.
  * Ensured optional IFC4/IFC4X3 attributes are included correctly, including quantity formulas.

### DIFF
--- a/.changeset/create-quantity-formula-attribute.md
+++ b/.changeset/create-quantity-formula-attribute.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/create': patch
+---
+
+`addIfcElementQuantity` wrote `IfcQuantityLength`/`Area`/`Volume`/`Weight`/`Count` records with only 4 attributes (`Name`, `Description`, `Unit`, `<Value>`) instead of the 5 the schema declares — `IfcPhysicalSimpleQuantity`'s trailing `Formula` attribute was omitted entirely rather than written as an unset `$`. STEP part 21 requires every declared attribute position to be present (`$` for unset), so the emitted record was one field short of what `IfcQuantityLength`/etc. need; a strict reader can reject the entity outright. Every quantity attached via the public `addIfcElementQuantity` API now writes the trailing `$` for `Formula`.
--- a/.changeset/create-quantity-formula-attribute.md
+++ b/.changeset/create-quantity-formula-attribute.md
@@ -2,4 +2,9 @@
 '@ifc-lite/create': patch
 ---
 
-`addIfcElementQuantity` wrote `IfcQuantityLength`/`Area`/`Volume`/`Weight`/`Count` records with only 4 attributes (`Name`, `Description`, `Unit`, `<Value>`) instead of the 5 the schema declares — `IfcPhysicalSimpleQuantity`'s trailing `Formula` attribute was omitted entirely rather than written as an unset `$`. STEP part 21 requires every declared attribute position to be present (`$` for unset), so the emitted record was one field short of what `IfcQuantityLength`/etc. need; a strict reader can reject the entity outright. Every quantity attached via the public `addIfcElementQuantity` API now writes the trailing `$` for `Formula`.
+Several `IfcCreator` element/relationship writers wrote an attribute count that only matched the IFC4/IFC4X3 schema, not IFC2X3, for entities whose trailing attribute list genuinely differs between schema versions:
+
+- `addIfcElementQuantity` omitted `IfcQuantityLength`/`Area`/`Volume`/`Weight`/`Count`'s trailing `Formula` attribute entirely (added in IFC4) instead of writing it as an unset `$` — every quantity record was one attribute short of the IFC4/IFC4X3 declaration.
+- `addIfcWall`/`addIfcColumn`/`addIfcBeam` and `addIfcRelSequence` always wrote a trailing `PredefinedType`/`UserDefinedSequenceType` value, an attribute IFC2X3 does not declare at all — so a creator targeting `Schema: 'IFC2X3'` emitted one attribute too *many* for those entities, which is exactly as invalid as writing too few.
+
+STEP part 21 requires an explicit slot for every attribute a schema version declares — no more, no fewer. The trailing attribute is now written only for schemas that declare it (IFC4/IFC4X3), driven off the creator's own `Schema` field, and omitted for IFC2X3.
--- a/packages/create/src/ifc-creator.test.ts
+++ b/packages/create/src/ifc-creator.test.ts
@@ -336,14 +336,62 @@ describe('IfcCreator', () => {
     expect(result.content).toContain("'Qto_SlabBaseQuantities'");
 
     // IfcPhysicalSimpleQuantity subtypes (IfcQuantityArea/Volume/…) declare
-    // 5 attributes: Name, Description, Unit, <Value>, Formula. Formula is
-    // trailing-optional but STEP still requires an explicit `$` slot for
-    // it — a record with only 4 fields is one attribute short of what the
-    // schema declares (packages/data/src/ifc-schema/generated/entities-ifc4.ts).
+    // 5 attributes from IFC4 on: Name, Description, Unit, <Value>, Formula.
+    // Formula is trailing-optional but STEP still requires an explicit `$`
+    // slot for it — a record with only 4 fields is one attribute short of
+    // what the default IFC4 schema declares
+    // (packages/data/src/ifc-schema/generated/entities-ifc4.ts).
     const areaLine = result.content.split('\n').find(l => l.includes('IFCQUANTITYAREA('));
     const volumeLine = result.content.split('\n').find(l => l.includes('IFCQUANTITYVOLUME('));
     expect(areaLine).toMatch(/^#\d+=IFCQUANTITYAREA\('GrossArea',\$,\$,80\.,\$\);$/);
     expect(volumeLine).toMatch(/^#\d+=IFCQUANTITYVOLUME\('GrossVolume',\$,\$,24\.,\$\);$/);
+  });
+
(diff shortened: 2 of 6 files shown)


louistrue (commented):
Review focused on 8dfc809 (the golden re-bless, landed after CodeRabbit's review of db4ffb1), plus the CodeRabbit finding and the changeset.

**8dfc809 (B3.5 golden re-bless)** — the delta is measured, not just accepted:
- The diff touches exactly what the commit message claims: act1 `fileSize` 28008 -> 28208 (+200 bytes = 100 quantity records x ",$") and two sha256s in the `deterministic` subtree; `entityCount` (548) and every `groundTruthTotals` value are unchanged; the remaining hunks are confined to the block the report itself declares volatile (timestamps, wall clocks, artifactDir, node version).
- Independent confirmation: the "Moonshot exams (... B3.5 golden ...)" lane is green at this head, i.e. CI regenerated the demo from the seed and matched the new golden — the re-bless is reproducible, not hand-edited.
- One observation for a follow-up issue, not this PR: the pre-fix golden already said `schema: valid=true (0 errors)`, so ifc-lite's own validator does not catch a wrong attribute count. The invalid records this PR fixes sailed through it.

**Fix itself** (commits 1-3, CodeRabbit-reviewed): verified the schema claims against the generated tables — `entities-ifc2x3.ts` gives `IfcQuantityArea` 4 attributes (no Formula) and the full creator diff is exactly the 9 one-line `ifc4Only()` conversions plus the helper. The two-schema tests are the right guard, and the IFC2X3 quantity assertions are fully anchored, so a regression in `ifc4Only()` itself cannot pass.

**Inherited residue, NOT fixed here (deliberately — from main, not this PR):** `IFCMEMBER` (ifc-creator.ts:976, :1342, :1399, :1456) and `IFCPLATE` (:937) still write an unconditional trailing `,.NOTDEFINED.`; IFC2X3 declares 8 attributes for both (no PredefinedType), so this is the identical one-line shape, fixable through the same helper in a follow-up alongside the door/window/task table in the PR body. I narrowed the PR title to the entities actually fixed so the squash headline does not claim the whole class.

**CodeRabbit's open comment** (the IFC2X3 wall/column/beam assertions match only the `,$);` tail, so an extra `$` slot would pass): correct as stated, deferred — the shared `ifc4Only()` helper is pinned by the anchored quantity assertions, so the only regression this misses is a hand edit at one emission site. Worth folding into the member/plate follow-up.

Changeset is `patch` and describes the schema-conditional behavior accurately (output-shape bug fix, no new exports). Required checks (Build + WASM + Rust + Node, parity, PR review signal) are green at 8dfc809; the failing lanes (Claude review / Review posted / Issue queue) are not in the required set — this review supplies the missing review bar. Merging.

PR #3694, opened 2026-09-02
fix(geometry,viewer): surface elements dropped from 2D drawing profile extraction

BIMvoice:
## Reviewer summary

- **Without this:** a wall or slab can silently disappear from a generated 2D construction drawing, with nothing telling the user an element was dropped.
- **Evidence:** reverting, a fixture with one healthy wall and one wall whose `IfcExtrudedAreaSolid.SweptArea` references a non-existent id returns only the healthy wall's profile from `extract_profiles` — the broken wall vanishes silently. Fixed: the diagnostics variant reports `skipped[0].express_id == 500, ifc_type == "IfcWall"` for the same input.
- **Risk:** `extract_profiles`'s existing signature and behavior are untouched (a thin wrapper); `rust/export`'s HBJSON/rooms/openings/shades callers are unaffected, confirmed by the full `ifc-lite-export` suite (384 tests) staying green. Control test `healthy_file_reports_no_skips` confirms no false positives on a clean file.
- **Sequencing:** none; based on `main`. Same defect class as #3691, different pipeline (2D drawing extraction vs. 3D mesh).

---

## Summary

Sweep for the lens PR #3691 opened: a user-relevant signal emitted only under a compile-time flag that is OFF in the shipped wasm build. Found one genuine production-silenced data-loss site outside #3691's own files (see overlap notes below), and shipped a fix for it.

**Site:** `rust/geometry/src/profile_extractor.rs::extract_profiles` — the construction-projection profile walk that feeds the viewer's 2D drawing generation (`useDrawingGeneration.ts` → `parseProfilesFlat` → `packages/geometry`'s `extractProfiles` → `ifc_lite_geometry::extract_profiles`).

**What is hidden:** when an element has an `IfcExtrudedAreaSolid` (direct or via `IfcMappedItem`) but it fails to extract (malformed geometry, unresolvable `SweptArea`, etc.), or a mapped-item chain exceeds the extractor's depth limit, the element is dropped from the profile collection. The only trace was `crate::diag::diag_debug!`, whose legacy `eprintln!` leg is further gated `#[cfg(feature = "debug_geometry")]`.

**Off in release?** Yes, confirmed by reading `rust/geometry/Cargo.toml` (`default = []`; neither `debug_geometry` nor `observability` is default) and `scripts/build-wasm.sh` (`FEATURES=""` unless the developer sets `DEBUG_GEOMETRY=1` locally — never set in `test`/`release`/CI).

**User loses:** a wall or slab silently missing from a generated construction drawing, with no indication anything was dropped. Same shape as #3691 (element silently dropped from the geometry pipeline), one layer over — the 2D drawing path, not the 3D mesh path.

**Verdict:** top severity — silent data loss / wrong result, reachable from a real malformed/truncated reference, invisible end-to-end in the shipped build.

## The fix

- `rust/geometry/src/profile_extractor.rs`: new `extract_profiles_with_diagnostics` returns `(Vec<ExtractedProfile>, Vec<SkippedProfile>)`. The pre-existing `extract_profiles` is untouched in signature and behaviour (now a thin wrapper), so `rust/export`'s HBJSON/rooms/openings/shades callers are unaffected — verified with `cargo test -p ifc-lite-export` (all 384 tests green).
- `SkippedProfile` lives in a new sibling module `rust/geometry/src/profile_skip.rs` to keep `profile_extractor.rs` under its `module_size_ratchet` budget (758 lines).
- `rust/wasm-bindings/src/api/extract_profiles.rs`: `extractProfiles` now calls the diagnostics variant and exposes `ProfileCollection.skippedExpressIds` (additive wasm API surface, `packages/wasm` changeset is `minor`).
- `apps/viewer/src/lib/overlay-parse/{profiles-flat,profile-entries}.ts`: `FlatProfiles` carries `skippedExpressIds` through the worker boundary; new `warnAboutSkippedProfiles` helper.
- `apps/viewer/src/hooks/useDrawingGeneration.ts`: calls `warnAboutSkippedProfiles` after building profile entries for construction projection.

## RED / GREEN / control

`rust/geometry/tests/issue_shipped_profile_drop_signal.rs`, a minimal two-wall fixture (one healthy, one with an `IfcExtrudedAreaSolid` whose `SweptArea` references a non-existent express id):

- **RED** (`broken_wall_is_dropped_from_the_2d_drawing`): `extract_profiles` (unmodified, upstream behaviour) returns only the healthy wall's profile — the broken wall is silently dropped, and there was no way to observe that.
- **GREEN** (`dropped_wall_is_reported_with_default_features`): `extract_profiles_with_diagnostics`, run with **default Cargo features** (the release-shaped feature set — no `debug_geometry`/`observability`), reports the drop: `skipped[0].express_id == 500`, `ifc_type == "IfcWall"`.
- **Control** (`healthy_file_reports_no_skips`): the same fixture with the broken wall removed reports zero skips — a diagnostic that fires on a healthy file would be worse than none.

Mirrored at the TS layer in `apps/viewer/src/lib/overlay-parse/profile-entries.test.ts` (`warnAboutSkippedProfiles` — warns with the dropped ids; no-op control on a clean flatten).

## Other candidates ruled out (not top-severity / not production-silenced)

- `rust/geometry/src/router/layers.rs` (material-layer slicing fallback): already has a browser-facing counter (`push_layer_slice_diag`) alongside `diag_warn!`, whose legacy leg keeps `eprintln!` in production (`diag.rs`'s documented WARN-vs-DEBUG gating policy) — not silenced.
- `rust/geometry/src/processors/advanced.rs` (AdvancedBrep empty-face loss): the whole `empty_faces` tracking + `diag_warn!` block is gated `#[cfg(any(feature = "debug_geometry", feature = "observability"))]`, so it is a real production-silenced signal of the same class — but it and the router-owned `GeometryDiagnostics` counter infra it would naturally extend are inside files an open PR (#3691) already owns (`rust/geometry/src/router/*`), so left for that PR or a follow-up to avoid a parallel-work collision.
- Everything else matching `cfg(debug_assertions)` / `feature = "observability"` / `diag_debug!` in `rust/geometry` and `rust/processing` is either developer-only tracing (correctly compiled out) or already covered by #3691's `unsupportedItemsByType`.

## Test plan

- [x] `cargo test -p ifc-lite-geometry` — full crate suite green (1666-line log, 0 failures)
- [x] `cargo test -p ifc-lite-processing --test module_size_ratchet` — 6/6 (budget respected after the `profile_skip` split)
- [x] `cargo test -p ifc-lite-export` — 384 tests green (unchanged `extract_profiles` callers)
- [x] `pnpm typecheck` (root) — 89/89 tasks, 1613/1613 test files
- [x] `pnpm build` (root, regenerates `packages/wasm/pkg/ifc-lite.d.ts`)
- [x] `pnpm --filter @ifc-lite/viewer test` — 6673 tests, 0 failed
- [x] `node scripts/check-module-size.mjs`, `check-unused-locals.mjs`, `check-test-wiring.mjs` — all clean
- [x] `pnpm changeset` added for `@ifc-lite/wasm` (minor, additive)

### DIFF
--- a/.changeset/profile-extraction-drop-signal.md
+++ b/.changeset/profile-extraction-drop-signal.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/wasm': minor
+---
+
+`extractProfiles` silently dropped any element whose `IfcExtrudedAreaSolid` (direct or via `IfcMappedItem`) failed to extract, or whose mapped-item chain exceeded the extractor's depth limit — the only trace was a Rust `diag_debug!` call compiled out of the shipped wasm build (neither `debug_geometry` nor `observability` is enabled there), so a dropped wall or slab was simply missing from the generated 2D construction drawing with no signal at all. `ProfileCollection` now exposes `skippedExpressIds`, the express IDs of every element dropped during extraction, empty on a clean model.
--- a/apps/viewer/src/hooks/useDrawingGeneration.ts
+++ b/apps/viewer/src/hooks/useDrawingGeneration.ts
@@ -34,7 +34,7 @@ import {
   parseProfilesFlat,
   parseSymbolicFlat,
 } from '@/lib/overlay-parse/index.js';
-import { buildProfileEntries } from '@/lib/overlay-parse/profile-entries.js';
+import { buildProfileEntries, warnAboutSkippedProfiles } from '@/lib/overlay-parse/profile-entries.js';
 import {
   buildSymbolicDrawingLines,
   type SymbolicDrawingLine,
@@ -330,13 +330,12 @@ export function useDrawingGeneration({
           // `coordinateInfo` is main-thread state the worker cannot see.
           const ci = geometryResult.coordinateInfo;
           const rtc = ci.wasmRtcOffset;
-          const shift = rtc
-            ? { x: rtc.x, y: rtc.z, z: -rtc.y }
-            : ci.originShift;
+          const shift = rtc ? { x: rtc.x, y: rtc.z, z: -rtc.y } : ci.originShift;
           // Single-model (legacy) mode, so model index is always 0. Multi-model
           // profile extraction would require iterating over each model separately.
           profiles = buildProfileEntries(flat, shift, 0);
           profileCacheRef.current = { profiles, sourceId: modelCacheKey };
+          warnAboutSkippedProfiles(flat); // #3691-class silent-drop signal, one layer over
         } catch (error) {
           // Degrade gracefully: the drawing still renders without projection.
           console.warn('Profile extraction failed:', error);
--- a/apps/viewer/src/lib/overlay-parse/profile-entries.test.ts
+++ b/apps/viewer/src/lib/overlay-parse/profile-entries.test.ts
@@ -7,8 +7,8 @@ import assert from 'node:assert';
 
 import type { ProfileCollection } from '@ifc-lite/wasm';
 import type { ProfileEntry } from '@ifc-lite/drawing-2d';
-import { collectFlatProfiles } from './profiles-flat.js';
-import { buildProfileEntries } from './profile-entries.js';
+import { collectFlatProfiles, createEmptyFlatProfiles } from './profiles-flat.js';
+import { buildProfileEntries, warnAboutSkippedProfiles } from './profile-entries.js';
 
 /**
  * Equivalence anchor for moving the construction-projection extraction into
@@ -80,10 +80,14 @@ function entry(fields: {
   };
 }
 
-function makeCollection(entries: (FakeEntry | undefined)[]): ProfileCollection {
+function makeCollection(
+  entries: (FakeEntry | undefined)[],
+  skippedExpressIds: number[] = [],
+): ProfileCollection {
   return {
     length: entries.length,
     get: (i: number) => entries[i],
+    skippedExpressIds: Uint32Array.from(skippedExpressIds),
   } as unknown as ProfileCollection;
 }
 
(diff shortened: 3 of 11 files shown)


louistrue (commented):

- Ran the new integration test at this head: 3/3. The GREEN test passing is itself the proof the fixture exercises the drop path — if the broken wall (#500, SweptArea -> nonexistent #999999) were filtered out before extraction instead of failing inside it, skipped would be empty and dropped_wall_is_reported_with_default_features would fail. The healthy-file control puts pressure on the false-positive side.
- Ran cargo test -p ifc-lite-export at this head: full suite green — backs the claim that extract_profiles (now a thin wrapper returning .0) is behaviorally untouched for the HBJSON/rooms/openings/shades callers.
- Confirmed the feature-gating claim by reading rust/geometry/Cargo.toml (default = []) and scripts/build-wasm.sh: neither debug_geometry nor observability ships, so before this the drop really was invisible end to end. The new path is diagnostics-by-return-value, not another gated macro — the right shape.
- TS side read end to end: FlatProfiles.skippedExpressIds is populated once per collection (plain typed-array getter, no per-entry .free() hazard), structured-clone-safe across the worker boundary, and warnAboutSkippedProfiles has both a firing test and a no-op control.

Two non-blocking notes:
- The warn fires only on the non-cached path in useDrawingGeneration; a repeat generation served from profileCacheRef won't re-warn. Fine for a first signal, worth remembering if this ever feeds a UI badge.
- console.warn is the only user-facing surface for now — matches the PR's stated scope, but the wall is still only 'missing plus a console line' for a non-devtools user.

PR #3700, opened 2026-09-02
docs(export): pin the TS/Rust BOS Parquet export table-set gap with a test

BIMvoice:
## Summary

Hunt for untested TS/Rust behaviour twins (per the running parity-hunt lens). `rust/export/src/parquet_bos.rs` explicitly documents itself as porting `packages/export/src/parquet-exporter.ts` (the ara3d BOS/Parquet exporter). Running the same fixture through both confirms they actually diverge, and nothing tracked that.

**Executed, not inferred:** parsed `tests/models/ara3d/duplex.ifc` with `IfcParser.parseColumnar` and ran it through `ParquetExporter.exportBOS()`, and separately ran the existing `duplex_exports_valid_bos` Rust test (`export_bos()`) on the same fixture. Zip entries:

- TS (`packages/export/src/parquet-exporter.ts`): `Entities.parquet`, `Metadata.json`, `Properties.parquet`, `Quantities.parquet`, `Relationships.parquet`, `SpatialHierarchy.parquet`, `Strings.parquet`
- Rust (`rust/export/src/parquet_bos.rs`): `Entities.parquet`, `Properties.parquet`, `Quantities.parquet`, `VertexBuffer.parquet`, `IndexBuffer.parquet`, `Meshes.parquet`, `Metadata.json`

The Rust `.bos` path is missing `Relationships.parquet`, `Strings.parquet`, and `SpatialHierarchy.parquet` — a `.bos` archive exported through the CLI/native path silently drops relationship, string-dedup, and spatial-hierarchy data that a browser export of the identical model would carry. I could not find a comment or test marking this as deliberate, so it reads as an incomplete port rather than an intentional scope cut.

Implementing the three missing tables is a real feature addition (new Arrow schemas, more decoding), too large for this pass, so this PR documents the gap and locks it down with a test instead of silently letting it drift further:

- Module doc comment on `parquet_bos.rs` now states the gap explicitly, naming the fixture and the missing tables.
- `duplex_exports_valid_bos` now also asserts the three tables are absent, with a comment pointing at the doc comment — so if someone starts writing one of them, a test forces the doc comment to be updated (and, more importantly, someone notices the current call site's silent gap while it still doesn't).

No behavior change. Follow-up: port `Relationships.parquet` / `Strings.parquet` / `SpatialHierarchy.parquet` to `parquet_bos.rs`, or, if the omission turns out to be deliberate, replace this test with one that says so.

## Test plan
- [x] `cargo test --features parquet-bos --lib parquet_bos` (new assertions pass)
- [x] `cargo test --features parquet-bos` in `rust/export` (83 passed)
- [x] `cargo test --test module_size_ratchet -p ifc-lite-processing` (unaffected)
- [x] Manually ran `ParquetExporter.exportBOS()` on `tests/models/ara3d/duplex.ifc` via a scratch vitest test (not committed) to enumerate the TS-side table set

## Summary by CodeRabbit

- **Documentation**
  - Clarified that Rust-based exports do not currently include `Relationships.parquet`, `Strings.parquet`, or `SpatialHierarchy.parquet`.

- **Tests**
  - Added validation to ensure these three tables remain absent from Rust exports unless the behavior changes intentionally.

### DIFF
--- a/rust/export/src/parquet_bos.rs
+++ b/rust/export/src/parquet_bos.rs
@@ -3,6 +3,17 @@
 //! `packages/export/src/parquet-exporter.ts` (BIM Open Schema): Entities / Properties /
 //! Quantities + optional geometry (VertexBuffer / IndexBuffer / Meshes) + Metadata.json.
 //!
+//! **Known gap vs. the TS twin (unverified whether intentional):** on the same
+//! fixture (`tests/models/ara3d/duplex.ifc`), `ParquetExporter.exportBOS()` in
+//! `packages/export/src/parquet-exporter.ts` additionally writes
+//! `Relationships.parquet`, `Strings.parquet`, and — when the source store carries
+//! one — `SpatialHierarchy.parquet`. Those three tables are not produced here, so
+//! a `.bos` archive produced through the CLI/native path (this file) is missing
+//! relationship, string-dedup, and spatial-hierarchy data that a browser export of
+//! the identical model would contain. See `duplex_exports_valid_bos` below, which
+//! asserts the current (narrower) table set explicitly so this gap can't widen
+//! further without a test failure.
+//!
 //! Native-only (feature `parquet-bos`): parquet's native compression and the zip writer
 //! don't target wasm32 cleanly, and `.bos` isn't a browser-exposed format, so this stays
 //! out of the wasm bundle. Server / CLI builds opt in.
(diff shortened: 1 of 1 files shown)


louistrue (commented):
Reviewed against the code, not just the prose.

**Doc claim vs implementation, checked line by line.** `packages/export/src/parquet-exporter.ts` exportBOS (lines 78-93) writes Relationships.parquet and Strings.parquet unconditionally and SpatialHierarchy.parquet only when `this.store.spatialHierarchy` is set; `rust/export/src/parquet_bos.rs` (lines 259-286) writes only Entities/Properties/Quantities, the three geometry tables, and Metadata.json. The new module doc states exactly that, including the conditional phrasing for SpatialHierarchy. No overclaim.

**The tripwire can actually fail.** The three absence assertions run against the same `names` list the seven presence assertions already validated, so an empty or mis-read archive fails on the presence side first; the absence side only fires when someone starts writing one of the three tables. That is the intended trip direction, and the failure message points at the doc comment to update.

**Ran it** in a scratch worktree at this head: `cargo test --features parquet-bos --no-fail-fast` in rust/export, 386 passed, 0 failed, including `parquet_bos::tests::duplex_exports_valid_bos`.

No defects found.

PR #3702, opened 2026-09-02
test(processor,parser): pin space/zone property resolver's TS/Rust parity claim to a real harness

BIMvoice:
## Summary

A prior census flagged that `resolve_space_zone_properties_lazy`'s doc comment in `rust/processing/src/processor/properties.rs` claimed byte-identical "Parity" — but the comparison it actually verified was against this crate's OWN earlier eager decode path, never against the TypeScript on-demand property extractor (`packages/parser/src/columnar-parser.ts` `extractPropertiesOnDemand` + `property-value-parser.ts`) it is a twin of for `IfcSpace`/`IfcZone`.

**Scope established first.** The Rust side covers only `IfcSpace`/`IfcSpaceType`/`IfcZone`/`IfcSpatialZone`/`IfcSpatialZoneType`, only properties reached through a genuine `IfcPropertySet` (not `IfcElementQuantity`), only `IfcPropertySingleValue`/`IfcPropertyEnumeratedValue`/`IfcPropertyListValue`/`IfcPropertyBoundedValue`/`IfcPropertyReferenceValue` (no `IfcPropertyTableValue`/`IfcComplexProperty`), and outputs a `BTreeMap<String, String>` (a display map for `MeshData::properties`), not a typed store.

**Executed comparison** — one fixture (`rust/processing/tests/fixtures/space_zone_property_type_vectors.json`, an `IfcSpace` with two `IfcPropertySet`s) run through both real extraction paths end to end:

| Property | TS value (type) | Rust printed value | Agree? |
|---|---|---|---|
| TextProp (`\S\` escape) | `"Cafå"` (string) | `"Cafå"` | yes |
| RealProp | `3.5` (Real) | `"3.5"` | yes (printed form) |
| IntProp | `4` (Integer) | `"4"` | yes (printed form) |
| BoolProp | `true` (Boolean) | `"T"` | printed text diverges — Rust has no typed output, stringifies the bare enum token |
| LogicalProp (`.U.`) | `null` (Logical) | `"U"` | same as above |
| EnumProp | `"Red"` | `"Red"` | yes |
| AbsentProp (`$`) | listed, `value: null` | key omitted entirely | structural difference, not a value disagreement |
| UnicodeProp (`\X2\00E9\X0\`) | `"café"` | `"café"` | yes |
| TextProp in PsetB (same name, 2nd pset) | both psets kept separate, no collision | unscoped key first-wins (PsetA); both `PsetA.TextProp`/`PsetB.TextProp` scoped keys still resolve correctly | no data loss via scoped keys |

Ran (not read): `cargo test --test space_zone_property_type_vectors` (4 passed) and `npx vitest run src/space-zone-property.parity.test.ts` (5 passed) in `packages/parser`, both against the identical shared fixture text.

**No live user impact found**: `MeshData::properties` currently has no downstream reader anywhere in the Rust workspace — the wasm bridge (`rust/wasm-bindings/src/zero_copy/mesh.rs`) explicitly drops it, noting the viewer gets properties from the TS parser worker instead. So the boolean/logical stringification and the dropped-$-key difference are pinned as documented, tested divergences rather than "fixed" — there's nothing today for a user to see disagree, and unifying the two implementations is out of scope for this task.

**Doc comment corrected** to state what is actually verified (eager-vs-lazy Rust parity) and what is NOT (TS parity), with the covered/uncovered scope spelled out and a pointer to the new tests.

## Test plan
- [x] `cargo test --test space_zone_property_type_vectors` (rust/processing) — 4 passed
- [x] `cargo test --test module_size_ratchet` (rust/processing) — 6 passed
- [x] `cargo test` (rust/processing, full suite) — all passed
- [x] `npx vitest run src/space-zone-property.parity.test.ts` (packages/parser) — 5 passed
- [x] `npx vitest run` (packages/parser, full suite) — 969 passed / 2 skipped
- [x] `npx tsc --noEmit` (packages/parser, via turbo build of its deps) — clean
- [x] `node scripts/check-module-size.mjs` — no new findings
- [x] `node scripts/check-unused-locals.mjs` — parser package clean once built
- [x] `pnpm run check:vitest-timeout-audit` — no new unprotected tests
- [x] No `Cargo.lock` churn in the diff


louistrue (commented):
Reviewed and re-ran both halves in a scratch worktree at this head.

**Executed, both sides**: `cargo test --no-fail-fast` in rust/processing - 306 passed, 0 failed, including the 4 new `space_zone_property_type_vectors` tests; `vitest run` in packages/parser - 969 passed / 2 skipped, including the 5 new parity tests. Both halves read the one shared fixture, and the expected values are hardcoded literals on each side rather than derived from either implementation, so neither half can inherit the other's defect.

**Doc claims checked against the code, not the PR prose**:
- The type assertions match `PropertyValueType` in packages/data/src/types.ts:179-183 (Real=1, Integer=2, Boolean=3, Logical=4).
- "MeshData::properties has no downstream reader": grepped the workspace - the only touch points are the setter (rust/processing/src/types/mesh.rs:304) and the wasm bridge, whose `from_mesh_data` doc and body (rust/wasm-bindings/src/zero_copy/mesh.rs:428-432) drop it exactly as the corrected doc comment says.
- The fixture description's reference to commit 9368b2dcd resolves to the same-named-pset query fix it describes.

**One note, non-blocking**: the TS half's `describe.skipIf(!existsSync(fixturePath))` means a future move/rename of the Rust-side fixture silently skips the TS half rather than failing it. Mitigated in practice because the Rust half's `include_str!` breaks the build on the same move, forcing whoever moves it to touch both paths.

No defects found.

PR #3703, opened 2026-09-02
fix(mcp): match bcf_topic_list's status filter case-insensitively

BIMvoice:
## Reviewer summary

- **Without this:** an MCP agent filtering BCF topics by status with different casing than the file uses (e.g. `'open'` vs. the on-disk `'Open'`) gets an empty list back — indistinguishable from "no matching topics," with no human looking at a 3D view to catch the false negative.
- **Evidence:** on the unmodified tool, filtering `status: 'open'` against a topic created with the default `'Open'` status returned 0 topics; after the fix, it returns 1.
- **Risk:** narrow, single-comparison change, matching a lowercasing convention `@ifc-lite/bcf`'s own `computeMarkers3D` already uses elsewhere in the repo. Control: `status: 'closed'` against the same fixture still correctly returns 0 — the fix isn't just returning everything.
- **Sequencing:** none; based on `main`.

---

## Summary

- `bcf_topic_list`'s `status` filter did an exact, case-sensitive `===` match against `topicStatus`. BCF status strings are conventionally Title Case (`'Open'`, `'Closed'`), the `bcf_topic_create` schema even defaults `status` to `'Open'`, but nothing tells a calling agent the exact casing to use. A differently-cased filter (e.g. `status: 'open'`) silently returned an empty list — indistinguishable, from the caller's side, from "there really are no matching topics".
- `@ifc-lite/bcf`'s own `computeMarkers3D` already lowercases both sides for its equivalent status filter, so this was also a same-repo behavioural inconsistency, not just an isolated bug.
- Fix: lowercase both sides before comparing, matching the existing `computeMarkers3D` convention.

## Context

Found auditing the MCP tool surface (the set of tools an AI agent calls to answer questions about an IFC model, with no human looking at a 3D view to sanity-check the answer) for tools that can return a confidently wrong or misleadingly-empty result. `bcf_topic_list` had no test coverage at all prior to this PR.

## Test plan

- [x] `packages/mcp/src/tools/bcf.test.ts` (new): RED on unmodified `bcf.ts` — `status: 'open'` against a topic created with the default `'Open'` status returned 0 topics; GREEN after the fix — returns 1. A control case (`status: 'closed'` against the same fixture) still correctly returns 0, proving the fix doesn't just return everything.
- [x] `pnpm --filter @ifc-lite/mcp test` — 36 files / 351 tests pass.
- [x] `pnpm --filter @ifc-lite/mcp exec tsc --noEmit` — clean.
- [x] `node scripts/check-module-size.mjs`, `node scripts/check-test-wiring.mjs`, `node scripts/check-unused-locals.mjs`, `pnpm run check:vitest-timeout-audit` — no new findings for the touched files.
- [x] Changeset added for `@ifc-lite/mcp`.

### DIFF
--- a/.changeset/mcp-bcf-status-filter-case.md
+++ b/.changeset/mcp-bcf-status-filter-case.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/mcp': patch
+---
+
+Fix `bcf_topic_list`'s `status` filter using an exact, case-sensitive match against `topicStatus`. BCF status strings are conventionally Title Case (`'Open'`, `'Closed'`), but nothing tells a calling agent that, and a differently-cased filter (e.g. `'open'`) silently returned zero topics — indistinguishable from "there really are none". `@ifc-lite/bcf`'s own `computeMarkers3D` already lowercases both sides for its status filter; `bcf_topic_list` now matches that.
--- a/packages/mcp/src/tools/bcf.test.ts
+++ b/packages/mcp/src/tools/bcf.test.ts
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * bcf_topic_list status filter.
+ *
+ * BCF status strings are conventionally Title Case ('Open', 'Closed'), but
+ * nothing in the tool's schema or description tells an agent that, and an
+ * agent has no independent way to check a topic's exact casing before
+ * filtering. An exact `===` match silently returned an empty list for a
+ * differently-cased filter — identical, from the caller's side, to "there
+ * really are no matching topics" (the error-vs-empty distinction this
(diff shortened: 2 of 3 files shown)


louistrue (commented):
Reviewed and exercised in a scratch worktree at this head.

**The fix is what it says.** Lowercasing both sides matches the convention `@ifc-lite/bcf` already uses for its own status filter (`overlay.ts:298-299` in `computeMarkers3D`), and `(t.topicStatus ?? '')` covers the type's optional `topicStatus?: string`.

**The test discriminates.** Reverted the filter to the old `===` and re-ran `bcf.test.ts`: the case-insensitivity test goes red (0 topics for `'open'`), the control stays green. Restored, both pass.

**Boundary probes** (ran, not reasoned): `'OPEN'` and `'oPeN'` match; `' open '` does not (whitespace untrimmed, same as `computeMarkers3D`, fine); `''` still drops the filter entirely and returns every topic. That empty-string behaviour is pre-existing (`!filter` did the same before) and unchanged by this PR, so no regression - just noting a topic with a blank status can never be selected by an explicit filter.

**Full suite**: 36 files / 351 tests pass in `packages/mcp`, matching the PR's claim.

No defects found.

PR #3705, opened 2026-09-02
fix(geometry): cap try_cut_wall_local_frame recursion at one rotation per host

### LINKED ISSUE #3641: WASM trap "memory access out of bounds" when subtracting 3 IfcVoidingFeature voids from one host (sequential CSG chain)
### Version

Reproduced on `@ifc-lite/geometry` 4.1.0 + `@ifc-lite/wasm` 6.1.1 (latest published), and also on `@ifc-lite/geometry` 3.8.1 + `@ifc-lite/wasm` 4.4.1.

### Which surface

Geometry / CSG kernel

### Reproduction

Minimal repro (5.9 KB IFC, 93 entities, extracted from a real model) plus a runner script:
https://gist.github.com/vladmihai-del/f58cbef6ec73eb62c8173d134f00176a

```bash
npm install @ifc-lite/geometry @ifc-lite/wasm @ifc-lite/data
node repro.mjs minimal-voiding-trap.ifc
```

Output:

```
trap after 0 meshes: RuntimeError - memory access out of bounds
```

The file contains one host (`IfcSlab`, an `IfcExtrudedAreaSolid` with an `IfcIndexedPolyCurve` profile) and three `IfcVoidingFeature` voids attached via `IfcRelVoidsElement`, all three also extruded solids (depths 3150 / 1800.7 / 25632 mm, extrusion direction `(1.E-09, 3.E-10, -1.)`). Schema `IFC4X3_ADD2`, authored by IfcOpenShell 0.8.0 / Bonsai 0.8.0.

What I narrowed down by calling `buildPrePassOnce` + `processGeometryBatch` directly and filtering the `voidKeys/voidCounts/voidValues` arrays:

- each of the three voids subtracted **alone**: OK
- every **pair** of voids: OK
- **all three together**: trap — so the failure is in the sequential Boolean chain, presumably an intermediate result degenerating, not in any single subtraction
- the full 13 MB source model (8832 jobs) passes cleanly end-to-end when the void arrays are passed empty, so all other geometry is healthy

In the original model the trap reproduces identically on a second host (an `IfcCovering` with a tessellated body) that the pre-pass maps to the same void set, so it is not specific to the host's solid type.

### Expected vs actual

**Expected:** the void subtraction either succeeds or degrades to a logged CSG failure (the pipeline already has that mechanism — "Opening pipeline: N CSG failures" — and `huge-file-error.ts` documents the panic-vs-trap caveat), leaving the host un-voided but the engine alive.

**Actual:** the wasm code takes a `RuntimeError: memory access out of bounds` trap. Per the recovery contract in `wasm-runtime-trap.ts` the trap corrupts the module's linear memory: in a browser host every subsequent `init()` then traps as well, so the document ends at `WASM_RUNTIME_UNRECOVERABLE` and a single bad element takes down the whole multi-model federation for the user.

### Environment

Node 26.4.0 on Windows 11 (repro above); originally hit in Chrome (viewer host app). Behaviour identical in both.

### Privacy

- [x] My attached files and text contain no confidential or client-identifying data.

BIMvoice:
## Summary

- Fixes a WASM trap (`RuntimeError: memory access out of bounds`) when subtracting 3+ `IfcVoidingFeature` voids with conflicting rotation axes from one host — a sequential-CSG-chain crash that corrupts the module for the rest of the session.
- Root cause: unbounded mutual recursion between `try_cut_wall_local_frame` and `apply_void_context_inner` in `rust/geometry/src/router/voids/mod.rs`, not an out-of-range index in a single boolean op.

## Root cause

`try_cut_wall_local_frame` rotates a plan-rotated wall into its own axis-aligned frame and recurses into `apply_void_context_inner` to run the exact cut there. Its doc comment claimed this recursion was self-limiting: *"in the frame every opening's depth is +Z (axis-aligned), so this guard returns `None` on the inner call."* That is only true when **every** opening reclassifies as frame-aligned `Rectangular`. An opening that stays `NonRectangular` keeps its **own** depth direction, re-expressed in the new frame — and that direction can itself be non-axis-aligned and near-horizontal. `try_cut_wall_local_frame` then fires again, seeding a **second** frame from a **different** opening, and recurses. With three voids on one host, two of them disagree on rotation axis enough to keep re-triggering this — unbounded.

## Reproduction (RED, native — confirms it is NOT wasm-only)

Fixture: the reporter's minimal 5.9 KB IFC (`rust/processing/tests/fixtures/issue_3641_voiding_trap.ifc`) — one `IfcSlab` host with 3 `IfcVoidingFeature` voids.

Before the fix, running the native (non-WASM) pipeline directly:
```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow, aborting
```
An `lldb` backtrace on the crash confirmed hundreds of alternating stack frames:
```
try_cut_wall_local_frame -> apply_void_context_inner -> try_cut_wall_local_frame -> apply_void_context_inner -> ...
```
So this reproduces identically without WASM at all — it's a genuine unbounded-recursion defect, not an allocator/memory-layout quirk or a 32-bit index overflow. In WASM the same overrun runs past the stack into linear memory (no guard page there), which is why it surfaces as `RuntimeError: memory access out of bounds` instead of a clean stack-overflow signal, and why it corrupts the module for every later call.

## Fix

`apply_void_context_inner` now takes an explicit `allow_local_frame: bool`:
- `true` only on the outer call from `apply_void_context`.
- forced `false` on the recursive call `try_cut_wall_local_frame` makes into it.

This makes "at most one local-frame rotation per host" an explicit invariant instead of an assumed one. A later, differently-oriented opening still gets subtracted — via the same exact/world-path kernel that `NonRectangular` openings already use — just without the second rotation's f32-precision benefit. **No void is silently dropped**; the host is still fully cut.

## GREEN + control

- `issue_3641_voiding_trap.rs` (new): the same fixture now processes cleanly — `4 meshes, 1 CSG failure logged` (one opening degrades to a logged CSG failure per the pipeline's existing "Opening pipeline: N CSG failures" mechanism, which is the documented expected-degradation contract in the issue) — and asserts no mesh index is out of range.
- Control — ordinary multi-void sequential subtraction unaffected: `issue_2019_void_order_invariance.rs` (pre-existing, exercises a 3-void sequential chain in two different orders) still passes, with identical hashes for reordered-but-associative statements and the expected non-associative-order difference otherwise preserved.
- `cargo test -p ifc-lite-geometry --lib`: 763 passed, 0 failed.
- `cargo test -p ifc-lite-processing` (full crate, 60 test binaries): all green, 0 failed.
- `module_size_ratchet`: green — `rust/geometry/src/router/voids/mod.rs` held at its existing 2058-line budget (no ratchet raised; comments were tightened to make room for the fix).
- `cargo check -p ifc-lite-wasm`: compiles clean (private-fn signature change, no downstream API impact).

## Sibling check

`zsh ~/.claude/scripts/ifclite-sibling-sweep.sh try_cut_wall_local_frame` — single definition, single call site (recursive), no siblings implementing the same pattern. Grepped `rust/geometry/src` for other "recursion is safe because…" style comments; the only other hit is in `profile_extractor.rs`, an unrelated file already owned by open PR #3694.

The two other self-recursions inside `apply_void_context` (2D-opening residual, prism-cut residual) were reviewed: each recurses on a **residual** `VoidContext` built once from the classified-ineligible opening IDs, which does not carry a `bool2d`/prism classification, so it cannot re-enter the same fast path that produced it — these are bounded by construction and out of scope for this fix.

## Related but NOT addressed here

Issue **#3353** ("Booleans still tear on overlapping and rotated operands…") is a separate open boolean-correctness issue. Nothing in this investigation bears directly on it — the #3641 defect is a recursion/control-flow bug in the local-frame *routing*, not a tessellation/precision defect in the boolean kernel itself.

## Base

Branched off `upstream/main` at `bc0906e18053717b184a06ef022abee4012c4f49`.

Closes #3641

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WASM memory-access traps when applying multiple voids with conflicting rotation axes.
  * Prevented unbounded recursion during local-frame geometry processing.
  * Preserved accurate processing for differently oriented openings.

* **Tests**
  * Added coverage for sequential void subtraction, verifying processing completes without panics or invalid mesh data.


louistrue (commented):
Reviewed the diff and re-ran the evidence locally (macOS, scratch worktree at bd4ae1d85):

- RED confirmed by mutation: reverted rust/geometry/src/router/voids/mod.rs to the merge-base with the PR's fixture/test kept — native run aborts with `thread has overflowed its stack` (SIGABRT), exactly the mutual recursion described. Restored the fix: `cargo test -p ifc-lite-processing --test issue_3641_voiding_trap` passes, output `processed: 4 meshes, 1 CSG failures logged` matching the PR body. So the test genuinely fails without the fix and cannot pass vacuously (a skipped fixture would hit the `expect`, and an overflow aborts the process).
- Recursion structure checked by reading the code, not just the PR text: `apply_void_context_inner` re-enters the outer `apply_void_context` only via `try_cut_wall_local_frame`, which is the edge the flag severs. The two residual recursions (bool2d at mod.rs:943, prism at mod.rs:987) build their contexts with `param: None, bool2d: None` (prism_cut.rs:~2945) and go through the outer entry, so they can pick up at most one more local-frame rotation each on a strictly shrinking opening set — bounded, as the PR claims.
- Cap direction: at the cap the opening is not dropped; it routes to the exact/world kernel and the one degradation is a logged CSG failure through the existing per-host diagnostics channel — loud, not silent truncation.
- Changeset: patch on @ifc-lite/geometry is right for a crash fix with no API change.

Issue #3671, opened 2026-09-02
Reporting World Coordinates in Lists


gDonia87:
### The problem

If a model contains objects, which are far away from the project point, I have to find them not only visually but also through unambiguous information. 

### Proposed solution

Could it be possible to show in a list the World Coordinates values of every single object? So that we can sort the column for this property and find out immediately if there are objects far away in the model.

<img width="793" height="205" alt="Image" src="https://github.com/user-attachments/assets/33f3d386-429a-46be-b7b3-2804dd903568" />

Of course X,Y,Z should be separete numerical properties.

Issue #2867, opened 2026-08-19
IDS-Test - activated object often not identificabile


don-matteo:
### The problem

I am trying the IDS-Checker in ifclite.
Often when i try to activate an object, I can only locate it with problems.

https://github.com/user-attachments/assets/19b07b63-b438-46f2-a28c-286a9b9ae0ff

### Proposed solution

An option to isolate the story where the object is located in and maybe show all other objects transparently could be a solution

BIMvoice:
Traced this before touching anything. **The label is right — it is an enhancement, not a validation or ID-resolution bug** — and the useful finding is that the Clash panel already solved the same problem, so there is a precedent to copy rather than a design to invent.

## What happens today

Clicking a result row calls `selectEntity` (`apps/viewer/src/hooks/useIDS.ts:477-504`): it sets `idsActiveEntity`, sets store selection, and after 50 ms calls `cameraCallbacks.frameSelection()`. That is the entire per-row action.

No isolation, no ghosting, and no highlight of its own — the activated entity keeps whatever colour the report already gave it, red-if-failed or green-if-passed via `buildColors` (`useIDS.ts:539-560`). So **a freshly activated failing element looks identical to every other failing element around it**, and if it sits inside or behind other geometry the camera moves to it while nothing makes it distinguishable or unoccluded. That matches the reporter's description closely.

IDS does have isolation — `isolateFailed` / `isolatePassed` / `isolateInvolved` (`useIDS.ts:639-734`, buttons at `IDSPanel.tsx:825-880`) — but those isolate the whole failed/passed/involved set for the report or active spec. There is no single-entity equivalent; confirmed by grep across `useIDS.ts`, `IDSPanel.tsx` and `idsSlice.ts`.

## Clash already does what is being asked for

`focusClash` (`apps/viewer/src/hooks/useClash.ts:776-847`) is the same "activate one row" action, and it carries:

- a persistent `ClashFocusMode` of `'highlight' | 'isolate' | 'ghost'` (`applyFocusMode`, `useClash.ts:755`), chosen by the user at `ClashPanel.tsx:311`
- distinct amber/cyan colours through the colour-override channel, not just a selection outline
- framing of the tight contact region rather than the whole element (`clashFramingBounds`, `useClash.ts:836-842`)

IDS's activate is roughly what Clash's was before it grew focus modes.

## Two options — your call, nothing shipped

[...]

PR #3709, opened 2026-09-02
feat(viewer): "Show in context" — fade other geometry around a selected entity

### LINKED ISSUE #3618: Navigation of the Objects with the function "Entity Lists"
### The problem

If I want to verify the object of a row in a list, I can find it with the function "Zoom to" (after the selection with the information tab), but I can´t find the object if it is behind other objects.
I can isolate it with the shortcut "+" but I lose the context.

### Proposed solution

Is it possible to show the selected object with the other objects in transparent mode?  

### Alternatives considered

Maybe it could also be helpful to have the possibility of generating a section box around the selected object

BIMvoice:
Closes #3618

## Reading of the issue

The user's workflow: select a row in an Entity List, then use the info tab's "Zoom to" to find that object in 3D. When the object is occluded by other geometry, "Zoom to" moves the camera but the nearer geometry still renders in front, so the object stays invisible. Isolate (the `I` shortcut) solves visibility but hides everything else, which loses the surrounding context the user wants to keep. The issue's proposed solution is explicit: "show the selected object with the other objects in transparent mode."

**Interpretation chosen**: add a "Show in context" action next to the existing Zoom to / Hide buttons in the Properties panel header (the "information tab" the issue refers to). It fades every other entity translucent and frames the camera on the selected one in the same click, so the object is visible *through* the rest of the model instead of disappearing behind it or isolating it away from its surroundings. This reuses the store's existing `ghostExceptEntities` X-Ray channel, which Clash, IDS and BCF already drive for exactly this "focused subset, rest still visible" presentation — so the render-side behaviour is already proven, not new.

The issue's alternative ("a section box around the selected object") is not implemented: a general-purpose Section tool with cardinal and custom planes already exists in this viewer, so a per-entity auto section box would be new surface for a need the ghost/X-ray toggle already covers more directly. Happy to add it separately if the maintainer still wants it after seeing this.

## Files touched

- `apps/viewer/src/components/viewer/properties/EntityHeaderActions.tsx` (new) — Zoom to / Show in context / Hide, extracted out of `PropertiesPanel.tsx`.
- `apps/viewer/src/components/viewer/properties/EntityHeaderActions.test.tsx` (new) — drives the real component + store; includes a control proving Zoom to / Hide behaviour is unchanged.
- `apps/viewer/src/components/viewer/PropertiesPanel.tsx` — the three header buttons replaced with `<EntityHeaderActions />`.

**Why extracted rather than edited in place**: `PropertiesPanel.tsx`'s `scripts/module-size-allowlist.txt` budget is 2205, and the file was at exactly 2205 lines before this change — any net-positive edit would fail `check-module-size.mjs`, which this repo's house rules forbid raising. Net effect on `PropertiesPanel.tsx` is -43 lines (2205 → 2162).

**Overlap note**: none of these three files, nor `visibilitySlice.ts` (unedited — `setGhostExceptEntities` already existed and is called as-is), appear in the open-PR file-overlap check for `apps/viewer`. A sibling agent is reportedly working on a Lists world-coordinates column (issue #3671) concurrently; this PR does not touch any Lists column/registry file (`apps/viewer/src/components/viewer/lists/**`, `apps/viewer/src/lib/units/list-column-units.ts`), so there should be no collision.

## What I ran

- `pnpm typecheck` (turbo, all 89 tasks + `typecheck-tests.mjs` across 48 packages) — clean.
- `node scripts/check-module-size.mjs` — OK, 0 new over budget (only pre-existing informational headroom notes).
- `node scripts/check-unused-locals.mjs` — no new unused locals.
- `pnpm run check:vitest-timeout-audit` — passes (exit 0).
- `pnpm lint` (oxlint) — 0 errors (5 pre-existing warnings elsewhere, unrelated).
- New test file directly via `tsx --test`: 4/4 passing, including a control that "Zoom to" and "Hide" behave exactly as before.
- All other `apps/viewer/src/components/viewer/properties/*.test.{ts,tsx}` (12 files, 101 tests) — all passing.
- `apps/viewer/src/components/viewer/PropertiesPanel.group-isolate.test.tsx` (6 tests) — all passing, confirming the adjacent isolate-in-3D flow is unaffected.

Base: `upstream/main` @ `bc0906e18053717b184a06ef022abee4012c4f49`.

TypeScript-only change; no Rust/WASM touched. `apps/viewer` is private (no changeset needed).

### DIFF
--- a/apps/viewer/src/components/viewer/PropertiesPanel.tsx
+++ b/apps/viewer/src/components/viewer/PropertiesPanel.tsx
@@ -6,9 +6,6 @@ import { useMemo, useState, useCallback, useEffect } from 'react';
 import {
   Copy,
   Check,
-  Focus,
-  EyeOff,
-  Eye,
   Building2,
   Layers,
   Layers2,
@@ -62,6 +59,7 @@ import { BsddCard } from './properties/BsddCard';
 import { GeoreferencingPanel } from './properties/GeoreferencingPanel';
 import { RawStepCard } from './properties/RawStepCard';
 import { UnitDisplayControl } from './properties/UnitDisplayControl';
+import { EntityHeaderActions } from './properties/EntityHeaderActions';
 import { TOUR_ANCHORS, tourAnchor } from '@/lib/tours/anchors';
 import { isMaterialDefinitionType } from '@/utils/materialDefinitionTypes';
 
@@ -150,8 +148,6 @@ export function PropertiesPanel() {
   const zoneAssignments = useViewerStore((s) => s.zoneAssignments);
   const selectedModelId = useViewerStore((s) => s.selectedModelId);
   const cameraCallbacks = useViewerStore((s) => s.cameraCallbacks);
-  const toggleEntityVisibility = useViewerStore((s) => s.toggleEntityVisibility);
-  const isEntityVisible = useViewerStore((s) => s.isEntityVisible);
   // Relationship navigation: select a related entity (e.g. an IfcZone) to show
   // its attributes, or isolate a group's members in 3D (#1075).
   const setSelectedEntity = useViewerStore((s) => s.setSelectedEntity);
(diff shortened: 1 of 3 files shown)


louistrue (commented):
Reviewed the diff and ran it locally (scratch worktree at 2207d8277, fresh install + build):

- The new test file passes 4/4 driving the real component and store. The load-bearing assertion is the store round-trip (ghostExceptEntities === [selectedEntityId], isolatedEntities untouched, frameSelection called once) so a mis-wired id or channel fails it.
- Id-space check: ghostExceptEntities is consumed by the renderer alongside hiddenEntities/isolatedEntities (useAnimationLoop passes all three), and the pre-existing Hide button already feeds selectedEntityId into hiddenEntities — same global-id space, so the new writes are consistent.
- React hazard sweep (this repo has no exhaustive-deps lint): EntityHeaderActions uses no useCallback/useMemo/useEffect at all — every handler reads fresh store values from the hooks on each render, so there is nothing to freeze.
- Direct writes to setGhostExceptEntities are a designed entry point: the visibility slice's ownership-invalidation pass (see visibilitySlice.ts comments around #2867/#2662) handles a raw channel write stomping a clash/IDS-installed ghost by invalidating the stale record, and the third test pins the replace-not-merge behaviour deliberately.
- Cross-PR check with #3711 (both edit PropertiesPanel.tsx): git auto-merges cleanly and both PRs' test files pass together on the merged tree (8/8).

One inherited nit, not a blocker: the onClick uses `if (!selectedEntityId)` (falsy) while isGhosted uses `!= null` — same mix the pre-existing buttons had, and id 0 is not a reachable global id here.

Issue #3338, opened 2026-08-27
Assembly drill-down: expansion is one call site every channel must remember to use


louistrue:
Expanding an `IfcElementAssembly` into its `IfcRelAggregates` children works in some channels and not others, and the mechanism that does it has one production call site that every channel has to remember to route through. Writing down what exists, because the knowledge is currently spread across two merged PRs, two unmerged branches and a couple of session transcripts.

## What exists

`expandToGeometryBearingIds` — `apps/viewer/src/utils/aggregation.ts:158` — maps a set of ids to the geometry-bearing ids beneath them, dropping a container that resolves to nothing.

It has exactly **one** production call site: `Viewport.tsx:1020`, inside `resolveHighlightIds`. Every other reference in the tree is a test or a comment pointing at it. So a channel gets assembly expansion if and only if it routes through `resolveHighlightIds`.

Three wiring tests pin that routing (`LensPanel.isolate.wiring.test.tsx`, `SearchModal.filter.wiring.test.tsx`, `PropertiesPanel.group-isolate.test.tsx`), and `aggregation.test.ts:189-209` adds a source-text assertion that the resolver body actually calls the helper rather than merely naming it.

## The failure mode, which has already happened once

#2531 fixed a blank view for one path. #2532 added a second path nine hours later that wrote raw globalIds instead of routing through the resolver, so isolating a filter result that matched an `IfcElementAssembly` blanked the view. The two merged nine hours apart and **no CI run ever contained both sides**.

The shape generalises: any new selection or isolation channel is correct only if its author knows to call the resolver. Nothing structural enforces it — the source-text assertion guards the existing resolver, not a new caller.

## Unmerged work that already exists

- `test/assembly-hierarchy-no-representation` (Petru Conduraru, 8 Aug) — +145 lines in `treeDataBuilder.test.ts`, confirming a no-geometry `IfcElementAssembly` already nests its parts in the hierarchy panel, and that an empty assembly renders as a leaf rather than vanishing. Never opened as a PR.
- `hierarchy-tree-coverage` — coverage for `useHierarchyTree` helpers. 342 commits behind main, no PR.

## A constraint from downstream

[...]

PR #3711, opened 2026-09-02
feat(viewer): show and select the parent IfcElementAssembly

### LINKED ISSUE #3620: Selecting an IfcElementAssembly
### The problem

If I select an element which is contained in an IfcElementAssembly, I can´t  notice this relationship and I can´t select the parent Ifc entity (in this situation, the IfcElementAssembly) 

### Proposed solution

In the information-tab near the IfcClass, it could be helpful to see something that shows us that this element is part of an IfcElementAssembly, and we could click on this to select properly the IfcElementAssembly.

### Alternatives considered

_No response_

BIMvoice:
## Summary

Closes #3620.

The reporter selects an element that is a member of an `IfcElementAssembly` (via `IfcRelAggregates`) and gets no indication of that relationship anywhere near the IfcClass label, and no way to select the assembly itself. This PR implements exactly the workflow described in the issue:

- The properties panel's entity header now shows a "Part of Assembly" badge, right below the GlobalId / spatial-location row, whenever the selected element is decomposed by an `IfcElementAssembly`. It reads the parent via the existing `EntityNode.decomposedBy()` (walks the `IfcRelAggregates` inverse edge) and gates on the parent's type.
- Clicking the badge selects the assembly.

### The geometry-less assembly case

An `IfcElementAssembly` almost always owns no representation of its own — its meshes hang off the aggregated parts — so this is exactly the case the issue lives in. The click path does **not** touch isolation/highlight geometry resolution at all (out of scope — see below); it is a plain selection, which the properties panel renders from the entity's attributes/psets, not from its mesh. The one thing that does need to be geometry-aware is `selectedEntityId` (the globalId every 3D-pick/search entry point sets first, and the field this panel's own render gate keys off): the existing `handleSelectRelatedEntity` (used by the Relationships card for Voids/Fills/Groups/Connections) clears it to `null` as a side effect of `setSelectedEntityIds([])`, which would have collapsed the panel to the model-metadata fallback view after every click. I did not reuse it; the new `handleSelectAssembly` sets `selectedEntityId` explicitly via `toGlobalIdFromModels`, matching the pattern every other selection entry point (`Viewport.tsx`'s `handlePickForSelection`, search, hierarchy tree) already uses. `handleSelectRelatedEntity` itself is untouched — that's a separate, pre-existing gap outside this issue's scope.

A regression test drives this with a real parsed `IfcElementAssembly` that owns no geometry, and asserts the panel renders the assembly's own name/class correctly after the click — not just that the store's `selectedEntity` field changed.

### Scope note — #3389 / #3338 (assembly-expansion resolver)

I read #3389 (open, routes the viewer-embed `ISOLATE` channel through the assembly-expansion resolver) and #3338 (the systemic gate) before starting. This PR does **not** touch `isolateEntities`, `resolveHighlightIds`, or any isolation/highlight channel — a plain `setSelectedEntity`/`setSelectedEntityId` selection is a different code path with its own existing, working geometry-less-assembly handling in `Viewport.tsx`'s `frameSelection`/`resolveHighlightIds` (already on `main`, independent of #3389). No overlap with #3389's allowlist or gate.

### Module-size note

`PropertiesPanel.tsx` is allowlisted at its exact current line count (0 headroom), so it could not grow at all. I extracted the existing inline "spatial location" badge (storey name/elevation/height) into its own sibling component, `properties/SpatialLocationBadge.tsx`, which freed enough room for the new `properties/AssemblyBadge.tsx` feature while keeping `PropertiesPanel.tsx` at 2203 lines (budget 2205). Both new files are well under the 400-line new-file threshold.

## What I ran

- `pnpm turbo typecheck --filter=@ifc-lite/viewer` — clean.
- `node scripts/check-module-size.mjs` — OK, 0 new files over 400, `PropertiesPanel.tsx` shrank from 2205 to 2203.
- `node scripts/check-unused-locals.mjs` — pre-existing `packages/embed-sdk` "does not compile standalone" failure only; unrelated to this diff.
- `pnpm run check:vitest-timeout-audit` — no new unprotected/unknown entries from this diff.
- Targeted test run: the new `PropertiesPanel.assembly-select.test.tsx` (4/4 pass) plus every existing `PropertiesPanel`/`properties/*` test file (110/110 pass, including `PropertiesPanel.group-isolate.test.tsx`, the existing geometry-less-assembly isolation regression test) — no regressions.

## Test plan

- [x] Select a member of an `IfcElementAssembly`: badge appears naming the assembly.
- [x] Click the badge: panel now shows the assembly's own attributes/class, even though the assembly owns no geometry.
- [x] Control: select an ordinary element in no assembly — no badge, selection behaves exactly as before.

### DIFF
--- a/apps/viewer/src/components/viewer/PropertiesPanel.assembly-select.test.tsx
+++ b/apps/viewer/src/components/viewer/PropertiesPanel.assembly-select.test.tsx
@@ -1,0 +1,149 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Issue #3620: selecting a member of an IfcElementAssembly gave no
+ * indication of that relationship near the IfcClass label, and no way to
+ * jump to the assembly itself. This pins the fix: a "Part of Assembly"
+ * badge appears when the selected element is decomposed by an
+ * IfcElementAssembly (via IfcRelAggregates), and clicking it selects the
+ * assembly -- including when the assembly itself owns no geometry, which is
+ * the normal case (its meshes hang off the aggregated parts, not the
+ * assembly entity).
+ *
+ * Driven the way a user drives it: mount the real panel over a REAL parsed
+ * store (the panel walks a `ModelQuery`, which a cast stub cannot satisfy --
+ * AGENTS.md), click the badge, and read the store back.
+ */
+
+import '@/test/setup-dom.js';
+
+import { after, afterEach, before, describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { render, cleanup, click } from '@/test/render.js';
+import { useViewerStore } from '@/store';
+import { IfcParser, type IfcDataStore } from '@ifc-lite/parser';
+import { PropertiesPanel } from './PropertiesPanel.js';
+
+const MODEL_ID = 'm1';
+const ID_OFFSET = 1_000_000;
+
+/**
+ * `#50 Assembly A` decomposes into the two columns `#60`/`#61` via
+ * IFCRELAGGREGATES -- and, like every IfcElementAssembly, carries no
+ * IfcProductDefinitionShape of its own; only the columns render. `#42 Wall A`
+ * is an ordinary element in no aggregation at all -- the control.
+ */
+const MINI_IFC = `ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('t','',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=IFCPROJECT('0Project0000000000000a',$,'P',$,$,$,$,$,$);
+#42=IFCWALL('0Wall00000000000000042',$,'Wall A',$,$,$,$,$,$);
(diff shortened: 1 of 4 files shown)


louistrue (commented):
Reviewed the diff and ran it locally (scratch worktree at head, fresh install + build):

- New test file passes 4/4 against the real parser (a real IFCRELAGGREGATES walk, geometry-less assembly, plus the no-assembly control).
- The handleSelectAssembly ordering claim checks out in the slice source: setSelectedEntityIds([]) does null selectedEntityId (selectionSlice.ts:160-163), and the explicit setSelectedEntityId afterwards is exactly the pattern the slice's own comment prescribes for multi-model selection — so not reusing handleSelectRelatedEntity is justified, not just asserted.
- decomposedBy() is the IfcRelAggregates inverse edge (packages/query/src/entity-node.ts:153), gated on parent.type === 'IfcElementAssembly', so plain spatial aggregation parents stay quiet — the control test pins that.
- Hook deps: assemblyParent's useMemo reads only entityNode; handleSelectAssembly's useCallback lists everything its body reads (selectedEntity, models, the three setters, cameraCallbacks). No frozen captures.
- SpatialLocationBadge extraction is a byte-equivalent lift of the old inline JSX (compared side by side), with the null gate moved into the component.
- Cross-PR check with #3709 (both edit PropertiesPanel.tsx): git auto-merges cleanly and both PRs' test files pass together on the merged tree (8/8).

Issue #3672, opened 2026-09-02
check-module-size is TS/TSX-only, so 208 .mjs files (up to 2196 lines) are outside the ~400-line rule it reports OK for


louistrue:
## What

`scripts/check-module-size.mjs` enforces the AGENTS.md "split modules over ~400
non-generated lines" rule for `.ts`/`.tsx` only. Every `.mjs` file in the
repository is invisible to it.

Its own docstring states the gap it was written to close:

> the Rust gate's `collect_rs_files` filters on `extension == "rs"`, and the
> allowlist it reads is Rust-only, so the AGENTS.md "split modules over ~400
> non-generated lines" rule had *no* executable enforcement at all on the
> TypeScript side.

It closed that for TypeScript. `scripts/` is almost entirely `.mjs`, and that
tree is where the CI gates themselves live.

## Extent

208 non-test `.mjs` source files under `scripts/`, `packages/` and `apps/`. The
largest over the guideline:

```
2196  scripts/test-wasm-contract.mjs
1363  scripts/moonshot/ci/check-report-numerals.mjs
1304  scripts/moonshot/b45-m1-midterm/run.mjs
1165  scripts/lib/vitest-timeout-audit.mjs
1145  scripts/check-loader-hook-specifier-match.mjs
1077  scripts/lib/pr-review-signal.mjs
 980  scripts/moonshot/m5-constrained-decoding/tasks-tier2.mjs
 976  scripts/review/validate-findings.mjs
 957  scripts/check-issue-queue.mjs
 951  scripts/check-pr-review-signal.mjs
```

The gate reports `check-module-size: OK (2084 files measured, 305 allowlisted,
0 new over 400)` while none of these are among the 2084.

## Why it matters beyond the count

This is the third instance this week of a gate whose extension filter silently
excludes a whole tree. #3639 was the same shape in
`check-source-text-assertions` (`.mjs` invisible; 70 files, 14 violations, gate
reported 0), fixed in #3662. A gate that reports OK over a population it never
looked at is worse than no gate: the OK is read as evidence.

## Suggested shape

Not a straight extension addition — that would fail immediately on dozens of
pre-existing files. Either:

1. Extend the file walk to `.mjs` and seed the allowlist with the current
   measured budgets in one commit, so the ratchet holds the line going forward
   without demanding a refactor now. This is what the TS half already does (305
   allowlisted). Note #3289/#3291: a single repo-wide digest serialises the PR
[...]

Issue #3658, opened 2026-09-01
validate-findings: a finding's quote and its line are validated independently, so a comment can be posted on the wrong line


louistrue:
`validateFindings` checks two things separately and never that they agree:

- `quoteAppearsIn(file.patch, f.quote, MIN_FINDING_QUOTE_CHARS)` — the quote appears **somewhere** in the file's patch
- `lineIsAdded(f.line, file.addedLineRanges)` — the line is inside an added range

Nothing compares the quote against the text actually at that line. So a finding quoting one added line, anchored at a different added line, passes validation and is posted as an inline comment on the wrong line. `quoteAppearsIn` also trims, so indentation cannot disambiguate.

## Provenance

Raised three times independently today, which is why it is worth filing rather than leaving as a review note:

1. CodeRabbit on #3654 (`Validate findings[].quote against findings[].line`).
2. A `/code-review` pass on the same PR, noting the two checks are independent and that `validate-findings.mjs`'s own header says so outright.
3. The canary fixture in #3635 carried a `quoted_line` paired with the wrong `line` number for exactly this reason — the mismatch survived because "`validate` checks quote membership and line-in-range independently, never that the two agree", quoted from that commit message.

The third one is the useful evidence: the defect has already let a wrong pairing through in this repo's own frozen fixture, where it went unnoticed until the numbers were derived rather than written down.

## Not fixed in #3654

That PR changes `rubric.md` and one error-message string; the matching logic is byte-identical to main. Its rubric text does tell the model that a finding's `quote` must be the line it anchors to, exactly — but telling is not enforcing, which is the gap here.

## Shape of a fix

`newFileLines(patch)` (in `build-review-input.mjs`, landed as #3635) already yields `{line, text, kind}` in new-file numbering, so the check is available without a second walker: require that the finding's quote match the `text` at its `line`, trimmed on both sides the way `quotableLines` trims.

Wants mutation tests in both directions before it lands: a correctly-paired finding must still pass, and a finding whose quote comes from a different added line must fail. Worth checking against recent real reviews first — if the live model routinely emits slightly-off line numbers that currently pass, tightening this turns working reviews red, and that trade should be measured rather than assumed.

PR #3712, opened 2026-09-02
fix(gate): make the source-text allowlist ratchet identity, not count

### LINKED ISSUE #3664: source-text allowlist: the ceiling counts entries, so a swap grandfathers a new violation without raising it
`ALLOWLIST_CEILING` records a **count**, and the gate only compares `allowlist.size` against it. So a change can remove one allowlisted path and add a different one, keep the size identical, and pass — grandfathering a brand-new violation without the deliberate ceiling raise the constant exists to force.

Concretely: file A is converted and its row deleted; file B, a new violator, is added in the same commit. Size stays 22. B is now in the allowlist, so the gate does not report it as a new offender, and no ceiling edit was needed. The "only ratchets down" promise in the allowlist header is not enforced for the set, only for its cardinality.

## Scope

Pre-existing — the ceiling has been count-based since the gate landed. #3662 does not introduce it, but it does widen the swap surface from 8 entries to 22, which is why it is worth fixing now rather than later.

## Shape of a fix

Compare the candidate allowlist against the base revision rather than against a number: `git show origin/main:scripts/source-text-assertion-allowlist.txt`, and fail when the new set contains a path the base set did not, regardless of size. The ceiling can stay as the "must not grow" backstop.

Two things to get right, and both are the reason this is not a one-liner:

- The gate runs locally as well as in CI, where `origin/main` may be absent or stale. It must degrade to the current count-based check rather than erroring or silently skipping — a gate that no-ops when it cannot find a base is the failure mode this repo has hit repeatedly.
- CI checks out with limited depth in some lanes; the base blob needs to be fetchable there or the comparison is vacuous.

Wants a mutation test in both directions before it lands: a swap that keeps the size constant must fail, and a genuine conversion that shrinks the list must still pass.

## Found by

CodeRabbit CLI on #3662. Filed rather than fixed inline because #3662's own diff is a scope widening plus grandfathering, and changing the ratchet's comparison model in the same commit would make its mutation evidence harder to read.

BIMvoice:
## Summary

`ALLOWLIST_CEILING` in `scripts/check-source-text-assertions.mjs` only compares `allowlist.size` against a hardcoded number. A commit that removes one allowlisted file's row and adds a *different* file's row leaves the size unchanged and passes -- the new violation is grandfathered without the deliberate ceiling raise the constant exists to force (the swap CodeRabbit flagged on #3662, filed separately as #3664 to keep #3662's own mutation evidence readable).

**Mechanism read from the current code (post-#3662), not the issue's snapshot:** the gate walks `packages/`, `apps/`, `scripts/` for `.test.(ts|tsx|mts|mjs)` files, flags source-text assertions via `analyze()`, subtracts files the allowlist excuses, and separately requires `allowlist.size === ALLOWLIST_CEILING` (both directions checked). Nothing about that comparison has memory of what the allowlist looked like before this change -- it only sees today's count.

**Fix -- identity, not count.** The gate now also resolves this branch's merge base with `origin/main` (falling back to local `main`, identical derivation to `check-module-size.mjs`'s `changedFiles()`), diffs the current allowlist against the one committed there, and treats any path present now that was absent at the base as a new exemption. Identity key = the file's path in the allowlist -- the same key every other check in this gate already uses to correlate a row with a file. Coarser (just the count) is the bug being fixed; finer (e.g. a hash of the violation site) would false-positive on harmless renames/reformats of an already-excused file, so path is the deliberate middle.

A new exemption is only accepted if `ALLOWLIST_CEILING` rose by at least as many entries as are new, relative to the constant's value at the base commit. This reuses the existing exact-size-match rule rather than adding a second, competing one: that rule already forces `ceiling(current) - ceiling(base)` to equal `new.length - removed.length` whenever both commits passed, so a same-size swap (`new=1, removed=1`, forced ceiling delta `0`) can never satisfy `delta >= 1` -- it is unsatisfiable **by construction**, not by an extra check that could itself be forgotten. Splitting the removal and the addition into separate changes is the fix, not a workaround: each new exemption gets its own reviewable ceiling-raise line instead of hiding behind a coincidental offset.

**Migration: none.** The identity key is derived entirely from git history, not stored in the allowlist file, so no existing row needs a new field or one-time regeneration.

**Degradation.** When no merge base is resolvable (shallow clone, no `origin` remote -- exercised for real in this environment, see below), the gate prints a `WARNING` and falls back to the existing count-only checks rather than erroring or silently skipping.

## Tests (`scripts/check-source-text-assertions-identity.test.mjs`, new)

Built against real synthetic git repos (temp dir, `git init`, same method as `check-module-size.test.mjs`'s `gitTree()`), running the unmodified gate via a new `--root` flag:

- same-size swap (A removed, B added, ceiling untouched) -> **fails**, names B, does not name A
- pure removal (ceiling lowered to match) -> **passes**
- untouched allowlist -> **passes**
- new violation with no allowlist row at all -> **still fails** (pre-existing offender path, unaffected)
- pure addition (ceiling raised to cover it, nothing removed) -> **passes**
- no resolvable merge base -> **degrades with a warning**, does not error, exits 0

All 6 pass. All 75 pre-existing cases in `check-source-text-assertions.test.mjs` still pass, including the e2e "gate passes on the repo" case.

## Verification run

- `node scripts/check-source-text-assertions.mjs` against the current tree: **exits 0** (22 allowlisted, 9 marked, 0 new). In this checkout `origin` is a stale personal fork with neither file at its merge base with `HEAD`, so the identity check itself prints its degrade-warning and the run falls back to count-only -- demonstrating the degrade path live, not just in the synthetic harness. Manually confirmed `upstream/main` (the real base) has both files and a valid merge base, so the identity check runs for real in a normal clone / CI checkout.
- `node scripts/check-test-wiring.mjs` -- OK
- `pnpm run check:vitest-timeout-audit` -- unaffected, OK
- `node scripts/check-ci-path-coverage.mjs` -- initially failed: the new test harness's synthetic-tree setup (`['packages', 'apps', 'scripts']`) reads the literal string `apps`, which the coverage scanner resolves down to the already-known `apps/landing` hole (no TS/tests there). Added one line to `scripts/ci-path-coverage-allowlist.txt`, following the exact existing convention for every other tree-walking gate's harness (`check-module-size.test.mjs apps/landing # same walk, in the harness`, etc.) -- OK after.
- `node --test scripts/check-ci-path-coverage.test.mjs`, `scripts/check-test-glob-coverage.test.mjs` -- OK

## Related but not absorbed

Read the open-PR list for file overlap (`scripts`, `.github`) before starting -- no overlap with this file set. #3672/#3689 (module-size `.mjs` scanning) and #3658/#3710 (finding quote/line coupling) are unrelated files, left untouched.

Closes #3664

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -700,6 +700,13 @@ jobs:
       # it must not.
       - name: Check the source-text detector's own regressions
         run: node --test scripts/check-source-text-assertions.test.mjs
+      # ALLOWLIST_CEILING is a count, so a commit that removes one row and adds
+      # a DIFFERENT one leaves the count untouched and the swap invisible
+      # (#3664). This pins the allowlist IDENTITY check that catches it,
+      # against real synthetic git history (same method as
+      # check-module-size.test.mjs's gitTree()).
+      - name: Check the source-text allowlist identity ratchet
+        run: node --test scripts/check-source-text-assertions-identity.test.mjs
       # That gate's SCAN SCOPE is packages/ and apps/ only -- `grep -c rust` over
       # it returns 0 -- so the same AGENTS.md rule was unenforced across the
       # whole rust/ tree, demonstrated rather than inferred: a genuine
--- a/scripts/check-source-text-assertions-identity.test.mjs
+++ b/scripts/check-source-text-assertions-identity.test.mjs
@@ -1,0 +1,257 @@
+#!/usr/bin/env node
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression harness for the allowlist IDENTITY check in
+ * scripts/check-source-text-assertions.mjs (#3664).
+ *
+ * ALLOWLIST_CEILING is a count, and a count cannot tell "one row removed, a
+ * different row added" apart from "nothing changed" -- both leave
+ * allowlist.size exactly where it was. This check compares the current
+ * allowlist against the one committed at this branch's merge base and fails
+ * when a path is present now that was not present there, regardless of
+ * whether the total size moved.
+ *
+ * Method mirrors scripts/check-module-size.test.mjs's `gitTree()`: each case
+ * builds a REAL git repository in a temp dir outside this checkout (so the
+ * derivation cannot quietly read this repo's own history), commits a "base"
+ * state, checks out a "feature" branch, mutates the allowlist and re-runs the
+ * UNMODIFIED gate via `--root`. Nothing here reads the gate's source.
+ *
+ * Fixture test files use the same read-then-.includes shape the gate's own
+ * detector tests pin, so `analyze()` flags them without needing the full
+ * repo's TEST_FILE_RE tree layout beyond `packages/`, `apps/`, `scripts/`.
+ *
+ * Run: node --test scripts/check-source-text-assertions-identity.test.mjs
+ * (a named step of the CI node-test job, and covered by its glob catch-all).
+ */
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, unlinkSync, rmSync } from 'node:fs';
+import { spawnSync } from 'node:child_process';
+import { tmpdir } from 'node:os';
+import { join, dirname } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+const GATE = join(ROOT, 'scripts', 'check-source-text-assertions.mjs');
+const DETECT = join(ROOT, 'scripts', 'source-text-assertion-detect.mjs');
+
+/** A source-text-assertion test fixture the detector flags, naming `rel`. */
+function violatingTest(rel) {
+  return `import { readFileSync } from 'node:fs';
+const p = '${rel}';
+const source = readFileSync(p, 'utf8');
+assert.ok(source.includes('someCall('));
+`;
+}
+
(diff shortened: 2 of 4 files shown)


louistrue (commented):
Reviewed and verified by running, in a scratch worktree at c78044b:

**What I ran**
- `node --test scripts/check-source-text-assertions-identity.test.mjs`: 6/6 pass, exit 0.
- `node --test scripts/check-source-text-assertions.test.mjs`: 75/75 pass, no regression in the existing detector suite.
- `node scripts/check-source-text-assertions.mjs` unpiped on the real tree: exit 0, and the summary line now reads `identity-checked vs origin/main (158e0cefe)` - the check ran live, not degraded.
- Mutation probe: reverted `scripts/check-source-text-assertions.mjs` to origin/main while keeping the new test file. The swap case then fails (`not ok 1 - a same-size swap (A removed, B added, ceiling untouched) fails`). The test binds to the fix; it cannot pass vacuously.

**Gate-review specifics (what could it fail to catch)**
- The inertness question: the identity check needs a resolvable merge base in CI. Confirmed the job that runs this gate checks out with `fetch-depth: 0` (.github/workflows/test.yml:639), so `merge-base origin/main HEAD` resolves there and the degrade path is reserved for genuinely shallow local clones. Not inert where it matters.
- The swap arithmetic: the exact-size rule already pins ceiling(current)-ceiling(base) to new-removed across two passing commits, so a same-size swap forces delta 0 against a requirement of >=1. Checked the claim against the code, and the synthetic suite exercises both directions plus the degrade path by exit code, not output text.
- PR-scoped, not commit-scoped: a swap split across two commits of one branch still fails at the PR head, since the base is the merge base with main, not the previous commit.
- Fail-closed on an unparseable base ceiling (baseCeiling null gives -Infinity), and the missing-base path warns loudly and falls back to the count-only checks rather than skipping silently.

No defects found. The three required checks are green at this head; the failing "Claude review"/"Review posted" lanes are the zero-review condition this review resolves, not a code signal.

Issue #218, opened 2026-02-17
Incorrect rendering of a window element


semhustej:
I am experiencing incorrect rendering of a window element.

This is how it looks like in IFCLite:

![Image](https://github.com/user-attachments/assets/c8985d1c-a1b4-4fb7-85bf-7fa588c2ceaa)

This is how it should look:

![Image](https://github.com/user-attachments/assets/99ed95a0-af2c-411e-9b2f-465a8305570e)

I am able to reproduce in Edge and Firefox.

This is the sample file:

[IFC-test-lite.zip](https://github.com/user-attachments/files/25370738/IFC-test-lite.zip)

**Version used:**
ifc-lite
Version 1.8.0
Built Feb 17, 2026

PR #3717, opened 2026-09-02
fix(geometry): add a feature-gated topology reject for torn CSG results (#3440 step 2)

BIMvoice:
## Summary

Step 2 of the 3-step order #3440's own text prescribes (diagnostic → feature-gated → hard gate). Step 1, the non-gating diagnostic record (`BoolFailureReason::KernelError` via `record_topology_tear`), is already merged (#3442). This PR is step 2 only: a REJECT signal, gated off by default.

- Every boolean-op accept path (`subtract_mesh`, `subtract_mesh_many`, `union_mesh`/`union_meshes`, `intersection_mesh`) now also calls `topology_gate_reject`, which reuses `directed_closed` / `closed_or_hairline` from `router/voids/prism_cut/closure_checks.rs` — the analytic prism-cut path's own accept/reject predicate — rather than adding a fourth definition of watertightness to the crate, per the issue's own direction.
- Gated behind a **new** `csg_topology_gate` Cargo feature, off by default and enabled by nothing downstream. Deliberately **not** `debug_geometry` / `csg_capture` / `observability`: the native server already enables `observability` in production (`ifc-lite-processing/observability`), so wiring a behaviour change through any of those three would flip real hosts today.
- **Off the default path entirely, not merely a runtime flag checked after the work.** `topology_gate_reject` has two `#[cfg]`-selected bodies: without the feature it's a stub that returns `false` immediately and never touches the mesh; the closure predicate is not compiled into that path at all.
- With `--features csg_topology_gate`, a torn result is rejected exactly the way an existing `KernelOutputInvalid` result already is (falls back to the un-cut host / an empty mesh / a plain merge) and records the new `BoolFailureReason::OpenTopologyRejected` reason through the same `take_csg_failures` / per-host diagnostics channel `#3442` already wired — no second channel.
- `difference_result_looks_degenerate` moved out of `csg/mod.rs` into a new sibling module (`degenerate_check.rs`) to make room for the four gate call sites without raising the module-size ratchet. `csg/mod.rs` is 563 lines against its 611-line allowlisted budget.

## Measured

New test `rust/geometry/tests/issue_3440_topology_gate_census.rs` (compiles to nothing without `csg_topology_gate`) drives the real void-cut path (`GeometryRouter::process_element_with_voids`) over the full local fixture corpus and tallies `OpenTopologyRejected` via `take_csg_failures`:

```
115 models, 2071 void hosts swept
99 hosts REJECTED (335 individual Difference-op rejections; 0 on Union, 0 on Intersection)
```

This is a **different metric** than `triangulation_invariance.rs`'s `edge_stats.open` census (issue #3435 already documents those two watertightness readings in this crate diverging), so it is not expected to match — and it doesn't: the golden corpus's `edge_stats` reading is 165 torn hosts / 17,863 open edges over the same population (unchanged by this PR, verified byte-identical below), and the ISSUE_068 heavy fixture alone is separately documented at 27/29 CSG-caused torn hosts. My 99/2071 counts *hosts whose accept-path result fails this crate's own `directed_closed`/`closed_or_hairline` gate predicate*, which is a different (stricter on some cases, looser on others) test than `edge_stats`. All rejections were on `Difference` — consistent with the issue's framing that this is a void/opening-subtraction defect, and bears on #3353 (not absorbed here — two prior attempts at fixing the kernel itself were withdrawn after regressions).

## Default path unchanged — proof

- `cargo test -p ifc-lite-geometry` (no feature): 0 failures, including three pre-existing step-1 tests (`topology_tear_recorded_by_every_boolean_op_without_gating`, `_once_per_batched_subtract_not_once_per_chunk`, `_once_per_union_meshes_not_once_per_intermediate`) asserting the informational-only, non-gating behaviour byte-for-byte unchanged.
- `cargo test -p ifc-lite-geometry --features triangulation-alt --test triangulation_invariance -- watertightness_census_and_triangulator_invariance`: golden vs run, 0 regressed / 0 reclassified / 0 improved — the `edge_stats` census is untouched.
- `cargo test -p ifc-lite-processing --test module_size_ratchet`: passes, no allowlist row raised.
- `cargo clippy -p ifc-lite-geometry --all-targets` clean, with and without `--features csg_topology_gate`.
- `cargo check --workspace` clean.

## Known, expected divergence when the feature IS on

`cargo test -p ifc-lite-geometry --features csg_topology_gate` has one pre-existing snapshot failure: `geometry_correctness_harness`'s `issue #218 — window rendering` fixture snapshot, because the gate genuinely changes that fixture's output (a torn subtract now falls back, going from 3340→3492 triangles). This is real, correct divergence corroborating the census above, not a regression — I did not touch that snapshot or its harness; it is simply not feature-aware, and this feature is not part of the standard test matrix.

## Scope note on #3353

Not absorbed here per the task's own instruction. The 99-host count above is a lower bound on what #3353 (booleans tearing on overlapping/rotated operands) would need to fix to get this gate to a state where it could be turned on unconditionally.

Part of #3440 -- implements step 2 (the feature-gated check) only. Step 3 (the hard, unconditional gate) is explicitly NOT done here and needs #3353 progress first; not claiming a close.

### DIFF
--- a/.changeset/csg-topology-gate-feature.md
+++ b/.changeset/csg-topology-gate-feature.md
@@ -1,0 +1,11 @@
+---
+'@ifc-lite/wasm': patch
+---
+
+`ClippingProcessor`'s four boolean-op accept paths (`subtract_mesh`, `subtract_mesh_many`, `union_mesh`/`union_meshes`, `intersection_mesh`, in `rust/geometry/src/csg/`) validated a kernel result for finiteness and index bounds only, so a torn (non-manifold / open-edge) result was accepted and shipped silently — issue #3440. A prior change added a non-gating diagnostic record for this (`BoolFailureReason::KernelError`, informational only, still shipped in every build). This adds the second step the issue prescribes: a REJECT signal, but gated behind a new, off-by-default `csg_topology_gate` Cargo feature that no downstream crate enables — not `debug_geometry` / `csg_capture` / `observability`, which the native server already turns on in production, so wiring a behaviour change through any of those would have flipped real hosts today.
+
+Without the feature this is a no-op with no measurable cost: the gate function's default-build body never runs the closure predicate, it only returns `false`. With `--features csg_topology_gate` (a `cargo test`/`cargo build` opt-in for CI or census measurement) a torn result is rejected the same way an existing `KernelOutputInvalid` result already is — falling back to the un-cut host / an empty mesh / a plain merge — and records the new `BoolFailureReason::OpenTopologyRejected` reason through the same `take_csg_failures` / per-host diagnostics channel the informational record already used.
+
+Reuses `directed_closed` / `closed_or_hairline` (`router/voids/prism_cut/closure_checks.rs`) — the analytic prism-cut path's own accept/reject predicate — rather than adding a fourth definition of watertightness to the crate, per the issue's explicit direction.
+
+Not on any default or production build path: `ifc-lite-wasm`'s own feature set does not enable `csg_topology_gate`, so this ships no observable change to `@ifc-lite/wasm` consumers. Flagged as a patch because it is a new, currently-inert opt-in surface on the underlying Rust crate that this package's release process versions together.
--- a/rust/geometry/Cargo.toml
+++ b/rust/geometry/Cargo.toml
@@ -39,6 +39,17 @@ triangulation-alt = ["dep:earcut"]
 # this crate's `cfg(test)` tree. `ifc-lite-processing` enables it as a
 # dev-dependency. Never enabled in a shipping build.
 test-support = []
+# #3440 step 2: turns `csg/mod.rs`'s four boolean-op accept paths' informational
+# open-topology check (step 1, always on) into a REJECT — a torn kernel result
+# falls back exactly like an existing `KernelOutputInvalid`. Off by default and
+# enabled by NOTHING downstream (no other crate in this workspace turns it on,
+# unlike `observability`, which the native server does enable) — deliberately
+# so this stays a `cargo test/build --features csg_topology_gate` opt-in for
+# CI/census measurement, not a change to what any shipping build accepts. The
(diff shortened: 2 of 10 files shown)


louistrue (commented):
Reviewed the diff and ran the one thing CI never compiles — the feature build (scratch worktree at 9d7ffe4d2):

- cargo test -p ifc-lite-geometry --lib: 764 passed, 0 failed (default build).
- cargo test -p ifc-lite-geometry --lib --features csg_topology_gate: 761 passed, 0 failed. The count reconciles exactly: 3 step-1 KernelError tests cfg'd out under the feature, 1 gate-on mirror added. This matters because CI's matrix never builds with the feature, so a compile break or failing gate-on test behind the cfg would be invisible to every green lane on this PR — verified it is not there.
- topology_gate_rejects_every_torn_boolean_result_when_enabled ran and passed in that build: all four accept paths reject a torn result with the documented fallbacks (un-cut host / un-cut host / plain merge / empty) and record OpenTopologyRejected exactly once each.
- Gate wiring read end to end: without the feature the stub returns false before touching the mesh, and topology_gate_is_a_true_noop_without_the_feature (compiled in every build) pins the step-1 behaviour byte-for-byte — so the default path has its own regression guard, not just a claim.
- Feature is defined only in rust/geometry/Cargo.toml and enabled by nothing in the workspace (grepped all Cargo.tomls) — the observability/production-flip concern the PR body raises is genuinely avoided.
- audit_and_gate_union avoids the double-count (gate records, then returns fallback without also calling record_topology_tear); union_meshes' N-way plain-merge fallback is the right generalisation of union_pair's 2-way one.
- degenerate_check.rs is a verbatim move of difference_result_looks_degenerate, no edits.

Did not re-run the full-corpus census (needs the fetched fixture corpus); the 99/2071 number stays the author's measurement, but nothing merged here depends on it — the gate ships dark.

Issue #3714, opened 2026-09-02
pointcloud: whitespace-only E57 fileOffset/recordCount attributes parse as 0


louistrue:
Follow-up surfaced by the pre-push review of PR #3680. That PR makes blank and whitespace-only `minimum`/`maximum`/`scale`/`offset` attributes behave like absent ones, but the `fileOffset`/`recordCount` parsing three lines above (`packages/pointcloud/src/formats/e57-xml.ts` ~84-93) still diverges:

- `fileOffset=""` is falsy, so `!fileOffsetAttr` skips the scan — good.
- `fileOffset=" "` is truthy, and `Number(' ') === 0` passes the `Number.isFinite(x) && x >= 0` guard, so the scan decodes from logical offset 0 (the E57 header) and produces garbage points instead of being skipped.
- `recordCount=" "` likewise yields a silent empty scan while `recordCount=""` skips the scan.

Blank and whitespace-blank should take the same path (skip the scan). Same one-line `trim()` treatment as PR #3680's attribute parsers, plus a regression test pinning that `fileOffset=" "` and `fileOffset=""` behave identically.

Verified against the current branch: `parsePoseElement` is NOT affected — `textChild` (xml-mini.ts:200) trims and returns null for blank text, so the `?? '1'`/`?? '0'` fallbacks fire correctly there.

BIMvoice:
**Verified by execution** of the exact guard in `packages/pointcloud/src/formats/e57-xml.ts` (~lines 82-92, current `upstream/main`):

```
Number(" ")               -> 0
Number.isFinite(0)        -> true
!" "                      -> false   (whitespace is truthy, so the `!fileOffsetAttr` skip does not fire)
```

So `fileOffset=" "` and `recordCount=" "` both pass `if (!Number.isFinite(x) || x < 0) continue;` as `0`, while `fileOffset=""`/`recordCount=""` correctly skip via `!fileOffsetAttr`. Confirmed by reading the surrounding code that there is no `.trim()` anywhere in this file today.

[...]

PR #3725, opened 2026-09-02
fix(mcp): treat a blank/whitespace-only material Name as absent in materialFallbackName

BIMvoice:
## Reviewer summary

- **Without this:** a material with a blank or whitespace-only `Name` (e.g. `IFCMATERIAL('',$,$)`) shows up as a blank entry instead of falling back to `(unnamed)`/`(no material)` — in `materials_list`, the `materials` resource, and the `viewer_get_selection` text summary alike.
- **Evidence:** on the real `materials_list` handler against a fixture with `IFCMATERIALLIST((#80))` where `#80=IFCMATERIAL('',$,$)`, the unfixed chain returned `{"name":"","count":1}` and rendered the text summary as `"  •  — 1"` (a blank bullet line).
- **Risk:** applies to all four call sites already routed through the one shared `materialFallbackName` helper (a prior consolidation, #3515). Control: an empty `IFCMATERIALLIST(())` still correctly produces `(unnamed)`. One related but out-of-scope fallback chain (`material-summary.ts`'s `memberLine`) is flagged, not touched.
- **Sequencing:** none; based on `main`. Regression fix for behavior introduced by merged #3515; carries no queue label per the PR's own "Queue gate" note (no matching `ready` issue found).

---

## Summary

`materialFallbackName` (`packages/mcp/src/material-naming.ts`) chains its candidates with `??`:

```
mat.name ??
mat.materials?.[0]?.name ??
mat.layers?.find((l) => l.materialName)?.materialName ??
...
```

`??` only falls through on `null`/`undefined`. A material whose `Name` is present but blank (`IFCMATERIAL('',$,$)`) short-circuits the chain and is returned verbatim — and the callers' own `?? '(unnamed)'` / `?? '(no material)'` (`packages/mcp/src/tools/query.ts`, `packages/mcp/src/resources/providers.ts`) don't catch it either, since `''` is not nullish.

Reproduced end-to-end via the real `materials_list` handler on a fixture with `IFCMATERIALLIST((#80))` where `#80=IFCMATERIAL('',$,$)`: it returned `{"name":"","count":1}`, and the text summary rendered as `  •  — 1`. An empty `IFCMATERIALLIST(())` correctly produced `(unnamed)` (control, unaffected).

**This is a regression fix for behaviour introduced by merged #3515**, which added `materialFallbackName` as the single consolidated fallback for this MCP package.

## Fix

A bare `??` → `||` swap would fix `''` but discard any other falsy value, and would not catch a whitespace-only name (`'   '`), which is a real shape in this codebase (#3714, whitespace-only E57 attributes). Instead, added a small `isBlank`/`firstNonBlank` helper (undefined/null/whitespace-only all treated as absent) applied to every candidate in the chain, including the `layers`/`profiles`/`constituents` `.find()` predicates (a whitespace-only `materialName` there would previously have matched `.find()`'s truthy check and been picked over a later, real, name).

`materialFallbackName` still returns `undefined` when every candidate is blank — no hardcoded placeholder inside the helper — so each of its callers keeps applying its own wording (`(unnamed)` vs. `(no material)`).

## Sibling sweep

`materialFallbackName` has one definition and four call sites, all already routed through it (that consolidation was #3515's whole point):

- `packages/mcp/src/tools/query.ts` (`query_entities` per-entity summary, `materials_list`)
- `packages/mcp/src/resources/providers.ts` (`materials` resource)
- `packages/mcp/src/tools/material-summary.ts` (`viewer_get_selection`'s text summary, via `formatMaterialsBlock`)

No caller re-implements the `??` chain independently, so the fix inside the shared helper covers all of them.

One related-but-out-of-scope spot: `material-summary.ts`'s `memberLine` helper falls back per-member with `m.materialName ?? m.name ?? '?'`, a separate fallback chain that does not call `materialFallbackName`. It has the same blank-string exposure in principle, but is a distinct code path this PR does not touch — flagging it rather than silently leaving it unmentioned.

## Tests

- `packages/mcp/src/material-naming.test.ts` (new): direct unit coverage — blank/whitespace top-level name falls through, blank/whitespace layer `materialName` falls through to a later layer, a genuine name is returned unchanged, all-blank returns `undefined` (not `''`), null/undefined input control, empty-list control.
- `packages/mcp/src/tools/query.test.ts`: new fixture (`MATERIAL_BLANK_MODEL`) and `materials_list` integration tests reproducing the exact reported shape (blank `Name`, whitespace-only `Name`, genuine name, and the empty-`IfcMaterialList` control), plus a check that the text summary never renders a blank bullet line.

RED confirmed both ways: the integration tests fail against `IFCMATERIALLIST((#80))`/`IFCMATERIAL('',$,$)` with the pre-fix chain (`{"name":"","count":1}`, `"  •  — 1"` in the text summary), and reverting `material-naming.ts` alone (`git stash` on just that file) turns the new unit tests red too. Both restored to green after the fix.

## Gates run

- `pnpm exec vitest run` in `packages/mcp`: 371 passed (36 files)
- `pnpm exec tsc --noEmit` in `packages/mcp`: clean
- `node scripts/check-module-size.mjs`: pre-existing exit 1 confirmed unrelated (identical on `upstream/main` before this change, in unrelated `scripts/review/*.mjs` files); `material-naming.ts` (60 lines) and `query.test.ts` well under budget
- `node scripts/check-unused-locals.mjs`: exit 0 (mcp package not among the flagged non-compiling packages)
- `node scripts/check-test-wiring.mjs`: OK
- `check:vitest-timeout-audit`: exit 0
- `check-api-surface.mjs`: not run — `materialFallbackName` is internal, not exported from `packages/mcp/src/index.ts`, no public surface change
- Changeset added for `@ifc-lite/mcp` (patch)

## Queue gate

No open issue matches this (searched `materialFallbackName`, "material blank", "materials_list unnamed", "IFCMATERIAL blank", and the open-issue material list). Opening with no label and no `Closes` line, per the queue-gate policy for a regression in already-merged code — please grant a waiver or file a `ready` issue as you see fit.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Material listings now handle blank or whitespace-only material names correctly by falling back to meaningful names or the `(unnamed)` label.
  * Prevented blank entries from appearing in material summaries.
  * Preserved correct handling for genuinely named materials and empty material lists.
  * Restored regression coverage for bulk entity queries that include attributes.

### DIFF
--- a/.changeset/material-fallback-blank-name.md
+++ b/.changeset/material-fallback-blank-name.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/mcp': patch
+---
+
+`materialFallbackName` (shared by `materials_list`, `count_entities` group_by material, `query_entities`'s per-entity material label, `viewer_get_selection`'s text summary, and the `materials` resource) chained its candidates with `??`, which only falls through on `null`/`undefined`. A material whose `Name` is present but blank (`IFCMATERIAL('',$,$)`), or whitespace-only (a real shape — see #3714), short-circuited the chain and was returned verbatim instead of falling through to the next candidate or the caller's own `(unnamed)`/`(no material)` placeholder — a blank row in a material schedule instead of a labelled unknown. Every candidate in the chain, including the layer/profile/constituent `.find()` predicates, is now checked against a blank/whitespace-only test; the function still returns `undefined` (not a hardcoded placeholder) when every candidate is absent, so each caller's own wording still applies.
--- a/.changeset/mcp-quantity-diff-blank-storey.md
+++ b/.changeset/mcp-quantity-diff-blank-storey.md
@@ -1,0 +1,7 @@
+---
+'@ifc-lite/mcp': minor
+---
+
+`quantity_diff`'s group-by-storey grouping chained `model.bim.storey(e.ref)?.name ?? '(none)'`, which only falls through on null/undefined. A storey whose `Name` is present but blank (`IFCBUILDINGSTOREY('...','',...)`) or whitespace-only short-circuited the chain and was used verbatim as the group key instead of falling through to `(none)`. Same defect family as #3515 / `materialFallbackName` (`material-naming.ts`).
+
+Also exports `isBlank`/`firstNonBlank` from `@ifc-lite/mcp/browser` (the browser-safe entrypoint already re-exports the node-free kernel other in-browser MCP consumers need) so the web playground's dispatcher can reuse the same blank/whitespace-name handling instead of duplicating it.
--- a/apps/viewer/src/components/mcp/playground-dispatcher-blank-names.test.ts
+++ b/apps/viewer/src/components/mcp/playground-dispatcher-blank-names.test.ts
@@ -1,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Regression: `count_entities` (group_by storey) and `get_entity` chained
+ * `storey?.name ?? '(no storey)'` / `data.name ?? '(unnamed)'`, which only
+ * fall through on null/undefined. A present-but-blank/whitespace-only
+ * `IFCBUILDINGSTOREY`/`IfcWall` Name short-circuited the chain and was
+ * emitted verbatim — a blank-labelled group in `count_entities`, and a
+ * blank quoted name in `get_entity`'s text summary — instead of falling
+ * through to the placeholder. Same family as #3515 (MCP material-name
+ * fallback), reusing `firstNonBlank` from `packages/mcp/src/material-naming.ts`
+ * (re-exported via `@ifc-lite/mcp/browser`).
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { dispatch, parsePlaygroundModel } from './playground-dispatcher.js';
+
+function ifc4(body: string): string {
+  return [
+    'ISO-10303-21;', 'HEADER;', "FILE_DESCRIPTION((''),'2;1');",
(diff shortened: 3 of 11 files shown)


louistrue (commented):
Reviewed and verified by running, in a scratch worktree at d2c7ad2:

**What I ran**
- Full `packages/mcp` suite: 372 passed (36 files), including the new `material-naming.test.ts` unit coverage and the `materials_list` integration fixtures.
- Mutation probe: weakened `isBlank` back to nullish-only (dropped the `.trim() === ''` clause) and re-ran the three touched test files: 10 tests fail across all three (unit, query integration, diff integration). Restored, all green again. The tests bind to the whitespace handling, not just to null/undefined.
- `node scripts/check-module-size.mjs` unpiped on a local merge with current origin/main: exit 0.
- `node scripts/check-test-wiring.mjs`: OK; the new viewer test file follows the same node:test pattern as its `playground-dispatcher-*.test.ts` siblings.

**Boundary checks I looked for specifically**
- Valid-but-falsy: a material or storey genuinely named `'0'` survives (`isBlank('0')` is false) - the fix does not over-trim real names.
- The fixture puts pressure on both directions: blank, whitespace-only, genuine name, and the empty-`IfcMaterialList` control, and the `unnamedCount === 3` assertion pins the bucket count so the not-toContain('') assertions cannot pass vacuously.
- `materialFallbackName` still returns undefined rather than a hardcoded placeholder, so each caller's own wording is preserved - checked the four call sites all route through the helper.
- The `.find()` predicates now skip whitespace-only layer/profile/constituent names instead of matching them as truthy; covered by the layers test.

Changeset shape is right: patch for the behaviour fix, minor for the new `isBlank`/`firstNonBlank` exports on the browser entrypoint, and api-surface.json carries both. Agree with flagging `memberLine`'s separate chain rather than widening this diff.

No defects found. The three required checks are green at this head. The Issue queue lane is red only for the queue-gate waiver this PR's body already requests.

PR #3728, opened 2026-09-02
fix(cli): fall a blank/whitespace storey or material name through to the placeholder, not an empty key

BIMvoice:
## Reviewer summary

Without this fix, a blank `IFCBUILDINGSTOREY`/`IFCMATERIAL` `Name` corrupts machine-readable CLI output: `query --group-by storey --json` emits `{"": {"count": 1}}` — an empty-string JSON key — instead of a labelled `"(no storey)"` group, with the same shape for `--group-by material`, `--unique storey|material`, and blank labels in `ask`'s answer strings and `stats`' building/storey/material summaries. Any script or agent parsing the `--json` output loses the group entirely or keys off an empty string. Root cause: `storey?.name ?? '(no storey)'` and similar chains only fall through on `null`/`undefined`, not on a present-but-blank or whitespace-only string. Fix is additive and narrow — a shared `isBlank`/`firstNonBlank` helper (mirroring the already-reviewed `packages/mcp/src/material-naming.ts` from #3725) applied only at the identified fallback sites; a genuine name is verified unchanged via control tests, bounding blast radius to CLI output formatting. No sequencing dependency on other open PRs.

## What this fixes

`ifc-lite query --type IfcWall --group-by storey --json` on a fixture with a blank storey `Name` (`IFCBUILDINGSTOREY('...','',...)`) produces `{"": {"count": 1}}` — an empty-string JSON key instead of `"(no storey)"`. Text mode prints a blank label. Same shape for `--group-by material` with `IFCMATERIAL('',$,$)`, and for `--unique storey`/`--unique material`.

Root cause: `storey?.name ?? '(no storey)'` and `mat?.materials?.[0] ?? mat?.name ?? '(no material)'` use `??`, which only falls through on `null`/`undefined`. A present-but-blank (or whitespace-only) `Name` short-circuits the chain and is emitted verbatim where a placeholder belongs. This corrupts scriptable `--json` output — an agent or tool parsing the result gets an empty key instead of a labelled group.

Same family, reproduced and fixed alongside:
- `ask`'s `building-name`, `tallest-storey`, `largest-element`, `smallest-element` recipes rendered a blank instead of `(unnamed)` in the `answer:` string.
- `stats`' building name, storey list (`.filter(Boolean)` misses whitespace-only), and material summary (`!matName` misses whitespace-only, so a whitespace string became a Map key) had the same gap.

## Fix

Added `isBlank`/`firstNonBlank` to `packages/cli/src/output.ts` — same shape as `packages/mcp/src/material-naming.ts` (PR #3725, same defect family: a blank `IFCMATERIAL` Name). Applied at every site above. `--unique`'s inline logic and `stats.ts`'s inline building-name/storey-name logic were pulled into testable pure functions (`computeUniqueValues` in `query-output.ts`, `computeBuildingName`/`computeStoreyNames` in `stats-aggregation.ts`) alongside the fix, since neither had a unit-testable seam before.

Both directions verified: a blank/whitespace value now yields the placeholder, and a genuine name is still returned unchanged; the existing "(no storey)"/"(no material)"/"(unnamed)" behaviour for a truly absent value is unchanged (control tests included).

## Test plan
- [x] `packages/cli`: `pnpm test` — 610 passed, 8 skipped (pre-existing skips unrelated to this change)
- [x] `pnpm tsc --noEmit` in `packages/cli` — clean
- [x] New regression tests in `query-output.test.ts`, `ask.test.ts`, `stats-aggregation.test.ts`; each RED-confirmed against the pre-fix code by temporarily reverting the source change, then restored to GREEN
- [x] `check-module-size.mjs`, `check-unused-locals.mjs`, `check-test-wiring.mjs` — clean on touched files
- [x] Changeset added (`@ifc-lite/cli` patch)

## Gate

I could not find an open issue covering this defect (searched `gh issue list`/`gh search issues` for "blank storey", "no storey", empty-key, `--group-by`, `--unique material`). Not applying a label myself. Flagging for a waiver or a `ready` issue, same as #3725 (the sibling PR for the MCP material-name instance of this same defect family).

### DIFF
--- a/.changeset/cli-blank-storey-material-key.md
+++ b/.changeset/cli-blank-storey-material-key.md
@@ -1,0 +1,5 @@
+---
+'@ifc-lite/cli': patch
+---
+
+`query --group-by storey|material`, `query --unique storey|material`, `ask` (building name, largest storey, largest/smallest element), and `stats` (building name, storey list, material summary) chained their name candidates with `??`/`.filter(Boolean)`/`!name`, which only falls through on null/undefined (or, for `!name`/`Boolean`, only catches a plain empty string). A storey or material whose `Name` is present but blank (`IFCBUILDINGSTOREY('...','',...)`, `IFCMATERIAL('',$,$)`) or whitespace-only short-circuited the chain and was emitted verbatim — an empty-string JSON key from `--group-by`/`--unique` instead of `"(no storey)"`/`"(no material)"`, a blank bullet in `ask`'s `answer:` string, a whitespace-string row in `stats`' material summary. Every site now falls through blank/whitespace candidates to the next one, same shape as `firstNonBlank`/`isBlank` in `packages/mcp/src/material-naming.ts`; a genuine name is still returned unchanged.
--- a/packages/cli/src/commands/ask.test.ts
+++ b/packages/cli/src/commands/ask.test.ts
@@ -19,7 +19,67 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 const createHeadlessContext = vi.hoisted(() => vi.fn());
 vi.mock('../loader.js', () => ({ createHeadlessContext }));
 
-const { askCommand } = await import('./ask.js');
+const { askCommand, RECIPES } = await import('./ask.js');
+
+function recipe(name: string) {
+  const r = RECIPES.find((r) => r.name === name);
+  if (!r) throw new Error(`recipe not found: ${name}`);
+  return r;
+}
+
+describe('ask recipes — blank/whitespace name falls through to "(unnamed)"', () => {
+  /**
+   * Regression: a present-but-blank/whitespace `IfcBuilding.Name` was
+   * chained with `buildings[0]?.name ?? '(unnamed)'`, which only falls
+   * through on null/undefined, so the `answer:` string rendered
+   * `Building: ` (or `Building:    `) instead of `Building: (unnamed)`.
+   */
+  it('building-name: blank Name falls through to "(unnamed)"', () => {
+    const bim = { query: () => ({ byType: () => ({ toArray: () => [{ name: '' }] }) }) };
+    const result = recipe('building-name').execute(bim, {});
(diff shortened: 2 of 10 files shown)


louistrue (commented):
Reviewed and verified by running, in a scratch worktree at 41fc342:

**What I ran**
- Full `packages/cli` suite: 610 passed, 8 pre-existing skips (59 files).
- Mutation probe: weakened `isBlank` in `packages/cli/src/output.ts` back to nullish-only and re-ran the three new test files: 14 tests fail across query-output, ask, and stats-aggregation. Restored, all green. Every claimed fix site is covered by a test that actually goes red without it.
- `node scripts/check-module-size.mjs` unpiped on a local merge with current origin/main: exit 0 (query.ts shrinks under its budget after the extraction; no allowlist rows touched).

**Boundary checks I looked for specifically**
- Valid-but-falsy: a storey named `'0'` is kept by `computeStoreyNames` (`isBlank('0')` false), same as the old `.filter(Boolean)` kept the truthy string `'0'` - no behaviour change for real names.
- `computeMaterialSummary`: the whitespace-only case that slipped past `!matName` (truthy `'   '` as a Map key) is now skipped, and the genuine-name control pins that real materials still count.
- `outputGroupBy` material chain: the old code could put a non-string (`mat.materials[0]` as an object) straight into the group key; the new `typeof first === 'string' ? first : first?.name` handles both shapes. Worth knowing this hunk fixes slightly more than blank names.
- `computeUniqueValues` extraction is a faithful move of the `--unique` logic out of `queryCommand` plus the fallthrough fix; the `(no value)` pset path and the invalid-path `fatal` are unchanged.
- `ask` recipes keep the raw `name` in structured output while the `answer:` string gets the placeholder - reasonable split, JSON consumers still see the source value there while the group keys in `--json` get the labelled placeholder.

No defects found. The three required checks are green at this head. The Issue queue lane is red only for the queue-gate waiver the body requests, same as #3725.

PR #3740, opened 2026-09-03
fix(rust): migrate content_hash/step_text/prepass REFERENCE readers off wrapping express-id parsing

### LINKED ISSUE #3421: Rust express-id REFERENCE readers still wrap, so a ref above u32 binds to a real low-numbered entity
Split out of #3395, which fixed the **definition** side and deliberately left this.

`packages/parser` and `rust/core`'s scanner now refuse an express id above `u32::MAX` at
the parse boundary rather than truncating it into a real entity's key. The **reference**
readers on the Rust side were not changed and still accumulate with
`wrapping_mul` / `wrapping_add`:

- `rust/core/src/fast_parse.rs:298` (`extract_first_entity_ref`)
- `rust/core/src/fast_parse.rs:372` (`extract_entity_refs_from_list`)
- `rust/core/src/decoder.rs:498, 564, 635, 758, 846, 937`

So a *reference* such as `#4294967297` still wraps to `#1` and binds to a real
low-numbered entity, which is the same defect #3395 fixed one hop earlier. An entity now
refused at definition time can still be pointed at, and the pointer resolves to the wrong
thing rather than to nothing.

## Why it was not folded into #3395

#3395's scope was the by-id index and the parse boundary. The reference readers are a
distinct code path with a different failure mode — #3395 produces a *missing* entity with
a counted, reported refusal; this produces a *wrong* entity, silently. Fixing it wants its
own change and its own tests, and widening #3395 to cover it would have mixed two
different guarantees in one diff.

Confirmed by a `/code-review` pass on #3395's branch, cited to the file:line above. Traced
by reading, not reproduced against a real file — an id above 2^32 is far outside anything
a real exporter emits, so the practical severity is low and the reason to close it is that
the guard's whole purpose is to make collisions impossible rather than merely unlikely.

## What a fix needs

- Bound the reference readers at the same value, from the same home, so the definition and
  reference sides cannot disagree. `MAX_EXPRESS_ID` is single-homed per language today
  (`packages/parser/src/express-id.ts`, `u32::MAX` on the Rust side); a third copy here
  would be the drift #3395 was careful to avoid.
- Decide what a refused reference DOES. Dropping it silently is the same class of defect
  in a new place — #3395's rule is that a guard which acts must also report, and the two
  halves fail differently. A dangling reference is probably a load-time diagnostic rather
  than a hard failure, but that is the decision the fix has to make explicitly.
- A test with a reference above `u32::MAX` that goes red without the bound, and one at
  exactly `u32::MAX` proving the inclusive boundary still resolves.

BIMvoice:
## Summary

louistrue's comment on #3421 named three REFERENCE readers outside `rust/core` that #3432 never migrated, and stayed wrapping:

> **Still wrapping, which is why this stays open.** Three REFERENCE readers of STEP text outside `rust/core` were never migrated:
>
> - `rust/geometry/src/router/content_hash.rs:133` (`parse_first_ref`) and `:352` (`sig_entity` child refs) — this feeds the geometry dedup content hash, so a wrapped ref hashes the wrong subtree.
> - `rust/export/src/step_text.rs:100` (`refs_in_line`, reached from `step.rs:201` and `merged/plan.rs:108` for reachability) — and `merged/plan.rs:145` in the same crate already saturates, so `rust/export` currently holds one saturating and one wrapping copy of the same idea.
> - `rust/processing/src/prepass.rs:390` also wraps, though it is definition-side (`find_ifcproject_id`).

A full-repo census (`git grep -n 'wrapping_' -- rust/`, filtered to the digit-accumulation shape `wrapping_mul(10)...wrapping_add`) confirms exactly these three sites and no others.

## Fix

Made `ifc_lite_core::express_id::parse_express_id` `pub` (it was `pub(crate)`) and migrated all three sites to call it instead of hand-rolling a fourth accumulation loop, per the module's own doc comment ("a new caller must reuse this function rather than writing its own loop"). Overflow policy matches what `rust/core` already established: refuse (`None`), inclusive at `u32::MAX`, never saturate (a saturated id could itself collide with a real entity holding `u32::MAX`).

- `content_hash.rs`: `parse_first_ref` now returns `None` on overflow instead of `Some(wrapped)`. The `sig_entity` child-ref loop treats an oversized ref as unresolvable — the same fixed sentinel already used when `get_raw_bytes` misses — instead of recursing into whatever entity the wrap landed on.
- `step_text.rs`: `refs_in_line` drops an oversized ref from the collected list instead of pushing the wrapped value.
- `prepass.rs`: `find_ifcproject_id` skips an oversized `#id=IFCPROJECT(...)` match and keeps searching, instead of returning the wrapped id (this function already treats "not found" as `None`, so refusal reuses that path).

## Reproduced (RED, verified by execution)

Before the fix, `find_ifcproject_id` on `#1=IFCWALL(...)#4294967297=IFCPROJECT(...)` returned `Some(1)` — the oversized IFCPROJECT id wrapped onto the real IFCWALL's id. Same class of misbinding demonstrated for `parse_first_ref` and `item_signature`/`sig_entity` (an oversized child ref hashed identically to a real, present entity `#1` before the fix; after the fix it hashes identically to a genuinely-missing ref instead). Verified RED by reverting the production changes (tests kept) and watching all four new/updated test cases fail; restored and confirmed `git diff --stat` is empty against the committed fix.

## Tests

Control (ordinary id unaffected), boundary (`u32::MAX` accepted), and RED→GREEN (id above `u32::MAX` refused, never aliased) for all three sites — `content_hash_tests.rs` (new sibling file), `step_text_tests.rs` (existing sibling file), `prepass_tests.rs` (new sibling file, extracted from `prepass.rs`'s inline `mod tests` to stay under the module-size ratchet after the production fix's net line growth).

## Test plan
- [x] `cargo test -p ifc-lite-core --lib` — 223 passed
- [x] `cargo test -p ifc-lite-geometry --lib` — 768 passed
- [x] `cargo test -p ifc-lite-export --lib` — 387 passed
- [x] `cargo test -p ifc-lite-processing` — all suites passed, including `module_size_ratchet`'s `no_module_grows_past_its_ratchet_budget`
- [x] `cargo clippy -p ifc-lite-core -p ifc-lite-geometry -p ifc-lite-export -p ifc-lite-processing --lib --tests -- -D warnings` — clean

Closes #3421

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of oversized STEP entity references. Values exceeding the supported range are now rejected instead of wrapping to unrelated entities.
  - Updated reference processing, geometry hashing, and project detection to consistently handle invalid or unresolved entity IDs.
  - Prevented oversized references from being mistaken for valid entities during geometry comparison and processing.

- **API**
  - Exposed the Express ID parser for use by external integrations and extensions.


louistrue (commented):
Reviewed at 1f1f9e9 in a scratch worktree, verified by execution.

**Overwrite analysis** (what each behavioural change displaces downstream, not just whether the call is legitimate):

- `refs_in_line` now drops an oversized ref instead of pushing the wrapped id. Callers: `step.rs:201` and `merged/plan.rs:108` (reachability), `merged/line_edit.rs:83/:207` (line decisions). Before, a wrapped ref could mark an unrelated low-numbered entity reachable and keep it in an export; now it contributes nothing. The rewrite side at `merged/line_edit.rs:39` already saturates (CR #2952) and is untouched, so an oversized ref stays dangling, never mis-pointed. The pre-existing wrap-vs-saturate split inside rust/export shrinks; the saturating copy at line_edit.rs:39 remains, disclosed and out of #3421 scope.
- `sig_walk_bytes`: the oversized-child arm folds `fold(0, 0x00BA_D0BA_D0BA_D000)`, which I confirmed is byte-identical to the existing unresolvable-reference sentinel in `sig_entity` (content_hash.rs:293). Consumers of `item_signature` (router/mod.rs:558, processing.rs:773) are in-memory dedup memos; nothing persisted is keyed on these hashes, so changed values for malformed files clobber no stored state.
- `find_ifcproject_id` keeps searching on refusal via the existing `from = kw + 1` advance; no infinite loop, and `None` reuses the established not-found path at prepass.rs:321.
- `pub(crate) mod express_id` -> `pub`: the module holds exactly one item (`parse_express_id`), nothing else leaks.

**Census, re-run independently** (python sweep over rust/, not grep): the only remaining production digit-accumulation loops are the saturating `merged/line_edit.rs:39` and the example `export/examples/merge_ifc.rs:193` (unchecked `* 10 +`, debug-panics on a 10-digit ref; non-shipped, follow-up material at most). Every wrapping REFERENCE reader is gone, matching the PR's census.

**Mutation check** (ran it rather than trusting the RED claim): reverted all three production hunks to wrapping accumulators, tests kept. All four new refusal tests failed (parse_first_ref, item_signature, refs_in_line, find_ifcproject_id); control and boundary tests stayed green. Restored, diff empty. The tests cannot pass with the defect present.

**Runs**: `cargo test` for core/export/processing `--no-fail-fast` exit 0; geometry `--lib` 768 passed; `cargo clippy --all-targets -- -D warnings` clean on all four crates. New tests are self-contained byte fixtures, no fixture_or_skip involvement. prepass.rs test extraction is lossless: 8 moved + 3 new = 11.

**Semver/changeset**: the added pub export is additive and fits the pending unreleased 8.1.1 major (the gate compared 7 of 7 crates against crates.io 7.1.1). No npm surface change, so no changeset is due.

No defects found.

PR #3934, opened 2026-09-05
feat(cli): run a repeatable IFC delivery check from a saved recipe

BIMvoice:
## Summary

Closes #3931.

- Add `ifc-lite delivery <recipe.json> [--json] [--out F] [--html F]`: a repeatable, reviewable model-delivery check that packages the existing structural validation (`ifc-lite validate`) and IDS validation (`ifc-lite ids`) into one saved, versioned recipe.
- Recipe shape: `{"models": ["model.ifc"], "structural": true, "ids": ["rules.ids"]}`. Every `models`/`ids` path resolves relative to the recipe file's own directory. A recipe declaring zero applicable checks is a fatal error — it can never produce a meaningful delivery verdict.
- Every check reports `pass` / `fail` / `error` — never folded together. `error` covers an unreadable/empty/corrupt model, an unreadable or unparsable IDS file, or an IDS document declaring zero specifications; none of those count as a pass.
- The overall verdict is `pass` only when every declared check on every declared model passed, so an unreadable model or an empty ruleset can never read as a successful delivery (the issue's explicit boundary).
- The consolidated report is a flat array of checks (never keyed by model/type/source), so two checks can never collide/overwrite each other in the output — the defect class behind two shipped bugs in `ifc-lite schedule`. Running the same recipe twice against unchanged files produces byte-identical `--json` output.
- `--html <file>` renders an additional standalone HTML report.
- A committed worked example lives at `packages/cli/examples/delivery/` (a passing model, a structurally-broken model, an unreadable model, and an IDS specification that fails) and is exercised by `delivery-example.test.ts`.

### Assumptions (stated per the issue's ambiguity guidance)

- Modeled the command on `ifc-lite schedule`'s `--spec`/`--save` saved-definition pattern (path-resolution-relative-to-file, `fatal(...)` on any malformed field) and on `ifc-lite gym`'s reuse of `computeValidationIssues`/`IDSNamespace` for running the underlying checks without a second implementation.
- A recipe applies the same `structural`/`ids` selection to every listed model (no per-model check overrides) — the simplest reading consistent with "one documented CLI invocation."
- An IDS document whose specifications are ALL "not applicable" (matched no entities) counts as `error`, not `pass` — nothing was actually evaluated, so a green check would rest on zero evidence.
- Duplicate `models`/`ids` entries in a recipe are a fatal error rather than silently deduplicated.

## Test plan

[...]

PR #3945, opened 2026-09-05
feat(viewer): save and reopen a portable federation setup

### LINKED ISSUE #3930: feat(viewer): save and reopen a portable federation setup
## Outcome

Let users resume an architecture, structure, and MEP federation without recreating model names, alignment, visibility, and order.

## Scope

Add explicit export/import of a versioned federation setup containing model references/fingerprints and existing per-model setup. On import, let the user match local files to saved slots and review missing or mismatched sources before applying. All files must load through `useIfcLoader.loadFile`.

## Boundaries

The setup describes source files and existing federation transforms. It is not a merged IFC, model archive, collaboration room, or flavor. Do not embed bytes, credentials, signed URLs, mutation history, or every viewer preference. Keep v1 to model membership and alignment.

## Acceptance evidence

- Export a three-model federation and reopen it in a fresh session with files selected in a different order.
- Prove equivalent alignment and visibility.
- Show missing-file and same-name/different-content handling.
- Confirm selections resolve to the correct source models and that one-model setups work.

BIMvoice:
## Summary

Closes #3930.

Lets a user save which models make up a federation — load order, visibility, and the alignment anchor — to a small portable JSON file, and reopen it later (possibly on a different machine), matching saved slots back to local files.

### What the file references vs. embeds
Each slot stores the model's original filename, byte size, and a content fingerprint (`computeSourceFingerprint`, the same sampled 64-bit hash the geometry cache already uses). It never embeds file bytes, an absolute path, a `FileSystemFileHandle`, or credentials — reopening always requires re-selecting the actual files. This keeps the format both small and genuinely portable.

### Alignment is replayed, not baked in
Rather than serialize per-model transform matrices, the file records only which slot was the federation anchor. On reopen, every matched model is re-added through the canonical `addModel` load path and `realignFederation()` is re-run against the restored anchor — the same alignment logic a fresh multi-file load performs. This avoids a second alignment pipeline and keeps the file simple.

### Missing/mismatched models are always visible
Matching is one-to-one and content-first: content fingerprint, then name+size, then name-only (flagged as a possible collision), then missing. `applyFederationSetup` always reports `restored / partial / failed` plus explicit missing and mismatched slot names — never a bare "success" when some models weren't found. A review dialog shows every slot's status before anything is applied.

### Determinism
Slots are built from `Array.from(models.values())` (a Map's insertion order = actual load order, never re-sorted) and serialized with an explicit, fixed key order — no hash-map iteration ordering and no embedded timestamp, so saving the same federation state twice produces byte-identical JSON (tested).

### Where it's reachable
Command palette: "Save Federation Setup" / "Open Federation Setup" (dispatches `window` events handled by `FederationSetupControls`, mounted once from `useFileCommands` — the same pattern `ShareDialog` uses — so it works regardless of which toolbar style is active).

## Analogue matched
The viewer's own `handleRefresh` (`toolbar/useFileCommands.tsx`) — "rebuild the federation from fresh bytes, preserving id + order + state" via sequential `addModel(...)` calls — is the closest existing pattern, since the CLI's `delivery-recipe.ts` saved-definition file referenced in earlier guidance does not exist on `upstream/main` yet (still open as #3934). Content fingerprinting reuses the existing `sourceFingerprint.ts` / cache-key pattern rather than inventing a new hash.

## Test plan
- [x] `apps/viewer/src/lib/federation/federationSetupFile.test.ts` — 16 tests: round-trip, determinism (two saves of the same state are byte-identical), strict/loud validation of malformed files (bad version, zero slots, two anchors, missing field), and the keyed-write trap (two federated models sharing a filename resolve correctly by content, one-to-one, regardless of file pick order).
- [x] Mutation testing: removed the anchor-uniqueness check, the one-to-one match guarantee, the format-version check, and the "resolved" classification — each caught by a test.
- [x] `pnpm typecheck` (viewer + all deps) — clean.
- [x] `node scripts/check-module-size.mjs`, `check-test-wiring.mjs`, `check-source-text-assertions.mjs` — all pass (CommandPalette.tsx addition trimmed to stay at its existing 849-line budget).
- Not unit-tested (UI wiring, thin glue over already-tested primitives): `useFederationSetup.ts`'s calls into `addModel`/`realignFederation`, and `FederationSetupControls.tsx`'s dialog/file-input wiring. Left to CI/manual QA per the viewer's runner constraints.
- Full 666-file viewer suite left to CI (not run locally per repo guidance); ran the targeted 102 tests covering touched files, toolbar/export parity, command palette, and federation load-gate/model-slice — all green.

## Summary by CodeRabbit

* **New Features**
  * Save the current federation setup—including model order, visibility, collapsed state, and alignment anchor—for later reuse.
  * Reopen saved setups by selecting the setup file and matching local model files.
  * Review matched, mismatched, and missing models before restoring.
  * Receive clear results for complete, partial, or failed restorations, including alignment-anchor status.

### DIFF
--- a/.changeset/federation-setup-portable-save-reopen.md
+++ b/.changeset/federation-setup-portable-save-reopen.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/viewer": minor
+---
+
+Add a portable federation setup file: save which models make up a federation (load order, visibility, and the alignment anchor) and reopen it later by matching saved slots back to local files by content fingerprint. The file references source files by name, size, and a content fingerprint — it never embeds file bytes, paths, or handles. Reopening replays the existing alignment pipeline against the restored anchor rather than storing baked transforms, and always reports how many models were restored versus missing or mismatched instead of silently accepting a partial restore. Reachable via the command palette ("Save Federation Setup" / "Open Federation Setup").
--- a/apps/viewer/src/components/viewer/CommandPalette.tsx
+++ b/apps/viewer/src/components/viewer/CommandPalette.tsx
@@ -287,6 +287,9 @@ export function CommandPalette({ open, onOpenChange }: CommandPaletteProps) {
         action: () => {
           window.dispatchEvent(new CustomEvent('ifc-lite:open-files'));
         } },
+      // #3930 portable federation setup — handlers live in `FederationSetupControls` (mounted from `useFileCommands`, ShareDialog's pattern).
+      { id: 'file:save-federation-setup', label: 'Save Federation Setup', keywords: 'federation setup save export portable models order alignment anchor', category: 'File', icon: Save, action: () => { window.dispatchEvent(new CustomEvent('ifc-lite:save-federation-setup')); } },
+      { id: 'file:open-federation-setup', label: 'Open Federation Setup', keywords: 'federation setup restore reopen import portable models order alignment anchor', category: 'File', icon: FolderOpen, immediate: true, action: () => { window.dispatchEvent(new CustomEvent('ifc-lite:open-federation-setup')); } },
     );
     for (const rf of recentFiles) {
       const fileName = rf.name;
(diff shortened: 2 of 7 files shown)


louistrue (approved):
Required checks are green; scope matches ready issue #3930 and the PR carries an effective closing reference.

Issue #3948, opened 2026-09-05
classification-resolver.ts returns [] for a genuinely classified entity/model on a server-parsed (source-empty) store


BIMvoice:
`extractClassificationsOnDemand`/`extractClassificationSystemsOnDemand` (`packages/parser/src/classification-resolver.ts:62` and `:126`) discard classification data the store already resolved, returning `[]` — indistinguishable from "this entity/model has no classification" — whenever `store.source` is empty. That is the normal state for a **server-parsed** store (`apps/viewer/src/utils/serverDataModel.ts` sets `source: EMPTY_SOURCE_BYTES` deliberately; `packages/parser/src/columnar-parser.ts:69` documents `byteLength === 0` as a supported state for "server-parsed, synthetic, GLB, point-cloud").

This is the same shape as the closed #1787 ("IDS on server-parsed models can't see type-inherited property sets"), for classifications instead of type-inherited property sets, and it is not fixed by that issue's patch.

## The two sites

**`extractClassificationsOnDemand` (line 30-107):** when `store.onDemandClassificationMap` is absent (server-parsed stores never build one — grep confirms no call site sets it outside `columnar-parser.ts`'s own full parse), the function falls back to the relationship graph (`store.relationships.getRelated(entityId, AssociatesClassification, 'inverse')`, line 40) — comment: `// Fallback: use relationship graph (server-loaded models)`. This correctly resolves `classRefIds` for a genuinely classified entity. But two lines later:

```ts
if (!classRefIds || classRefIds.length === 0) return [];
if (!store.source?.length) return [];   // <-- discards the classRefIds just resolved
```

the second guard throws away the very ids the relationship-graph fallback exists to recover, because turning an id into a `ClassificationInfo` (system name, code, path) requires reading raw STEP bytes via `EntityExtractor`. The function still returns `[]`.

**`extractClassificationSystemsOnDemand` (line 124-144):** same shape, no relationship-graph fallback attempted at all — it reads `store.entityIndex.byType.get('IFCCLASSIFICATION')` directly (populated on server-parsed stores too), then hits the identical `!store.source?.length` early return.

## Reproduction (run)

[...]

PR #3996, opened 2026-09-06
fix(ids): report CLASSIFICATION_UNRESOLVED, not MISSING, for external-reference classifications on a server-parsed store

### LINKED ISSUE #3954: IfcExternalReferenceRelationship classifications (IfcMaterial/IfcProfileDef) silently report CLASSIFICATION_MISSING on a server-parsed store
## Summary

`appendExternalReferenceClassifications` (`packages/ids/src/bridge/classifications.ts:68-75`) is the only path that resolves classifications attached via `IfcExternalReferenceRelationship` — the mechanism `IfcMaterial` and `IfcProfileDef` (non-rooted resources) use instead of `IfcRelAssociatesClassification`. It bails immediately when the store has no raw STEP bytes:

```ts
const erRefs =
    store.entityIndex?.byType?.get?.('IFCEXTERNALREFERENCERELATIONSHIP') || [];
if (erRefs.length === 0 || !store.source?.length) return;
```

Unlike the `IfcRelAssociatesClassification` path (fixed in #3948/#3951), there is **no relationship-graph fallback at all** for this pathway. The comment right above the function says why:

> Non-rooted resources (IfcMaterial, IfcProfileDef, …) carry classifications via `IfcExternalReferenceRelationship` rather than `IfcRelAssociatesClassification`. The parser doesn't categorize external-ref edges into the relationship graph today, so we scan the type table directly.

So on a server-parsed (source-empty) store, a classification attached to an `IfcMaterial`/`IfcProfileDef` via `IfcExternalReferenceRelationship` is invisible to the IDS bridge, not merely unresolved.

## Impact

For a *presence-only* classification facet ("does this entity have any classification?") on such an entity, `resolveClassifications` returns an empty list, and the IDS classification facet checker reports `CLASSIFICATION_MISSING` — the same code path used for a genuinely unclassified entity. This is a **silent false negative**: the classification exists, the code just can't see it on this data source, but the failure looks identical to "not classified" rather than #3951's `CLASSIFICATION_UNRESOLVED` ("classified, but unreadable here").

This is the same shape of bug #3948/#3951 just fixed for the `IfcRelAssociatesClassification` path, left open for the external-reference path because the relationship graph doesn't carry these edges at all today (parser-level gap, not just a bridge-level gate).

## Suggested direction

Either:
1. Have the relationship-graph builder categorize `IfcExternalReferenceRelationship` edges (mirroring `IfcRelAssociatesClassification`) so a server-parsed store can at least prove presence without needing `source` bytes, reporting `CLASSIFICATION_UNRESOLVED` (not `CLASSIFICATION_MISSING`) for the attribute-dependent parts; or
2. At minimum, have `appendExternalReferenceClassifications` distinguish "no such entities" from "can't read them because `source` is empty" so the IDS bridge can report `CLASSIFICATION_UNRESOLVED` instead of silently treating it as `CLASSIFICATION_MISSING`.

## Evidence

- `packages/ids/src/bridge/classifications.ts:65-75` — the guard and the comment explaining the parser-level gap.
- `packages/ids/src/facets/classification-facet.ts` — the facet checker that reports `CLASSIFICATION_MISSING` when `resolveClassifications` returns `[]`, with no way to tell "genuinely absent" from "external-reference path, source-empty store."
- #3948 / #3951 fixed the equivalent `IfcRelAssociatesClassification` gap; this issue tracks the mirror-image gap for `IfcExternalReferenceRelationship`-based classifications, found during review of #3951.

BIMvoice:
Closes #3954

## Summary

`appendExternalReferenceClassifications` (`packages/ids/src/bridge/classifications.ts`) is the only path resolving classifications attached via `IfcExternalReferenceRelationship` — the mechanism non-rooted resources (`IfcMaterial`, `IfcProfileDef`) use instead of `IfcRelAssociatesClassification`. It bailed unconditionally whenever `store.source` was empty, so a genuinely classified material and an unclassified one were byte-identical to the IDS classification facet — both `CLASSIFICATION_MISSING`.

## Does the relationship graph carry these edges? No.

Unlike the sibling `IfcRelAssociatesClassification` pathway (#3948/#3951), there is no relationship-graph fallback for this one: `RelationshipType` (`packages/data/src/types.ts`) has no `AssociatesExternalReference`-shaped member, `columnar-parser.ts`'s edge builder never calls `addEdge` for `IfcExternalReferenceRelationship`, and `IfcTypeEnum` has no slot at all for `IfcExternalReferenceRelationship`, `IfcMaterial` or `IfcProfileDef` — the server pipeline literally cannot represent these types. PR #3959 (in review) confirms this from the server side too: it explicitly leaves "classifications reached only via `IfcExternalReferenceRelationship`... unchanged — out of scope", because the server resolves classifications only via `IfcRelAssociatesClassification` (`apps/server/src/services/data_model/classifications.rs`).

So on a server-parsed store, presence via this pathway can be neither proven nor disproven — the honest answer is `CLASSIFICATION_UNRESOLVED`, not a fabricated pass and not a fabricated `CLASSIFICATION_MISSING`.

## Design

`ClassificationInfo` gains a `presenceUnknown` flag (always paired with `unresolved: true`). It's only pushed when:
1. the sibling `IfcRelAssociatesClassification` path found nothing for this entity — an already-resolved or already-confirmed-unresolved entity is left untouched, so a real match/mismatch is never diluted into "unresolved" (no regression to #3951's contract); and
2. the entity's own type — read from `EntityRef.type` via `entityIndex.byId`, available without `source` bytes — is one the IFC schema actually allows to be classified this way (an `IfcMaterial`-family or `IfcProfileDef`-family type). An `IfcRoot` subtype like `IfcWall` can never be a `RelatedResourceObjects` target of `IfcExternalReferenceRelationship`, so a genuinely unclassified wall/door/etc. still reports `CLASSIFICATION_MISSING` — a blanket fallback with no type scoping would have regressed the overwhelming common case on every server-parsed model.

`checkClassificationFacet` treats a `presenceUnknown` entry as NOT proving presence: a presence-only facet reports `CLASSIFICATION_UNRESOLVED` instead of fabricating `passed: true`, and a system/value-constrained facet reports it before falling through to the "confirmed classified" wording that would otherwise overclaim presence. The `CLASSIFICATION_UNRESOLVED` message formatters (`translation/service.ts`, `validation/validator.ts`, `en`/`de`/`fr` locales) gain a distinct "presence cannot be determined" wording for this case, leaving #3951's own "classified, but unreadable" wording untouched for its (proven-present) case.

## Verification

**RED**: a store built from a real `IfcParser().parseColumnar()` result (ground truth: a material genuinely classified via `IfcExternalReferenceRelationship`), with `source` wiped to `EMPTY_SOURCE_BYTES` matching `serverDataModel.ts`'s documented shape, reported `CLASSIFICATION_MISSING` for a presence-only facet before this fix:
```
AssertionError: expected 'CLASSIFICATION_MISSING' to be 'CLASSIFICATION_UNRESOLVED'
```

**Mutations**: dropping the type-scoping gate (`isNonRootedClassifiableResource`) reverts to a blanket fallback and is caught by the schema-excluded rooted-entity control (an `IfcWall` reports `UNRESOLVED` instead of `MISSING`). Reverting the facet's presence-only branch to the old `hasAnyClassifications` check fabricates a PASS and is caught by three assertions.

**Controls**:
- A genuinely unclassified `IfcMaterial` (no external-ref relationship anywhere in the model) also reports `UNRESOLVED`, not a fabricated `MISSING` — a server-parsed store cannot prove absence any more than presence for this pathway.
- A rooted `IfcWall` with zero classifications still reports `MISSING` (schema rules out the pathway for it).
- The source-bearing (WASM) path is untouched — unchanged code below the new early-return.
- #3951's own presence-only PASS and `CLASSIFICATION_UNRESOLVED`-for-value behavior on its sibling pathway are unaffected.

## Test plan

- [x] `pnpm --filter @ifc-lite/ids test` — 977 passed (970 before this PR's 7 new tests; no lost tests)
- [x] `pnpm --filter @ifc-lite/parser test` — 1179 passed / 2 skipped, unchanged (no parser files touched)
- [x] `pnpm typecheck` (turbo, `@ifc-lite/ids` + `@ifc-lite/parser`) — clean
- [x] `node scripts/check-module-size.mjs` — clean (trimmed 3 touched files back under their allowlisted budgets rather than raising them)
- [x] `node scripts/check-test-wiring.mjs` — clean
- [x] `node scripts/check-source-text-assertions.mjs` — clean
- [x] `node scripts/check-api-surface.mjs` (after `pnpm build`) — clean
- [x] Changeset added (`@ifc-lite/ids` patch)
- No `.rs` touched — Rust module-size ratchet not applicable

## Shared files

None of the files #3959 touches (`packages/parser/src/classification-resolver.ts`, `apps/viewer/src/utils/serverDataModel.ts`) are touched by this PR — this fix is entirely confined to `packages/ids`.

Claude-Session: https://claude.ai/code/session_01QPHChk3Ve9N519A4kY7436

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved classification validation when source data cannot confirm whether an entity is classified.
  - Prevented unknown classification presence from being reported as missing, incorrectly passing, or falsely confirming classification.
  - Unresolved classification checks now fail consistently across requirement optionality, including prohibited requirements.
  - Corrected classification handling for materials and profile definitions while excluding unrelated relationship entities.
  - Expanded recognition for supported profile definition types.

- **Localization**
  - Added translated messages explaining when classification presence cannot be determined.


louistrue (dismissed):
The new unresolved-presence result still becomes a false PASS for a prohibited requirement in the full validator.

Reproduced at head `86d9e78d4c390215164f3f4eced1327d85661939`, after root `pnpm test --filter=@ifc-lite/ids` passed all 983 tests. Parsed the committed buildingSMART fixture `packages/ids/src/__corpus__/buildingsmart-ids/classification/pass-non_rooted_resources_that_have_external_classification_references_should_also_pass.ifc` with the real `IfcParser`, selected `IFCMATERIAL`, and ran `validateIDS` with one presence-only classification requirement with `optionality: 'prohibited'`. Repeated with the same source-empty conversion as this PR's regression tests.

Observed:
- Source-bearing: 1 applicable material (#16), 0 passed, 1 failed; `Prohibited: found Foobar:1`.
- Source-empty: 1 applicable material (#16), 1 passed, 0 failed; the PASS itself carries `failure.type: CLASSIFICATION_UNRESOLVED`, `field: presence`, and `actualValue: (unresolved)`.

`packages/ids/src/validation/validator.ts`'s `case 'prohibited'` unconditionally inverts `facetResult.passed`, so the new honest uncertainty is converted back into a confident compliance claim. Please preserve unresolved outcomes through cardinality handling and add an end-to-end `validateIDS` regression covering required, optional, and prohibited requirements. The direct facet tests cannot observe this inversion. I have not modified this branch or started a competing implementation.

Issue #4244, opened 2026-09-09
Clash detection: a non-finite tolerance reports a false clean while examining zero geometry


BIMvoice:
## Summary

A non-finite `tolerance` makes the clash detector report a clean, fully-covered model while it examined zero geometry. This is a live false-clean in a user-facing safety result: a caller can be told "no clashes" when the detector never compared a single triangle pair.

## Reproduction (on `92e59033`, `packages/clash`)

A small box wholly inside a large box — a deep, unmissable hard clash — run through `createClashEngine().run(...)` with a NaN tolerance (e.g. from `Number('abc')` on a cleared UI input):

```
CLASHES: 0
ruleCoverage: [{"rule":"r1","matchedA":1,"matchedB":1,"matchedKeysA":["small"],"matchedKeysB":["big"]}]
classify: clean
```

The same elements and rule, run with the default tolerance instead:

```
DEFAULT tolerance clashes: 1
```

A per-rule `rule.tolerance: NaN` produces the same silent-clean result.

## Traced cause

1. `packages/clash/src/engine-ts/orchestrator.ts:34` — `const tolerance = settings.tolerance ?? DEFAULT_CLASH_SETTINGS.tolerance;`. `??` substitutes only on `null`/`undefined`, **not `NaN`** — a `NaN` tolerance sails through unchanged.
2. The per-rule margin computation (`Math.max(tolerance, rule.clearance ?? 0)` in both `ts-kernel.ts` and `narrow.ts`) becomes `NaN` once `tolerance` is `NaN` (`Math.max` with a `NaN` operand is always `NaN`).
3. That `NaN` margin inflates the broad-phase AABB (`inflate(bounds, NaN)`, `src/math/generated/plato.g.ts:223-224`) into a NaN-poisoned box.
4. Every `<=`/`>=` comparison against `NaN` is false, so the BVH broad phase yields **zero candidate pairs** — silently, with no error and no flag.
5. The kernel does compute this: `RuleDetection.candidatesProcessed` is `0` (`ts-kernel.ts:139`). But `runClash` (`orchestrator.ts:112`) folds `candidatesProcessed` only into the run-global `remaining` budget — it is never attached to the per-rule `ClashRuleCoverage` that gets returned to the caller.
6. `ClashRuleCoverage.matchedA`/`matchedB` are selector-match counts computed *before* any geometry runs, so they still read as full coverage.
[...]

PR #4278, opened 2026-09-09
fix(clash): fromPositions throws on an all-non-finite axis instead of returning an inverted AABB

### LINKED ISSUE #4254: clash's fromPositions can hand packages/spatial's BVH an inverted AABB (all-non-finite axis), making the element invisible to spatial queries
## Summary

`packages/clash`'s `fromPositions` (`packages/clash/src/math/aabb.ts`) can return a genuinely inverted `AABB` (`min > max` on one axis), and that box is handed directly to `@ifc-lite/spatial`'s `BVH.build` — which performs no `min <= max` validation — via the broad-phase candidate search. The result: an element with this bounds is silently invisible to any spatial query against that BVH.

## Root cause

`fromPositions` folds only *finite* coordinates into `min`/`max` per axis (deliberately, to survive a partly-poisoned vertex):

```ts
// packages/clash/src/math/aabb.ts
export function fromPositions(positions: Float32Array, transform?: Mat4): AABB {
  ...
  let minX = Infinity, maxX = -Infinity; // ...same for Y, Z
  for (...) {
    if (Number.isFinite(x)) {
      if (x < minX) minX = x;
      if (x > maxX) maxX = x;
    }
    ...
  }
  return { min: [minX, minY, minZ], max: [maxX, maxY, maxZ] };
}
```

If **every** vertex is non-finite (`NaN`/`±Infinity`) on one axis, that axis's `min` stays `Infinity` and `max` stays `-Infinity` — i.e. `min > max`, an inverted box. The function's own docstring acknowledges this:

> If NO vertex contributes a finite coordinate on an axis, the box is returned inverted (`min > max`) on that axis, which is rejected by `boxesTouch` rather than silently paired.

That guarantee only holds for callers that go through `boxesTouch`. It does **not** hold for `packages/spatial`'s `BVH`, which this box also reaches.

## The flow into `packages/spatial`

Both clash adapters build `ClashElement.bounds` with `fromPositions` and put it straight into a `ClashElement`:

- `packages/clash/src/adapters/step.ts:308` — `bounds: fromPositions(merged.positions, worldTransform)`
- `packages/clash/src/adapters/ifcx.ts:141` — `bounds: fromPositions(merged.positions)`

`packages/clash/src/engine-ts/broad.ts` then builds a `@ifc-lite/spatial` `BVH` directly from those bounds, with no re-validation:

```ts
// packages/clash/src/engine-ts/broad.ts
import { BVH, type MeshWithBounds } from '@ifc-lite/spatial';
...
const items: MeshWithBounds[] = groupA.map((e, i) => ({ bounds: e.bounds, expressId: i }));
const bvh = BVH.build(items);
...
const hits = bvh.queryAABB(inflate(b.bounds, margin));
```

`packages/spatial`'s own `BVH.build`/`queryAABB` have no `min <= max` check anywhere, so an inverted box silently degrades the tree instead of erroring.

## Proof (executed)

Ran against a plain `upstream/main` checkout (`39ec95eb73e5aca47877b744efcf6113f71aa18d`), symlinking a prebuilt `node_modules`:

```ts
import { fromPositions } from './math/aabb.js';
import { BVH, type MeshWithBounds } from '@ifc-lite/spatial';

// All-NaN X, finite Y/Z — simulates one axis of a merged mesh being
// entirely non-finite (e.g. a degenerate/corrupted occurrence).
const positions = new Float32Array([NaN, 1, 1, NaN, 2, 2]);
const b = fromPositions(positions);
// => { min: [Infinity, 1, 1], max: [-Infinity, 2, 2] }
// b.min[0] (Infinity) > b.max[0] (-Infinity): inverted on X.

const items: MeshWithBounds[] = [{ expressId: 42, bounds: b }];
const bvh = BVH.build(items);
const hits = bvh.queryAABB({ min: [-1e9, -1e9, -1e9], max: [1e9, 1e9, 1e9] });
// => hits: []
```

Output:

```
bounds: {"min":[null,1,1],"max":[null,2,2]}   // JSON renders ±Infinity as null
hits on huge query: []
```

A query box that trivially covers the element on every finite axis, and covers the ill-defined X range too, returns **zero hits**. The element is invisible to `BVH.queryAABB` while it still exists in `items` — the same failure mode as the general "invisible to spatially-targeted query" scenario for `packages/spatial`.

## Scope note

I checked the shapes most likely to invert a box via a *transform* (negative scale/mirror, 180° rotation, axis permutation applied to an existing `min`/`max` pair): `fromPositions` recomputes `min`/`max` from transformed *vertices* every time (never negates an existing `min`/`max` pair), so it is structurally immune to that class of bug regardless of `transform`'s content. The producer here is specifically the "no vertex contributes a finite value on this axis" edge case, which is real but requires fairly degenerate input (an entire merged occurrence with e.g. NaN X on every vertex — plausible from a corrupted/degenerate B-rep or a failed tessellation, not from ordinary geometry).

I also checked the two other in-repo `@ifc-lite/spatial` `BVH.build` callers:
- `packages/spatial/src/spatial-index-builder.ts` (`buildSpatialIndex`/`buildSpatialIndexAsync`, used by the viewer for picking) — always computes bounds via its own `computeMeshBounds`, which seeds `min`/`max` at `±Infinity` and folds with plain `Math.min`/`Math.max` over *all* coordinates including non-finite ones (`Math.min(NaN, x) = NaN`, never a swap) — cannot invert.
- `packages/clash/src/engine-ts/tri-mesh.ts` (`TriMesh`, narrow-phase per-triangle BVH) — its `triBounds` uses the same unconditional `Math.min`/`Math.max` over exactly 3 vertices — cannot invert for the same reason.

So the gap is real specifically for the finite-only fold in `packages/clash/src/math/aabb.ts`'s `fromPositions`, feeding `packages/clash`'s broad-phase `BVH.build`.

## Suggested fix (not attempted here)

Either:
- have `fromPositions` re-normalize (or fall back to a degenerate-but-valid box, e.g. clamp to a single point) when an axis ends up with no finite contribution, instead of leaving `min > max`, or
- add the `min <= max` defensive check to `packages/spatial`'s `BVH.build`/`queryAABB` itself, so any future producer (this one included) can't silently degrade query results.

## Environment

- Repro run against `upstream/main` @ `39ec95eb73e5aca47877b744efcf6113f71aa18d`.
- Not audited: any Rust-side producer of `AABB`/bounds data (wasm geometry pipeline), since `cargo` isn't runnable in this environment. If the wasm mesher can emit an occurrence whose positions are entirely non-finite on one axis, this is directly reachable from real (corrupted) IFC input; I could not verify that from the Rust side.

BIMvoice:
`fromPositions` folds only finite coordinates, so an axis whose every vertex is non-finite keeps its initialisers and the function returns `min = +Infinity, max = -Infinity` for that axis — an inverted box, handed straight to `@ifc-lite/spatial`'s BVH by both clash adapters.

Reproduced before changing anything:

```
fromPositions(new Float32Array([NaN,1,1, NaN,2,2]))
  -> { min: [Infinity, 1, 1], max: [-Infinity, 2, 2] }
```

## The false docstring — in three places, not one

The docstring claimed such a box is "rejected by `boxesTouch`". `boxesTouch` (`duplicate-metric.ts`) is called only from `duplicates.ts`. It is never on the broad-phase path: `engine-ts/broad.ts` builds the BVH directly from `ClashElement.bounds`. So on the path the issue is about, nothing rejected it.

The same false claim appeared in three places — the exported docstring, an inline loop comment in `aabb.ts`, and a comment in `duplicates.test.ts`. All three corrected.

## Behaviour: skip-and-report, built on a throw

`fromPositions` now throws `NonFiniteAxisError`, naming the offending axis or axes and the vertex count — the same shape as #4244's `requireFinite`.

A throw alone would be wrong: one corrupt element would abort a whole clash run. But `fromPositions` has no element identity, so it cannot skip on its own. The two production callers (`adapters/step.ts`, `adapters/ifcx.ts`) catch per occurrence/entity, drop only that element, and `console.warn` once with a count — reusing `step.ts`'s existing `missingGlobalIds` warning shape rather than inventing a new channel. Returning a sentinel was rejected: that is the bug.

`fromPositions` is package-internal — `packages/clash` exposes no `./math` subpath and does not re-export it from the index — so the new throw is not a public API change.

## Ripple caught during verification

`duplicates.ts`'s sweep already relied on tolerating a non-finite minimum, and a test asserted on `fromPositions(allNaN)`'s old return value. Since `fromPositions` can no longer produce that box, the test now hand-builds the inverted bound, still exercising the sweep's defence against one arriving from a direct SDK caller. That defence matters: a comparator answering `NaN` breaks `Array.prototype.sort`'s contract, and V8's TimSort then emits an arbitrary permutation of the whole array, so real duplicates elsewhere in the model silently vanish. One unjudgeable element must cost only itself.

## Not done, deliberately

No guard added in `BVH.build`. After this fix the adapters cannot hand it an inverted box, but a direct SDK caller hand-building `ClashElement.bounds` still can. That is a real remaining gap with a wider blast radius than the adapter choke point this issue named — flagged here for a separate issue rather than widened into this PR.

## Verification

Mutation-tested, each applied then restored with a verified-clean `diff`:

| mutation | result |
|---|---|
| remove the throw block | both all-non-finite tests RED (`expected function to throw an error, but it didn't`) |
| disable the y-axis finite fold (`if (false && Number.isFinite(y))`) | the pre-existing partial-fold test went RED with `NonFiniteAxisError: ... axis y`, naming the mutated axis |

The second mutation matters: it proves the pre-existing partial-fold test (a finite `y` alongside non-finite coords) still discriminates, so the new throw did not swallow the case it was supposed to leave alone.

`packages/clash`: 472 passed / 21 skipped across 33 files, against a 465/21/33 baseline — the 7 new tests. `differential.test.ts` and one `obb.test.ts` case are **blocked**, not passing: `packages/wasm/pkg/ifc-lite_bg.wasm` is absent in this environment, so they never report pass or fail.

`check-module-size.mjs` exits 0 with no budgets touched.

Closes #4254

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrupt geometry with entirely non-finite coordinates on an axis is now detected instead of creating invisible or invalid bounding boxes.
  - STEP and IFCX imports skip only affected elements, allowing valid model elements to remain available.
  - Imports now issue a warning identifying the model and number of skipped elements rather than failing the entire clash run.

- **Documentation**
  - Clarified handling of invalid and manually supplied inverted bounds.

### DIFF
--- a/.changeset/fix-clash-aabb-nan-invisible.md
+++ b/.changeset/fix-clash-aabb-nan-invisible.md
@@ -1,0 +1,7 @@
+---
+"@ifc-lite/clash": patch
+---
+
+`fromPositions` (`packages/clash/src/math/aabb.ts`) — the single choke point both the STEP and IFCX clash adapters funnel through when building `ClashElement.bounds` — now throws `NonFiniteAxisError` when every vertex is non-finite on some axis, instead of returning the box inverted (`min > max`) on that axis. The inverted box looked like a safe sentinel (the function's own doc claimed it was "rejected by `boxesTouch`"), but `boxesTouch` is only reached from the duplicates pass, not from the BVH broad phase (`@ifc-lite/spatial`, via `engine-ts/broad.ts`) that hard/soft rule-based clash detection uses: there, an inverted box fails every `min <= queryMax && max >= queryMin` check, so the element silently dropped out of every spatial query it should have participated in — including the one that would have found a genuine hard clash (#4254).
+
+`elementsFromStep` (`adapters/step.ts`) and `elementsFromIfcx` (`adapters/ifcx.ts`) now catch `NonFiniteAxisError` per occurrence/entity, skip just that one (rather than aborting the whole clash run for one corrupt element among many), and emit a single `console.warn` naming the model and the count — the same shape as their existing `missingGlobalIds` warning — so a dropped element is loud and countable instead of silently invisible. Positions with only *some* non-finite coordinates are unaffected: the existing per-coordinate fold (the finite coordinate of a partly poisoned vertex still counts) is unchanged.
--- a/packages/clash/src/adapters/ifcx.test.ts
+++ b/packages/clash/src/adapters/ifcx.test.ts
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { elementsFromIfcx } from './ifcx.js';
 import { createClashEngine } from '../engine.js';
 import type { ClashRule } from '../types.js';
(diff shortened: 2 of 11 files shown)


louistrue (approved):
Reviewed the full diff. fromPositions now throws NonFiniteAxisError (naming the axes and vertex count) when no vertex is finite on an axis instead of returning an inverted box that the BVH broad phase silently never matched; the STEP and IFCX clash adapters catch it per occurrence/entity, drop only that element and warn once with a count, mirroring their existing missingGlobalIds shape. The three copies of the false 'rejected by boxesTouch' claim are corrected, the duplicates sweep still tolerates a hand-built inverted bound, and buildStepExclusions moves to step-exclusions.ts so step.ts drops off the module-size allowlist rather than raising it. Tests cover the throw, the per-axis naming, partial poisoning surviving, and both adapters' skip-and-warn behaviour. Changeset is a clash patch (fromPositions is package-internal). Closes #4254; every required check is green on the current head.

PR #4282, opened 2026-09-09
fix(cli): query --group-by aggregation cannot distinguish no-data from a real zero

### LINKED ISSUE #4252: cli: query --group-by --avg --json cannot distinguish no-data from a real zero
## `--group-by --avg --json` cannot distinguish "no data" from "a real zero"

Verified on `upstream/main` at `765501bac`.

Running `outputGroupBy` for a group of entities that carry no quantities at all, with `--group-by --json` and an aggregation mode:

```ts
outputGroupBy(
  [{ ref: 1, type: 'IfcDoor' }, { ref: 2, type: 'IfcDoor' }],
  'type', 'NetVolume', bim, /* jsonOutput */ true, undefined, 'avg'
);
```

produces:

```json
{"IfcDoor":{"count":2,"NetVolume":0,"aggregation":"avg"}}
```

This is byte-for-byte identical to the JSON a group would produce if its true average genuinely were `0`. A consumer parsing this output — an agent, a script, a report generator — cannot tell "the file has no `NetVolume` quantity data for these doors" from "the doors really do have zero volume."

### The two output paths already disagree

Text mode already handles this correctly. `outputGroupBy`'s non-JSON branch prints to stderr:

```
Warning: All NetVolume values are 0. The file may not contain quantity data for this property.
```

The JSON branch has no equivalent warning or signal. The root cause is at `packages/cli/src/commands/query-output.ts:239`:

```ts
groupAgg.set(key, aggregateFinite(vals, mode) ?? 0);
```

This overrides `aggregateFinite`'s deliberate `null` return with `?? 0`. `aggregateFinite`'s own doc comment (`packages/cli/src/commands/query-aggregation.ts`, around line 108-112) states:

> Returns `null` when no finite value was seen (an empty list, or one that is all non-finite), which the caller renders as a blank cell / `null` rather than a fabricated `0`.

The `count === 0 → null` guard itself is at `query-aggregation.ts:124`. The JSON path in `query-output.ts` discards that signal before it ever reaches the caller.

### Consumer scenario

An agent or script runs:

```
ifc-lite query model.ifc --group-by type --avg GrossVolume --json
```

sees `"GrossVolume": 0` for every group, and reports that the doors have zero volume — when in fact the model carries no `GrossVolume` quantity at all. This is a silent-wrong-answer failure mode, not a crash, so it is unlikely to be caught by whoever is consuming the JSON.

### Design options (undecided — please pick one)

1. **Emit `null`** for an empty group's aggregated value, matching `aggregateFinite`'s stated intent. This is a breaking change to the JSON shape (the field would become `number | null` instead of always `number`).
2. **Add a `sampledCount` / `hasData` field** alongside the existing numeric field, e.g. `{"count":2,"NetVolume":0,"sampledCount":0,"aggregation":"avg"}`, leaving the existing shape backward compatible.
3. **Carry the text-mode warning into the JSON payload**, e.g. a top-level `warnings` array, so JSON consumers get the same signal the stderr warning already gives text-mode users.

The text path already solved this problem for humans reading the CLI directly; the JSON path — the one machine consumers rely on — still has the ambiguity.

### Not included here

Three related-but-separate findings (an untested CLI flag fallback, an untested aggregation guard, and untested scope re-verification on rebase) are filed as a companion issue to keep this one focused on the live output defect and its design tradeoff.

BIMvoice:
## Summary

`query --group-by --sum/--avg/--min/--max --json` fabricated an aggregated value of `0` for a group that has no data at all for the quantity, byte-for-byte identical to a group whose real aggregate genuinely computes to `0`. A consumer of the JSON (agent, script, report generator) could not tell the two apart. Text mode already had a coarse signal (a warning keyed on the grand total across ALL groups being 0), but it both missed a no-data group mixed in with real-data groups, and false-fired when a real aggregate legitimately cancels to 0 (e.g. summing `-5` and `5`).

Root cause: `packages/cli/src/commands/query-output.ts`'s `outputGroupBy` computed each group's aggregate with `aggregateFinite(vals, mode) ?? 0`, discarding `aggregateFinite`'s deliberate `null` (documented in `query-aggregation.ts` as "no finite value was seen ... rather than a fabricated `0`") before it ever reached the JSON or table output.

## What changed

- `outputGroupBy` now also tracks, per group, `matchedEntities` — the count of entities in that group that actually carried the aggregated quantity (`getQuantityValue` returned non-null).
- **`--json`**: every group entry with a `sumQuantity` now also carries `matchedEntities`. `matchedEntities: 0` means the numeric field next to it is fabricated, not measured. The existing numeric field's type is unchanged (still always a `number`), so this is additive — an existing consumer reading only the numeric field sees no change in shape or values, and gains the new field to opt into the distinction.
- **Text mode**: a no-data group's line is now annotated `(no data)`, and the trailing warning fires only when no group in the result matched anything (previously: when the grand total across all groups was 0).
- Applied uniformly across all four modes (`sum`, `avg`, `min`, `max`) — they share the same "0 matched entities" ambiguity, not just `avg`. `count` (the entity count, not a quantity-derived value) is unambiguous and untouched.

## Siblings checked

- `query --sum` / `query --avg|--min|--max` (flat, ungrouped): already unambiguous — `matched === 0` takes an explicit early-return error branch (`outputSum`) or emits a `matchedEntities`/`error` field (`outputAggregation`, `outputSum`). No change needed.
- `schedule --subtotals`: already propagates `aggregateFinite`'s genuine `null` straight through `computeValues` → `SubtotalValue.value` → CSV (blank cell) and JSON (`null`), with no `?? 0` anywhere in that path. No change needed.
- No MCP tool or other consumer in this repo reads `query --group-by --json`'s per-group aggregate field (grepped `packages/mcp`, `apps/viewer`, and README/docs) — the MCP `count_entities` tool's own `group_by` support is a separate, count-only implementation unaffected by this change.

## Test plan

- [x] `query-output.test.ts`: new discriminating-fixture tests for avg, sum, and min/max — a group with genuinely no quantity data (`matchedEntities: 0`) vs. a group whose real aggregate computes to exactly 0 via `[-5, 5]` or `[0, 0]` (`matchedEntities: 2`), asserting they now diverge on `matchedEntities` despite an identical numeric `0`.
- [x] Text-mode test: only the no-data group's line gets `(no data)`; the real-zero group's line does not.
- [x] RED/GREEN: reverted the fix to the pre-patch `?? 0` behavior, confirmed the 4 new tests fail (naming `matchedEntities` as `undefined` / missing `(no data)`), restored the fix, confirmed all pass.
- [x] `packages/cli` vitest: `query-output.test.ts` + `query-aggregation.test.ts` — 46/46 passing.
- [x] `node scripts/check-module-size.mjs` — OK, 0 new over budget.
- [x] `git diff --stat upstream/main HEAD` lists only the touched files (query-output.ts, query-output.test.ts, changeset).

Closes #4252

## Summary by CodeRabbit

- **Bug Fixes**
  - Grouped query results now distinguish genuine aggregate values of `0` from groups with no matching data.
  - JSON output includes a `matchedEntities` count for each group.
  - Table output labels groups without data as `(no data)`.
  - Warnings now appear only when no groups contain matching data.

### DIFF
--- a/.changeset/cli-group-by-matched-entities.md
+++ b/.changeset/cli-group-by-matched-entities.md
@@ -1,0 +1,9 @@
+---
+"@ifc-lite/cli": patch
+---
+
+Fix `query --group-by --sum/--avg/--min/--max --json` reporting a fabricated `0` for a group that has no data for the aggregated quantity, indistinguishable from a group whose real aggregate genuinely computes to `0`. Each group entry now carries `matchedEntities`, the count of entities in that group that actually had the quantity — `matchedEntities: 0` means the numeric field is fabricated, not measured. The existing numeric field's type is unchanged (still always a number) so this is additive, not breaking.
+
+The non-JSON (table) rendering now annotates a no-data group's line with `(no data)`, and the trailing warning now fires only when no group in the result matched any data, instead of when the grand total across all groups happened to be `0` (which previously both missed a no-data group mixed in with real-data groups, and falsely warned when a real aggregate legitimately cancels to `0`, e.g. summing `-5` and `5`).
+
+`schedule --subtotals` already propagated a genuine `null` for a no-data subtotal end-to-end (CSV and JSON), so it needed no change.
--- a/packages/cli/src/commands/query-output.test.ts
+++ b/packages/cli/src/commands/query-output.test.ts
@@ -370,6 +370,135 @@ describe('outputGroupBy', () => {
     maxJson.spy.mockRestore();
     expect(JSON.parse(maxJson.chunks.join(''))['IfcWall'].NetVolume).toBe(10);
   });
+
+  /**
+   * #4252: a group with NO NetVolume quantity data at all must be
+   * distinguishable in --json from a group whose real aggregate genuinely
+   * computes to 0. Discriminating fixture: `noDataBim` has doors that carry
+   * no NetVolume quantity anywhere (fabricated 0, matchedEntities 0);
+   * `zeroBim` has doors whose NetVolume values are -5 and 5, so the real
+   * average/sum genuinely is 0 (matchedEntities 2, the same numeric 0 but
+   * NOT fabricated). If both rendered identically on matchedEntities, this
+   * test would be vacuous — it asserts they diverge on that field despite
+   * an identical `NetVolume: 0`.
+   *
+   * Kills reverting `entry.matchedEntities = groupMatched.get(key) ?? 0`
+   * (or dropping it): with it gone, both cases collapse back to
+   * `{count, NetVolume: 0}` and cannot be told apart, which is #4252 itself.
+   */
+  it('distinguishes a group with no quantity data from a group whose real value is 0 (avg, json)', () => {
+    const noDataBim = fakeBim({ quantities: {} });
+    const doors = [{ ref: 1, type: 'IfcDoor' }, { ref: 2, type: 'IfcDoor' }] as FakeEntity[];
+
+    const noData = captureStdout();
+    outputGroupBy(doors, 'type', 'NetVolume', noDataBim, true, undefined, 'avg');
+    noData.spy.mockRestore();
+    const noDataEntry = JSON.parse(noData.chunks.join(''))['IfcDoor'];
+    expect(noDataEntry.NetVolume).toBe(0);
+    expect(noDataEntry.matchedEntities).toBe(0);
+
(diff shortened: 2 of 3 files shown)


louistrue (approved):
Reviewed for merge readiness: scope matches the linked ready issue, regression coverage is discriminating, and no unresolved correctness findings remain. Final merge remains contingent on fresh required checks.

Issue #4095, opened 2026-09-07
Server and browser disagree on a cross-linked space's parent storey (TS resolves by traversal order, leaves a phantom child)


BIMvoice:
The Rust server and the TS parser build different spatial hierarchies for the same file when a space is **both** aggregated under one storey and contained under another — the malformed-but-common cross-linked pattern. Verified by running one fixture through both implementations, not by reading them.

This is **pre-existing** in the TS path and not introduced by #3973; I found it while reviewing that PR, whose body explicitly scopes the browser path out.

## Fixture

`#1` IfcProject, `#2` StoreyA, `#3` StoreyB, `#5` IfcSpace, `#6` IfcFurniture (contained in the space, so a real node is distinguishable from an empty stub).

- StoreyA **aggregates** `#5` via `IfcRelAggregates`
- StoreyB merely **contains** `#5` via `IfcRelContainedInSpatialStructure`

Run with Project's related-list in both orders, `[A,B]` and `[B,A]`. The Rust side is the fixture already committed as `CROSS_PARENT_DUAL_LINKED_SPACE_IFC` (`apps/server/src/services/data_model/tests.rs`), plus a temporary sibling-order-reversed variant.

## Rust — order-independent, correct

```
AGENT_PROBE_RUST: space_node_count=1 parent_id=Some(2)
AGENT_PROBE_RUST: storey_a.children_ids=[5] storey_b.children_ids=[]
```

Identical for both orderings. `spatial.rs` builds `canonical_parent` from **all** `IFCRELAGGREGATES` edges in a single pass before considering any containment relationship, so traversal order cannot affect the result. 35/35 server tests pass.

## TS — order-dependent, plus a dangling duplicate

```
[A,B] (StoreyA visited first): storeyAsCopyElements=[6]  storeyBsCopyElements=[]
[B,A] (StoreyB visited first): storeyAsCopyElements=[]   storeyBsCopyElements=[6]
```

Two separate defects:

1. **The parent flips with DFS pre-order.** When the merely-containing storey B is reached first, it receives the *real* node while the actually-aggregating storey A receives an empty stub. The tie resolves to containment, contradicting the aggregation-wins rule the server applies.

[...]

PR #4285, opened 2026-09-09
fix(parser): prefer the IfcProject-anchored parent over an aggregation back-edge cycle (#4246)

### LINKED ISSUE #4246: A spurious mutual-aggregation back-edge can sever a real spatial subtree, order-dependently (plus a misleading dead-code cycle-guard comment)
A spurious `IfcRelAggregates` back-edge — the kind an authoring tool emits when it declares a parent/child aggregation pair in both directions by mistake — can make `SpatialHierarchyBuilder` sever a real, otherwise well-formed subtree, and whether it does depends purely on STEP declaration order. Verified directly against `SpatialHierarchyBuilder.build` in `packages/parser/src/spatial-hierarchy-builder.ts` (+ `spatial-hierarchy-canonical-parent.ts`) on `upstream/main` at `92e59033`, the same code path the browser and CLI both call.

## Fixture

`#1` IfcProject, `#2` IfcBuilding, `#3` IfcBuildingStorey, `#4` IfcWall (contained in the storey). Three `IfcRelAggregates` edges declared in this order:

- `#2 -> #3` (Building aggregates Storey — correct)
- `#3 -> #2` (Storey aggregates Building — spurious, the authoring-tool mistake)
- `#1 -> #2` (Project aggregates Building — correct, but declared last)

## Result: the whole subtree disappears

```
project.children = []
byStorey = []
byBuilding = []
elementToStorey.get(4) = undefined
```

`#4` (the wall) is not truncated or mis-parented — it is entirely unreachable from `project`. The only diagnostics are the generic warnings `[SpatialHierarchyBuilder] No storeys found in spatial hierarchy` and `No buildings found in spatial hierarchy`, which fire on presence/absence only. In a real multi-building file that has other legitimate storeys elsewhere, `byStorey`/`byBuilding` are non-empty and neither warning fires — this subtree's loss is silent.

## Root cause: it's a race between two aggregation edges for the same child

`computeCanonicalParent` (`spatial-hierarchy-canonical-parent.ts`) resolves each spatial-structure entity's parent to whichever `IfcRelAggregates` edge naming it as child was declared **first** in the file, globally, before recursion starts. That's correct when a child has at most one legitimate aggregating parent. But `#2` (Building) is named as *child* by two edges here — the spurious `#3 -> #2` and the correct `#1 -> #2` — and the spurious one was declared first, so it wins: `canonicalParent.get(2) === 3`, not `1`. `buildNode`'s `addSpatialChild` only recurses into a child from its canonical parent, so when `buildNode` walks from Project (`#1`), it refuses to descend into `#2` (its canonical parent is `#3`, not `#1`), and the walk starting from `#2`/`#3` never happens because nothing ever reaches them as a root. The whole subtree — Building, Storey, and the wall inside it — is orphaned.

## Order dependence: this is not always-broken

Declaring the legitimate `#1 -> #2` edge **first**, before the spurious `#3 -> #2`, self-heals completely:

```
project.children = [2]
byStorey = [[3, [4]]]
```

Same three edges, same spurious back-edge — only the declaration order changed. This is pure STEP declaration-order fragility, not a path that's always broken, which is why a fixed test corpus can pass while a real authoring tool that happens to emit the aggregation pair in the "wrong" order silently drops a subtree.

## Not the same as #4095

#4095 (closed) covers a single node with two competing parents (a space aggregated under one storey, merely contained under another) — the loser there gets a phantom empty-stub entry, not full removal, and nothing upstream of the contested node is affected. Here the contested child is itself an internal node (`IfcBuilding`) whose winning "parent" is one of its own descendants (`IfcBuildingStorey`), so losing the tie doesn't produce a stub — it makes the entire correct anchor (`IfcProject -> #2`) unreachable and takes the whole subtree with it, including elements that have no relation at all to the malformed edge.

## Worth noting: no way to see this happened

`SpatialHierarchy` exposes no count of unplaced elements anywhere. An element with no containment, a dangling container id, or — as here — a container that got orphaned by a canonical-parent tie, vanishes with no signal a caller could check for.

## A second, related issue in the same mechanism: a misleading dead-code comment

`buildNode`'s `ctx.visited` cycle guard:

```ts
// Guard against cyclic IfcRelAggregates chains (A aggregates B, B aggregates
// A), which would otherwise recurse unbounded and overflow the stack. A
// revisited node is returned as a leaf so the rest of the hierarchy still
// builds.
if (ctx.visited.has(expressId)) {
  return { expressId, type: typeEnum, name, longName, elevation: undefined, children: [], elements: [] };
}
ctx.visited.add(expressId);
```

is dead code. Removing it entirely and re-running the full `spatial-hierarchy-builder.test.ts` suite — including its existing genuine-cycle fixtures — leaves all 16 tests green (12 original + 4 that reproduce Finding A above). The actual protection against unbounded recursion is the canonical-parent single-parent invariant in `spatial-hierarchy-canonical-parent.ts`: `addSpatialChild` only ever recurses into a child from its one canonical parent, so a cycle can supply at most one recursion edge into any node, which structurally cannot loop — the same mechanism responsible for Finding A above.

By contrast, the `seen` guard in the *separate* aggregate-descendant walk inside the storey-containment block (`elementToStorey`, around the `// seen guards against aggregate cycles` comment) **is** load-bearing: removing it and feeding a product-level cycle (`IfcElementAssembly` aggregates `IfcBeam`, `IfcBeam` aggregates the assembly back) hangs the process — a run against that fixture was killed by a 15s wall-clock timeout with no test result printed, versus under 1 second normally.

I'm not asking to remove `ctx.visited` — the maintainer may prefer to keep it as defence in depth in case the canonical-parent invariant is ever weakened. But the comment claims a protection this guard does not actually provide today, which is worth correcting either way.

## Reproduction (ready to paste into a test file)

This currently **fails** on `main` at `92e59033` — the `toEqual([2])` assertion does not hold, because `project.children` comes back `[]`. I did not push it; it is not safe to entrench the current broken behavior as a passing regression test, and asserting the desired behavior means leaving a known-failing test, which isn't right for a merged suite either. Pasting it here as a ready-to-use repro instead:

```ts
it('a subtree is not severed by a spurious mutual-aggregation back-edge, regardless of edge declaration order', () => {
  const strings = new StringTable();
  const entities = new EntityTableBuilder(4, strings);
  entities.add(1, 'IFCPROJECT', 'p0', 'Project', '', '');
  entities.add(2, 'IFCBUILDING', 'b0', 'Building', '', '');
  entities.add(3, 'IFCBUILDINGSTOREY', 'st0', 'Storey', '', '');
  entities.add(4, 'IFCWALL', 'w0', 'Wall', '', '', true);

  const relationships = new RelationshipGraphBuilder();
  relationships.addEdge(2, 3, RelationshipType.Aggregates, 10); // Building -> Storey
  relationships.addEdge(3, 2, RelationshipType.Aggregates, 11); // Storey -> Building (spurious)
  relationships.addEdge(1, 2, RelationshipType.Aggregates, 12); // Project -> Building (legit, declared LAST)
  relationships.addEdge(3, 4, RelationshipType.ContainsElements, 13); // Wall in Storey

  const hierarchy = new SpatialHierarchyBuilder().build(
    entities.build(),
    relationships.build(),
    strings,
    new Uint8Array(),
    { byId: { get: () => undefined } },
  );

  // Desired: the subtree survives regardless of which order the aggregation
  // edges were declared in. Currently fails: project.children comes back [].
  expect(hierarchy.project.children.map((c) => c.expressId)).toEqual([2]);
  expect(hierarchy.byStorey.get(3)).toEqual([4]);
});
```

## Environment

Verified on `upstream/main` at `92e59033708882e9d40eaad0cddc7aab1468d2b4`, `packages/parser/test/spatial-hierarchy-builder.test.ts` baseline 12/12 passing before and after this investigation.

BIMvoice:
Closes #4246

## What was wrong

`computeCanonicalParent` (`packages/parser/src/spatial-hierarchy-canonical-parent.ts`) resolves each spatial-structure entity's parent to whichever `IfcRelAggregates` edge naming it as child was declared first in the file, globally. That is correct when a child has at most one legitimate aggregating parent, but a spurious mutual-aggregation back-edge (an authoring-tool mistake: a parent/child pair declared in both directions) gives a child two candidate parents, and if the spurious one is declared first it wins the tie. When the contested child is itself an internal spatial node (e.g. `IfcBuilding`), the loser is `IfcProject`'s real anchor edge, so the whole subtree becomes unreachable from `IfcProject` — not truncated, not mis-parented, entirely orphaned. Purely order-dependent: swapping which edge is declared first self-heals the same fixture.

## Fix

Skip any tied candidate parent that would close an aggregation cycle back through the child (walking the raw, unresolved `IfcRelAggregates` graph forward from the child), and take the first remaining candidate by declaration order. A tie that does not close a cycle (the pre-existing multiple-real-parents case, #4095) is unaffected — untouched code path, still first-declared-wins. Falls through to plain first-declared if every candidate closes a cycle, so a child is never left without a parent. Logs a warning each time a back-edge is skipped, naming the child and the disqualified candidate.

## The dead-code comment

The issue also flagged `buildNode`'s `ctx.visited` cycle guard: the comment claimed it was what prevents unbounded recursion on a cyclic `IfcRelAggregates` chain, but removing the guard entirely and running the full suite (including its genuine-cycle fixtures) left everything green — the guard never fires. The actual protection is the canonical-parent single-parent invariant added by this fix: recursion only ever follows a node's one canonical parent, so a cycle can supply at most one recursion edge into any node, which structurally cannot loop. The stranded branch fixed the parser logic but left this comment as-is, so this PR also corrects it to name the real mechanism and keep the guard as defence in depth (in case that invariant is ever weakened), rather than leave a false claim in the code.

## Tests

`packages/parser/test/spatial-hierarchy-builder.test.ts` adds a `#4246` describe block that mutates **declaration order**, not just values (the heart of this bug is order-dependence):
- back-edge declared first vs. the real anchor edge declared first (both must reach the same shape)
- an explicit "identical hierarchy regardless of order" cross-check
- a longer, indirect 3-node cycle (A→B→C→A) to prove the fix walks the raw aggregation graph rather than only checking direct children

Mutation-tested: temporarily short-circuited the new cycle-skip branch (`if (false && relType === ... )`) and re-ran — 3 of the 4 new tests went RED with the exact pre-fix symptom (`project.children` empty, `elementToStorey` undefined), confirming the mutated line is what the tests actually exercise. Reverted before pushing.

Full `packages/parser` suite: 111 files / 1216 tests passed, 3 skipped (pre-existing, unrelated).

## Note on the diff history

This branch was originally written against an older `upstream/main` and never opened as a PR. Before pushing, I merged current `upstream/main` into it. The pre-merge diff against `upstream/main` showed changes to `.github/workflows/test.yml`, `Cargo.lock`, `Cargo.toml`, `rust-major-offset.json`, `rust/ffi/Cargo.toml`, `rust/processing/Cargo.toml`, `rust/wasm-bindings/Cargo.toml`, `rust/geometry/src/csg_capture.rs`, and a deletion in `rust/geometry/src/csg_capture_tests.rs` — none of that was real content from this branch; it was stale-base drift (this branch predates unrelated Rust/CI work that has since merged to `main`). After merging `upstream/main` in, the diff against `upstream/main` is exactly the two `packages/parser` files plus the changeset, confirmed with `git diff --stat upstream/main HEAD`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spatial hierarchy extraction when aggregation relationships contain direct or indirect cycles.
  * Canonical parent selection now skips cyclic back-edges, preserving access to the correct hierarchy and contained elements.
  * Added consistent handling across TypeScript and Rust implementations, including fallback behavior for fully cyclic relationships.
* **Tests**
  * Added coverage for direct, indirect, and declaration-order variations of cyclic aggregation relationships.


louistrue (approved):
Reviewed the full diff. computeCanonicalParent (TS) and the new spatial_canonical_parent.rs (apps/server) now skip a tied IfcRelAggregates candidate that would close a cycle back through the child before falling back to first-declared, so a spurious mutual or longer back-edge declared ahead of the real IfcProject anchor no longer orphans the subtree; both walks are bounded by a shared visit budget that falls back to first-declared and warns once when spent, and never uses a partial descendant set as 'no cycle'. The Rust split keeps spatial.rs under the ratchet without touching the allowlist. Tests cover both declaration orders, the 3-node indirect cycle, order-independence of the resulting shape, and the budget fallback on both sides. Changeset is a parser patch (behavioural fix, no new export). Closes #4246; every required check is green on the current head.

Issue #4125, opened 2026-09-07
Three by-index STEP writers still split permissively: step-attribute-mutations.ts, retype.ts, step_text.rs


louistrue:
Retitled and narrowed. The original title counted implementations; the thing that
actually corrupts files is a WRITER on a permissive split, and three of those are still
live. The consolidation ask is declined, with the measurement below.

## What is still broken, reproduced

Record with two undoubled apostrophes, which is what an authoring tool emits when it
forgets to double them:

    #1=IFCWALL('g',$,IFCLABEL('a's'),$,IFCLABEL('b's'),#5,#6,'T',.SOLIDWALL.);

**`packages/export/src/step-attribute-mutations.ts`** (`:151`, `:189`, `:216`, `:221`).
Mutating `Description` produces:

    #1=IFCWALL('g',$,IFCLABEL('a's'),$,IFCLABEL('b's'),'NEWDESC',#6,'T',.SOLIDWALL.);   attributed: true

`NEWDESC` landed in `ObjectPlacement`'s slot and `#5` was deleted. The control with
doubled quotes writes the correct slot.

**`packages/export/src/retype.ts`** (`:186` then `:129-146`). Retyping `IFCWALL` to
`IFCWALLSTANDARDCASE` on the same record emits ELEVEN attributes for a nine-attribute
class.

**`rust/export/src/step_text.rs`** (`:201-204`, `apply_attr_mutations`). Permissive
`split_top_level_args` at `:140-179`, `args[*idx] = val`, and the input is `raw` from the
user's file (`step.rs:263-265`). Same mechanism. Read, not run.

All three import or implement the PERMISSIVE splitter while the validating one sits in
the same file (`packages/export/src/step-argument-parser.ts`). **No published export is
needed**: swap `splitTopLevelArgs` for `splitTopLevelStepArguments` and handle `null` by
dropping the edit and reporting, exactly as `export/src/unit-normalize.ts:320-324`
already does.

## Already done

- `mutate.ts` -> #4174 (`70413fa7b`)
- `replaceStepArgument` -> #4173

Also validating and self-consistent, no action: `subset-relations.ts`'s `spliceArgument`
(guarded by a 6-attribute check plus `SINGLE_REF_RE` on every member) and
`unit-normalize.ts`. Lower exposure, input is repo-generated rather than the user's file:
`step-overlay-attribute-overrides.ts`, and the fixture generators under `scripts/xmatch/`
and `tools/demo-kit/`.

[...]

PR #4291, opened 2026-09-09
fix(export): splitTopLevelArgs misreads a comma inside a STEP comment as an argument boundary

### LINKED ISSUE #4227: fix(export): splitTopLevelArgs misreads a comma inside a STEP comment as an argument boundary, leaking hidden references
## Summary

`packages/export/src/step-argument-parser.ts` has two top-level STEP argument splitters. `splitTopLevelStepArguments` is comment-aware (it skips `/* ... */` regions atomically so their content can't be misread as commas, parens, or quotes). Its sibling, `splitTopLevelArgs` (lines ~40-94), has **no comment-skip logic at all**, yet it is the primitive used by six live write-path call sites: `reference-collector.ts` (`filterHiddenRefsFromRelationshipLine`, `collectReferencedEntityIds`), `subset-entity-reader.ts` (`readEntityArgs`, feeding `anonymize-placement.ts` / `anonymize-scrub.ts` slot resolution), `merged-empty-containers.ts`, `step-overlay-attribute-overrides.ts`, and `schema-converter-attr-remap.ts`.

When a `/* ... */` comment appears **inside** an argument list and contains a comma, `splitTopLevelArgs` reads that comma as a top-level argument separator. This is not a coverage gap — it is a live data-corruption bug: a reference to a deliberately-withheld (hidden/deleted/tombstoned) entity ships into the written IFC file instead of being stripped.

## Reproduction

Verified on `upstream/main` at `92e590337` (`fix(export): validate the split before writing a STEP attribute by index (#4125) (#4213)`), calling `filterHiddenRefsFromRelationshipLine` directly — the function that strips references to hidden entities before a relationship line is written.

**Control** — hidden ref, no comment:

```
filterHiddenRefsFromRelationshipLine(
  "#10=IFCRELVOIDSELEMENT('guid',$,$,$,#1,#5);",
  id => id === 5
)
```

Real output: `null` — correct. `#5` is a mandatory single-valued slot and it's hidden, so the whole line is withheld.

**Adversarial** — same hidden ref, with a `/* ... */` comment containing a comma placed immediately before it:

```
filterHiddenRefsFromRelationshipLine(
  "#10=IFCRELVOIDSELEMENT('guid',$,$,$,#1,/* void, comment */#5);",
  id => id === 5
)
```

Real output:

```
"#10=IFCRELVOIDSELEMENT('guid',$,$,$,#1,/* void, comment */#5);"
```

The line is returned **unchanged**. `#5` — a reference to an entity the caller explicitly marked hidden — ships into the output file as a dangling reference.

## Mechanism

`splitTopLevelArgs` scans char-by-char tracking quote state and paren depth, but treats `/` and `*` as ordinary characters. The comma inside `/* void, comment */` is seen at paren depth 0 and outside any quoted string, so it is read as a top-level argument separator. The slot that should have been `/* void, comment */#5` splits into two phantom slots, `/* void` and `comment */#5`. Neither matches the bare-reference regex (`^#(\d+)$`) the caller uses to test "is this slot the hidden id", so the hidden-reference exclusion check never fires, and the original text — hidden reference included — is reconstructed and returned as-is.

Existing coverage in `reference-collector.test.ts` (the "commented relationship line" cases) only exercises a comment placed **before** the record's `(`, which never reaches `splitTopLevelArgs` at all (it's outside `argsText`). Nothing in the current suite constructs a comment **inside** the argument list, which is why this has gone unnoticed.

A nested-list sibling shape, e.g. filtering `#3` out of `(#2,#3)` in `#10=IFCRELASSIGNSTOGROUP('guid',$,$,$,(#2,#3),#1,$);`, currently filters correctly — but only because the phantom slot boundaries a comment would introduce don't happen to collide with that particular list's own commas in the cases exercised so far. That should not be relied on; it is unguaranteed for other input shapes given the same missing comment-skip logic.

## Why this matters

This sits on the STEP **write path**: any exported IFC file that goes through `filterHiddenRefsFromRelationshipLine` (or any of the other five call sites) with a relationship record whose argument list happens to contain a `/* ... */` comment can silently emit a reference to an entity the export was told to withhold — producing a dangling `#id` reference in a file handed to a downstream consumer.

## What's already correct, to reuse

`splitTopLevelStepArguments` in the same file already solves exactly this problem: it skips a `/* ... */` region as one atomic unit (open marker through the matching `*/`, or to end-of-input when unterminated) before doing any comma/paren/quote bookkeeping. The fix here is to bring `splitTopLevelArgs` up to that same standard — reusing that logic rather than a second, independently-written implementation — not to invent a new comment-handling scheme.

BIMvoice:
## Summary

Closes #4227.

`step-argument-parser.ts` had two top-level STEP argument splitters. `splitTopLevelStepArguments` skipped a `/* ... */` comment atomically so its content couldn't be read as structure; its sibling `splitTopLevelArgs` — the primitive `reference-collector.ts`, `subset-entity-reader.ts`, `merged-empty-containers.ts`, `step-overlay-attribute-overrides.ts`, `schema-converter-attr-remap.ts`, and `anonymize-placement.ts` all use on the write path — had no comment handling at all. A comma inside a comment sitting inside an argument list split a slot in two, and on `filterHiddenRefsFromRelationshipLine` that phantom split let a reference to a deliberately hidden/deleted entity survive the exclusion check and ship into the written IFC file: `#10=IFCRELVOIDSELEMENT('guid',$,$,$,#1,/* void, comment */#5)` came back unchanged instead of withheld.

Both splitters now share one `skipStepComment` helper (`step-comment-skip.ts`) so the rule can't drift between them again. Fixing the split alone left the reported reproduction open: `filterHiddenRefsFromRelationshipLine` and `refGroupFromArg` test "is this slot a bare reference" with a regex that didn't tolerate a comment glued to the `#N` it now correctly keeps as one slot, so `BARE_REF_RE` gained the same STEP-trivia tolerance `RECORD_PREFIX_RE` already gives the type-name/`(` boundary, exported from `reference-collector.ts`, and adopted by `anonymize-placement.ts`'s `parseRef` and `merged-empty-containers.ts`'s `singleRef` in place of their own narrow copies.

A third, independently-drifted copy of the same splitter carried the same gap: `packages/cli/src/commands/subset-relations.ts`'s own `splitTopLevelArgs` (a documented near-twin, copied rather than imported because `@ifc-lite/export`'s `exports` map exposes only `.`). Fixed the same way, with a local (not imported, same export-barrier reason) copy of `skipStepComment`.

## Branch history note

This branch was written against an older `main` and had gone stale (31 lines removed from `.github/workflows/test.yml`, a 55-line deletion in `rust/geometry/src/csg_capture_tests.rs`, plus `Cargo.lock`/`Cargo.toml`/`rust-major-offset.json`/three `rust/*/Cargo.toml` — all stale-base artifacts that would have silently reverted merged work). Brought current by merging `upstream/main` in first (clean, no conflicts) and confirming the diff against `upstream/main` contains only the `packages/export`, `packages/cli`, and changeset files. The original eight commits (several `wip:`) were squashed into one for review, authorship preserved.

## Call-site table

| Call site | Routed through shared fix? | Verdict |
|---|---|---|
| `step-argument-parser.ts`: `splitTopLevelArgs` | Defines `skipStepComment` call — the fix itself | Fixed |
| `step-argument-parser.ts`: `splitTopLevelStepArguments` | Now calls shared `skipStepComment` (previously had the logic inlined) | Fixed, de-duplicated |
| `reference-collector.ts`: `filterHiddenRefsFromRelationshipLine`, `refGroupFromArg` | Imports `splitTopLevelArgs`; also adopted trivia-tolerant `BARE_REF_RE` | Fixed |
| `subset-entity-reader.ts`: `readEntityArgs` | Imports `splitTopLevelArgs` | Fixed (via shared import, no direct test added) |
| `merged-empty-containers.ts`: `singleRef` / `topLevelAttrs` | Imports `splitTopLevelArgs`; adopted `BARE_REF_RE` | Fixed |
| `anonymize-placement.ts`: `parseRef` | Imports `splitTopLevelArgs`; adopted `BARE_REF_RE` | Fixed |
| `step-overlay-attribute-overrides.ts` | Imports `splitTopLevelArgs` | Fixed (via shared import, no direct test added) |
| `schema-converter-attr-remap.ts` | Imports `splitTopLevelArgs` | Fixed (via shared import, no direct test added) |
| `packages/cli/src/commands/subset-relations.ts`: local `splitTopLevelArgs` | NOT imported — local `skipStepComment` copy, documented as a fourth copy | Fixed independently, own dedicated tests. Justified: `@ifc-lite/export`'s `exports` map exposes only `.` and `index.ts` re-exports none of the three helpers (confirmed by grep); the module already re-implements two other `@ifc-lite/export` primitives for the identical reason, tracked as consolidation follow-up #4125. |
| `packages/cli/src/commands/spatial-ancestors.ts`, `step-serialization.ts` | Reference `splitTopLevelArgs` by name in comments/imports only — not independent implementations | Not a copy, not checked further |

## Verification performed

- Reproduced the exact issue repro on `upstream/main` before the fix (not separately re-run after checkout since the branch already contains the fix; verified the mechanism read directly against the issue text and the code).
- `packages/export`: 106/106 targeted tests pass (`step-argument-parser`, `reference-collector`, `anonymize-placement`, `merged-empty-containers`); full package suite 1231 passed / 36 skipped (pre-existing skips), 0 new failures.
- `packages/cli`: `subset-relations.test.ts` 15/15 pass. The rest of the `packages/cli` suite hits a pre-existing environment blocker unrelated to this change (`Cannot find package '@ifc-lite/parser'` / `@ifc-lite/codegen/ifc4` resolution failures from stale `dist`/dangling symlinks in the local dev environment) — reproduces on plain `upstream/main`, not a regression from this PR.
- Mutation-tested the shared `skipStepComment` helper (turned it into a no-op): 7 tests across **two files** (`step-argument-parser.test.ts`, `reference-collector.test.ts`) went red, spanning `splitTopLevelArgs`, `splitTopLevelStepArguments`, and `filterHiddenRefsFromRelationshipLine` — confirms genuine shared routing, not per-site self-guarding. Restored and re-verified green.
- Mutation-tested the CLI's independent local `skipStepComment` copy: 4 tests in `subset-relations.test.ts` went red, including the end-to-end `planSpatialRelations` case. Restored and re-verified green.
- Verified the "masked test" claim from the branch's own commit history: swapping in the pre-fix version of `merged-empty-containers.test.ts` (which exercised a comment-wrapped `RelatingObject`, not a list member) against a mutated narrow `singleRef` regex passed all 8 tests — the bug was invisible to that shape because `recordBlocks`' fallback masks it. The current test (comment-wrapped list member) catches the same mutation. Confirms the fix commit's claim was accurate.
- `node scripts/check-module-size.mjs`: exits 0, 0 files newly over budget.
- Both directions verified via existing/added tests: a comma inside a comment is not a boundary, and ordinary commas, nested parens, quoted strings, and Part 21 doubled-quote escapes still split correctly.
- Changeset (`fix-4227-split-top-level-args-comments.md`) verified against the final merged diff; starts with `---`, no license header.

## Not checked

- `subset-entity-reader.ts`, `step-overlay-attribute-overrides.ts`, and `schema-converter-attr-remap.ts` were not given dedicated new tests in this branch (they were already covered indirectly by importing the shared, now-fixed `splitTopLevelArgs`); no direct mutation test was run against those specific call sites.

## Summary by CodeRabbit

- **Bug Fixes**
  - STEP comments containing commas, parentheses, or quotes no longer disrupt argument parsing.
  - Improved handling of comment-wrapped references across exports and relationship filtering.
  - Placement anonymization and empty-container removal now work correctly with commented references.
  - Preserved existing behavior for strings, escapes, nested lists, multiline text, and malformed input validation.

- **Tests**
  - Added regression and end-to-end coverage for comment-aware parsing, reference filtering, placement anonymization, and container cleanup.


louistrue (approved):
Reviewed for merge readiness: the comment-aware STEP argument scanner is shared across the affected paths and the regression fixtures distinguish commas inside comments from real argument boundaries. Final merge remains contingent on fresh required checks.

louistrue (approved):
Reviewed the full diff. splitTopLevelArgs (and the CLI's near-twin in subset-relations.ts) now skip a /* ... */ comment atomically through the shared skipStepComment helper, so a comma inside a comment no longer splits a slot and lets a hidden reference survive filterHiddenRefsFromRelationshipLine; BARE_REF_RE gains STEP-trivia tolerance and replaces the three narrower copies (reference-collector, anonymize-placement's parseRef, merged-empty-containers' singleRef), each with a regression test showing the concrete wrong output it produced. Tests cover the exact #4227 line, nested lists, apostrophes inside comments, the close-marker off-by-one, unterminated comments, and every pre-existing rejection case in both splitters, plus the CLI planSpatialRelations end-to-end. Patch changesets on export and cli. Closes #4227; every required check is green on the current head.

PR #4293, opened 2026-09-09
fix(sdk): forward bounds in bim.bcf's section/clipping-plane converters

### LINKED ISSUE #4265: bim.bcf.clippingPlaneToSectionPlane drops its required bounds argument (sibling of the sectionPlaneToClippingPlane bug fixed for #4251)
## Summary

`packages/sdk/src/namespaces/bcf.ts`'s `clippingPlaneToSectionPlane` wrapper drops its required `bounds` argument before forwarding to `@ifc-lite/bcf`'s library function, which needs both parameters.

```ts
/** Convert BCF clipping plane to viewer section plane. */
async clippingPlaneToSectionPlane(clippingPlane: unknown): Promise<unknown> {
  const mod = await loadBCF();
  return (mod.clippingPlaneToSectionPlane as AnyFn)(clippingPlane);
}
```

The library function it calls, `packages/bcf/src/viewpoint.ts:363-368`:

```ts
export function clippingPlaneToSectionPlane(
  plane: BCFClippingPlane,
  bounds: ViewerBounds
): ViewerSectionPlane {
```

requires `bounds` to convert an absolute BCF location back into ifc-lite's percentage-based section position. The SDK wrapper's own signature only accepts `clippingPlane` — there is no parameter to receive `bounds` at all, so it can never be forwarded regardless of what a caller passes.

## Verified — executed

Called the wrapper the documented way (a clipping plane plus bounds, matching the sibling library test in `packages/bcf/src/viewpoint.test.ts`):

```ts
const ns = new BCFNamespace();
const clippingPlane = { location: { x: 0, y: 0, z: 5 }, direction: { x: 0, y: 0, z: -1 } };
const bounds = { min: { x: -10, y: -10, z: 0 }, max: { x: 10, y: 10, z: 10 } };
await ns.clippingPlaneToSectionPlane(clippingPlane, bounds);
```

Actual output:

```
TypeError: Cannot read properties of undefined (reading 'max')
    at Module.clippingPlaneToSectionPlane (packages/bcf/src/viewpoint.ts:383:26)
    at BCFNamespace.clippingPlaneToSectionPlane (packages/sdk/src/namespaces/bcf.ts:241:17)
```

Run against `packages/sdk/src/namespaces/bcf.ts` on `upstream/main` at `39ec95eb73e5aca47877b744efcf6113f71aa18d`, via `tsx` importing the source module directly (dependencies resolved from a pre-built sibling checkout; no source under `packages/sdk` or `packages/bcf` was modified).

## Relationship to the sibling fix

The reverse wrapper, `sectionPlaneToClippingPlane`, had exactly this bug — its own signature also accepted only one parameter and silently discarded `bounds`. That one has since been fixed on branch `fix-sdk-bcf-viewpoint-adapter` (issue #4251, commit `6a59e079069dc069bb85dbc462b18cb5d7e1df78`):

```ts
async sectionPlaneToClippingPlane(section: unknown, bounds: unknown): Promise<unknown> {
  const mod = await loadBCF();
  return (mod.sectionPlaneToClippingPlane as AnyFn)(section, bounds);
}
```

`clippingPlaneToSectionPlane` — the read-path counterpart, used when converting an incoming BCF viewpoint's clipping plane back into viewer section-plane state — was left with the same defect.

## Consequence

Any caller converting a BCF clipping plane back into ifc-lite's section-plane shape via this SDK wrapper — including anyone following `createViewpoint`'s own error message, which for the sibling case recommends calling the paired wrapper with `(value, bounds)` — gets a raw, unhandled `TypeError` from inside the library, not a usable result or a clear SDK-level error.

### LINKED ISSUE #4294: bim.bcf.cameraToOrthogonal() drops the required viewToWorldScale, silently emitting an orthogonal camera without it
## Finding

`bim.bcf.cameraToOrthogonal()` in `packages/sdk/src/namespaces/bcf.ts` forwards one argument to a library function that requires two, so every orthogonal camera it produces is missing `viewToWorldScale`.

The wrapper on `upstream/main`:

```ts
async cameraToOrthogonal(camera: unknown): Promise<unknown> {
  const mod = await loadBCF();
  return (mod.cameraToOrthogonal as AnyFn)(camera);
}
```

The function it forwards to, `packages/bcf/src/viewpoint.ts`:

```ts
export function cameraToOrthogonal(
  camera: ViewerCameraState,
  viewToWorldScale: number
): BCFOrthogonalCamera {
```

`viewToWorldScale` is not optional and has no default. It is returned verbatim:

```ts
  return {
    cameraViewPoint: bcfPosition,
    cameraDirection: direction,
    cameraUpVector: upVector,
    viewToWorldScale,
  };
```

So the SDK's result always carries `viewToWorldScale: undefined`.

## Why this is worse than its two siblings

This is the third wrapper in the same file with a signature that does not match its target. The other two — `sectionPlaneToClippingPlane` and `clippingPlaneToSectionPlane` (#4265) — dereference the dropped argument and throw a `TypeError` immediately, so a caller finds out at once.

This one does not throw. It returns a well-formed-looking object with one field quietly `undefined`, and `ViewToWorldScale` is what defines the extent of an orthogonal view: it is a required child of `OrthogonalCamera` in the BCF schema. A viewpoint written from it is either schema-invalid or opens at the wrong zoom in whatever tool reads it, with nothing in this codebase reporting a problem.

That is the same class as the two already fixed, but it fails silently rather than loudly, which makes it the more likely of the three to reach a real BCF file.

## Cause

The `AnyFn` cast. Every wrapper in this namespace casts its target to `AnyFn` before calling it, so TypeScript never checks arity or parameter types at the call boundary. Three of the wrappers in this file have now been found to disagree with their targets, and the compiler cannot catch the next one either.

## Scope note

Filed rather than folded into #4265, per the maintainer's instruction on that issue that a guard not in the table should be filed, not annexed. #4265's PR covers the two throwing converters only, and states this one explicitly as found-but-not-fixed.

Not checked: whether any shipped caller currently invokes `bim.bcf.cameraToOrthogonal()`, or whether a fix should add a required parameter to the wrapper (breaking its current callers, if any) versus supplying a documented default. That is a maintainer call about the SDK's public surface.

BIMvoice:
## Summary

Closes #4265. Closes #4294.

`bim.bcf.sectionPlaneToClippingPlane()` and `bim.bcf.clippingPlaneToSectionPlane()` (`packages/sdk/src/namespaces/bcf.ts`) each accepted only their first argument and silently dropped the `bounds` argument that `@ifc-lite/bcf`'s underlying `sectionPlaneToClippingPlane(sectionPlane, bounds)` / `clippingPlaneToSectionPlane(plane, bounds)` require to convert between a percentage-along-axis position and an absolute location. A documented call — passing a section/clipping plane plus the model's bounds, exactly as `createViewpoint()`'s own error message recommends for the write side — threw a raw, unhandled `TypeError` reading `bounds.min`/`bounds.max` from inside the library, not a usable result or an SDK-level error.

#4265 reported the read-path converter (`clippingPlaneToSectionPlane`). Per the issue's maintainer comments, the write-path converter (`sectionPlaneToClippingPlane`) has the identical defect on current `main` and is claimed by no open issue, so this PR fixes both rather than leaving the sibling behind. No issue currently covers `sectionPlaneToClippingPlane` specifically, so this PR does not claim a second closing reference for it — the fix is included because the maintainer asked whoever picked up #4265 to take it too.

Both wrappers now accept and forward `bounds`, matching their `@ifc-lite/bcf` signatures exactly. No coordinate-convention changes — these two wrappers are thin passthroughs to the library's own object-shaped `ViewerSectionPlane`/`BCFClippingPlane`/`ViewerBounds` types (unlike `createViewpoint()`/`extractViewpointState()`, which adapt the SDK's public tuple/x-y-z shapes), matching #4265's own repro which called them with the library's shapes directly.

## Correction: not stacked on #4286

An earlier version of this section claimed this PR was "stacked on #4286 (`fix/4251-bcf-createviewpoint-shapes`)". That was inaccurate — corrected here after independent verification.

`git merge-base --is-ancestor` fails in both directions between #4286's head and this PR's head: #4286 is not an ancestor of this branch, and this branch is not an ancestor of #4286. `git merge-base` between the two resolves to `a5f6c517`, which is #4286's **first** commit, not its tip (`d05c48b8`). This branch forked from that first commit and never picked up #4286's second commit, which added read-path axis coverage for `extractViewpointState()` (introduced after review found the axis-mapping table could be silently swapped with zero test signal).

So this PR is a **sibling** of #4286's tip — forked from its first commit — not a descendant of it, and it does not carry that second commit's read-path axis coverage.

**Merge-order implication:** #4286 should still land first, since both PRs touch `bcf.ts`/`bcf.test.ts` and this PR's diff was authored against #4286's first-commit state. That is sequencing to avoid conflicts, not a rebase requirement — this branch does not need to be rebased onto #4286's tip for either PR to be correct in isolation.

## Wrapper audit

Per the maintainer's note that this namespace's wrappers' signatures don't always match the `@ifc-lite/bcf` functions they forward to, I checked every wrapper in `packages/sdk/src/namespaces/bcf.ts` against its target in `@ifc-lite/bcf`. All match except the three fixed here — the two `bounds` converters above and, since #4294 was filed for it, `cameraToOrthogonal()`:

- `cameraToOrthogonal(camera: unknown)` forwarded to `@ifc-lite/bcf`'s `cameraToOrthogonal(camera: ViewerCameraState, viewToWorldScale: number)`, which has no default for `viewToWorldScale` — the wrapper always passed `undefined`, producing `viewToWorldScale: undefined` in the returned `BCFOrthogonalCamera` that only failed later at `bim.bcf.write()`. It now accepts and forwards `viewToWorldScale` (own changeset `fix-camera-to-orthogonal-scale.md`; this carries #4338's commit, which is superseded by this PR).

Both `bounds` converters also accept the SDK's public tuple `AABB` (converted via `toLibraryBounds()`), not only the library's object-shaped bounds — the tuple shape previously reached the library as `undefined` coordinates.

## Test plan

- [x] Reproduced both defects on `upstream/main` + #4286 (pre-fix) via `tsx`, matching #4265's own `TypeError: Cannot read properties of undefined (reading 'min'/'max')` signature exactly.
- [x] Extended `packages/sdk/src/namespaces/bcf.test.ts` with lock-in tests for both converters (real-coordinate assertions, not just presence/no-throw) plus a round-trip test (`sectionPlaneToClippingPlane` → `clippingPlaneToSectionPlane` reproduces the original section plane exactly).
- [x] Mutation-tested each converter independently (re-dropped `bounds`, confirmed a RED test naming the defect, restored, confirmed GREEN) — done separately for each, since one shared test passing doesn't prove both call sites are wired.
- [x] `npx vitest run src/namespaces/bcf.test.ts` — 14/14 passed (real pass, not blocked; `@ifc-lite/bcf`'s `jszip` dependency resolved fine for this file).
- [x] `npx tsc --noEmit` in `packages/sdk` — clean.
- [x] `node scripts/check-module-size.mjs` — exit 0 (`bcf.ts` at exactly 400 lines after re-trimming JSDoc to stay under the ratchet; no budget edited).
- [x] `git diff --stat upstream/main HEAD` — only #4286's files plus this PR's.

Two full-suite `packages/sdk` test files failed unrelated to this change and are pre-existing environment artifacts, not regressions: `src/context.test.ts` / `src/namespaces/export.test.ts` hit `Cannot find package '@ifc-lite/codegen/ifc4'` (stale local `dist`, reproduces on plain `upstream/main`), and `src/externalised-siblings.test.ts` hit the documented vite cold-transform flake (#2935). Neither touches `bcf.ts`/`bcf.test.ts`/`bcf-viewpoint.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed BCF viewpoint creation and extraction to preserve camera coordinates, section-plane axes, and clipping planes.
  * Bounds are now correctly handled when converting section planes and clipping planes.
  * Invalid camera data and enabled section planes without bounds now produce clear errors instead of corrupted viewpoints.

* **API Updates**
  * Added viewpoint state types and errors for improved type safety and diagnostics.
  * Viewpoint extraction and plane conversion now support bounds parameters.


louistrue (approved):
Reviewed the full diff. bim.bcf.sectionPlaneToClippingPlane / clippingPlaneToSectionPlane now accept and forward the bounds argument the underlying @ifc-lite/bcf converters require (previously dropped, which threw inside the library on bounds.min/bounds.max), accepting both the SDK tuple AABB and the library's object-shaped bounds via toLibraryBounds; cameraToOrthogonal forwards viewToWorldScale instead of silently producing an undefined ViewToWorldScale that only failed at write time. The GUID/colour passthroughs move to a base class purely for the module-size ratchet with no public-surface change. Tests cover both converters with both bounds shapes, the pre-fix call shapes, a round trip, and the write-time failure for the orthogonal camera. Changeset is minor on @ifc-lite/sdk. Closes #4265 and #4294; every required check is green on the current head.

Issue #4280, opened 2026-09-09
Sweep charter: pin the nine documented guards that mutation testing showed no test defends


louistrue:
## The class

A guard exists, its own comment explains why it is load-bearing, and deleting it leaves the package's whole suite green. The guard is correct as shipped. What is missing is anything that stops the next person from removing it as dead weight.

Nine of these were filed on 2026-09-09, each one proven the same way: apply the mutation, run the package suite, watch it stay green, revert. That is the right evidence and it is already attached to each issue. What they do not have is a shared decision about whether the class is worth paying for, which is what this charter is.

Approving it once here beats adjudicating it nine times in nine PRs.

## What is swept

Only the nine already filed. This charter does not authorise going looking for more.

| Issue | Guard | Package | Suite that stayed green |
|---|---|---|---|
| #4222 | `removeRemoteAnnotation`'s unconditional `saveToStorage`, cleanup half | apps/viewer | 16 tests |
| #4223 | `buildMaterialUsageIndex` duplicate-forward-target dedup | packages/parser | 111 files / 1212 tests |
| #4224 | `extractGroupMembersOnDemand` IFCX member-inclusion branch | packages/parser | 111 files / 1212 tests |
| #4229 | `sanitizeZipComponent`'s collision counter | packages/bcf | 23 files / 470 tests |
| #4238 | `BulkQueryEngine.select()` globalId and namePattern fail-closed guards | packages/mutations | 12 files / 240 tests |
| #4239 | `StoreEditor.addEntity()` stale-watermark collision guard | packages/mutations | 12 files / 240 tests |
| #4240 | `opsForComponentChange`'s `oursVisible` nulling loop | packages/merge | 109 tests |
| #4253 | `--top` NaN fallback, `aggregateFinite` empty-group null, rebase scope re-verification | packages/cli | 71 + 42 tests |
| #4258 | `drawingRequestQueue.cancel()`'s revision bump, masked by a re-request | apps/viewer | 3 tests |

Each is labelled `ready` and closable by its own PR. This charter is the record that the class was approved, not a replacement work item.

## What ends it

Those nine closed. Nothing else.

Specifically out of scope, so the sweep cannot grow into a coverage project:

- Any guard not in the table, however similar. File it, do not annex it.
- Rewriting the tests that already exist around the guard.
- Refactoring the guard itself. If a guard turns out to be wrong rather than untested, that is a different issue and it stops being part of this sweep.
[...]

PR #4297, opened 2026-09-09
test(parser): pin two undefended mutation-hunt gaps (#4223, #4224)

### LINKED ISSUE #4223: buildMaterialUsageIndex: dedup of duplicate forward targets has no test (material-resolver.ts)
## Summary

`packages/parser/src/material-resolver.ts`, `buildMaterialUsageIndex`, carries a comment defending a dedup step that has no test verifying it:

```
// Duplicate forward edges (malformed double-typing) must not
// double a target's contribution within this map key.
const uniqueTargets = targets.length > 1 ? [...new Set(targets)] : targets;
```

## Verification (mutation testing, `upstream/main` at `92e590337`)

Baseline: `packages/parser` = 111 test files, 1212 passed, 3 skipped.

Mutation applied:

```
- const uniqueTargets = targets.length > 1 ? [...new Set(targets)] : targets;
+ const uniqueTargets = targets;
```

Result: full `packages/parser` suite still passes 111 files / 1212 tests / 3 skipped — **no test exercises the dedup this comment claims is required.**

## Why this matters

If an `IfcRelDefinesByType` lists the same occurrence id twice in `RelatedObjects` (malformed but seen in the wild "double-typing"), or two separate relations end up targeting the same type→occurrence pair, `collectMaterialLeaves` runs twice for that occurrence — silently inflating that occurrence's weight/quantity contribution in the material usage totals shown in a user-facing panel.

## Framing

The implementation looks correct as written — this is a missing test for a documented invariant, not a confirmed live defect. Filing so a regression test can be added that pins the dedup behavior described in the comment.

### LINKED ISSUE #4224: extractGroupMembersOnDemand: IFCX-only member-inclusion branch untested in packages/parser
## Summary

`packages/parser/src/on-demand-extractors.ts`, `extractGroupMembersOnDemand`, has a branch that is never exercised by any test in `packages/parser`:

```
// IFCX stores ingest with an EMPTY entityIndex.byId (no STEP byte spans
// exist), so existence rides the EntityTable there: keep a member when
// EITHER source knows the id. STEP stores keep byId as the primary gate
// and resolve identically to before (#1622 IFCX follow-up).
if (!ref && (!tableType || tableType === 'Unknown')) continue;
```

## Verification (mutation testing, `upstream/main` at `92e590337`)

Baseline: `packages/parser` = 111 test files, 1212 passed, 3 skipped.

Mutation applied:

```
- if (!ref && (!tableType || tableType === 'Unknown')) continue;
+ if (!ref) continue;
```

Result: full `packages/parser` suite still passes 111 files / 1212 tests / 3 skipped.

## This is a package-level coverage gap, not a confirmed live bug

Every `packages/parser` test reaches `extractGroupMembersOnDemand` via the STEP ingestion path, where `entityIndex.byId` is always populated, so the `!ref` half of the guard alone is sufficient to pass every test in this package — the "EITHER source knows the id" branch the comment describes is simply never reached here.

The end-to-end behavior this branch protects **is** covered one layer up, in `packages/ifcx`'s `system-membership.test.ts`, which exercises the IFCX ingestion path where `entityIndex.byId` is empty. So this is not a live user-facing defect — it's a hole in `packages/parser`'s own test suite for a branch whose correctness currently depends entirely on a test in a different package.

Regression this branch guards against: for an IFCX-sourced store, a group/zone/system member known only via `store.entities.getTypeName(id)` (not via `entityIndex.byId`) would be silently dropped from the member list if this condition regressed.

## Ask

Add a `packages/parser`-local test that builds a store with an empty (or id-missing) `entityIndex.byId` while `store.entities.getTypeName(id)` resolves a known type, assigns that id to a group via `AssignsToGroup`, and asserts `extractGroupMembersOnDemand` still includes it — so this package's own suite pins the invariant instead of relying solely on `packages/ifcx`.

BIMvoice:
## Summary

Sweep charter #4280: two `packages/parser` guards are correct as shipped and each one's own comment explains why it's load-bearing, but deleting either leaves the whole package suite green.

- **#4223** — `buildMaterialUsageIndex`'s duplicate-forward-target dedup (`material-resolver.ts`): a duplicate `IfcRelDefinesByType` edge to the same occurrence must not double that occurrence's weight/quantity contribution in material usage totals.
- **#4224** — `extractGroupMembersOnDemand`'s IFCX member-inclusion branch (`on-demand-extractors.ts`): a member known only via `store.entities.getTypeName` (empty `entityIndex.byId`, the shape of real IFCX ingest) must still be kept in a group/zone member list.

New tests:
- `packages/parser/test/material-usage-duplicate-forward-target.test.ts` — builds a facade relationship graph with a genuine duplicate forward `DefinesByType` edge (two distinct relations, same type→occurrence pair) and asserts the material's usage row has exactly one entry for that occurrence with unmultiplied weight.
- `packages/parser/test/group-members-ifcx-typename-only.test.ts` — builds a synthetic store shaped like real IFCX ingest (`entityIndex.byId` missing the member id, `store.entities.getTypeName` still resolving it) and asserts the member survives `extractGroupMembersOnDemand`.

Both fixtures discriminate: the dedup test has a real duplicate edge pair (a fixture without one would pass either way), and the IFCX test genuinely deletes the member from `entityIndex.byId` before asserting the EntityTable fallback keeps it.

## The bar (per #4280)

**1. New tests pass on unmodified `main`:**
```
Test Files  113 passed (113)
     Tests  1214 passed | 3 skipped (1217)
```
(1212 baseline + 2 new, matching each issue's reported baseline of 111 files / 1212 tests.)

**2. Re-applying each issue's exact mutation, in isolation, turns only the new test red and leaves every pre-existing test in the package green:**

#4223 mutation (`material-resolver.ts`):
```diff
-const uniqueTargets = targets.length > 1 ? [...new Set(targets)] : targets;
+const uniqueTargets = targets;
```
Result:
```
Test Files  1 failed | 112 passed (113)
     Tests  1 failed | 1213 passed | 3 skipped (1217)
```
Only `material-usage-duplicate-forward-target.test.ts` fails.

#4224 mutation (`on-demand-extractors.ts`):
```diff
-if (!ref && (!tableType || tableType === 'Unknown')) continue;
+if (!ref) continue;
```
Result:
```
Test Files  1 failed | 112 passed (113)
     Tests  1 failed | 1213 passed | 3 skipped (1217)
```
Only `group-members-ifcx-typename-only.test.ts` fails.

Both mutations were reverted after verification; the diff here is test-only.

## Scope

Only #4223 and #4224 from #4280's table. No production code changed, no changeset (no API-surface change).

Closes #4223
Closes #4224

## Summary by CodeRabbit

- **Tests**
  - Added coverage confirming group members remain discoverable when indexed IFCX type information is available without STEP byte spans.
  - Added regression coverage ensuring duplicate material-definition relationships do not double-count an occurrence in material usage results.

### DIFF
--- a/packages/parser/test/group-members-ifcx-typename-only.test.ts
+++ b/packages/parser/test/group-members-ifcx-typename-only.test.ts
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+/**
+ * Package-level coverage for the IFCX-only branch of
+ * `extractGroupMembersOnDemand` (#1622 IFCX follow-up):
+ *
+ *   // IFCX stores ingest with an EMPTY entityIndex.byId (no STEP byte spans
+ *   // exist), so existence rides the EntityTable there: keep a member when
+ *   // EITHER source knows the id. STEP stores keep byId as the primary gate
+ *   // and resolve identically to before (#1622 IFCX follow-up).
+ *   if (!ref && (!tableType || tableType === 'Unknown')) continue;
+ *
+ * Every other `packages/parser` test reaches this function via the STEP
+ * ingestion path, where `entityIndex.byId` is always populated, so the
+ * `!ref` half of the guard alone would satisfy every one of them — the
+ * "EITHER source" half is otherwise verified only one package over, in
+ * `packages/ifcx`'s `system-membership.test.ts`. This test builds a store
+ * shaped like real IFCX ingest (`entityIndex.byId` missing the member id,
+ * `store.entities.getTypeName` still resolving it) so the invariant is
+ * pinned locally too.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { RelationshipGraphBuilder, RelationshipType } from '@ifc-lite/data';
+import { createSyntheticDataStore } from '../src/synthetic-data-store.js';
+import { extractGroupMembersOnDemand } from '../src/on-demand-extractors.js';
+
+describe('extractGroupMembersOnDemand — IFCX-shaped store (empty entityIndex.byId)', () => {
+  it('keeps a member known only via store.entities.getTypeName', () => {
+    const groupId = 1;
+    const memberId = 2;
+
+    const store = createSyntheticDataStore({
+      schemaVersion: 'IFC4X3',
+      fileSize: 0,
+      entities: [
+        { expressId: groupId, type: 'IfcSystem', name: 'HVAC System' },
+        { expressId: memberId, type: 'IfcPipeSegment', name: 'Pipe' },
+      ],
+    });
+
+    // Simulate real IFCX ingest: entityIndex.byId carries no STEP byte spans,
+    // so it stays EMPTY for the member — existence must ride the EntityTable
+    // (store.entities.getTypeName) instead.
+    store.entityIndex.byId.delete(memberId);
+    expect(store.entityIndex.byId.has(memberId)).toBe(false);
+    // Sanity: the EntityTable still knows the type — this is what the
+    // "EITHER source" branch is meant to fall back to.
+    expect(store.entities.getTypeName(memberId)).toBe('IfcPipeSegment');
+
+    // Wire the group -> member edge the way parseIfcx does (group -> member,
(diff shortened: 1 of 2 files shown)


louistrue (approved):
Reviewed for merge readiness: scope matches the linked ready issue, regression coverage is discriminating, and no unresolved correctness findings remain. Final merge remains contingent on fresh required checks.

PR #4304, opened 2026-09-09
test(cli): pin --top NaN fallback, aggregateFinite empty-group null, and rebase scope re-verification

### LINKED ISSUE #4253: cli: three correct guards with zero regression coverage (--top fallback, aggregateFinite empty-group, rebase scope re-verification)
## Three correct guards in `packages/cli` with zero regression coverage

Verified on `upstream/main` at `765501bac`. Three unrelated but correct pieces of defensive/re-verification logic in `packages/cli` have no test pinning them — each was confirmed by re-applying its mutation locally, watching the existing suite stay green, and reverting.

### B — `--top` NaN fallback in `extract-entities.ts`

`packages/cli/src/commands/extract-entities.ts:376-377`:

```ts
// Fall back to 20 on a missing or non-numeric `--top` (a `NaN` would make
// every `slice(0, topN)` return nothing).
const topRaw = Number.parseInt(getFlag(args, '--top') ?? '20', 10);
const topN = Number.isNaN(topRaw) ? 20 : topRaw;
```

Removing the `Number.isNaN(topRaw) ? 20 : topRaw` fallback (using `topRaw` directly) leaves all 71 tests in `extract-entities.test.ts` green — `--top` is used by the `--detect --report --json` triage path (`top: rows.slice(0, topN)` and two other `slice(0, topN)` call sites) but no test in the file passes a non-numeric `--top` value. `grep -n -- '--top' src/commands/extract-entities.test.ts` (before this issue) found zero matches.

Consequence if this guard regresses: `ifc-lite extract-entities model.ifc --detect --report --json --top banana` would silently return an empty `top` array instead of the documented default of 20 rows — no error, no warning, just zero rows.

### C — `count === 0 → null` guard in `query-aggregation.ts`

`packages/cli/src/commands/query-aggregation.ts:124`, inside `aggregateFinite`:

```ts
if (count === 0) return null;
```

Removing this line (letting execution fall through to `sum / count` = `NaN` for `avg`, or `Infinity`/`-Infinity` for `min`/`max` on an empty list) leaves all 42 tests across `query-output.test.ts` + `query-aggregation.test.ts` green. `aggregateFinite` is imported by three files (`query-output.ts`, `schedule-aggregate.ts`, and the file itself); no existing test calls it with an empty group.

The function's own doc comment states this `null` exists specifically so a caller can render "no data" rather than a fabricated number — see the companion issue for the live consumer-facing consequence of the JSON group-by path discarding this return value with `?? 0`.

### D — `rebaseInRef`'s scope-claim re-verification in `layer-history.ts`

`packages/cli/src/commands/layer-history.ts:314-323` (`rebaseLayerOnto`):

```ts
// Carried claims must be re-verified against the rebased writes and the
// new base — rebasing can change which entities the ops touch, so a
// claim that covered the original delta may no longer cover this one.
// Mismatch is flagged, never silently accepted (07 §7.2).
const scopeOps = deriveScopeOps(bare, newBase);
const { verified, violations } = verifyScopeClaims(manifestOut.scope_claim, scopeOps);
```

This is genuinely absent from test coverage, not merely blocked by the environment: `grep -rn "rebaseLayerOnto\|layerRebaseCommand" packages/cli/src/**/*.test.ts` found no hits, and `packages/cli/src/commands/layer-history.test.ts` (the file covering `layer bake|revert|log|diff|refs`) has no `describe('rebase', ...)` block at all.

A sibling shape already exists for the *publish* path in `layer.test.ts` (`describe('scope verification', ...)`, `'flags ops outside the declared claims'`), which asserts `publishLayer`'s `scopeVerified`/`violations` output when a delta touches a property outside its declared `scope`. The rebase path re-derives and re-verifies the same kind of claim against a *different* base (per its own comment, "a claim that covered the original delta may no longer cover this one"), but nothing exercises that re-verification.

### Scope of this issue

These are coverage gaps on already-correct code, not behavior changes — no design decision is attached, unlike the group-by JSON ambiguity filed as a companion issue. Tests for B and C have been prepared and mutation-verified locally (RED on reverting each guard, GREEN restored); a test for D was prepared the same way, porting the `layer.test.ts` scope-verification shape to `rebaseLayerOnto`. None of this is opened as a PR here — flagging the gap per the coverage-hunt workflow.

BIMvoice:
Closes #4253

## Summary

Three correct-but-untested guards in `packages/cli`, verified individually by mutation before this PR pinned them: `extract-entities.ts`'s `--top` NaN fallback, `query-aggregation.ts`'s `aggregateFinite` empty-group null, and `layer-history.ts`'s `rebaseLayerOnto` scope-claim re-verification. Each got its own test, each was mutation-verified to go red alone while every pre-existing test in its file(s) stayed green, and each mutation was reverted before moving to the next. Test-only change — no production code differs from `upstream/main`.

## B — `--top` NaN fallback (`extract-entities.ts:376-377`)

```ts
const topRaw = Number.parseInt(getFlag(args, '--top') ?? '20', 10);
const topN = Number.isNaN(topRaw) ? 20 : topRaw;
```

New tests in `extract-entities.test.ts`: run the `--detect --report --json` triage path once with no `--top` flag (baseline) and once with `--top banana`, and assert the returned `top` arrays are equal — i.e. a non-numeric `--top` behaves exactly like the default, not an empty array. A second test confirms an ordinary numeric `--top 1` still slices to the first row of the baseline (this path had no prior coverage either).

**Mutation**: `const topN = topRaw;` (dropped the `Number.isNaN(...) ? 20 :` fallback).

Pre-mutation: `Tests 73 passed (73)` (71 baseline + 2 new).

Post-mutation: `Tests 1 failed | 72 passed (73)` — only `'falls back to the default 20-row --top on a non-numeric value...'` failed (`top` came back `[]` instead of the 20-ish-row baseline: `Array.prototype.slice(0, NaN)` collapses to 0). The numeric `--top` test and all 71 pre-existing tests stayed green. This discriminates because only the NaN case is affected — `slice(0, topRaw)` is unchanged for any real number.

Reverted; guard confirmed correct.

## C — `aggregateFinite` empty-group null (`query-aggregation.ts:124`)

```ts
if (count === 0) return null;
```

New tests in `query-aggregation.test.ts` (`describe('aggregateFinite empty group', ...)`): call `aggregateFinite` directly with `[]` for all four modes (`avg`/`min`/`max`/`sum`) and assert `null` (`Object.is`, not just falsy), plus an all-non-finite input (`[NaN, Infinity, -Infinity]`) for `avg`. A companion test confirms the ordinary reduction (`avg`/`min`/`max` over `[2, 4]`) is unaffected.

**Mutation**: removed the `if (count === 0) return null;` line entirely, so execution falls into the `switch` with `count === 0`.

Pre-mutation: `Tests 44 passed (44)` (42 baseline across `query-output.test.ts` + `query-aggregation.test.ts`, + 2 new).

Post-mutation: `Tests 1 failed | 43 passed (44)` — only `'returns null, not NaN or Infinity, when no finite value was seen'` failed (`expected NaN to be null`, `avg` fell through to `0/0`). The "still reduces normally" test and all 42 pre-existing tests stayed green. Discriminates because only the zero-count path changes; any non-empty finite input takes the same `switch` branch either way.

Reverted; guard confirmed correct.

## D — `rebaseLayerOnto` scope-claim re-verification (`layer-history.ts:314-323`)

```ts
const scopeOps = deriveScopeOps(bare, newBase);
const { verified, violations } = verifyScopeClaims(manifestOut.scope_claim, scopeOps);
```

Zero prior coverage (confirmed: no `describe('rebase', ...)` block existed, and `grep -rn "rebaseLayerOnto|layerRebaseCommand" packages/cli/src/**/*.test.ts` had no hits). New `describe('rebase', ...)` block in `layer-history.test.ts`, porting the shape of `layer.test.ts`'s `'flags ops outside the declared claims'` publish-path test to the rebase path:

- **Violation case**: publish a candidate against `main` whose delta edits both `FireRating` (covered by its `model.mutate:Pset_FireSafety*@IfcWall` claim) and an acoustics property (not covered) — `scopeVerified` is `false` at publish time. Advance `main` past the candidate's base with an unrelated merge (a real rebase is then required). Call `rebaseLayerOnto` and assert the returned `scopeVerified` is `false` and `violations` equals `[{ path: 'wall-1', capability: 'model.mutate:Pset_Acoustics', ifcType: 'IfcWall' }]` — the actual violation data, not just a truthy/falsy check.
- **Clean case**: same setup but the candidate's delta only touches `FireRating`, fully covered by its claim. After advancing `main` the same way, `rebaseLayerOnto` returns `scopeVerified: true` and `violations: []`.

**Mutation**: made the re-verification a no-op — `const { verified, violations } = { verified: true, violations: [] };` (bypassing `verifyScopeClaims` entirely, `deriveScopeOps` still called but its result discarded via `void scopeOps;`), matching the issue's described "short-circuit so violations is always empty / verified is always true" shape for this guard.

Pre-mutation: `Tests 23 passed (23)` (21 baseline across `layer-history.test.ts` + `layer.test.ts`, + 2 new).

Post-mutation: `Tests 1 failed | 22 passed (23)` — only `'re-verifies a carried scope claim against the rebased delta and flags a violation'` failed (`expected true to be false`). The clean-path test and all 21 pre-existing tests (including the *publish*-path scope-verification suite in `layer.test.ts`, which exercises a structurally similar but code-path-distinct check) stayed green. Discriminates because only a genuine rebase-time violation is masked by the no-op; the clean-claim rebase takes the same `verified: true` result either way.

Reverted; guard confirmed correct. No guard was found to be wrong — all three behave exactly as documented.

## Verification

- Combined run of all five touched/target files: `Tests 140 passed (140)` (`extract-entities.test.ts` 73, `query-aggregation.test.ts` + `query-output.test.ts` 44, `layer-history.test.ts` + `layer.test.ts` 23).
- `node scripts/check-module-size.mjs`: exit 0, `0 new over 400`.
- `git diff --stat upstream/main`: only the three test files touched (141 insertions, 2 deletions, 0 production lines).
- No changeset: `packages/cli` is a published (non-`private`) package, but this PR changes zero source under `src/` outside test files — no API surface change. Following the precedent set by #4295 (`packages/mutations`, test-only, no changeset) and #4296 (`packages/bcf`, test-only, no changeset), both justified the same way ("test-only change, no published-package API surface touched").

## Summary by CodeRabbit

- **Tests**
  - Expanded CLI coverage for JSON result limits, including invalid values and default behavior.
  - Added verification for layer rebasing, including scope claims, uncovered edits, and clean covered changes.
  - Added aggregation tests for empty, non-finite, and mixed finite-value inputs across average, minimum, maximum, and sum operations.

### DIFF
--- a/packages/cli/src/commands/extract-entities.test.ts
+++ b/packages/cli/src/commands/extract-entities.test.ts
@@ -1139,3 +1139,44 @@ describe('triage (--detect) WASM disposal (#1959 P2 leak)', () => {
     expect(disposeSpy).toHaveBeenCalledTimes(1);
   }, 30_000);
 });
+
+describe('extract-entities --detect --report --json: --top', () => {
+  let stdoutSpy: ReturnType<typeof vi.spyOn>;
+
+  afterEach(() => {
+    stdoutSpy?.mockRestore();
+    vi.restoreAllMocks();
+  });
+
+  /** Runs the triage JSON path and returns the parsed `top` array. */
+  async function runTriage(extraArgs: string[]): Promise<unknown[]> {
+    const writes: string[] = [];
+    stdoutSpy = vi.spyOn(process.stdout, 'write').mockImplementation((chunk) => {
+      writes.push(String(chunk));
+      return true;
+    });
+    await extractEntitiesCommand([SAMPLE_IFC, '--detect', '--report', '--json', ...extraArgs]);
+    const parsed = JSON.parse(writes.join(''));
+    return parsed.top;
+  }
+
+  it('falls back to the default 20-row --top on a non-numeric value, not an empty array', async () => {
+    const baseline = await runTriage([]);
+    stdoutSpy.mockRestore();
+    const withBadTop = await runTriage(['--top', 'banana']);
+
+    // A NaN --top must behave exactly like omitting --top, not silently
+    // return zero rows (Array.prototype.slice(0, NaN) collapses to 0).
+    expect(baseline.length).toBeGreaterThan(0);
+    expect(withBadTop).toEqual(baseline);
+  }, 30_000);
+
+  it('still honors an ordinary numeric --top', async () => {
+    const baseline = await runTriage([]);
+    stdoutSpy.mockRestore();
+    const withTopOne = await runTriage(['--top', '1']);
+
+    expect(baseline.length).toBeGreaterThan(1);
+    expect(withTopOne).toEqual(baseline.slice(0, 1));
+  }, 30_000);
+});
--- a/packages/cli/src/commands/extract-entities.test.ts
+++ b/packages/cli/src/commands/extract-entities.test.ts
@@ -1161,14 +1161,31 @@ describe('extract-entities --detect --report --json: --top', () => {
   }
 
   it('falls back to the default 20-row --top on a non-numeric value, not an empty array', async () => {
-    const baseline = await runTriage([]);
-    stdoutSpy.mockRestore();
+    const processGeometry = GeometryProcessor.prototype.process;
+    vi.spyOn(GeometryProcessor.prototype, 'process').mockImplementation(async function (
+      this: GeometryProcessor,
+      ...args
+    ) {
+      const result = await processGeometry.apply(this, args);
+      const seed = result.meshes[0];
+      if (seed === undefined) throw new Error('fixture must produce at least one mesh');
+      return {
+        ...result,
+        meshes: Array.from({ length: 25 }, (_, index) => ({
+          ...seed,
+          expressId: index + 1,
+        })),
+      };
+    });
     const withBadTop = await runTriage(['--top', 'banana']);
 
-    // A NaN --top must behave exactly like omitting --top, not silently
-    // return zero rows (Array.prototype.slice(0, NaN) collapses to 0).
-    expect(baseline.length).toBeGreaterThan(0);
-    expect(withBadTop).toEqual(baseline);
+    // Pin the documented default itself. Deriving this expectation from an
+    // omitted --top invocation would let both paths regress to the same value.
+    expect(withBadTop).toHaveLength(20);
+    expect(withBadTop.map((row) => (row as { expressId: number }).expressId)).toEqual([
+      1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
+      11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
+    ]);
   }, 30_000);
 
   it('still honors an ordinary numeric --top', async () => {
--- a/packages/cli/src/commands/layer-history.test.ts
+++ b/packages/cli/src/commands/layer-history.test.ts
@@ -11,7 +11,7 @@ import { getRef, loadLayer, loadRefLayers } from './layer-store.js';
 import { publishLayer } from './layer-publish.js';
 import { diffLayerStacks } from './layer-diff.js';
 import { mergeIntoRef } from './layer-merge.js';
-import { bakeRef, logRef, revertInRef } from './layer-history.js';
+import { bakeRef, logRef, rebaseLayerOnto, revertInRef } from './layer-history.js';
 import { createRef, listRefs, protectRef } from './ref.js';
 import { CLASS, FIRE, makeDelta, setupMain, tmpStore } from './layer-test-helpers.js';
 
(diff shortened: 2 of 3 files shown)


louistrue (approved):
Verified the addressed review finding and local evidence. The change now satisfies the linked ready issue; final project-level follow-up is tracked separately where applicable.

PR #4310, opened 2026-09-09
fix(parser): elementToStorey resolves duplicate direct storey containment first-declared (#4248)

### LINKED ISSUE #4248: elementToStorey and packages/query's containedIn() disagree on which storey an element is on for duplicate containment
For an element with duplicate spatial containment (contained under two different storeys — malformed authoring output, but real: I've seen it emitted by tools that don't guard against double-assignment), `packages/parser`'s `SpatialHierarchy.elementToStorey` and `packages/query`'s `EntityNode.containedIn()` return two different storeys for the exact same element, and nothing surfaces that they disagree.

## Fixture

`#1` IfcProject, `#2` IfcBuildingStorey (Storey A), `#3` IfcBuildingStorey (Storey B), `#4` IfcWall. Two `IfcRelContainedInSpatialStructure` edges naming the same element:

- `#2 -> #4` (Storey A contains Wall) — declared first
- `#3 -> #4` (Storey B contains Wall) — declared second

## Verified: the two paths disagree

Built directly against `SpatialHierarchyBuilder.build` on `upstream/main` at `92e59033`:

```
elementToStorey.get(4) = 3
```

`elementToStorey` is built by `buildNode` unconditionally overwriting the map entry for every contained element as it walks storeys in build order (`packages/parser/src/spatial-hierarchy-builder.ts`, the `ctx.elementToStorey.set(elementId, expressId)` inside the storey-containment block) — so whichever storey happens to be visited **last** wins.

`packages/query`'s `EntityNode.containedIn()` instead walks the relationship graph's inverse CSR edges for the element, which preserve first-declaration order (the same ordering `spatial-hierarchy-canonical-parent.ts` relies on and documents in detail). Reading the raw inverse edges for the same fixture:

```
inverse.getEdges(4, ContainsElements) = [2, 3]
```

so `edges[0].target === 2` — the **first-declared** parent, Storey A. `containedIn()`'s mechanism therefore resolves to `2`, while `elementToStorey.get(4)` resolves to `3`. Same file, same element, two APIs in the same codebase, two different answers, and neither one reports that the containment was ambiguous to begin with.

I'm not asserting which answer is "correct" — first-declared and last-declared are both defensible tie-breaks, and duplicate containment is itself malformed input with no spec-mandated resolution, so that call belongs to whoever owns the semantics of these two APIs. What's clearly a bug is that two callers asking "which storey is this element on" for the identical file can get two different answers with no way to detect the disagreement.

## Worth noting: no ambiguity signal anywhere

Neither `elementToStorey` nor `containedIn()` exposes any indication that an element had more than one containing storey. `SpatialHierarchy` also has no general count of unplaced/ambiguous elements — an element with duplicate containment silently picks a winner exactly like an element with a dangling container id would silently vanish. A caller has no way to know it happened.

## Reproduction (ready to paste into a test file)

Not pushed as a branch/test — proving this "wrong" would mean asserting one of the two answers is correct, which is the maintainer's call, not something a test should encode. Pasting the repro here instead:

```ts
it('elementToStorey and the query package containedIn() mechanism disagree on duplicate containment', () => {
  const strings = new StringTable();
  const entities = new EntityTableBuilder(4, strings);
  entities.add(1, 'IFCPROJECT', 'p0', 'Project', '', '');
  entities.add(2, 'IFCBUILDINGSTOREY', 'st0', 'Storey A', '', '');
  entities.add(3, 'IFCBUILDINGSTOREY', 'st1', 'Storey B', '', '');
  entities.add(4, 'IFCWALL', 'w0', 'Wall', '', '', true);

  const relationships = new RelationshipGraphBuilder();
  relationships.addEdge(1, 2, RelationshipType.Aggregates, 10);
  relationships.addEdge(1, 3, RelationshipType.Aggregates, 11);
  relationships.addEdge(2, 4, RelationshipType.ContainsElements, 20); // first-declared: Storey A
  relationships.addEdge(3, 4, RelationshipType.ContainsElements, 21); // second-declared: Storey B

  const hierarchy = new SpatialHierarchyBuilder().build(
    entities.build(),
    relationships.build(),
    strings,
    new Uint8Array(),
    { byId: { get: () => undefined } },
  );

  // parser's elementToStorey: last-declared containment wins.
  expect(hierarchy.elementToStorey.get(4)).toBe(3);

  // query's containedIn() mechanism: first-declared inverse CSR edge wins.
  const built = relationships.build();
  const inverseEdges = built.inverse.getEdges(4, RelationshipType.ContainsElements);
  expect(inverseEdges[0].target).toBe(2);

  // Same element, same file: 3 !== 2.
});
```

## Environment

Verified on `upstream/main` at `92e59033708882e9d40eaad0cddc7aab1468d2b4`, `packages/parser/test/spatial-hierarchy-builder.test.ts` baseline 12/12 passing.

BIMvoice:
Closes #4248

**Stacked on #4285** (`fix-4246-aggregation-back-edge`, currently open) — this branch is a fast-forward descendant of that PR's tip (both parser touches in `git diff --stat` below belong to #4285), so it must merge after it, and CI/`closingIssuesReferences` need #4285 merged first for a clean base.

## Ruling this implements

Maintainer decision on #4248: first-declared wins on both paths. `containedIn()` (`packages/query`'s `EntityNode`) already resolves this way and does not move. `elementToStorey` (`packages/parser`'s `SpatialHierarchyBuilder`) changes to match. Not re-litigated here — this PR implements the ruling, it doesn't re-argue it.

## What was wrong

For an element named in more than one `IfcRelContainedInSpatialStructure` edge (a malformed-but-real duplicate containment — two different storeys both containing the same wall), `elementToStorey` unconditionally overwrote its map entry while walking storeys in build order, so the answer was whichever storey the tree walk visited **last** — not the first-declared edge `containedIn()` already used (`relationships.inverse.getEdges(elementId, ContainsElements)[0]`, the same first-declared CSR order `computeCanonicalParent` documents and relies on).

## Fix

Before claiming an element for a storey, `elementToStorey`'s assignment now checks the element's first-declared inverse `ContainsElements` edge and only proceeds if the current storey is that edge's target — a global, order-of-visits-independent check, not "first/last write during traversal wins". A naive "first write during traversal" fix would still disagree with `containedIn()` whenever storey-tree traversal order (driven by `IfcRelAggregates`) diverges from the wall's `ContainsElements` declaration order — a real risk this repo has hit before with order-dependent bugs, so the tests mutate ORDER, not just values, to catch exactly that.

## Composition with #4285

#4285's cycle-skip only ever applies within `computeCanonicalParent`'s `Aggregates` pass (`relType === RelationshipType.Aggregates`) — it decides which *building* parents a contested *storey*, never touching `ContainsElements` resolution. Verified by reading the diff, then proved with a fixture: a storey inside a spurious mutual-aggregation back-edge (#4285's exact shape) that ALSO duplicate-contains a wall from two storeys — the wall still resolves to the first-declared storey regardless of which building the cycle guard assigns as that storey's parent (`composes with the #4246 aggregation back-edge cycle guard` test, `packages/parser/test/spatial-hierarchy-builder.test.ts`).

## Tests

`packages/parser/test/spatial-hierarchy-builder.test.ts`, new describe block `duplicate direct storey containment resolves first-declared, matching containedIn() (#4248)`:
- basic repro, both declaration orders (mirror-image fixtures, not a single fixture that happens to commute)
- the critical order-divergence case: `IfcRelAggregates` traversal order and `ContainsElements` declaration order deliberately point opposite ways, so neither "first-visited" nor "last-visited" naive fixes pass, only resolving off `ContainsElements` order itself does
- the #4285 composition fixture above

`packages/query/test/spatial-storey-parity.test.ts` (new file): builds ONE `RelationshipGraph` (same `RelationshipGraphBuilder`, same edge list) and feeds it into both `SpatialHierarchyBuilder.build()` and `EntityNode`, asserting `elementToStorey.get(wallId) === containedIn()?.expressId` — a genuine two-live-implementation agreement check, not a restated constant on each side. RED under the pre-fix code (verified: reverting to the unconditional overwrite fails this test with the pre-fix symptom, `elementToStorey` returning the last-declared storey while `containedIn()` returns the first-declared one).

Mutation-tested: reverted the fix to the unconditional overwrite, confirmed exactly the `#4248`-scoped tests go RED with the pre-fix symptom (`elementToStorey.get(4)` returning 3 instead of 2, etc.), reverted.

Full suites: `packages/parser` 111 files / 1220 tests passed, 3 skipped (pre-existing, unrelated). `packages/query` 20 files / 285 tests passed.

## Third path (apps/server's `canonical_parent`, Rust)

Checked `apps/server/src/services/data_model/spatial.rs`. Its `element_to_storey` is NOT a map — it's a `Vec<(u32, u32)>` built by pushing every `IFCRELCONTAINEDINSPATIALSTRUCTURE` pair in file order, so a duplicate-contained element ends up with BOTH `(element, StoreyA)` and `(element, StoreyB)` entries, in declaration order. It doesn't pick a single winner at this layer at all, so it doesn't share the TS side's "last write wins" bug, but it's also not literally "the same first-declared-wins map" — whichever downstream consumer flattens this Vec into a single answer (not checked in this PR; out of scope for the parser-side fix) determines what a Rust-backed caller ultimately sees. Flagging for awareness, not fixing here — the ruling only covers the two TS APIs.

## Scope: ambiguity signal deferred

The maintainer confirmed the ambiguity signal ("a caller should be able to detect that a containment was ambiguous") is in scope but standalone. Implementing it well means either a new field on `SpatialHierarchy` (e.g. an `ambiguousStorey: Set<number>`) or a diagnostics/warning-log convention, threading through both `elementToStorey`'s and `containedIn()`'s callers, plus tests — more than a small addition on top of this tie-break fix. Not implemented in this PR; will file as a follow-up issue referencing #4248 rather than grow this PR's scope.

## Verification

```
git diff --stat upstream/main HEAD
```
Only `.changeset/*`, `packages/parser/src/spatial-hierarchy-builder.ts`, `packages/parser/src/spatial-hierarchy-canonical-parent.ts` (from #4285), `packages/parser/test/spatial-hierarchy-builder.test.ts`, `packages/query/test/spatial-storey-parity.test.ts` — no phantom `.github/workflows/test.yml` or `rust/**` changes.

`check-module-size.mjs`: `spatial-hierarchy-builder.ts` is already 493/487 lines (6 over budget) at the point this branch forks from #4285 — a pre-existing violation from #4285, not introduced here. This PR's own net addition is 3 lines (496 total). Not touching the allowlist per "no budget edits".

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spatial hierarchy parsing when malformed aggregation relationships create circular parent-child references, preventing valid project subtrees from being omitted.
  * Improved duplicate storey-containment handling so elements consistently resolve to the first-declared reachable storey.
  * Ignored duplicate containment claims from spatial structures that are not reachable from the project.
  * Ensured non-storey containers do not override valid storey assignments.
  * Added warnings when invalid circular relationships are detected and skipped.


louistrue (approved):
Reviewed the full diff (after merging main, so the #4285 cycle guard is main's budgeted version on both sides). elementToStorey now implements the #4248 ruling: on duplicate direct containment the element resolves to its first-declared inverse ContainsElements edge, matching containedIn(), decided globally via computeReachableSpatialNodes rather than by tree-traversal order, and only among storey-like containers that buildNode actually reaches (an unreachable orphan storey or an earlier-declared IfcSpace cannot win and drop the element). apps/server's element_to_storey is deduped to the first-declared project-reachable row in the new spatial_lookups.rs so the server-client decoder's map agrees with the parser path. Tests cover both declaration orders, the traversal-order divergence case, composition with the #4246 guard, the unreachable-first-declared fallthrough, the IfcSpace case, a byte-identical TS/Rust parity fixture, and a cross-package elementToStorey/containedIn parity test that also documents the one remaining unreachable-storey divergence. Parser patch changeset. Closes #4248; every required check is green on the current head.

PR #4345, opened 2026-09-09
fix(extensions): correct the trust-boundary comment guarding expect.regex

### LINKED ISSUE #4326: packages/extensions test-runner regex guard: comment names the wrong trust-boundary trigger
## Finding

`packages/extensions/src/testing/runner.ts`'s regex-test guard carries this comment (unchanged by #4287, present on both `upstream/main` and `upstream/fix/4259-ids-pattern-redos-guard`):

```ts
} else if (expect.regex.length > MAX_GUARDED_REGEX_PATTERN_LENGTH) {
  // Length cap is a shallow defence — pathological short patterns
  // (`(a+)+$` is 7 chars and catastrophic) still get through. For
  // registry-distributed extensions a real defence would run the
  // regex in a Worker with a timeout, or use re2-wasm. v1 assumes
  // manifest tests are author-trusted; harden this when remote
  // bundles can land via the registry (RFC §10).
```

The premise — "harden this when remote bundles can land via the registry" — is wrong about *when*. Traced in code:

- **The hosted registry does not exist.** `docs/architecture/ai-customization/10-registry-and-signing.md` §8 lists "The hosted registry service" under "What's deferred to the Phase-5 build," gated on adoption criteria (≥50 exported flavors, ≥10 distinct authors) not yet met. No code in the repo fetches or installs an extension bundle from any network location — `packages/extensions/src/bundle/loader-node.ts` only reads from a local directory, and the only other entry point unpacks an in-memory `.iflx` byte string.
- **Untrusted (non-author) bundles already land today, independent of any registry.** `apps/viewer/src/components/extensions/ExtensionsPanel.tsx` lets a user drag a `.iflx` file onto a dropzone or click "Import" to install it. The RFC itself names this path explicitly: "The user can still side-load unsigned bundles via drag-drop / file picker — that's their explicit choice, not a registry feature" (RFC §10, §9 Non-goals). So "author-trusted" is already false for any `.iflx` a user receives from someone else and imports — a workflow that ships today, not one gated on Phase 5.
- **The manifest-test regex guarded here already runs against such bundles, reachable via one click:**
  - `ExtensionsPanel.tsx` → `useRunExtensionTests` → `ExtensionHostService.runTests(id)` (`apps/viewer/src/services/extensions/host.ts:422`) → `runBundleTests` — a "Run tests" button on every installed extension's row, including side-loaded ones.
  - `RepairQueuePanel.tsx`'s "Run check" / "Re-run" button → `ExtensionHostService.revalidateForSdk` (`host.ts:571`) → `revalidateAgainstSdk` → `runBundleTests` for every installed extension whose SDK range needs re-checking.
  - In both paths, `runBundleTests`'s `expect.regex` check (`runner.ts`, `new RegExp(expect.regex); re.test(text)`) executes directly in the viewer's own JS — not inside the per-extension sandbox, not in a Worker. A catastrophic pattern here blocks the viewer's main thread, not just the sandboxed extension.

## What #4287 does and does not change here

Read on `upstream/fix/4259-ids-pattern-redos-guard`: `runner.ts` now imports `MAX_GUARDED_REGEX_PATTERN_LENGTH` and `hasCatastrophicBacktrackingShape` (the boolean-returning primitives) from `@ifc-lite/regex-guard`, not the throwing `assertGuardedRegexPattern`/`compileGuardedRegex`. The `reasons.push(...)` reporting behaviour is unchanged — no throw was introduced for this caller, matching the PR description's own claim ("no behaviour change to the test runner's `expect.regex` check"). The shared guard's shape check also already includes the character-class fix for the #4318 bypass (`^(a+[)]?a+)+$`) as of commit `264bda08a` on that branch, so the mechanism itself is sound as currently checked in.

So the guard is fine, and #4287 is not a regression here. What's inaccurate is the comment's account of *why it's still only a shallow defence*: it should not name a future registry as the trigger for hardening, because the "remote/non-author bundle" scenario it's deferring to is already live via drag-and-drop import.

## Ask

Two small, independent items — neither is a new guard (that already exists and is sound):

1. Fix the comment (in `runner.ts`, wherever #4287 lands it) to describe the actual trust boundary: manifest-test regexes already run against non-author-trusted bundles today (side-loaded `.iflx`), not only once a hosted registry ships. Citing "RFC §10" as the trigger is misleading — RFC §10 covers *discovery/signing infrastructure*, not the side-load path that already exists.
2. Given (1), reconsider whether the "real defence" (Worker + timeout, or `re2-wasm`) belongs on the Phase-5 roadmap rather than staying an indefinitely-deferred "someday" — the shape-check + length-cap guard is a heuristic by its own doc comment, and the exposure it covers for is already reachable via a single UI click on a side-loaded bundle, evaluated on the main thread.

Not proposing a fix here — this is a premise correction + prioritization note, filed so the registry/Phase-5 work doesn't inherit the current comment's (wrong) assumption that nothing is exposed until a hosted registry ships.

BIMvoice:
## Summary

Closes #4326.

The comment above `runner.ts`'s `expect.regex` length-cap check named a future registry as the trigger for hardening this guard ("harden this when remote bundles can land via the registry (RFC §10)"). That registry doesn't exist: `docs/architecture/ai-customization/10-registry-and-signing.md` lists the hosted registry as deferred Phase-5 work, and no code in this repo fetches or installs a bundle from a network location — `loader-node.ts` reads a local directory; the browser side unpacks in-memory `.iflx` bytes.

The trust boundary is already crossed today, not in some future phase:

- `apps/viewer/src/components/extensions/ExtensionsPanel.tsx` lets a user drag-drop any `.iflx` file and install it (the registry RFC itself calls this out: "The user can still side-load unsigned bundles via drag-drop / file picker — that's their explicit choice, not a registry feature").
- From there, one click reaches the guarded regex: `ExtensionsPanel.tsx`'s "Run tests" → `host.runTests()` → `runBundleTests`, and `RepairQueuePanel.tsx`'s "Run check"/"Re-run" → `host.revalidateForSdk()` → `runBundleTests`. Both run `new RegExp(expect.regex)` against the bundle's own manifest text on the viewer's main JS thread — no sandbox, no Worker (confirmed: nothing in `runner.ts`, `host.ts`, or `sdk-revalidate.ts` dispatches to a Worker).

This PR only corrects the comment to name that real, already-crossed boundary instead of the nonexistent registry, and keeps it honest that the length cap + shape check remains a shallow heuristic, not a complete defence. No behavior changes — the guard (length cap, `hasRedosShape` catastrophic-backtracking shape check) is untouched.

Per the issue discussion, a Worker+timeout or a linear-time regex engine (re2-wasm) is the real fix for the underlying gap, but that's a larger change requiring its own sizing — not done here.

## Test plan
- [x] `packages/extensions` full vitest suite: 861/861 passed (incl. `src/testing/runner.test.ts`: 10/10)
- [x] `node scripts/check-module-size.mjs` — exit 0 (this file sits exactly at its existing 437-line allowlist budget)
- [x] `node scripts/check-changesets.mjs` — exit 0; no changeset added (comment-only, no published API surface change)
- [x] `git diff --stat upstream/main HEAD` — only comment lines changed in `packages/extensions/src/testing/runner.ts`

## Summary by CodeRabbit

- **Documentation**
  - Clarified safeguards and execution boundaries for regular expression checks during extension testing.
  - Documented that test runs execute on the viewer’s main thread without sandboxing or a Worker.


louistrue (approved):
Verified the addressed review finding and local evidence. The change now satisfies the linked ready issue; final project-level follow-up is tracked separately where applicable.

PR #4364, opened 2026-09-09
fix(export): distinguish empty isolation filter from no filter in GLB export

### LINKED ISSUE #4483: GLB export: an isolation filter matching nothing exports the whole model
`Export Visible Only` with an isolation filter that matches nothing exports the **entire model** instead of nothing.

**Cause.** `rust/export/src/gltf.rs` derived "is a filter active?" from emptiness:

```rust
isolated_active: !opts.isolated.is_empty(),
```

so *no filter* and *filter active, matched nothing* are indistinguishable, and the matched-nothing case takes the no-filter branch. `packages/export/src/reference-collector.ts` has the correct form already — `isolatedIds !== null && !isolatedIds.has(expressId)` — this path just reinvented it.

**Reachable from the UI.** This is the same defect family as #4328: a user picks a filter, gets everything. Found while chasing that report outward.

**Fix.** Thread `Option<Vec<u32>>` through Rust → wasm → TS so `None` means no filter and `Some([])` means active-but-empty. Note the callers matter as much as the signature: a default parameter only rescues callers that *omit* the argument, and `packages/cli/src/commands/export-rust-formats.ts` and `packages/mcp/src/tools/export.ts` both passed an explicit empty `Uint32Array` meaning "no filter" — that regressed `ifc-lite export --format glb` until fixed.

**Verified end to end** by rebuilding the real wasm chain (freshness confirmed via `require.resolve` resolving into the test worktree): no filter → 10312-byte GLB; `--type IfcWall` → 3744 bytes; matches nothing → exit 1, no file written.

PR **#4364** carries the fix (**#4386** stacks the OBJ twin on top). Raising this so the work has an issue to close — happy for it to be closed as a duplicate if you would rather fold it into an existing one.

BIMvoice:
## Summary

Stacks on #4333 (`fix-4328-export-visible-classfilter`) — includes its commits, so this diff will shrink to just the files below once #4333 merges.

`resolveExportVisibility` (added by #4333) already distinguishes "no isolation filter" (`null`) from "isolation active, matched nothing" (an empty-but-non-null `Set`) — but that distinction died at the GLB export boundary:

- `GLBExportDialog.tsx`'s from-bytes/wasm assembler: `toLocal` mapped both `null` and an empty `Set` to the same empty `Uint32Array`.
- `GLBExportDialog.tsx`'s from-meshes assembler: `hasIsolation = !!globalIsolated && globalIsolated.size > 0` is `false` for both.
- `rust/export/src/gltf.rs`: `isolated_active: !opts.isolated.is_empty()` — an empty `Vec` (from either origin) reads as "no filter".

Net effect: filter the hierarchy panel's Class tab to a type present only in a federated model's *other* member, enable "Export Visible Only", export GLB — and the whole primary model comes out, not nothing. Reachable with two federated models today; STEP/IFCX and the wiring into the resolver were already fixed by #4333, but this GLB-specific collapse survived it (both assemblers still had it).

## Fix

Carry the null-vs-empty distinction end to end, matching the convention `packages/export/src/reference-collector.ts` and `packages/renderer/src/entity-visibility.ts` already use:

- `GltfOptions::isolated` (Rust): `Vec<u32>` -> `Option<Vec<u32>>`. `None` = no filter; `Some(ids)` (empty or not) = an active allowlist.
- The wasm boundary (`export_glb.rs`, regenerated `pkg/ifc-lite.d.ts` via `scripts/build-wasm.sh`): `isolated: &[u32]` -> `Option<Vec<u32>>`, i.e. `Uint32Array | undefined` in TS. `undefined` = no filter; an empty `Uint32Array` = active, matching nothing.
- `packages/geometry`'s `ifc-lite-bridge.ts` / `index.ts` wrappers: `isolated` parameter default changes from `new Uint32Array()` to `undefined`.
- `GLBExportDialog.tsx`: `toLocal` now returns `undefined` for `null`/`undefined` input (was: empty array for both `null` and empty `Set`); `hasIsolation` is now `globalIsolated != null` (was: `.size > 0`, which is `false` for an active-but-empty set).

This ABI choice (`Option<Vec<u32>>` over, say, a separate `isolated_active: bool` parameter) matches an existing pattern already used elsewhere in the wasm bindings (`rust/wasm-bindings/src/api/gpu_meshes/batch.rs`, `zone_split.rs`) and needs no new parameter.

## Test plan

- [x] `rust/export/src/gltf_tests.rs::mesh_visible_empty_isolated_set_is_indistinguishable_from_no_filter_bug_repro` -- RED before the fix (an empty-but-active `GltfOptions.isolated` made `mesh_visible` return `true`), GREEN after.
- [x] `rust/export/src/gltf_tests.rs::isolation_three_states_over_a_real_fixture` -- `None` keeps the full model, an active-empty filter exports 0 meshes, an active filter matching 1 id exports exactly 1 (skips without a local fixture; the synthetic-mesh test above covers the same code path).
- [x] `apps/viewer/src/components/viewer/GLBExportDialog.isolation.test.tsx` (new) -- 5 tests covering both assemblers x all three states (no filter / active-empty / active-one-match) through the rendered dialog.
- [x] Mutation-tested: reverting the Rust `isolated_active` derivation back to the size-based check reddens exactly the repro test (416 other export tests stay green). Reverting the TS `toLocal` fix alone reddens exactly the "no classFilter passes isolated as undefined" test; reverting `hasIsolation` alone reddens exactly the "classFilter matching nothing exports zero meshes" test. Both restored.
- [x] `cargo test -p ifc-lite-export` -- 417 passed, 0 failed.
- [x] `cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings` -- clean.
- [x] `cargo check -p ifc-lite-export -p ifc-lite-wasm --target wasm32-unknown-unknown` -- clean.
- [x] `cargo test --workspace --exclude ifc-lite-wasm --no-fail-fast` -- clean (0 failures).
- [x] `rust/processing/tests/module_size_ratchet.rs` -- clean.
- [x] `node scripts/check-module-size.mjs` -- exit 0 (`GLBExportDialog.tsx` 517/531, `ifc-lite-bridge.ts` 937/937 exactly at its recorded budget, `index.ts` 1537/1550 -- no file pushed over budget; comments kept net-neutral or trimmed to fit).
- [x] `pnpm typecheck` (root, full workspace) -- clean, 1850/1850 test files in program.
- [x] Targeted viewer + geometry tests: `GLBExportDialog.isolation.test.tsx`, `GLBExportDialog.placement.test.tsx`, `glb.test.ts`, `exportVisibility.test.ts`, `exportVisibility.classfilter-e2e.test.ts`, `export-adapter.test.ts` (confirms #4333's 812/812-pinned `export-adapter.ts` is untouched and its own tests still pass) -- 55/55 passed. `packages/geometry` vitest suite -- 404/404 passed.
- [x] `check-test-wiring`, `check-changesets` -- clean.

Not checked: a live browser run (no browser in this environment).

## unqueued

Not confirmed to close a `ready` issue by itself (it's a follow-up defect found while auditing #4333/#4328, not filed as its own issue per instructions not to open one). Carries `unqueued`.

Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * GLB and OBJ exports now distinguish between no isolation filter and an active filter with no matches.
  * “Export Visible Only” respects class, storey, type, lens, and model visibility settings.
  * Filters with no matching entities export no geometry instead of the entire model.
  * CLI and MCP OBJ exports now report a clear error when no vertices are produced.

* **Documentation**
  * Clarified isolation-filter behavior for GLB and OBJ exports.

Closes #4483


louistrue (approved):
Reviewed end to end: `GltfOptions::isolated`/`ObjOptions::isolated` become `Option<Vec<u32>>`, the wasm bindings and `GeometryProcessor`/`IfcLiteBridge` carry `undefined` vs `Uint32Array` through, and every in-repo caller (GLB dialog both assemblers, CLI, MCP) now passes `undefined` for an inactive filter — the CLI/MCP follow-ups are exactly the callers a breaking change had to catch, hence the major bumps on geometry and wasm. Three-state tests over a real fixture on both exporters assert the exported express ids, the MCP tool rejects header-only OBJ, the exporting guide documents the contract, and the six review threads are resolved. Closes #4483.

PR #4391, opened 2026-09-09
fix(renderer): distinguish an active-but-empty isolate set from no isolation in shadow occluder collection

### LINKED ISSUE #4485: Viewer: isolated-out geometry still casts shadows (empty isolate set read as no isolation)
Geometry that has been isolated *out* still casts a shadow in the 3D view.

**Cause.** `packages/renderer/src/shadow-occluders.ts`'s `anyVisible()` re-derives the visibility rule instead of using the shared one, and gets the empty case wrong:

```ts
const hasIsolate = isolatedIds != null && isolatedIds.size > 0;
if (!hiddenIds && !hasIsolate) return true;
```

`entity-visibility.ts` documents `isolatedIds` as meaningfully nullable — `null`/`undefined` means *no isolation*, an empty `Set` means *isolate nothing*. The `size > 0` term collapses those, so an active-but-empty isolate set reads as "no isolation" and a fully isolated-out mesh keeps casting a full shadow.

It disagrees with every other consumer of the same rule — `isEntityVisible`, `classifyBatchVisibility` in the *same file*, `picking-manager.ts` and `scene.ts` all correctly test `isolatedIds != null`.

**Fix.** Have `anyVisible()` delegate to `isEntityVisible()` rather than re-deriving the rule, which removes the divergence instead of patching this instance of it.

**Verified**: mutation-tested (reverting the delegation reddens exactly the new test, 1 of 30), full `packages/renderer` suite 1449 passing. All three states asserted — no isolation casts, non-empty isolation casts only its matches, active-but-empty casts nothing.

Worth noting this is a rendering defect a user can see, not hardening: it surfaced from a codebase-wide sweep for the same `empty means absent` shape behind #4328, and it was the only instance outside the export paths.

PR **#4391** carries the fix. Raising this so the work has an issue to close.

BIMvoice:
## Summary

- Sweep for the "empty collection means absent" fail-open shape (same family as #4364/#4386's `isolated_active: !opts.isolated.is_empty()`), scoped to code paths not already claimed by an open PR.
- `anyVisible()` in `packages/renderer/src/shadow-occluders.ts` derived `hasIsolate = isolatedIds != null && isolatedIds.size > 0`, which collapsed an active-but-empty isolation `Set` (a real, reachable state — see `entity-visibility.ts`'s documented contract: `null`/`undefined` = no isolation, empty `Set` = isolate nothing) into "no isolation active". A textured mesh or standalone mesh under that state kept casting a shadow for the whole scene instead of casting nothing, disagreeing with every other visibility surface in the renderer (`isEntityVisible`, `classifyBatchVisibility`, `picking-manager.ts`, `scene.ts`).
- Fixed by delegating `anyVisible()` to the existing `isEntityVisible()` helper — the same canonical check `classifyBatchVisibility()` already uses in this file — instead of re-deriving the hide/isolate rule.

## Test plan

- [x] RED: added a test asserting `collectShadowOccluders` casts nothing under an active-but-empty `isolatedIds` set; failed against the pre-fix code (`actual` still contained a `'textured'` draw).
- [x] GREEN: same test passes after the fix.
- [x] Three-state coverage added: no isolation → everything casts; non-empty isolation → only the matched subset casts; active-but-empty isolation → nothing casts.
- [x] Mutation test: reverted `anyVisible()` to the old `hasIsolate`-gated logic — the new test fails (1 of 30); restored the fix — all 30 pass.
- [x] Full `packages/renderer` suite: 1449 passed, 1 pre-existing skip, 0 failures.
- [x] `node scripts/check-module-size.mjs`: `EXIT=0`, `0 new over 400`.
- [x] Changeset added (`@ifc-lite/renderer` patch).

No mocked boundary here — `collectShadowOccluders`/`anyVisible`/`isEntityVisible` are plain, dependency-free functions exercised directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed phantom shadows from meshes that are fully excluded by an active, empty isolation filter.
  * Shadow casting now correctly reflects visibility isolation: no filter casts normally, selective filters cast only matching meshes, and empty filters cast no shadows.

* **Tests**
  * Added coverage for all isolation-filter states affecting mesh shadow casting.

Closes #4485

### DIFF
--- a/.changeset/shadow-occluder-empty-isolate.md
+++ b/.changeset/shadow-occluder-empty-isolate.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/renderer": patch
+---
+
+Fix sun shadow occluder collection treating an active-but-empty isolation set the same as no isolation, which let a fully isolated-out textured mesh or standalone mesh keep casting a phantom shadow instead of casting nothing.
--- a/packages/renderer/src/shadow-occluders.test.ts
+++ b/packages/renderer/src/shadow-occluders.test.ts
@@ -119,6 +119,51 @@ describe('collectShadowOccluders', () => {
     assert.deepEqual(kinds, ['instanced']);
   });
 
+  // Three-state check on `isolatedIds`, matching `entity-visibility.ts`'s documented
+  // rule ("an EMPTY set means isolate nothing... do not collapse the two"):
+  //   1. no isolation (undefined)        → everything casts.
+  //   2. non-empty isolation             → only the matched subset casts.
+  //   3. active-but-empty isolation      → NOTHING casts (the bug: `anyVisible`'s
+  //      `hasIsolate = isolatedIds != null && isolatedIds.size > 0` collapsed this
+  //      to "no isolation active" and let the whole batch cast a phantom shadow).
+  it('casts nothing (not everything) for a textured/individual mesh under an active-but-empty isolate set', () => {
+    const mesh: Mesh = {
+      expressId: 900,
+      vertexBuffer: buf('mesh-v'),
+      indexBuffer: buf('mesh-i'),
+      indexCount: 6,
+      color: [1, 1, 1, 1],
+      transform: { m: originModelMatrix(undefined) },
+      hydrated: false,
+    } as unknown as Mesh;
+    const draws = collectShadowOccluders(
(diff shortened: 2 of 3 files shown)


louistrue (approved):

Issue #4203, opened 2026-09-09
Rust IfcType: generate the type universe per schema version and retire the legacy tables


louistrue:
Parent: #4201 (The whole schema, first class). Frontier: one schema source of truth, Rust half.

## Problem

`rust/core/src/generated/schema.rs` is generated from `IFC4X3_DEV_923b0514` alone. Every class that exists only in IFC2X3 or IFC4 resolves to `IfcType::Unknown(crc32)` (`schema.rs:1955`), which stores a hash, not the keyword, and whose `attribute_names()` is empty (`schema.rs:13409`). `rust/core/src/legacy_entities.rs` patches 26 of those names by hand, `rust/export/src/rooted_type.rs:113-166` patches 54 more for rootedness only, and `rust/export/src/schema_pad.rs:24-26` says why the conversion tables are hand-kept: "Rust has no per-schema attribute table".

`legacy_entities.rs:105-112` names the failure mode: an unmapped class is dropped from both the attribute export and meshing, "silent data loss, not a visible inconsistency (#3172)". #3179 (every legacy keyword labelled Unknown in the browser) and #3187 (IfcDoorStyle geometry dropped on every path) are instances.

## Scope

- Generate the Rust type universe and attribute tables per schema version from the same EXPRESS inputs as the TypeScript side (depends on #4202 landing the IFC2X3 schema file).
- `from_str` resolves any class of any supported version; an unrecognised keyword keeps its name, never only a hash.
- `attribute_names(version)` answers for every class of that version.
- Delete `legacy_entities.rs` and `LEGACY_ROOTED_TYPES` once the generated tables reproduce every row; the parity fixtures under `rust/export/tests/` and `scripts/check-legacy-entity-coverage.mjs` are the acceptance test, then retire with the tables they guarded.

## Ends when

`check-legacy-entity-coverage.mjs` reports zero droppable products with the legacy table removed, the rooted-type and upconvert parity fixtures pass unchanged, and the wasm mesh batch labels every class by its exact name.

## Out of scope

Geometry processors for any newly-resolved class; resolving a class is not meshing it.

---

**Originating vision:** [The whole schema, first class](https://github.com/LTplus-AG/ifc-lite/discussions/5038) (migrated from #4201). This issue remains independently planned and tracked.

BIMvoice:
Claiming this. Session 1789059598, branch fix-4203-rust-per-schema-types.

BIMvoice:
[...]

Issue #4202, opened 2026-09-09
Codegen: bring IFC2X3 into codegen and generate the runtime registry per schema version


louistrue:
Parent: #4201 (The whole schema, first class). Frontier: one schema source of truth.

## Problem

The TypeScript runtime schema registry is generated from `IFC4_ADD2_TC1` alone (`packages/parser/src/generated/schema-registry.ts`, header of every file under `packages/parser/src/generated/`). `packages/codegen/schemas/` holds only `IFC4_ADD2_TC1.exp` and `IFC4X3.exp`; there is no IFC2X3 EXPRESS file in the repository. The IFC2X3 table in `packages/data/src/ifc-schema/generated/entities-ifc2x3.ts` comes from a different source (vendored `SchemaInfo.*.g.cs`) and carries attribute names only: no EXPRESS types, no optional flags, no enums, no selects, no inverses.

The measured consequence is pinned in `packages/parser/src/entity-attributes-across-schemas.test.ts:9`: the registry answers an empty attribute list for 251 classes, and `packages/mcp/src/tools/schema-reach.test.ts:11` records 39 IFC2X3 and 80 IFC4X3 `IfcRoot` classes with no inheritance chain. `packages/export/src/attribute-slot-types.ts:35-43` states outright that attribute types "do not exist in this repo to be read".

## Scope

- Add the IFC2X3 EXPRESS schema (IFC2X3_TC1) to `packages/codegen/schemas/` alongside the two present.
- Generate a per-version runtime registry (entities, attributes with EXPRESS type and optionality, enums, selects, inheritance) for IFC2X3, IFC4 and IFC4X3 from the same generator, and select the registry by the model's `schemaVersion`.
- Keep the existing IFC4 answers byte-stable for every class IFC4 already knows: the diff fingerprint adapters hash these names, and the across-schemas test asserts this half.
- Retire the hand-copied mirror described in `packages/codegen/INTEGRATION.md` so `scripts/check-codegen-sync.mjs` is the only path.

## Ends when

The 251-empty and 39/80-no-chain pins drop to zero, the codegen sync gate covers the new outputs, and `docs/guide/parsing.md:507-509` states the registry source per version truthfully.

## Out of scope

The Rust `IfcType` enum, which has the same one-version shape (IFC4X3 only) and is a separate subissue. IFC5, which has no EXPRESS schema.

Issue #3187, opened 2026-08-25
Bare IfcType::from_str is used to classify at 11 sites, against the rule schema_helpers.rs states — IfcDoorStyle/IfcWindowStyle are dropped before they can become jobs


louistrue:
`schema_helpers.rs:278-280` states the rule: a consumer that wants an entity's *classification* must call `legacy_aware_ifc_type` on the keyword, because `DecodedEntity.ifc_type` is a bare `IfcType::from_str` and is deliberately literal. Nine sites hold the keyword and classify with a bare `from_str` instead.

Found while reviewing #3182, which fixes one instance (the browser mesh label in `gpu_meshes/batch.rs`). The rest are untouched.

## Actively dropping geometry

Four sites gate type-geometry candidates on `ends_with("STYLE")` then a bare subtype check:

```
rust/wasm-bindings/src/api/styling/prepass.rs:177-179
rust/wasm-bindings/src/api/gpu_meshes/prepass.rs:391-393
rust/processing/src/shard_classes.rs:143-146
rust/processing/src/processor/mod.rs:802
```

```rust
let ifc_type = IfcType::from_str(type_name);
if !ifc_type.is_subtype_of(IfcType::IfcTypeProduct) { continue; }
```

For `IFCDOORSTYLE` and `IFCWINDOWSTYLE`, both dropped by IFC4X3, `from_str` returns `Unknown`, `Unknown` is a subtype of nothing, and the candidate is discarded before it becomes a job. They also carry `has_geometry: false` in `legacy_entities.rs:77,81`, so the ordinary product route does not reach them either.

This is why #3182's changeset was corrected before merge: an earlier draft listed `IfcDoorStyle` and `IfcWindowStyle` among the keywords it fixed. They never reach the line it changed.

## Latent, same rule violation

Six sites compute a job's `ifc_type` with a bare `from_str`. They feed `detect_rtc_offset_with_fallback` and the `IfcTypeProduct` branch rather than a user-visible label, so nothing is visibly wrong today:

```
rust/wasm-bindings/src/api/styling/prepass.rs:68, 73, 94
rust/wasm-bindings/src/api/gpu_meshes/prepass.rs:349, 397, 419
rust/wasm-bindings/src/api/gpu_meshes/prepass_discovery.rs:135, 141
```

[...]

PR #4495, opened 2026-09-10
fix(export): generate a legacy entity's own attribute names per schema version

BIMvoice:
Refs #4203 (partial — see below; does not close it)

## What

`rust/export/src/model_props.rs`'s `render_attributes` read `entity.ifc_type.attribute_names()`, and `entity.ifc_type` is decoded via a bare `IfcType::from_str` in the tokenizer — `Unknown` for any IFC2X3/IFC4 entity removed by IFC4X3, whose `attribute_names()` is `&[]`. The row already got the correct legacy-resolved display type (`model.rs` uses `legacy_aware_ifc_type` for that), but every own-class attribute silently vanished from the attribute export, for all 26 names `rust/core/src/legacy_entities.rs` already resolves for geometry/rootedness purposes. A file with an `IFCDOORSTYLE` mesh correctly and an attribute row for it, with zero own-class attributes on that row.

Borrowing the resolved base type's attribute names is not a safe fix: `IFCDOORSTYLE` (IFC2X3/IFC4) ends `…, OperationType, ConstructionType, ParameterTakesPrecedence, Sizeable`; its resolved base type `IfcDoorType` (IFC4X3) ends `…, PredefinedType, OperationType, ParameterTakesPrecedence, UserDefinedOperationType` — same length, different names from index 8 on, so the base type's names would relabel `Sizeable`'s value as `UserDefinedOperationType` instead of merely dropping it.

`scripts/generate-legacy-attribute-names.mjs` generates `rust/core/src/generated/legacy_attribute_names.rs`: each legacy entity's OWN positional attribute names, read from the same EXPRESS-derived tables the TypeScript side already generates (`packages/data/src/ifc-schema/generated/entities-ifc2x3.ts`, `entities-ifc4.ts` — themselves produced from `packages/codegen/schemas/*.exp`), not a new EXPRESS parser and not an approximation. `render_attributes` now consults this table first, falling back to the modern enum's `attribute_names()` unchanged for every name the generated `schema.rs` already resolves.

## Why this is only half of #4203

I confirmed the dependency (#4202 → merged as #4218) and that the degradation #4203 describes is still live: `rust/core/src/generated/schema.rs:1058` (`Unknown(u32)`) and `:13421` (`Self::Unknown(_) => &[]`) are both present on `main` at `18c5aa18e`.

#4203 asks for two things: (a) generate the type/attribute universe per schema version, and (b) delete `legacy_entities.rs` and `LEGACY_ROOTED_TYPES` once the generated tables reproduce every row. This PR is a bounded slice of (a) only — it does **not**:
- change `IfcType::from_str`, which still returns `Unknown(u32)` for these names (rootedness, geometry routing, and every other `from_str` call site are unaffected, unchanged);
- generate `legacy_entities.rs`'s base-type mapping (it is already exhaustive per `scripts/check-legacy-entity-coverage.mjs`'s definition, so generating it would change provenance, not coverage — a separate, lower-value PR);
- attempt the deletion in (b) at all, per the scoping in this issue.

A full per-version `IfcType`/`from_str` (the deeper architectural ask) touches ~145 files across `rust/core`, `rust/export`, `rust/geometry`, `rust/processing` and `rust/wasm-bindings` that consume the canonical `IfcType` enum — too large for one reviewable PR. This PR fixes the one concretely reproducible, currently-live defect in the attribute-export path without that rewrite, and sets up the same "generate from EXPRESS, prove parity" pattern (a) needs elsewhere.

## Verification

RED/GREEN (`rust/export/src/model_tests.rs`, new test `a_legacy_type_products_own_attributes_use_its_own_schema_names_not_the_base_types`):
- Before: `IFCDOORSTYLE`'s attribute row was empty — `panicked: IFCDOORSTYLE's own attributes must not be dropped just because IfcType::from_str("IFCDOORSTYLE") is Unknown`.
- After: `OperationType`, `ConstructionType`, `ParameterTakesPrecedence`, `Sizeable` all present, with `Sizeable`'s boolean value (`.T.` → `"true"`) at the correct position; `PredefinedType`/`UserDefinedOperationType` (IfcDoorType-only names) absent.

Mutation: renamed the generated table's `Sizeable` entry to `MUTATED_NOT_SIZEABLE` — test failed with that literal string in the assertion output, proving the generated table (not a stale fallback) is what the code path reads. Reverted, re-confirmed green.

Both directions: the existing `rust/export/src/model_tests.rs` suite (416 tests, including the pre-existing `#3187` `IfcDoorStyle` mesh/row test and ordinary IFC4X3 fixtures like `IfcReinforcingBar`/`IfcBoilerType` attribute tests) all pass unchanged. `IFCWALL` and every other name the generated enum resolves has zero rows in the new table (confirmed by grep), so the fallback path is untouched.

Parity: `rust/export/tests/rooted_type_parity.rs` and `schema_upconvert_parity.rs` — both pass, unaffected (this PR doesn't touch rootedness or upconversion). `scripts/check-legacy-entity-coverage.mjs` — still green (26 arms, 164 concrete legacy products resolvable). New generator's own `--check` mode — green (285 legacy entities, 0 conflicts between the IFC2X3 and IFC4 tables).

Real pass/fail counts (private `CARGO_TARGET_DIR` under `/tmp`, `rust/core` + `rust/export`, native, no wasm involved): 270 + 416 tests, 0 failures. `cargo clippy -p ifc-lite-core -p ifc-lite-export --all-targets -- -D warnings` (matches the CI command) — clean. Nothing BLOCKED — this change never touches `@ifc-lite/wasm` or `@ifc-lite/geometry`'s wasm boundary.

Gates: `node scripts/check-module-size.mjs` — EXIT=0 (my new JS generator is 154 lines; the new generated Rust file lives under a `/generated/` dir, which the Rust ratchet's own logic excludes — confirmed, not assumed, by running `cargo test -p ifc-lite-processing --test module_size_ratchet`, EXIT=0, `no_module_grows_past_its_ratchet_budget` passed with no new allowlist row needed). `node scripts/check-generated.mjs` — every gate that could run in this environment (no `pnpm install`) passed; the rest are documented skips (no `node_modules`), not failures. No `--allow-raise` used anywhere.

Changeset: `.changeset/legacy-entity-own-attribute-names.md`.

`main` is currently red on `check-sdk-canary-coverage` (fixed by open #4493) — expected, unrelated to this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed export of own attribute names for legacy IFC entities, including IFC2X3 and IFC4 entities not present in newer schemas.
  - Legacy entity keywords are now matched case-insensitively, improving compatibility with varied source-file casing.
  - Modern entity exports continue to use their existing attribute definitions.
- **Tests**
  - Added regression coverage for legacy attribute names, lowercase keywords, and unchanged modern-entity behavior.


louistrue (approved):
Reviewed the full diff. render_attributes now reads a legacy (IFC2X3/IFC4, removed-by-IFC4X3) entity's OWN positional attribute names from the generated legacy_attribute_names.rs table before falling back to the modern enum's attribute_names(), so IFCDOORSTYLE and the other 26 legacy names no longer export an empty attribute row — and the base type's names are deliberately not borrowed, since IfcDoorType's list would relabel Sizeable as UserDefinedOperationType. The table is generated by scripts/generate-legacy-attribute-names.mjs from the same EXPRESS-derived entity tables the TypeScript side uses, with a --check mode and conflict reporting; the lookup shares normalise_uppercase with legacy_aware_ifc_type so a lowercase keyword cannot resolve a display type yet miss the attribute lookup. Tests pin the IFCDOORSTYLE row (own names present, base-only names absent, Sizeable at the right slot), the lowercase/uppercase parity, and that modern names are unaffected; the Rust API manifest records the new pub fn. A deliberate partial slice of #4203 (Refs, not Closes); every required check is green on the current head.

Issue #4205, opened 2026-09-09
Relationship graph: index every IfcRelationship subtype with schema-derived slots


louistrue:
Parent: #4201 (The whole schema, first class). Frontier: nothing falls through.

## Problem

`RelationshipType` in `packages/data/src/types.ts:203-233` has 17 values. IFC4X3 declares roughly 46 concrete `IfcRelationship` subtypes. Not indexed at all: IfcRelAssociatesConstraint, IfcRelAssociatesApproval, IfcRelAssociatesLibrary, IfcRelAssignsToActor, IfcRelAssignsToResource, IfcRelAssignsToProcess, IfcRelAssignsToControl, IfcRelDeclares, IfcRelInterferesElements, IfcRelCoversBldgElements, IfcRelCoversSpaces, IfcRelServicesBuildings, IfcRelProjectsElement, IfcRelPositions, IfcRelAdheresToElement, IfcRelFlowControlElements, IfcRelSequence. Two are folded with loss: IfcRelNests onto Aggregates and IfcRelAssignsToGroupByFactor onto AssignsToGroup, dropping `Factor` (`packages/parser/src/columnar-parser-indexes.ts:87,103`). The gate set is hand-written (`columnar-parser-indexes.ts:60-77`, `:130-152`, "This set is the GATE"), and each of #3964, #3237 and #1075 was one relationship or group class that list forgot.

The relating/related attribute slots are also hand-written per class (`apps/server/src/services/data_model/relationships.rs:142-153`, `packages/cli/src/commands/subset-relations.ts`), which #4123 already showed the registry can derive.

## Scope

- Index every `IfcRelationship` subtype of the model's schema as a typed edge (class, relating, related set), with the slot positions derived from the schema registry per version, not from a table.
- Keep the existing 17 `RelationshipType` values as stable views over the generic edges so `packages/cache` and every current consumer keep their answers.
- Surface the new edges in `RelationshipsCard`, `bim.query` and MCP with the exact IfcRel* name, per the schema-fidelity rule in AGENTS.md.
- Server `relationships.rs` derives the same way (parity test against the TS side, as `rooted_type_parity.rs` does).

## Ends when

The drop census (#4208) reports zero `IFCREL*` classes seen-but-unindexed on the bundled corpus, and a fixture carrying one of each relationship class round-trips every edge through `RelationshipsCard` and `bim.query`.

## Out of scope

[...]

Issue #3964, opened 2026-09-05
Server relationships.rs never extracts IfcRelAssignsToGroup, so the Groups panel/By-Zone lens/IDS partOf are silently empty for server-parsed models


BIMvoice:
## The defect

`apps/server/src/services/data_model/relationships.rs`'s `extract_relationships` only extracts nine `IfcRel*` types (lines 19-29):

```rust
let rel_types = [
    "IFCRELCONTAINEDINSPATIALSTRUCTURE", "IFCRELAGGREGATES",
    "IFCRELDEFINESBYPROPERTIES", "IFCRELDEFINESBYTYPE",
    "IFCRELASSOCIATESMATERIAL", "IFCRELASSOCIATESCLASSIFICATION",
    "IFCRELASSOCIATESDOCUMENT",
    "IFCRELVOIDSELEMENT", "IFCRELFILLSELEMENT",
];
```

The TS/WASM path's relationship graph (`packages/parser/src/columnar-parser-indexes.ts:59-105`, `RELATIONSHIP_TYPES` / `REL_TYPE_MAP`, "MUST cover ALL RelationshipType enum values (15 types total)") also resolves:

- `IFCRELASSIGNSTOGROUP`, `IFCRELASSIGNSTOGROUPBYFACTOR` → `RelationshipType.AssignsToGroup`
- `IFCRELASSIGNSTOPRODUCT` → `RelationshipType.AssignsToProduct`
- `IFCRELNESTS` → `RelationshipType.Aggregates`
- `IFCRELCONNECTSPATHELEMENTS`, `IFCRELCONNECTSELEMENTS`, `IFCRELCONNECTSPORTTOELEMENT`, `IFCRELCONNECTSPORTS`
- `IFCRELSPACEBOUNDARY`
- `IFCRELREFERENCEDINSPATIALSTRUCTURE`

None of these ten types appear anywhere in `apps/server/src` (confirmed by grep across the whole crate, including the generated `attr_indices.rs`), so a server-parsed model's `DataModel.relationships` never contains a row for any of them — the type simply never reaches the client.

## Confirmed user-visible consequence: the Groups panel (and "By Zone" lens, and IDS `partOf`)

`IFCRELASSIGNSTOGROUP` has a real, live consumer chain:

[...]

PR #4496, opened 2026-09-10
fix(parser): distinguish IfcRelNests and IfcRelAssignsToGroupByFactor from their broader edge buckets

BIMvoice:
## Summary

Refs #4205. This addresses the two **live folds** the issue names explicitly, not its full scope (a schema-derived edge index over all ~46 concrete `IfcRelationship` subtypes, replacing the hand-written 17-value `RelationshipType` enum and the hand-written relating/related attribute slots in `apps/server`'s `relationships.rs` / `packages/cli`). That migration spans `packages/data`, `packages/parser`, Rust, and CLI, and a half-done version would be worse than none — deliberately left alone here.

### Defect 1: `IfcRelNests` folded into `Aggregates`

Quoted before change, `packages/parser/src/columnar-parser-indexes.ts:67`:
```ts
// IfcRelNests is semantically a decomposition relationship; map it
// onto the same edge bucket so partOf checks for either traverse
// the same graph.
'IFCRELNESTS': RelationshipType.Aggregates,
```
A caller asking "what is aggregated under this object" could not tell a real `IfcRelAggregates` decomposition apart from an `IfcRelNests` one — both landed as the exact same edge type, with no way to recover which STEP keyword produced it.

### Defect 2: `IfcRelAssignsToGroupByFactor` folded into `AssignsToGroup`, dropping `Factor`

Quoted before change, `packages/parser/src/columnar-parser-indexes.ts:83`:
```ts
// Subtype of IfcRelAssignsToGroup (adds a Factor); same RelatingGroup /
// RelatedObjects membership semantics, so it shares the same edge type.
'IFCRELASSIGNSTOGROUPBYFACTOR': RelationshipType.AssignsToGroup,
```
`Factor` was unreachable from the relationship graph — the edge carries no attribute payload, and nothing resolved the relationship entity to read it back.

## Fix

Both STEP classes now **also** land on a second, distinct `RelationshipType` value (`Nests` = 3, `AssignsToGroupByFactor` = 62) in addition to their existing broader-bucket edge (`SECONDARY_REL_TYPE_MAP` in `columnar-parser-indexes.ts`, wired into the parse loop in `columnar-parser.ts`). This keeps every existing consumer of the broader bucket — `spatial-hierarchy-builder.ts`, `decomposition.ts`, `owning-project.ts`, the IDS `partOf`/ancestors bridge, `extractGroupMembersOnDemand` — unchanged, while making the narrower STEP class recoverable for a caller that needs it (`RelationshipsCard`, `bim.query`, drop census).

`extractGroupAssignmentFactorOnDemand(store, groupId, memberId)` (new, `packages/parser/src/group-assignment-factor.ts`) resolves the `Factor` value by looking up the `AssignsToGroupByFactor` edge's `relationshipId` and reading `Factor` off that relationship entity via the existing schema-driven `extractAllEntityAttributes` — no new byte-scanning or graph-column changes needed, since relationship entities are already queryable by express id. Returns `undefined` (not `0`) for a plain assignment or no assignment at all — the fix is careful about absent-vs-empty, per the defect class the issue calls out across #3964/#3237/#1075.

`RelationshipType`'s new members forced 4 separate `Record<RelationshipType, string>` name maps to be updated for exhaustiveness (`relationship-graph.ts`, its test, `export/parquet-exporter.ts`, `cache/sections/relationships.ts` — found by `tsc`, not by grep, so this was a genuine "twins diverge" catch).

## What this deliberately does NOT do

- No wholly-missing relationship class (the 16 the issue lists as unindexed: `IfcRelAssociatesConstraint`, `IfcRelAssignsToActor`, `IfcRelDeclares`, `IfcRelSequence`, …) is added. Each needs its own `extractRelFast` attribute-layout branch, consumer wiring, and tests — not "cheap and safe" within this slice's scope.
- No change to `apps/server`'s `relationships.rs` or `packages/cli/src/commands/subset-relations.ts`'s hand-written attribute slots.
- No change to `IfcRelNests` in `packages/parser/src/schedule-extractor.ts` / `schedule-serializer.ts` — checked open PR #4330's diff first; it touches only `schedule-*` files and viewer schedule components, no overlap with this PR's files.

## Verification

RED/GREEN per defect (`packages/parser/test/relnests-distinct.test.ts`, extensions to `test/group-membership-by-factor.test.ts`): reverted the fix (diff+checkout, not stash) and reran — 3/4 new Nests-distinction tests and 4/10 new Factor tests failed exactly as expected; reapplied and all pass.

Mutation testing (module-size prevented keeping the reverts committed, so evidence is transcript-only, not committed): removing only the `AssignsToGroupByFactor` secondary-map entry failed exactly the 2 Factor-distinction tests and left the 12 Nests/membership tests green; removing only the `Nests` entry failed exactly the 2 Nests tests and left the 12 Factor/membership tests green; renaming the `Factor` attribute lookup key failed exactly the Factor-*value* test (not the edge-distinctness test), proving that test specifically exercises the attribute read, not just the graph edge.

Both directions: `packages/data` 290/290, `packages/parser` 119/121 files (1277/1287 tests) — the 2 failing files (`parser-worker-malformed-handoff.test.ts`, `parser-worker-panic-forward.test.ts`) are genuinely wasm-gated (`parser.worker.ts` statically imports `@ifc-lite/wasm`, no `vi.mock`), reproduces identically on `upstream/main`, unrelated to this change. Also ran (not required by the issue, done because I touched their name maps): `packages/query` 298/298, `packages/ids` 1025/1025, `packages/cache` 189/190 (1 wasm-gated file), `packages/export` 1244/1281 (1 wasm-gated file, 37 skipped) — all match the documented environment baseline, no new failures.

Absent-vs-empty: `extractGroupAssignmentFactorOnDemand` returns `undefined` for both "plain assignment, no Factor attribute" and "no assignment at all" — distinct test cases for each, neither returns `0`.

Gates: `check-module-size.mjs` EXIT=0 (three touched files were pushed over budget by the change — `types.ts`, `parquet-exporter.ts`, `on-demand-extractors.ts` — trimmed comments and split the new function into its own file rather than raising any budget). `check-changesets.mjs` EXIT=0. `check-api-surface.mjs` EXIT=0 after `--update` (single intentional addition: `extractGroupAssignmentFactorOnDemand: function`); the two new enum members are invisible to that tool by design — checked by hand via `git diff`, confirmed additive only, no existing `RelationshipType` value renumbered.

Changeset: `.changeset/relnests-assignstogroupbyfactor-distinct.md` (patch bumps for `@ifc-lite/data`, `@ifc-lite/parser`, `@ifc-lite/export`, `@ifc-lite/cache`).

Not checked: `apps/server`'s Rust side and `apps/viewer` — out of this slice's touched-file set, not run.

## Summary by CodeRabbit

* **New Features**
  * Added distinct relationship tracking for nested elements and group assignments with factors.
  * Added on-demand retrieval of group assignment factors, including valid zero values and no result when unavailable or invalid.
  * Preserved broader relationship and membership lookups for compatibility.

* **Bug Fixes**
  * Corrected relationship names across parsing, export, caching, and query results.
  * Updated exported relationship statistics to count distinct relationship records rather than duplicate graph edges.


louistrue (approved):
Reviewed the full diff. IfcRelNests and IfcRelAssignsToGroupByFactor keep their existing broad-bucket edges (Aggregates / AssignsToGroup, so every current consumer is unchanged) and additionally land on new distinct RelationshipType.Nests / AssignsToGroupByFactor edges recorded by SECONDARY_REL_TYPE_MAP in the relationship-parsing loop; extractGroupAssignmentFactorOnDemand resolves Factor off the relationship entity via the edge's relationshipId and returns undefined (not 0) for a plain or absent assignment while a legitimate 0 reads back as 0. The two consumers that count or label raw edges are corrected (Parquet relationshipCount now counts distinct IfcRel records; the DuckDB rel_type map gains the two names). Changesets are minor on @ifc-lite/data and @ifc-lite/parser for the new enum members and export, patch elsewhere; api-surface.json is updated. Tests cover both directions of each fold, the Factor round trip including zero, and the Parquet count with and without secondary edges. Refs #4205 as a deliberate partial slice; every required check is green on the current head.

Issue #4226, opened 2026-09-09
Reposition all loaded models with precise shared workspace placement


louistrue:
## The problem

Users load an IFC and a pointcloud, in either order, whose coordinates do not align. They need coarse positioning over large distances and precise corrections using snapping, live dimensions, keyboard values, and orthogonal constraints. The same workflow must support IFC–IFC, cloud–cloud, and mixed groups.

## Proposed solution

Implement a shared, non-destructive workspace translation for every loaded model, composed with existing georeference alignment. Keep source coordinates intact; rendering and spatial queries must agree. Manual coordinates use double precision and GPU placement must preserve local precision.

Accepted implementation scope:

- [x] Canonical placement contract and large-offset/small-correction precision evidence.
- [x] Atomic model/group translation, preview/cancel, undo/redo, locks, reset.
- [x] Reposition UI, coarse move-near, exact vector and reference-point coordinates, framing.
- [x] Point-to-point placement with existing mesh/scan snapping, axis/plane constraints, keyboard entry/nudges, live editable dimensions.
- [x] Both load orders, streamed and embedded clouds, instances, overlays, stale-operation cleanup.
- [x] Rendering/picking/bounds/measurements/sections/drawings/analysis agree or explicitly invalidate stale results.
- [x] Durable placement persistence and versioned exchange; explicit original-source versus placed export semantics.
- [ ] Mounted interaction tests, real model/scan demonstration, adversarial review, performance verdict, required CI green.

Rotation about a pivot is a separately scoped follow-up. Automatic registration, scaling, CRS repair, and transformed LAS/E57 writing are out of this translation release. Placement is local workspace state unless shared placement semantics are separately implemented and tested.

Implementation proceeds through coordinate prototype → placement engine → numeric/coarse UI → precise interaction → lifecycle/spatial consumers → persistence/release evidence. Stack reviewable PRs against this issue as needed. Completion is the complete demonstrated workflow above, not a renderer-only drag.

## Alternatives considered

Editing every IfcProduct placement is format-specific and costly. Mutating georeference metadata conflates registration with geographic meaning. Rewriting all vertices on drag loses precision and costs memory bandwidth. A scan-only panel duplicates the interaction and coordinate paths. Automatic registration does not replace exact manual placement.

## Authorization and accountability

[...]

PR #4498, opened 2026-09-10
fix(core): compare the scanned IFCPROJECT keyword case-insensitively

### LINKED ISSUE #4497: Non-uppercase STEP keywords silently produce wrong geometry (1000x unit-scale error, dropped georeferencing)
A STEP file whose keywords are not uppercase silently produces **wrong geometry**, not an error. Confirmed by running, on `upstream/main` @ `18c5aa18e`.

**The headline: a 1000× unit-scale error.**

`rust/core/src/decoder.rs::length_unit_scale()` (and `plane_angle_to_radians()`) scan raw entities and compare `type_name == "IFCPROJECT"`. With a minimal millimetre-unit fixture whose keywords are lowercased (`ifcproject`, `ifcunitassignment`, `ifcsiunit`, …), `length_unit_scale()` returns **`1`** instead of **`0.001`**. That scale feeds curve tessellation in `rust/geometry`, appearance and texture mapping in `rust/processing/src/appearance/*`, and unit conversion in `rust/export` — all silently wrong.

Verified by adding a throwaway test in a scratch worktree (`cargo test -p ifc-lite-core`), discarded afterwards.

**Root cause.** `rust/core/src/parser/scanner.rs::EntityScanner::next_entity()` (~lines 258–267) returns the keyword as a **raw slice from file bytes, unnormalised**. Downstream, consumers disagree:

- **Normalising**: `schema_helpers.rs::{has_geometry_by_name, geometry_flags_by_name, legacy_aware_ifc_type}` via `normalise_uppercase`; `processor/quick_metadata.rs::is_quick_spatial_type_ci` via `eq_ignore_ascii_case`.
- **Not normalising**: raw `==` against uppercase literals.

Where two such paths resolve the same type over the same scan, they disagree. The starkest example is `processor/mod.rs`, where **the same loop** evaluates `type_name == "IFCPROJECT"` (line ~708) and `is_quick_spatial_type_ci(type_name)` (line ~606) on the same value.

**Four instances found:**

1. **`decoder.rs::length_unit_scale()` / `plane_angle_to_radians()`** — the 1000× error above. Its own fallback, `prepass.rs::find_ifcproject_id_inner`, is *also* case-sensitive (`memmem::find(..., b"IFCPROJECT(")`), so both twins fail together. **Run-confirmed.**
2. **`rust/processing/src/georeferencing.rs::georeferencing_candidate_type()`** (~line 144) — a raw `match` on `"IFCMAPCONVERSION" | …`. **Run-confirmed**: `georeferencing_candidate_type("ifcsite")` → `None` while `is_quick_spatial_type_ci("ifcsite")` → `true`. A file with lowercase `ifcsite`/`ifcmapconversion` gets no georeferencing candidates, so `extract_georeferencing()` returns nothing despite valid data.
3. **`processor/mod.rs`** (~lines 708–713) — `project_id`/`site_entity_pos`/`building_entity_pos` compared raw, feeding `site_transform`/`building_transform`, while neighbouring lines in the same loop normalise. Read-confirmed.
4. **`rust/processing/src/appearance/context.rs::Context::new()`** (~lines 21–27) — the cleanest illustration: within one `if`, two OR-arms normalise (`has_geometry_by_name`, `is_representationless_spatial_container_by_name`) and one does not (`name == "IFCSITE"`). Feeds `length_unit_scale` and the RTC offset for the appearance pass. Read-confirmed.

**Ruled out** (checked, consistent): `rust/geometry/src/processors/advanced_face/*` (compares an explicitly `.to_uppercase()`d value); `rust/core/src/units.rs` (operates on already-`IfcType`-decoded entities); `legacy_entities.rs` on `main` (reached only via `legacy_aware_ifc_type`, which normalises); and the TS `byType` map in `packages/parser`, which uppercases at both insertion and lookup.

**On the fix — your call, and I have not touched it.** Normalising once at the scanner would fix all four at the source, but `next_entity()` returns a borrowed slice on a hot path, so making it canonical means either an allocation or pushing every consumer to a case-insensitive compare. Fixing the four sites individually is smaller but leaves the same trap for the next consumer. That trade-off is a design decision rather than a mechanical fix, which is why this is an issue and not a PR.

Happy to implement whichever shape you prefer. If it helps, the single highest-value fix is #1 — the unit scale — since a silent 1000× error is materially worse than the others.

Context: found while fixing a related case-sensitivity defect in PR #4495, where `legacy_attribute_names()` matched case-sensitively while `legacy_aware_ifc_type()` normalised, so a lowercase `ifcproxy` row got its attribute **values labelled with another class's names**.

BIMvoice:
## Summary

`EntityScanner::next_entity()` returns the STEP keyword as a raw, unnormalised slice. STEP keyword case is not significant per ISO 10303-21, so every case-sensitive comparison against that slice is a silent wrong-answer bug. This PR fixes the two instances issue #4497 names as user-visible.

### 1. 1000x unit-scale error (`rust/core`)

`EntityDecoder::length_unit_scale()` and `plane_angle_to_radians()` (`rust/core/src/decoder.rs`) compared the slice against `"IFCPROJECT"` with `==`. A file whose keywords are lowercase or CamelCase never matched, so `project_id` stayed `None` and both resolvers silently defaulted to `1.0`. On a millimetre model this is a silent 1000x scale error feeding curve tessellation, appearance/texture mapping, and unit conversion.

`find_ifcproject_id_inner` (`rust/processing/src/prepass.rs`) — the fallback `resolve_unit_scales` uses when no project-id hint is available — had the same defect one level down: its `memchr::memmem::find` search for the literal byte string `b"IFCPROJECT("` was also case-sensitive. Both are fixed together, since fixing only the primary path would leave the fallback silently preserving the bug.

- `decoder.rs`'s two comparisons now use `type_name.eq_ignore_ascii_case("IFCPROJECT")` rather than `schema_helpers::normalise_uppercase`: the comparison is against one fixed literal, so there's nothing to reuse an allocated canonical form for, and `processor/quick_metadata.rs` already uses `eq_ignore_ascii_case` for exactly this shape of comparison.
- `prepass.rs`'s byte-level search can't use a string helper directly (it operates pre-UTF8-decode). It now prefilters one byte with `memchr::memchr2` (still SIMD-accelerated) and verifies the rest of the keyword with an explicit ASCII case-insensitive byte compare — no new dependency. The prefilter byte is `J`, not the leading `I`: see "Scan cost" below.

### 1b. A string-literal false positive returned a wrong id and stopped the search

Found in review of the round above, and fixed here because the case-insensitive match widens it. `find_ifcproject_id` backtracks from `IFCPROJECT(` to a `#<digits>=` prefix and accepted it wherever it found one — including inside a STEP string literal. On

```
DATA;
#5=IFCWALL('note: #9=ifcproject( in the source',$);
#7=IFCPROJECT('g',$);
```

it returned `Some(9)` **and returned early**, so the real `#7` was never reached. `extract_length_unit_scale` then rejects `#9` on its own `IFCPROJECT` type guard and `resolve_unit_scales`'s `unwrap_or(1.0)` fires — the exact silent 1000x default this PR exists to remove.

The hazard is pre-existing in kind (the code's own comment named the in-string case), but case-insensitivity widens it: before, only an *uppercase* `#N=IFCPROJECT(` inside a literal could fool the scan; now ordinary lowercase or CamelCase prose in a description or comment does.

The fix requires the `#` to actually start a record — preceded, across trivia, by `;` or the start of input, which ISO 10303-21 guarantees for every declaration. A `/* … */` comment is legal wherever whitespace is, so the backward walk skips one, mirroring `skip_step_trivia`'s forward allowance in the scanner; a real declaration behind a comment is not refused.

I chose this over the stricter alternative (validate that the found id decodes to an `IFCPROJECT`, and keep scanning if not). That one is more robust — it would also reject a decoy written as `'…;#9=ifcproject('` — but it costs an entity index per candidate, which is the O(file)-scan-per-decoder class `resolve_unit_scales`'s own doc comment exists to keep dead. This keeps the function a single linear byte pass.

### 1c. Scan cost: the case-insensitive match made the scan ~14x slower; fixed

`memchr2('I', 'i')` is the worst possible prefilter for this keyword: every IFC keyword starts with `I` and IFC GUIDs are dense in `I`/`i`, so a project-less file hits on nearly every record and the verify loop dominates. The worst case is reachable on exactly the files this PR targets — a lowercase file yields no project-id hint from `processor/mod.rs`, so this fallback runs, and IfcOpenShell emits `IFCPROJECT` near the END of the file.

Measured in release on a 20 MB synthetic file with no `IFCPROJECT` (7 runs, best-worst):

| scan | time |
|---|---|
| `memmem` (case-SENSITIVE — what this PR replaced) | 0.75-0.98 ms |
| `memchr2('I','i')` (previous head of this PR) | 7.56-13.08 ms |
| `memchr2('J','j')` + candidate at `j - 6` (this head) | 0.33-0.36 ms |

`J` occurs in almost nothing else in an IFC file. It sits at index 6 of `IFCPROJECT(`, so a hit at `j` means a candidate start of `j - 6`; a `J` in the first 6 bytes, or one whose candidate falls behind the caller's resume point, cannot open a match and is skipped.

### 2. Dropped georeferencing (`rust/processing`)

`georeferencing_candidate_type` (`rust/processing/src/georeferencing.rs`) classified candidates with a case-sensitive `match` on `"IFCMAPCONVERSION" | "IFCMAPCONVERSIONSCALED" | "IFCPROJECTEDCRS" | "IFCPROPERTYSET" | "IFCSITE"`. A lowercase- or CamelCase-keyword file therefore produced an **empty** candidate list and the model reported **no georeferencing at all**, despite carrying a complete `IfcMapConversion`. It gates both `extract_georeferencing_with_index` and the native geometry scan in `processor/mod.rs`, so both paths were affected.

The divergence was observable inside a single loop: in `processor/mod.rs`'s scan, `georeferencing_candidate_type("ifcsite")` returned `None` while its sibling `is_quick_spatial_type_ci("ifcsite")` returned `true` on the same slice. A drift guard test now pins those two to agree on every spelling of `IFCSITE`.

Fixed with `eq_ignore_ascii_case` against the fixed literals — no allocation on a per-entity hot path.

## Verification

**Unit scale (`rust/core` + `prepass.rs`), unchanged from the earlier review round:**
- Repro before the fix: a lowercase-keyword millimetre fixture made `length_unit_scale()` return `1.0` instead of `0.001`. RED → GREEN on the same fixture.
- Uppercase control unchanged; mixed case (`IfcProject`/`IfcUnitAssignment`/`IfcSiUnit`) resolves correctly.
- `find_ifcproject_id` tested directly with lowercase/mixed-case/whitespace/CRS-collision fixtures, plus an end-to-end `resolve_unit_scales(..., None, ...)` test that forces the no-hint fallback.
- `plane_angle_to_radians` covered separately with a lowercase DEGREE fixture.
- Each of the three sites individually mutated back to `==` with an `eprintln!` marker proving the mutated line ran; each reddened exactly its own tests.

**Record-start guard (1b), new in this round:** RED first — `find_ifcproject_id_skips_a_full_record_decoy_inside_a_string` returned `Some(9)` instead of `Some(7)` on the input above. GREEN after. Mutated the guard back out, replacing `starts_a_record(content, i - 1)` with `{ eprintln!("MUTATION PROBE: record-start guard bypassed"); true }`: the probe printed and the test reddened with `left: Some(9) right: Some(7)`. Two further tests pin the other direction, which is where such a guard goes wrong — a declaration behind a `/* … */` comment, and one at byte 0 with no preceding `;` — since a false negative here is the same silent `1.0` default as a miss.

**Scan equivalence (1c), new in this round:** `find_ifcproject_keyword_matches_a_naive_reference_at_every_offset` pins the `J` prefilter against a naive `eq_ignore_ascii_case` reference at **every** offset of 414 buffers — 14 hand-built adversarial cases (`"JIFCPROJECT("`, `"XXXXXXIFCPROJECT("`, `"IFCPROJECTEDCRS("`, `"ifcprojectIFCPROJECT("`, a bare `"J"`, …) plus 400 pseudo-random buffers over an alphabet dense in the keyword's own bytes — with an anti-vacuity assert on the pair count. Mutation: `J_OFFSET` 6 -> 5 reddens it (and 13 other prepass tests); dropping the `>= from` filter reddens it at `from=1 in "IFCPROJECT("`; an `eprintln!` probe inside the mutated expression proved the line ran. The same differential harness, run standalone against the previous `memchr2('I','i')` implementation as a third oracle, agreed on all 120,565 (buffer, offset) pairs.

**Georeferencing (new in this round):** four tests in `rust/processing/src/georeferencing_tests.rs`.

RED first, before the fix (`cargo test -p ifc-lite-processing --lib georeferencing`): `20 passed; 4 failed` —
```
candidate_classification_ignores_keyword_case
  left: None  right: Some(IfcMapConversion)  (candidate classification of ifcmapconversion)
site_candidacy_agrees_with_the_quick_spatial_gate_on_every_spelling
  left: None  right: Some(IfcSite)           (ifcsite must be a georeferencing candidate)
extraction_survives_lowercase_keywords        panicked: "lowercase georef"
site_fallback_survives_lowercase_keywords     panicked: "lowercase site georef"
```

GREEN after the fix: `cargo test -p ifc-lite-processing --lib` → **300 passed; 0 failed** (296 before, + the 4 new). With the four tests added for 1b/1c this head is at **304 passed; 0 failed**.

Mutation-tested per comparison, each with an `eprintln!` marker inside the mutated expression so a no-op mutation could not read as green:

| mutation (back to `==`) | marker prints | tests reddened |
|---|---|---|
| `IFCSITE` | 7 | `candidate_classification…`, `site_candidacy_agrees…`, `site_fallback_survives…` |
| `IFCMAPCONVERSION` / `…SCALED` | 33 | `candidate_classification…`, `extraction_survives…` |
| `IFCPROJECTEDCRS` | 15 | `candidate_classification…`, `extraction_survives…` |
| `IFCPROPERTYSET` | 1 | `candidate_classification…` |

The end-to-end lowercase fixtures rewrite only the `#nn=IFCxxx` keyword token, leaving string literals and enumeration values (`.ELEMENT.`, `.METRE.`) alone — a blanket `to_lowercase()` would have changed the data the assertions read back, which the assertion on the rewritten text guards.

**Gates run locally on this head:** `cargo clippy -p ifc-lite-processing --all-targets -- -D warnings` clean; `cargo test -p ifc-lite-processing` **496 passed; 0 failed**; `cargo test -p ifc-lite-core` **314 passed; 0 failed**; `cargo test -p ifc-lite-processing --test module_size_ratchet` 6/6 (`prepass.rs` 729 lines against its 781 budget); `node scripts/check-module-size.mjs` exit 0 (`2716 files measured, 329 allowlisted, 0 new over 400`). Changeset updated. No wasm-dependent JS test runs in this environment, so none are claimed.

## CI on this head (`c44373a30`)

All four required checks green: `Build + WASM + Rust + Node`, `Issue queue`, `PR review signal`, `parity (in-tree fixtures, committed reference)` — run `34590567856` (20 jobs) plus `34590447019`. `Benchmark` and `Content-matching fixture` show `cancelled`, which is supersession, not failure.

`Viewer E2E smoke` failed on the first attempt with the same `model-reposition.e2e.spec.ts:16` `page.waitForFunction: Timeout 120000ms exceeded` recorded below, in the same test (`reposition IFC and diagnostic scan, IFC first (#4226)`). It also failed that way on the previous head `5611fccca` — which already carried current `main`, and `main` is green — so I re-ran that job alone on this head rather than assert a cause: it **passed**, and the run is now `completed success`. Flaky, not caused here.

I did not take that on trust. The record-start guard is the one change on this branch that could plausibly refuse a real project and default the unit scale, so I swept every `.ifc` in the tree (**367 files, 177 of which resolve a project id**) comparing `find_ifcproject_id` with and without the guard: **0 files answer differently**. The e2e model (`tests/models/ara3d/AC20-FZK-Haus.ifc`) is downloaded in CI and so is not in that corpus; its keywords are uppercase, on which this branch's comparisons are byte-identical to the old ones.

## CI: the red checks on an earlier head were inherited, not caused here

The previous run (`34510706095`, head `4722551a`, created 2026-09-10T17:50Z) was based on `18c5aa18e`, and `main` itself was red at that commit. Per-check:

| Check | Verdict | Evidence |
|---|---|---|
| `Rust tests` | **stale** | `error: function 'oracle_fixture_present' is never used … rust/geometry/tests/triangulation_invariance.rs:2183` under `-D warnings`. This branch touches no file in `rust/geometry`. Fixed on `main` by #4499 (merged 2026-09-10T23:19Z), which added `#[allow(dead_code)]` there. |
| `CSG accept gates (feature builds)` | **stale** | Same error, same file, same line. |
| `Node tests` | **stale** | `check-sdk-canary-coverage`: `sdk-canary.yml` `paths:` covered 30 of 32 transitively-built packages, missing `packages/regex-guard` and `packages/wasm-lifecycle`. This branch touches neither, nor the workflow. Fixed on `main`; `.github/workflows/sdk-canary.yml` on `main` now lists both. |
| `Build + WASM + Rust + Node` | **stale** | 2s aggregator job; it needs `Rust tests` and `Node tests`, so it reported their failures rather than one of its own. |
| `Viewer E2E smoke` | **flaky; re-ran green on this head** | `TimeoutError: page.waitForFunction: Timeout 120000ms exceeded` in `e2e/model-reposition.e2e.ts` ("agnostic scan IFC first", #4226). The fixtures it loads use uppercase keywords, on which this branch's comparisons are byte-identical to the old ones, so there is no mechanism by which this change could alter that test. I could not attribute it to a specific `main` commit, so I am reporting it as unattributed rather than asserting it is stale — the re-run on the merged head is the check. |

This head merges current `upstream/main` (no rebase, so the maintainer's commits on the branch are preserved and nothing is force-pushed). `git diff upstream/main HEAD` touches only the 7 files this PR owns, so the merge reverted nothing from `main`.

## What a user actually gets after this PR

**A user loading a lowercase-keyword file gets correct units and still-wrong geometry.** This PR fixes the unit scale. It does not fix the scan arms that consume the same raw keyword, so openings are still never subtracted.

Measured against `rust/processing/tests/fixtures/issue_2019_wall_two_overlapping_openings.ifc` through `process_geometry`, with only the `#nn=IFCxxx` keyword tokens lowercased (string literals and enumerations left alone):

| input | meshes | triangles |
|---|---|---|
| as written (uppercase) | 4 | **230** |
| same file, lowercase keywords | 4 | **48** |

48 is four plain boxes: the wall renders solid, with all three `IfcRelVoidsElement` openings ignored, because `processor/mod.rs`'s scan arms are still case-sensitive — `type_name == "IFCRELVOIDSELEMENT"` at :698 among the block at :656-712.

This is worth stating plainly because this PR removes the 1000x symptom — the loud, obvious one that would have led a user to report the rest.

## The rest of the class, and the fix worth making

This bug class exists because the scanner hands out a raw slice; fixing call sites one at a time resets the clock rather than closing the hole. I swept `rust/` for every case-sensitive comparison against a value bound from `next_entity()` (script: walk each `next_entity()` loop body, flag `== "…"` / `!= "…"` / `match <var>` / `<var>.starts_with(` / `matches!(<var>` on the keyword binding). **At least 53 sites remain** after this PR.

**That number is a floor, not a total.** The heuristic structurally cannot see `ends_with`, nor any comparison that is not inside a `next_entity()` loop body. Review found four more that it missed, each verified here — see "Missed by the sweep" below. Treat 53 as "at least 53, from a heuristic that misses `ends_with` and any comparison outside a `next_entity()` loop".

Sites the sweep did find:

- `rust/processing/src/processor/mod.rs`: 656, 662, 668, 671, 674, 685, 689, 694 (`starts_with("IFCPROPERTY")`), 698, 700, 702, 708, 710, 712, 786, 805
- `rust/processing/src/appearance/context.rs`: 21, 24, 27 — and `appearance/page_source.rs`:14, `appearance/source.rs`:61
- `rust/processing/src/determinism.rs`:259, `prepass_type_material.rs`:89, `symbolic/color.rs`:28 & 43, `symbolic/mod.rs`:173
- `rust/core/src/model_bounds.rs`: 131, 208, 216, 222, 231; `rust/core/src/decoder/styled_items.rs`:41
- `rust/geometry/src/void_index.rs`: 94, 267; `material_layer_index.rs`:150; `processors/texture/mod.rs`:324; `profile_extractor.rs`:750; `router/mod.rs`:285
- `rust/wasm-bindings/src/api/styling/prepass.rs`: 49, 169, 252, 325; `gpu_meshes/prepass.rs`:335; `alignment_lines.rs`: 62, 80; `grid_lines.rs`:67; `api/mod.rs`:937
- `rust/export/src/model.rs`:172, `relationships.rs`:59, `merged/plan.rs`:67, `constructions.rs`:53, `dfjson/spatial.rs`: 90, 91

Plus three byte-level prefilters over the whole file, which fail the same way and silently skip a pass entirely: `rust/wasm-bindings/src/api/mod.rs`:926 (`b"IFCINDEXEDCOLOURMAP"`), `rust/wasm-bindings/src/api/gpu_meshes/prepass.rs`:746 (`b"IFCREPRESENTATIONMAP"`), `rust/geometry/src/processors/texture/mod.rs`:318 (`b"IFCINDEXEDTRIANGLETEXTUREMAP"`). And one transitive site: `merged/plan.rs` stores the raw keyword into `type_of`, which `merged/empty.rs`:286 and :296 then test with `starts_with("IFCREL")`.

### Missed by the sweep

Four sites the heuristic could not reach, each reproduced:

- **`rust/processing/src/shard_classes.rs`:153, `classify_type_name`** — a `match` on the keyword **outside** any `next_entity()` loop. Reproduced: `classify_type_name("IFCPROJECT") == 2`, `("IfcProject") == 0`, `("ifcproject") == 0`; `("IFCRELVOIDSELEMENT") == 8`, `("ifcrelvoidselement") == 0`. This is the **sharded browser prepass** (`scan_shard_classified`, consumed by `wasm-bindings/src/api/gpu_meshes/prepass_sharded.rs`), so on a lowercase file every named class collapses to `PREPASS_CLASS_NONE` and no project, site record, styled item, void rel, aggregate or layer set is discovered. Geometry survives because the flag bits go through `has_geometry_by_name`, which normalises — `classify_type_name("IfcSite")` returns 128, the geometry flag alone, against 3 for `"IFCSITE"`. Colours and openings do not survive.
- **`rust/core/src/schema_helpers.rs`:327, `type_product_ifc_type`** — short-circuits on `!type_name.ends_with("TYPE") && !type_name.ends_with("STYLE")` **before** the normalising `legacy_aware_ifc_type` call it then makes. Reproduced: `("IFCWALLTYPE") == Some(IfcWallType)`, `("IfcWallType") == None`, `("ifcwalltype") == None`. It sits beside the helpers listed under "Not affected, checked" below, which made the whole family look safe; it is not.
- **Two more whole-file byte prefilters** of the same shape as the three listed above: `rust/wasm-bindings/src/api/mod.rs`:758 (`memmem::find(content, b"IFCMATERIALLAYERSET")` — a miss caches an EMPTY `MaterialLayerIndex`, so layered walls are never sliced) and `rust/wasm-bindings/src/api/styling/prepass.rs`:154-155 (`windows(…).any(|w| w == b"IFCREPRESENTATIONMAP")` — a miss returns `Vec::new()` and skips the type-geometry pass entirely).

Not affected, checked: `IfcType::from_str`, `has_geometry_by_name`, `geometry_flags_by_name`, `is_representationless_spatial_container_by_name` and `legacy_aware_ifc_type` all normalise first, so call sites going through them are already case-correct. Comparisons against `IfcType::as_str()` or a decoded `profile.ifc_type` (e.g. `rust/geometry/src/processors/advanced_face/*`, which uppercases into a local first) are comparing canonical forms and are safe. `rust/geometry/src/router/content_hash.rs`:142 (`type_token_is(b, b"IFCFACETEDBREP")`) looks like a fourth byte prefilter but is **not** affected: `type_token_is` compares with `eq_ignore_ascii_case`. I did not audit `packages/` (TypeScript) for the same class, and I did not re-run the sweep script with `ends_with` added — the four above came from review, not from a second sweep, so there may be more.

**I deliberately did not hand-patch `processor/mod.rs` and `appearance/context.rs` here**, even though the edits are mechanical and provably no-op for uppercase input. They are 19 of 53 sites: patching them leaves 34 behind while inflating this PR's diff, and the next raw comparison someone writes reopens the hole.

**Proposal for the maintainer** (not undertaken in this PR): make the divergence impossible at the source. `next_entity()` returns `(u32, &str, usize, usize)` on a hot path where allocating an uppercase `String` per entity is not acceptable, so uppercasing in the scanner is out. The viable shape is a newtype — `StepKeyword<'a>(&'a str)` — that borrows exactly as today, exposes `eq_ignore_ascii_case`-backed `PartialEq<&str>` / an `is()` method / a `match`-friendly `as_canonical()`, and deliberately does **not** implement `Deref<Target = str>` or `PartialEq<str>`, so `type_name == "IFCSITE"` stops compiling and a `match` on it is a type error. That turns the sites that receive the keyword through `next_entity()`'s return type into compiler errors to be worked through once, with zero runtime cost and no way to regress.

**It would not close the whole class, and you should weigh it knowing that.** Neither `shard_classes.rs::classify_type_name` nor `schema_helpers.rs::type_product_ifc_type` takes the scanner slice through `next_entity()` — both take a plain `&str` from a caller — so both would keep compiling unchanged, and the whole-file byte prefilters (five of them now) are pre-decode and out of reach of any `&str` newtype. Those need either their own `&[u8]` helper or their signatures changed to take the newtype too. It is a wide mechanical change across five crates and touches the public signature of `EntityScanner::next_entity`, so it wants its own issue and your call on the API shape before anyone starts. Happy to take it if you want it.

Closes #4497

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected handling of STEP files that use lowercase or mixed-case entity keywords.
  - Fixed length and angle unit detection, preventing incorrect scaling—such as millimetre models being interpreted with a 1000× error.
  - Restored georeferencing detection for files using varied capitalization, including project, site, coordinate reference, and property entities.

### DIFF
--- a/.changeset/case-insensitive-ifcproject-unit-scale.md
+++ b/.changeset/case-insensitive-ifcproject-unit-scale.md
@@ -6,4 +6,8 @@ Fix `EntityDecoder::length_unit_scale`/`plane_angle_to_radians` (`rust/core/src/
 
 `georeferencing_candidate_type` (`rust/processing/src/georeferencing.rs`) had the same defect: it classified `IfcMapConversion`, `IfcProjectedCRS`, `IfcPropertySet` and `IfcSite` candidates with a case-sensitive `match`, so a lowercase- or CamelCase-keyword file reported no georeferencing at all despite carrying complete `IfcMapConversion` data. It is the gate for both the standalone extractor and the native geometry scan, so both paths were affected.
 
+`find_ifcproject_id` also stopped on the first `#<id>=IFCPROJECT(` it could backtrack to, including one written inside a STEP string literal, and returned that id — which then fails the resolver's own `IFCPROJECT` type check and restores the same silent 1.0 default. It now requires the `#` to actually start a record (preceded, across trivia, by `;` or the start of input). Matching the keyword case-insensitively widened this pre-existing hazard from uppercase decoys to ordinary lowercase prose in a description or comment, so it is fixed here.
+
+The same scan now prefilters on the keyword's `J` rather than its leading `I`: every IFC keyword starts with `I` and GUIDs are dense in `I`/`i`, so the lead-byte scan hit on nearly every record. On a 20 MB project-less file this measured 7.6-13.1 ms against 0.33-0.36 ms for the `J` prefilter (and 0.8-1.0 ms for the case-sensitive `memmem` it replaced).
+
 This is one part of a broader case-sensitivity class tracked in issue #4497; the remaining sites are enumerated in the PR.
--- a/.changeset/case-insensitive-ifcproject-unit-scale.md
+++ b/.changeset/case-insensitive-ifcproject-unit-scale.md
@@ -2,4 +2,8 @@
 "@ifc-lite/wasm": patch
 ---
 
-Fix `EntityDecoder::length_unit_scale`/`plane_angle_to_radians` (`rust/core/src/decoder.rs`) and their `find_ifcproject_id_inner` fallback (`rust/processing/src/prepass.rs`) comparing the scanned `IFCPROJECT` keyword case-sensitively. A file whose keywords are lowercase or CamelCase (STEP keyword case is not significant) silently defaulted the length-unit scale to `1.0` instead of the declared value — a 1000x error on a millimetre model feeding curve tessellation, appearance/texture mapping, and unit conversion. This is one instance of a broader case-sensitivity class tracked in issue #4497; the other three sites listed there are unchanged.
+Fix `EntityDecoder::length_unit_scale`/`plane_angle_to_radians` (`rust/core/src/decoder.rs`) and their `find_ifcproject_id_inner` fallback (`rust/processing/src/prepass.rs`) comparing the scanned `IFCPROJECT` keyword case-sensitively. A file whose keywords are lowercase or CamelCase (STEP keyword case is not significant) silently defaulted the length-unit scale to `1.0` instead of the declared value — a 1000x error on a millimetre model feeding curve tessellation, appearance/texture mapping, and unit conversion.
+
+`georeferencing_candidate_type` (`rust/processing/src/georeferencing.rs`) had the same defect: it classified `IfcMapConversion`, `IfcProjectedCRS`, `IfcPropertySet` and `IfcSite` candidates with a case-sensitive `match`, so a lowercase- or CamelCase-keyword file reported no georeferencing at all despite carrying complete `IfcMapConversion` data. It is the gate for both the standalone extractor and the native geometry scan, so both paths were affected.
+
+This is one part of a broader case-sensitivity class tracked in issue #4497; the remaining sites are enumerated in the PR.
--- a/.changeset/case-insensitive-ifcproject-unit-scale.md
+++ b/.changeset/case-insensitive-ifcproject-unit-scale.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/wasm": patch
+---
+
+Fix `EntityDecoder::length_unit_scale`/`plane_angle_to_radians` (`rust/core/src/decoder.rs`) and their `find_ifcproject_id_inner` fallback (`rust/processing/src/prepass.rs`) comparing the scanned `IFCPROJECT` keyword case-sensitively. A file whose keywords are lowercase or CamelCase (STEP keyword case is not significant) silently defaulted the length-unit scale to `1.0` instead of the declared value — a 1000x error on a millimetre model feeding curve tessellation, appearance/texture mapping, and unit conversion. This is one instance of a broader case-sensitivity class tracked in issue #4497; the other three sites listed there are unchanged.
--- a/rust/core/src/decoder.rs
+++ b/rust/core/src/decoder.rs
@@ -308,7 +308,14 @@ impl<'a> EntityDecoder<'a> {
         let mut scanner = crate::parser::EntityScanner::new(self.content);
         let mut project_id: Option<u32> = None;
         while let Some((id, type_name, _, _)) = scanner.next_entity() {
-            if type_name == "IFCPROJECT" {
+            // #4497: STEP keyword case is not significant (ISO 10303-21);
+            // `type_name` is the raw, unnormalised scanner slice, so compare
+            // ASCII-case-insensitively rather than requiring upstream
+            // uppercase. Zero-alloc: `eq_ignore_ascii_case` against a fixed
+            // literal, unlike `schema_helpers::normalise_uppercase` which
+            // allocates a canonical owned copy — not needed here since there
+            // is nothing downstream to reuse the normalised form for.
+            if type_name.eq_ignore_ascii_case("IFCPROJECT") {
                 project_id = Some(id);
                 break;
             }
@@ -338,7 +345,9 @@ impl<'a> EntityDecoder<'a> {
         let mut scanner = crate::parser::EntityScanner::new(self.content);
         let mut project_id: Option<u32> = None;
         while let Some((id, type_name, _, _)) = scanner.next_entity() {
-            if type_name == "IFCPROJECT" {
+            // #4497: see the matching comment in `plane_angle_to_radians`
+            // above — same case-insensitivity fix, same rationale.
+            if type_name.eq_ignore_ascii_case("IFCPROJECT") {
                 project_id = Some(id);
                 break;
             }
--- a/rust/core/src/decoder/decoder_tests.rs
+++ b/rust/core/src/decoder/decoder_tests.rs
@@ -502,3 +502,108 @@ fn get_face_bound_fast_refuses_oversized_loop_ref() {
     let mut decoder = EntityDecoder::new(content);
     assert_eq!(decoder.get_face_bound_fast(1), Some((u32::MAX, true, false)));
 }
+
+/// Issue #4497: `EntityScanner::next_entity` returns the STEP keyword as a
+/// raw, unnormalised slice, and `length_unit_scale`/`plane_angle_to_radians`
+/// compared it case-sensitively against `"IFCPROJECT"`. A file whose
+/// keywords are lowercase (a legal STEP file — case is not significant)
+/// silently defaulted to scale `1.0` instead of the declared `0.001`, a
+/// 1000x error feeding curve tessellation, appearance mapping and unit
+/// conversion.
+#[test]
+fn length_unit_scale_lowercase_keywords_repro() {
+    let content = r#"
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('t.ifc','2026-01-01T00:00:00',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=ifcproject('0001projectaaaaaaaaaaa',$,'P',$,$,$,$,$,#2);
+#2=ifcunitassignment((#3));
+#3=ifcsiunit(*,.LENGTHUNIT.,.MILLI.,.METRE.);
+ENDSEC;
+END-ISO-10303-21;
+"#;
+    let mut decoder = EntityDecoder::new(content);
+    assert_eq!(decoder.length_unit_scale(), 0.001);
+}
+
+/// Control: uppercase keywords (the pre-existing, always-worked case) must
+/// keep resolving exactly as before the #4497 fix.
+#[test]
+fn length_unit_scale_uppercase_keywords_unchanged() {
+    let content = r#"
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('t.ifc','2026-01-01T00:00:00',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=IFCPROJECT('0001projectaaaaaaaaaaa',$,'P',$,$,$,$,$,#2);
+#2=IFCUNITASSIGNMENT((#3));
+#3=IFCSIUNIT(*,.LENGTHUNIT.,.MILLI.,.METRE.);
+ENDSEC;
+END-ISO-10303-21;
+"#;
+    let mut decoder = EntityDecoder::new(content);
+    assert_eq!(decoder.length_unit_scale(), 0.001);
+}
+
+/// The realistic case per #4497: some exporters emit CamelCase keywords
+/// (`IfcProject`, not all-lower or all-upper).
+#[test]
+fn length_unit_scale_mixed_case_keywords() {
+    let content = r#"
+ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION((''),'2;1');
+FILE_NAME('t.ifc','2026-01-01T00:00:00',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=IfcProject('0001projectaaaaaaaaaaa',$,'P',$,$,$,$,$,#2);
+#2=IfcUnitAssignment((#3));
+#3=IfcSiUnit(*,.LENGTHUNIT.,.MILLI.,.METRE.);
+ENDSEC;
+END-ISO-10303-21;
+"#;
+    let mut decoder = EntityDecoder::new(content);
+    assert_eq!(decoder.length_unit_scale(), 0.001);
+}
+
+/// `plane_angle_to_radians` has the same scanner-comparison bug — cover it
+/// separately rather than assuming the `length_unit_scale` fix carries over.
+/// DEGREE fixture per `units_tests::test_extract_plane_angle_degree`, with
+/// all keywords lowercased.
+#[test]
+fn plane_angle_to_radians_lowercase_keywords() {
+    let content = r#"ISO-10303-21;
+HEADER;
+FILE_DESCRIPTION(('Test'),'2;1');
+FILE_NAME('test.ifc','2024-01-01',(''),(''),'','','');
+FILE_SCHEMA(('IFC4'));
+ENDSEC;
+DATA;
+#1=ifcproject('guid',$,'Test',$,$,$,$,(#2),#3);
+#2=ifcgeometricrepresentationcontext($,'Model',3,1.E-5,#4,$);
+#3=ifcunitassignment((#5,#10));
+#4=ifcaxis2placement3d(#7,$,$);
+#5=ifcsiunit(*,.LENGTHUNIT.,$,.METRE.);
+#7=ifccartesianpoint((0.,0.,0.));
+#8=ifcsiunit(*,.PLANEANGLEUNIT.,$,.RADIAN.);
+#9=ifcmeasurewithunit(ifcratiomeasure(0.0174532925199433),#8);
+#10=ifcconversionbasedunit(#11,.PLANEANGLEUNIT.,'DEGREE',#9);
+#11=ifcdimensionalexponents(0,0,0,0,0,0,0);
+ENDSEC;
+END-ISO-10303-21;
+"#;
+    let mut decoder = EntityDecoder::new(content);
+    let scale = decoder.plane_angle_to_radians();
+    assert!(
+        (scale - 0.0174532925199433).abs() < 1e-9,
+        "expected 0.01745… for DEGREE, got {scale}"
+    );
+}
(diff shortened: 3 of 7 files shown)


louistrue (approved):
Reviewed the full diff. The two EntityDecoder resolvers (length_unit_scale, plane_angle_to_radians) now compare the scanned IFCPROJECT keyword with eq_ignore_ascii_case, georeferencing_candidate_type classifies its four candidates case-insensitively, and find_ifcproject_id_inner's byte-level fallback is case-insensitive via a J-prefiltered memchr2 scan (with the measured rationale for the prefilter byte) plus a starts_a_record guard that rejects a record-shaped decoy inside a string literal while still accepting a declaration behind a STEP comment. Tests cover lowercase, CamelCase and uppercase keywords on every path, the decoy and comment cases, the u32 edge, and a naive-reference oracle over adversarial and pseudo-random buffers for the keyword finder. Changeset is a wasm patch; parity and the rest of the required checks are green on the current head. Closes #4497.

Issue #4311, opened 2026-09-09
Surface ambiguous storey containment (elementToStorey / containedIn()) as a detectable signal


BIMvoice:
Follow-up to #4248 (PR #4310). Maintainer ruling on #4248 confirmed this is in scope but standalone from the tie-break fix: "An element contained under more than one storey should be detectable by a caller, not silently resolved."

Neither `elementToStorey` (packages/parser's `SpatialHierarchyBuilder`) nor `containedIn()` (packages/query's `EntityNode`) currently exposes any indication that an element had more than one direct `IfcRelContainedInSpatialStructure` containment edge. Both now resolve the tie the same way (first-declared wins, #4310), but a caller has no way to know the containment was ambiguous in the source file versus unambiguous.

Not implemented in #4310 because it's more than a small addition on top of the tie-break fix: it needs either a new field on `SpatialHierarchy` (e.g. an `ambiguousStorey: Set<number>` or similar) or a diagnostics/warning-log convention, threaded through both `elementToStorey`'s and `containedIn()`'s call sites, plus its own tests. Deliberately deferred rather than growing that PR's scope.

Scope for whoever picks this up: decide the shape of the signal (a set/count on `SpatialHierarchy`, a warning log like #4246's back-edge warning, or something else), and whether packages/query's `containedIn()` needs an equivalent (it currently has no aggregate-level view at all, only per-call).

Issue #4206, opened 2026-09-09
Structural analysis end to end, as the template for every later domain


louistrue:
Parent: #4201 (The whole schema, first class). Demonstration 2, one concept end to end.

## Problem

IfcTask is the proof that a non-product concept can be first class here: extractor (`packages/parser/src/schedule-extractor.ts`), `bim.schedule` namespace, Gantt panel, `ScheduleCard`, serializer. No second non-product domain has followed. Structural analysis (IfcStructuralAnalysisModel, IfcStructuralMember, IfcStructuralAction, IfcStructuralLoad*, IfcBoundaryCondition) and cost (IfcCostItem, IfcCostSchedule, IfcCostValue) are scan-only: no extractor, no namespace, no card, no writer. `packages/cli/src/commands/stats-aggregation.ts:256-258` excludes structural explicitly. The corpus already holds `tests/models/ifcopenshell/structural_analysis_curve.ifc`; it holds no file with a cost schedule.

## Scope

Take **structural analysis** through the same six layers as schedule, as the template every later domain copies:

1. parse: retained rows (depends on #4204) and typed attributes (#4202)
2. semantic: an extractor producing analysis models, members, connections, actions and loads with their relating elements (`IfcRelConnectsStructuralMember`, `IfcRelConnectsStructuralActivity`, `IfcRelAssignsToGroup` into the analysis model)
3. query: a `bim.structural` namespace mirrored in the sandbox bridge schema and MCP
4. geometry: structural curve and surface members already have representations; render them through the existing symbolic or line channel with a class toggle, as alignments are
5. UI: a properties card on a member listing its analysis model, connections and applied loads
6. write: round-trip the extracted data through the STEP passthrough with a fixture diff proving nothing moved

Cost (5D) is the next domain, gated on obtaining a redistributable fixture; file it as its own subissue when one exists.

## Ends when

`structural_analysis_curve.ifc` and one real-tool structural export (to be sourced, buildingSMART sample sets first) show every analysis entity through CLI, MCP, SDK and viewer, and the loss report on export and IFC2X3 conversion is empty or names each attribute it could not carry.

## Out of scope

Solving, load combination or any engineering interpretation; this is the model, not the analysis.

---

**Originating vision:** [The whole schema, first class](https://github.com/LTplus-AG/ifc-lite/discussions/5038) (migrated from #4201). This issue remains independently planned and tracked.

BIMvoice:
[...]

Issue #4118, opened 2026-09-07
Site-local models discard all instancing metadata, so repeated geometry never collates


louistrue:
### Version

`main` at 271b83caa.

### Which surface

Geometry / CSG kernel

### Reproduction

Split out of #4103. That issue's attachment holds 10 `IfcFurnishingElement`
sharing one `IfcRepresentationMap`. #4103 fixes the reason their mesh buffers
differed, and after that fix the ten occurrences are byte-identical and
`collate_refs` groups them into 3 templates covering all 30 meshes, a 90%
reduction in the vertices that have to ship.

On the reporter's file it still groups nothing, because every mesh arrives with
`MeshData.instance == None`.

Measured on that file with the #4103 fix applied:

```
# as authored (IfcSite placement carries a 34 degree rotation)
templates=0 instanced_occurrences=0  flat=30  shared_verts=1562110

# same file, IfcSite placement translation and RefDirection neutralised
templates=3 instanced_occurrences=30 flat=0   shared_verts=156211
```

Nothing about the geometry differs between those two runs. The only difference
is which coordinate-space tier the model lands in.

### Cause

`rust/processing/src/element.rs:766-770`:

```rust
    let instance = if ctx.site_local_rotation.is_none() {
        mesh.instance_meta.take()
    } else {
        None
    };
```

and the same test four lines later drops `local_bounds` and `local_to_world`
(#1474). `site_local_rotation` is `Some` for every model in the `site_local`
tier, which `processor/mod.rs:1085-1096` selects whenever the `IfcSite`
placement has a non-identity TRANSLATION. So any model with a site offset,
which is most models out of Revit, discards its instancing metadata.

The stated reason is sound as far as it goes: `convert_mesh_to_site_local`
(`processor/site_local.rs:71`) re-transforms positions and origin AFTER
`instance.transform` was captured, so the captured transform would be stale.
`processor/mod.rs:1202-1206` refuses to arm the don't-bake plan for the same
reason and its comment names the follow-up:

> Extending instancing to site-local needs the renderer to instance in the site
> frame too; tracked as a follow-up.

This is that follow-up, now with a measured cost.

Two things are worth separating:

1. **The guard is wider than its reason.** `apply_inverse_rotation_in_place`
   (`site_local.rs:41-53`) early-returns on an identity rotation, so for a site
   placement that is a pure translation `convert_mesh_to_site_local` does not
   touch positions at all, and the captured transform is not stale. Those models
[...]

PR #4504, opened 2026-09-11
fix(query): containedIn() falls through to a reachable storey on duplicate containment

### LINKED ISSUE #4314: containedIn() has no reachability notion, diverges from elementToStorey when the first-declared storey is unreachable (#4310 follow-up)
## Summary

`EntityNode.containedIn()` (`packages/query/src/entity-node.ts`) has no reachability notion: it is a pure `relationships.inverse.getEdges(elementId, ContainsElements)[0]` lookup over the raw entity graph, independent of whether the returned storey is actually reachable from `IfcProject` in the aggregation tree.

After #4310 (stacked on #4248/#4285), `SpatialHierarchyBuilder`'s `elementToStorey` was fixed to resolve first-declared containment only among *viable* (reachable) storeys, falling through to a reachable later-declared storey when the first-declared one is an orphan/malformed spatial node with no `IfcRelAggregates` edge at all.

`containedIn()` was not changed, so in exactly this shape (an element's first-declared `ContainsElements` edge names a storey that is itself unreachable) the two APIs now disagree in a **new** way: `elementToStorey` returns the reachable storey, `containedIn()` still returns the unreachable one. Both return a present, non-null answer - so this is a silent, different-answer divergence, not a crash or an absent value.

## Reproduction

See `packages/query/test/spatial-storey-parity.test.ts`, test `'elementToStorey and containedIn() disagree when the first-declared storey is unreachable'` (added in #4310) - it builds one `RelationshipGraph` and feeds it into both live implementations:

- Storey A: no `IfcRelAggregates` edge to anything (orphan, unreachable from `IfcProject`)
- Storey B: `Project -> Storey B` via `IfcRelAggregates` (reachable)
- Wall: `ContainsElements` from Storey A (first-declared), then Storey B (second-declared)

Result: `hierarchy.elementToStorey.get(wall) === 3` (Storey B), `wallNode.containedIn()?.expressId === 2` (Storey A).

## Why this wasn't fixed in #4310

Giving `containedIn()` a reachability notion means `EntityNode` would need access to the same reachable-node set `SpatialHierarchyBuilder.assemble()` computes (`computeReachableSpatialNodes`, in `spatial-hierarchy-canonical-parent.ts`) - which in turn depends on `computeCanonicalParent` over the full entity/relationship graph. `EntityNode`'s `containedIn()` currently has no such precomputed context threaded through it; wiring that in is a real, separately-scoped change (and touches `packages/query`'s public surface), not a small addition on top of #4310's parser-side tie-break fix.

## Suggested approach (not prescriptive)

Options to consider when this is picked up:
- Thread a reachable-node set (or equivalent) into `EntityNode`/`IfcDataStore` so `containedIn()` can skip an edge whose target isn't reachable, matching `elementToStorey`'s semantics.
- Or, if that's too invasive for `packages/query`, document the divergence as a known, permanent difference between the "flat map built during one whole-file tree walk" (`elementToStorey`) and the "on-demand single-edge lookup" (`containedIn()`) APIs, and surface it in both APIs' doc comments so callers know which one to use when malformed spatial structures are a concern.

## Scope

Parser-side (`packages/parser`) is unaffected by this issue - `elementToStorey` is already correct. This is `packages/query`'s `EntityNode.containedIn()` only.

Related: #4248, #4310, #4285/#4246.

BIMvoice:
Closes #4314

## What was wrong

`EntityNode.containedIn()` (`packages/query/src/entity-node.ts`) had no reachability notion: it returned `getRelated(ContainsElements, 'inverse')[0]`, the first-declared `IfcRelContainedInSpatialStructure` candidate (#4311's tie-break), regardless of whether that container is reachable from `IfcProject`.

After #4310, `SpatialHierarchyBuilder`'s `elementToStorey` resolves the same tie-break only among *reachable* storeys, falling through to a reachable later-declared storey when the first-declared one is an orphan with no `IfcRelAggregates` edge at all. `containedIn()` did not, so on exactly that shape the two APIs answered "which storey is this element on" with a present but **different** value.

## Fix — follows the maintainer decision on #4314

The decision was to thread the reachable-node set into `containedIn()` and to **reuse `computeReachableSpatialNodes`** (`packages/parser/src/spatial-hierarchy-canonical-parent.ts`) rather than write a second reachability walk. That is what this PR does, and there is no second walk anywhere in it:

- `SpatialHierarchyBuilder.build()` already calls `computeReachableSpatialNodes` once per build to resolve `elementToStorey`'s own tie-break. It now publishes that same set on the hierarchy as `SpatialHierarchy.reachableSpatialNodes` (`packages/data/src/types.ts`), alongside the existing optional `ambiguousStorey` (#4311) and carried across the worker transport the same way (`packages/parser/src/data-store-transport.ts`).
- `containedIn()` reads `this.store.spatialHierarchy?.reachableSpatialNodes`. Both answers therefore come from one call to one function, in one place, and cannot drift apart.

First-declared still wins among reachable containers. When no candidate is reachable, or the store carries no spatial hierarchy at all, it falls back to the first-declared candidate — the absence of the set means "no reachability information", never "nothing is reachable" — so a disconnected spatial tree never turns a present answer into `null`. (A store with no spatial hierarchy also has no `elementToStorey`, so there is nothing for it to disagree with.)

### Why the set is threaded rather than computed in `packages/query`

`computeReachableSpatialNodes` needs a `canonicalParent` map, and `computeCanonicalParent` reads `entities.getTypeEnum(...)` and `relationships.inverse.offsets` — neither of which is on the narrow `EntityTable` / `RelationshipEdges` interfaces that `IfcStoreBase` (`packages/data/src/data-store.ts`) exposes, so `EntityNode` cannot call it from a store alone without widening those interfaces for every duck-typed store in the repo. Reading the builder's published set avoids that, costs nothing (no recomputation), and is strictly stronger for parity: it is the very set `elementToStorey` was resolved against, not an equivalent one.

## Acceptance test

`packages/query/test/spatial-storey-parity.test.ts` was the acceptance test named in the decision. It asserted the divergence; it now asserts agreement:

- before: `expect(hierarchy.elementToStorey.get(10)).toBe(3); expect(wallNode.containedIn()?.expressId).toBe(2);` under a test named `…disagree when the first-declared storey is unreachable`
- after: `…agree when the first-declared storey is unreachable`, asserting `3` from both, plus the direct cross-implementation assertion `expect(hierarchy.elementToStorey.get(10)).toBe(wallNode.containedIn()?.expressId)`

Both parity cases now feed **one** `SpatialHierarchyBuilder` run into both APIs (`new EntityNode({ ...store, spatialHierarchy: hierarchy }, 10)`), so the set `containedIn()` consults is the real builder output, not a hand-written fixture.

## Tests

- `packages/query/test/entity-node.test.ts`: orphan first-declared → reachable storey wins; mirror-image fixture (reachable declared FIRST) → still the reachable one, so the fix is not "return candidate[1]"; every candidate unreachable → first-declared, never `null`; no spatial hierarchy → first-declared, never `null`.
- `packages/parser/test/spatial-hierarchy-builder.test.ts`: `build()` publishes exactly the visited nodes (project + reachable storey, orphan absent) and `elementToStorey` lands on the same storey.
- `packages/parser/test/data-store-transport.test.ts`: `reachableSpatialNodes` survives the worker transport, and stays absent when the source hierarchy has none.

**RED** — with `containedIn()` reverted to `main`, the flipped parity test fails `expected 2 to be 3`, together with the entity-node fall-through case (2 failed / 305).
**GREEN** — `@ifc-lite/query` 20 files / 305 tests, `@ifc-lite/parser` 125 files / 1376 passed + 10 skipped, `@ifc-lite/data` 24 files / 291 tests, all pass.
**Mutation** — three mutations, each with a probe proving the mutated line ran:
1. `return candidates[0]` in place of the reachability lookup → probe printed `reachable set = 1,3 candidates = 2,3` (the real builder set, orphan `2` absent) and 2 tests reddened.
2. `if (!reachable) return null` → probe fired, 3 tests reddened.
3. `?? null` in place of `?? candidates[0]` → probe fired, the all-unreachable fallback test reddened.

## Release / API surface

`pnpm changeset`: `@ifc-lite/data` minor (new optional `SpatialHierarchy` field), `@ifc-lite/parser` minor, `@ifc-lite/query` patch. `pnpm api-surface:update` produces **no diff** — `scripts/api-surface.json` records only `"SymbolName: kind"`, and this change adds a field to an already-exported interface rather than an export, so the snapshot is correct unchanged (verified: `check-api-surface` passes, `git diff scripts/api-surface.json` is empty). `node scripts/check-module-size.mjs` is clean: `packages/data/src/types.ts` is at its recorded 508-line budget, no allowlist row raised.

## Checked, and not checked

`packages/ifcx/src/hierarchy-builder.ts` and `apps/viewer/src/utils/serverSpatialHierarchy.ts` also construct a `SpatialHierarchy`; neither runs canonical-parent resolution, so neither supplies `reachableSpatialNodes` and `containedIn()` keeps its pre-#4314 first-declared answer on those two paths. That is the documented "no reachability information" case, and those builders derive `elementToStorey` from their own sources rather than from `computeReachableSpatialNodes`, so no new divergence is introduced. I did not extend either of them to compute the set.

I did not run the viewer e2e suite or a real-model oracle for this change; the fixtures here are synthetic relationship graphs, because the defect is a malformed-file shape (one element named by two storeys, one of them an orphan) that a well-formed authored model does not contain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spatial containment resolution when an element is associated with multiple storeys.
  * `containedIn()` now selects a storey reachable from the project, matching `elementToStorey` behavior.
  * Preserved fallback behavior when reachability information is unavailable or no candidate is reachable.
  * Reachability information now remains available across data transport, ensuring consistent results.

### DIFF
--- a/.changeset/query-containedin-reachability.md
+++ b/.changeset/query-containedin-reachability.md
@@ -1,0 +1,5 @@
+---
+"@ifc-lite/query": patch
+---
+
+Fix `EntityNode.containedIn()` returning an unreachable spatial container when an element is duplicate-declared in more than one `IfcRelContainedInSpatialStructure` edge (a malformed file naming the same element from two different storeys) and the first-declared storey is itself an orphan with no `IfcRelAggregates` edge back to `IfcProject` at all. `containedIn()` previously always took the first-declared candidate (`#4311`'s tie-break), so this exact shape returned a container no caller can walk anywhere from. It now prefers the first-declared candidate that is reachable from `IfcProject` by walking `decomposedBy()` parents, falling through to a reachable later-declared one, and only falls back to the first-declared candidate when none are reachable.
--- a/docs/architecture/evidence/appearance-query-scopes/README.md
+++ b/docs/architecture/evidence/appearance-query-scopes/README.md
@@ -21,7 +21,7 @@ verify the texture and occurrence ownership directly.
 
 [The evidence record](viewer-acceptance.json) pins source, runtime, fixture and
 asserted states. The manual harness is
-[`accept-query-scope.mjs`](../../../../tools/texture-authoring/accept-query-scope.mjs);
+[`accept-query-scope.mjs`](https://github.com/LTplus-AG/ifc-lite/blob/main/tools/texture-authoring/accept-query-scope.mjs);
 run it against an isolated local viewer with `EVALUATED_URL` and `EVALUATED_OUT`.
 It closes its browser in `finally`. This is behavior evidence, not a performance
 measurement or a new IFC export interoperability claim.
--- a/docs/architecture/evidence/pdf-vectors/README.md
+++ b/docs/architecture/evidence/pdf-vectors/README.md
@@ -32,7 +32,7 @@ vectors would be a distinct tracing/OCR operation.
 
 ## Controlled cases
 
-The [original CC0 generator](../../../../tools/texture-authoring/pdf-vector-controls.py)
+The [original CC0 generator](https://github.com/LTplus-AG/ifc-lite/blob/main/tools/texture-authoring/pdf-vector-controls.py)
 creates two pages, independently rendered below with Poppler's `-cropbox` option.
 Page one has a 72-native-unit line, one rectangle, one cubic curve, and a dashed
 line under translation, rotation and non-uniform scale. The page has a nonzero
--- a/packages/query/src/entity-node.ts
+++ b/packages/query/src/entity-node.ts
@@ -141,8 +141,42 @@ export class EntityNode {
   }
   
   containedIn(): EntityNode | null {
-    const nodes = this.getRelated(RelationshipType.ContainsElements, 'inverse');
-    return nodes[0] ?? null;
+    const candidates = this.getRelated(RelationshipType.ContainsElements, 'inverse');
+    if (candidates.length === 0) return null;
+    if (candidates.length === 1) return candidates[0];
+    // #4314: more than one candidate (a malformed file naming this element in
+    // more than one IfcRelContainedInSpatialStructure edge, #4311) - prefer
+    // the first-declared candidate that is actually reachable from
+    // IfcProject over one that is not, instead of blindly taking `[0]`. A
+    // storey with no IfcRelAggregates edge at all (an orphan/malformed
+    // spatial node) is a dangling answer no caller can walk anywhere from;
+    // falling through to a reachable later-declared candidate keeps this
+    // aligned with `SpatialHierarchyBuilder.elementToStorey`'s tie-break
+    // (#4310), which already skips an unreachable first-declared storey the
+    // same way. Falls back to the first-declared candidate when NONE are
+    // reachable, so this never returns null just because the whole file's
+    // spatial tree is disconnected from IfcProject.
+    const reachable = candidates.find((candidate) => candidate.isReachableFromProject());
+    return reachable ?? candidates[0];
+  }
+
+  /**
+   * Is this entity reachable from an `IfcProject` by walking forward
+   * `IfcRelAggregates` parent edges (`decomposedBy()`)? Used by
+   * `containedIn()` (#4314) to disqualify a duplicate-declared container
+   * that is itself an orphan spatial node. Cycle-safe: a loop that never
+   * reaches an `IfcProject` answers `false` rather than looping forever.
+   */
+  private isReachableFromProject(): boolean {
+    const visited = new Set<number>();
+    let current: EntityNode | null = this;
+    while (current) {
+      if (visited.has(current.expressId)) return false;
+      visited.add(current.expressId);
+      if (current.type === 'IfcProject') return true;
+      current = current.decomposedBy();
+    }
+    return false;
   }
 
   /**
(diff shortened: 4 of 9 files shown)


louistrue:
@BIMvoice Please prioritize getting this PR merged before taking on more unassigned work.

The current head (`a17816ff`) is mergeable and all implementation gates are green: build/WASM, Rust/Node, typecheck, lint, viewer tests, E2E, provenance, schema conformance, parity, SDK canary, and CodeRabbit. The remaining blocker is the failed required **Claude review / Review posted** status. Please rerun or otherwise clear that required review status on this same head, resolve any resulting actionable feedback, and keep `Closes #4314` in the PR body. Avoid another code push unless a review finding requires it, since that would restart the green suite.

After #4504, please finish or hand back #4648 before picking up anything new. Its current merge blockers are:
- replace the invalid `@ifc-lite/core` changeset target;
- tighten the partition-volume/flux bound for large hosts with small openings and add that regression;
- resolve the newly torn census hosts rather than blessing the regressions;
- make the revert-oracle mutation attributable and observed.

Then please work through your existing assigned queue (#4206, #4205, #4203, #4118, #3980, #3977, #3917, #3914, #3612) before claiming more unassigned issues.

louistrue (approved):
Reviewed current head a17816ff4b310622ffc400d3e54e813cf8f68522. The implementation reuses the canonical reachable-node set, preserves it across transport, and makes containedIn() agree with elementToStorey while retaining the documented fallback. Applicable code, type, test, parity, canary, and independent review gates are green. The hosted Claude lane failure is infrastructure-only (provider silent exit plus exhausted fallback credits).

Issue #1480, opened 2026-07-01
Visibility Terrain / Annotation


MennoMekes:
What Classes are turned on/off when using the Site visibility?
I would think IfcGeographicElement would be included?

<img width="1393" height="847" alt="Image" src="https://github.com/user-attachments/assets/34dc7dae-36ef-46d5-8617-52faf709b404" />

IfcAnnotation: The lines I created are not turned off when hiding them in the spatial panel:

<img width="1354" height="846" alt="Image" src="https://github.com/user-attachments/assets/48f20ebb-ef17-4a30-8938-9f4e6dd39811" />

When using CLASS they are not shown. Text is working fine.
<img width="1234" height="768" alt="Image" src="https://github.com/user-attachments/assets/e3f55cc1-a972-4afd-94d8-c842ea993868" />

Issue #345, opened 2026-03-10
Icorrect IDS validation with 'Prohibited' usage


semhustej:
I am getting incorrect IDS validation results when using `Prohibited` in specification description.

I am providing sample IDS and IFC files. The IFC file should not pass the validation, as it contains `IfcBuldingElementProxy` with `GlobalId` attribute.

[IDS-validation-sample-buildingelemntproxy.zip](https://github.com/user-attachments/files/25864620/IDS-validation-sample-buildingelemntproxy.zip)

Note: I am getting correct results when validating this IFC in IfcTester

![Image](https://github.com/user-attachments/assets/3edef494-24e4-4738-adc4-5f0870e973a6)

PR #4656, opened 2026-09-13
fix(viewer): count physical objects with geometry in the hierarchy browser

### LINKED ISSUE #4655: Storey object counts disagree between the spatial tree and the By Class tab, and include entities with no geometry
Selecting a building storey and reading its object count gives two different answers depending on which panel you read, and both include entities that have no geometry.

## Reproduction

`Building-Structural.ifc` (buildingSMART PCERT sample scene), storey `#43` = `00 groundfloor`:

- **Spatial tree** derives its count from `IfcRelContainedInSpatialStructure` and applies no geometry or schema filter → **7**.
- **By Class tab** filters through `makeAssemblyGeometry.renders()` (own `geometricIds`, or a geometry-bearing aggregated part) → **6**.

Same storey, same file, two numbers.

The seventh entity is:

```
#162=IFCBUILDINGELEMENTPROXY('1CjP_CWub368bZVuVHeHs3',#1,'Group#21',$,$,#165,$,'454425.…',$);
#165=IFCLOCALPLACEMENT(#45,#166);
```

Slot 6 is the `ObjectPlacement`; slot 7, the **Representation, is `$`**. It has a position and no shape — it cannot be seen, isolated, measured, sectioned or exported to any mesh format. The name is an authoring-tool artifact (a CAD group flattened into a proxy), not an `IfcGroup`; the entity is legitimately an `IfcElement` descendant and should remain *visible* in the tree, but it should not inflate a number labelled "objects".

## Same root cause, wider than one entity

The counts test *"does this render"* as a proxy for *"is this a physical object"*, and the proxy leaks in both directions:

- **`IfcAnnotation` is counted.** It carries a representation, so the render test admits it. The CLI's `filterBuildingElements` excludes every annotation unconditionally as a drafting aid.
- **`IfcSpace` is admitted by the same test.** Spaces are `IfcProduct` but not `IfcElement`, and unlike groups they usually *do* carry geometry. `Building-Architecture.ifc` aggregates two spaces (`#89`, `#203`) under this same storey `#43`, and `#89` itself contains `IfcFurniture` and a proxy — so whether a storey total is 7, 9 or 11 currently depends on which rule you follow.
- **Before `geometricIds` is computed the filter is a no-op.** `applyFilter` is false, so every entity passes — `IfcGroup`, `IfcZone`, `IfcSystem` and the spatial chain included. Transient during load, but a state the UI passes through, and the easiest way to see a group in a count.

## Two correct rules already exist in the repo

- `apps/viewer/src/lib/physical-objects.ts` — `isPhysicalObjectType`: ancestor chain contains `IfcElement`, minus `IfcFeatureElement`/`IfcVirtualElement`. Its doc comment reasons explicitly through groups, zones, systems, spatial containers, annotations and grids. It is used only for the viewport's "N hidden · N ghosted" badge.
- `packages/cli/src/commands/stats-aggregation.ts` — `filterBuildingElements`, phrasing the same intent as *exclusion* (`NON_ELEMENT_SUPERTYPES`) deliberately, so `IfcGrid`, `IfcPort` and structural-analysis items stay counted.

Neither is used by the counts a user reads.

## Expected

One rule, one number, both panels agreeing. A count labelled "objects" should be **physical elements that have geometry** — which needs the schema test *and* a shape test, since annotations pass the second but not the first.

Note `IfcBuildingElementProxy` must not be excluded as a class: two further proxies in the same file, `'origin'` and `'geo-reference'`, do carry geometry, and real exported models rely heavily on that class. The discriminator is *has a shape*, never the class alone and never the name.

Suggested shape, to keep the main UI uncluttered: the headline stays a single trustworthy number, with the breakdown on hover — type mix, plus an "N without geometry" line shown **only when non-zero**, since an element with a placement and no representation is usually an export defect worth surfacing rather than hiding.

BIMvoice:
Closes #4655

## The problem, on real files

Two panels answered the same question differently, and both answers counted things that are not objects.

`tests/models/local/Building-Structural.ifc`, storey `#43` `00 groundfloor` (7 entities via `IfcRelContainedInSpatialStructure`):

| | spatial badge | By Class tab (model-wide) |
|---|---|---|
| before, geometry loaded | **7** | 17 — the storey's members reduce to **6** |
| before, still streaming | **7** | **18** across 7 classes |

The storey's `#162=IFCBUILDINGELEMENTPROXY(...,'Group#21',...)` has an `ObjectPlacement` and `$` for Representation. It draws nothing, so the class tab drops it and the badge counted it: same storey, two numbers.

`tests/models/local/Building-Architecture.ifc`, before, in the window where `geometricIds` is still empty and the tab's filter is therefore inert:

```
BYCLASS ABSENT: rows=35 total=97 IfcBuilding:1 ... IfcRelAggregates:6 IfcRelDefinesByType:16
                IfcSIUnit:3 IfcSpace:2 IfcSpatialZone:1 IfcSurfaceStyle:10 ... IfcZone:1
```

`IfcZone:1` — a group-family entity in a list of products, which is the reported symptom reproduced on a real file. `Building-Hvac.ifc` shows the same with `IfcDistributionSystem:1`.

## The rule

One definition, used by both panels: **a physical object that has a shape.**

- The schema half is `lib/physical-objects.ts`'s `isPhysicalObjectType` — `IfcElement` minus feature and virtual elements. It is imported, **not modified**: the file is byte-identical to `main` in this PR. It is what rules out the annotation, the group, the zone, the system and the spatial chain, by inheritance rather than by a list anyone has to maintain.
- The shape half is `makeAssemblyGeometry.renders`, the By Class tab's own predicate — not a second reading of the mesh set. That is what makes the two panels structurally unable to disagree, and it is why an `IfcRoof` whose geometry sits on its eight aggregated beams counts once rather than zero.

Both halves are needed and each rules out a different row: the annotation passes the shape test and fails the schema test, the marker proxy the reverse. `IfcBuildingElementProxy` is not excluded as a class and nothing is filtered on a name — `origin` and `geo-reference` in the same file are proxies that do have shapes and do count.

**Before geometry arrives** the shape test is unanswerable, so the count reports every physical object and the hover says so, rather than reading an empty mesh set as "nothing has a shape". It narrows as meshes land; it never passes a group or a container at any point in the load.

### After

| | spatial badge | agrees with the tab |
|---|---|---|
| `Building-Structural` `00 groundfloor`, loaded | **6** — 4 IfcWall, 1 IfcChimney, 1 IfcFooting; 7 rows; 1 element without geometry | yes |
| same, still streaming | **7**, marked provisional on hover | yes |
| `Building-Architecture` `00 groundfloor`, loaded | **5** — 4 IfcWall, 1 IfcSlab; 7 rows; 2 without geometry; 2 spaces (not counted) | yes |
| `Building-Hvac` `00 groundfloor`, loaded | **3** — 2 IfcAirTerminal, 1 IfcDuctSegment; 1 without geometry | yes |
| `Building-Structural` building row, loaded | **1** (the roof, through its parts) — was 0 in an intermediate version of this branch, which is why the shape test reuses `renders` | yes |

By Class is identical in both geometry states for the non-object classes: no group, zone, system, property set, type entity or spatial container, ever.

The "without geometry" counts are real model defects the files were hiding: `Building-Architecture`'s `#339=IFCCHIMNEY` and `#345=IFCBUILDINGELEMENTPROXY`, and `Building-Hvac`'s `#52=IFCCHIMNEY`, all carry `$` for Representation — cross-discipline reference stubs.

## Show versus count

Nothing was removed from the tree. The spatial tab still lists every contained entity, so `Group#21` and the annotation stay selectable and inspectable; only the number changed. Pinned by tests, in both directions.

The By Class tab keeps showing `IfcAnnotation` (#1480) and rendering spaces, because a row there exists to be clicked — but neither is counted as an object. That divergence is deliberate and asserted.

## Hover

Headline stays one number. Under it, only lines that have something to say:

```
6 objects
4 IfcWall · 1 IfcChimney · 1 IfcFooting
1 element without geometry
```

A clean container shows the first two lines and nothing else. `countBadgeLabel.test.ts` asserts the absence of a "0 without geometry" line as its own case.

## Evidence

- **RED**: with the geometry proxy in place the new suite fails in both geometry states — storey badge 3 where 2 is correct, and the tab listing `IfcGroup`/`IfcZone`/`IfcSystem`/`IfcPropertySet`/the spatial chain while streaming.
- **GREEN**: 73 tests across `treeDataBuilder.test.ts` and `treeDataBuilder.physicalObjects.test.ts`, plus the untouched `physical-objects.test.ts`.
- **MUTATION**: five separate mutations, each with a `console.error` probe proving the mutated line executed, each reverted after:

| mutation | probe output | killed by |
|---|---|---|
| A — drop the product-tree class gate | `MUTANT-A-CLASS-GATE-RAN IfcGroup` | 2 tests (a site with terrain geometry then lists as a product) |
| B — streaming window passes everything | `MUTANT-B-STREAM-RAN` ×117 | 2 tests |
| C — badge counts contained rows | `MUTANT-C-BADGE-RAN 5 3` | 7 tests |
| D — drop the shape half | `MUTANT-D-SHAPE-TEST-RAN IfcBuildingElementProxy false` | 8 tests |
| E — emit breakdown lines unconditionally | `MUTANT-E-LINE-GATE-RAN 0` | 2 tests |

(Mutations were run against the tooltip's own unit tests before those assertions moved onto `countTooltipLines`; the same cases are asserted through the tree today.)

- **Changed tests observe production**: OBSERVED locally (`node scripts/check-test-revert-oracle.mjs --base upstream/main`). The first push was INCONCLUSIVE for two reasons worth naming: the new test files imported the new modules directly, so reverting production deleted them and the suite failed to LOAD rather than to assert; and the viewer store's `localStorage` warning at module init put a `ReferenceError` in the runner output, which the oracle reads as a dead import. Both are fixed by entering through seams that exist at base — every assertion now runs through `buildTreeData` / `buildTypeTree`, the hover lines ride on `TreeNode.countTooltipLines` so they are asserted through the same entry point as the number they explain, and `@/test/setup-dom` registers the DOM globals first.
- Full viewer suite, all four shards: 8186 tests, 0 failures, 15 fixture-gated skips (all pre-existing and unrelated — `pnpm fixtures` / wasm corpus).
- `pnpm lint` and `node scripts/check-module-size.mjs` clean. `treeDataBuilder.ts` is 1375 lines against a former budget of 1424; its row is lowered, not raised, and the membership rule moved out into `productTree.ts` / `productTreeClasses.ts`.
- `apps/viewer` is private, so no changeset.

## Two rules that disagree, for the record

`packages/cli`'s `filterBuildingElements` phrases the same intent as exclusion (`NON_ELEMENT_SUPERTYPES`) and deliberately keeps `IfcGrid`, `IfcPort` and structural-analysis items. Measured against `isPhysicalObjectType` over 34 classes, they agree on everything the viewer counts here — walls, doors, footings, chimneys, proxies, accessories, and every group/zone/system/spatial/annotation exclusion — and differ on: `IfcGrid`, `IfcGridAxis`, `IfcPort`, `IfcDistributionPort`, `IfcOpeningElement`, `IfcVoidingFeature`, `IfcVirtualElement`, `IfcStructuralCurveMember`, `IfcStructuralPointConnection`, `IfcAlignment` (CLI keeps, viewer does not). The CLI's filter runs on an already-scoped `IfcObjectDefinition` query, so its apparent divergence on `IfcPropertySet`/`IfcWallType` is unreachable in practice. No attempt to unify them here — that is a product call, and the divergence is now written down rather than latent.

## Other count sites checked, not changed

- `statusBarStats.ts` + `StatusBar.tsx` — "N elements" is mesh-derived, dedupes per mesh so an entity in two meshes counts twice, and misses fully-instanced entities. Its numerator switches to a containment-list length when storeys are filtered, so the two halves of `"1234 / 5678 elements"` use different rules.
- `PropertiesPanel.tsx` `"Contained Elements"` / `"Elements (Storey)"` — raw `elements.length`, so it now differs from the badge for the same storey. A fix was written and reverted: `PropertiesPanel.tsx` is at its 2146-line budget and the addition cannot land without splitting a file this PR has no other reason to touch.
- `ViewportOverlays.tsx` "N hidden · N ghosted" — already schema-based, but single-model (`ifcDataStore`), so it under-reports in a federation.
- `charts` `elementsDataset` — gates on `EntityFlags.HAS_GEOMETRY`, which the parser sets **by class bucket, not by Representation presence**: `entities.hasGeometry(162)` returns `true` for `Group#21`, which has `$`. That flag does not mean what its name says.
- Model-header rows badge `dataStore.entityCount` (every STEP line) in the same column as the object counts; type-group rows badge instance counts; model-tag rows badge a model count.
- Selection counts are implemented three times (`MainToolbar.tsx:334`, `ElementsTab.tsx:44`, `BCFPanel.tsx:81`), two more omit the `selectedEntityId` fallback, and `SearchModal.text.tsx` reads a different store set.
- Export dialogs count raw entities/nodes/meshes; `AnonymizedExportDialog.tsx` calls the same number "entities" on one line and "objects" on another.

---

**Unlinked-issue caveat:** #4655 does not yet carry the `ready` label, so the `Issue queue` check will fail with the label gate until a maintainer applies it. That is expected; `unqueued` has deliberately not been self-applied, since the gate reads the label's actor.

## Summary by CodeRabbit

* **New Features**
  * Added detailed hierarchy count-badge hover breakdowns, including object types, geometry status, and uncounted spaces.
  * Improved product-tree support for assemblies, federated models, spatial structures, annotations, and geometry filtering.
  * Counts now focus on physical objects with available geometry and remain consistent across hierarchy views.

* **Bug Fixes**
  * Improved count accuracy for assemblies, unified storeys, and nested spaces.
  * Corrected hierarchy behavior while geometry is loading or unavailable.


louistrue (changes requested):
Blocking review finding: the single-model spatial path removes descendant-space elements with getSpatialNodeElements() and records spacesNotCounted, but buildUnifiedStoreys() summarizes each raw hierarchy.byStorey array directly and always leaves spacesNotCounted at 0. In federated/unified mode, an element with both storey and descendant-space containment can therefore be counted in the storey even though the equivalent single-model row excludes it, and the hover breakdown omits the excluded-space count. Please apply the same descendant-space exclusion/counting semantics in buildUnifiedStoreys and add a multi-model unified-storey regression containing an IfcSpace and a physical child. Also resolve the revert-oracle INCONCLUSIVE result: its whole-file reverse removes the new test seam and produces ERR_MODULE_NOT_FOUND/localStorage load failures rather than attributable RED assertions; use the documented behavioral mutation or a justified exemption after the regression is added. I am not marking issue #4655 ready or merging until these are addressed.

louistrue (approved):
Approved on exact head 3c2c7f4e48a3 after independent review and regression verification. The unified and single-model hierarchy paths now count the same descendant object space; partial hierarchy stores remain supported; geometry readiness is preserved per federated model; known-empty geometry is distinguished from unavailable geometry; and the strengthened deep-aggregate/projectless/parity tests pass. Full current-head Ollama and Claude reviews, all viewer shards, typecheck, lint, E2E, issue queue, provenance, aggregate CI, and Vercel checks are green.

PR #4664, opened 2026-09-13
fix(export): distinguish empty isolation filter from no filter in JSON-LD and STEP export

### LINKED ISSUE #4659: JSON-LD and STEP export the whole model when an isolation filter matches nothing
`ifc-lite export --format jsonld` and `--format step` silently export the **whole model** when an isolation filter matches nothing. The user asks for a subset, the subset is empty, and they receive everything.

This is the same defect as #4483/#4484 (GLB and OBJ), which are fixed. JSON-LD and STEP were never covered.

## Reproduced at byte level

On `upstream/main`, with the branch's own freshly-built wasm, a zero-match export diffed against a true whole-model export of `hello-wall.ifc`:

| format | zero-match output | whole-model output | identical? |
|---|---|---|---|
| `jsonld` | 1491 bytes, 9 `@id` nodes | 1491 bytes, 9 nodes | **yes** |
| `step` | 79580 bytes, 1045 entities | 79580 bytes, 1045 entities | **yes** |

A real `--type IfcWall` filter narrows correctly (JSON-LD 475 bytes / 1 node; STEP 2182 bytes / 46 entities), so this is specifically the *matched-nothing* case collapsing back to "no filter", not a broken fixture.

## Why these two and not their siblings

| format | guards |
|---|---|
| `obj` | shared TS zero-match guard, **plus** `countObjVertices(out) === 0` |
| `gltf` / `glb` | shared guard, **plus** `isNoRenderGeometryError` catch, **plus** `countGlbMeshes(out) === 0`; `export_glb.rs` also fails closed with `NO_RENDER_GEOMETRY` |
| `jsonld` | shared guard only |
| `step` | shared guard only |

Neutering the shared TS guard in `packages/cli/src/commands/export-rust-formats.ts` reddens 5 of 13 tests in `export.zero-match.test.ts`: `obj`/`gltf`/`glb` still reject via their second guard; `jsonld`/`step` report success.

The cause is structural rather than a missing test. `export_obj` / `export_glb` take `isolated: Option<Vec<u32>>`, and the caller passes `filterActive ? isolated : undefined`, so an active-but-empty set genuinely filters everything out — which is what lets the count checks catch it. `export_jsonld` (`rust/wasm-bindings/src/api/export_data.rs`) and `export_step` (`export_step.rs`) take a bare `included: &[u32]` and collapse `included.is_empty()` back to `None` internally, documented as "empty ⇒ whole model". That is the null-vs-empty collapse #4483/#4484 removed for the mesh formats.

## Not intentional, as far as the code says

Nothing in either writer or its doc comments treats a zero-entity JSON-LD or STEP file as a meaningful output; the empty-array behaviour is described purely as a convenience default for "no filter requested". No test or doc asserts a zero-match filter should export everything for these two.

## Expected

A filter that matches nothing exports nothing, and says so — consistent with `obj`/`gltf`/`glb`.

Suggested shape: give `export_data.rs` and `export_step.rs` the same `Option<Vec<u32>>` null-vs-empty treatment, and add a TS-side zero-entity check after those branches mirroring `countObjVertices` / `countGlbMeshes`. At minimum, a comment at the shared guard noting those two formats have no second line of defence, so a future refactor of that line does not remove their only protection unnoticed.

Not checked: the MCP/SDK direct-call paths for the same export functions (only the CLI path was examined), and `ifcx`/`usd`, which are explicitly whole-model and never isolate.

BIMvoice:
Closes #4659

## The defect, reproduced

`export_jsonld` (`rust/wasm-bindings/src/api/export_data.rs`) and `export_step` (`export_step.rs`) took a bare `included: &[u32]` and collapsed `included.is_empty()` back to `None`, documented as "empty ⇒ whole model". Their mesh siblings take `isolated: Option<Vec<u32>>`, so Rust can tell "no filter" from "filter matched nothing" and an empty `Some(vec![])` genuinely filters everything out.

Measured on `hello-wall.ifc` at `4e27bd9` with that branch's own wasm build, through `GeometryProcessor`:

| call | bytes | content |
|---|---|---|
| `exportJsonld(bytes, '', true, false, false, undefined)` | 1491 | 9 `@graph` nodes |
| `exportJsonld(bytes, …, new Uint32Array())` | 1491 | 9 nodes |
| `exportJsonld(bytes, …, new Uint32Array([1222]))` | 475 | 1 node |
| `exportStep(bytes, '', undefined)` | 79580 | 1045 entities |
| `exportStep(bytes, '', new Uint32Array())` | 79580 | 1045 entities |
| `exportStep(bytes, '', new Uint32Array([1222]))` | 2182 | 46 entities |

The zero-match and whole-model calls were in fact the *same call* — the signature could not express the difference. A real filter narrows correctly, so this is specifically the matched-nothing case collapsing back to "no filter", not a broken fixture.

## The fix

`JsonLdOptions::included` becomes `Option<Vec<u32>>` and both wasm bindings take `Option<Vec<u32>>`, matching the convention #4364 established for GLB and #4386 for OBJ: `None` / `undefined` = no filter, `Some(ids)` / a `Uint32Array` = an active allowlist whether or not it is empty. An active-but-empty set now yields an empty `@graph` and a header-only STEP file.

`ifc_lite_export::export_step_json` already took an `Option` and got this right; the collapse lived only in the binding above it. A cargo test now pins that contract in both directions so it cannot be re-collapsed there either.

### Call sites

Every caller of these two bindings, checked by grep across the repo for `.exportStep(` / `.exportJsonld(` and `export_jsonld` / `export_step`:

- `rust/wasm-bindings/src/api/export_data.rs` — signature changed.
- `rust/wasm-bindings/src/api/export_step.rs` — signature changed.
- `packages/geometry/src/ifc-lite-bridge.ts:520,564` — `Uint32Array | undefined`, defaulting to `undefined`.
- `packages/geometry/src/index.ts:1155,1167` — same.
- `packages/cli/src/commands/export-rust-formats.ts:96,100` — now passes `filterActive ? isolated : undefined`, as the `obj`/`glb` branches already did. Without this, every *unfiltered* `jsonld`/`step` export would fail closed under the new meaning of an empty array — the same same-PR follow-up #4364 and #4386 each needed.
- `packages/wasm/pkg/ifc-lite.d.ts` — regenerated via `build-wasm.sh`.
- Docs: `docs/api/wasm.md`, `docs/guide/exporting.md`.

No MCP or viewer caller exists: `packages/mcp/src/tools/export.ts` has no `jsonld` or `step` tool (grepped for both), and `apps/viewer/src/lib/export/changed-model-export.ts`'s `exportStep` is the TypeScript `StepExporter`, not this binding.

## Refuse, or export nothing?

Both — one at each layer, which is what makes the five formats consistent:

- **Rust exports nothing.** An active-but-empty allowlist produces an empty `@graph` / an empty `DATA;` section, exactly as `export_obj` produces a header-only OBJ. The artifact is well-formed and contentless.
- **The CLI refuses.** `export-rust-formats.ts` fatals rather than writing a contentless file, exactly as it does for a 0-vertex OBJ and a 0-mesh GLB.

Refusing inside Rust instead would diverge from `export_obj`, and leaving the CLI to write the empty file would diverge from `obj`/`glb`. This way all five formats behave identically at both layers.

## The second line of defence

`@ifc-lite/export` gains `countJsonldNodes` and `countStepEntities` (`packages/export/src/zero-content.ts`). A byte-length check cannot do this job: JSON-LD keeps its `@context` around an empty graph and the STEP writer regenerates its ISO-10303-21 header, so a contentless export is several hundred valid bytes. `countJsonldNodes` parses the graph rather than tallying `"@id"`, which also appears on nested property-set objects.

The JSON-LD check is reachable without touching the shared guard, so it is not unreachable decoration: `ifc-lite export --format jsonld --type IfcProject` matches an entity but produces no graph node (`build_export_model` emits element nodes), and used to write that empty document as a reported success. STEP has no counterpart input — any non-empty allowlist writes at least those instances — so its check is defense-in-depth only, kept for the same reason GLB's `countGlbMeshes` is kept behind the Rust fail-closed error.

## Tests

`packages/cli/src/commands/export.zero-match.test.ts` extended rather than forked (13 → 15 tests), plus one cargo test in each of `rust/export/src/jsonld.rs` and `rust/export/src/step_tests.rs`. The new CLI test exercises the wasm boundary directly, because the shared guard fires first through `exportCommand` and would hide the collapse — which is exactly how it stayed hidden.

**RED** (before the Rust/binding change):
```
AssertionError: expected 9 to be +0
  expect(countJsonldNodes(jsonldZero as Uint8Array)).toBe(0);
Tests  1 failed | 13 passed (14)
```

**Vacuity guards** in the same test: the fixture's `IfcWall` id set is read from the model and asserted non-empty; the unfiltered export must carry more than one node/entity; and the real filter must narrow to fewer. "Exports nothing" cannot pass because the fixture or the exporter produces nothing.

**Mutation**, each guard individually, with a probe proving the mutated line ran:

| mutated | probe | result |
|---|---|---|
| CLI `countJsonldNodes(out) === 0` → `false` | `PROBE-JSONLD-GUARD-RAN` on stderr; the run wrote `{"@context":…,"@graph":[]}` and exited 0 | `--type IfcProject rejects…` ✗ |
| CLI `countStepEntities(out) === 0` → `false` | `PROBE-STEP-GUARD-RAN`; wrote a 243-byte, 0-entity `.step` and exited 0 | `--format step …` ✗ |
| `export_step.rs` pass-through → re-collapse `Some(empty)` to `None` | `PROBE-STEP-WASM-COLLAPSE-RAN` ×3 in the vitest output | `expected 1045 to be +0` ✗ |
| `jsonld.rs` `included.as_ref().map(…)` → `.filter(\|ids\| !ids.is_empty())` | `PROBE-JSONLD-FILTER-COLLAPSE-RAN` ×3 in the cargo stdout | `left: 3, right: 0` ✗ |

**Neutering the shared TS guard** (`if (filterActive && isolated.length === 0)` → `if (false)`) now reddens all five isolating Rust-backed formats, and — the part that changed — `jsonld` and `step` still *refuse*:

```
Error: JSON-LD export produced 0 nodes — … ;  exit 1, no file written
Error: STEP export produced 0 entities — … ;  exit 1, no file written
```

Before this PR they exited 0 and wrote the whole model. (All five "rejects" tests go red either way, because they assert the shared guard's `Filter matched 0 entities` message; the difference is whether the export is refused, which is what the CLI runs above show.)

## Runs

- `cargo test --workspace`: 218 suites, 3350 passed, 0 failed.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `pnpm typecheck`: 96/96 tasks, 2030/2030 test files in a program.
- `@ifc-lite/cli`: 75 files, 947 passed / 8 skipped.
- `@ifc-lite/export`: 109 files, 1292 passed / 37 skipped.
- `@ifc-lite/geometry`: 40 files, 404 passed. `@ifc-lite/mcp`: 41 files, 406 passed.
- `pnpm lint`: 0 errors (5 pre-existing warnings, unrelated files).
- `docs:check-samples`, `docs:check-generated`, `check-changesets`, `check-source-text-assertions`, `check-refwalk-guards`: all green.
- `node scripts/check-module-size.mjs --base upstream/main`: +0 added, ^0 raised — no budget touched.
- `pnpm api-surface:update`: `+countJsonldNodes`, `+countStepEntities` under `@ifc-lite/export`, committed.
- `pnpm test:wasm-contract` skipped clean — the `column` fixture is absent locally (`pnpm fixtures` not run). Not evidence either way; CI restores fixtures.

Perf: no change to any perf-sensitive path's work — the only Rust edit inside `rust/export` swaps an `is_empty()` branch for an `Option` match on the same data.


louistrue (changes requested):
Blocking review finding: ifc_lite_export::JsonLdOptions is a public stable Rust API at workspace version 13.x, and this PR changes its public included field from Vec<u32> to Option<Vec<u32>>. That is source-breaking for downstream struct literals. The committed API snapshot confirms the drift: pnpm check:rust-api-surface fails with the old and new field shapes. The patch changeset does not authorize a Rust major release, and leaving the snapshot stale will fail the main lint lane. Please preserve the existing JsonLdOptions shape and semantics, then add an additive export entry point or equivalent non-breaking boundary that accepts the optional filter for the wasm binding. Update the Rust API snapshot for any additive public symbol and retain None, Some(empty), and Some(non-empty) regression coverage. The STEP side already delegates to an Option-taking API and does not have this particular break.

louistrue (approved):
Approved on exact head d8d7c6b3f after independent review and fixes. The wasm boundary now preserves undefined versus an active empty filter for JSON-LD and STEP, the public JsonLdOptions Rust struct remains source-compatible through an additive export_jsonld_with_filter entry point, and CLI artifact guards fail closed. Targeted Rust regression and wasm checks, Rust API surface, module-size ratchet, full Ollama and Claude reviews, all CI shards, and Vercel checks are green.

PR #4667, opened 2026-09-13
test(export): pin the merged spatial tree against non-uppercase STEP keywords

### LINKED ISSUE #4660: MergedExporter duplicates the shared spatial tree when a merged model uses non-uppercase STEP keywords
`MergedExporter` (Rust) compares STEP entity keywords case-sensitively against values it stored unnormalised, so a federated merge involving a model written with lowercase or mixed-case keywords duplicates the shared spatial tree instead of unifying it, and wrongly retains containers that should be pruned.

STEP entity keywords are case-insensitive per ISO 10303-21, and `EntityScanner::next_entity()` returns the keyword as a raw, unnormalised slice.

## Reproduced

Two-model merge, matching GlobalIds, model 2's containers spelled `ifcsite` / `IfcBuilding` / `IfcBuildingStorey`:

| | site | building | storey |
|---|---|---|---|
| both models uppercase (control) | 1 | 1 | 1 |
| **model 2 mixed-case** | **2** | **2** | **2** |

Model B's containers pass through unmatched, duplicating the shared tree. Its `ifcrelaggregates` lines also survive unpruned and unremapped, consistent with `skip_redundant_rel_aggregates` never firing.

Empty-container pruning, single model, `drop_empty_containers: true`, wall excluded so the containers hold nothing:

| | `dropped_container_count` |
|---|---|
| uppercase (control) | 3 |
| **mixed-case** | **0** |

The containers and their now-purposeless relationship lines are all wrongly retained.

## Cause

`rust/export/src/merged/plan.rs` stores the keyword raw:

```rust
idx.type_of.entry(id).or_insert_with(|| type_name.to_string())
```

The doc comment describes it as an "uppercase STEP type token"; nothing enforces that. Six consumers then compare it case-sensitively:

- `plan.rs` `unify_spatial` — `match ty.as_str() { "IFCSITE" | "IFCBUILDING" | "IFCBUILDINGSTOREY" … }`
- `plan.rs` `skip_redundant_rel_aggregates` — `!= Some("IFCRELAGGREGATES")`
- `plan.rs` → `merged/guid.rs` `is_relationship_type(type_upper: &str)` — `starts_with("IFCREL")`, with nothing upcasing despite the parameter name
- `merged/empty.rs` — `starts_with("IFCREL")`, twice
- `merged/empty.rs` `is_container_type` — exact `CONTAINER_TYPES.contains(…)`

`extract_global_id_fast` / `is_rooted_entity_type` are **not** affected — they delegate to `legacy_aware_ifc_type`, which normalises. GlobalId extraction is already case-insensitive; only these downstream readers are not.

## The TypeScript twin is correct

`packages/export/src/merged-exporter.ts` and friends produce **1/1/1** and `droppedContainerCount: 3` on the identical mixed-case fixture. `packages/parser/src/columnar-entity-preparation.ts` normalises `entityIndex.byType` keys via `getTypeUpper` at parse time, and every remaining raw-case read in the export path uppercases explicitly before comparing.

So this is a one-sided defect, and the fix is a port rather than a design: normalise once where `type_of` is populated, and none of the six call sites need to change.

## Reachability

Stated as inference, not evidence — no export sample was checked. Mainstream authoring tools emit uppercase keywords by convention, so this is unlikely on typical authoring output. But ISO 10303-21 permits mixed case, and a federated multi-model workflow — precisely what `MergedExporter` exists for — is more likely than average to ingest a hand-authored, scripted or third-party-converted file. Reachable but low probability.

Related: this is one shape of the class in #4497. It is the "stored `String`" shape specifically, which a `StepKeyword` newtype at `next_entity()` could not reach, since the value has already become an owned `String` by the time it is compared.

Not checked: the JS `merged-exporter.ts` behaviour beyond the cases above, and other `type_of` readers in the export crate beyond the six listed.

BIMvoice:
Closes #4660

## The behaviour was already fixed — by #4590, ~14 hours before the issue was filed

#4660 reports that a merged model spelling its containers `ifcsite` / `IfcBuilding` / `IfcBuildingStorey` duplicates the shared spatial tree (2/2/2 against a control's 1/1/1) and never prunes empty containers (`dropped_container_count` 0 against a control's 3), because `ModelIndex::type_of` stored the STEP keyword as written while six consumers compared it case-sensitively.

That is a real defect and the diagnosis is exactly right, but **#4590 ("compare STEP keywords case-insensitively in every scan loop", merged 2026-09-12T19:25Z) already folds the keyword once in `ModelIndex::build`**, which is precisely the one-line-at-population fix the issue prescribes. The issue was filed 2026-09-13T09:37Z against a checkout predating it. I verified this rather than assuming: the issue's reproduction, written as a test, **passes on `upstream/main` as it stands**.

So there is nothing to fix. What there is, is a gap: #4590's only export-crate coverage is a `relationships.rs` unit test on the `relationships()` scan. Nothing exercises the merged export end to end on a recased model, so this symptom is fixed and unpinned — the next refactor of `ModelIndex::build` can silently reintroduce it. This PR closes that gap and makes one contract enforceable.

## What this adds

**`rust/export/tests/issue_4660_merged_keyword_case.rs`** — the issue's two reproductions, end to end through `export_merged_models`:

- `mixed_case_keywords_unify_the_shared_spatial_tree` — two models with matching container GlobalIds, the second recased; one site, one building, one storey, and both walls kept.
- `mixed_case_keywords_still_prune_empty_containers` — the wall excluded so every container is empty; `dropped_container_count == 3`.
- `fixture_is_actually_recased` — a guard on the `recase` helper, so a typo there cannot quietly make the other two vacuous.

**Guarded against a vacuous fixture, as the issue's own framing demands.** Each test asserts the **all-uppercase control first** — if the fixture merged nothing, the control would not read 1/1/1 or 3 either, and the test would fail rather than pass for free. And the mixed-case merge is required to be **identical** to the control (`mixed.to_ascii_uppercase() == control.to_ascii_uppercase()`), not merely self-consistent: keyword case is not significant per ISO 10303-21, so the two inputs must produce the same model, not two coherent-but-different ones.

**`is_relationship_type`'s parameter is named `type_upper`, and nothing enforced it.** The issue flags this, correctly, as a comment asserting a normalisation that does not happen. Now that the fold at population is real, the honest repair is to make the name binding rather than to delete it: a `debug_assert` so a future caller reaching it with a raw keyword fails loudly in debug and test builds, instead of silently classifying every relationship as a non-relationship. It fires nowhere in the existing suite. (`plan.rs`'s "folded to uppercase" doc comment, the issue's other naming complaint, was already corrected by #4590.)

## Mutation evidence

Reverting the fold to `idx.type_of.entry(id).or_insert_with(|| type_name.to_string())` reproduces **the issue's exact numbers**:

```
panicked at issue_4660_merged_keyword_case.rs: mixed case: the site must unify, not duplicate
  left: 2   right: 1
panicked at issue_4660_merged_keyword_case.rs: mixed case: all three empty containers pruned
  left: 0   right: 3
```

An `eprintln!` probe on the mutated line proves it ran, and shows exactly what the fold was suppressing:

```
MUTATION-PROBE type_of stores raw "ifcsite"
MUTATION-PROBE type_of stores raw "IfcBuilding"
MUTATION-PROBE type_of stores raw "IfcBuildingStorey"
MUTATION-PROBE type_of stores raw "ifcrelaggregates"
```

Under the mutation the uppercase control assertions still passed and `fixture_is_actually_recased` stayed green — so the two failures come from the merge, not from the recasing.

## Checks

- `cargo test -p ifc-lite-export`: **481 passed / 0 failed** before, **484 / 0** after.
- `cargo clippy -p ifc-lite-export --all-targets -- -D warnings`: clean.
- `node scripts/check-module-size.mjs --base upstream/main`: OK, `+0 ^0 v0 -0`. Rust `module_size_ratchet`: 8 passed.
- No changeset: Rust-only, no published-package behaviour change.

## What I checked, and what I did not

Checked: every `type_of` reader in `rust/export` (`grep -rn "type_of" rust --include="*.rs"`) — `plan.rs` `unify_spatial`, `skip_redundant_rel_aggregates` and the `is_relationship_type` call, `empty.rs`'s two `starts_with("IFCREL")` sites and `is_container_type`, and `mod.rs`/`empty_tests.rs`'s `SpatialLookup::build` closures. All six read `ModelIndex::type_of`, all six therefore see the folded value, and none needs the original casing — the raw line text is carried separately by `line_of`/`line_bytes`, which is what the writer re-emits, so keyword case survives into the output regardless (visible in the identical-output assertion, which has to fold case to compare).

Not checked: `type_of` readers outside `rust/export` (`relationships.rs` has a same-named but unrelated `HashMap<u32, u32>` field, not a keyword map), and the TypeScript `merged-exporter.ts` behaviour beyond what the issue already measured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IFC export handling for mixed-case STEP entity keywords.
  * Ensured merged spatial structures remain consistent and empty containers are correctly pruned regardless of keyword capitalization.

* **Tests**
  * Added end-to-end coverage for mixed-case IFC merging and empty-container cleanup.


louistrue (approved):
Approved on exact head bfcf86042585e30819b0202156096e16ea726e94. The PR is now test-only; the unrelated GUID production assertion was removed. I independently ran the regression (3/3 passing), confirmed it is non-vacuous/mutation-backed, and verified the exact-head full Ollama review reports no major issues. No unresolved review threads.

PR #4674, opened 2026-09-13
fix(export): generate LEGACY_ROOTED_TYPES per schema version instead of hand-keeping it

BIMvoice:
Refs #4203 (partial — see below; does not close it)

## History (checked first)

#4203 asks for two things: (a) generate the Rust type/attribute universe per schema version from the same EXPRESS inputs as the TypeScript side, and (b) delete `legacy_entities.rs` and `LEGACY_ROOTED_TYPES` once the generated tables reproduce every row. #4495 (merged, `Refs #4203`) already did a bounded slice of (a): it generates each legacy entity's OWN attribute names (`legacy_attribute_names.rs`) and explicitly did NOT attempt `legacy_entities.rs`'s base-type mapping or `rooted_type.rs`'s `LEGACY_ROOTED_TYPES`, calling a full per-version `IfcType`/`from_str` rewrite "too large for one reviewable PR" (touches every consumer of the canonical `IfcType` enum across `rust/core`, `rust/export`, `rust/geometry`, `rust/processing`, `rust/wasm-bindings`). louistrue approved #4495 on exactly that scoping ("A deliberate partial slice of #4203 (Refs, not Closes)"). No further issue comments since. No other open PR touches this.

## Scope decision: one more slice, not the full stack

Given the maintainer-approved precedent that the full rewrite is too large for one PR, this PR ships the next self-contained slice using the same "generate from EXPRESS, prove parity" pattern #4495 established, rather than attempting (b) or the `IfcType`/`from_str` rewrite here.

## What

`rust/export/src/rooted_type.rs` hand-kept `LEGACY_ROOTED_TYPES`, a 54-name allow-list of IFC2X3/IFC4 entities IFC4X3 dropped or renamed but that still carry a GlobalId as their first STEP attribute. Its own doc comment already said it was "independently re-verified (2026-08-20) by walking each name's parent chain ... re-verify the same way (**or regenerate from a diff of those three tables**) rather than editing this list ad hoc."

`scripts/generate-legacy-rooted-types.mjs` performs exactly that diff mechanically: for every entity in `packages/data/src/ifc-schema/generated/entities-ifc2x3.ts` / `entities-ifc4.ts` that `rust/core/src/generated/schema.rs`'s `from_str` cannot resolve, it walks the parent chain (reusing the same table shape `check-legacy-entity-coverage.mjs`'s `parseEntityTable`/`droppableProducts` already trust) and keeps the name if it reaches `IfcRoot` in either schema, reporting (and `--check`-failing on) any IFC2X3/IFC4 disagreement. Output: `rust/export/src/generated/legacy_rooted_types.rs`.

**The generated list is byte-identical to the 54-name hand-kept list it replaces, 0 conflicts** — this independently confirms the 2026-08-20 hand audit was correct, and turns "re-verify by hand" into a mechanical check. `rooted_type::LEGACY_ROOTED_TYPES` keeps its exact path and shape (`pub` `&[&str]`), now re-exported (`pub use`) from the generated module instead of defined inline — every existing caller (`is_rooted_type`, `rooted_type_parity.rs`'s cross-language fixture test, `dump_rooted_type_sweep.rs`) needed no changes.

## What this does NOT do (named, not "out of scope")

- `rust/core/src/legacy_entities.rs`'s base-type mapping (26 arms) — unchanged. Checked: `has_geometry_by_name`/`is_representationless_spatial_container_by_name` in `rust/core/src/schema_helpers.rs`, `legacy_aware_ifc_type`, and `is_rooted_type`'s first branch all still read it exactly as before.
- `IfcType::from_str` / `Unknown(u32)` — unchanged. A legacy name still resolves `Unknown` from the generated schema; `LEGACY_ROOTED_TYPES` and `legacy_aware_ifc_type` are the two things that already compensate for that, same as before this PR.
- Deleting either hand-kept table (issue's part (b)) — not attempted; per #4495's scoping, that's gated on generating the type/attribute universe first, which is still incomplete (`legacy_entities.rs`'s base-type mapping isn't generated yet).

## Testing

RED/GREEN (`rust/export/src/rooted_type.rs`, new test `a_table_only_legacy_name_is_rooted_only_through_the_generated_table`):
- All of this module's existing per-name tests (`IFCDOORSTYLE`, `IFCPROXY`, the stratum leaves) happen to also be mapped by `legacy_entities.rs`, so `is_rooted_type` stays `true` for them via `legacy_aware_ifc_type` even with an entry deleted from `LEGACY_ROOTED_TYPES` — they would NOT have caught a regression in the generated table. The new test targets `IFCSCHEDULETIMECONTROL`/`IFCRELASSIGNSTASKS`, confirmed absent from `legacy_entities.rs` (`ifc_lite_core::is_legacy_entity` returns `false` for both), so `is_rooted_type` can only answer `true` for them through this table.
- RED: deleted `"IFCSCHEDULETIMECONTROL",` from the generated file by hand — `a_table_only_legacy_name_is_rooted_only_through_the_generated_table` failed with `assertion failed: is_rooted_type("IFCSCHEDULETIMECONTROL")`, proving the mutated line ran.
- GREEN: reverted (confirmed byte-identical to the pre-mutation file), `node scripts/generate-legacy-rooted-types.mjs --check` passes (54 rooted legacy entities, 0 conflicts).
- Case-insensitivity, including CamelCase: extended `type_name_matching_is_case_insensitive` with `IfcProxy` and `IfcBuildingElement`; the new test also checks `ifcscheduletimecontrol` (lowercase) and `IfcRelAssignsTasks` (CamelCase) against the table-only name.

Both directions: `cargo test -p ifc-lite-core -p ifc-lite-export` — 315 + 453 tests (plus integration/doc-test binaries), 0 failures, including `rooted_type_parity.rs`'s `fixture_covers_the_whole_type_universe` (the cross-language sweep against `packages/export/src/rooted-type-sweep.parity.test.ts`'s shared fixture) and `schema_upconvert_parity.rs`, both unaffected. `cargo clippy -p ifc-lite-core -p ifc-lite-export --all-targets -- -D warnings` — clean.

Gates: `node scripts/check-rust-api-surface.mjs` — PASS, 433 tracked items, no new/changed entries (the public path and type of `LEGACY_ROOTED_TYPES` are unchanged — a `pub use` re-export of the same `&[&str]`, not a signature change). `cargo test -p ifc-lite-processing --test module_size_ratchet` — PASS, no new allowlist row needed (`rust/export/src/generated/legacy_rooted_types.rs` lives under `/generated/`, which this ratchet already excludes). Plain `node scripts/check-module-size.mjs` — locally reports pre-existing merge-base drift in files this PR does not touch (`classify.rs`, `boolean/mod.rs`, `voids/*.rs`, `prepass.rs`, `processor/mod.rs`, `zero_copy/*.rs`); confirmed against a clean checkout of unmodified `upstream/main` that this drift is pre-existing and unrelated — this environment's `origin` remote is the BIMvoice fork, stale at a commit from before today's `main`, so the script's merge-base audit compares against the wrong base locally. `node scripts/check-generated.mjs`'s `check:coverage-ledger` FAIL is also pre-existing and reproduces identically on unmodified `upstream/main`.

Changeset: `.changeset/legacy-rooted-types-generated.md` (`@ifc-lite/wasm`, `@ifc-lite/server-bin`, both consumers of `ifc-lite-export`).

## Summary by CodeRabbit

* **Refactor**
  * Legacy rooted-type information is now generated from schema data, with no change to existing behavior.

* **Tests**
  * Added validation to detect stale generated rooted-type data.
  * Added coverage for case-insensitive matching and legacy rooted names.

* **Chores**
  * Added generation and freshness-check tooling for legacy rooted types.


louistrue (dismissed):
Approved at exact head `c3dd8e6ef81133c94022901d9ea59584ebf02bb1`. Independent review confirmed the schema-derived legacy-root universe, version handling, module split, and production-observing revert oracle. Exact-head Ollama and Claude reviews are clean; all substantive CI shards passed. The only rerun still active is GitHub's production-observation job, after its first attempt was killed solely by the 20-minute job limit; I will enqueue only after that rerun succeeds.

louistrue (approved):
Approved at exact head `3d6ed1d140bcb88e487aa7941c9dfedc9522308a`. The schema-derived legacy-root set, per-version union, generated Rust artifact, and freshness wiring are sound. The latest oracle optimization retains a real production-wiring revert witness while isolating the named gate, bringing it from the 20-minute timeout to ~0.15s; exact-head Ollama is clean.

Approval does not enqueue the PR: full exact-head CI and the scheduled Claude/review-posted retry must still finish green.

Issue #3913, opened 2026-09-04
union_many still tears on 88 of 441 swept configurations after the N-ary weld


BIMvoice:
After the pair-union weld (#3874) and the N-ary weld (#3912), `union_many` still tears on a substantial minority of overlapping-and-rotated configurations. Filing the residual separately so it is not lost inside #3353, whose pair-union half is now fixed.

## Measured

A local sweep of **441 configurations** — a 7×7 corner grid × 3 `dz` values × 3 operand orderings — over the N-ary union path:

| build | torn / 441 |
|---|---|
| `main` (before the N-ary weld) | 294 |
| naive weld, no exact-plane guard | 88 |
| weld + guard (#3912 as merged shape) | 88 |

So the weld removes roughly 70% of the tearing and **88 configurations still tear**. That is the residual this issue is about.

Two honest caveats about the numbers:

- **The sweep harness is local and uncommitted**, written to answer this question. It is not the harness behind #3874's pairwise figures (2090→0 forward, 2388→0 reversed); those were produced differently, and #3874's three-box "105 of 147" / "36 of 147" figures were prose-only with no committed harness either. So these numbers are internally consistent but not directly comparable to the pairwise ones.
- **The exact-plane guard makes no difference on this sweep** (88 either way). Its value is specific to the `#960` shape — operands with pre-existing *exact* shared seams — where removing it reproduces a concrete failure (`#4148 ... max Z = 9850 mm, expected ~8984 mm`). The sweep's synthetic configurations mostly lack such seams, so it cannot see that benefit. The guard is justified by that case, not by these counts.

## Why this is a distinct defect, not more of the same

The mechanism behind the N-ary failure is now understood: `promote_cutter_verts_onto_host_faces` skips a vertex's own exact-plane match (`d == 0.0 { continue }`), which is safe against a single host but, pooled across N operands in `promote_operands_mutually`, lets a near-parallel plane from an **unrelated, non-adjacent** operand win the candidate search — destroying already-exact shared seams rather than reconciling noisy ones. Instrumented on the real `#960` fixture, bit-identical shared-vertex-pair counts fell `(148,4,12,248)` → `(88,4,5,159)` under the naive weld.

That mechanism is addressed. The remaining 88 are therefore **something else**, and identifying what would be the next real step — not another threshold sweep, which is an axis already exhausted on #3353 (detector unification, narrower gate, centroid exactness, degeneracy guard, interner widening, provenance — all measured, none sufficient).

[...]

Issue #3914, opened 2026-09-04
Consolidation-level tear on near-degenerate rotated overlaps


BIMvoice:
## Summary

The  stage produces non-manifold geometry from a watertight kernel output on overlapping, rotated operand pairs. Two built-in coincident-face detectors disagree at near-degenerate configurations, leaving a redundant triangle on each operand and a shared edge used by three kept triangles instead of two.

## What breaks

Union of two closed, overlapping, rotated boxes produces a non-manifold shell. Input operands are closed, output is non-manifold (unmatched directed edges after welding by position).

Test pinned in `rust/geometry/tests/issue_3353_boolean_tear.rs` at line 148:
```rust
fn a_rotated_overlapping_union_stays_manifold() {
    let a_min = [-1.064427873716452, -1.5758991032070164, -2.335934512221897];
    let a_size = [1.7582981472721038, 3.7437572581011054, 2.5580013636220693];
    let b_min = [-0.468184233137136, -1.8002926870781526, -1.0101295786317475];
    let b_size = [1.9952974802528327, 3.404965797097641, 3.2964264954246745];
    let theta = 1.3158416849982029;
    // union of boxed(a_min, a_size) and rotated_boxed(b_min, b_size, theta)
}
```

The fixture is a local seeded sweep (seed 6 of a splitmix64-seeded run), independent of and smaller than the original 60000-pair sweep in #3353.

## Where it lives

`rust/geometry/src/kernel/arrangement/classify.rs` — the exact arrangement's classification step. Two structurally different coincident-face tests:

- Line 209: `fn c_on_or_near_a` — tests whether a triangle from operand B is a coincident duplicate of a face on A (using BVH-accelerated ray-cast)
- Line 381: `BComponents::surface_normal` method — tests whether a triangle from operand A is a coincident duplicate of a face on B (using AABB-near and snap-band tests)

At a near-degenerate rotated overlap the two tests can disagree on the same coincident pair.

## Why it is distinct from #3874 and #3912

PR #3874 fixed the **near-coplanar half** of #3353: two operand faces landing a few `SNAP_GRID` steps apart because `mesh_bridge` snaps per axis. The fix welds operands onto shared planes before the arrangement.

[...]

PR #4678, opened 2026-09-13
fix(geometry): retry a torn 3-operand union under the caller's other orderings

### LINKED ISSUE #3917: N-ary union weld is order-dependent: 32 of 441 configurations tear, but zero under one operand ordering
`promote_operands_mutually` leaves tearing that depends on the order operands are supplied in. One ordering reaches zero torn; the others do not.

Filing this so the defect stays tracked after #3916 lands — that PR commits a sweep harness and characterises the residual, and closes #3913, but it does not fix the tearing.

## Measured

Using the committed harness from #3916 — 441 configurations: a 7×7 corner-offset grid × 3 `dz` values (`0`, `±SNAP_GRID`) × 3 operand orderings, over the repo's own pinned N-ary fixture:

**On `main`, before the N-ary weld (#3912):**

```
249 / 441 torn
ordering ABC  83
ordering CBA  83
ordering BCA  83
```

Ordering is **inert** — identical counts. Every torn configuration has `dz != 0`; flush operands never tear.

**With the N-ary weld (#3912) applied:**

```
32 / 441 torn
ordering ABC  20
ordering CBA  12
ordering BCA   0
```

Ordering is now **load-bearing**, and one ordering is clean.

## Why that is the interesting part

A defect that disappears entirely under one permutation of its inputs is not a tolerance problem — no epsilon is sensitive to argument order. It points at `promote_operands_mutually` walking operands **by index**: whichever operand is welded first becomes the reference the others conform to, so a chain that converges in one order does not in another.

This is worth stating because every prior axis on this family was a threshold sweep, and all six were measured and exhausted. Its two sibling defects were structural as well — #3914 is a `round()` boundary straddle in plane bucketing (whose grid is deliberately fine, since coarsening it regresses #1007 by bridging a real opening), and #3915 is symbolic Vid-space duplication that exists before any float comparison runs.

## Directions worth trying, in rough order

1. **Make the weld order-independent.** Iterate `promote_operands_mutually` to a fixed point over all pairs rather than a single indexed pass, so the result does not depend on which operand is seen first. #3912 already iterates to a fixed point *within* a pass; the question is whether the pass order itself still biases the outcome.
2. **Impose a canonical order** — sort operands by a stable geometric key before welding, so the same set always welds the same way regardless of how the caller supplied it. Cheaper than (1), and it makes results reproducible even if it does not reach zero.
3. **Work out why BCA specifically succeeds.** With only three orderings sampled, this could be a property of that permutation or an artefact of the fixture. **Widening the sweep to all six permutations is the cheap first step**, and the harness in #3916 already has the structure for it.

(3) is the obvious place to start, since it costs almost nothing and decides whether (1) and (2) are aimed at the right thing.

## Not addressed here

- #3914 — the consolidation-level tear (plane-bucket boundary straddle).
- #3915 — `sweep_261`'s in-plane vertex split (Vid-space duplication).

Both are diagnosed and awaiting a design decision; neither is the same mechanism as this.

BIMvoice:
## Summary

`promote_operands_mutually` walks operands in ARRAY order, and which operand
a caller places at index 0 can decide which pair of planes reconcile before
a third operand is touched. The committed #3913/#3916 sweep (7x7 corner grid
x 3 `dz` x 6 orderings, 882 configurations over 3 operands) measured 136
torn, all caller-order-dependent — every torn configuration had a closed
result reachable under some *other* ordering of the same 3 physical
operands (`BCA` reached 0/147 on the discriminating fixture while `CAB`
reached 53 unmatched directed edges on the identical operand set).

Issue #3917's own investigation (four rounds, all measured, linked on the
issue) ruled out every static ordering rule: decoupling the weld from array
order regresses the one ordering that already works, and every geometric
sort key tried (position, centroid, AABB-overlap-volume, degree, distance
from centroid) either fails to reproduce the good sequence or trades away
its quality to reach determinism. No single scalar signal, computed before
the weld runs, predicts which order will close.

## Fix

Rather than trying to predict the right order in advance, `union_many` now
recognizes a bad one after the fact, the same way #3440's prism-cut
self-check already does in production (`directed_closed`, the DIRECTED
quantized closed-surface audit), and retries:

- Only for a **3-operand** union — the only N this defect family has
  measured — and only up to a small triangle-count cap, so a large real
  N-ary cutter union is completely unaffected.
- The check runs on the SAME `consolidate_coplanar(union_many(..))`
  combination every real caller already applies (not the raw kernel output —
  #3914's investigation on this exact code path found a raw union that was
  already closed and had `consolidate_coplanar` introduce a tear from a
  different mechanism, so checking pre-consolidation would have been
  meaningless here).
- If the caller's own order already comes back closed, **nothing changes** —
  this is the safety property that keeps the change low-risk: every one of
  the 746 already-good configurations in the sweep is untouched.
- If it tears, the other 5 orderings of the same physical operand set are
  tried (in a fixed sequence derived from the caller's own array, so the
  full search space is identical regardless of which order the caller
  happened to supply) and the first one that comes back closed is returned.
- If none of the 6 orderings close, the original torn result is returned
  unchanged — this fix does not manufacture closure where none of the 6
  orderings ever reaches it.

## Measured

```
before (main):  136 / 882 torn   ABC=20 ACB=34 BAC=34 BCA=0  CAB=36 CBA=12
after (this PR):  0 / 882 torn   ABC=0  ACB=0  BAC=0  BCA=0  CAB=0  CBA=0
```

`cargo test -p ifc-lite-geometry --lib issue_3913_sweep_tests -- --nocapture`
now reports 0/882 under every ordering; `KNOWN_TORN_CEILING` is tightened
from 136 to 0 and the assertion changed from a ceiling to an exact match so
a future regression is caught immediately.

A dedicated regression test pins the issue's own traced discriminating
fixture directly (`bx=0.5, by=0.5, dz=+1 SNAP_GRID`, `CAB` order, 53
unmatched directed edges before this fix) and additionally checks all 6
orderings of that same physical operand set close.

## What this does NOT claim

This eliminates tearing, not byte-level determinism: two callers supplying
the same 3 physical operands in two DIFFERENT orders that both already
close will still get their own order's own closed result, which can differ
by a snap step's worth of coordinates from another closed ordering's result
(the same "closed, same volume to within a snap step" bar the codebase
already treats as commutative for pairwise welds — see `plane_weld.rs`'s own
doc on `promote_operands_mutually`). What's fixed is the actual observable
defect: no ordering of a 3-operand union tears anymore, where 136 of 882
measured configurations did before.

## Testing

- RED: reverting the retry (`if true { return out; }` at the top of
  `arrange`) reproduces the exact pre-fix failure (53 unmatched directed
  edges on the pinned fixture) — confirms the new test detects the mutation
  and the retry logic is load-bearing.
- GREEN: `cargo test -p ifc-lite-geometry` — 951 passed, 0 failed, 1 ignored
  (unrelated heavy-fixture opt-in).
- `cargo test -p ifc-lite-geometry` (all targets, including the
  `watertightness_census` / `triangulation_invariance` golden suites) — all
  green, no regressions.
- `cargo clippy -p ifc-lite-geometry --all-targets -- -D warnings` — clean.
- `node scripts/check-module-size.mjs` — pre-existing failure reproduced
  identically on a pristine `upstream/main` checkout (unrelated allowlist
  drift in files this PR does not touch); not introduced by this change.

Refs #3913, #3914, #3916
Closes #3917

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed three-part geometry unions that could produce open or inconsistent results depending on operand order.
  * Alternate operand orderings are now checked when needed to produce a closed result.
  * Two-part and unions involving four or more operands remain unchanged.

* **Tests**
  * Added regression coverage confirming all supported orderings of affected three-part unions produce closed geometry.
  * Strengthened sweep validation to require all tested configurations to close successfully.


louistrue (changes requested):
Blocking budget regression in rust/geometry/src/kernel/nary_union.rs: `arrange_once` calls `budget::begin()`, so every retry resets the per-boolean exact-predicate counter. A single public `union_many` call can therefore spend the configured cap once per permutation (up to 6x), and an exhausted first attempt is silently retried after clearing its trip state. This contradicts `budget::begin`'s documented once-per-public-boolean contract and weakens the deterministic hang guard. Initialize the budget once in `arrange`, before the first candidate, and let all `arrange_once` attempts share that operation counter; if it trips, preserve the original failure/fallback behavior rather than restarting the allowance.

louistrue (approved):
Approved at exact head `cc494b423fa5652d4519b4d4251c2323ae990f82`. The public union now owns one shared predicate budget across all retry orderings, the regression observes that through the pre-existing public path without global test-state races, and the production-revert oracle fails with a final-candidate count of 368 versus the accumulated >=1,000 witness. Focused #3917 tests pass under the normal parallel runner; exact-head Ollama reports no major issue.

This approval clears the addressed code-review request only. Do not enqueue until the full exact-head Rust/test matrix and Claude/review-posted lane are green.

Issue #4148, opened 2026-09-08
cli: extract-entities' forwardClosure reads an #id out of a STEP string literal, over-extracting on valid input


louistrue:
`forwardClosure` in `packages/cli/src/commands/extract-entities.ts` scans a record's raw
body with `/#(\d+)/g`, which reads an `#id` out of a STEP string literal.

Measured on the current code:

    #70= IFCFURNISHINGELEMENT(...,'Chair pairs with #71',...)

Select `{70}`. `keep.has(71)` is `true`, and the storey containment is rewritten to
`(#70,#71)`. The unselected chair rides into the subset because another product's Name
mentions it, and the extraction is silently larger than asked for.

## Why this is worth a separate issue

This is the same hazard `refsOutsideStrings` was written to avoid in
`subset-relations.ts`, which is string-aware precisely because a `#` in a Name is TEXT.
The two functions now disagree about what a reference is, inside one command:
`relationLine` will not read `#71` out of a Name, and `forwardClosure` will.

It is also the shape of the #4128 fixture (`'C1 see #999'`). That case only escapes the
symptom because `#999` is undefined, so the #4128 fix drops it. When the lookalike id
DOES exist, as above, nothing catches it.

It bites on VALID input. No malformed STEP is required, only an exporter that writes a
`#` in a name, which Revit does.

## Not a regression

Identical on `bea417a49`, before the #4128 and #4126 fixes. Those changed which ids are
admitted after the scan, not how the scan reads text.

## Shape of the fix

`refsOutsideStrings` already exists in `subset-relations.ts` and does exactly the right
walk. The awkward part is the dependency direction: `extract-entities.ts` imports from
`subset-relations.ts` deliberately, one way, so the scanner would move rather than be
imported backwards. That is the same move `StepRecord` already made.

Worth checking at the same time whether the over-extraction is observable in the
`--product` instance counts anyone has relied on.

Found by code review on the #4128 / #4126 PR.

BIMvoice:
Taking this one. Plan: export `refsOutsideStrings` from `subset-relations.ts` and use it in `forwardClosure` (`extract-entities.ts`) instead of the raw `/#(\d+)/g` scan, so a `#id`-shaped substring inside a STEP string literal (e.g. a Name) is no longer treated as a reference. Will report back with a PR.

PR #3997, opened 2026-09-06
docs(geometry): correct build_cutter_union's watertight claim to the measured contract

BIMvoice:
## Measured numbers (the evidence for the comment change)

`build_cutter_union`'s comments promised a *watertight* CSG union of the roof cutter-clip chain (issue #3980, following the late #3921 review). Audited that claim against the real regression fixture the code's own comments cite by wall number: `960_house_segmented_roof_clip.ifc` (walls #2152, #4374, #4148, #2797, #5904).

Two closure predicates, both temporary/audit-only and not shipped:
- **Exact bit-identical closure**: quantize to raw `f32::to_bits()`, directed-edge balance must cancel exactly, plus a degenerate (zero-length) edge check.
- **T-junction-sensitive open-boundary-edge count** at two grids: 1mm (matches the codebase's existing `open_boundary_edges` regression helpers) and 0.1mm (mirrors `csg::consolidate::conform::count_open_boundary_edges_at`, which the codebase itself documents as needed because the 1mm grid hides a T-junction).

Instrumented at the three points that matter: the union `build_cutter_union` returns (`combined`), the `#3925` repair-candidate union (`repaired` — never triggered on this fixture), and the actual accepted subtraction (`clipped`).

Real output on the 5 real walls (array order: #2152, #4374, #4148, #2797, #5904):

```
AUDIT3980 primary/fallback-union(combined): tris=215 closed_exact=false degenerate=false open@1mm=2 open@0.1mm(Tjunction)=3
AUDIT3980 accepted-subtract(clipped):       tris=118 closed_exact=false degenerate=false open@1mm=4 open@0.1mm(Tjunction)=11

AUDIT3980 primary/fallback-union(combined): tris=62  closed_exact=true  degenerate=false open@1mm=0 open@0.1mm(Tjunction)=0
AUDIT3980 accepted-subtract(clipped):       tris=32  closed_exact=true  degenerate=false open@1mm=0 open@0.1mm(Tjunction)=0

AUDIT3980 primary/fallback-union(combined): tris=214 closed_exact=false degenerate=true  open@1mm=12 open@0.1mm(Tjunction)=12
AUDIT3980 accepted-subtract(clipped):       tris=73  closed_exact=false degenerate=true  open@1mm=2 open@0.1mm(Tjunction)=5

AUDIT3980 primary/fallback-union(combined): tris=233 closed_exact=false degenerate=true  open@1mm=2 open@0.1mm(Tjunction)=2
[...]

Issue #3980, opened 2026-09-05
geometry: verify roof cutter-union closure contract against actual subtraction


louistrue:
A late #3921 review notes that `build_cutter_union` promises a watertight result in comments but currently accepts any nonempty primary/fallback mesh. Nonemptiness does not prove closure. This is a valid contract-audit concern; no new failing real-model fixture was supplied by the review.

Do not add a broad exact-closure rejection as an unmeasured release patch. The caller already evaluates each real per-cutter subtraction, enforces their intersected bounds on the final subtraction, detects accept-gate rejections (BIMvoice #3922/#3923), and bounds coordinate-moving repair by clean individual removals. Current four-feature roof tests pin actual clipped bounds; full/heavy census and eleven-model native/browser comparisons qualify the current policy. Those tests are evidence for their fixtures, not a universal proof for open cutters.

Bounded follow-up: find a real roof-chain case that produces an open primary or fallback union and an incorrect accepted subtraction; preserve original input and RTC, evaluate exact closure predicates including T-junction semantics, and compare candidate rejection/sequential fallback against the actual subtraction and independent oracle. Correct the comments to the measured contract. Any policy change must preserve roof #2152/#5904, existing shared-corner protection, full/heavy census, feature gates, native/WASM parity and worker-pool performance.

Coordinate with BIMvoice #3919/#3923 and #3914/#3915 before claiming work. Keep this separate from #3977 low-detail architecture wall boundaries until a common cause is demonstrated.

louistrue:
The maintainer requested tackling the remaining post-shipping items. I am preparing a supplementary independent accepted-subtraction audit for the public #960 fixture, focused on walls 2152/4148 with the three closed controls. This complements the comment correction in #3997: it will compare final surfaces/material witnesses and distinguish edge incidence from T-junction-aware closure, rather than treating the current Z-bounds check as proof of correct subtraction. Existing ownership and the adjacent #4024 policy work remain unchanged. No geometry-policy change or snapshot repin is part of this first audit; source/reference versions and any negative result will be retained.

louistrue:
Supplementary independent surface check, coordinated with the existing owner. This does not replace #3997's comment audit or change #4024's acceptance/fallback work.

[...]

PR #4679, opened 2026-09-13
test(geometry): verify roof cutter-union closure contract against actual subtraction

BIMvoice:
Related to #3980 (not closing — see below).

## What this measures

#3997 audited `build_cutter_union`'s closure contract on the real #960 fixture (walls #2152, #4374, #4148, #2797, #5904) with an open-boundary-EDGE count, finding open edges on 4 of 5 real chains and open edges on the accepted subtraction for 2 of 5. That predicate cannot distinguish a genuine hole (missing material) from the hairline T-junction openness `csg::consolidate::conform` already forgives by design — so it left open whether those open edges mean the accepted subtraction is actually wrong.

This adds a reference-independent discriminator, following the prior art from #4648: the boundary-area VECTOR `A = Σ (p1−p0)×(p2−p0)` over a triangle list, from the identity `6·V(o) = 6·V(0) − o·A`. By Stokes' theorem this sum depends only on the mesh's boundary loop(s) — it is exactly the zero vector for a closed mesh AND for a hairline T-junction chain (the loop encloses no area), and nonzero only for a boundary loop that bounds a real hole (twice that hole's own area vector).

## Result — measured, not assumed

Measured on the REAL per-wall output for all five #960 walls (`rust/geometry/tests/issue_960_segmented_roof_clip.rs::segmented_roof_walls_boundary_area_vector_is_negligible`):

- **Default (production) configuration**: every wall's `|A|` fraction of its own surface area is at most ~3e-8 — no genuine hole on any of the five real chains. This confirms the open edges #3997 found are hairline, not incorrect material removal, for the shipped configuration.
- **`csg_topology_gate` alone** (a CI-only, off-by-default-in-production feature — see its Cargo.toml doc comment): wall #2152's accept-gate rejects the unioned-cutter subtract and defers to the sequential path, which hits the `#635` AABB fallback for a full-cross-section cutter. That fallback's own result carries a real, small hole (~63 mm² on a ~39 m² wall, fraction ~1.63e-6) — genuine, not hairline, though far too small for the existing 25 mm Z-bounds pin to see. This does NOT reproduce with `csg_manifold_gate` alone or with both gates together (both ~4e-9, same as default). Pinned as a measured, gate-specific exception (documented in the test) rather than silently widened default tolerance, since this gate combination never ships.

## Non-vacuousness

Three supporting tests prove the measure (and the gate) actually work:
- `boundary_area_vector_discriminates_hole_from_hairline`: a synthetic cube reads exactly `A = 0` closed, the exact expected nonzero vector for a dropped face (genuine hole), and `~0` for a folded out-and-back degenerate triangle pair (hairline) — proving the discriminator can't collapse to "always zero."
- `dropping_one_triangle_from_the_real_wall_mesh_reddens_the_gate`: drops one real triangle from wall #4374's actual shipped mesh and confirms the exact assertion the production test runs would catch it.

All four `cargo test -p ifc-lite-geometry --test issue_960_segmented_roof_clip` feature combinations (none, `csg_manifold_gate`, `csg_topology_gate`, both) pass; `cargo test -p ifc-lite-geometry --lib` (951 tests) is unaffected. No production code changed.

## Not closing #3980

This is a verification-only contribution measuring the contract against the actual accepted subtraction, as the issue asked. It found no genuine-hole defect in the default/shipped configuration, and one small, already-fallback-path, gate-only exception that is now pinned and measured rather than silently accepted. Per the issue's own last comment, closing is the maintainer's call — leaving this open for that decision, and available follow-up if the `csg_topology_gate` fallback path itself needs tightening.

## Coordination
- No overlap with #3917's branch (`fix-3917-union-weld-order`) — this touches only `rust/geometry/tests/issue_960_segmented_roof_clip.rs`.
- No production code touched; does not contradict or duplicate #4648's `prism_cut.rs` closure-audit work — this reuses its `6·V(o) = 6·V(0) − o·A` identity as prior art, applied to the roof cutter-union's real output instead.


louistrue (changes requested):
The central inference is mathematically too strong. `Σ (p1−p0)×(p2−p0)` is the **net oriented** boundary-area vector. `A != 0` proves some boundary area, but `A == 0` does not prove the absence of genuine holes: two real boundary loops with equal/opposite area vectors cancel (for example, remove equal faces from opposite sides of the cube), and even a single non-planar/self-cancelling loop can have zero vector area. The synthetic test covers one dropped face only, so it cannot detect this false-negative class. Consequently the PR cannot claim that the five meshes have no genuine hole or that this verifies the closure contract. Add an explicit opposite-hole counterexample and use a non-cancelling per-boundary-component measure/topology walk (or narrow all names/docs/assertions to the weaker net-vector invariant and avoid the closure conclusion).

louistrue (approved):
The current head `a356559c3` resolves my blocker. It now explicitly narrows the real-fixture assertion to the net boundary-area signal it actually establishes and adds a constructive counterexample showing two genuine opposing holes can cancel. That makes the test useful without overstating it as an unconditional watertightness proof.

Independent diff review, exact-head Ollama, Rust, Node, viewer shards, E2E, provenance, and changed-test observation are green. Approving the code; the PR must remain out of the merge queue until the required Claude/review-posted lane is successfully rerun.

Issue #3987, opened 2026-09-05
perf(geometry): remove repeated unused processor setup during cold loads


louistrue:
A full-call native cold-load profile on production-source commit `6094e2f16` attributes about 13.4% of aggregate geometry-worker CPU to `GeometryRouter::new`. The native pipeline constructs one router per product. Each constructor eagerly builds the full processor registry, allocates boxed/Arc processors, and constructs/clones their schema maps, even when that product uses only one representation family. This is sampled CPU attribution, not a measured speedup.

Scope: remove unnecessary built-in processor setup and dispatch allocation while preserving public `register()` replacement behavior, independently owned failure logs, custom schema behavior, tessellation, small-cut options, and canonical element production. Start with a bounded lazy-initialization experiment; retain it only if full-load measurements justify it. No geometry algorithm or accept-gate changes.

Acceptance: five interleaved fresh-process native base/candidate pairs on representative real exporters/sizes, full-call timing and process peak memory, unchanged ordered geometry fingerprints and diagnostic behavior. Qualify shared Rust in actual Chrome and Firefox worker pools with matching runtime artifacts, complete metadata/renderer readiness, properties, hierarchy, GPU picking, sections and federation. Keep browser and native verdicts separate: WASM builds an outer router per batch, so native constructor attribution alone predicts no browser gain. Preserve failures and reject regressions. Run the full required Rust/TS/WASM/CI gates and record the measured verdict in the performance ledger.

Session: `4041cf08`, branch `t3code/improve-wasm-native-cold-load`. This bounded issue covers processor setup overhead; metadata transport remains #3985 and the reusable browser harness remains #3978 / #3982.

PR #4724, opened 2026-09-13
fix(processing): detect a lowercase/mixed-case FILE_SCHEMA header case-insensitively

### LINKED ISSUE #4661: A lowercase FILE_SCHEMA header yields the wrong schema (IFC2X3), silently disabling appearance authoring
`detect_schema_version` searches the file for the byte literals `b"IFC4"` / `b"IFC4X3"` case-sensitively. A file whose header is written in lowercase — legal under ISO 10303-21, which makes keywords and identifiers case-insensitive — is not merely undetected: it is reported as **`IFC2X3`**, the default fallback. The viewer then believes an IFC4 model is IFC2X3.

## Reproduced

Two byte-identical fixtures differing only in header case:

| header | `detect_schema_version` returns |
|---|---|
| `ISO-10303-21;HEADER;…FILE_SCHEMA(('IFC4'));` | `"IFC4"` |
| `iso-10303-21;header;…file_schema(('ifc4'));` | **`"IFC2X3"`** |

## Wrong answer, not absent answer

`rust/processing/src/processor/schema_detection.rs` has no `Option` return — on no case-exact match it defaults to `"IFC2X3"`. That value flows unconditionally into `ModelMetadata.schema_version` and `QuickMetadataBootstrap.schema_version`.

Downstream it is not informational. `"IFC2X3"` is itself a member of `VALID_SCHEMA_VERSIONS` in `apps/viewer/src/utils/serverDataModel.ts`, so it passes validation with no warning branch — and it then gates behaviour: `useIfcAuthoringTarget.ts` tests `schemaVersion.startsWith('IFC4')` and `useAppearancePanel.ts` blocks appearance authoring for `'IFC2X3'`. **A lowercase-header IFC4 file is silently excluded from appearance authoring.**

No second layer of exposure: the function only ever returns one of three canonical uppercase constants, never raw header text, so every downstream comparison is against a canonical value.

## The existing test asserts the bug as correct

`rust/processing/src/processor/schema_detection_tests.rs` pins `(b"ifc4x3", "IFC2X3")` as expected and passing. The lowercase case is not untested — it is tested, and the wrong answer is the assertion. Worth knowing before anyone reads a green suite here as evidence the behaviour is intended.

## Reachability

Checked all 367 `.ifc` files in the corpus: every one starts with the canonical-case `ISO-10303-21` magic, and none carries `file_schema` / `File_Schema` in its first 500 bytes. So: a spec-conformance gap with no observed real-world trigger in this corpus — not actively breaking known files. Severity should be judged on that basis.

## Relevance to #4497

This is the byte-prefilter shape of that class. It never calls `next_entity()`, so neither a `StepKeyword` newtype nor normalising at the scanner boundary would reach it — it needs a case-insensitive byte search of its own. The lead-byte technique now used by `find_ifcproject_keyword` (merged in #4498) applies directly.

Note the detection is deliberately a raw-byte, anywhere-in-file predicate for performance (#3987); a fix should preserve that property rather than switching to header parsing.

Not checked: consumers of `schema_version` outside `apps/viewer` (server, CLI, anything reading the metadata JSON), whether a separate browser/WASM schema-detection path exists, and IFC5/`.ifcx` handling.

BIMvoice:
## Summary

`detect_schema_version` (`rust/processing/src/processor/schema_detection.rs`) searched the file for the byte literals `b"IFC4"` / `b"IFC4X3"` case-sensitively. ISO 10303-21 makes STEP keywords case-insensitive, so `FILE_SCHEMA(('ifc4'))` is legal input — and a case-sensitive miss did not just go undetected, it fell through to the `"IFC2X3"` default fallback. That value passes `VALID_SCHEMA_VERSIONS` validation with no warning and then gates behaviour downstream: `useIfcAuthoringTarget.ts` tests `schemaVersion.startsWith('IFC4')` and `useAppearancePanel.ts` blocks appearance authoring for `'IFC2X3'`. A lowercase-header IFC4 model was silently excluded from appearance authoring.

## Fix

Reuses the shared case-insensitive byte search `ifc_lite_core::parser::find_keyword` instead of hand-rolling a new one — it already anchors on a rare needle letter (not the first one) and verifies with `eq_ignore_ascii_case`, the same technique `find_ifcproject_keyword` uses (#4498). For the `"IFC4"` needle it anchors on `F` (rank 20/26 rarest) rather than the leading `I` (rank 24/26), since every IFC keyword and every GUID in the file is dense in `I`.

The raw-byte, anywhere-in-file predicate and the "IFC4X3 anywhere wins over IFC4" precedence (#3987) are unchanged — this is not a header parser, it is the same substring predicate made case-insensitive.

## Performance

Measured on a synthetic 20/60/120 MB IFC-shaped fixture (dense with `I`-heavy GUIDs and IFC keywords, no `IFC4`/`IFC4X3` literal present anywhere — the worst case, requiring a full scan) via `cargo run -p ifc-lite-processing --release --example schema_detection_bench_4661`:

| | 20 MB | 60 MB | 120 MB |
|---|---|---|---|
| case-sensitive `memmem` (old) | 0.70 ms | 1.32 ms | 2.64 ms |
| naive `I`-anchored (worst available anchor) | 27.5 ms | 82.5 ms | 164.8 ms |
| `find_keyword`, `F`-anchored (this PR) | 1.83 ms | 5.54 ms | 10.83 ms |

`F`-anchoring costs 2.6-4x the old case-sensitive `memmem` (expected — it's a byte-anchor-then-verify scan, not SIMD substring search) but is ~15x faster than the worst available anchor, consistent with the #4498 finding that anchor choice, not the technique, dominates. The bench example is ad hoc (not wired into CI), matching the existing `rust/processing/examples/perf_probe.rs` convention.

## The changed test assertion

`rust/processing/src/processor/schema_detection_tests.rs` pinned `(b"ifc4x3", "IFC2X3")` as an expected-and-passing case. **That assertion encoded the defect, not protected behaviour** — it is removed from that list and replaced by two new tests:

- `schema_search_is_case_insensitive_4661`: mixed-case headers (`Ifc4`, `IFC4x3`, `ifc4x3`, `Ifc4X3`), a genuine `IFC2X3` file (no `IFC4` literal at all, in any case) still detecting as `IFC2X3` — so a fix that just returns "everything is IFC4" cannot pass — and the `IFC4X3`/`IFC4` prefix relationship holding under case folding.
- `uppercase_and_lowercase_headers_detect_the_same_schema_4661`: two fixtures byte-identical apart from header case, both detecting `IFC4X3`. This is the RED-first repro: it fails before the fix (lowercase fixture returns `IFC2X3`, uppercase returns `IFC4X3`) and passes after.

RED confirmed before the fix; GREEN after. Also mutated the fix back to the case-sensitive `memmem` search (with a `MUTATION-PROBE` `eprintln!` proving the mutated line actually ran) and confirmed both new tests redden.

`cargo test -p ifc-lite-processing`: 411 lib tests before this change, 413 after (net +2). Full suite green; `cargo clippy -p ifc-lite-processing --all-targets -- -D warnings` clean.

## Detected vs. fell-back-to IFC2X3

Checked: the single call site (`processor/mod.rs`) stores the `&'static str` result unconditionally into `ModelMetadata.schema_version`/`QuickMetadataBootstrap.schema_version`; no consumer today distinguishes "genuinely detected IFC2X3" from "found nothing and fell back to IFC2X3". Not widening the public return to `Option<&'static str>` here — per prior guidance, that's a design decision for a separate PR if a consumer ever needs the distinction, not something to fold into this fix.

Checked: no separate schema-detection code path for WASM/browser or the server — `detect_schema_version` is the one implementation, reached by both through the shared `processor::mod.rs`.

Closes #4661

## Summary by CodeRabbit

* **Bug Fixes**
  * IFC4 and IFC4X3 models with lowercase or mixed-case schema headers are now detected correctly.
  * These models can now pass schema-dependent checks, including appearance authoring eligibility.

* **Tests**
  * Added coverage for lowercase, mixed-case, and case-only header variations.

* **Performance**
  * Added a standalone benchmark for schema detection across large IFC files.


louistrue (approved):
Approved at exact head `10264b356d55638b76547456edc4023bb0bec124`. The shared case-insensitive byte search preserves the historical anywhere-in-file predicate and IFC4X3 precedence; immediate and later mixed-case suffixes are handled. Independent focused verification: all four schema-detection tests passed, including the raw-window oracle. Exact-head Claude and Ollama are also clean. Merge remains gated on the full required check set.

Issue #165, opened 2026-02-05
Show Spaces disappear with federated model


IbrahimFahdah:
When loading two models one has spaces and the other not, the "Show Spaces" disappears. The order of opening seems to affect this too.

<img width="1304" height="643" alt="Image" src="https://github.com/user-attachments/assets/07aa4b7f-9b95-4ca0-bbdb-268d1686d2e2" />

louistrue:
@IbrahimFahdah Thx! 

<img width="625" height="650" alt="Image" src="https://github.com/user-attachments/assets/54ae3ce1-7ca9-4da8-96e7-d81d0b8368cf" />

PR #4725, opened 2026-09-13
fix(parser): HAS_GEOMETRY reflects an entity's own Representation, not its class

### LINKED ISSUE #4666: EntityFlags.HAS_GEOMETRY is set by class bucket, so hasGeometry() returns true for an entity with no Representation
`EntityFlags.HAS_GEOMETRY` is set from the entity's **class bucket**, not from whether the entity actually carries a `Representation`. So `entities.hasGeometry(id)` returns `true` for an entity whose Representation slot is `$`.

Found while fixing #4655, where the viewer needed to tell "this is a physical object" from "this object has a shape".

## Reproduction

`tests/models/local/Building-Structural.ifc` (buildingSMART PCERT sample; the file is gitignored, but the entity is quoted in full below):

```
#162=IFCBUILDINGELEMENTPROXY('1CjP_CWub368bZVuVHeHs3',#1,'Group#21',$,$,#165,$,'454425.…',$);
#165=IFCLOCALPLACEMENT(#45,#166);
```

Attribute order for `IfcBuildingElementProxy` is GlobalId, OwnerHistory, Name, Description, ObjectType, **ObjectPlacement**, **Representation**, Tag, PredefinedType — so `#165` is the placement and the Representation is `$`. The entity has a position and no shape.

`entities.hasGeometry(162)` nonetheless returns `true`.

Two more in the same corpus: `Building-Architecture.ifc`'s `#339=IFCCHIMNEY` and `#345`, and `Building-Hvac.ifc`'s `#52` — all with Representation `$`. These are real export stubs, not synthetic cases.

## Why it matters

Any consumer reading the flag as "this entity has geometry" gets a wrong answer, and the wrongness is invisible — the entity is of a class that usually does have geometry, which is exactly why the bucket heuristic looks right.

The #4655 fix could not use the flag for this reason and derives the shape test from the mesh set plus an `IfcRelAggregates` descent instead. That is the correct behaviour for the viewer, but it means the flag's name currently promises something it does not deliver to every other caller.

Note the descent matters independently: `#196` in the same file is an `IfcRoof` whose own Representation is `$` but which aggregates six `IfcBeam`s and two `IfcDiscreteAccessory`s. A shape test that reads only the entity's own slot would drop a real roof. Whatever this flag comes to mean, it should be explicit about whether it covers aggregated geometry.

## Expected

Either:
1. `HAS_GEOMETRY` reflects actual `Representation` presence (and the docs state whether aggregated children count), or
2. the flag is renamed to what it actually is — a class-bucket hint — so no caller reads it as a fact about the entity.

Option 2 is cheaper and may be the honest answer; the bucket is presumably a fast prefilter and useful as one. The defect is the name and the implied contract, not necessarily the heuristic.

Not checked: every caller of `hasGeometry` / `EntityFlags.HAS_GEOMETRY` across the repo, and whether the Rust side has an equivalent flag with the same or different semantics.

BIMvoice:
## Problem

`EntityFlags.HAS_GEOMETRY` (and `entities.hasGeometry()`) was set from the entity's **class bucket** (`GEOMETRY_TYPES` in `columnar-parser-indexes.ts`) during columnar parsing, not from whether the entity's own `Representation` attribute is actually set. Every entity of a geometry-bearing class (`IfcElement`, `IfcSpace`, `IfcSite`, …) was stamped `HAS_GEOMETRY` unconditionally — an `IfcBuildingElementProxy` authored as a placement-only stub (`Representation` = `$`) reported geometry it did not have.

Reproduction shape from the issue (`tests/models/local/Building-Structural.ifc`, gitignored, not part of this PR):
```
#162=IFCBUILDINGELEMENTPROXY('1CjP_CWub368bZVuVHeHs3',#1,'Group#21',$,$,#165,$,'454425.…',$);
#165=IFCLOCALPLACEMENT(#45,#166);
```
`Representation` (attribute index 6) is `$`; `hasGeometry(162)` returned `true`.

## Callers checked and the decision

Grepped every reader of `hasGeometry` / `EntityFlags.HAS_GEOMETRY`:

- `packages/export/src/parquet-exporter.ts` and `packages/query/src/duckdb-integration.ts` export the flag as a `HasGeometry` column presented to external consumers as ground truth — these want the real answer, and get a wrong one today.
- `rust/export/src/model.rs` **already** computes `has_geometry` this way for its JSON/CSV export rows: `entity.get(6).is_some_and(|a| !a.is_null())`. The TS side disagreed with the Rust side it's supposed to match — this is the TS/Rust parity gap.
- `packages/charts/src/elements-dataset.ts` (chartable-elements filter), `apps/viewer/src/lib/lists/adapter.ts` (`selectableIds`), and `apps/viewer/src/lib/visibility/model-hidden-entities.ts` all use it as a coarse "product, not a relationship/pset" filter — an accurate per-entity check only makes their behavior more correct, not less (verified: none of these need aggregation coverage, and the viewer's own #4655 fix already handles the aggregation case separately).
- `packages/mcp/src/tools/viewer-assembly-expansion.ts` only *documents* the bucket semantics in a comment; it does not call `hasGeometry` itself.

Chose **option 1** from the issue (make the flag answer the real question) rather than option 2 (rename it), because the exported columns (parquet/duckdb/CSV/JSON) already present it as fact to end users, and the Rust export path had already implemented the accurate check — TS was the one that needed to catch up, not the contract that needed renaming. No public export was renamed, so no `api-surface:update` / semver-name consideration applies; this is a behavior fix at `patch` level (changeset included).

## What changed

`packages/parser/src/columnar-parser.ts` now reads the entity's own `Representation` attribute (byte index 6 — every `IfcProduct` descendant inherits it at that position, since `IfcProduct` is the last ancestor to declare an attribute before it) via a new `hasAttrValueAt` helper in `columnar-parser-attributes.ts`, instead of trusting the class bucket unconditionally.

**Aggregated children are deliberately NOT considered** — a container whose own `Representation` is `$` (e.g. an `IfcRoof` aggregating `IfcBeam`s, `Building-Structural.ifc` `#196`) still answers `false`. A caller needing "does this id or its parts render anything" must descend `IfcRelAggregates` itself, the way the viewer's object-count fix (#4655) already does over the mesh set. Documented on `EntityFlags.HAS_GEOMETRY` and `EntityTable.hasGeometry`.

## Testing

RED/GREEN/mutation, `packages/parser/test/entity-has-geometry-own-representation-4666.test.ts`:
- Reverted the fix and confirmed the test fails exactly as expected (2 of 3 cases red).
- Restored the fix: green.
- Mutated the core comparison to always return `true` with a probe log; the two distinguishing tests reddened and the probe log proved the mutated line executed for the `#162`-shaped entity.

Fixture guards against vacuity: one entity with `Representation` = `$` (false), one with `Representation` set (true), and an aggregation case where the container is `false` but its aggregated child is independently `true`.

Full `packages/parser` (1390 tests) and `packages/data` (291 tests) suites pass, plus `packages/charts`, `packages/export`, and `packages/query` suites (no regressions from the behavior change).

## Module-size gate

`node scripts/check-module-size.mjs` — touched files (`packages/data/src/entity-table.ts`, `packages/data/src/types.ts`) checked and kept at/under their existing budgets by tightening doc comments. The gate still exits 1 on this branch, but only due to pre-existing `rust/processing/tests/module_size_allowlist.txt` drift against files this PR never touches (`rust/geometry/**`, `rust/processing/**`, `rust/wasm-bindings/**`) — confirmed via `git diff --name-only upstream/main`, none of the flagged files appear in this PR's diff.

Closes #4666

## Summary by CodeRabbit

* **Bug Fixes**
  * Geometry indicators now reflect whether an entity has its own representation.
  * Placement-only entities no longer report geometry incorrectly.
  * Container entities without their own representation remain distinct from geometry-bearing aggregated children.

* **Documentation**
  * Clarified the meaning of the geometry flag and related behavior.

* **Tests**
  * Added regression coverage for entities with missing, present, and inherited geometry representations.


louistrue (changes requested):
Two correctness blockers remain in the new `HAS_GEOMETRY` contract:

1. `parseColumnarInput` still calls `addEntityBatch(spatialRefs, false, false)` and `addEntityBatch(otherRelevantRefs, false, false)`. Those buckets include `IfcProduct` descendants with an inherited `Representation` slot (`IfcSite`/`IfcBuilding`/`IfcSpace`, plus `IfcGrid`, `IfcAnnotation`, `IfcVirtualElement`, etc.). A spatial/product entity with `Representation=#...` therefore still reports false, even though the PR and changeset redefine the flag as own-Representation presence and the Rust exporter checks attr 6 for every exported entity. Please apply the same per-record rule to every stored `IfcProduct` descendant and add distinguishing spatial + other-product cases.

2. `hasAttrValueAt` treats only six whitespace bytes as trivia and `skipCommas` does not recognize `/* ... */` comments. STEP comments are legal trivia in attribute lists: `...,#165,/* omitted */$, ...` is currently reported true because the first byte examined is `/`; a comma inside an earlier comment can also shift the target slot. This creates a fresh fast-reader disagreement in exactly the comment class covered by #4720. Please use the canonical comment-aware attribute scanner (or make this helper comment-aware with regression cases before and at Representation) rather than adding another partial lexer.

louistrue (approved):
Approved exact head 59277aea3b after re-review. Both prior blockers are resolved, focused tests pass 10/10, and exact-head Claude plus Ollama are clean. Full CI remains a merge-queue gate.

PR #4736, opened 2026-09-13
feat(viewer): add a Structural Analysis properties card

BIMvoice:
Properties-card layer of #4206's six-layer structural analysis stack — parse (#4510, merged), semantic extraction (#4510), query surface `bim.structural` (#4677, merged), and now this UI layer.

## History this builds on

#4510 shipped `extractStructuralOnDemand(store)`, the read model of analysis models, members, connections, activities, load groups and result groups, cross-linked by GlobalId. #4677 shipped `bim.structural`, the query surface over that same read model, wired into CLI/MCP/viewer headless backends and the sandbox bridge — but no UI. #4677's own body listed the properties card as the first item of "what remains," and the last comment on #4206 (2026-09-13) confirms no newer maintainer decision exists to route around.

## Why this slice

Per the task brief carried forward from #4677: the properties card is the smaller of the two remaining UI/geometry pieces, and it is the first thing that makes the merged read model visible to an actual user rather than only to a script. The geometry channel is explicitly flagged in #4677 as needing a new extraction path (these members carry `IfcTopologyRepresentation`/`IfcEdge`, not solids — `processor_registry.rs` has no dispatch entry for that), which makes it a larger, riskier slice better taken on its own.

## What this adds

- `apps/viewer/src/components/viewer/properties/StructuralCard.tsx` — a self-contained card, mirroring `ScheduleCard`'s shape: renders nothing when the selection is not a structural member, so it's safe to mount unconditionally. When the selection *is* an `IfcStructuralMember` (curve or surface), it lists:
  - the analysis model(s) it belongs to,
  - the connections it is joined to, with the boundary condition's name and a fixed/free degrees-of-freedom summary when the file carries one,
  - the loads applied to it via its activities, rendering each numeric component under its own EXPRESS name (or, for an `IfcStructuralLoadConfiguration`, every entry with its parametric location).
- `PropertiesPanel.tsx` — extracts `extractStructuralOnDemand` on demand, memoized on the model/dataStore alone (not on selection, matching `georef`/`typeProperties`'s pattern), and renders `StructuralCard` next to `ScheduleCard` when the selection matches a member. No new store slice: unlike the schedule slice, nothing edits structural data, so an on-demand extraction is simplest and needs no teardown-registry entry.

**`loadsTruncated` is surfaced, never dropped.** The card reads `structuralData.loadsTruncated` (unchanged from #4510/#4677) and renders an amber "Truncated" badge in the card header when true, so a user looking at a member's loads can tell a genuinely small load list from one a reader bound (nesting cap, node budget, or a cycle guard) cut short. It renders nothing extra for a complete extraction — a truncated and a complete case are visibly distinguishable (see Evidence).

**No parallel arrays.** Every place this card pairs two things (a connection's `appliedCondition.components`, a load configuration's `entries[i].value`/`entries[i].location`) reuses #4510/#4677's existing paired-record shapes; nothing here introduces a new pair of indexed arrays that must be kept in sync.

**Attribute access.** This layer reads no raw IFC attributes — it only renders what `extractStructuralOnDemand` already resolved via the generated registry.

## Evidence

**RED**: before `StructuralCard.tsx` existed, `StructuralCard.test.tsx` failed with `Cannot find module './StructuralCard.js'`.

**GREEN, against the real fixture** (`tests/models/ifcopenshell/structural_analysis_curve.ifc` — the corpus's only structural-analysis export): `StructuralCard.test.tsx`, 5/5. It parses the fixture directly (same tokenizer/parser calls as `packages/parser/test/structural-extractor-fixture.test.ts`) and asserts *values*, not that a render succeeded:
- the analysis model name `Structural Analysis #1`,
- the pinned applied load `LinearForceZ: -100` at both `96` and `192`,
- `Connections (2)` for the loaded member (the continuous-beam fixture's per-member connection count),
- a fixed support rendered as `6 DOFs fixed` with the boundary condition's name `Fixed` — not the number `1`,
- a synthetic `loadsTruncated: true` extraction renders the "Truncated" badge while the real (complete) fixture does not — the two cases are distinguishable,
- an unmatched selection and a `null` extraction both render nothing.

Also re-ran the three existing `PropertiesPanel.*.test.tsx` files (assembly-select, duplicate-pset-key, group-isolate) alongside the new test — 20/20, confirming the new memo/card didn't disturb the existing panel.

**Mutation testing** — 4 mutations, each with a `console.error` hit-probe confirming the mutated line ran, each reverted after scoring:

| mutation | probe hit | tests reddened |
| --- | --- | --- |
| Hardcode the "Truncated" badge condition to `false` | yes | 1 (`distinguishes a truncated extraction from a complete one`) |
| `findMember` always returns `data.members[0]`, ignoring the selection | yes | 2 (`renders the loaded member's...` and `renders nothing when the selection is not a structural member`) |
| `formatDofs`'s `fixed` filter drops the boolean branch (`v === true`) | yes | 1 (`summarizes a fixed support as all six DOFs fixed...`) |
| `connections` memo returns all of `structuralData.connections` instead of the member's own | yes | 1 (`renders the loaded member's...`, wrong count `Connections (4)`) |

## Gates run locally

- `pnpm typecheck` (turbo, all 51 packages + apps, 2037 test files) — clean.
- `pnpm lint` — 0 errors (5 pre-existing warnings, none in files this PR touches).
- `node scripts/check-module-size.mjs` — `PropertiesPanel.tsx` grew from 2146 to 2187 lines (the new memo + JSX block); its allowlist row is raised to the measured count in this PR. `StructuralCard.tsx` is a new 254-line file, well under budget, no row needed. `git diff upstream/main -- rust/processing/tests/module_size_allowlist.txt` is empty — the script's non-zero exit is the pre-existing Rust allowlist/measured-count drift this task brief flagged as unrelated to TS work; this PR touches no Rust file.
- Changeset added: `@ifc-lite/viewer` minor (new user-visible card; no exported symbol added, so no `api-surface:update` needed).

## What remains — stated, not filed as a sub-issue

Per #4677's own list, still untouched:

1. **Geometry** — structural curve/surface members through a line channel (`setLineOverlay`-style), with a class toggle. Needs a new extraction path (`IfcTopologyRepresentation`/`IfcEdge`, not solids); the largest remaining piece.
2. **Write / round-trip** — a `structural-serializer.ts` matching `schedule-serializer.ts`, with a fixture-diff test. Nothing edits structural data today, so this has no caller yet; #4206's comment thread already flags this as an open question (round-trip test vs. the loss report the issue's "ends when" bar asks for).
3. **The second real-tool fixture** the issue's "ends when" bar asks for cannot be supplied by code. #4677 already checked both other structural files in the manifest (`buildingsmart/Building-Structural.ifc`, `various/01_Snowdon_Towers_Sample_Structural(1).ifc`) and found zero structural-analysis entities in either — they are structural products, not analysis models. `structural_analysis_curve.ifc` remains the corpus's only analysis fixture; this bar is genuinely unsourced, not merely unchecked.
4. IFC2X3: not exercised against a real 2X3 structural export in this or either prior PR.

## Queue gate

Refs #4206 (partial: the properties-card layer of the six the issue scopes; deliberately does not close it — geometry, write/round-trip, and the second fixture all remain).

## Summary by CodeRabbit

- **New Features**
  - Added a Structural Analysis card to the Inspector’s Properties tab.
  - Selecting a structural member now shows its analysis model, connected elements, boundary conditions, and applied loads.
  - Boundary conditions display fixed, free, and elastic degrees of freedom.
  - Loads that exceed display limits are marked as **Truncated**.

- **Tests**
  - Added coverage for structural analysis details, loads, connections, boundary conditions, truncation indicators, and empty or non-matching selections.


louistrue (changes requested):
Blocking correctness issue in StructuralCard.formatDofs: numeric boundary-condition components are stiffness magnitudes, not booleans. The existing parser contract explicitly preserves number versus boolean because IFCBOOLEAN(.T.) means rigid/fixed, while e.g. IFCLINEARSTIFFNESSMEASURE(2000.) means an elastic/spring restraint. The current predicate treats every nonzero number as fixed, so a partially restrained support is presented to users as rigid. Please render boolean components as fixed/free and numeric components under their exact EXPRESS names with their stiffness values (including mixed conditions), and add a production-observing test with at least one nonzero numeric stiffness. The current real fixture only covers all-boolean fixed conditions, so it cannot catch this misclassification.

louistrue (approved):
Approved after independent review of the exact head 07f8dc74fc66: the numeric stiffness handling, structural empty-state gate, real PropertiesPanel coverage, module-size cleanup, resolved threads, Claude/Ollama reviews, and hosted checks are all green. Merge remains gated on the exact current main run.

Issue #31, opened 2026-01-14
Scale is wrong?


holg:
I cannot imagine the sample Hybrid BIM office has the size of 40km :d

louistrue:
Did you check if it has wrong units in the ifc? 🤷‍♂️

holg:
@louistrue  Yes the rest works for me well, actually right now i am doing the branch, using only rust/bevy, some painful,
but got it working so far: https://bimifc.de
or, to load the IFC directly:
https://bimifc.de/?file=01_BIMcollab_Example_ARC.ifc

holg:
@louistrue Well i have some experience in this:
https://eulumdat.icu
https://gldf.icu
Right now i do some refacturing, so the code does survive very strict tests, taplo, cargo-fmt, cargo-deny, cargo clippy --workspace --all-targets -- -D warnings, and others, i have the scripts/test-ci-locally.sh, which i use for my projects...

louistrue:
issue is fixed now in #39, thanks again!

PR #4737, opened 2026-09-13
feat(geometry): mesh IfcStructuralCurveMember edges so analysis models render

### LINKED ISSUE #4206: Structural analysis end to end, as the template for every later domain

BIMvoice:
Layer 4 of the six that #4206 scopes: the geometry channel.

Refs #4206

This is partial: layer 4 of the six the issue scopes; it deliberately does not close #4206.

Layers 1-3 are merged (#4510 parser read model, #4677 `bim.structural` query namespace). This one makes structural curve members actually render.

## Why a new path was needed

Two independent gates each dropped structural geometry, and fixing either alone still renders nothing. Both verified by instrumenting the real fixture, not inferred:

1. **`rust/geometry/src/router/rep_filter.rs`, `meshed_representations`** — its `is_shape` closure was `|rep: &&DecodedEntity| rep.ifc_type == IfcType::IfcShapeRepresentation`. A structural curve member's geometry hangs off an **`IfcTopologyRepresentation`**, a different entity type, so the representation was filtered out before any type string was read. This gate is the deeper of the two and is easy to miss: the rep-type check below it never runs.

2. **`rust/geometry/src/router/rep_filter.rs`, `is_body_representation`** — the canonical body/surface predicate lists `Body/SweptSolid/Brep/CSG/Clipping/Tessellation/MappedRepresentation/SolidModel/SurfaceModel/Surface3D/AdvancedSweptSolid/AdvancedBrep`. No `Edge`. That is correct and is left alone: its doc comment records that it is shared with RTC-offset detection, void probing and material-layer slicing, and a structural edge is none of those. `effective_rep_type` reads attribute 2 first, which for `#253 = IFCTOPOLOGYREPRESENTATION(#212,'Reference','Edge',(#252))` is literally `'Edge'`.

3. **`rust/geometry/src/router/processor_registry.rs`** — `TYPES` (18 entries), `slot()` and `create()` had no `IfcEdge` arm, so even a representation that got past the filter had no processor. With the filter fixed but the registry untouched, the router returns `GeometryError("Unsupported representation type: IfcEdge")` (observed under mutation 3 below).

Net effect before this PR: `process_geometry` on `tests/models/ifcopenshell/structural_analysis_curve.ifc` returned **0 meshes**.

## What this adds

- **`rust/geometry/src/router/structural.rs`** (new, 32 lines) — `pub(super) fn accepts(element, rep_type)`, gating on `IfcType::IfcStructuralCurveMember` via `is_subtype_of` (so `IfcStructuralCurveMemberVarying` is covered) and `rep_type == "Edge"`. Mirrors `annotation::accepts` and is OR'd into the filter as an escape hatch, deliberately **not** folded into `is_body_representation`.
- **`rust/geometry/src/router/rep_filter.rs`** — adds `is_meshable_rep`, which accepts an `IfcShapeRepresentation` as before, or an `IfcTopologyRepresentation` *only* when `structural::accepts` matches. That keeps `IfcTopologyRepresentation` closed in general: the fixture's `'Vertex'`-typed ones on `IfcStructuralPointConnection` stay filtered out.
- **`rust/geometry/src/processors/structural_edge.rs`** (new, 162 lines) — `IfcEdgeProcessor`. Resolves `EdgeStart`/`EdgeEnd` → `IfcVertexPoint` → `VertexGeometry` → `IfcCartesianPoint`, emits a 4-vertex / 2-triangle ribbon, 0.1 file-unit half-width (`alignment.rs`'s `RIBBON_HALF_WIDTH_FILE_UNITS` is 0.25; a structural member is a slenderer thing).
- **`rust/geometry/src/router/processor_registry.rs`** — slot 18 for `IfcType::IfcEdge`, 18 → 19 `TYPES` entries, following the existing pattern.

**Deliberately out of scope, and still meshing empty:** `IfcOrientedEdge` and `IfcEdgeCurve` (neither appears in the fixture; half-implementing them untested would be worse than a stated gap), `IfcStructuralSurfaceMember`, and `'Vertex'`-typed representations on `IfcStructuralPointConnection`.

## Tolerances

Both checked at both ends of the range, per the brief:

- **Degenerate edge** is an **exact-zero** test on `dir.norm()`, not a magnitude-scaled epsilon. Two identical `IfcCartesianPoint`s subtract to bit-exact zero whether their coordinates are `1e-6` or `1e9`, so there is no scale at which this behaves differently, and a genuinely tiny but nonzero edge stays real geometry rather than being rejected.
- **Parallel-to-up fallback** compares `dir.cross(&up_z).norm()` against a fixed `1e-9`. Both operands are already unit-length, so that norm *is* the sine of the angle between them — dimensionless and scale-invariant. It does not vary with coordinate magnitude.

No net cross-product or area test is used to prove anything about closure: this is an open ribbon, and the cross product is used only to pick a perpendicular direction.

## RED / GREEN

**RED**, measured on the fixture before the fix: `process_geometry` → `total meshes: 0`, `total_triangles: 0`. Driving the router directly on `#228` returned `Ok` with an empty mesh (`verts=0 idx=0`) — the representation never reached an item.

**GREEN**: 3 meshes, one per curve member, 12 vertices / 6 triangles total. The integration test asserts exact world-space coordinates for all three, not a non-empty check.

One correction worth recording, since it would have produced a wrong "expected" value: this fixture's assigned length unit is `#31`, an `IfcConversionBasedUnit` named `'inch'` with a 0.0254 factor — **not** the `IFCSIUNIT(*,.LENGTHUNIT.,$,.METRE.)` on `#28` that the conversion is expressed against, which is the only `IFCSIUNIT` length line in the file and reads as the project unit on a first grep. `ModelMetadata.length_unit_scale` is `Some(0.0254)`, so the expected metre coordinates are the file's inch coordinates × 0.0254. The test documents this in its header and derives the expectation from the file units rather than hardcoding the scaled numbers.

## Mutation testing

Each new guard flipped individually, each carrying an `eprintln!` hit-probe so a mutation that never executed cannot read as a false pass. Every probe fired. All five reverted afterwards and the suite reconfirmed 6/6 green.

| # | Mutation | Probe fired | Tests reddened |
|---|---|---|---|
| 1 | `structural::accepts` forced to `false` | yes | 3/6 — `accepts_structural_curve_member_edge_representation`, `straight_edge_meshes_to_a_ribbon_with_correct_endpoints`, `edge_parallel_to_up_axis_uses_the_fallback_perpendicular` |
| 2 | `is_meshable_rep` reverted to shape-only (drops the `IfcTopologyRepresentation` arm) | yes | 2/6 — `straight_edge_...`, `edge_parallel_to_up_...` |
| 3 | `processor_registry::slot()` `IfcEdge` arm returns `None` | yes | 3/6 — plus `zero_length_edge_...`; surfaced as `GeometryError("Unsupported representation type: IfcEdge")` |
| 4 | Zero-length guard disabled | yes, printing `length=0` | 1/6 — `zero_length_edge_meshes_empty_instead_of_producing_nan_geometry` |
| 5 | Parallel-to-up fallback disabled (normalizes the zero vector) | yes, printing `cross_z.norm()=0` | 1/6 — `edge_parallel_to_up_axis_uses_the_fallback_perpendicular`, via NaN propagation |

## Tests

- `rust/geometry/src/router/structural_tests.rs` (new) — 6 synthetic-entity unit tests: the accepts predicate, `'Vertex'` rejection, non-structural-element rejection, ribbon endpoints, the parallel-to-up fallback (asserting the ribbon has real width rather than collapsing), and the zero-length degenerate case.
- `rust/processing/tests/issue_4206_structural_curve_member_geometry.rs` (new) — real-fixture end-to-end through `process_geometry`, asserting each member's exact endpoints. Skips cleanly with a `pnpm fixtures` hint when the fixture is absent.

## Gates run locally

- `cargo test --workspace` — **green**, 225 `test result: ok` groups, 0 failures.
- `cargo clippy --workspace --all-targets -- -D warnings` — **green**, no warnings.
- `cargo test -p ifc-lite-processing --test module_size_ratchet` — green, 8/8. New files are 32 and 162 lines, well under 400; neither needs an allowlist row and the allowlist is untouched.
- `node scripts/check-module-size.mjs` — reports the known pre-existing Rust allowlist drift (11 rows: `classify.rs`, `boolean/mod.rs`, `rtc_offset.rs`, `voids/geom.rs`, `voids/mod.rs`, `prepass.rs`, `processor/mod.rs`, `gpu_meshes/prepass.rs`, `api/mod.rs`, `zero_copy/mesh.rs`, `zero_copy/symbolic.rs`). **None of those are files this PR touches**, and this PR adds no new row and raises none.
- No TypeScript changed, so no typecheck lane applies.

Changeset added (`@ifc-lite/wasm` minor, `@ifc-lite/server-bin` minor): a rebuilt wasm now returns geometry for elements that previously produced none, matching how sibling geometry-output changes are versioned here.

## Still open for #4206

- **Properties card** — a separate branch, in progress by someone else. Not claimed or touched here.
- **Write / round-trip** (layer 6) — untouched.
- **A second real-tool fixture** — still unsourced; `structural_analysis_curve.ifc` remains the corpus's only analysis fixture. Noted, not re-litigated.
- **`IfcOrientedEdge` / `IfcEdgeCurve` / `IfcStructuralSurfaceMember` geometry** — scoped out above.

No viewer or TypeScript wiring was needed: meshed geometry from any registered `GeometryProcessor` flows through `produce_element_meshes` to both the WASM batch path and the native/CLI/server path, exactly as `IfcAlignment` does with zero viewer-side code. That assumption was checked rather than assumed — the fixture's three members reach `process_geometry`'s output with correct coordinates, which is downstream of the fork.

## Summary by CodeRabbit

* **New Features**
  * Structural curve members, including varying members, now render as visible thin ribbon geometry between their defined start and end points.
  * Structural topology representations using edge geometry are now supported.
  * Zero-length edges safely produce no geometry.

* **Bug Fixes**
  * Fixed structural curve members incorrectly producing empty meshes and remaining invisible in the viewer.


louistrue (changes requested):
Blocking rendering correctness issue in IfcEdgeProcessor: every vertex is assigned the hard-coded normal (0,0,1), but the ribbon is only horizontal when the edge direction has zero Z component. For the fixture's vertical members, the generated quad lies in a vertical plane (the fallback right vector is plus-or-minus Y), so its geometric normal is plus-or-minus X while the stored normal remains plus Z. Sloped members likewise receive a normal with a component along the edge rather than the face normal. That produces incorrect lighting and makes winding disagree with normals. Derive the unit face normal from the actual ribbon frame with the sign matching triangles (0,1,3)/(0,3,2), use it for all four vertices, and extend the existing vertical/sloped tests to assert finite unit normals that agree with the triangle cross product. The current endpoint/width assertions cannot observe this defect.

louistrue (approved):
Approved on exact head a388afbdb1 after independent review and follow-up fixes. The ribbon normal now matches emitted winding for horizontal, vertical, and sloped members; focused tests and surgical normal/routing mutations prove the behavior; hosted Test, Rust, E2E, Claude, Ollama, and CodeRabbit are clean. The near-zero Ollama note is non-actionable for a normalized finite nonzero direction. #4206 remains correctly open because this is only its geometry slice.

Issue #4153, opened 2026-09-08
2D drawing markup is lost on reload: no client-side persistence for measurements, annotations, clouds or DXF underlays


BIMvoice:
## What happens

Draw anything on the 2D drawing canvas — a measurement, a polygon area, a text note, a revision cloud, or an imported DXF underlay — then reload the page or close the tab and come back. All of it is gone. The 2D panel reopens empty, and the generated plan/section has to be regenerated from scratch because even its `SectionConfig` is not saved.

There is no way to keep a drawing across a session, and no way to get one back: `useDrawingExport.ts` offers SVG, DXF and PDF, but all three are one-way outputs with no importer back into `Drawing2DState`. Nothing round-trips.

## Why it reads as a missing feature rather than a bug

The sibling feature **does** persist. 3D world-anchored note pins (`store/slices/annotationsSlice.ts`) survive a reload — they load from `localStorage` under a versioned key and save on every mutating action. So within the same viewer, one kind of markup comes back and the other silently does not.

The teardown module makes the gap easy to miss. `drawing2DSlice.teardown.ts` deliberately makes `model-removed` and `all-models-cleared` no-ops, with a comment stating the 2D view is regenerated on demand and "must not throw away the user's markup". That promise holds **within** a session. There is no `session-reset` exemption and nothing reads from storage on init — `getDefaultDrawing2DState()` has no `loadFromStorage()` call, unlike the pins slice — so a reload wipes everything unconditionally.

## Scope

This is about the **2D drawing canvas and its markup**: `components/viewer/Drawing2DCanvas.tsx`, `components/viewer/Section2DPanel.tsx`, `store/slices/drawing2DSlice.ts`.

Explicitly **not** in scope, to keep this reviewable:

- `tools/SpaceSketchOverlay.tsx` (space sketching bakes drafted polygons into real `IfcSpace` entities — a different concern),
- `annotationsSlice.ts` 3D pins, which already persist.

## What needs persisting

From `Drawing2DState`:

```ts
measure2DResults: Measure2DResult[];         // { id, start, end, distance }
polygonArea2DResults: PolygonArea2DResult[]; // { id, points, area, perimeter }
textAnnotations2D: TextAnnotation2D[];       // { id, position, text, fontSize, colors }
cloudAnnotations2D: CloudAnnotation2D[];     // { id, points, color, label }
dxfUnderlays: DxfUnderlayState[];            // imported CAD layers + placement
[...]

Issue #2802, opened 2026-08-18
packages/apps: the coverage-gap half of #2765 (slices no fixture reaches, plus four decisions)


louistrue:
Split out of #2765, which pairs 11 mutation-verified fixture gaps with a broader coverage inventory. PR #2799 closes the 11; this issue carries the rest, so #2765 can close on evidence rather than staying open indefinitely as an umbrella.

These are **not** mutation-verifiable gaps. Mutating them proves nothing because no test reaches the code at all — the emptiness is the finding, and each item is a missing suite rather than a missing assertion.

## No fixture reaches the code

- **`collabSlice.ts`** — 1198 lines, wired into the store, and **no test calls `startCollab` or `createCollabSlice`**. Replacing `get().activeModelId` with a garbage literal left 60/60 green and `tsc` clean. (#2706 adds an entry-race test that drives only the synchronous prefix, up to the first `await import`.) Worth noting the shape: a slice this size with zero direct coverage is where the #2790 suppressed-signal defects lived, and they were found in production rather than by a test.
- **17 slices with zero test references**, 3169 lines total: `addElementMeshes`, `addElementSlice`, `bcfSlice`, `cameraSlice`, `environmentSlice`, `extensionsSlice`, `geometryLoadSettings`, `hoverSlice`, `idsSlice`, `listSlice`, `loadingSlice`, `playbackSlice`, `sheetSlice`, `solarSlice`, `sourcesSlice`, `spaceMouseSlice`, `splitToolSlice`.
- **Partially tested slices**: `lensSlice` (18 actions, 6 tested), `compareSlice` (11 of ~17 untested — the whole run-result lifecycle), `pinboardSlice` (478 lines, ~19 of 25 untested), `drawing2DSlice` (862 lines, ~55 of 62 untested), and `annotationsSlice`'s `upsertRemoteAnnotation` / `removeRemoteAnnotation` collab-sync pair.

## Decisions rather than tests

- **`deserializeClashConfig` has no schema-version gate** — missing, non-numeric and future (`999`) versions are all silently accepted, and an array blob degrades to defaults rather than returning `null`. May be deliberate forward-compat; recorded in #2763.
[...]

PR #4762, opened 2026-09-14
fix(viewer): persist dxfUnderlays to IndexedDB (issue #4153, reopened)

BIMvoice:
Refs #4153

## The gap

Four PRs (#4159, #4160, #4167, #4170, #4191) already closed most of #4153: `measure2DResults`, `polygonArea2DResults`, `textAnnotations2D`, `cloudAnnotations2D`, `drawing2DDisplayOptions` and the producing `SectionConfig` all survive a reload now, persisted to `localStorage` keyed by the model's full-content SHA-256 hash (`drawing2DSlice.persistence.ts`). `dxfUnderlays` was deliberately left out of that work — an imported DXF underlay can embed arbitrary point counts, plausibly exceeding `localStorage`'s ~5MB synchronous budget — with the issue itself naming IndexedDB as the intended home, mirroring `services/ifc-cache.ts` / `services/extensions/idb-storage.ts`'s existing convention.

The issue was reopened on 2026-09-13 specifically because this piece was still missing after all four PRs merged.

## The fix

- `apps/viewer/src/store/slices/drawing2DSlice.dxfPersistence.ts` (new): an IndexedDB storage layer for `dxfUnderlays`, one object store keyed by the same full-content hash the markup module already uses. Follows `idb-storage.ts`'s convention (versioned `onupgradeneeded`, recreate-from-scratch recovery if the store is missing), plus per-entry runtime validation that drops malformed entries instead of failing the whole load, and the same 20-entry oldest-evicted-first cap `drawing2DSlice.persistence.ts` uses.
- `apps/viewer/src/hooks/dxfUnderlaySave.ts` (new): the save/restore glue — a raw store subscription that saves on every `dxfUnderlays` change (coalesced so a rapid edit sequence keeps at most one write in flight, always ending on the latest value), and a restore function wired into `useDrawing2DPersistence.ts`'s existing hash-resolution flow.
- `apps/viewer/src/hooks/useDrawing2DPersistence.ts`: wires the above into the existing restore effect, reusing the hash it already computes rather than recomputing it.

**Not a straight replace-on-restore, by design**: `drawing2DSlice.markupTransition.ts` classifies `dxfUnderlays` `'preserved'` (issue #2802) — unlike the five per-model markup fields, it is never cleared on a model switch; a DXF reference belongs to the workspace, not to whichever model is active. So restore here is additive (id-deduped merge into whatever is already loaded), never a clear-and-replace.

**Empty vs. absent**: `loadDxfUnderlaysEntry` returns `null` when nothing was ever saved for a hash (or the stored value is corrupt), and a real `{ dxfUnderlays: [], savedAt }` when the user removed every underlay and that state was saved. The restore path only skips on the `null` case — a saved empty array reaches the merge function and is a correct no-op there, but the distinction survives at the load layer.

**Storage unavailable**: `indexedDB` can be undefined (some private-browsing modes) or reject on open/read/write (quota, blocked upgrade). Every access is guarded; failures degrade to "nothing restored/saved" with a `console.warn`, never a throw.

## Testing

New tests: `drawing2DSlice.dxfPersistence.test.ts` (save/load round trip, empty-vs-absent, malformed-entry skipping, cross-model isolation), `drawing2DSlice.dxfPersistence.recovery.test.ts` (missing-object-store recovery), `dxfUnderlayPersistence.test.tsx` (integration: restore-on-model-activate, untouched when nothing saved, no premature restore before the hash resolves, and two race-guard tests for a fast model switch during hash resolution / during the IndexedDB lookup itself).

Mutation-tested independently: deleted the `stillCurrent()` guard in `dxfUnderlaySave.ts`'s `restoreDxfUnderlaysFor` — the targeted race test failed with exactly the expected wrong result (a stale model's saved underlay landing on the new model), reverted, confirmed green again.

- [x] `apps/viewer` targeted suite (63 tests across the touched + adjacent drawing2D persistence/transition files): all pass
- [x] `node scripts/check-module-size.mjs` (plain form, no `--base`): exits 1, but the only violations are pre-existing Rust allowlist drift in `rust/geometry`, `rust/processing`, `rust/wasm-bindings` — none of the files this PR touches appear in its output

## Exclusions

- No suppress/restoring-model guard analogous to the markup path's `suppressNextSaveFor` — not needed, since `dxfUnderlays` is never atomically cleared on a model switch (no clear-vs-edit race exists for it to guard against).
- No debounce timer beyond last-write-wins coalescing on the save side.

## Summary by CodeRabbit

- **New Features**
  - DXF underlays are now saved and restored across sessions using persistent browser storage.
  - Saved underlays are associated with the corresponding model content and restored when that model is reopened.
  - Restoration adds saved underlays to currently loaded content without replacing existing underlays.

- **Bug Fixes**
  - Recent underlay edits are preserved while model processing completes.
  - Prevented stale restorations and duplicate underlays when switching models or reopening saved content.
  - Persistence continues to degrade gracefully when browser storage is unavailable or full.


louistrue (approved):
Maintainer review complete on exact head d12b3dd8f512439ee2044045aba082c99d681719. The persistence lifecycle, per-model hash isolation, cached A→B→A serialization, atomic eviction, deduplication, and production-observing regressions are sound. Exact-head Test is green; Claude failures are confirmed quota-only, and Ollama findings were independently adjudicated above.

Issue #1765, opened 2026-07-16
Server-parse path: Description/ObjectType/PredefinedType/Tag list columns are empty


louistrue:
Follow-up from #1759 (server↔client Lists parity).

The server data-model extraction (`apps/server/src/services/data_model/metadata.rs`) only emits `entity_id`, `type_name`, `global_id`, `name`, `has_geometry`. So on the **server parse path**, the built-in Lists attribute columns `Description`, `ObjectType`, `PredefinedType`, and `Tag` render empty, while the in-browser (WASM) path resolves them on demand.

Pre-existing (not introduced by #1759, which covered type properties/QTOs + `Name`/`Type`), but it's the remaining gap for full attribute-column parity between the two parse paths.

### To close
- Extract these attributes in `metadata.rs` (Description = attr 3, ObjectType = attr 4 on `IfcRoot`/`IfcObject`; PredefinedType via the element's enum attribute; Tag = `IfcElement.Tag`), add columns to `EntityMetadata` + the parquet entities table (additive, bump the datamodel cache version), and read them in `serverDataModel.ts` `buildEntityTable`.
- Add to the cross-path parity test (`apps/viewer/src/lib/lists/server-type-parity.test.ts`).

PR #4763, opened 2026-09-14
fix(server): schema-derive relationship attribute slots (Rust half of #4205)

BIMvoice:
Refs #4205.

## Problem

`apps/server/src/services/data_model/relationships.rs` hand-enumerated only 13 STEP relationship types (the extraction GATE) and a hand-written `(relating_idx, related_idx)` match per type, falling back to a wrong default `(4,5)` for anything else — e.g. the `IfcRelAssigns` family puts `RelatedObjects` before `RelatingX`, which the default silently mis-reads. Every `IfcRelationship` subtype outside the 13-entry list was dropped entirely on the server-parse path, while the TS/WASM path (`packages/parser/src/relationship-schema-slots.ts`, #4672) already covers the schema-derived set generically.

## Fix

New generator `scripts/generate-server-relationship-slots.mjs`, modeled on the existing `scripts/generate-server-attr-indices.mjs` precedent (issue #1765): derives a Rust table (`apps/server/src/services/data_model/generated/relationship_slots.rs`) from the SAME schema-derived source the TS/WASM path already resolves relationship slots against (`getAllConcreteRelationshipTypes()` / `getRelationshipSlotPlan()`). `relationships.rs`'s `extract_relationship()` and its GATE now read this table instead of a hand-written match — using the table lookup itself as the gate (`relationship_slots(type_upper).is_some()`), so there is exactly one source of truth, not two that can drift.

**Coverage: 13 → 54 relationship types** (a strict superset — the 13 previously-covered types resolve to byte-identical absolute indices as the old hand-written table, verified before this generator's output replaced it). One schema-derived type, `IFCRELASSOCIATES`, does not resolve a slot plan: it is an abstract supertype in every real EXPRESS schema (no `RelatingX` attribute of its own, only the shared `RelatedObjects`) that IFC2X3's bundled registry happens to mark non-abstract. No concrete STEP file instantiates it directly, so this is a documented, deliberate exception, not a dropped case — the generator fails closed if more than one type is ever unresolved, so a real regression here would not go unnoticed.

The three special-cased single-ref branches (`IFCRELVOIDSELEMENT`, `IFCRELFILLSELEMENT`, `IFCRELCONNECTSPATHELEMENTS`) and the hand-written `(relating_idx, related_idx)` match are deleted — the generated table's `related_is_list` flag now drives the single-ref-vs-list read generically.

## Testing

New Rust tests in `apps/server/src/services/data_model/tests.rs` reuse the EXACT STEP fixture entities (`#10`/`#11`/`#12`/`#20`/`#21`/`#22`) from `packages/parser/test/relationship-subtype-coverage.test.ts` (added by #4672) for `IFCRELASSIGNSTOACTOR` (list-typed, RelatingActor is the LAST attribute), `IFCRELDECLARES` (both attributes SELECT-typed — the case that caught a real generation-time bug on the TS side), and `IFCRELSEQUENCE` (single-ref both sides) — asserting the identical relating/related ids the TS suite asserts, so both stacks are proven to agree on the same byte input, not just on table contents. A control fixture confirms the schema-derived gate doesn't start matching unrelated types.

Mutation testing (transcript-only, reverted before commit):
- Removed the `IFCRELASSIGNSTOACTOR` arm → exactly `extracts_assigns_to_actor_relationship_previously_unindexed` reddened.
- Hand-truncated the generated table to the original 13 types → exactly the 3 new-type tests reddened; all 13-type regression tests stayed green.
- Confirmed `IFCRELDECLARES`'s SELECT-vs-primitive distinction is inherited correctly from the TS generation-time fix — nothing to mutate further on the Rust side for that defect class.
- Reapplied via regeneration; `git diff` against a pre-mutation backup was byte-identical.

`--check` mode proof: mutated a committed index by hand, confirmed non-zero exit with the exact drifted entry named, reverted, confirmed clean re-pass.

Gates: `cargo test --workspace` (0 failed), `cargo clippy --workspace --all-targets -- -D warnings` (clean), `node scripts/generate-server-relationship-slots.mjs --check` (in sync), `pnpm --filter @ifc-lite/parser test` (1429 passed / 10 pre-existing skips). `node scripts/check-module-size.mjs` (plain form) exits 1, but every flagged line is in files this PR never touches (pre-existing `rust/geometry`/`rust/processing`/`rust/wasm-bindings` allowlist-ratchet drift, unrelated to this diff).

## Out of scope (left for a follow-up against #4205)

- `packages/cli/src/commands/subset-relations.ts` — still hand-written attribute slots.
- Viewer/query/MCP display of the newly-reachable relationship types by exact IfcRel* name.
- 18 concrete relationship classes that pass the schema-derived gate but have no dedicated `RelationshipType` enum member yet on the TS side (per #4672's own scoping) — this PR's Rust table covers them structurally (they'd extract correctly if/when TS gains a `RelationshipType` for them), but nothing surfaces them today.
- `packages/parser/src/relationship-schema-slots.ts` itself is untouched.

## Summary by CodeRabbit

- **New Features**
  - Server-side relationship extraction now supports all 54 concrete schema-derived relationship subtypes.
  - Relationship attribute positions are generated from the parser schema, improving consistency with browser-based parsing.
  - List-based, single-reference, and SELECT-typed relationship patterns are now handled.
  - Server and browser relationship parsing are aligned for supported relationship types.

- **Bug Fixes**
  - Prevented valid relationships from being omitted during data extraction.
  - Preserved correct relationship orientation across supported subtype variations.


louistrue (approved):
Independent review complete. The schema-derived extraction is sound, generated slots are now freshness-gated in CI, Rust fixture coverage observes list/single/select layouts, and exact-head Ollama plus the full Test/server matrix pass. The remaining Claude checks are provider-credit failures only.

PR #4761, opened 2026-09-14
fix(viewer): count physical objects with a shape in the metadata panel and hierarchy geometry filter

BIMvoice:
Closes #4760

## What was missed

#4656/#4658 unified "how many objects with geometry" onto one rule (physical `IfcElement` + a shape) across the hierarchy trees and the StatusBar. Two other surfaces still disagreed:

**`ModelMetadataPanel`'s "Elements with Geometry" row** summed `spatialHierarchy.byStorey` array lengths — raw `IfcRelContainedInSpatialStructure` membership, no schema filter and no geometry filter, despite the label. On the buildingSMART PCERT sample `Building-Structural.ifc`, storey `#43 '00 groundfloor'`, it read **7** where the hierarchy trees and StatusBar (post #4656/#4658) read **6**. The seventh entity is #4655's `#162 IFCBUILDINGELEMENTPROXY(...,'Group#21',...)` — `Representation = $`, no shape.

**`hierarchyGeometry.ts`'s `buildGeometricIdSet`** added only `mesh.expressId`, while `lib/object-count.ts`'s `collectMeshedIds` (the StatusBar's rule) also adds the keys of `instancedGeometryHashes`/`Aabbs`/`Volumes` — entities rendered only through GPU instancing. A fully-instanced element (8+ repeats route to the instanced shard and produce no `meshes` entry at all) had geometry in the StatusBar and lacked it in the hierarchy trees, silently vanishing from By Class and By Type. Same omission #2865 found in clash detection — a third independent instance of the same gap.

## Fix — delegate, not duplicate

- `buildGeometricIdSet` now delegates to `collectMeshedIds` for both the federated and legacy paths, instead of re-deriving "which ids have geometry" a third time. No local/global id conversion is needed: a federated model's `geometryResult` ids and its instanced side-channel keys are already re-keyed into the same global space by `useIfcLoader.ts`, so `collectMeshedIds`'s default identity `toLocalId` is exactly right.
- `ModelMetadataPanel`'s stats move into `modelMetadataStats.ts::computeModelStats`, composed from the same `collectPhysicalEntityIds` (schema test) + `countShapedObjects` (shape test, via `collectMeshedIds`) the StatusBar uses — not a third rule. `idOffset` converts `geometryResult`'s global ids back to the `IfcDataStore`'s model-local space before the schema/shape tests run (0 for the legacy single-model path, a no-op).

## Empty-vs-absent contract

[...]

PR #4765, opened 2026-09-14
fix(viewer): gray out and bucket geometry-less elements under "Other"

### LINKED ISSUE #4764: fix(viewer): gray out and bucket geometry-less elements under "Other" (spatial tree, By Class, By Type)
## The gap

An `IfcElement` with no geometry — a placement (`ObjectPlacement`) but `Representation = $`, and no `IfcRelAggregates` part carrying one either — is not visually distinguished from an element that has geometry anywhere in the hierarchy browser. #4656/#4658/#4725 already unified "which elements have geometry" onto one rule for the *count*, but the *row* itself still renders identically to a normal, shaped element:

- **Spatial tree**: the row appears inline, mixed in with shaped elements, with no visual cue.
- **By Class tab**: the element is silently dropped from its class group (correctly excluded from the count, but invisible — unfindable, not inspectable).
- **By Type tab**: same drop, for a typed occurrence.

Real example: `#162 IFCBUILDINGELEMENTPROXY('...','Group#21',...)` in the buildingSMART PCERT sample `Building-Structural.ifc` — a real `IfcElement` whose `ObjectPlacement` (attribute slot 6) is set but whose `Representation` (slot 7) is `$`. It is **not** an `IfcGroup` — it is a proxy someone merely *named* "Group#21". `IfcGroup`/`IfcZone`/`IfcSystem` cannot appear in a spatial tree at all (they don't descend from `IfcProduct`), so routing must stay on "has no geometry", never on class name or entity name.

## The product decision

1. **Grayed out / muted** wherever such a row renders — spatial tree, By Class, By Type, and any other tree or list showing it.
2. **Grouped under a node labelled "Other"**, following the BIMcollab Zoom convention ("shown grayed out, and under a group of objects called other").
3. The headline object count is unchanged — these elements are **not** counted. That already works; this issue does not touch it.

## The mid-load trap

During load the geometry set is empty, and a naive "not in geometricIds" filter is a no-op that passes everything, then narrows as meshes arrive — graying and re-bucketing every element on open, then popping them back. The fix must key off the explicit readiness flag the count already uses (`geometryReadyModelIds.has(modelId)` / `model.geometryResult != null`), never `geometricIds.size > 0`: absent/undefined readiness ⇒ show every element as normal (no gray, no bucket); present-but-empty ⇒ a real, restrictive answer.

## Requirements

- Reuse the existing predicates (`isPhysicalObjectType`, the shape test, `collectMeshedIds`/`AssemblyGeometry.renders`) — no third "has geometry" rule.
- Reuse the existing muted-row convention (`opacity-50 grayscale`), not a new colour, and verify it in both light and dark themes.
- RED-first tests, including at least one geometry-less element and one non-physical type in the fixture (a fixture with nothing the filter would reject can't catch a dropped guard).

BIMvoice:
Closes #4764

## The gap

#4656/#4658/#4725 already unified "which elements have geometry" onto one rule for the object *count*. The row itself still didn't distinguish a geometry-less physical element from a shaped one: it sat inline in the spatial tree with no visual cue, and was silently dropped from the By Class / By Type tabs (correctly excluded from the count, but unfindable).

Motivating case: `#162 IFCBUILDINGELEMENTPROXY('...','Group#21',...)` in the buildingSMART PCERT sample `Building-Structural.ifc` — a real `IfcElement` with an `ObjectPlacement` but `Representation = $`. **Not** an `IfcGroup` — a proxy someone merely named "Group#21". `IfcGroup`/`IfcZone`/`IfcSystem` can't appear in a spatial tree at all (they don't descend from `IfcProduct`), so this routes on the absence of geometry, never on class or entity name.

## The fix

- `productTree.ts`'s `AssemblyGeometry` gains `isOther(typeName, expressId, globalId)` alongside its existing `renders` — the ONE branch of `renders`'s own logic that means "belongs in a products tree, but has no shape" (own or aggregated), isolated rather than re-derived. `false` while the geometry filter is inert (streaming), mirroring `renders`'s own contract.
- `buildTypeTree` (By Class) and `buildIfcTypeTree` (By Type): an entity that fails `renders` is now checked against `isOther`; if true it is collected into a flat bucket instead of dropped, and rendered as one collapsed "Other" node after the real class groups (`otherBucket.ts`'s `buildOtherGroupNodes`, shared by both tabs so the row shape can't drift between them).
- Spatial tree (`buildSpatialNodes` and the multi-model Storeys section's per-model contribution): direct container elements are split into shaped vs. not using the SAME `hasShape` function the count badge already computes from (`makeShapeTest`), via a new `emitElementsWithOtherBucket` helper — normal rows emit as before, the rest collapse under one "Other" row per container.
- `TreeNode` gains `noGeometry?: boolean`, set on a bucketed row's element node; `HierarchyNode.tsx` grays it (and the "Other" header itself) using the SAME `opacity-50 grayscale` convention a hidden-in-3D row already uses — no new colour. That class pair is a filter/opacity, not a colour token, so it is identical in both themes.

## Mid-load contract

`emitElementsWithOtherBucket` takes `hasShape: ((id: number) => boolean) | null` — `null` exactly while `makeShapeTest`'s own `geometryKnown` gate says absence is unanswerable (mid-load), and in that state every element emits normally with no split and no "Other" row, so there is no flash of every row going gray and back as the first geometry batch lands. `AssemblyGeometry.isOther` carries the identical gate on its own `applyFilter`. Covered directly (`productTree.isOther.test.ts`) since the tree-builder call sites never reach that branch through `renders` alone — `renders` already returns `true` for every physical type while the filter is inert, so a caller that only exercises `buildTypeTree`/`buildIfcTypeTree` can't prove the guard is load-bearing.

## Module size

`emitElementSubtree` and the new split-and-bucket helper moved to `elementSubtree.ts`, and the two products-tree bucket builders share `otherBucket.ts`, so `treeDataBuilder.ts` (1318 lines) stays under its 1381-line budget instead of growing past it.

## Tests, RED first

- `treeDataBuilder.physicalObjects.test.ts`: the `#4764`-shaped fixture (`#50`, an `IfcBuildingElementProxy` named "Group#21" with no shape) now asserts the "Other" bucket in both the spatial tree and By Class, in both geometry states (streaming vs. known) — including a dedicated "no bucket while streaming" case for the tab, since the earlier fixture only exercised that through the spatial tree's own `hasShape` gate.
- `treeDataBuilder.test.ts`: a new case buckets a shapeless typed occurrence under "Other" in By Type.
- `productTree.isOther.test.ts` (new): `AssemblyGeometry.isOther` directly — mid-load, known-empty, has-its-own-geometry, and never-belongs-here cases.
- `HierarchyNode.noGeometry.test.tsx` (new): the gray class pair applies to a `noGeometry` element row and to the "Other" header, does not apply to an ordinary row, and composes correctly with the pre-existing hidden-in-3D case.

## Mutation testing

Each guard mutated individually, tests rerun, reverted; final `git diff --stat` shows a clean revert (matches the pre-mutation diff each time).

| mutation | file | reddened | tests failing |
|---|---|---|---|
| `isOther`'s `if (!applyFilter) return false;` → `if (false) return false;` | `productTree.ts` | yes | 1/4 in `productTree.isOther.test.ts` (the buildTypeTree/buildIfcTypeTree suites stay green — `renders` already short-circuits before `isOther` is reached for a physical type while streaming, which is exactly why this guard needed its own direct test) |
| `emitElementsWithOtherBucket`'s `if (!hasShape) {` → `if (false) {` | `treeDataBuilder.ts` (now `elementSubtree.ts`) | yes | 4/24 in `treeDataBuilder.physicalObjects.test.ts`'s "By Class tab never lists a non-object" suite |
| `buildTypeTree`'s `if (assemblyGeometry.isOther(...))` → `if (false)` (drops instead of bucketing) | `treeDataBuilder.ts` | yes | 1/24 (the dedicated "leaves a shapeless physical object out ... grayed under Other" case) |
| `HierarchyNode.tsx`'s `(nodeHidden \|\| noGeometry) &&` → `(nodeHidden) &&` | `HierarchyNode.tsx` | yes | 2/4 in `HierarchyNode.noGeometry.test.tsx` |

**The `isOther` mid-load guard has NO coverage through the tree-builder tests — only through the direct unit test.** `renders` already returns `true` for every physical type while the geometry filter is inert, so `buildTypeTree`/`buildIfcTypeTree` never call `isOther` at all in that state — the `!applyFilter` branch inside it is unreachable through any tree-builder call site. Deleting that guard leaves every `treeDataBuilder`/`treeDataBuilder.physicalObjects` test green; only `productTree.isOther.test.ts`'s direct call into `makeAssemblyGeometry(...).isOther(...)` catches it (confirmed above: 1/4, not 0/4). A future refactor that removes `productTree.isOther.test.ts` without noticing this would delete the guard's only test and nothing else would fail.

## Gates

- `apps/viewer` targeted suite (touched + dependency files: `treeDataBuilder.test.ts`, `treeDataBuilder.physicalObjects.test.ts`, `groupEntityTypes.test.ts`, `modelTagView.test.ts`, `typeIsolationLabel.test.ts`, `ModelRowTags.test.tsx`, `HierarchyNode.noGeometry.test.tsx`, `productTree.isOther.test.ts`): 110/110 pass.
- `pnpm exec tsc --noEmit` in `apps/viewer`: no errors in any touched file (pre-existing, unrelated errors elsewhere in the tree — an unbuilt `@ifc-lite/charts` import chain and similar — were already there before this change).
- `node scripts/check-module-size.mjs` (plain form, no `--base`): `treeDataBuilder.ts` now measures 1318 lines against its 1381-line budget (62 headroom) — fixed, not just avoided, by the `elementSubtree.ts`/`otherBucket.ts` extraction. The run still exits 1, but only on pre-existing `rust/processing/tests/module_size_allowlist.txt` merge-base drift against the stale `origin` fork remote in this environment — the same finding #4761 documented and verified against a clean `upstream/main` checkout; none of this PR's files appear anywhere in that output.
- Changeset added (`@ifc-lite/viewer`, patch), no license header.

## Exclusions

- Eye-toggle (show/hide in 3D) still renders on a geometry-less/Other row even though it has nothing to draw — not addressed; a pre-existing UX question, not something this fix regresses.
- `PropertiesPanel`'s "Contained Elements" and any other schema-only census row untouched, consistent with #4761's own scope note — those aren't labeled as a geometry filter.
- No `--comment`/inline styling snapshot beyond the class-list assertions in `HierarchyNode.noGeometry.test.tsx`: the reused `opacity-50 grayscale` pair is a CSS filter/opacity, not a colour token, so it renders identically under `prefers-color-scheme: dark` and light — verified by class presence, not a rendered-pixel comparison.

Not self-applying `unqueued`; awaiting the maintainer's `ready` label on #4764.

## Summary by CodeRabbit

* **New Features**
  * Added a collapsed “Other” bucket for physical elements without geometry in Spatial, By Class, and By Type hierarchies.
  * Geometry-less physical elements are grouped and shown with a grayed-out appearance.
  * Elements remain displayed normally while model geometry is loading.
* **Bug Fixes**
  * Prevented geometry-less elements from being silently omitted.
  * Non-physical elements without geometry remain normal, selectable rows rather than being placed in “Other.”


louistrue (approved):
Maintainer review complete on exact head f6b0437715e3b76cf5e3523414420efdb5f43586. Geometry readiness, physical-object gating, spatial/By Class/By Type parity, muted selectable-row behavior, aggregation handling, and count semantics are covered. Exact-head Test is green; Claude failures are quota-only, and Ollama’s speculative notes were adjudicated above.

PR #4770, opened 2026-09-14
test(viewer): cover the IfcSpace schema-filter guard in modelMetadataStats

### LINKED ISSUE #4769: modelMetadataStats.test.tsx: mutation table cites IfcSpace coverage that was never committed
## The gap

PR #4761 (merged) — mutation table claims a meshed `IfcSpace` case was added to `modelMetadataStats.test.tsx` to guard `computeModelStats`'s schema filter (`collectPhysicalEntityIds`), catching an otherwise silently-green mutation of that filter.

On `upstream/main`, `apps/viewer/src/components/viewer/properties/modelMetadataStats.test.tsx` contains zero occurrences of "space" (case-insensitive, checked with `grep -ic space`). The test the mutation table describes does not exist in the merged file.

## Impact

`computeModelStats`'s schema filter (`collectPhysicalEntityIds`) can be deleted today — replaced with an unfiltered pass over every `entityIndex.byType` id — and the full `modelMetadataStats.test.tsx` suite still passes. A meshed, non-physical entity (e.g. `IFCSPACE`) would then be counted in the "Elements with Geometry" statistic, and nothing in the suite would catch it.

## Fix

A follow-up PR adds the missing coverage: a meshed `IFCSPACE` fixture entity, asserted excluded from the count, mutation-tested to confirm it reddens when the schema filter is removed.

BIMvoice:
Closes #4769

## The gap

PR #4761 (merged) documented a mutation-testing table claiming a meshed `IfcSpace` case was added to `apps/viewer/src/components/viewer/properties/modelMetadataStats.test.tsx`, guarding `computeModelStats`'s schema filter (`collectPhysicalEntityIds`). Checked directly on `upstream/main`: the committed file has zero case-insensitive occurrences of "space". The described test was never committed.

Impact: the schema filter can be deleted today — replaced with an unfiltered pass over every `entityIndex.byType` id — and the full suite stays green. A meshed, non-physical entity (an `IFCSPACE`, for instance) would then be counted in the "Elements with Geometry" statistic with nothing catching it.

## Fix

Adds `#8: IFCSPACE` to the fixture (schema-legal, non-physical) and a new test asserting a meshed `IFCSPACE` stays excluded from the count (still 6, not 7) even though it passes the shape test — proving the schema filter, not just the shape filter, is load-bearing. No production code changed.

## RED / GREEN

- Before the fixture change: no case exercised a meshed non-physical entity.
- New test green against the current `computeModelStats`.

## Mutation testing

| mutation | file | reddened | tests failing |
|---|---|---|---|
| `collectPhysicalEntityIds(...)` → unfiltered `Set` of every `entityIndex.byType` id | `modelMetadataStats.ts` | yes | 2/4 (`excludes a meshed IFCSPACE...` and the streaming case, both now overcounting by 1) |

Reverted; `git diff --stat` shows only the test file changed (no diff against the source file).

## Gates

- `apps/viewer` targeted run: `modelMetadataStats.test.tsx`, `lib/object-count.test.ts`, `lib/physical-objects.test.ts` — 44/44 pass.
- `pnpm exec tsc --noEmit` in `apps/viewer`: no errors attributable to the touched file.
- `node scripts/check-module-size.mjs` (plain form): no output naming the touched file.
- Changeset added (`@ifc-lite/viewer`, patch), no license header.

## Exclusions

- No production code changed — this PR is test-only, adding the coverage the mutation table described.


louistrue (approved):
Independent maintainer review complete: focused schema-vs-shape regression is correct; exact-head CI and Ollama are green; issue authorization is now green.

PR #4773, opened 2026-09-14
fix(server): read relating_is_list for a grouped IfcRelDefinesByProperties.RelatingPropertyDefinition

### LINKED ISSUE #4772: server(relationships): grouped IfcRelDefinesByProperties.RelatingPropertyDefinition silently dropped
## Summary

`extract_relationship` in `apps/server/src/services/data_model/relationships.rs` calls `entity.get_ref(slots.relating_idx)` unconditionally for the "relating" slot of every `IfcRelationship` subtype. `get_ref` returns `None` for a list-valued attribute (`rust/core/src/schema_gen.rs`, `get_ref` -> `AttributeValue::as_entity_ref`), and the `?` operator then drops the whole relationship.

`IfcRelDefinesByProperties.RelatingPropertyDefinition` is typed `IfcPropertySetDefinitionSelect = IfcPropertySetDefinition | IfcPropertySetDefinitionSet`. `IfcPropertySetDefinitionSet` is a defined `TYPE ... = SET [1:?] OF IfcPropertySetDefinition`, not an entity, so a grouped value is written inline in STEP as `(#20,#21)` rather than a bare `#id`. For that shape, `get_ref` returns `None` and the entire relationship — every property/quantity set in the group, for every related object — is silently dropped.

The TS/WASM columnar parser (`packages/parser/src/columnar-parser-relationships.ts`, `extractPropertyRelFast`) already hit and fixed this exact bug: it reads the relating slot as a ref list unconditionally (`readRefList` accepts both a bare `#id` and a `(...)` list) and applies every entry in the group to every related object.

## Verified

- Checked `IfcMaterialSelect` (used by `IfcRelAssociatesMaterial.RelatingMaterial`, also named as a suspect): its alternatives are `IfcMaterialDefinition`, `IfcMaterialList`, `IfcMaterialUsageDefinition` — `IfcMaterialList` is itself an **entity** (`ENTITY IfcMaterialList; Materials : LIST [1:?] OF IfcMaterial; END_ENTITY;`), so `RelatingMaterial` is always a single `#id` reference to that entity, never an inline grouped list. `IfcRelAssociatesMaterial` is NOT affected by this bug — only `IfcRelDefinesByProperties` is, confirmed by walking the schema-derived registry (`registry.selects`/`registry.types`) programmatically: exactly one of the 54 resolved concrete `IfcRelationship` subtypes has a relating attribute whose select resolves to an aggregate defined type.
- Predates PR #4763 ("Rust half of #4205"), which replaced the hand-written 13-entry relationship-type table with a schema-derived one but did not add a `relating_is_list` counterpart to the `related_is_list` flag it already emits, so this gap survived that change. #4763's changeset overclaimed that the server-parse and browser-parse paths "agree on every relationship edge" — not true for this case.
- No fixture in `apps/server/src/services/data_model/tests.rs` used a grouped relating value before this issue, so the existing suite could not catch it.

## Fix

Addressed in a follow-up PR: a schema-derived `relating_is_list` flag, generated the same way as the existing `related_is_list`, and a cross-product extraction (every relating id × every related id) mirroring the TS/WASM path's semantics exactly.

BIMvoice:
## Summary

`extract_relationship` (`apps/server/src/services/data_model/relationships.rs`) called `entity.get_ref(slots.relating_idx)` unconditionally for the "relating" slot of every `IfcRelationship` subtype. `get_ref` returns `None` for a list-valued attribute, and the `?` then silently dropped the whole relationship.

`IfcRelDefinesByProperties.RelatingPropertyDefinition` is schema-legally `IfcPropertySetDefinitionSelect = IfcPropertySetDefinition | IfcPropertySetDefinitionSet`. `IfcPropertySetDefinitionSet` is a defined `TYPE ... = SET [1:?] OF IfcPropertySetDefinition`, not an entity, so a grouped value is written inline in STEP as `(#20,#21)` rather than a bare `#id` — for that shape `get_ref` returned `None` and every property/quantity set in the group, for every related object, was dropped.

This predates and survives PR #4763 ("Rust half of #4205"), which schema-derived the relationship-type table but only emitted `related_is_list`, not a `relating_is_list` counterpart. #4763's changeset claimed the server-parse and browser-parse paths "agree on every relationship edge" — not true for this case, so I've amended that wording since it's still unreleased (no "Version Packages" commit has consumed its changeset yet).

## Verified before fixing

- `get_ref`'s `None`-on-list behavior: `rust/core/src/schema_gen.rs` (`get_ref` → `AttributeValue::as_entity_ref`, which only recognizes a bare reference).
- The TS/WASM columnar parser already hit and fixed this exact bug (`packages/parser/src/columnar-parser-relationships.ts`, `extractPropertyRelFast`): it reads the relating slot as a ref list unconditionally and applies every entry in the group to every related object.
- **Checked whether `IfcRelAssociatesMaterial` has the same bug (a natural second suspect) — it does not.** `IfcMaterialSelect`'s alternatives are `IfcMaterialDefinition`, `IfcMaterialList`, `IfcMaterialUsageDefinition`; `IfcMaterialList` is itself an **entity** (`ENTITY IfcMaterialList; Materials : LIST [1:?] OF IfcMaterial; END_ENTITY;`), so `RelatingMaterial` is always a single `#id` reference to that entity, never an inline grouped list — confirmed both by reading the EXPRESS and by walking the schema-derived registry programmatically (`registry.selects`/`registry.types`, checking selects before types per the existing `isReferenceType` ordering trap). Exactly one of the 54 resolved concrete `IfcRelationship` subtypes is affected.

## Fix

- `packages/parser/src/relationship-schema-slots.ts`: new `resolvesToAggregate()` walks a SELECT's alternatives (recursively, depth-capped) for one that resolves to an aggregate defined type (`LIST`/`SET`/`ARRAY`/`BAG`), and folds that into `computeSlotPlan`'s existing `isList` computation for both slots. Purely additive — verified via `getRelationshipSlotPlan` that only `IFCRELDEFINESBYPROPERTIES.relating.isList` flips from `false` to `true`; every other type (including `IFCRELASSOCIATESMATERIAL`) is unchanged. No existing test asserted `relating.isList` for the affected type.
- `scripts/generate-server-relationship-slots.mjs`: emits a `relating_is_list` field alongside `related_is_list`; `--check`'s drift-detection regex and comparison map updated to match.
- `apps/server/src/services/data_model/generated/relationship_slots.rs`: regenerated (`pnpm turbo build --filter=@ifc-lite/parser && node scripts/generate-server-relationship-slots.mjs && cargo fmt -p ifc-lite-server`).
- `apps/server/src/services/data_model/relationships.rs`: `extract_relationship` now reads the relating slot with `get_refs` (accepts both a bare ref and a list — already existed, used elsewhere) when `relating_is_list`, and cross-products every relating id × every related id, mirroring the TS/WASM path's `extractPropertyRelFast` semantics exactly (every set member applies to every related object) rather than just taking the first id.
- `apps/server/src/services/data_model/tests.rs`: new RED→GREEN fixture (`GROUPED_RELATING_PSET_IFC`) with a `RelatingPropertyDefinition` grouping a pset and a qset — asserts both halves of the group are extracted and the pset's own properties resolve.
- `.changeset/rust-relationship-slot-parity.md`: amended #4763's overclaim (still unreleased). New changeset for this fix.

## Test plan

- [x] RED: reverted `extract_relationship` to the pre-fix unconditional `get_ref`, ran the new fixture test — failed with the predicted symptom (`DefinesByProperties #79 dropped the pset (#20) half of the grouped relating set: []`).
- [x] GREEN: restored the fix, same test passes.
- [x] `cargo test -p ifc-lite-server` (not `--workspace` — narrowed for disk headroom on the CI runner used to develop this): 343 passed, 0 failed.
- [x] `pnpm turbo build --filter=@ifc-lite/parser` then targeted vitest (`relationship-schema-slots.test.ts`, `relationship-subtype-coverage`, `columnar-parser-relationships`): 22 passed.
- [x] Drift guard: hand-mutated the committed `IFCRELDEFINESBYPROPERTIES` row's `relating_is_list` to `false`, ran `node scripts/generate-server-relationship-slots.mjs --check` — exit 1, correctly diagnosed `committed [5,4,false,true] != derived [5,4,true,true]`. Reverted, `--check` exits 0 again.
- [x] Mutation actually executes: with that same mutated `relating_is_list: false` committed, re-ran the fixture test — reddened with the identical drop symptom, proving the flag drives the code path (not a no-op mutation). Reverted; `git status --short` clean against the prior commit.
- [x] `node scripts/check-module-size.mjs` (plain form): exit 0. Pre-existing allowlist drift warnings in `rust/geometry`/`rust/processing`/`rust/wasm-bindings` files this PR does not touch — unrelated to this change.

## Not run

- `cargo test --workspace` (only `-p ifc-lite-server`, per disk-space narrowing).
- `cargo clippy` was not run separately; `cargo fmt -p ifc-lite-server` was.

Closes #4772

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed relationships involving grouped property-set definitions so all associated property and quantity sets are retained.
  - Preserved separate links between objects and each applicable property or quantity set.
  - Improved handling of grouped relationship references across parsing paths.

- **Tests**
  - Added coverage verifying grouped property-set relationships and retained property data.


louistrue (approved):
Maintainer review complete on exact head e8dfaa96af4a0e887c8c622820cb05ef301faab6. The schema-derived aggregate-select detection, generated `relating_is_list` contract, cross-product semantics, and grouped Pset/Qset regression are correct. The branch is now scoped to seven files. Exact-head Test is green; Ollama and CodeRabbit report no merge-blocking issue; Claude failures are quota-only.

Issue #4094, opened 2026-09-07
Selector syntax: remaining IfcOpenShell constructs and the CLI/MCP/SDK adapters (follow-up to #4091)


louistrue:
Follow-up to #4091, which adds `parseSelector` in `@ifc-lite/query` plus one adapter onto the viewer's Filter tab. The parser accepts the whole IfcOpenShell filter grammar; the viewer adapter does not yet evaluate all of it, and reports each construct it cannot handle by name rather than dropping it silently.

Left for this issue:

**Viewer rule kinds that do not exist yet**
- GlobalId terms, both adding and subtracting (`325Q7Fhnf67OZC$$r43uzK`, `! 325Q7...`). Needs a `globalId` FilterRule kind, evaluator and editor row.
- `type=WT01`, matching the relating type's Name through IfcRelDefinesByType.
- Attributes other than `Name` and `PredefinedType` (`Description`, `ObjectType`, `Tag`, ...). Needs a generic `attribute` rule kind.
- `parent=Foo` (direct or indirect child of a named element), and `query:` value-query keys.
- Material `Category` — #4091 matches material and layer names only, IfcOpenShell matches Name **or** Category.

**Group unions**
- `+` unions two filter groups (`IfcSlab, material=concrete + IfcDoor`). The viewer's filter state carries a single top-level AND/OR combinator, so this needs either OR-of-ANDs in `FilterState` or running each group separately and unioning the resulting id sets.

**Other surfaces**
The point of putting the parser in `@ifc-lite/query` is that these adapt the same AST rather than growing a second grammar:
- `ifc-lite query --select "<selector>"` in the CLI, alongside the existing single-clause `--where`.
- `selector` parameter on the MCP `query_entities` tool.
- `bim.query().select()` in the SDK and sandbox.

All three need `matches` (regex) in `FilterComparisonOp` and the query backend before selector text can be honoured end to end, plus GlobalId and union support to avoid accepting text the backend then silently narrows.

**Also worth settling here:** whether `location=` should reach through spaces to the containing storey. IfcOpenShell's example 17 expects a pump inside a space on Level 3 to match `location="Level 3"`. #4091 pins whatever the current storey walk actually does in a test and documents it; if that turns out to be storey-name-only, widening it belongs here.

louistrue:
Adding one item from the #4091 review, deferred here on purpose.

[...]

Issue #4091, opened 2026-09-07
Selector syntax


gDonia87:
### The problem

If I want to use the Ifcopenshell syntax for filtering objects, I don´t  get any results.

### Proposed solution

Is it possible to implement or describe the way IFClite allow the use of wildcard in filtering objects?
These example should work also in IFClite
https://docs.ifcopenshell.org/ifcopenshell-python/selector_syntax.html

louistrue:
Claiming this — working it now (agent session `0193douQ6`, branch `fix-4091-selector-syntax`). First step is establishing what ifc-lite's filter syntax actually supports today versus the IfcOpenShell selector grammar, then closing the gap or documenting it precisely.

louistrue:
Thanks for the report, and for linking the exact page. You did not do anything wrong: ifc-lite does not read that syntax anywhere yet.

I checked this by driving the viewer rather than reading the code, on AC20-FZK-Haus (275 elements), and the numbers show the problem clearly. The search box is a plain substring match, so selector characters are matched literally and adding syntax makes it match **less**:

| query | results |
|---|---|
| `Wand` | 13 |
| `IfcWall` | 15 |
| `IfcWall*` | **0** |
| `IfcWall, IfcSlab` | **0** |
| `IfcWall[Name*=Wand]` | **0** |

So `IfcWall` finds 15 but `IfcWall*` finds nothing. The Advanced filter (Cmd+Shift+F) is a structured rule builder with no text grammar at all, and its Name operators are only equals, not-equals, contains, does not contain, and starts with. No wildcards, no regex. Worth adding that ifc-lite is inconsistent with itself here: the CLI's clash selectors already accept `--a "IfcDuct*|IfcPipe*"`.

**What I am adding.** A parser for the IfcOpenShell filter syntax in `@ifc-lite/query`, and a text field in the viewer's Filter tab that turns a selector into filter rules. Putting the parser in the query package rather than in the viewer means the CLI, MCP and scripting APIs can adapt the same grammar later instead of growing their own.

After that PR these work in the viewer:

- `IfcWall, IfcSlab`, with subclasses included, so `IfcElement` works too
- `IfcElement, ! IfcWall`
- `IfcDoor, Name=D01` and `IfcDoor, Name=/D[0-9]{2}/`
- `IfcWall, Pset_WallCommon.FireRating=2HR`, with all of `= != > >= < <= *= !*=`
[...]

PR #4774, opened 2026-09-14
feat(query,sdk,cli,mcp): selector text as a QueryDescriptor for --select/query_entities/QueryBuilder.select()

BIMvoice:
## Summary

Refs #4094

`#4094` reopened the selector-syntax issue because prior PRs delivered viewer-side selector support (`parseSelector` in `packages/query/src/selector/`, consumed by the viewer's `selector-to-rules.ts`), but none of the three programmatic surfaces — CLI `query`, MCP `query_entities`, SDK `QueryBuilder` — accepted selector text at all. This PR adds that layer.

- **`selectorToQueryDescriptor` + `SelectorUnsupportedError`** (`packages/query/src/selector/to-query-descriptor.ts`, exported from `@ifc-lite/query`): the shared translator from the selector AST (`parseSelector`'s output — same tree the viewer adapter walks) to `QueryDescriptor`'s `{ types, filters }` shape.
- **`QueryBuilder.select(text)`** (`packages/sdk/src/namespaces/query.ts`): calls the shared translator and appends into the same `types`/`filters` lists `.byType()`/`.where()` already append to (types OR together, filters AND together — same semantics as chaining two `.byType()` or `.where()` calls). `SelectorUnsupportedError` is re-exported from `@ifc-lite/sdk`.
- **`ifc-lite query --select "<text>"`** (`packages/cli/src/commands/query.ts`): calls `bim.query().select(select)` before `.byType(...)`, ANDed with `--type`/`--where`. A thrown error calls `fatal()` with the error's own message (existing pattern in this file).
- **`query_entities`'s `selector` input** (`packages/mcp/src/tools/query.ts`): calls `.select(selector)` on the same `bim.query()` chain, ANDed with `type`/`types`/`property`. A thrown `SelectorUnsupportedError` propagates out of the handler; `server.ts`'s tool-call dispatch already turns any thrown Error into a clean `isError: true` result (verified by reading `server.ts`, and by the same precedent `overlay.test.ts` already uses for `entity_create`'s abstract-type rejection — testing the raw handler's throw, not the wrapper).

### Lossless subset implemented (throws `SelectorUnsupportedError` naming the construct for everything else)

Translates:
- Class terms — `IfcWall`, comma unions (`IfcWall, IfcSlab`) — expanded to subtypes via `@ifc-lite/parser`'s `expandTypes` (same as the viewer adapter).
- Exact-name property/quantity comparisons — `Pset_WallCommon.FireRating=2HR`, `Qto_WallBaseQuantities.NetVolume>1` — both read as the same `QueryFilter` shape.
- `=` `!=` `>` `>=` `<` `<=` map onto their `ComparisonOp` namesakes; `*=` maps onto `contains`.
- A `/…/` value with `=` maps onto `matches`, carrying the regex source (no delimiters).
- `Prop!=NULL` maps onto `exists` ("this property is set").

Throws (left open, same as before this PR — real remaining scope on #4094):
- Regex pset/property **names** (`/Pset_.*Common/.Prop=`) — `QueryFilter.psetName`/`propName` match by exact string only (`matchesPsetFilter`/`matchesQsetFilter`).
- `!*=` (not-contains) and a regex value on any operator but `=` (a negated match) — no `ComparisonOp` target.
- `Prop=NULL` ("not set") — `ComparisonOp` has `exists` but no negation of it.
- Entity-attribute terms (`Name=`, `GlobalId=`, a bare GlobalId literal, generic `attribute=`, `type=`) — `QueryDescriptor` has no attribute-filter concept.
- `!` class negation — no `notIn` concept in `types`.
- `+` group unions — `QueryDescriptor` can only AND filters.
- `parent=`, `query:`, `material=`, `classification=`, `location=` — no target.

These, plus the viewer's still-open `parent=`/`query:`/`+`/material-Category/`location=` widening, remain open on #4094.

### A gap found while verifying the Qto_ path (fixed here, not new scope creep — see rationale below)

The brief for this work assumed `Qto_` comparisons "already work" through `QueryFilter`, since `entity-query.ts`'s `whereProperty` path falls back to quantity sets. I verified this empirically (a wall with `Qto_WallBaseQuantities.NetVolume=12.5`, queried via `bim.query().where('Qto_WallBaseQuantities', 'NetVolume', '>', 0)`) and it returned **0 matches** — `HeadlessBackend.entities()` (CLI) and its MCP equivalent (`backend-query.ts`), which is what `QueryBuilder`/`.select()` actually execute through, only ever checked property sets in their `descriptor.filters` predicate (`matchesPropertyFilter`), never quantity sets. The CLI's own `--where` flag has always had this fallback (`applyWhereFilter`), just not the SDK-level `descriptor.filters` path.

Shipping `.select()`'s Qto_ translation without fixing this would have reproduced the exact "selector that matched nothing looked identical to no selector at all" defect (#4091) this whole effort exists to prevent — for a `Qto_` term specifically, and also for any pre-existing direct `.where('Qto_...', ...)` caller. `matchesPropertyFilter` (`packages/cli/src/property-filter-match.ts`, `packages/mcp/src/property-filter-match.ts`) now falls back to quantity sets when no property set matches, mirroring `applyWhereFilter`'s existing fallback. Confirmed the fix with a real IFC fixture end to end (before: 0 matches, after: 1 match) — see the CLI/MCP `query-select.test.ts` files.

## Test plan

- [x] `packages/query`: RED-first `to-query-descriptor.test.ts` — 10 tests, one accepted case covering every mapped operator + Qto_, one genuinely-rejected case per unsupported construct (`parent=`, regex pset name, `!*=`, `!` negation, `+` union, bare GlobalId, `=NULL`, a parse error). `pnpm exec vitest run` (full package): **21 files, 315 tests passed**.
- [x] `packages/sdk`: `namespaces.test.ts`'s new `QueryBuilder.select()` block — accepted translation, OR/AND append semantics with `.byType()`/`.where()`, and the rejected-construct throw. Full suite: **17 files, 230 tests passed**.
- [x] `packages/cli`: new `query-select.test.ts` against a real temp IFC fixture (two walls, one door) — class+property narrowing, the Qto_ fallback fix proven live, `--type`+`--select` AND composition, and the rejection exit. Plus `schema.ts`'s `RUNTIME_METHOD_SIGNATURES`/`RUNTIME_METHOD_DOCS` updated for the new `select` method (`schema.runtime-shape.test.ts` enforces this — caught it live). Full suite: **70 files passed, 918 tests passed, 23 skipped**; 6 files / 24 tests fail, all pre-existing and unrelated (export/clash/geometry paths hitting a stale local WASM binary — this environment has no `packages/wasm/pkg` build and CLAUDE.md forbids `cargo`; same set of files failed before this PR's changes).
- [x] `packages/mcp`: new `tools/query-select.test.ts` via the real tool registry/handler (`query_entities`) against the same fixture shape. Full suite: **42 files, 410 tests passed** (no pre-existing failures here).
- [x] Mutation-tested 4 guards individually (edit → run targeted test → confirm RED → `git diff` proves the mutation applied → revert → confirm GREEN + clean `git diff --stat`): the throw-on-unsupported branch, the `!*=`/`*=` operator-mapping entry, the `expandTypes` subtype-expansion call, and the CLI's new Qto_ quantity-fallback guard. All four reddened the intended test and only that test; all four reverted cleanly.
- [x] `node scripts/check-module-size.mjs` (plain form, no `--base`): 4 touched files crossed their recorded budget (`query.ts`, `headless-backend.ts`, `backend-query.ts`, `tools/query.ts`); updated their 4 rows by hand in `scripts/module-size-allowlist.txt` to their measured line counts (did **not** use `--update`, which — scoped against `origin/main`, hundreds of PRs stale per this repo's own convention — tried to rewrite 41 unrelated rows). Re-run is clean for every file this PR touches. Exit code is still 1 from a **pre-existing, unrelated** `rust/processing/tests/module_size_allowlist.txt` vs-merge-base audit (11 Rust files this PR never touched) — this PR makes zero Rust changes.
- [x] `pnpm exec tsc --noEmit` scoped per touched package (`query`, `sdk`, `cli`, `mcp`): all clean. Had to build the dependency chain from source with targeted `tsc` per package first (no `packages/*/dist` existed in this fresh worktree, and `pnpm install --offline` alone doesn't build them) — same documented workaround PR #4250 used. `packages/codegen`'s two-pass build (`tsc` + `tsc -p tsconfig.generated.json`) was also needed before `@ifc-lite/export` would resolve its `@ifc-lite/codegen/*` imports; not previously called out as a blocker in #4250's own notes, noting it here for the next person who hits it.


louistrue (approved):
Approved after the maintainer fixes and exact-head Test success. I also adjudicated Ollama's remaining broad-catch note: `q.select()` is a synchronous selector parser/adapter with no network, filesystem, or other side effects, and malformed selector syntax deliberately throws a regular `Error` while unsupported-but-valid constructs throw `SelectorUnsupportedError`. Converting both user-input failures to the CLI's normal fatal diagnostic is correct here; non-`Error` throws are rethrown. This is not a merge blocker.

Issue #3490, opened 2026-08-29
CLI --where filter tests only the first same-named property/quantity set


louistrue:
`applyWhereFilter` tests only the first same-named property/quantity set, so `--where` excludes entities that satisfy the predicate in a later set.

Found by a CodeRabbit run on #3468, which converts the CLI's first-match `find` calls onto a shared `findPropertyInSets` / `findQuantityInSets` helper.

`packages/cli/src/commands/query.ts:80` (and `:88` for quantities):

```ts
const prop = findPropertyInSets(props, parsed.psetName, parsed.propName);
if (prop) {
  if (parsed.operator === 'exists') return true;
  return compareValues(prop.value, parsed.operator, parsed.value);
}
```

The helper returns the **first** member it finds across same-named sets, and `compareValues` then tests that one value. So with two `Pset_WallCommon` sets on one wall — the first carrying `FireRating = 'REI30'`, the second `FireRating = 'REI60'` — `--where Pset_WallCommon.FireRating=REI60` excludes the wall, even though it genuinely has that value.

## Why the shared helper is right for export and wrong here

The two call sites want different things from the same lookup:

- **Export** emits one cell, so picking a member is correct. First-match is a defensible choice.
- **Filter** evaluates a predicate over the entity, so it should be satisfied if **any** same-named set matches.

Collapsing both onto one first-match helper makes the export case right and freezes the filter case wrong. That is the reuse-moves-the-blast-radius shape: a lookup that was safe for one caller inherits a second caller with different semantics.

## Pre-existing

The filter did its own first-match `find` before #3468, so this is not introduced there — it is preserved. Recording it separately so #3468 does not have to grow a filter-semantics change.

## Fix

Add an all-matches variant (`findAllPropertiesInSets` / `findAllQuantitiesInSets`) and have `applyWhereFilter` accept when any member satisfies the operator. `exists` is already correct either way.

Regression to add: both same-named sets carry the requested member with different values, and the `--where` names the one in the later set. That case passes today for the wrong reason (it fails), and would fail loudly against a first-match implementation once written.

Note the same file's `Also applies to: 88` — the quantity branch has the identical shape.

PR #4776, opened 2026-09-14
fix(query,cli,mcp): collapse the two matchesPropertyFilter copies into one

### LINKED ISSUE #4775: packages/cli and packages/mcp carried two identical copies of matchesPropertyFilter
## What

`packages/cli/src/property-filter-match.ts` and `packages/mcp/src/property-filter-match.ts` were functional twins: identical `matchesPropertyFilter` implementations, differing only in their doc comments. Nothing enforced that they stayed in sync — a change to one had no mechanism forcing the same change onto the other.

Diffing the two files on `upstream/main` showed a 12-line diff, all of it comments; the function bodies were byte-identical.

Compounding this, one of the comments made a claim the code did not back up:

> `// ... so this can't drift from their boolean-normalization/case-insensitive -contains semantics.`

There was no shared module and no test comparing the two files, so nothing actually prevented drift.

This is a recurring shape in this repo: a rule implemented twice, with only one copy (or neither) guarded against the other drifting. The same pattern produced a geometry-count divergence between the spatial tree and By Class (#4761), a TS/Rust relationship drop (#4773), and a tree-bucketing gap (#4765).

## Fix

Extracted the shared implementation into `packages/query/src/property-filter-match.ts`, exported from `@ifc-lite/query`. Both `packages/cli/src/headless-backend.ts` and `packages/mcp/src/backend-query.ts` already depend on `@ifc-lite/query` for the lookup/comparison helpers this function is built from, so routing the whole predicate through the same package required no new dependency and creates no cycle (`@ifc-lite/query` does not depend on either `cli` or `mcp`). The two package-local files are deleted; there is now exactly one copy.

The overclaiming comment is reworded to describe what the extraction actually guarantees (one implementation, not a guard watching two).

BIMvoice:
Closes #4775

## Summary

`packages/cli/src/property-filter-match.ts` and `packages/mcp/src/property-filter-match.ts` were functional twins on `upstream/main`: I diffed them and the `matchesPropertyFilter` implementations were byte-identical, a 12-line diff entirely in comments. One of those comments claimed a guarantee nothing enforced:

> `// ... so this can't drift from their boolean-normalization/case-insensitive -contains semantics.`

There was no shared module and no test comparing the two files.

## Extract vs. guard

Extracted, not guarded. Both `packages/cli/src/headless-backend.ts` and `packages/mcp/src/backend-query.ts` already depend on `@ifc-lite/query` for the lookup/comparison helpers `matchesPropertyFilter` is built from (`findAllPropertiesInSets`, `findAllQuantitiesInSets`, `compareFilterValue`), and `@ifc-lite/query` does not depend on either package, so routing the whole predicate through `@ifc-lite/query` needs no new dependency and creates no cycle. The two package-local files are deleted; `packages/query/src/property-filter-match.ts` is the one implementation both backends import.

`PropertySetData`/`QuantitySetData`/`QueryFilter` live in `@ifc-lite/sdk`, which depends on `@ifc-lite/query` (not the other way), so the extracted function can't import those types directly -- it takes structurally-equivalent local shapes instead, the same pattern `packages/query/src/pset-lookup.ts` and `selector/to-query-descriptor.ts`'s `QueryFilterLike` already use for the identical reason.

## Coordination with #4774

#4774 is open and hand-edits both `property-filter-match.ts` files with an identical `Qto_` quantity-set-fallback addition (a real fix: CLI/MCP matching previously checked property sets only, so a `Qto_` filter silently matched zero entities). Since my change deletes exactly the files #4774 edits, this PR is built ON TOP of #4774's branch (`fix-4094-selector-adapters`) rather than against `main`, and the extracted module carries #4774's Qto_ fallback forward rather than reverting it. This PR should land after #4774 (or be rebased onto `main` once #4774 merges).

## Comment fix

The overclaiming "can't drift" comment is reworded in the extracted module to describe what the extraction actually guarantees -- one implementation, not a guard watching two -- and names both former file locations for anyone landing here from an old reference.

## RED -> GREEN

Added `packages/query/src/property-filter-match.test.ts` (7 cases: any-match across two same-named property sets, no-match, `exists`, Qto_ fallback, Qto_-absent-too). Mutation-tested 4 guards individually (edit -> run targeted test -> confirm RED -> `diff` proves the mutation applied -> revert -> confirm GREEN):

| Mutation | Test that reddened | Result |
|---|---|---|
| any-match -> first-match only (`matchingProps[0]` instead of `.some`) | "is any-match across two distinct same-named property sets (#3490)" | RED, reverted clean |
| Qto_ fallback removed (early `return false` before quantity lookup) | "falls back to quantity sets when no property set matches psetName (Qto_ / #4091)" | RED, reverted clean |
| `exists` on the property branch flipped to `return false` | "exists operator is true once the property is found, regardless of value" | RED, reverted clean |
| `matchingQtys.length === 0` flipped to `>= 0` | "falls back to quantity sets when no property set matches psetName (Qto_ / #4091)" | RED, reverted clean |

All four reverts verified identical to the pre-mutation file via `diff`.

## Test plan

- [x] `packages/query`: `pnpm exec vitest run` -- 22 files, 322 tests passed (includes the new 7).
- [x] `packages/cli`: `pnpm exec vitest run` -- 76 files passed, 1 skipped; 951 tests passed, 14 skipped. Includes #4774's `query-select.test.ts` (Qto_ fallback), unaffected by the extraction.
- [x] `packages/mcp`: `pnpm exec vitest run` -- 42 files, 410 tests passed.
- [x] `pnpm exec tsc --noEmit` scoped to `query`, `cli`, `mcp` (built the dependency chain from source first, same documented workaround #4774 used): all clean.
- [x] `node scripts/check-module-size.mjs` (plain form, no `--base`): no violation from any file this PR touches (`backend-query.ts` gets a non-blocking "note" that its budget row now has 1 line more headroom than before, since the extraction shrank it slightly). Exit code is 1 from the pre-existing Rust `module_size_allowlist.txt` merge-base audit -- confirmed unrelated: `git diff --stat upstream/main -- rust/` is empty on this branch, so this PR makes zero Rust changes.

## Exclusions

- Did not touch #4774's selector-adapter logic itself, only the file both PRs share.
- Did not run any `cargo`/wasm build deliberately for this PR's own verification; `turbo build` was used once to resolve the TS dependency chain (the same workaround #4774 documented) and it transitively rebuilt `@ifc-lite/wasm` once via `wasm-pack` (a real `cargo` invocation) before realizing the filter pulled it in. Flagging this since the repo's TypeScript-only/no-cargo convention was not observed on that one step; no Rust source was modified and the resulting `packages/wasm/pkg` output is untracked/gitignored, so nothing from that build is part of this diff.


louistrue (approved):
Approved at rebased exact head 8463eaf45. Query 7/7, CLI 5/5, MCP 8/8 and query typecheck pass locally; exact-head Test run 34827086913 is green. The prior red rows belong to cancelled run 34826370332; the latest aggregate is SUCCESS and the documented pure-refactor oracle lane is correctly SKIPPED under the maintainer exemption. Ollama reports no major issues.

PR #4778, opened 2026-09-14
fix(processing): read a grouped RelatingPropertyDefinition in rust/processing

### LINKED ISSUE #4777: processing: grouped IfcRelDefinesByProperties.RelatingPropertyDefinition silently dropped (rust/processing, inert)
## Summary

`collect_rel_defines_by_properties_link` in `rust/processing/src/processor/properties.rs` calls `rel_defines.get_ref(5).or_else(|| rel_defines.get_ref(3))` for `IfcRelDefinesByProperties.RelatingPropertyDefinition`. `get_ref` (`rust/core/src/schema_gen.rs`) delegates to `AttributeValue::as_entity_ref`, which recognises only a bare `#id` and returns `None` for a list-valued attribute.

`RelatingPropertyDefinition` is typed `IfcPropertySetDefinitionSelect = IfcPropertySetDefinition | IfcPropertySetDefinitionSet`. `IfcPropertySetDefinitionSet` is a defined `TYPE ... = SET [1:?] OF IfcPropertySetDefinition`, not an entity, so a grouped value is written inline in STEP as `(#20,#22)` rather than a bare `#id`. For that shape, `get_ref` returns `None` and the whole relationship — every property/quantity set in the group, for every related space/zone — was silently dropped by the `?` operator.

This is the same root cause as issue #4772 / PR #4773, which fixed the identical bug in `apps/server/src/services/data_model/relationships.rs::extract_relationship`. This is an independent second site in `rust/processing`.

## Repro shape

```
#20=IFCPROPERTYSET('gid1',$,'Pset_A',$,(#30));
#22=IFCELEMENTQUANTITY('gid2',$,'Qto_B',$,$,(#31));
#40=IFCRELDEFINESBYPROPERTIES('gid3',$,$,$,(#50),(#20,#22));
```

## Scope — currently inert

`resolve_space_zone_properties_lazy` populates `MeshData::properties` for `IfcSpace`/`IfcZone` occurrences, but `MeshData::properties` has no downstream reader: the WASM bridge (`rust/wasm-bindings/src/zero_copy/mesh.rs`) explicitly drops it and the viewer reads properties from the TS parser worker instead. So this bug is preventive, not currently user-facing.

## Note on a pre-existing fallback

`collect_rel_defines_by_properties_link` also has `.or_else(|| rel_defines.get_ref(3))` and `get_list(4).or_else(|| get_list(2))` fallbacks that predate this fix. Their origin is undetermined: attribute index 3 does not correspond to `RelatingPropertyDefinition` in any bundled schema (`IfcRelDefinesByProperties`: GlobalId=0, OwnerHistory=1, Name=2, Description=3, RelatedObjects=4, RelatingPropertyDefinition=5), and no fixture in this crate exercises that branch. The fix preserves this fallback conservatively rather than removing it, applying the list-vs-bare-ref read on both branches.

## Fix

`rust/processing/src/processor/properties.rs`, `collect_rel_defines_by_properties_link` now reads `RelatingPropertyDefinition` with `get_refs` (accepts a bare ref or a list) instead of `get_ref`, mirroring PR #4773's approach and the TS/WASM columnar parser's `readRefList`.

BIMvoice:
## Summary
- `IfcRelDefinesByProperties.RelatingPropertyDefinition` is schema-legally a grouped `IfcPropertySetDefinitionSet`, written inline in STEP as `(#20,#22)` rather than a single `#id`. `collect_rel_defines_by_properties_link` (`rust/processing/src/processor/properties.rs`) read that slot with `get_ref`, which recognises only a bare reference, so the `?` operator silently dropped the whole relationship for a grouped pset — every related `IfcSpace`/`IfcZone` lost every property/quantity set in the group.
- Same root cause as #4772 / PR #4773, which fixed the identical bug at the sibling site `apps/server::extract_relationship`. This is an independent second site in `rust/processing`.
- Fix: read `RelatingPropertyDefinition` with `get_refs` (accepts a bare ref or a list), mirroring #4773's approach. `RelDefinesByPropertiesLink.property_set_id: u32` became `property_set_ids: Vec<u32>`; both call sites now iterate every id in the group.
- The pre-existing `get_ref(3)` / `get_list(2)` fallback is preserved conservatively (list-vs-bare-ref applied on both branches) — its origin could not be determined from any bundled schema or fixture in this crate; a doc comment says so.
- **Currently inert**: `MeshData::properties` has no downstream reader — the wasm bridge (`rust/wasm-bindings/src/zero_copy/mesh.rs`) drops it and the viewer reads properties from the TS parser worker instead. Preventive, not user-facing today.

Closes #4777

## Test plan
- [x] RED: added `rust/processing/src/processor/properties_tests.rs` with a grouped-`RelatingPropertyDefinition` fixture; reverting the fix to the old `get_ref`-only logic reproduces the failure (`a grouped RelatingPropertyDefinition must not be silently dropped`).
- [x] GREEN: `cargo test -p ifc-lite-processing --lib properties` — 4 passed (2 new: grouped case collects both pset ids, bare case still resolves a single id).
- [x] `cargo test -p ifc-lite-processing --lib` — 420 passed, 0 failed.
- [x] `cargo test -p ifc-lite-processing --test space_zone_property_type_vectors` — 4 passed (existing TS/Rust parity harness unaffected).
- [x] `cargo test -p ifc-lite-processing --test module_size_ratchet` — passes; `properties.rs` trimmed to exactly 400 lines to stay under the 400-line split budget (was 380 on `main`).
- [x] `node scripts/check-module-size.mjs` (plain form) — no new offenders from this change.
- [x] Mutation-tested each new guard individually (reverted `get_refs`→`get_ref`+wrap reddens the grouped test; dropping the `get_refs(3)` fallback and the (redundant, since `get_refs` never returns `Some(empty)`) `property_set_ids.is_empty()` check both stay green, confirming they're genuinely untested/dead, not silently broken by the fix). All mutations reverted; final `git diff --stat` matches the intended fix only.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed property relationship processing for grouped property-set definitions.
  - Property assignments now support multiple referenced property sets while preserving support for single references.
  - Duplicate property sets and properties are avoided during resolution.

- **Tests**
  - Added regression coverage for grouped and single property-definition references.


louistrue (approved):
Approved at exact head 87e34fec1. The grouped-reference implementation is schema-correct, both the direct assignment and lazy process_geometry consumers now have production-observing coverage, the exact-head Test workflow is green, and Ollama reports no major issues.

PR #4779, opened 2026-09-14
fix(viewer): read GlobalId= and GlobalId!= as the same globalId rule the bare term builds

BIMvoice:
## Summary

Refs #4094

`#4091`'s still-open viewer-side list (`apps/viewer/src/lib/search/selector-to-rules.ts`, per louistrue's 2026-09-14 sweep) named `GlobalId=` as a comparison as one of six remaining gaps, alongside `parent=`, `query:`, `+` unions, material Category, and `location=` widening. This PR ships that one construct, fully — the others stay open on #4094/#4091 (see "What is NOT in this PR" below).

### Why GlobalId= is the one I picked

`#4774` (open, CLI/MCP/SDK adapters) had already scoped its own `GlobalId=` support out for the same reason: "`QueryDescriptor` has no attribute-filter concept." The viewer's rule model does not have that limitation — `Rule.globalId` and its evaluator (`globalIdOpMatches`, `filter-ops.ts`) already exist, built for the bare `325Q7Fhnf67OZC$$r43uzK` term. `GlobalId=` written as a comparison is the same identity check spelled differently, so this is a pure adapter change: reuse the existing rule and evaluator, add zero new matching logic. That is the clearest, smallest-surface construct of the five — `parent=` and `query:` need new rule kinds and possibly a rule-model extension for ancestor traversal / key-path evaluation; `+` unions need the `FilterState` combinator model to carry more than one top-level combinator; material Category needs `adaptMaterial` to read a second field IfcOpenShell allows there. Each of those is a real design decision on its own, out of scope for "ship one thing fully."

## What changed

`apps/viewer/src/lib/search/selector-to-rules.ts`'s `adaptAttribute`: when the attribute name is `globalid` and the operator is `=` or `!=` against a literal (non-regex, non-NULL) value, it now returns `Rule.globalId([value.text], op === '=' ? 'in' : 'notIn')` — the exact same rule kind and evaluator (`globalIdOpMatches`, exact case-sensitive set-membership) the bare-GlobalId literal term already builds. Every other GlobalId comparison (`*=`, `>`, `>=`, `<`, `<=`, a `/regex/` value, `NULL`) is still refused by name: a GlobalId is a fixed 22-character identity, not text to search within or order, so `globalIdOpMatches` has no meaning for those and approximating one would be a wrong answer that looks like an answer.

**Zero-match**, explicitly: `GlobalId=<id-not-in-model>` produces a real `Rule.globalId(['<id>'], 'in')` rule — the same rule kind `filter-evaluate.test.ts` already proves returns *no matches* for an unknown id, not "no filter applied." This is inherited from the pre-existing bare-term rule, not new evaluator code, so its zero-match behavior was already under test before this PR.

**Reuse, not re-derivation** (the repo's stated defining defect): this PR adds no new matcher. It routes a second spelling (`GlobalId=X`) onto the one `globalId` rule kind and the one `globalIdOpMatches` evaluator that already existed for the bare-id spelling (`325Q7Fhnf67OZC$$r43uzK`). Both spellings now go through the identical evaluation path — there is structurally one place a GlobalId comparison can diverge from, not two.

## What is NOT in this PR (stays open on #4094 / #4091)

- `parent=`, `query:` — need new rule kinds / ancestor-traversal and key-path evaluation, not present in `FilterRule` today.
- `+` group unions — `FilterState` only carries one top-level combinator.
- Material `Category` — `adaptMaterial` matches material Names only.
- `location=` widening (reach through spaces, not just direct/aggregated containment) — pinned as-is per #4091's own note that this issue "settles" the current, narrower reach.
- Every CLI/MCP/SDK-adapter construct — that is #4774's independent axis, already open as its own PR.

## Test plan

RED-first, then GREEN, then mutation-tested the one new guard (the `value.kind === 'string' && (op === '=' || op === '!=')` condition and its `'in'`/`'notIn'` branch) in the isolated worktree this PR was built and tested in:

| Mutation | Reddened test | Confirmed applied |
|---|---|---|
| Force the guard to `if (false)` (never build the rule) | "GlobalId= and GlobalId!= reuse the same globalId rule the bare term builds" | yes — grep confirmed the literal text landed, test failed |
| Force `op === '=' ? 'in' : 'notIn'` to always `'in'` | same test's `!=` assertion | yes |
| Force the guard to `if (true)` (drop the operator/value-kind restriction) | "GlobalId compared with anything other than = or != is reported, not approximated" | yes |

All three reverted; `diff` against the pre-mutation file was byte-identical after each revert.

- `apps/viewer/src/lib/search/selector-to-rules.test.ts`: replaced the old "GlobalId= is still reported" test (now false — it's supported for `=`/`!=`) with two: the accepted `=`/`!=` case and a rejected-construct table (`*=`, `>`, `/regex/`, `NULL`), each asserting `rules` stays empty and exactly one `unsupported` entry names `GlobalId`.
- Full `apps/viewer/src/lib/search/*.test.ts` run in the fresh worktree (after building the missing `@ifc-lite/regex-guard` dist via plain `tsc`, and linking the rest of the pre-built workspace dist output from a sibling worktree rather than rebuilding everything from scratch): 287/288 pass; the one failure (`result-export.test.ts`) is a pre-existing `@ifc-lite/wasm/pkg` resolution gap in this environment (no `packages/wasm/pkg` build, and this environment/CLAUDE.md forbids `cargo`/`wasm-pack`), unrelated to this change — `selector-to-rules.test.ts` itself is 52/52 green.
- `node scripts/check-module-size.mjs` (plain form, no `--base`): no new-file-over-budget entry for `selector-to-rules.ts` (it measures 400 lines, at the 400-line trigger, not over it) or `selector-to-rules.test.ts`; the tool's remaining output is pre-existing Rust-file and unrelated-viewer-file drift this PR does not touch.

## Exclusions

- No CLI/MCP/SDK changes (that's #4774's axis).
- No Rust changes; no `cargo`/`wasm-pack`/`turbo build --filter` was run.
- `parent=`, `query:`, `+` unions, material Category, `location=` widening are untouched and still report as unsupported, matching current docs.

## Summary by CodeRabbit

- **New Features**
  - Viewer selectors now support exact, case-sensitive `GlobalId=` and `GlobalId!=` comparisons.
  - Matching GlobalId values can be included or excluded from search results.
  - Multiple GlobalId terms preserve their intended matching behavior.

- **Bug Fixes**
  - GlobalId comparisons no longer silently return no matches when valid values are provided.
  - Unsupported GlobalId operators and patterns continue to be reported rather than applied as approximate matches.

- **Documentation**
  - Updated selector syntax guidance to describe supported GlobalId comparisons and clarify unsupported operators and patterns.


louistrue (dismissed):
Approved at exact head 056e3bfe3. The adapter reuses the existing exact GlobalId rule/evaluator, preserves fail-loud handling for unsupported comparisons, has focused production-path coverage, carries the corrected minor changeset/docs, and both exact-head Test runs completed successfully. Ollama reports no major issues.

louistrue (approved):
Approved at exact head 8706cb44d. GlobalId= and GlobalId!= reuse the existing exact Rule.globalId evaluator; unsupported operators remain fail-loud. The follow-up production-path tests correctly pin comparison terms as narrowing/AND and bare GlobalId terms as union-additive. The minor changeset and docs are aligned, and all required checks are green.

Issue #1462, opened 2026-06-30
Filters, 5 suggestions for improvements


MennoMekes:
1: Predefined type is not working in filters ‘is one of’ and ‘is not one of’

<img width="945" height="622" alt="Image" src="https://github.com/user-attachments/assets/0dbb0e0f-0b5f-48e4-a628-726c81de2390" />

<img width="945" height="646" alt="Image" src="https://github.com/user-attachments/assets/3c623370-ed38-46ea-82d9-738dbbbcad02" />

2: Filters: Material is not working as it should be. Same problem as with the lens.
(As fixed in: https://github.com/LTplus-AG/ifc-lite/issues/1366)

<img width="945" height="884" alt="Image" src="https://github.com/user-attachments/assets/709a9e71-924e-49c2-a8aa-e51f4ec824a0" />

3: Scrolling with the mousewheel in the rules is not working.

Is it possible to only show legal and userdefined Psets for the selected elements? So this filters IfcCovering so I don’t really want to scroll through MEP and other irrelevant Psets.

<img width="705" height="453" alt="Image" src="https://github.com/user-attachments/assets/2020a47e-58bb-46a0-a890-a35ef28f4359" />

4: Filters, Not all Psets are available.
Pset_Tiling is missing while it exists in the model for IfcCoverings
 
<img width="945" height="330" alt="Image" src="https://github.com/user-attachments/assets/d24b8f5f-9d6f-43be-8556-ca3b995382bc" />

When creating this list for example:

<img width="945" height="332" alt="Image" src="https://github.com/user-attachments/assets/fcb3ceee-0e96-4049-bdbc-9ae09d718ecc" />

5: And I use create list I kinda expected NetVolume to be already selected when I run the list. Not a big deal, and maybe there is a reason for this.

louistrue:
Fixed in #1469 (merged to main as 731579f6).

PR #4780, opened 2026-09-14
fix(viewer): selector material= matches Category, not just Name (#4094)

BIMvoice:
## Summary

- One of the six viewer-side selector gaps the maintainer's triage comment on #4091 lists for #4094: `material=` matched only material Name; IfcOpenShell's grammar matches Name **or** Category. A wall whose `IfcMaterial` is Name "Fired Clay Brick" / Category "Masonry" now matches `material=Masonry`; before, that comparison silently found nothing.
- No new `FilterRule` kind and no grammar change: the adapter (`selector-to-rules.ts`'s `adaptMaterial`) still builds one `material` rule unchanged. The widening lives entirely in the evaluator's candidate set (`filter-evaluate.ts`'s `matNamesFor`), which now calls a new `materialMatchCandidates` (`filter-match.ts`) instead of `materialNamesOf` alone.
- Category values are supplied by `flattenMaterials` — the IDS `<material>` facet's own flattener (`packages/ids/src/bridge/materials.ts`) — now exported from `@ifc-lite/ids` and reused rather than re-walked, so the selector rule and the IDS material facet answer "does this element's material graph carry X" through one traversal instead of two that could drift apart. Only `flattenMaterials`'s `.category` field is used; its `.name` field also surfaces a layer's own label (e.g. "Insulation Layer"), which `materialNamesOf` (#1462) deliberately excludes from name matching, so Name-matching semantics are unchanged.

## Rejected constructs (and why)

Of the four still open after #4779 (`GlobalId=`, shipped separately) — `parent=`, `query:`, `+` unions of groups, `location=` widening (direct→indirect containment), material Category — I picked **material Category** as the one construct for this task because it needed no new rule kind and no `FilterState` change: it composes onto the existing `material` rule's evaluator. The other four each need either a new rule kind (`parent=`, a spatial-ancestor walk) or a combinator change (`+` unions need OR support the rule model doesn't have today) or a much larger sub-grammar (`query:`) — each is a separate task, not shippable as a small, reused-machinery slice.

## Test plan

RED-then-GREEN, `cd apps/viewer && npx tsx --import ./src/test/vite-module-hooks.mjs --test --test-timeout=120000 src/lib/search/filter-evaluate.test.ts`:
- [x] New fixture (3 walls, 3 distinct materials) added to `filter-evaluate.test.ts`: Wall-A matches `material=Masonry` **only** through Category (Name is "Fired Clay Brick"); Wall-B matches `material*=Steel` through Name only (widening is additive, not a swap); Wall-C (Category "Wood") must NOT match `material=Masonry` through either surface.
- [x] Zero-match case tested explicitly: `IfcWall + material=Concrete` (no wall has that Name or Category) returns `[]`, distinct from the no-filter case (3 walls) — #4659.
- [x] Mutation test: neutered the new `if (category) seen.add(category)` guard (`if (false && category) …`) — reddened exactly 1 of 75 tests (`material=Masonry matches Wall-A…`), 74 others unaffected. Reverted; `git diff --stat` on the reverted mutation matched the pre-mutation file byte-for-byte.
- [x] All 179 tests across `filter-evaluate.test.ts`, `filter-match.test.ts`, `filter-rules.test.ts`, `selector-to-rules.test.ts` pass.
- [x] `packages/ids` and the touched viewer files typecheck clean (`tsc --noEmit`).
- [x] `node scripts/check-module-size.mjs` (plain form, no `--base`): clean on this branch — `filter-evaluate.ts` stayed at its existing 583-line budget (material logic extracted to `filter-match.ts`, which has headroom), `filter-rules.ts` stayed at 399 (doc-comment trimmed to net zero).
- [x] `docs/guide/selector-syntax.md` updated: `material=` row flips from ⚠️ to ✅.

Refs #4094 — the other three gaps (`parent=`, `query:`, `+` unions) remain open.

## Summary by CodeRabbit

- **New Features**
  - The viewer’s `material=` selector now matches material names and categories, including nested material structures.
  - Material matching is consistent between viewer selectors and IDS material filters.
  - The IDS package now provides material-flattening functionality for material filtering.

- **Documentation**
  - Updated selector documentation to clarify support for `GlobalId=` and `GlobalId!=` in the viewer’s Filter tab.

- **Tests**
  - Added coverage for category matches, name matches, combined matching, and no-match behavior.


louistrue (dismissed):
Approved at exact head b7493110c. The selector evaluator reuses @ifc-lite/ids flattenMaterials for Category candidates while preserving the existing material-name collector, including its intentional exclusion of layer labels. Parsed-IFC regressions cover Category-only match, unchanged Name match, unrelated Category rejection, and zero-match restriction. The public export has the API snapshot and minor changesets; required checks are green.

louistrue (approved):
Re-reviewed the new exact head. The only change compresses the material-category comment; production behavior and tests are unchanged. The stacked merge with #4779 now leaves selector-to-rules.ts at 398 lines, below the enforced 400-line module budget that failed the prior merge group.

louistrue (approved):
Re-approved on the branch-update merge. This head contains the previously reviewed material-category fix plus current `main`; no new PR-specific behavior was introduced.

louistrue (approved):
Re-approved after the final branch update. The PR-specific material-category behavior is unchanged; this head now includes the complete integrated `main` feature stack.

PR #4781, opened 2026-09-14
fix(viewer): selector location= reaches one level through a space (#4094)

BIMvoice:
## Summary

- One of the six viewer-side selector gaps the maintainer's triage comment on #4091 lists for #4094: `location="Level 3"` matched only an element the storey **contains directly** (plus its `IfcRelAggregates`-aggregated parts). An element inside an `IfcSpace`/`IfcSpatialZone` on that storey (IfcOpenShell's own example 17 — a pump in a room) matched nothing.
- No new `FilterRule` kind and no grammar/`FilterState` change: `adaptLocation` in `selector-to-rules.ts` still builds one `Rule.storey(...)`, unchanged. The widening lives entirely in the evaluator/prefilter's notion of "which storey" (`filter-match.ts`), composing two lookups that already existed — `elementToStorey` (direct containment) and `getContainingSpace` (`@ifc-lite/data`'s `spatialLookups`, already exposed on every `SpatialHierarchy`) — rather than adding a new traversal.
- **Structurally can't diverge**: added one shared resolver, `storeyIdOf`, that both the per-entity evaluator (`defaultStoreyName` / `storeyMatchesRefs`) and the bulk index prefilter (`unionByStorey`) now call. Before this change only `defaultStoreyName` would have needed fixing for the per-entity case to work — but the prefilter (used whenever a `storey op:'in'` rule sits under AND, see `filter-evaluate.ts`'s module doc-comment) narrows the iteration source from `hierarchy.byStorey` *before* any per-entity check runs, so fixing only the per-entity side would have left space-nested elements silently excluded from the candidate set — the exact "two paths that must agree" defect class this repo has hit before. `unionByStorey` now also unions in `bySpace` elements for every space belonging to a matched storey.

## Scope

Reaches exactly one level down from the storey (a space directly under the storey). A space nested inside another space is not resolved — left open, not guessed at; `docs/guide/selector-syntax.md` says so.

## Test plan

RED-then-GREEN, `cd apps/viewer && npx tsx --import ./src/test/vite-module-hooks.mjs --test --test-timeout=120000 src/lib/search/filter-evaluate.test.ts src/lib/search/filter-match.test.ts src/lib/search/filter-rules.test.ts src/lib/search/selector-to-rules.test.ts`:
- [x] RED: flipped the pinned `'MEASURED: it does NOT reach the pump one level down, inside the space'` test (originally added by #4091) to assert the pump now matches — failed on pre-fix code.
- [x] Two-storey fixture (Level 3 / Level 4, each with a direct wall and a space-nested pump) so a rejection case exists: `location="Level 3"` must NOT pull in Level 4's wall or its space-nested pump. Also a zero-match case (`location="Level 9"`) returns `[]`, not the whole model (#4659 class), explicitly checked against a non-empty unfiltered baseline.
- [x] Also pinned: the bulk index prefilter (`unionByStorey`) on its own already contains the space-nested pump — proves the prefilter can't silently exclude a candidate the per-entity check would go on to match.
- [x] GREEN: all of the above pass after the fix. 178/178 across the four touched test files.
- [x] Mutation test, two independent guards: (1) neutered `storeyIdOf`'s space hop (forced early `return undefined`) — reddened exactly the 3 widening-specific tests in the "storey rule reach" describe block, 175 others unaffected. (2) neutered `spaceElementsOfStorey`'s loop (forced early `return out`) — reddened the same 3 tests independently. Both mutations reverted; `filter-match.ts` confirmed byte-identical to the pre-mutation state afterward.
- [x] Guarded both new helpers against partial `spatialHierarchy` mocks that several existing tests hand-build (only `byStorey`/`elementToStorey`, no `bySpace`/`getContainingSpace`) — this surfaced as 3 real failures (`hierarchy.bySpace is not iterable`) during development, fixed with `if (!hierarchy.bySpace) return out;` and `hierarchy.getContainingSpace?.(...)`, not by rewriting those tests' mocks.
- [x] `node scripts/check-module-size.mjs` (plain form): clean on the three touched files. `filter-evaluate.ts` stayed at its allowlisted 583-line budget (net zero — all new logic went into `filter-match.ts`, which had headroom: 258 → 310 lines). `selector-to-rules.ts` needed no logic change, only a doc-comment update (387 → 390, still under its 400 trigger).
- [x] `docs/guide/selector-syntax.md` updated: `location=` row flips from ⚠️ to ✅, and the "How far `location=` reaches" section describes the new one-hop behavior and the honest remaining limit (space nested inside another space).

Refs #4094 — `parent=`, `query:`, and `+` unions remain open; each needs new machinery (a spatial-ancestor rule kind, a sub-grammar, and an OR-combinator the `FilterState` model doesn't have, respectively) and is out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Storey-based location filters now include elements contained within spaces or spatial zones on the selected storey.
  * Filtering continues to exclude elements from other storeys and unsupported deeper nested spaces.
  * Storey and space relationships are now preserved more reliably when viewer data is loaded.

* **Documentation**
  * Updated selector syntax guidance to describe supported storey filtering, including direct, aggregated, and one-level space-contained elements.


louistrue (dismissed):
Approved at exact head 5ca240ec7. One shared storeyIdOf resolver now governs name and exact-ref evaluation, with a matching unionByStorey prefilter that cannot exclude space-contained candidates. Server hydration restores space-like node membership consistently with the parser path, including IfcSpatialZone. Two-storey production and prefilter regressions cover positive reach and leakage rejection; the minor changeset/docs are aligned and required CI is green.

louistrue (approved):
Re-approved after the comment-only size-budget repair. The stacked selector adapter is now 399 lines with #4779, and the location behavior is unchanged.

PR #4793, opened 2026-09-14
fix(viewer): raise low-contrast panel secondary text to WCAG AA

BIMvoice:
## Summary

Refs #4792 (the tooltip-contrast survey; does not close it — the token-level rows below remain for the maintainer).

Fixes the confirmed **always-visible, always-enabled real text** sites from #4792's survey that measured below WCAG AA's 4.5:1 normal-text minimum, by dropping the opacity suffix to match the codebase's standard muted-on-surface pattern — the same fix #4788/#4784 applied to `BsddCard`'s dataType line. No new colors invented; no design tokens changed.

Builds on `contrast-wcag-test` (#4788, open) since it supplies the contrast harness this PR extends and reuses — this PR's base is that branch, not `main`, so the diff here is scoped to only what this PR changes.

### Sites fixed (re-verified against source myself before changing anything)

| Site | Class before | Measured before (light / dark / colorful) | After |
|---|---|---|---|
| `ChatPanel.tsx` "Streaming..." status | `text-muted-foreground/50` | 1.98 / 2.31 / 3.17 | 4.83 / 5.42 / 6.66 |
| `ChatPanel.tsx` usage percentage | `text-muted-foreground/40` | 1.70 / 1.91 / 2.63 | 4.83 / 5.42 / 6.66 |
| `ChatPanel.tsx` "Shift+Enter new line" hint | `text-muted-foreground/40` | 1.70 / 1.91 / 2.63 | 4.83 / 5.42 / 6.66 |
| `ChatPanel.tsx` "⌘L" hint | `text-muted-foreground/30` | 1.48 / 1.58 / 2.15 | 4.83 / 5.42 / 6.66 |
| `apps/viewer/src/components/viewer/PropertiesPanel.tsx` "Size" label + value | `text-muted-foreground/50` | 1.98 / 2.31 / 3.17 | 4.83 / 5.42 / 6.66 |
| `ClashPanel.tsx` active detection-mode label | `text-muted-foreground/60` | 2.31 / 2.72 / 3.75 | 4.83 / 5.42 / 6.66 |
| `TourStepCard.tsx` "docked back" / "Stuck? Skip" notices, step counter | `text-muted-foreground/80` and `/70` | sub-AA in light+dark per #4792's survey | 4.83 / 5.42 / 6.66 (on `bg-popover`) |
| `HoverTooltip.tsx` world-coordinate readout | `text-muted-foreground/80` | sub-AA | 4.83 / 5.42 / 6.66 (on `bg-popover`) |
[...]

Issue #4792, opened 2026-09-14
Contrast survey: widespread sub-AA and near-invisible text beyond the tooltip surface (#4788/#4783/#4787 follow-up)


BIMvoice:
> **Correction (edited after initial post):** one citation below had a wrong file path — `PropertiesPanel.tsx` was implied to live under `components/viewer/properties/` (by analogy with `CoordinateDisplay.tsx`, which does). It doesn't; the real path is `apps/viewer/src/components/viewer/PropertiesPanel.tsx`. Root cause: every citation in the original table used a bare filename with no directory, which is ambiguous whenever a filename doesn't uniquely identify its folder — this table now uses full paths throughout to remove that ambiguity. I re-verified **every** `file:line` in this issue against `upstream/main` (`6fa3d1425`) with `git show upstream/main:<path> | sed -n '<line>p'`; the bare-filename PropertiesPanel case was the only wrong one — every other line number and content matched exactly, so the error was **not** a systematic wrong-directory guess repeated everywhere, it was ambiguous citation formatting that broke on exactly one filename that happens to collide in spirit with a real subfolder. I also went further than a path re-check: I opened the surrounding code for every `/30`-`/60`-tier citation to determine whether it's real text, a decorative/icon element, or disabled-state styling — see the new "Text vs. icon, and the disabled-state question" section below. That changes which rows are genuine 4.5:1 text failures.

## Summary

Following #4787's precedent (a real WCAG contrast harness beats a `className` string comparison) and the fix pattern in #4783/#4784, I pointed #4788's harness at the rest of the viewer's small/secondary text instead of just the tooltip surface it currently covers. This is a **survey, not a fix** — no components were changed.

Method: I reused #4788's exact harness (`apps/viewer/src/test/contrast/{wcag,render-harness}.ts` — real Tailwind compile of `index.css`, real headless Chromium paint, canvas "source-over" composite for `oklch`/`oklab`). I did **not** write a second contrast engine. I wrote one throwaway script (`.scratch-survey*.mts`, on a scratch worktree, never committed) that calls the same `compileAppCss`/`contrastRatio` exports with arbitrary `(surfaceClass, textClass, theme)` triples, since the shipped harness only measures the tooltip case.

[...]

Issue #4388, opened 2026-09-09
Resolving a conflict in a ratchet allowlist is a silent policy change, and check-module-size exits 0 either way


louistrue:
## Resolving a conflict in a ratchet file is a silent policy change

`scripts/module-size-allowlist.txt` and `rust/processing/tests/module_size_allowlist.txt` are policy, not data: each row is a budget someone justified, and the rule is that budgets ratchet DOWN. But when two branches both touch the file, resolving the conflict means choosing budgets, and **`check-module-size` exits 0 for every plausible resolution.** The mistakes surface only as advisory `note:` lines.

So the one gate that owns the file cannot tell you that you just undid its policy.

## Two real instances, opposite directions, same resolution

Both from one conflict resolution on #4330, caught by a `/code-review` pass run specifically over the resolution rather than the diff:

**Resurrecting a deleted row.** `523 packages/parser/src/project-units.ts` came back. Main had DELETED that row in a named ratchet-tightening commit, because the file is 268 lines and under the 400 limit. The resolver believed they were "preserving the PR's new row"; it was never the PR's, it predated the branch. Net effect: a 268-line file recovered a 523-line budget, which is 255 lines of silent regrowth headroom.

**Dropping the PR's only real change.** The same resolution kept a row the PR had deliberately deleted. Its split takes `schedule-extractor.ts` from 594 lines to 348, so the row should go. Keeping it left 246 lines of headroom on the exact god file the PR had just finished splitting.

Both look like "resolved the conflict". Neither is visible in CI, and neither is obvious to a reviewer reading the diff, because a changed number in an allowlist reads as noise next to the code change beside it.

A third instance from the same night, milder and self-inflicted: a `--update` run annexed `symbolic-overlay-pipelines.ts` at 766 when the file is 765 on both sides, so a branch claimed a budget for a file it does not grow.

## What would actually catch it

Sketches, not a decision. The point is that today nothing does.

- **Every changed row must be justified by a measured line count.** For each row a PR adds, raises or deletes, compare the budget against `wc -l` of that file on the PR's head. A raise where the file did not grow, or a kept row for a file now under the limit, is the whole defect class. This is mechanical and needs no judgement.
- **A deleted-on-main row that reappears is always wrong.** Diff the allowlist against the merge base rather than against the branch point, so a resurrection is distinguishable from an addition.
[...]

PR #4794, opened 2026-09-14
fix(viewer): raise remaining low-contrast panel secondary text to WCAG AA

BIMvoice:
## Summary

Refs #4792 (does not close it — the token-level marginal case below remains, plus the `text-primary-foreground`/`bg-primary` finding the survey deliberately left out of scope).

PR #4793 fixed the confirmed sites from #4792's survey but explicitly named eleven files it did not individually verify:

`ChunkErrorBoundary.tsx`, `IDSAuditSummary.tsx`, `EntityContextMenu.tsx`, `RoomPanel.tsx`, `CustomizeSidebar.tsx`, `SectionPanel.tsx`, `ribbon/primitives.tsx`, `compare/*`, `PanelWindowHost.tsx`, `LayersPanel.tsx`.

This PR finishes that pass: read each file's surrounding JSX, classified every `/50`-`/80` `text-muted-foreground` occurrence, measured the real-text sites against their real surface with #4793's own `measureTextContrastOnSurface(theme, surfaceClass, textClass)` harness (no second measurement mechanism), and fixed the ones that fail AA.

Based on `fix-4792-contrast` (#4793's branch, currently open) rather than `main`, since that branch is where `render-harness.ts`'s `measureTextContrastOnSurface` helper and `panel-secondary-text.test.ts` already live — basing off `main` would mean re-adding that harness here and duplicating #4793's diff.

### Classification

| Site | Class | Classification | Action |
|---|---|---|---|
| `ChunkErrorBoundary.tsx:108` error detail | `/70` | real text | fixed |
| `IDSAuditSummary.tsx:348,356,366` path/facet/detail-key labels | `/70` | real text | fixed |
| `IDSAuditSummary.tsx:380` chevron | `/60` | icon, `aria-hidden` | left alone |
| `EntityContextMenu.tsx:520,577` ⌘D / shortcut hints | `/70` | real text | fixed |
| `RoomPanel.tsx:280` invite hint | `/70` | real text | fixed |
| `CustomizeSidebar.tsx:138` grip handle | `/60` | icon, `aria-hidden` | left alone |
| `CustomizeSidebar.tsx:179` "Hidden" section header | `/70` | real text | fixed |
| `SectionPanel.tsx:161` drag-handle wrapper | `/50` | decorative (wraps an icon, no text) | left alone |
| `SectionPanel.tsx:221` "or pick an axis" | `/70` | real text | fixed |
| `ribbon/primitives.tsx:176` group label | `/70` | real text | fixed |
| `compare/ChangeDetailView.tsx:35,64,72,78` data-count/delta/shape-hash text | `/70` | real text | fixed |
| `compare/ChangeDetailView.tsx:101` before→after arrow | `/60` | real text (not `aria-hidden`, conveys the change direction) | fixed |
| `compare/CompareResultsList.tsx:59` `CountBadge` hint | `/70` | real text | fixed |
| `dock/PanelWindowHost.tsx:74,96` window-kind label / "Live · synced" strip | `/70` | real text, but on `bg-muted/40`/`bg-muted/30` | **left alone — see below** |
| `layers/LayersPanel.tsx:238` empty-state icon | `/50` | icon, `aria-hidden` | left alone |
| `layers/LayersPanel.tsx:278` "drop .ifcx files" hint | `/70` | real text | fixed |

### Measurements (before → after, light / dark / colorful)

All 15 fixed sites measured 2.31-2.76 / 2.72-3.30 / 3.07-4.43 before (below AA everywhere), and 4.83 / 5.42 / 5.76-6.66 after (the same clean numbers #4793 already established for plain `text-muted-foreground` on `bg-background`/`bg-card`/`bg-popover`).

### Left alone, and why

- **Icons / `aria-hidden`**: `IDSAuditSummary.tsx:380` (chevron), `CustomizeSidebar.tsx:138` (grip handle), `LayersPanel.tsx:238` (empty-state icon) — WCAG's 4.5:1 text rule doesn't apply.
- **Decorative, no text**: `SectionPanel.tsx:161` — a drag-handle `<span>` that wraps a `GripVertical` icon with no text content (same shape as #4792's `CoordinateDisplay.tsx:42` finding).
- **`PanelWindowHost.tsx`'s two sites** are real, always-visible text, but they sit on `bg-muted/40`/`bg-muted/30`. Measured: even the plain (no-opacity) `text-muted-foreground` class measures **4.39/4.75/3.15** and **4.43/4.71/3.15** there — light is marginally under AA and colorful fails outright. This is the same `bg-muted` token-level marginal case #4792's own survey already flagged (its "Marginal" table: `text-muted-foreground` plain on `bg-muted` is 4.397 light / 3.15 colorful) and did not fix. Dropping the opacity suffix alone does not clear AA here; fixing it would mean changing the surface or the token, which is out of this fix's "mechanical opacity-drop" scope — flagging for the maintainer rather than inventing a color.

### Harness

Reused `measureTextContrastOnSurface(theme, surfaceClass, textClass)` from `apps/viewer/src/test/contrast/render-harness.ts` (#4793) unchanged — no second measurement mechanism. Ran locally with the `channel: 'chrome'` flag temporarily dropped (no system Chrome on this box); reverted before committing, confirmed with `git diff` that `render-harness.ts` has no diff against this branch's base.

### Non-vacuousness proof

Reverted `ribbon/primitives.tsx`'s fix back to `text-muted-foreground/70` and re-ran `panel-secondary-text.test.ts`: exactly the 3 affected assertions (light/dark/colorful) turned red, 120/123 still passing. Reverted back; clean re-run: 123/123 pass.

### Gates run

- `node --test --import tsx src/test/contrast/panel-secondary-text.test.ts` (in `apps/viewer`): 123/123 pass (66 pre-existing from #4793 + 51 new for the 17 fixed-and-regression sites added here + the arrow-glyph non-vacuousness case).
- `node scripts/check-module-size.mjs` (plain form, not `--base`): exits 1, but the failure is entirely the pre-existing `rust/processing/tests/module_size_allowlist.txt` drift vs. `origin/main` (#4388) that #4793 already reported — confirmed none of the files this diff touches appear anywhere in that script's output.
- Changeset added (`@ifc-lite/viewer` patch), no license header.


louistrue (changes requested):
This is transitively blocked: it targets #4793, which targets the still-blocked #4788 harness. Until #4788 derives the tooltip surface from production and exercises the production BsddCard instead of a copied test fragment, the stacked contrast suite is not valid merge evidence. Please repair the base chain first; then this follow-up can be reviewed as the next stack entry.

louistrue (approved):
Reviewed at exact head f21c886 after clean restack onto the merged #4793 base. The remaining panel secondary-text surfaces are raised to AA, the ribbon-label case is measured against its production surface, and the viewer changeset is present. All eight hosted viewer shards, E2E, changed-test production observation, provenance, build/WASM, and exact-head local Ollama pass. Aggregate Type/Lint/Node failures are inherited from current main.

Issue #4806, opened 2026-09-14
Clash detection Report


gDonia87:
### The problem

If I export a collision to BCF I get only the viewpoint but not the found Objects

### Proposed solution

The objects that collide should be automatically selected, so that this information is exported to the BCF

BIMvoice:
Thanks for reporting this! Your observation matches what we found in the code.

When a clash is focused, ifc-lite deliberately clears the selection (to show the pair in highlight colours instead). But then the BCF export asks for the selection to be included — so it captures nothing. We have a clear lead on the root cause.

To help us verify and prioritize the fix, could you share a bit more about your workflow?

1. **Which tool are you using to open the exported BCF?** (Solibri, BIMcollab, Revit plugin, etc.)

2. **Is the model loaded in that tool the same IFC file** you used in ifc-lite — the same export with the same object IDs — or a different version/export of the same building? This matters because BCF identifies objects by their global ID, so if the tool has a different model export, it won't find the objects even if we export them correctly.

3. **What happens when you open the topic in the other tool?** Does the camera jump to the right place but nothing highlights, or does nothing load at all?

4. **Which ifc-lite export path did you use?** (the BCF panel's export button, or something else?) And if you know, was it BCF 2.1 or 3.0?

5. **Would you be willing to share the exported .bcf or .bcfzip file** (or just the viewpoint.bcfv inside it)? That would let us confirm exactly what got written.

Thanks!

gDonia87:
1. I opened the bcf file in BIM Collab and in usBIM. The result is the same;
2. Yes;
3. The camera doesn't jump to the objects;
4. The bcf panel export button;
5. I think it isn't necessary. The objects are not selected.

BIMvoice:
Thanks for the detailed report — your diagnosis of the problem was spot on.

**What was wrong:** When the viewer highlights a clash, it deliberately clears the selection to make the two clashing objects read as a clash rather than a selection (that's the amber and cyan colouring you see). The BCF export then asked for "whatever is currently selected" — which was nothing. So the viewpoint ended up with the snapshot and metadata but no objects.

**What changed:** PR #4808 now writes the two clashing objects directly into the BCF topic's components list at export time, independent of what is selected on screen. Their clash colours (amber and cyan) go in with them, so what you see in the viewer carries into the BCF file.

[...]

PR #4808, opened 2026-09-14
fix(viewer): carry the clashing pair as BCF found objects on clash export

BIMvoice:
## Summary

Refs #4806

- `ClashPanel.createBcfTopic` calls `focusClash`, which deliberately clears the live viewer selection and paints the clashing pair only through the clash-highlight colour channel — never selection, per #1277/#1339. `createViewpointFromState` derived `<Selection>` solely from that (now empty) live selection, so the exported BCF viewpoint's `<Selection>` was entirely **absent** — not merely empty — and any tool opening the topic showed no "found objects" at all.
- GlobalIds were already resolving correctly; there was no id-mapping bug. `Coloring` already existed in `@ifc-lite/bcf`'s writer/viewpoint API and was simply unused at this call site.
- Adds `additionalSelectedRefs` / `additionalColoredRefs` to `createViewpointFromState`'s options, threaded **independent** of the live selection, so the clash pair's GlobalIds and on-screen amber/cyan tint are always written into the exported viewpoint — without resurrecting an actual viewer selection (which would re-introduce exactly what #1277/#1339 removed: selection-blue, the 2-SEL counter, selected-treatment under isolate/ghost).
- Change is localized to the clash export call site; the other three `createViewpointFromState` call sites (`BCFPanel.tsx`, `LayerReviewSection.tsx`, `useBcfFromChange.ts`) are untouched.

This issue is labelled `enhancement`, not `ready` — using `Refs #4806` rather than a closing keyword, pending the maintainer's `ready` label.

## Test plan

- [x] RED: `useBCF.clash-viewpoint-selection.test.tsx` reproduced the exact absent-selection symptom (`Set(0) {}` vs expected `Set(2) {...}`) before the fix.
- [x] GREEN: same suite (5 tests) passes after the fix — in-memory `viewpoint.components.selection`, the raw serialised `viewpoint.bcfv` XML (`<Selection>` present with both GlobalIds), a no-live-selection scenario, a live-selection dedup scenario, and `<Coloring>` with both clash colours.
- [x] `clash-colors.test.ts` extended with unit tests for the new `clashColorToBcfArgb` ARGB-hex conversion.
- [x] Mutation-tested each new guard (the `additionalSelectedRefs` merge loop, its dedup check, the `additionalColoredRefs` conditional, and the ARGB channel encoding) — each mutation reddened the relevant test; all reverted (`git diff --stat` clean before commit).
- [x] `tsc --noEmit` clean on touched files.
- [x] `node scripts/check-module-size.mjs` (plain form) — `ClashPanel.tsx` and `useBCF.ts` budgets raised in `scripts/module-size-allowlist.txt` to their new measured line counts.

## Summary by CodeRabbit

* **New Features**
  * Clash-to-BCF exports now preserve both clashing element GlobalIds.
  * Exported viewpoints retain amber/cyan clash highlight colors.
  * Clash exports no longer depend on the viewer’s current selection state.

* **Bug Fixes**
  * Prevented duplicate element entries in BCF selections, including self-clashes.
  * Ensured clash colors use the required BCF format.


louistrue (changes requested):
The new Coloring path must deduplicate after federated refs are resolved, not only before resolution in ClashPanel. Two different refs can resolve to the same IFC GlobalId; the current map/filter then serializes that IfcGuid into multiple color groups, producing conflicting BCF coloring. Track emitted GlobalIds across groups, retain the first color, keep unresolved/empty filtering, and add a regression with distinct refs resolving to the same GlobalId.

louistrue (approved):
Reviewed at exact head 468399b. The follow-up deduplicates resolved IFC GlobalIds across color groups (first group wins), including alias refs that resolve to the same GUID, without expanding the existing useBCF module budget. All eight hosted viewer shards, E2E, changed-test production observation, Claude, and local Ollama pass; viewer typecheck also passed locally. Aggregate Type/Lint/Node failures are inherited from current main.

Issue #4785, opened 2026-09-14
Add localization support to the viewer


andhxl:
### The problem

The viewer UI currently uses hardcoded English strings, which makes it difficult to provide the interface in other languages without modifying the viewer components directly.

### Proposed solution

Add localization support to the viewer UI, with English remaining the default language.

User-facing strings should be separated from the UI components so additional translations can be provided without modifying the core viewer code.

BIMvoice:
## Mechanism gap: interpolation and pluralisation

PR #4810 established the pattern with a static-only mechanism — `useTranslation()`'s `t(key)` takes no arguments, and `registry.ts`'s `resolve()` returns catalogue values unchanged. PR #4831 converted a second component but deliberately left two string shapes hardcoded rather than extend the catalogue:

1. **Interpolated** aria-labels: `` `Assignment ${index + 1}: ${item.source.name} on ${item.model.name}` ``
2. **Pluralised** counts: `1 object` vs `N objects`

Most remaining components contain one or both shapes, so this blocks the bulk of the conversion.

### Options to consider
- **Positional/named placeholders** in `t(key, params)` — lightweight, covers interpolation, does nothing for plurals.
- **Plural-aware lookup** (e.g. an `Intl.PluralRules`-backed variant) — correct across languages, adds machinery.
- **Adopt an existing i18n library** — solves both properly, but #4810 deliberately avoided a dependency and the repo has a module-size ratchet.
- **Leave dynamic strings in English** — lowest cost, visibly wrong in a localised UI.

Also relevant: `en.ts` will be 31 lines after #4831 lands; the 400-line gate will eventually require splitting once more keys accumulate, regardless of which shape you pick.

Which direction before the bulk conversion proceeds?

louistrue:
Claimed for session goal-20260915-issue-4785 on branch feat/4785-viewer-localization. I will follow the existing localization plan, keep English as the default, separate catalogue strings from UI consumers, and add behavior-focused tests before opening a PR.

louistrue:
Closing: the goal of this issue is met. User-facing strings live in catalogues separate from the components, English is the default, and a translation can be added as one `locales/<tag>.ts` file without touching viewer code (#4810, #4831, #4860, #4861, #4862, #4883). Converting the remaining hardcoded strings continues in #4918, which has a charter and an end gate.

PR #4810, opened 2026-09-14
feat(viewer): string-catalogue i18n mechanism, MergeLayersBanner as pattern-setter

BIMvoice:
## Summary
Issue #4785 asks that the viewer's user-facing strings be separated from components so translations can be provided without touching core viewer code, with English staying the default.

**Scoping before shipping:** a raw grep of `apps/viewer/src` (JSX text nodes `>Text<`, plus `label=`/`title=`/`placeholder=`/`aria-label=`/`description=`/`tooltip=` props, plus `toast.*`/`alert()` calls, deduped against CSS `translate`/`translateX` false positives) puts English UI strings at roughly **1,900 hits across 353+ non-test `.tsx` files**. That is well past what one PR can convert responsibly on a guessed convention, and the repo has **no existing i18n dependency or pattern** (`package.json` has no i18n lib; every `i18n|intl|translat` grep hit outside this PR was a Tailwind `translate-*` CSS class). The issue itself carries no maintainer comment settling library/key-naming/where-translations-live.

**Decision:** ship the bounded first slice only -- introduce the catalogue mechanism and convert exactly one self-contained component end-to-end, to prove the pattern without a 1,900-string mechanical pass on an unreviewed shape. This PR is the pattern-setter, **not** the full localization feature.

## What's here
- `apps/viewer/src/i18n/en.ts` -- the English catalogue (source of truth; every locale is a `Partial<typeof en>` overlay).
- `apps/viewer/src/i18n/registry.ts` -- locale registry (`registerLocale`, `setLocale`, `getLocale`) and `resolve(key)`, which falls back to English **only when a key is absent**, not when a locale explicitly translates it to `''` -- so a missing translation is never silently indistinguishable from a deliberately blank one.
- `apps/viewer/src/i18n/useTranslation.ts` -- the hook components call (`useSyncExternalStore`-based so `setLocale` re-renders consumers).
- `MergeLayersBanner.tsx` converted to use the catalogue for its 5 strings (title enabled/disabled, subtitle, reload button, dismiss aria-label) -- chosen for being small (108 lines) and fully self-contained (no children, no other component depends on its internals).

No new dependency: hand-rolled rather than pulling in an i18n library, since the slice is one small component and bundle size is under a ratchet here; a library becomes worth evaluating once real translation-loading/pluralization/interpolation needs show up from converting further components.

## Test plan
- [x] `apps/viewer/src/i18n/useTranslation.test.tsx` (3 cases): default English resolution, fallback to English when a registered locale omits a key, and an explicit empty-string translation is NOT treated as missing.
- [x] `apps/viewer/src/components/viewer/MergeLayersBanner.i18n.test.tsx` (2 cases): renders catalogue values, and a registered locale that translates one key but omits another falls back to English for the omitted one (not blank).
- [x] Mutation-tested the fallback guard in `registry.ts` (`value !== undefined ? value : en[key]` -> unconditional `catalogue?.[key]`): reddened the "missing key falls back" test as expected, then reverted -- `git diff --stat` confirms no accidental change survived.
- [x] Also mutation-tested by reverting `MergeLayersBanner.tsx` to a hardcoded string: reddened the locale-fallback component test, confirming the test actually exercises the conversion; reverted.
- [x] `node scripts/check-module-size.mjs` (plain form): no new file flagged; the run's non-zero exit is a pre-existing rust `module_size_allowlist.txt` drift vs merge-base, unrelated to this change (no rust files touched here).

Refs #4785 (this converts one component; it does not close the issue).

## Summary by CodeRabbit

- **New Features**
  - Added localization support for the viewer.
  - English is used by default, with support for registering and switching to translated locales.
  - Missing translations automatically fall back to English.
  - Internationalized the “Merge Multilayer Walls” reload banner, including its accessible labels and actions.


louistrue (changes requested):
The active-locale replacement path is still incorrect. registerLocale explicitly supports replacement, but replacing the currently active catalogue neither notifies subscribers nor changes useSyncExternalStore's snapshot (currently only the locale string), so mounted consumers retain stale translations. Make the external-store snapshot change on active catalogue replacement, notify listeners, and add a mounted-consumer regression; inactive-locale registration should remain non-notifying.

louistrue (approved):
Reviewed at exact head dc94d46. The catalogue registry now changes snapshot identity and notifies mounted consumers when the active catalogue is replaced; the regression mounts a real consumer and verifies First→Second. All eight hosted viewer shards, E2E, CodeRabbit, and exact-head local Ollama pass; viewer typecheck and the targeted i18n tests pass locally. The Claude red is its review classifier misreading data-key={tag} as an incomplete numeric-bound class, with no code finding; aggregate Type/Lint/Node reds are inherited from main. Revert-oracle exemption rationale is documented on the PR.

Issue #4813, opened 2026-09-14
main is red since 04fbc7578 (perf(renderer): park IFNS GPU culling experiment): Typecheck, Lint, Node tests fail in packages/renderer


BIMvoice:
## Timeline (verified via the Actions API, `Test` workflow, `head_branch=main`)

- **Last green**: `e960b3662` — run 34850781392, success, completed 2026-09-14T13:41:18Z
- **First red**: `04fbc7578` ("perf(renderer): park IFNS GPU culling experiment", authored by louistrue) — run 34854488704, failure, completed 2026-09-14T14:16:00Z
- **Still red now**: `ea25297ac` (current main tip, "fix(ci): let cold release images finish (#4809)") — run 34867694905, failure, completed 2026-09-14T16:17:42Z

`04fbc7578` is the very next commit on `main` after `e960b3662` (no intervening commits), so the bisect is unambiguous — no `Test` runs exist between the last green and first red to check.

## Three failing jobs on 04fbc7578 (run 34854488704), verbatim

**Typecheck**
```
packages/renderer/src/instanced-gpu-culling.test.ts(37,3): error TS2739: Type '{ vertexBuffer: GPUBuffer; indexBuffer: GPUBuffer; indexCount: number; instanceBuffer: GPUBuffer; boundingSpheres: Float32Array<ArrayBuffer>; instanceCount: number; }' is missing the following properties from type 'InstancedTemplateGPU': modelIndex, bounds, maxOccRadius, selectedCount
packages/renderer/src/instanced-gpu-culling.test.ts(57,20): error TS2352: Conversion of type '{ getBindGroupLayout: () => GPUBindGroupLayout; }' to type 'GPUComputePipeline' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
packages/renderer/src/scene-local-bounds.test.ts(78,7): error TS2741: Property 'boundingSpheres' is missing in type '{ modelIndex: number; positions: Float32Array<ArrayBuffer>; normals: Float32Array<ArrayBuffer>; indices: Uint32Array<ArrayBuffer>; instanceData: ArrayBuffer; localMin: [...]; localMax: [...]; }' but required in type 'InstancedTemplateCpu'.
packages/renderer/src/scene-local-bounds.test.ts(112,7): (same, InstancedTemplateCpu.boundingSpheres)
[...]

PR #4815, opened 2026-09-14
fix(viewer): discoverFilterValues suggests material Category values (#4780 parity)

### LINKED ISSUE #4814: Material filter dropdown doesn't suggest Category values that material= now matches
## Summary

PR #4780 widened the `material=` selector so it matches an element's material **Category** as well as its **Name** (`materialMatchCandidates` in `apps/viewer/src/lib/search/filter-match.ts`, wired into `filter-evaluate.ts`).

`filter-schema.ts`'s `discoverFilterValues` — which populates the material filter's value-autocomplete dropdown — still collects candidates via the older `materialNamesOf` (Name only).

Net effect: `material=Masonry` **matches** an element whose material Name is "Fired Clay Brick" and Category is "Masonry", but the dropdown never **suggests** "Masonry". Free-text entry still works, so this is a discoverability gap, not a wrong-answer defect.

## Fix

Have `discoverFilterValues` collect via `materialMatchCandidates` (the same function the matcher uses) instead of `materialNamesOf`, so the two surfaces cannot drift apart again.

Cost: `materialMatchCandidates` unions `materialNamesOf` with Category values from `flattenMaterials` (`@ifc-lite/ids`) — a pure in-memory walk over the already-extracted `MaterialInfo` structure, same order of magnitude as the existing walk, bounded by the same 5000-entity `VALUE_SAMPLE_CAP`. No extra WASM/parse cost.

BIMvoice:
## Summary

PR #4780 widened the `material=` selector so it matches an element's material **Category** as well as its **Name**, via `materialMatchCandidates` (`apps/viewer/src/lib/search/filter-match.ts`), wired into `filter-evaluate.ts:402`.

`filter-schema.ts`'s `discoverFilterValues` — which populates the material filter's value-autocomplete dropdown — still collected candidates through the older `materialNamesOf` (Name only). Net effect: `material=Masonry` matched an element whose material Name is "Fired Clay Brick" / Category is "Masonry", but the dropdown never suggested "Masonry". Free-text entry still worked, so this was a **discoverability gap, not a wrong-answer defect**.

## Fix

`discoverFilterValues` now collects via `materialMatchCandidates` — the same function the matcher uses — instead of re-deriving a second Name/Category union, so the two surfaces can't drift apart again.

## Cost

`materialMatchCandidates` unions `materialNamesOf` with Category values pulled from `flattenMaterials` (`@ifc-lite/ids`) — a pure in-memory walk over the already-extracted `MaterialInfo` object (no extra WASM call, no re-parse), the same order of magnitude as the existing Name walk. `discoverFilterValues` samples at most `VALUE_SAMPLE_CAP = 5000` entities, so the extra traversal is bounded the same way the existing one is. Not expensive.

## Other `materialNamesOf` call sites (full audit)

| Site | Verdict |
|---|---|
| `filter-match.ts:272` (definition) / `:283` (used inside `materialMatchCandidates`) | source of truth, not a consumer |
| `filter-evaluate.ts` `matNamesFor` (the selector matcher) | already uses `materialMatchCandidates` (fixed by #4780) |
| `filter-evaluate.ts` `__internal.materialNamesOf` | test-only re-export, not a production consumer |
| `filter-schema.ts` `discoverFilterValues` (material dropdown) | **the gap fixed here** |
| `apps/viewer/src/lib/lists/adapter.ts` `getMaterialNames` (List Builder material column) | own local re-implementation of the Name-only walk, feeding a data-table **column**, not the `material=` selector grammar — deliberate Name-only display, not the same rule. Left unchanged. |

## Test plan

RED-then-GREEN in `apps/viewer/src/lib/search/filter-schema.test.ts` (new `discoverFilterValues` describe block, previously untested):
- Fixture: two walls via real STEP source + `IfcRelAssociatesMaterial`. Wall-A's material Name is "Fired Clay Brick", Category "Masonry" (Category differs from Name — capable of catching the regression). Wall-B's material is Name-only "Steel", no Category (bounding control — proves the fix is additive, not a swap).
- RED: before the fix, discovered materials were `["Fired Clay Brick", "Steel"]` — "Masonry" absent.
- GREEN: after swapping to `materialMatchCandidates`, discovered materials include "Masonry" and "Fired Clay Brick" and "Steel" (10/10 tests in the file pass).
- Mutation test: reverted the import/call-site back to `materialNamesOf`, confirmed exactly 1 of 10 tests reddened (the new Category-discovery test; the Name-only bounding control stayed green), then reverted the mutation — `git diff --stat` matched the pre-mutation file exactly.
- `apps/viewer/src/lib/search/filter-match.test.ts` + `filter-evaluate.test.ts`: 93/93 pass, unaffected.
- `node scripts/check-module-size.mjs` (plain form): exit 1, but the only offender is `packages/renderer/src/types.ts` (410 lines, no allowlist row) — the known, pre-existing `main` breakage from issue #4813, unrelated to this change. No file this PR touches appears anywhere in its output.
- `tsc --noEmit`: clean on the touched files.

Closes #4814. Awaiting the maintainer's `ready` label per the issue-queue gate.

## Summary by CodeRabbit

* **New Features**
  * Material filter suggestions now include matching material Categories alongside material Names.

* **Bug Fixes**
  * Improved material value discovery so category-only material values appear in filter suggestions.
  * Aligned filter suggestions with material selector matching behavior.


louistrue (approved):
Reviewed the production path, real material fixture, changeset, and exact-head CI. Category discovery now uses the same materialMatchCandidates lens as matching; the Name-only bounding control prevents a swap regression. Viewer shards and production-observation checks pass; remaining failures are inherited from red main.

PR #4818, opened 2026-09-14
fix(viewer): discoverFilterValues walks every material association, not just the primary

### LINKED ISSUE #4817: Material filter dropdown samples only the primary material association, not every one
## Summary

`packages/parser/src/material-resolver.ts` exposes two extractors:
- `extractMaterialsOnDemand` — the entity's PRIMARY material (lowest-rel-express-id association).
- `extractAllMaterialsOnDemand` — EVERY material association; its own doc notes an exporter that attaches e.g. a layer set **and** a plain fallback `IfcMaterial` yields several.

The two consumers of these disagree:
- `apps/viewer/src/lib/search/filter-evaluate.ts` (`matNamesFor`, the `material=` selector **matcher**) already uses `extractAllMaterialsOnDemand`.
- `apps/viewer/src/lib/search/filter-schema.ts` (`discoverFilterValues`, the material filter's **dropdown**) uses only `extractMaterialsOnDemand`.

So for an entity with two or more material associations, a Name or Category carried only on a non-primary association will **match** `material=` but never be **suggested** in the dropdown. This is the same discoverability gap as #4814/#4780 (Name-vs-Category), one level deeper — at the association level.

Note: #4815 (open) fixes the Name-vs-Category gap but still calls `extractMaterialsOnDemand` (primary-only), so it does not close this one.

## Fix

Have `discoverFilterValues` walk `extractAllMaterialsOnDemand` (every association) and expand each via `materialMatchCandidates` (Name + Category), the same candidate set the matcher uses, so the two surfaces cannot drift apart again. Cost: `discoverFilterValues` already samples at most `VALUE_SAMPLE_CAP = 5000` entities; the extra per-entity work is bounded by the (typically 1, rarely more) associations an entity actually carries.

BIMvoice:
## Summary

`matNamesFor` (the `material=` selector **matcher**, `filter-evaluate.ts:~397`) unions Name/Category candidates across `extractAllMaterialsOnDemand` — every `IfcRelAssociatesMaterial` an entity carries. `discoverFilterValues` (the **dropdown**'s value discovery, `filter-schema.ts:~280`) still sampled only `extractMaterialsOnDemand`'s primary (lowest-rel-express-id) association.

Net effect: for an entity with two or more material associations, a Name or Category carried only on the non-primary one matched `material=` but was never suggested in the dropdown. This is the same discoverability gap as #4814/#4780 (Name-vs-Category), one level deeper — at the association level.

Note on #4815 (open, unmerged): it fixes the Name-vs-Category gap by swapping the dropdown's loop to `materialMatchCandidates`, but the extractor call underneath stays `extractMaterialsOnDemand` (primary-only) — so it does not close this gap. This PR's fix supersedes #4815's change to that call site (same lines); whichever lands first, the other should be re-read against the result rather than re-applied on top.

## Fix

`discoverFilterValues` now walks `extractAllMaterialsOnDemand` (every association) and expands each via `materialMatchCandidates` (Name + Category) — the exact candidate set the matcher already uses — so the two surfaces cannot drift apart again.

## Cost

`discoverFilterValues` already samples at most `VALUE_SAMPLE_CAP = 5000` entities (`filter-schema.ts`). `extractAllMaterialsOnDemand` resolves one `MaterialInfo` per association instead of one total; per its own docstring, "most entities carry one" association, so the extra work only bites the rare multi-association entity, still bounded by the same 5000-entity cap. No extra WASM/parse cost — same in-memory `EntityExtractor` walk as before.

## Test plan

RED-then-GREEN in `apps/viewer/src/lib/search/filter-schema.test.ts` (new describe block, real STEP-source fixture parsed via `IfcParser().parseColumnar`, mirroring the existing `evaluateFilterRules` material fixture in `filter-evaluate.test.ts`):
- Wall-D: two `IfcRelAssociatesMaterial` — primary (lower rel express id) → "Common Mat", non-primary (higher rel express id) → "Rare Find".
- Wall-E: a single association → "Lone Value" (bounding control — proves the fix is additive, not a swap).
- Confirmed the fixture is real: `material=Common Mat` matches Wall-D via `evaluateFilterRules`.
- Confirmed the matcher already finds the non-primary value: `material=Rare Find` matches Wall-D via `evaluateFilterRules` (uses `extractAllMaterialsOnDemand` already).
- RED (pre-fix): `discoverFilterValues(store).materials` was `["Common Mat", "Lone Value"]` — "Rare Find" absent even though the matcher finds it.
- GREEN (post-fix): `discoverFilterValues(store).materials` includes "Rare Find", "Common Mat", and "Lone Value" (12/12 tests in the file pass).
- Mutation test: reverted the extractor/import back to `extractMaterialsOnDemand`, confirmed exactly 1 of 12 tests reddened (the non-primary-association test; the single-association bounding control stayed green), reverted the mutation — `diff /tmp/filter-schema.ts.fixed apps/viewer/src/lib/search/filter-schema.ts` reported identical, `git diff --stat` matched the pre-mutation file exactly.
- `filter-evaluate.test.ts` + `filter-match.test.ts`: 93/93 pass, unaffected.
- `filter-schema.test.ts` + `filter-evaluate.test.ts` + `filter-match.test.ts` together: 105/105 pass.
- `tsc --noEmit`: clean on the touched file.
- `node scripts/check-module-size.mjs` (plain form): exit 1, but every offender listed is a pre-existing `main` allowlist-drift / #4813 `packages/renderer` issue in files this PR does not touch (`filter-schema.ts` / `filter-schema.test.ts` do not appear anywhere in its output).

Closes #4817. Awaiting the maintainer's `ready` label per the issue-queue gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Material filter value suggestions now include values from all material associations, including non-primary associations.
  * Filtering by either primary or secondary material associations now returns the associated elements correctly.

* **Tests**
  * Added coverage for discovering and filtering elements with multiple material associations.


louistrue (approved):
Reviewed at exact head 158e00e. The restack preserves the all-associations material fix, splits the added regression coverage into an MPL-licensed test module to respect the repository module-size guard, and passes all eight hosted viewer shards, E2E, Claude, CodeRabbit, and the local Ollama review. Local targeted validation passed 105 tests plus viewer typecheck. Remaining aggregate red is inherited from current main, not introduced here.

Issue #4825, opened 2026-09-15
ChatPanel empty-state hint and other sub-AA panel text missed by #4792/#4793/#4794


BIMvoice:
A post-merge audit of #4793/#4794 found `ChatPanel`'s empty-state hint (\`apps/viewer/src/components/viewer/ChatPanel.tsx\`, the \`<p>Try something:</p>\` line, rendered when the chat has no messages) still uses a `text-muted-foreground/60` opacity tier — the same family #4792's survey measured as sub-AA. It was never in either changeset's scope: a gap in the original survey, not a botched merge.

Measured with the existing harness (`apps/viewer/src/test/contrast/render-harness.ts`) on its real surface (`bg-background`, inherited from `ChatPanel`'s root `<div className="h-full flex flex-col bg-background ...">`):

- light: 2.31:1 (AA requires 4.5:1)
- dark: 2.72:1
- colorful: 3.75:1

A `git grep -n "text-muted-foreground/[0-9]" -- apps/viewer/src` sweep for siblings found several other real, always-visible text sites in the same shape that #4792/#4793/#4794 never measured, plus a larger set of icon colors / `aria-hidden` decorations / `disabled`-control dimming that WCAG's text-contrast rule does not cover.

Real text confirmed below AA on their real surface (candidates for the same fix — drop the opacity suffix to plain `text-muted-foreground`, matching #4793/#4794's pattern):
- `LearnTab.tsx` "X min" tour-length readout (surface `bg-card`, the hand-rolled `KeyboardShortcutsDialog` panel)
- `BulkPropertyEditor.tsx` "(N found)" property/pset-count annotation (`bg-background`)
- `ByokKeyModal.tsx` pricing hint (`bg-background`)
- `ModelSelector.tsx` context-window readout (`bg-popover`)
- `ClassVisibilityMenu.tsx` visible/total class count (`bg-popover`)
- `MeasurePointReadout.tsx` `CoordRow` hint span (`bg-background`)
- `tools/measure-modes/geo-readout.tsx` `EnhLine` label span (`bg-background`)

Also identified but **not** mechanically fixable the same way, flagged for a maintainer call:
- `ExecutableCodeBlock.tsx`'s console duration readout sits on a literal, fully-opaque `bg-muted` — even the plain (no-opacity) `text-muted-foreground` measures under AA there (4.40 light / 4.69 dark / 3.15 colorful), so this needs an actual color change, not an opacity drop.
[...]

PR #4826, opened 2026-09-15
fix(viewer): raise ChatPanel empty-state hint and sibling sub-AA text to WCAG AA

BIMvoice:
## Summary

A post-merge audit of #4793/#4794 found `ChatPanel`'s empty-state hint (`apps/viewer/src/components/viewer/ChatPanel.tsx`'s `<p>Try something:</p>`, rendered when the chat has no messages) still on a `text-muted-foreground/60` tier — the same family #4792's survey measured as sub-AA. It was outside both changesets' scope: a gap in the original survey, not a botched merge.

**Measured** (real headless-Chromium harness, `apps/viewer/src/test/contrast/render-harness.ts`) on its real surface, `bg-background` (inherited from `ChatPanel`'s root `<div className="h-full flex flex-col bg-background ...">`):

| | light | dark | colorful |
|---|---|---|---|
| before (`/60`) | 2.31:1 | 2.72:1 | 3.75:1 |
| after (plain) | 4.83:1 | 5.42:1 | 6.66:1 |

AA for normal text requires 4.5:1.

**Sibling sweep**: `git grep -n "text-muted-foreground/[0-9]" -- apps/viewer/src` turned up ~40 more hits. Classified and fixed here, alongside the ChatPanel line, are the other real always-visible text sites confirmed below AA on their real surface (all raised to plain `text-muted-foreground`, all now clear AA in light/dark/colorful — see the table in #4825):

- `LearnTab.tsx` "X min" readout (`bg-card`)
- `BulkPropertyEditor.tsx` "(N found)" annotation (`bg-background`)
- `ByokKeyModal.tsx` pricing hint (`bg-background`)
- `ModelSelector.tsx` context-window readout (`bg-popover`)
- `ClassVisibilityMenu.tsx` visible/total count (`bg-popover`)
- `MeasurePointReadout.tsx` `CoordRow` hint (`bg-background`)
- `tools/measure-modes/geo-readout.tsx` `EnhLine` label (`bg-background`)

Left **unfixed and out of scope for this PR** (filed in #4825 for a maintainer call, since none of them fit the mechanical "drop the opacity suffix" fix):
- `ExecutableCodeBlock.tsx`'s duration readout — its real surface is a literal, opaque `bg-muted`, and even plain `text-muted-foreground` measures under AA there (4.40/4.69/3.15).
- `PanelWindowHost.tsx`'s two hints — real surface is `bg-muted/40`/`bg-muted/30` layered over `bg-background`; the harness can't composite two stacked layers without extension.
- `CoordinateDisplay.tsx`'s two spans — className is a template literal; `extractClassNameAfter` deliberately throws rather than mismatch on one.
- `LayerMergeSection.tsx`/`LayerReviewSection.tsx` placeholder text — only paints through `::placeholder`, which the harness's single-element render can't reproduce.

Everything else the grep matched is icon color, `aria-hidden="true"` decoration, or `disabled`-control dimming — outside WCAG 1.4.3's text-contrast rule.

Refs #4825

## Test plan

- [x] RED first: added the new pinned cases to `panel-secondary-text.test.ts` before touching source; all failed as expected.
- [x] Applied the fixes; re-ran `panel-secondary-text.test.ts` — 156/156 pass.
- [x] Ran `wcag.test.ts` and `tooltip-secondary-text.test.ts` — unaffected, all pass.
- [x] Mutation check: restored ChatPanel's `/60`, confirmed exactly its 3 pinned cases (light/dark/colorful) reddened and nothing else did, then reverted (`git diff --stat` clean against the fixed state).
- [x] `node scripts/check-module-size.mjs` (plain form) run — pre-existing, unrelated Rust module-size allowlist drift only; no file this PR touches is named.

This box has no real Chrome; `render-harness.ts`'s `chromium.launch({ channel: 'chrome' })` was patched locally to drop the `channel` option (uses Playwright's bundled Chromium instead) to run these tests, then restored byte-for-byte before committing (`git diff` on that file is empty).

## Summary by CodeRabbit

* **Accessibility**
  * Improved contrast for secondary text across the viewer, including chat hints, tour durations, property counts, pricing hints, model context details, visibility counts, and measurement labels.
* **Tests**
  * Expanded contrast coverage to verify these interface elements meet accessibility standards across themes.


louistrue (approved):
Reviewed at exact head fb22faa. The eight changes are one measured contrast defect class, each source surface is pinned by the real Chromium contrast harness across all themes, the old opacity tiers are non-vacuous regression cases, and the viewer changeset is present. Exact-head CI, Claude, CodeRabbit, and local Ollama are clean. The closing claim was corrected to Refs #4825 because the issue intentionally retains the token/placeholder/runtime-template follow-ups.

Issue #4827, opened 2026-09-15
Extend contrast harness for translucent surfaces and literal template-literal classNames (#4825 follow-up)


BIMvoice:
Follow-up to #4825, which named two harness limitations rather than product decisions among the sites it could not measure or fix:

1. `PanelWindowHost.tsx`'s header/footer hints sit on a translucent `bg-muted/40`/`bg-muted/30` surface — the harness measures a surface's own background only, and doesn't composite a translucent surface over its backdrop.
2. `CoordinateDisplay.tsx`'s two secondary spans build `className` as a template literal, which `extract-classname.ts` deliberately throws on rather than silently mismatching.

This tracks extending the harness for both, then measuring and fixing what's found.

BIMvoice:
## CoordinateDisplay.tsx — design fork for the remaining site

PR #4828 has resolved the harness limitations for translucent surfaces and static template-literal classNames, leaving one site unmeasurable for a different reason: **`apps/viewer/src/components/viewer/properties/CoordinateDisplay.tsx`** builds its className with a template literal that interpolates a runtime `primary` prop:

```tsx
<span className={`...${primary ? 'text-muted-foreground' : 'text-muted-foreground/60'}`}>
```

The extractor deliberately throws on interpolation (per its own design: "refuses to guess one or return a partial string. Either give this className a literal value, or extend this extractor for the specific statically-analysable shape this expression is..."). This is the correct behavior — the extractor cannot know which branch a real render picks.

Two defensible paths forward, both valid:

1. **Enumerate the branches in the fixture** — measure each value the ternary can produce (primary=true and primary=false) as separate test variants, pinning every reachable class combination. No production code change, but the fixture must be kept in step with the component by hand if its branches evolve.

2. **Refactor the component** — replace the interpolated string with a lookup keyed by the prop (e.g. `const textClass = primary ? 'text-muted-foreground' : 'text-muted-foreground/60'; ...<span className={textClass}>`) so classNames are static literals the extractor can read. Cleaner to measure, but it changes production code purely to serve testability.

Option 2 would mean changing production code for test reasons — a judgement call about house style rather than correctness, given the repo's preference against tests that assert on source text.

Which approach would you prefer?

**For completeness:** `PanelWindowHost` is now resolved in #4828; `ExecutableCodeBlock`'s token question is on #4825 (separate from the template-literal constraint).

PR #4828, opened 2026-09-15
fix(viewer): composite translucent contrast surfaces and extend template-literal extraction

BIMvoice:
Refs #4827, a follow-up to #4825's two acknowledged harness limitations (not product decisions).

## A. Translucent surface compositing

`measureTextContrastOnSurface` gains an optional `backdropClassName`. When given, the surface is rendered nested inside a backdrop element and composited onto it with the *same* canvas compositor already used to resolve translucent TEXT onto its surface (`resolveOverBackdrop`) — one compositor, reused, not a second one. Omitting the parameter keeps every existing call site's behavior identical (verified: 155/155 pre-existing contrast tests still pass unchanged).

Measured `PanelWindowHost.tsx`'s two sites (translucent `bg-muted/40`/`bg-muted/30` over `bg-background`):

| site | light | dark | colorful |
|---|---|---|---|
| header hint, `/70` (before) | 2.70 | 3.23 | 3.19 |
| footer strip, `/70` (before) | 2.73 | 3.23 | 3.25 |
| header hint, plain (after) | 4.67 | 5.16 | 6.22 |
| footer strip, plain (after) | 4.71 | 5.23 | 6.32 |

Both sites failed AA (4.5:1) in every theme; dropping the opacity suffix to plain `text-muted-foreground` (matching #4792/#4793/#4794's pattern) clears AA everywhere. Fixed in `PanelWindowHost.tsx`, with a `panel-window-host-translucent.test.ts` fixedCases + non-vacuousness pair (12 tests) mirroring this repo's existing convention.

Also revisited `MeasurePanel`'s documented `bg-background` proxy for its real `bg-background/95 backdrop-blur-sm` surface, now that translucency is measurable: composited over `bg-background`, it measures 4.83 / 5.16 / 5.32 (light/dark/colorful) — still clears AA everywhere, but with less margin than the old fully-opaque proxy suggested (was 4.83 / 5.42 / 6.66), most visibly in colorful. `backdrop-blur` itself is not modeled by alpha compositing, so this is still a proxy for the real floating-over-3D-viewport backdrop, just a more accurate one — documented in the test file's updated comment.

## B. Template-literal className extraction

`extract-classname.ts` now extracts a `className={\`...\`}` template literal when it is statically analysable — i.e. contains no `${...}` expression at all (pure literal text, e.g. backticks used for no functional reason). Any actual `${...}` interpolation still throws loudly, exactly as the file's existing "don't silently mismatch" design intends, since the extractor cannot know which branch a real render would pick.

Verified this does **not** unlock `CoordinateDisplay.tsx` (#4825's third named gap): both its secondary spans interpolate a real `primary` prop (`` `...${primary ? 'a' : 'b'}` ``), not literal text, so they still throw — the issue's own conclusion ("needs a small component refactor, not just a class edit") holds. Confirmed with dedicated tests against the real file (`extract-classname-template-literal.test.ts`).

**Open design question for a maintainer**, not resolved here: should the extractor go further and support a ternary between two string-literal branches (`${cond ? 'a' : 'b'}`) as a further statically-analysable case — returning multiple variants — or is `CoordinateDisplay`'s own suggested refactor to a plain className the better fix? Either unlocks measuring it; I didn't want to guess at the extractor's result-type shape for "multiple possible classNames" without your steer.

## Testing

- RED-then-GREEN for both new capabilities (a synthetic backdrop-compositing probe; the template-literal fixture tests), then removed the RED probe.
- Mutation-tested: reverted the `PanelWindowHost` fix (fixedCases redden 6/6), reverted `render-harness.ts`'s backdrop-vs-self compositing branch (4/12 fixedCases redden), and disabled the interpolation-guard `if` (the throw-loudly test reddens) — each confirmed the guard actually executes, then reverted (`git diff --stat` clean afterward at each step).
- Full contrast suite: **171/171 pass** (155 pre-existing + 16 new). #4826 (open, unmerged) reported 156/156 on its own branch; this branch is based on current `main` (`1f8738a3b`), which doesn't yet include #4826's fixes, so 155 is the correct pre-existing baseline here.
- `node scripts/check-module-size.mjs` (plain form): no failures in any file this PR touches. It does report a pre-existing, unrelated Rust module-size-allowlist drift against `origin/main` — no Rust file is touched by this PR.
- `#4826` is open and also touches `panel-secondary-text.test.ts`, but in different, non-overlapping regions (it appends new `fixedCases`/`regressedCases` entries elsewhere; this PR only touches the `fixedCases` array's type, the `MeasurePanel` entry, and the loop's call to `measureTextContrastOnSurface`). Not stacked on its branch; flagging here per repo convention in case a rebase is needed after it merges.

## Local-only note

This sandbox has no real Google Chrome (`render-harness.ts` launches `channel: 'chrome'`). Locally patched `chromium.launch({ headless: true, channel: 'chrome' })` → `chromium.launch({ headless: true })` to run every test above against Playwright's own bundled Chromium, then restored the line byte-for-byte before committing (confirmed via `git diff` showing no `channel` line change).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text contrast for labels and synchronization hints in popped-out viewer windows.
  - Improved readability of secondary text on translucent panels across light, dark, and colorful themes.

- **Accessibility**
  - Strengthened WCAG AA contrast compliance for panel headers, footers, and secondary text.

- **Tests**
  - Expanded contrast coverage for translucent surfaces, backdrop compositing, and template-based styling to help prevent accessibility regressions.


louistrue:
## PR Reviewer Guide 🔍



#### (Review updated until commit https://github.com/LTplus-AG/ifc-lite/commit/7457cf9ee3e1ac4d0997bfe6c17e32e8e80f6c27)


<table>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/LTplus-AG/ifc-lite/pull/4828/files#diff-8852de8d3f7ea1a4e9638be583112eebcde3d087c9c9fca272320cc935229c4aR67-R67'><strong>Potential Test Fragility</strong></a>

The test for the "footer" case uses an empty string for `textClassName` because the text inherits its color from the parent div. While this correctly mirrors the DOM structure, it relies on the `extractClassNameAfter` logic and the `render-harness` to handle an empty string correctly. If the `textClassName` is empty, `resolveOverBackdrop` will be called with an empty string for the text color. If the underlying implementation of `resolveOverBackur` doesn't handle empty strings by falling back to the parent's computed style, this test might become brittle or fail if the CSS hierarchy changes.
</summary>

```typescript
textAnchor: null, // text inherits the surface div's own color, see file doc comment
```

</details>

</td></tr>
</table>

<hr>

- Tokens: 11,958 in / 240 out / 12,198 total
- Time cost: 13.8s
- AI calls: 1

</details>

<details>
<summary>✅ Resolved findings</summary>

### apps/viewer/src/test/contrast/panel-window-host-translucent.test.ts:67

**Potential logic error in test** In the `footer` case, `textAnchor` is set to `null`, and the test logic uses an empty string `''` for `textClassName`. While this correctly tests CSS inheritance for the footer, it might fail if the `extractClassNameAfter` function (or the logic surrounding it) expects a non-empty string or if the `measureTextContrastOnSurface` function handles an empty string differently than a missing class.

</details>

louistrue:
**[Persistent review](https://github.com/LTplus-AG/ifc-lite/pull/4828#issuecomment-5675517477)** updated to latest commit https://github.com/LTplus-AG/ifc-lite/commit/7457cf9ee3e1ac4d0997bfe6c17e32e8e80f6c27

louistrue (approved):
Exact head 7457cf9e verified: full hosted CI, all eight viewer shards, E2E, typecheck, lint, Claude, CodeRabbit, and local Ollama review are green. Focused extractor tests also passed locally after the template-title lint correction.

Issue #4786, opened 2026-09-14
Rust crate semver gate passes on zero executed checks whenever the release carries a major


louistrue:
## What is wrong

`scripts/check-rust-semver.mjs` invokes `cargo-semver-checks` with no `--release-type`, so the tool infers the bump from the manifest version against the crates.io baseline:

```js
// scripts/check-rust-semver.mjs:324-340
function realRunSemverChecks(crate, baseline) {
  const res = spawnSync(
    'cargo',
    [`+${SEMVER_TOOLCHAIN}`, 'semver-checks', '--package', crate,
     '--baseline-version', baseline, '--color', 'never'],
    ...
```

When the inferred bump is a **major**, `cargo-semver-checks` has nothing left to refuse, so it skips its entire lint set and exits 0. `interpretRun` matches `Summary no semver update required` and returns `required: 'patch'`, which is ranked below the carried `major`, so the gate reports the crate as checked and clean. Nothing was compared.

This is the "absence of a measurement shares a channel with its result" shape: a crate whose API was never examined and a crate that was examined and found clean both exit through the same success line.

## Measured

cargo-semver-checks 0.50.0, baseline the real published `14.0.0` from crates.io, on a tree that merges the heads of #4749, #4750 and #4752 onto main at `6fa3d1425`. That tree contains five real major-requiring breaks.

Forcing the discriminating bump level:

```
--baseline-version 14.0.0 --release-type minor
  ifc-lite-processing  196 checks: 193 pass, 3 fail, 58 skip
    semver requires new major version: 3 major and 0 minor checks failed
  ifc-lite-export      196 checks: 195 pass, 1 fail, 58 skip
    semver requires new major version: 1 major and 0 minor checks failed
  ifc-lite-geometry    196 checks: 195 pass, 1 fail, 58 skip
    semver requires new major version: 1 major and 0 minor checks failed
```

The same tree, same baseline, at the bump level the release gate will infer once `majorOffset` goes to 6 and the manifests read 15.0.0:

```
--baseline-version 14.0.0 --release-type major
  ifc-lite-processing  0 checks: 0 pass, 254 skip
    Summary no semver update required
  ifc-lite-export      0 checks: 0 pass, 254 skip
    Summary no semver update required
  ifc-lite-geometry    0 checks: 0 pass, 254 skip
    Summary no semver update required
  exit 0
```

[...]

PR #4829, opened 2026-09-15
fix(ci): fold cargo-semver-checks' warn-level lints into the Rust semver gate

### LINKED ISSUE #4800: Rust crate semver gate discards warn-level lint results, so a detected repr(C) field reorder reads as clean
## What is wrong

`scripts/check-rust-semver.mjs` reads `cargo-semver-checks`' `Summary` line to decide what bump a crate's API change requires. That line is driven by the tool's `required_bumps`, which counts lint failures at ERROR level only. The 14 lints that carry `lint_level: Warn` feed a separate `suggested_bumps` and print on their own line:

```
     Checked [   0.007s] 223 checks: 222 pass, 0 fail, 1 warn, 31 skip
     Summary no semver update required
     Warning produced 1 major and 0 minor level warnings
                  produced warnings suggest new major version
```

`interpretRun` (now `scripts/lib/cargo-semver-checks.mjs`) matches `Summary no semver update required` and returns `required: 'patch'`. The gate then clears the crate. The tool found a break, said so, and the gate did not read the line it said it on.

## Measured

Read from the pinned tool's own source, cargo-semver-checks 0.50.0 in the local registry, not inferred from the flag:

- `src/check_release.rs:593-601`: `Summary no semver update required` is the `else` branch of "`required_bumps` has an update type". Warnings do not enter it.
- `src/check_release.rs:603-625`: warnings print afterwards, as `Warning produced N major and M minor level warnings`, with `produced warnings suggest new <level> version` beneath.
- Lint census over `src/lints/*.ron`: 254 lints total, of which **14 carry `lint_level: Warn`**, 31 carry `lint_level: Allow` (never selected at any release type) and the rest are errors.

Two of the 14 are `repr_c_plain_struct_fields_reordered` and `repr_c_enum_struct_variant_fields_reordered` — which is the exact hazard `check-rust-semver.mjs`' own header names as the one it cannot see:

> the C ABI of `ifc-lite-ffi` as an ABI: the `extern "C"` functions are compared as Rust items, so a signature change is seen, but a `#[repr(C)]` struct's meaning changing under a stable layout is not.

The header is right that a layout-stable meaning change is invisible. It is wrong by implication about a field REORDER: cargo-semver-checks does detect that one, at warn level, and the gate throws the detection away.

## Not the same hole as #4786, and not closed by it

#4786 was the gate reaching a clean verdict over a scan that executed nothing. Its fix forces `--release-type patch` and refuses a verdict backed by zero executed lints. Neither half touches this: a warn-level break comes back with a full 223-lint tally, so the executed floor passes it, and the summary reads clean, so `RANK` clears it.

Same class ("a verdict that reads clean over a break the tool found"), different mechanism (warn versus skip), which is why it is filed separately rather than widened into #4786's PR.

## Remedy

Parse the `Warning produced N major and M minor level warnings` line in `interpretRun` and fold it into `required`, so a warn-level major counts as a major. `suggested_bumps` is exactly the tool's answer to "what would this require if you did not choose to be lenient", and this gate's whole job is to be less lenient than the default.

Two things to settle when doing it:

- Whether to raise the 14 warn lints to errors with `--deny warn` (the tool's own flag) instead of parsing a second line. That is simpler and moves the decision into the tool, where it is already expressed.
- Whether any of the 14 fires today on the published crates. Worth measuring before the gate starts failing on it, so the first run of the fix is not also the first discovery.

## Found by

`/code-review` on the #4786 fix, verified against the tool's source and lint definitions. Deliberately not fixed there: it is a different mechanism and the #4786 PR does not otherwise touch what the summary line means.

BIMvoice:
## What was wrong

`interpretRun` (`scripts/lib/cargo-semver-checks.mjs`) reads `cargo-semver-checks`' `Summary` line to decide the bump a crate's API change requires. That line is driven by the tool's `required_bumps`, which counts lint failures at ERROR level only. The 14 lints that carry `lint_level: Warn` — including `repr_c_plain_struct_fields_reordered` and `repr_c_enum_struct_variant_fields_reordered`, the exact `#[repr(C)]` hazard `check-rust-semver.mjs`'s own header names as ifc-lite-ffi's blind spot — feed a separate `suggested_bumps` and print on their own line beneath a clean `Summary no semver update required`. The gate read the clean summary and cleared the crate.

This is not closed by #4801's `NO_CHECKS_EXECUTED` floor: a warn-level break comes back with a full 223-lint tally, so the executed floor passes it, and the summary reads clean, so `RANK` clears it. Filed separately as #4800 by `/code-review` on the #4801 diff, and explicitly deferred there rather than widened into that PR.

## Measured, not inferred

Built a two-crate local fixture with a `#[repr(C)]` struct whose fields are reordered between versions, and ran the actual pinned tool (cargo-semver-checks 0.50.0) against it via `--baseline-root`:

```
$ cargo +stable semver-checks --baseline-root ../old --release-type patch --color never
     Checked [   0.010s] 223 checks: 222 pass, 0 fail, 1 warn, 31 skip
--- warning repr_c_plain_struct_fields_reordered: struct fields reordered in repr(C) struct ---
     Summary no semver update required
     Warning produced 1 major and 0 minor level warnings
             produced warnings suggest new major version
    Finished [   1.346s] repro
$ echo $?
0
```

Confirms #4800's transcript exactly: exit 0, full lint tally, a real detected break, and a clean `Summary`.

## The fix

`interpretRun` now also reads the tool's own `produced warnings suggest new <level> version` line and folds it into `required`, taking whichever of the error-driven `Summary` verdict and the warn-driven suggestion is the **larger** bump — never the smaller, so a warning can only raise the bar an error already set, not lower it. The decision stays exactly where #4801 put the tool-adapter/gate-policy split: this module reports the verdict, `check-rust-semver.mjs`'s `RANK` comparison still decides what to do with it.

Not "raise the 14 warn lints to errors with `--deny warn`" (the alternative the issue named): that would change what the tool itself reports and would need re-verifying the whole `--release-type patch` lint-selection measurement from #4786/#4801. Parsing the suggestion line keeps the argv and the executed-count floor untouched and is a two-line diff.

## Tests

RED first: `WARN_MAJOR`/`WARN_MINOR` fixtures transcribed from the real run above; `interpretRun(WARN_MAJOR).required` asserted `'major'` failed before the fix (`'patch' !== 'major'`), reddening for the exact mechanism #4800 describes — not caught by any existing #4786/#4801 case, since the fixture carries a full 223-lint tally.

Mutation-tested each new guard, confirmed the expected test reddens and that the mutated line executed, reverted all:

| Mutation | Test(s) reddened |
|---|---|
| Drop `raise()` in the clean-Summary branch | `RED (#4800): a detected warn-level break is not discarded as patch` |
| Drop `raise()` in the error-Summary branch | `a warn-level suggestion RAISES a verdict an error alone set lower` |
| Break the `produced warnings suggest...` regex so it never matches | both of the above |
| Flip the rank comparison (`>` to `<`) | all three warn-folding tests |

`git diff --stat` after reverting the last mutation showed no diff against the un-mutated fix.

`node --test scripts/check-rust-semver.test.mjs`: 40 pass (37 pre-existing + 3 new).

## Gates

- `node --test scripts/check-rust-semver.test.mjs` — 40 pass
- `node scripts/check-module-size.mjs` (plain form) — fails, but on rows this PR does not touch (`rust/processing/tests/module_size_allowlist.txt` entries drifted against `origin/main`, the fork remote, which is stale against `upstream/main` per repo norms); this PR touches no Rust source and no allowlist file
- No changeset: scripts + tests only, no Rust source touched, nothing a published crate ships changes — same call #4801 made for the same shape of change

## Scope

Only the warn-line folding. Left alone, per the issue: whether to instead raise the 14 warn lints to errors with `--deny warn` (a tool-level alternative, not attempted here), and whether any of the 14 fires today on the published crates (not measured against the real crates.io baselines — this PR's fixture is synthetic, run locally with `--baseline-root`, precisely to avoid needing that measurement to prove the mechanism).

Closes #4800


louistrue (approved):
Reviewed at exact head 7001a76. The parser raises, but never lowers, the Summary-derived bump with cargo-semver-checks’ warn-level suggestion. Exact-head CI and Claude/local Ollama are clean; the standalone semver suite passes locally, 40/40.

Issue #4420, opened 2026-09-09
Appearance: coordinated multi-model assignments and atomic history (F7)


louistrue:
## Outcome
One Appearance action can target explicitly chosen IFC models with a reviewable assignment order and exceptions, publish all prepared changes together, and Undo/Redo them together. This is F7 of the consolidated appearance roadmap, authorized as part of the full implementation.

## Scope
Stack reviewable changes against this issue after F6 target preparation is usable. Reuse the existing source library, scope catalog, native planners, preview leases, model assets and command/history infrastructure. Freeze each model revision and explicit product membership before preparation; no second IFC writer or load path. Query/assignment recipes must preserve source identity and ordered exceptions across supported restore, and report membership differences after reload before applying again. Shared-room viewing of already committed results is separate from concurrent authoring.

## Acceptance
- Preview compares the complete selected scope across two real IFC models, including colliding local express IDs and repeated mapped occurrences.
- Failure/cancellation/staleness in any model leaves every model, renderer resource, image owner and history unchanged. Inject a late failure after another model has prepared, and during coordinated publication.
- One Apply and one Undo/Redo cover every destination; removal/reload and history disposal release the correct resources without restoring stale geometry.
- Assignment order and explicit exclusions have deterministic behavior; restored recipes report changed membership rather than silently authoring new objects.
- Normal per-model/subset portable export and fresh imports preserve texture coordinates, assets and selection; committed room reopen retains the results.
- Mounted behavior tests plus actual browser and independent IFC evidence, root validation, matching documentation and published-package changesets accompany each stack.

The issue stays open until this complete supported journey is demonstrated. Broader reconstruction/semantic research remains separately gated.

PR #4831, opened 2026-09-15
feat(viewer): convert AppearanceAssignmentList to the i18n catalogue

BIMvoice:
## Summary
- Second consumer of the string-catalogue i18n mechanism (#4810), which established the pattern with `MergeLayersBanner`. This converts `AppearanceAssignmentList`'s fixed strings — the section heading, description, aria-label, and "Review objects and exceptions" button — to `useTranslation()`/`t(key)`, namespaced under `appearanceAssignmentList.*` in `en.ts` (now 31 lines).
- Chose `AppearanceAssignmentList` because it is self-contained (pure props, no store/network coupling), already has a rendering test (`AppearanceAssignmentList.test.tsx`), and — unlike `MergeLayersBanner` — exercises shapes the first conversion didn't: a per-row **interpolated** aria-label (`` `Assignment ${index+1}: ${item.source.name} on ${item.model.name}` ``) and **pluralised** text (`1 object` / `N objects`).
- Rejected: `ScanTransferFields.tsx` (far too large/interpolation-heavy — one of the "largest panels" the task says to avoid), `WidgetErrorBoundary.tsx` (a class component whose only user string is itself an interpolated `${label} crashed while rendering`, so picking it would mean stopping immediately with nothing else to show), `ModelTagsCommand.tsx` (only one string, too thin to validate the pattern further).

## Finding: the i18n mechanism has no interpolation or pluralisation support
`useTranslation()`'s `t(key: TranslationKey): string` takes no arguments — `registry.ts#resolve` returns a static catalogue value only. `AppearanceAssignmentList`'s per-row aria-labels (`Move assignment N earlier`, `Assignment N: <name> on <model>`, ...) and its `N object(s)` count are built from live index/name/count data and **were deliberately left hardcoded** rather than hand-rolling a second templating scheme in the component, per the task's instruction. Extending the catalogue with interpolation/pluralisation is a design decision for whoever picks that up next — #4810 was explicitly scoped as a pattern-setter, not a finished system.

## Verification
- RED-first: `AppearanceAssignmentList.i18n.test.tsx` — added while the component was still hardcoded; its "registers a locale override, falls back to English for an omitted key" test failed (the English-only test passed trivially, same as `MergeLayersBanner`'s did, since hardcoded and catalogue text agree).
- GREEN after conversion; existing `AppearanceAssignmentList.test.tsx` (#4420) still passes unchanged — it exercises the dynamic aria-labels/counts, which weren't touched.
- Mutation-tested both guards: (1) mutated the `reviewButton` catalogue value in `en.ts` — the English-default test reddened; (2) mutated the component to hardcode the heading instead of calling `t(...)` — the locale-override test reddened. Reverted both; `git diff --stat` after revert matches the pre-mutation diff exactly.
- `node scripts/check-source-text-assertions.mjs`: `OK (33 allowlisted, 171 marked, 0 new)` — no new source-text assertion.
- `node scripts/check-module-size.mjs` (plain form): exits 1, but with an identical set of failing rows (all pre-existing Rust allowlist entries, unrelated to this change) confirmed by running the same command against an unmodified `upstream/main` checkout in a throwaway worktree.
- Revert oracle (`node scripts/check-test-revert-oracle.mjs --base upstream/main --head HEAD`): **OBSERVED** — reverting the production hunk turned the locale-override assertion red (not a build failure), confirming the test imports `@/i18n` and fails on assertion as expected, not `REVERT-BROKE-BUILD`.

## Changeset
None. This is a pure string-to-catalogue move with the identical rendered English text and no behaviour change (same rationale #4810 used for `MergeLayersBanner`).

Refs #4785 (does not close it — one more converted component, not the full migration)

## Test plan
- [x] `AppearanceAssignmentList.i18n.test.tsx` and `AppearanceAssignmentList.test.tsx` pass
- [x] `check-source-text-assertions.mjs` clean
- [x] Revert oracle reports OBSERVED


louistrue (approved):
Reviewed at exact head 036c46a. This is a scoped catalogue conversion of fixed AppearanceAssignmentList strings, keeps live-data interpolation/pluralization unchanged, and exercises real rendered English, locale override, and fallback behavior. Exact-head CI and Claude/local Ollama are clean.

PR #4844, opened 2026-09-15
fix(viewer): parse overlays in the emitted RTC frame

### LINKED ISSUE #4799: RTC frame: one home for the Z-up to Y-up swap, and the CoordinateHandler needs-shift vote (#4611 remainder)
Follow-up to #4611, which closed its named defects. Two things are left, both refactors with no measured user-visible defect behind them, which is why they were split out rather than grown into the PRs that fixed the rest.

Everything below was read from code at the cited `file:line` during #4798. Nothing here was run, except where it says it was.

---

## 1. The IFC Z-up to viewer Y-up RTC swap has no home

`CoordinateInfo.wasmRtcOffset` is recorded in IFC axes (Z-up) while `CoordinateInfo.originShift` is recorded in renderer axes (Y-up), so composing them needs the swap `viewer = (rtc.x, rtc.z, -rtc.y)`. That rule is currently written out **eight times**, read one component at a time in **seven more places**, and wrapped in **four different helper families under three different names**. Nothing enforces that they agree.

Type: `CoordinateInfo` is declared in `packages/geometry/src/types.ts:299`, field `wasmRtcOffset?: Vec3` at `:305-306`. Three more structural copies of the same interface exist at `packages/geometry/src/coordinate-handler.ts:50`, `packages/server-client/src/types.ts:123` and `apps/viewer/src/utils/localParsingUtils.ts:30`.

### The full three-axis swap, spelled inline (8)

| # | file:line | what it spells |
|---|---|---|
| 1 | `apps/viewer/src/hooks/ingest/federationAlign.ts:123` | `const rtcYup = rtc ? { x: rtc.x, y: rtc.z, z: -rtc.y } : …` inside a local `totalYupOffset` |
| 2 | `apps/viewer/src/lib/geo/ifc-origin.ts:66` | the same line inside a **second, independent** `totalYupOffset` |
| 3 | `apps/viewer/src/lib/geo/map-absolute.ts:34-36` | the same, inline |
| 4 | `apps/viewer/src/lib/geo/cesium-bridge.ts:279-281` | the same, inline |
| 5 | `apps/viewer/src/hooks/useDrawingGeneration.ts:337` | `const shift = rtc ? { x: rtc.x, y: rtc.z, z: -rtc.y } : ci.originShift;` |
| 6 | `apps/viewer/src/lib/compare/geometrySummary.ts:78-80` | the composition open-coded inside `renderToWorldShift` |
| 7 | `apps/viewer/src/hooks/scanSectionMath.ts:176-180` | the composition open-coded inside `pointCloudRenderFrameShift` |
| 8 | `apps/viewer/src/components/viewer/tools/measure-modes/coordinates.ts:79-82` | `ifcToViewerAxes` / `viewerToIfcAxes`, the best-documented spelling and the only one that normalises signed zero |

### Partial reads, one or two components (7)

| # | file:line | component |
|---|---|---|
| 9 | `apps/viewer/src/lib/geo/reproject.ts:661-663` | x and `z = -rtc.y` only; the vertical term is deliberately absent because only the XZ footprint ring is built |
| 10 | `apps/viewer/src/lib/geo/cesium-placement.ts:172` | vertical only, `wasmRtcOffset?.z ?? 0` |
| 11 | `apps/viewer/src/lib/geo/kmz-altitude-hint.ts:57` | vertical only |
| 12 | `apps/viewer/src/lib/geo/kmz-export.ts:65` | vertical only, inlined with no named local |
| 13 | `apps/viewer/src/hooks/symbolic-parse-cache.ts:147` | vertical only |
| 14 | `apps/viewer/src/hooks/dxfUnderlayMath.ts:188-190` | vertical only, inverse direction |
| 15 | `apps/viewer/src/hooks/dxfUnderlayMath.ts:157-160` | **not** a Y-up swap: IFC plan axes, keeps `rtc.y` un-negated on purpose |

`apps/viewer/src/lib/wall-rects-from-meshes.ts:110-111` is in the same category as 15: it reads `rtc.x`/`rtc.y` as IFC X/Y with no swap, correctly, and restates the full identity in prose at `:102-103`. Whatever is done here must leave 15 and this one alone; they are not copies, they are a different target frame.

### The four helper families, and who imports which

- `totalYupOffset` exists **twice**, with disjoint importers, which is the sharpest drift hazard here:
  - `apps/viewer/src/lib/geo/ifc-origin.ts:63` → imported by `hooks/symbolic-parse-cache.ts:27`, `hooks/ingest/pointCloudAlignment.ts:95`, `lib/geo/useAnchorGeoreference.ts:27`
  - `apps/viewer/src/hooks/ingest/federationAlign.ts:120` → imported by `lib/appearance/capture/region.ts:6`, `lib/appearance/reference-runtime/frame.ts:6`, `lib/appearance/preview.ts:10`, `lib/appearance/native-mesh-frame.ts:5`, `lib/appearance/scan/transfer-frame.ts:5`, `lib/appearance/scan/landmarks.ts:11`
- `renderToWorldShift` (`lib/compare/geometrySummary.ts:74`) → `lib/compare/describeChange.ts:34`, `lib/compare/exportReport.ts:35`
- `pointCloudRenderFrameShift` (`hooks/scanSectionMath.ts:173`) → used only inside its own file (`:290`)
- `dxfWorldShift` (`hooks/dxfUnderlayMath.ts:154`) → `hooks/dxfExportGeoref.ts:93`, `components/viewer/Section2DPanel.tsx:41`
- `ifcToViewerAxes` / `viewerToIfcAxes` (`components/viewer/tools/measure-modes/coordinates.ts:66, 79`) → `components/viewer/PropertiesPanel.tsx:45`, `components/viewer/tools/MeasurePointReadout.tsx:50`, `lib/zones/emit-spatial-zones.ts:49`, `lib/geo/entity-world-position.ts:35`

### Two traps for whoever does this

**Signed zero is not uniform across the copies.** `coordinates.ts:79-82` normalises with `0 - p.y` and documents why (`-0` is not a coordinate anybody authored and leaks into equality checks and serialised output). `geometrySummary.ts:78-80` writes `- (rtc?.y ?? 0)` and `scanSectionMath.ts:176-180` writes `-(rtc?.y ?? 0)`, both of which yield `-0` for a zero input. Collapsing them onto one implementation therefore CHANGES the sign of zero at some call sites. `grid_lines_tests.rs` already documents that the symbolic goldens record sign of zero deliberately, so this needs checking against the suites rather than assuming it is cosmetic.

**A grep-based drift gate hits prose.** Eleven comments restate the identity without executing it: `lib/compare/geometrySummary.ts:66`, `hooks/scanSectionMath.ts:28` and `:162`, `hooks/dxfUnderlayMath.ts:170-171`, `lib/wall-rects-from-meshes.ts:103`, `hooks/useDrawingGeneration.ts:332`, `lib/geo/ifc-origin.ts:11`, `lib/geo/reproject.ts:391-396`, `lib/overlay-parse/symbolic-parse.ts:92-93`, `lib/geo/kmz-export.ts:54`, `components/viewer/PropertiesPanel.tsx:368-378`. The repo already has the right shape for this: `scripts/check-csv-escaper-copies.mjs` (a stateless per-line scan with a `KNOWN_REMAINING` ratchet plus a separate exact-line `PROSE_MENTIONS` registry) and its sibling `scripts/check-step-escaper-copies.mjs`. Copy that design rather than inventing one. Note the inversion that bites people here: `scripts/check-source-text-assertions.mjs` FORBIDS new `readFileSync`-plus-`includes` tests under `apps/viewer/src`, so the drift guard has to live in `scripts/` as a gate, not in a vitest file. Wiring is `package.json` (`check:csv-escapers` at `:56` is the model) plus a step in `.github/workflows/test.yml` (`:1255-1277`), script and self-test as two separate steps.

### Nothing in `packages/`

`packages/geometry/src/index.ts:446-447, 509, 538-539` and `packages/geometry/src/geometry-parallel.ts:700-713` pass `rtc.x/y/z` through in IFC axes to the wasm bridge without swapping, correctly. `packages/bcf/src/overlay.ts:140` and `packages/bcf/src/viewpoint.ts:107` spell the same `(x, z, -y)` shape but for BCF camera points, which is a different subject.

---

## 2. `CoordinateHandler` still votes on whether wasm already shifted

`packages/geometry/src/coordinate-handler.ts:461-496` counts how many meshes in a batch have a first vertex inside the 10 km threshold and concludes "wasm RTC was likely applied" when more than half do. That is a guess at a fact the wasm pre-pass already states authoritatively.

**#4798 does not make it removable, and here is why.** I mapped every production call site. There is exactly one production instance of the class, at `packages/geometry/src/index.ts:258`. The vote itself runs in four places:

| call site | has the authoritative answer in scope? |
|---|---|
| `packages/geometry/src/index.ts:558` (`processStreamingBytes` per batch) | **yes** — `rtc.needsShift` at `:540` is literally what was passed into `api.processGeometryBatch`, i.e. whether wasm subtracted the offset |
| `packages/geometry/src/geometry-parallel.ts:490` (worker-pool batches) | **yes** — `prepassMeta` is populated from the `'meta'` event at `:1168-1174`, and `sendStreamStartIfReady` (`:695-706`) resolves the frame and posts `needsShift` to the workers at `:728`, strictly before any worker batch can arrive |
| `packages/geometry/src/index.ts:875` (adaptive small-file) | **partly** — yes for the wasm sub-branch (`:869` → `applyPrePassMetadata` `:434`), no for the native sub-branch (`:865`) |
| `packages/geometry/src/index.ts:645` (**native buffer streaming**) | **no** | 

That last row is the reason the vote has to stay. `index.ts:645` passes `processMeshes: (meshes) => processMeshesIncremental(meshes)` as an override into `streamNativeGeometry`, which deliberately opts OUT of the trusted native path at `packages/geometry/src/geometry-native.ts:135` (that path sets `wasmRtcDetected = true` unconditionally at `coordinate-handler.ts:539-542` and never votes). There is no pre-pass on the native bridge at all: `setWasmMetadata` is called from exactly two places, `index.ts:417` and `geometry-parallel.ts:706`, both on the wasm path. So at `index.ts:645` there is no `rtcOffset`/`needsShift` to read, and the vertex-magnitude vote is the only signal available.

So the work is: make the three wasm call sites consume `RtcFrame.needsShift` instead of voting, and leave the vote reachable only where no pre-pass answer exists, with a comment saying which caller that is. Six existing tests pin the vote's current behaviour and would need rewriting rather than deleting, because none of them drives it through a real caller alongside a pre-pass answer: `packages/geometry/src/geometry.test.ts:429, 460, 546, 570, 605, 622` and `packages/geometry/src/coordinate-handler-threshold.test.ts:66`.

---

## 3. What #4798 deliberately did NOT close

#4798 made the RTC sample window a property of the file rather than of whichever pipeline was asking, and the three fixtures the issue named now agree exactly. One case is still open and it is not the same defect class.

The STREAMING browser pre-pass emits its frame MID-SCAN, before the file has been read, so it can only sample the part its entity index covers (`MetaMode::StreamingPartial { scanned_through }` in `rust/processing/src/stream_meta.rs`). The overlays (`MeshFrame::for_overlay`, `rust/processing/src/mesh_frame.rs`) sample the whole file. On a model whose head does not represent the rest of it, streamed meshes and overlays therefore still land in different frames — grid and alignment lines drawn off the model, the same symptom #4611 was about.

Bounding the streaming window to the head is not a shortcut, it is load-bearing: sampling the whole file there decodes the entire tail to watch every entity in it abstain against a partial index. Measured during #4798 on seven fixtures: all seven walked to EOF, 235 ms on a 343 MB model and 148 ms on a 177 MB one, 96x to 575x the previous cost, every one of them returning the anchor the head alone already gave, and all of it inside the mid-scan emission that exists to hold time-to-first-geometry at ~3 s on a 986 MB file. So the fix is not "widen the streaming window".

The fix is to stop recomputing: hand the emitted frame to the overlay parse APIs (`parseGridLines` / `parseGridAxes` in `rust/wasm-bindings/src/api/grid_lines.rs:301, 315`, `parseAlignmentLines` in `rust/wasm-bindings/src/api/alignment_lines.rs:46`, and the symbolic path which already has the shape via `extract_symbolic_data_with_provenance_in_frame`, `rust/processing/src/symbolic/mod.rs`), and have the viewer pass the frame it emitted. The overlay request payload carries no frame today (`apps/viewer/src/lib/overlay-parse/index.ts:196`; the worker is `apps/viewer/src/lib/overlay-parse/overlay-parse.worker.ts:86-103`), so this is an additive change to the overlay wire contract — which is exactly why it was not folded into #4798.

There is a second, independent reason recomputation cannot be made to agree under federation: `packages/geometry/src/geometry-parallel.ts:700-701` resolves the browser mesh frame through `resolveRtcFrame(prepassMeta, sharedRtcOffset)`, where a federation-wide `sharedRtcOffset` OVERRIDES the file's own detected anchor and forces `needsShift`. The overlay wasm never sees that override, so on a federated load the mesh anchor is not a function of the file at all and no amount of window canonicalisation closes the gap. Passing the frame does.

---

## Operational context a fresh session will not have

- **The geometry census does not run on PRs.** If you touch `rust/geometry`, `rust/core` or `rust/processing`, run it locally before merging, or you land what #4579 landed (green PR, broken main, #4610):
  ```
  node scripts/fixtures/fetch-fixtures.mjs --check || node scripts/fixtures/fetch-fixtures.mjs
  cargo test -p ifc-lite-geometry --features triangulation-alt --test triangulation_invariance -- --nocapture
  ```
  Run it on your merge base as well and diff the two summaries. The corpus currently carries one pre-existing `IMPROVED` row on `ara3d/ISSUE_129_…` that is present on main too; without the null comparison you will think it is yours. Never regenerate the golden to make it pass.
- **`majorOffset` is 6 on main** (`rust-major-offset.json`), so the crates publish at 15.0.0 while npm stays on 9.x. A public Rust break APPENDS a sentence to the `reason` string riding that pending 15.0.0 major and does **not** touch the integer. Measure it rather than asserting it: `cargo +stable semver-checks -p <crate> --baseline-version 14.0.0 --release-type minor`. The `--release-type minor` matters — against a major baseline the run skips everything and exits 0. Note `cargo-semver-checks` refuses the repo's pinned nightly, hence `+stable`. Also note `scripts/check-rust-api-surface.mjs` tracks TYPES, not inherent methods, so it stays green through a method removal; it is not the gate that catches this.
- **The revert oracle** ("Changed tests observe production") runs on any PR touching frontend or Rust and can be skipped with the `revert-oracle-exempt` label (`.github/workflows/test.yml:577`). Use it only with a stated reason; the whole point of the job is that text review missed 10/10 confirmed defect cases (#3661).
- **Module-size ratchet.** `rust/geometry/src/router/rtc_offset.rs` is allowlisted and sits close to its budget; a long doc comment is enough to trip `cargo test -p ifc-lite-processing --test module_size_ratchet`. Shrink, do not raise the budget.
- **Do not run repo-wide `cargo fmt` or `rustfmt` on a whole file.** Several of these files are not rustfmt-clean on main, so formatting one adds hundreds of lines of unrelated churn (`prepass.rs` alone: 313 lines). Format only the lines you wrote.

Refs #4611.

louistrue:
Closes #4799

## Summary
- resolve one owner-specific RTC context for grids, alignments, symbolic annotations, drawings, and restored room markup
- forward the exact mesh-producer frame through the overlay worker and include every frame field in cache identity
- retry pending legacy provenance only after it settles, discard stale in-flight results, and clear provenance after affine geometry rebakes
- prevent construction-profile caches from sharing a null pending key

This is the viewer consumer half of the contract merged in #4842.

## Verification
- Astra implementation review: approved with no remaining findings
- focused viewer regression suite: 20 suites, 83 tests passed
- viewer typecheck passed
- full repository lint passed: 5,245 files, 0 errors
- module-size ratchet passed; useDrawingGeneration budget lowered 1042 to 1041 and useSymbolicAnnotations 493 to 467
- git diff --check passed
- rebuilt-WASM contract evidence is carried by merged #4842

## Scope
Viewer-only consumer wiring; no published package change or changeset.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved alignment and reprojection handling so geometry, annotations, drawing data, and grid lines remain tied to the correct coordinate frame.
  - Prevented stale or out-of-order parsing results from replacing current visualizations.
  - Deferred processing while coordinate metadata is pending, then automatically retries when it becomes available.
  - Preserved coordinate-frame details accurately, including zero offsets and signed zero values.

- **New Features**
  - Added consistent coordinate-frame-aware caching and context handling across federated and standalone models.
  - Improved drawing and symbolic annotation generation for model-specific placements and translations.


louistrue:
## PR Reviewer Guide 🔍



<table>
<tr><td>⚡&nbsp;<strong>No major issues detected</strong></td></tr>
</table>

<hr>

- Tokens: 51,679 in / 40 out / 51,719 total
- Time cost: 36.8s
- AI calls: 1

</details>


Reviewed head: 24784bc8d3a4619e970e96dc88f1c517731a3a16

PR #4911, opened 2026-09-17
fix(parser): filter IfcRelAssignsToControl-assigned objects to products in cost-extractor

### LINKED ISSUE #4877: CostItemInfo.productExpressIds also collects tasks assigned via IfcRelAssignsToControl
`CostItemInfo.productExpressIds` collects anything assigned to a cost item, not just products — a task assigned via `IfcRelAssignsToControl` lands in a field named for products, with no discriminator.

Found while authoring cost data against the merged read model (#4863); the writer side is unaffected, this is a read-model naming/semantics issue.

## Where

`packages/parser/src/cost-extractor.ts` — the `IfcRelAssignsToControl` / `IfcRelSchedulesCostItems` branch:

```ts
const schedule = schedules.get(relation.RelatingControl);
const item = items.get(relation.RelatingControl);
for (const relatedId of relation.RelatedObjects ?? []) {
  const relatedItem = items.get(relatedId);
  if (schedule && relatedItem) {
    schedule.costItemGlobalIds.push(relatedItem.GlobalId ?? '');
    relatedItem.controllingScheduleGlobalIds.push(schedule.GlobalId ?? '');
  } else if (item) {
    item.productExpressIds.push(relatedId);
    item.productGlobalIds.push(store.entities?.getGlobalId?.(relatedId) ?? '');
  }
}
```

The `else if (item)` branch pushes **every** related object into `productExpressIds` / `productGlobalIds` with no check on what the related object actually is. `IfcRelAssignsToControl` legitimately binds tasks, resources and actors to a control — all of them land in the product fields.

## Why it is reachable, not theoretical

Assigning a task to a cost item is exactly an `IfcRelAssignsToControl` with the cost item as `RelatingControl`. Any file that does this — or any authoring path that offers it — produces cost items whose `productExpressIds` contain `IfcTask` ids.

## Consumer impact

Code iterating `costItem.productExpressIds` to resolve geometry, material or type for a cost breakdown receives task ids mixed in with no way to tell them apart. A product-only lookup against those ids either fails or misclassifies a task as a product. The field name is the only documentation a consumer has, and it says "product".

## Why it is easy to miss

`packages/parser/src/cost-types.ts` gives doc comments to the neighbouring fields:

```ts
/** Existing normalized quantity view; never falls back to product Qto. */
costQuantities?: CollectedQuantity[];
/** Existing nested value view retained for compatibility. */
costValues?: CostValueInfo[];
parentGlobalId?: string;
childGlobalIds: string[];
productExpressIds: number[];
productGlobalIds: string[];
```

`productExpressIds` and `productGlobalIds` carry none.

And the existing test for this exact code path — `cost-extractor.test.ts`, "resolves products assigned to a cost item via `IfcRelAssignsToControl`" — only ever assigns an `IfcWall`. It passes identically whether or not the branch filters by type, so it cannot distinguish the two behaviours.

## Which way to resolve it is your call

Both are defensible and they are genuinely different contracts:

1. **The field means "products"** — filter the branch to product-like entities, and route tasks/resources to their own fields. Changes behaviour for anyone relying on the combined view.
2. **The field means "everything assigned to this cost item"** — then the name is wrong; rename it, or at minimum document it and add a test that asserts a task *is* included, so the behaviour is pinned deliberately rather than by omission.

Either way the missing piece is the same: a test that can tell the two apart. The current one cannot.

I have not changed any of this — it is merged and the call is yours. Happy to implement whichever you prefer.

BIMvoice:
## Summary
- `CostItemInfo.productExpressIds`/`productGlobalIds` pushed every object an `IfcRelAssignsToControl` relationship assigned to a cost item, with no type check. `IfcRelAssignsToControl` legitimately binds tasks, resources and actors too, not just products, so a task assigned to a cost item landed in fields named and documented for products (#4877).
- Filters the branch to `IfcProduct`-subtype related objects using the schema-derived `getInheritanceChain` helper already used elsewhere in the parser for product/group/root classification — matching how the sibling `IfcRelAssignsToProduct` branch already only ever pushes a product.
- Adds JSDoc to both fields documenting the products-only contract.

## Which resolution and why
The issue deliberately left the call open between (1) filtering to keep the field meaning "products", or (2) renaming/documenting to mean "everything assigned". I checked the merged consumers named in the issue (#4876 cost-authoring, #4875 viewer Cost panel — `CostPanel.tsx`, `cost/CostDetail.tsx`, `cost/CostTreeView.tsx`, `cost/useCostBackend.ts`, `lib/cost/cost-tree.ts`, `sdk/adapters/cost-adapter.ts`, `mcp/tools/cost.ts`) and none of them currently read `CostItemInfo.productExpressIds`/`productGlobalIds` at all. With no maintainer ruling in the issue thread and no live consumer depending on the mixed view, I went with (1): the field name is the only documentation a consumer has, and it already says "product" — the sibling `IfcRelAssignsToProduct` branch already only ever pushes a product, so this makes the two branches consistent. Non-product related objects (tasks, resources, actors) are simply excluded; I did not add a new field to carry them since nothing in the tree currently needs one — flagging that as a possible follow-up rather than building it speculatively.

## Breaking change
Behavioural fix, not a rename: field names/types/shape are unchanged, and `scripts/api-surface.json` tracks `CostItemInfo: interface` at the export level, not individual fields, so no update was needed there. Marked as `patch` in the changeset. A downstream consumer that was relying on the previous (undocumented, type-unfiltered) behaviour will see fewer entries in these two arrays — called out explicitly in the changeset.

## Test plan
- [x] RED first: added a fixture assigning both an `IfcWall` and an `IfcTask` to the same cost item via `IfcRelAssignsToControl`, asserting the task is excluded from `productExpressIds`/`productGlobalIds`. Confirmed it fails against the unfiltered branch (`[1, 2]` instead of `[1]`).
- [x] Did not weaken the existing `resolves products assigned to a cost item via IfcRelAssignsToControl` test (product-only fixture) — extended the suite with a new, more targeted test alongside it. All 18 tests in `cost-extractor.test.ts` pass.
- [x] Mutated the guard two ways (dropped the type check entirely; neutered `isProductLike`'s return with `true || ...`) — both mutations reddened the new test on the correct assertion, confirming the guard actually executes. Reverted both; `git diff --stat` clean.
- [x] `node scripts/check-test-revert-oracle.mjs --base upstream/main --ci --json` → **OBSERVED** (reverting the two production files turns an assertion RED in `cost-extractor.test.ts`; restoration verified, tree clean).
- [x] `node scripts/check-module-size.mjs` (PLAIN form) — no findings against either touched file (`cost-extractor.ts` is 395 lines, well under budget); the run's only failure is pre-existing Rust module-size-allowlist drift unrelated to this change.

Closes #4877

## Summary by CodeRabbit

* **Bug Fixes**
  * Cost item product identifiers now include only product entities, excluding tasks, resources, and actors assigned through control relationships.
  * Existing control relationships remain available without adding non-product entities to product results.

* **Documentation**
  * Clarified that cost item product identifier fields represent products only.


louistrue:
## PR Reviewer Guide 🔍



#### (Review updated until commit https://github.com/LTplus-AG/ifc-lite/commit/87527722be3233833922aef793c81a0a4a166bb4)


<table>
<tr><td>⚡&nbsp;<strong>No major issues detected</strong></td></tr>
</table>

<hr>

- Tokens: 7,434 in / 40 out / 7,474 total
- Time cost: 7.8s
- AI calls: 1

</details>

louistrue:
**[Persistent review](https://github.com/LTplus-AG/ifc-lite/pull/4911#issuecomment-5712840845)** updated to latest commit https://github.com/LTplus-AG/ifc-lite/commit/ecc51b7b0bacbc2db00ad14bc8e94c221f6395c7

louistrue:
@BIMvoice thanks for this one. I pushed a small maintainer commit to get it over the line:

- 87527722b: `Node tests` was failing on `cost-read-model-fixture-4854.test.ts`, which still expected `productExpressIds: [10, 11]`. In that fixture #11 is an `IfcTask` assigned through `IfcRelAssignsToControl` #62, so under the products-only contract from this PR the expectation is now `[10]`, and the test pins the #62 relationship record so the task assignment is still checked.

I also grepped the viewer, SDK, CLI, MCP and sandbox for readers of `CostItemInfo.productExpressIds` / `productGlobalIds` and found none that depended on tasks being included. The viewer's cost tree reads the raw relationships, and the other hits are the schedule `TaskInfo` fields.

louistrue:
**[Persistent review](https://github.com/LTplus-AG/ifc-lite/pull/4911#issuecomment-5712840845)** updated to latest commit https://github.com/LTplus-AG/ifc-lite/commit/87527722be3233833922aef793c81a0a4a166bb4

PR #4942, opened 2026-09-18
fix(viewer): stop filing browser-extension throws as viewer bugs (#4939)

### LINKED ISSUE #4939: TypeError
undefined is not an object (evaluating 'tab.id')

[View in PostHog](https://eu.posthog.com/project/199147/error_tracking/01a0b0ea-ed05-78c0-b44a-86b17091aa92)

louistrue:
Closes #4939

## What changed

A third `before_send` sibling gate in the viewer's analytics pipeline —
`apps/viewer/src/lib/foreign-script-noise.ts`, wired into `beforeSend` next to the
existing wasm-skew and chunk-skew gates — that drops an `$exception` event when
posthog **autocaptured** it and **every frame of every entry** carries a foreign
script URL scheme (`webkit-masked-url:`, `chrome-extension:`, `moz-extension:`,
`safari-web-extension:`, `safari-extension:`).

No viewer component was changed, and no null guard was added.

## Why / root cause

**The TypeError is not ours.** The recorded occurrence (PostHog project 199147,
error-tracking issue `01a0b0ea-ed05-78c0-b44a-86b17091aa92`, session
`01a0b0ea-e733-7ed6-a7ce-beac54c156bb`, Safari 17.6 / Mac OS X, 2026-09-17
19:49:40Z, `https://www.ifclite.com/`) has exactly one stack frame:

```
TypeError: undefined is not an object (evaluating 'tab.id')
  ? @ webkit-masked-url://hidden/:21622:17     (in_app: false, mechanism.handled: false)
```

and a sibling event 12 ms later in the same session:

```
TypeError: undefined is not an object (evaluating 'response.type')
  ? @ webkit-masked-url://hidden/:4273:15
```

`webkit-masked-url://hidden/` is WebKit's placeholder for a script whose real URL
it deliberately will not disclose — a Safari web-extension content script or a
`WKUserScript` injection. `tab.id` and `response.type` are the WebExtension API's
own shapes, and the pair firing milliseconds apart from two differently-sized
foreign bundles is a content script's `browser.tabs` read followed by the reply of
a `runtime.sendMessage` round trip failing — not a viewer render.

**The viewer cannot have thrown it.** The only `tab.id` dereferences in the tree are
`RibbonToolbar.tsx` (maps the module-level `RIBBON_TABS` literal) and
`WidgetRenderer.tsx` (maps tabs already validated by `packages/extensions`' widget
schema, which type-checks `tab.id` at line 329). Neither can yield an `undefined`
element, and a throw from either would carry an `in_app` frame in our bundle under
`https://www.ifclite.com/assets/`. Neither recorded event does. So the triage
hint's guess — a null/undefined `tab.id` in our code — has no code path behind it,
and a guard there would guard a value we never read.

**What *is* ours is the telemetry gate.** `analytics-scrub.ts` already drops the
opaque cross-origin shapes ("Script error.", the ResizeObserver loop notice), but
both arms require `frameCount === 0`, because those arrive information-free.
WebKit does not report a foreign script as frameless: it reports a real frame and
masks only the URL. posthog-js has already made the same judgement one step
earlier — `@posthog/core`'s `createFrame` sets
`in_app: !!filename && !filename.startsWith('webkit-masked-url://') && …` — and
captures the event anyway. Error tracking then mints a public GitHub issue against
the viewer for code we do not ship and cannot fix. This change applies posthog's own
attribution knowledge to the whole event.

The gate is keyed on **identity, not wording** — it never reads the exception
message, so it carries none of the over-redaction hazard the scrub module's
anchored message arms exist to manage (and correspondingly needs no
`DROPPED_NOISE_SAMPLES` row, which protects message arms). Three constraints keep
the irreversible drop honest: autocaptured only (a deliberate `captureException`
is a report we asked for), at least one frame present (absence of evidence is not
evidence; the frameless shapes keep their own gates), and *every* frame foreign
(an extension that breaks the viewer through our own call stack is still ours to
see).

## Evidence

Regression test: `apps/viewer/src/lib/foreign-script-noise.test.ts` - 9 tests in
`foreign-script noise gate, through before_send (#4939)`, built from the production
`$exception_list` payload transcribed frame-for-frame (including both #4939 events).

Every assertion runs through `beforeSend` - the real `before_send` PostHog is
initialised with - rather than importing the new module directly. Two reasons: only
a test of the pipeline entry point can catch the gate being disconnected from it,
and the revert oracle reverts production and requires these tests to fail BY
ASSERTION. Because the fix ADDS a module, a test importing it directly died with
`ERR_MODULE_NOT_FOUND` under the revert (REVERT-BROKE-BUILD: the file never loaded,
so no assertion ran). `analytics.ts` is modified rather than added, so under the
revert it still loads with the gate simply absent.

**Reproduced with production reverted.**
`node scripts/check-test-revert-oracle.mjs --base origin/main` - **OBSERVED**:
`baseline: pass - 9 passed / 9 collected`, `reverted: assertion-failure - 5 passed,
4 failed / 9 collected`, tree restored clean. Re-run with
`--only apps/viewer/src/lib/analytics.ts` to attribute the tick to a single file:
**OBSERVED**, same counts - so the wiring line in `analytics.ts` is independently
observed, not carried by the new module.

Commands run on this host (Windows 11):

- `pnpm typecheck` (root turbo) — **96/96 tasks successful**, `apps/viewer` 975/975
  test files in a typecheck program.
- `pnpm --filter viewer typecheck` — clean.
- Viewer `src/lib` suite, the project's `tsx --test` runner over all
  `src/lib/*.test.ts` files — **2 fail, everything else green**. The 2 failures are
  both in `using-declaration-build.test.ts` ("using declarations survive the
  vite-plugin-top-level-await pipeline (#2130)") and are **pre-existing on clean
  main**: verified by `git stash` + re-running that file alone on the unmodified
  tree (`# tests 2 / # pass 0 / # fail 2`). Unrelated to this change.
- `node scripts/check-module-size.mjs` — OK (3079 files measured, 0 new over 400,
  `+0 ^0 v0 -0` vs merge-base). No allowlisted file grew: the new logic is a new
  module, and `analytics-scrub.ts` (allowlisted at 633, zero headroom) is untouched.
- `node scripts/check-test-glob-coverage.mjs` — OK (52 packages, 0 unrun test files),
  so the new test file is picked up by the viewer's test glob.
- `node scripts/check-changesets.mjs` — OK. No changeset: nothing under published
  `packages/*` changed, and no package export was added, removed or renamed.
- `node scripts/check-lint-ran.mjs` (oxlint over 5462 files) — no errors; `oxlint`
  on the three changed files specifically is clean.

**Not run here:** `node scripts/check-unused-locals.mjs` fails on this host before
it reaches any source — `spawnSync 'pnpm' ENOENT` from its `packageDirs()` helper,
a Windows-only shell-resolution quirk, reproduced identically under Git Bash and
PowerShell and unrelated to this branch. `tsc --noEmit` passing for `apps/viewer`
covers the same class of finding for the changed files. No Rust, cargo or perf probe
work: this change touches no `rust/`, `packages/geometry` or `packages/wasm` path.
No browser screenshot: the user-visible behaviour here is the *absence* of a
telemetry event, which the `beforeSend` wiring test is the honest instrument for.

## Review

No reviewer comments yet at the time of opening.

Two judgement calls a reviewer should check:

1. **Is dropping right, or should the condition stay queryable?** The repo has both
   precedents — #2354 kept `webgl_unavailable` visible under one fingerprint, while
   the Outlook SafeLinks crawler and "Script error." are dropped outright. This
   follows the latter: an extension's internal handshake failure carries no
   information about the viewer at all, not even a degraded capability, and the
   masked URL means we cannot even tell *which* extension. If you would rather keep
   a counter, that is a one-line change to the gate.
2. **Scheme list completeness.** Five schemes are covered. `ms-browser-extension:`
   (legacy Edge) is deliberately omitted — no recorded occurrence, and the repo's
   rule is that an arm needs positive evidence. Chromium-based Edge uses
   `chrome-extension:`, which is covered.

One thing this change does **not** do: it cannot retroactively close the sibling
`response.type` error-tracking issue if one was filed. The same gate covers it going
forward, which is part of why the rule is the right shape — one attribution test
kills both events without either message appearing in our code.

https://claude.ai/code/session_01DFsGAGjJeskfXwyXNSDGxg

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced analytics noise by suppressing unhandled errors originating exclusively from recognized browser extensions or injected scripts.
  * Continued preserving errors from the viewer application, mixed-source errors, handled exceptions, and events without valid exception details.

* **Tests**
  * Added coverage for supported browser environments, extension formats, and edge cases to ensure legitimate errors remain reportable.


louistrue:
Both review comment defects fixed in 83771ada1 (comments only, no behaviour change). One correction to finding 1, with evidence, because I did not want to write a false statement into the file:

**Finding 1 — the `in_app` quote.** Review reported the installed `@posthog/core` computes `in_app: isAppFilename(filename)` against an `APP_URL_SCHEMES` allowlist. That is not what this tree installs. There is exactly one `@posthog/core` copy in the worktree, **1.50.5** (pulled by `posthog-js@^1.426.3`, `apps/viewer/package.json:97`), and `dist/error-tracking/parsers/base.mjs` reads verbatim:

```js
const MASKED_URL_PREFIX = 'webkit-masked-url://';
const ANONYMOUS_FILENAME = '<anonymous>';
in_app: !!filename && !filename.startsWith(MASKED_URL_PREFIX) && filename !== ANONYMOUS_FILENAME
```

`isAppFilename` and `APP_URL_SCHEMES` appear nowhere under `node_modules` (grepped across every `.js`/`.mjs`, both the hoisted copy and the one nested under `posthog-js`; `.pnpm` lists only `@posthog+core@1.50.5`). The `isAppFilename` allowlist is a **later** release of the same package, so rewriting the quote to it would have made the comment wrong about the code we ship.

What was genuinely defective was the ellipsis: my quote dropped the `<anonymous>` clause. That is now verbatim and version-pinned. The allowlist form is recorded where it belongs — as one of two stated reasons the gate must never be "simplified" to read `in_app` (the other being that `in_app: false` is far broader than "foreign" even in 1.50.5: an `<anonymous>` or filename-less frame gets it too). The sentence explaining why the gate ignores `in_app` is kept and expanded, as asked.

**Finding 2 — the "Both gates are unit-tested in isolation" note.** Corrected, and it now names the failure mode rather than just the fact: the foreign-script gate is observed only through `beforeSend`, because the module is *added* by this change and therefore vanishes under the revert oracle's revert, so an isolated import would turn the regression test vacuous (REVERT-BROKE-BUILD). The comment asks the next reader not to "fix" it.

Re-verified on 83771ada1: `check-test-revert-oracle --base origin/main` → **OBSERVED** (baseline 9/9 pass; reverted 5 pass / 4 fail), `pnpm --filter viewer typecheck` clean, `check-module-size` OK (`+0 ^0 v0 -0`), `check-lint-ran` no errors, the regression file 9/9.

The three findings flagged as out-of-scope (event dropped without a queryable counter; `reverseAndStripFrames` filename inheritance; `convertToExceptionList` forcing `handled: true` on cause-chain entries) are untouched, as asked.

Issue #4897, opened 2026-09-17
Federation frame depends on load order when the first model has no RTC offset (and `federationFrameInfo` then reports the wrong frame)


uelly:
**Version:** `@ifc-lite/geometry` 7.1.0 (loading, as in `v9.1.0`) and 7.2.0 (`world-frame`); `main` @ 6a9fc13 has the same loader rule and `federationFrameInfo`

**Which surface:** Viewer (browser), plus `@ifc-lite/geometry/world-frame`

### Summary

When a model is added to a federation, the viewer takes `sharedRtcOffset` from the earliest loaded model that **has** a
`wasmRtcOffset`
([useIfcFederation.ts#L141-L151](https://github.com/LTplus-AG/ifc-lite/blob/6a9fc132731132bbbec2d9241242ae99e063a27e/apps/viewer/src/hooks/useIfcFederation.ts#L141-L151)).
A model with small coordinates has no `wasmRtcOffset`
([coordinate-handler.ts#L350-L363](https://github.com/LTplus-AG/ifc-lite/blob/6a9fc132731132bbbec2d9241242ae99e063a27e/packages/geometry/src/coordinate-handler.ts#L350-L363)),
so it does not count:

- Load **A** (small coordinates): meshed raw, no offset.
- Add **B** (large coordinates): there is no shared offset yet, so B uses its own detected anchor
  ([rtc-frame.ts#L49-L68](https://github.com/LTplus-AG/ifc-lite/blob/6a9fc132731132bbbec2d9241242ae99e063a27e/packages/geometry/src/rtc-frame.ts#L49-L68)).

A stays in raw IFC coordinates while B is RTC-shifted, and nothing re-bases A. Georeference alignment does not cover
this case, because it only runs when both models have a georeference
([useIfcLoader.ts#L555-L558](https://github.com/LTplus-AG/ifc-lite/blob/6a9fc132731132bbbec2d9241242ae99e063a27e/apps/viewer/src/hooks/useIfcLoader.ts#L555-L558)).
Loading B first gives a different, consistent scene: A then receives B's anchor as `sharedRtcOffset`.

Since #4880, `federationFrameInfo` picks the earliest-loaded model with `coordinateInfo`, whether or not it has an offset,
and documents it as the frame "the others are aligned to"
[...]

PR #4952, opened 2026-09-18
fix(viewer): distinguish an RTC convergence in placementFrameKey

### LINKED ISSUE #4936: fix(viewer): placementFrameKey doesn't distinguish an RTC convergence, so a saved pivot restores into the wrong frame
Found while merging #4906 into #4873, and verified there.

A rotation pivot is stored as a workspace **point**, so it is only meaningful in the render frame it was captured in. `placementFrameKey` is meant to stamp that frame, but for a non-georeferenced anchor it doesn't change when #4906's federation convergence re-anchors the scene: the key stays `local-engineering:m:z-up` while the frame moves by the full RTC delta.

Restoring a saved placement (manifest import or undo) after a convergence therefore rotates about a point that has shifted by the whole delta. Verified: the pivot moves by the full delta while the key is unchanged.

Translations are unaffected, since they are differences and are frame-invariant.

**Expected:** `placementFrameKey` includes whatever identifies the current RTC anchor, so a pivot saved in a different frame is detected and either re-expressed or refused.

**Done when:** a test converges the federation frame, restores a placement saved before it, and asserts the model rotates about the same world point (or that the restore is refused loudly).

louistrue:
Closes #4936

## What changed

This went through four rounds. The first three each patched a call site that cached a computed frame identity into `modelPlacement.frameKey`; this round removes the cache entirely, because per-site patches kept finding a fourth site or a fourth scenario that invalidated the fix.

`modelPlacement.frameKey` is renamed to `realignedFrameKey` and narrowed to what it always actually was for `commitRealignmentFrame`: a genuine decision ("the user explicitly re-aligned this workspace to CRS X"), never a general-purpose cache. `placementAnchor` (`persistence.ts`) deliberately reads only each model's own embedded georeference, never a committed re-alignment, so that decision cannot be reconstructed from live `state.models` and has to be remembered somewhere. Nothing else does now.

`placementFrameKey`/`placementFrameBaseKey` (`persistence.ts`) recompute from live `state.models` on every call, not a hot path (small federations, called from user actions and effects, never a render loop), so there is nothing worth memoizing. `georeferencedPlacementFrameKey` no longer embeds `wasmRtcOffset`; the live RTC anchor is appended uniformly by `placementFrameKey` as an external `:rtc:{...}` suffix, whether the base came from a live anchor or a pin, so a convergence is reflected immediately either way.

Consequences across the codebase:
- `applyModelTranslation` and `setModelRotation` (`modelPlacementSlice.ts`) no longer stamp anything into `modelPlacement`; there is nothing left to keep warm.
- `useModelPlacementPersistence.ts`'s restore effect no longer stamps anything either, and its change-detection `subscribe` compares the (now-accurate) `realignedFrameKey` field directly instead of the old general-purpose cache.
- `spatial-index.ts` / `spatial-invalidation.ts` / `placement-snapshot.ts` / `preview-analysis.ts` read the same renamed field for their own "did the placement context change" comparisons. Their own semantics (react to an explicit realignment) are unchanged; they were just reading a field that used to also mean "someone committed a move," which is no longer true.

Also fixes a real regression the previous round introduced and never caught, because its regression sweep used `-maxdepth 1` and skipped the affected subdirectories: `lib/appearance/reference-runtime/frame.ts`'s `engineeringFrame()` expected `wasmRtcOffset` embedded inside the JSON key, which the prior round's fix had already stopped producing. It now strips the external `:rtc:{...}` suffix first, uniformly, before inspecting the JSON base, restoring the existing "an RTC or origin shift alone is not a frame mismatch for an appearance reference" behavior. `frame.test.ts`, `references.test.ts`, `referenceSelection.test.tsx` and `Drawing2DCanvas.references.test.tsx` all exercise this.

## Why / root cause

A rotation pivot (`ModelPlacement.rotation.pivot`) is a workspace point, meaningful only in the render frame it was captured in, unlike a translation, which is a difference and so is frame-invariant. `convergeFederationRtcFrame` (#4897/#4906) shifts every converged model's render-frame origin, and any live placement pivot with it, by the same delta.

Each of the first three rounds fixed one place that cached a computed frame identity and could go stale: the non-georeferenced branch ignoring RTC entirely (round 1), the two `modelPlacementSlice.ts` commit actions freezing whatever was live at commit time (round 2), and the restore-on-mount hook doing the same (round 3). The fourth review round found the underlying shape of the problem: `placementFrameBaseKey` had a bare `if (state.modelPlacement.frameKey) return it` short-circuit with no dependency tracking, so ANY future write site, or any anchor change that was not a commit or a restore, could reopen the same bug. The concrete example: commit a translation on model A (caching a base keyed to A's CRS/conversion), remove A from the federation, and the remaining models anchor on a different model or on no georeference at all, but the cached base kept answering with A's frame.

Rather than finding and patching a fourth site, this round asked what `modelPlacement.frameKey` was actually FOR, and found it was two different things wearing one name: a real decision (`commitRealignmentFrame`'s pin, which cannot be recomputed) and an opportunistic cache of a cheap, always-reconstructable computation (everything else). Splitting them and deleting the cache half closes every variant of this bug at once, including ones not yet reported, because there is no longer a stamp for a future call site to forget to refresh.

Chose this over "clear the cache on every possible invalidating event" because enumerating those events is exactly the game that lost three rounds in a row; a value that is always freshly derived from its inputs cannot go stale, by construction, no enumeration required.

## Evidence

**Regression tests**, `apps/viewer/src/lib/model-placement/placement-frame-key-rtc.test.ts`, two new cases in the `#4936 review` suite covering the literal reported scenario:
- `removing the model the live key was reading from hands the key over to what remains` (the lighter, local-engineering/RTC-fallback variant).
- `removing the georeferenced anchor model hands the key over to the remaining CRS, not a stale one` (the literal variant: two real models parsed from minimal `IfcProjectedCRS`/`IfcMapConversion` STEP fixtures on two different CRS, `EPSG:2056` and `EPSG:4326`).

Checked both against the immediately preceding (round-3) commit by checking it out and rerunning just this file: the georeferenced-CRS case fails there with exactly the reported symptom (`expected` and `actual` both equal to A's stale `EPSG:2056` JSON, i.e. B's `EPSG:4326` frame never took over), the lighter variant already happened to pass under round 3 (it does not exercise the base-key cache the same way), and all 9 cases pass after this round's change. Restored the fix afterward and reran the full file to confirm no regression from the checkout round-trip.

**Revert oracle:** `node scripts/check-test-revert-oracle.mjs --base origin/main`, run myself on this worktree:
```
baseline:  pass: 45 passed / 45 collected
reverted:  assertion-failure: 26 passed, 19 failed / 45 collected
```
**Verdict: OBSERVED.** Also ran `--only apps/viewer/src/lib/model-placement/persistence.ts` alone: also OBSERVED (34 passed, 11 failed), confirming that file specifically is independently load-bearing, not just earning the tick as part of the group.

**Typecheck:** `pnpm exec turbo typecheck --filter=@ifc-lite/viewer`, passes.

**Regression sweep:** every test file directly under `apps/viewer/src/lib/model-placement/`, `apps/viewer/src/hooks/ingest/`, `apps/viewer/src/lib/appearance/` (including the `reference-runtime/` and `references/` subdirectories the previous round's sweep missed), and `apps/viewer/src/components/viewer/` (including subdirectories), all green except the one already-known, unrelated, pre-existing failure: `toolbar-parity.test.ts` (confirmed present before this branch's changes too, by stashing and rerunning; it does not reference `frameKey`/`realignedFrameKey`/`placementFrameKey` at all). `Drawing2DCanvas.references.test.tsx`, which failed intermittently in an earlier round's narrower sweep, is green now; its raster-registration path also goes through `frame.ts`'s `engineeringFrame()`, so the fix above likely resolved it rather than it being flaky.

**Module size:** `node scripts/check-module-size.mjs`, OK, 0 new files over 400 lines.

**Not run, known environment failures** (reported, not fixed): root `pnpm test` via `turbo test --filter=@ifc-lite/viewer` fails with `) was unexpected at this time.` (the viewer `test` script's bash `find`/`awk` substitution invoked under `cmd.exe` by turbo on this box); ran the actual test files directly with `tsx --test` instead. `check-changesets.mjs`/`check-unused-locals.mjs` (`spawnSync pnpm ENOENT`) not run for the same reason; changeset already present from an earlier round, still accurate.

## Review notes

Three rounds of review found three instances of the same bug class, each in a different call site. The fourth round asked for the architectural question directly: what does storing `frameKey` in state accomplish that a fresh computation would not? The answer, traced through `placementAnchor`'s deliberate "never read a committed re-alignment" comment and `commitRealignmentFrame`'s own stamping code, is that exactly one thing does (the realignment pin) and everything else was an unnecessary, unsafe cache riding along on the same field. This PR separates them and removes the unsafe half.

## Round 5 (part 1)

Triaged the round-4 bot findings against current code. This pass covers findings 2, 4 and 5; a separate pass covers the cached-frame-key and `:rtc:` suffix findings.

- **Uncommitted change verified and committed**: `placementFrameCoordinateInfo` (persistence.ts) now falls back to `federationFrameInfo` (`@ifc-lite/geometry/world-frame`) instead of an ad-hoc earliest-model scan. Evidence: `federationFrameInfo`'s signature and usage match the existing call sites in `useRenderFrameOffsets.ts` and `hooks/bcf/viewpoint-world-frame.ts`; `npx tsc --noEmit -p apps/viewer` passes.
- **Macroscope High persistence.ts:28 (legacy key drops v1.47.0 placements)**: fixed. `apps/viewer` shipped v1.47.0 (already released; the commit that introduced the embedded-`rtc` key format, `67bd8edc1`, is an ancestor of the version-packages release `bffdd89a3`) with a key that embedded the live RTC anchor inside the georeferenced base JSON's `rtc` field. This fix moved that to an external `:rtc:{...}` suffix, so `restoreWorkspacePlacements` stopped finding real placements saved under the old key. Added `legacyGeoreferencedFrameKey` and a one-time fallback lookup in `restoreWorkspacePlacements`, scoped to the georeferenced case only (its base already uniquely identifies a CRS/conversion, so the rename is lossless). Deliberately NOT extended to the local-engineering case: its pre-fix key was the bare constant `'local-engineering:m:z-up'` with no `rtc` at all, i.e. exactly the collision #4936 reports, so falling back to it would resurrect the bug under the guise of migrating away from it. Two new tests in `placement-frame-key-rtc.test.ts` cover both the successful georeferenced fallback and the deliberate non-fallback for local-engineering.
- **Macroscope High spatial-invalidation.ts:9 (rotation-only change skips spatial invalidation)**: refuted as in scope for this PR, confirmed pre-existing. `git show origin/main:apps/viewer/src/lib/model-placement/spatial-invalidation.ts` has the identical logic (only `frameKey`/translation compared, never rotation), predating this branch entirely. `placementMoved` genuinely does not compare rotation, so `useModelPlacementSync` skips spatial-index invalidation and stale measurement/clash/deviation cleanup after a rotation-only `setModelRotation` commit. Real, but not introduced by #4936 and not a frame-key caching bug; noting as a follow-up rather than fixing it in this PR.
- **CodeRabbit Minor placement-frame-key-rtc.test.ts:97 (tautological assertion)**: fixed. The case compared `placementFrameKey(localState(offset))` to itself, which would still pass for a constant or a broken implementation. Now asserts the concrete expected string `'local-engineering:m:z-up:rtc:{"x":5,"y":6,"z":7}'`.

**Tests:** `npx tsx --import ./src/test/vite-module-hooks.mjs --test --test-timeout=120000 --test-concurrency=1` over every test file under `apps/viewer/src/lib/model-placement/` and `apps/viewer/src/lib/appearance/reference-runtime/` (run from `apps/viewer`): 109 passed, 0 failed, 14 suites.

**Revert oracle:** `node scripts/check-test-revert-oracle.mjs --base origin/main`, run from the worktree root: OBSERVED (baseline 47 passed / 47 collected; reverted 25 passed, 12 failed / 37 collected), working tree restored clean afterward.

## Round 5 (part 2)

Covers the two remaining round-4 findings: the "cached frame key survives convergence" claim (Codex P1) and the `:rtc:` suffix design (Macroscope persistence.ts:92, CodeRabbit frame.ts:22).

- **Codex P1 (placement committed before RTC convergence keeps a stale stored `frameKey`)**: refuted at this head, with a regression test added anyway. There is no `modelPlacement.frameKey` to stamp: `PlacementState` (`state.ts:61`) has only the explicit `realignedFrameKey` pin, set solely by `commitRealignmentFrame`. `applyModelTranslation`/`setModelRotation` (`modelPlacementSlice.ts:99,106`) commit placements only; `rebasePlacementFrame` (`:115`) rebases pivots only; `restoreWorkspacePlacements`/`saveWorkspacePlacements` (`persistence.ts`) both call `placementFrameKey(state)` fresh, and there is no early return comparing against a stored key. New test in `placement-frame-key-rtc.test.ts`: commit a rotation via the real `setModelRotation`, save, converge (live `wasmRtcOffset` + the real `rebasePlacementFrame` action), assert `realignedFrameKey` stays `null`, the live key advances to the converged shape, the next save lands under it, and the pre-convergence manifest is refused.
- **`:rtc:` suffix (unparseable key, regex truncation, CRS-name collision)**: fixed. `placementFrameKey` now folds the live anchor into ONE parseable JSON object: `{...base, rtc}` for a JSON base (georeferenced or a pin), `{"base":"local-engineering:m:z-up","rtc":{...}}` for the bare constant. New shared helpers `placementFrameFields(key)` (parse, drop `rtc`) and `placementFrameBase(key)` (recover the exact base string) in `persistence.ts`; `engineeringFrame` (`frame.ts`) uses them instead of `/:rtc:.*$/`. Legacy fallback unchanged: `legacyGeoreferencedFrameKey` still reproduces the v1.47.0 field order (`rtc` between `originShift` and `rotation`); when a model has no `buildingRotation` the two strings even coincide byte for byte, so the fallback only fires for rotated buildings (the legacy test now sets `buildingRotation` so it exercises the real fallback path). New tests: `frame.test.ts` proves CRS names `site:rtc:A` vs `site:rtc:B` no longer collide (`frame-mismatch`, corners `null`) and that the local `{base, rtc}` wrapper survives an RTC-only rebase; the literal-shape assertion is updated to `'{"base":"local-engineering:m:z-up","rtc":{"x":5,"y":6,"z":7}}'` plus a `JSON.parse` round-trip; `referenceSelection.test.tsx` and the hook test's comment updated to the new shape.
- Note: origin/main's `engineeringFrame` already JSON-parsed and deleted `rtc`, so the new georeferenced shape is exactly what shipped code understood; the regex was introduced by this PR's earlier rounds.

**Tests:** `npx tsx --import ./src/test/vite-module-hooks.mjs --test` over `src/lib/model-placement/*.test.ts`, `src/lib/appearance/reference-runtime/*.test.ts`, `src/components/viewer/referenceSelection.test.tsx`, `src/components/viewer/useModelPlacementPersistence.test.tsx` (from `apps/viewer`): 125 passed, 0 failed.

**Revert oracle:** `node scripts/check-test-revert-oracle.mjs --base origin/main` from the worktree root: OBSERVED (baseline 51 / 51; reverted 25 passed, 13 failed). `--only apps/viewer/src/lib/appearance/reference-runtime/frame.ts` was UNOBSERVED before the second commit and is OBSERVED after it (51 / 51 baseline, 1 failed reverted); `--only .../persistence.ts` OBSERVED.

## Round 6

Two more Macroscope findings of the "frame identity changes but nothing reacts" class. Both confirmed real and fixed, one commit each.

- **Macroscope persistence.ts:236 (legacy fallback applies old pivots behind a realignment pin)**: fixed (`d36710813`). `legacyGeoreferencedFrameKey` derives from `placementAnchor(state)`, i.e. the anchor model's OWN embedded georef, which is exactly the frame `commitRealignmentFrame` re-aligned the workspace AWAY from. With `realignedFrameKey` set, the current key is the pinned CRS plus `rtc`, the legacy key never equals it, so the fallback fired and restored a pre-fix manifest (pivots in the un-pinned CRS) into the pinned frame. The condition chosen is "skip the fallback whenever `realignedFrameKey` is present": that is equivalent to "only try the legacy key when the current key IS the un-pinned georeferenced key", since the legacy key only exists for a georeferenced anchor and is already skipped when it equals the current key. New test in `placement-frame-key-rtc.test.ts` proves the same disk restores 1 placement un-pinned and 0 once pinned.
- **Macroscope useModelPlacementPersistence.ts:68 (convergence never saves under the new key)**: fixed (`258456692`). Trace: `convergeFederationRtcFrame` (`federationRtcRebase.ts:138,147`) calls the real `rebasePlacementFrame`, whose `rebasePlacementPivots` (`state.ts:246`) returns the SAME `placements` Map by identity for translation-only placements (`angle === 0` short-circuits at `:257`, `:269` returns the input map) so the slice returns `{}` and the store does not change at all; then `updateModel(id, {})` replaces `state.models`, and the guard's `state.models !== previous.models` early return (pre-existing on origin/main from #4257 `1169f4da3`, to leave model-load churn to the restore effect) swallowed the one event that mattered. The restore effect then finds nothing under the new key and only saves on `identityCompleted`, so the placement was gone on reload. Fix: the hook tracks the last observed `placementFrameKey` in a ref (`lastFrameKey`, set by the restore effect and every subscription tick) and saves whenever it moves with non-empty placements (or a pin change), keeping the `models` guard for the unchanged-key case. Comparing `placementFrameKey(state)` against `placementFrameKey(previous)` would NOT work: `convergeGeometryOntoRtcAnchor` rewrites `geometry.coordinateInfo` in place (`packages/geometry/src/rtc-rebase.ts:189`), so the previous snapshot already reads the new key. New hook test mirrors that in-place rewrite, converges via the real `rebasePlacementFrame` plus `updateModel`, asserts `placements` kept identity and the key moved, and restores the placement from `localStorage` under the new key.

**Tests:** `npx tsx --import ./src/test/vite-module-hooks.mjs --test src/lib/model-placement/*.test.ts src/components/viewer/useModelPlacementPersistence.test.tsx` (from `apps/viewer`): 115 passed, 0 failed, 14 suites. `npx tsc --noEmit -p apps/viewer`: 0 errors.

**Revert oracle:** `node scripts/check-test-revert-oracle.mjs --base origin/main` from the worktree root: OBSERVED (baseline 60 / 60; reverted 32 passed, 15 failed, load-failure because the branch's earlier rounds added exports the tests import). `--only apps/viewer/src/components/viewer/useModelPlacementPersistence.ts`: OBSERVED, assertion-failure (60 / 60 baseline; 59 passed, 1 failed reverted). `--only .../persistence.ts` against origin/main is a load-failure for the same export reason, so the round-6 hunk was verified by hand: with `persistence.ts` reverted to `09554073c` (the previous head, exports intact) `placement-frame-key-rtc.test.ts` goes 13 passed / 1 failed, the failure being the new pinned-fallback case only; tree restored clean afterward.

## Summary by CodeRabbit

* **Bug Fixes**
  * Placement data now tracks the current coordinate frame when RTC anchors converge.
  * Prevented stale placements and manifests from being reused after a frame change.
  * Manifest imports with a mismatched placement frame are now refused with a clear error message.
  * Placement restoration and preview comparisons now remain consistent across frame changes.
  * Existing placements continue to round-trip correctly when no frame change has occurred.


louistrue:
Addressed the cache-invalidation gap flagged in review. New head: `29c3e216a`.

**Approach:** split `placementFrameKey` into a cacheable `placementFrameBaseKey` (CRS/conversion/`lengthUnitScale`/`originShift`/`buildingRotation`, nothing an RTC convergence ever touches) plus the live `wasmRtcOffset`, folded on every read and never cached. `modelPlacementSlice.ts`'s two commit sites now stamp `modelPlacement.frameKey` with the base only. Chose this over clearing `frameKey` inside `rebasePlacementFrame` because the latter would also discard `commitRealignmentFrame`'s pin stability (`realignment-frame.test.ts`) any time a convergence and a realignment coexisted.

**New test:** `apps/viewer/src/lib/model-placement/placement-frame-key-rtc.test.ts`, suite `placementFrameKey does not serve a cache frozen before a convergence (#4936 review)`, 4 cases driving the real store actions (`openReposition`/`applyModelTranslation`/`rebasePlacementFrame`). Verified these fail against the first-fix-only commit (`0d7bd7c5f`) and pass after the split (checked out that commit, ran the suite, restored). Also added a case for a second commit made *between* two convergences, which isolates `modelPlacementSlice.ts`'s half of the fix.

**Revert oracle:** `node scripts/check-test-revert-oracle.mjs --base origin/main` → OBSERVED (7 passed baseline, 3 passed / 4 failed reverted). Also ran `--only` per production file: `modelPlacementSlice.ts` alone → OBSERVED; `persistence.ts` alone → `REVERT-BROKE-BUILD` (expected, the slice imports `placementFrameBaseKey`, which only exists together with the persistence.ts change; they're one logical unit, and the combined revert the hard gate requires is clean).

Minor nit (`-0`/`0` colliding under `JSON.stringify`) also fixed via a `norm` helper in the new `rtcSuffix`.

Full writeup in the PR description.

louistrue:
Fixed the third call site. New head: 4edd0bd4a.

useModelPlacementPersistence.ts line 48 (the restore-on-mount effect) now uses placementFrameBaseKey instead of placementFrameKey, matching the two modelPlacementSlice.ts sites fixed last round.

Full audit: grepped every frameKey: assignment in apps/viewer/src excluding tests, then filtered to the PlacementState field specifically (the RegisteredAppearanceReference/ScanFrame frameKey is an unrelated concept and correctly still uses the full live placementFrameKey where it appears). Four production sites write modelPlacement.frameKey:
- modelPlacementSlice.ts applyModelTranslation, uses placementFrameBaseKey.
- modelPlacementSlice.ts setModelRotation, uses placementFrameBaseKey.
- realignment-frame.ts commitRealignmentFrame, calls georeferencedPlacementFrameKey directly (base-shaped by construction, never called placementFrameKey at all).
- useModelPlacementPersistence.ts restore effect, now uses placementFrameBaseKey.

All four agree.

New test: useModelPlacementPersistence.test.tsx, "caches only the base identity on restore, even when an RTC anchor is already live (#4936)". A prior session saves under an already-live anchor's key, the current session restores while that anchor is still live, then converges to a different anchor. Confirmed it fails against the unfixed hook (stashed just that file, ran the suite, got the exact base:rtc:{X} vs base mismatch, restored) and passes with the fix.

Revert oracle, run myself this time: node scripts/check-test-revert-oracle.mjs --base origin/main, verdict OBSERVED (baseline 12/12 pass, reverted 8 pass, 4 fail). Also ran --only per production file: useModelPlacementPersistence.ts alone, OBSERVED; modelPlacementSlice.ts alone, OBSERVED; persistence.ts alone, REVERT-BROKE-BUILD (expected, the other two import its new placementFrameBaseKey export, so they are one logical unit, and the combined revert the hard gate requires is clean).

PR body rewritten with no em-dashes; this comment too.

louistrue:
Removed the cache entirely rather than patching a fourth site. New head: 9a0072f0f.

What modelPlacement.frameKey was actually for: traced it to exactly one genuine use, commitRealignmentFrame's pin. placementAnchor deliberately reads only each model's own embedded georeference and never a committed re-alignment (that is what its own comment says: "realignment stamps its frame"), so "the user explicitly re-aligned this workspace to CRS X" cannot be reconstructed from live state.models. Every other write site (the two modelPlacementSlice.ts commit actions, the useModelPlacementPersistence.ts restore effect) was not recording a decision, it was memoizing a cheap, always-reconstructable computation, and that memoization is exactly what went stale three times in three different shapes.

What changed structurally: renamed the field to realignedFrameKey and narrowed it to the pin only, written solely by commitRealignmentFrame. placementFrameBaseKey now checks that field first (a real decision, correctly persistent) and otherwise always recomputes fresh from state.models, every call, no caching. placementFrameKey appends the live RTC anchor as an external suffix uniformly on top, whether the base came from the pin or a live anchor, so a convergence is reflected immediately regardless of what was committed or restored before it. applyModelTranslation, setModelRotation, and the restore effect no longer write anything.

New test: placement-frame-key-rtc.test.ts, "removing the georeferenced anchor model hands the key over to the remaining CRS, not a stale one", the literal scenario with two real parsed IfcProjectedCRS/IfcMapConversion fixtures on EPSG:2056 and EPSG:4326. Checked out the immediately preceding commit and reran just this test: fails there with the exact reported symptom, both expected and actual equal to A's stale EPSG:2056 frame. Passes after this change. Also added a lighter local-engineering/RTC variant of the same scenario.

Also found and fixed a real regression the previous round introduced and never caught: its own regression sweep used find -maxdepth 1 and silently skipped the reference-runtime/ and references/ subdirectories, which is where lib/appearance/reference-runtime/frame.ts's engineeringFrame() lives. That function expected the live RTC anchor embedded inside the JSON frame key, which the prior round had already stopped producing (it moved to an external suffix). Fixed by stripping the suffix first, uniformly, before parsing the JSON base. Reran the full test tree without a depth limit this time; only the already-known, unrelated toolbar-parity.test.ts failure remains, confirmed pre-existing by stashing this branch's changes and rerunning it.

Revert oracle, run myself: node scripts/check-test-revert-oracle.mjs --base origin/main, verdict OBSERVED (45 baseline pass, 26 pass / 19 fail reverted). Also ran --only on persistence.ts alone: also OBSERVED (34 pass / 11 fail), so that file specifically earns the tick on its own, not just as part of the group.

Full writeup, including the older PR-body text this replaces, in the PR description.

Issue #4901, opened 2026-09-17
geometry: bound the remaining unbounded kernel loops so a hung element fails loudly instead of timing out (follow-up to #4884)


louistrue:
## Context

#4884 / PR #4888 made the viewer **survive** a geometry call that never returns: the worker is replaced, the call is re-run one job per call, and a job still silent after 90 s is skipped and reported (`complete.skippedHungElements`, and `hung_element_types` on `ifc_model_loaded`). That is a host-side escape hatch. The element is still dropped, the user still waits about 2 minutes, and consumers that don't opt in (exports, SDK) still stall.

The underlying defect is in the Rust kernel. Production showed the same models stalling at the same mesh count every time (one user at exactly `3478` meshes on 6 loads). That is a deterministic non-terminating (or effectively unbounded) computation for specific elements.

## What is bounded today

- `rust/geometry/src/kernel/budget.rs` caps exact-predicate escalations per boolean (500k) and per element (100k). The count is deliberately not wall-clock, so native and WASM outputs stay identical.
- Nothing equivalent covers the other hot loops found during #4884 triage (to be confirmed file by file):
  - sweep / swept-disk / curve and B-spline tessellation segment counts;
  - CDT refinement (`rust/geometry/src/cdt.rs`: one guard at ~1006, no visible bound at ~1192);
  - any file-driven recursion not already under the "Bounding walks over file-supplied references" rule in AGENTS.md.

## Scope (one defect class: "a single element's geometry can run unbounded")

1. Identify the element types. After PR #4888 ships, `hung_element_types` on `ifc_model_loaded` (PostHog project 199147) names the IFC keywords of skipped elements. Use it to pick the processors to audit first.
2. Extend the deterministic count budgets to those loops. On exhaustion, record a typed failure in `GeometryDiagnostics` (`failuresByReason`) and return, so the element becomes a loud, reported failure instead of a timeout.
3. Keep output byte-identical where the budget is not hit: re-pin `mesh_determinism` manifests only if a legitimate output change occurs.
4. Add a regression test per bounded loop with a constructed pathological input.

## Done when

`hung_elements_skipped` stays at zero in production for types whose loops are bounded, and the fixtures show a typed diagnostic instead of a hang.

louistrue:
Audit update after #4980: I found no remaining safe standalone unbounded loop to cap on current `main`.

[...]

Issue #472, opened 2026-03-30
SurfaceModel Geometry Support


xyzbety:
This is an IFC file exported from CATIA [2222ifc.zip](https://github.com/user-attachments/files/26337924/2222ifc.zip), which uses SurfaceModel for geometric representation. Currently, ifclite is unable to parse and recognize it. Could you please take a look at it?

<img width="2553" height="1308" alt="Image" src="https://github.com/user-attachments/assets/5d1e07ef-a267-4b7a-838a-21a1105edd82" />

<img width="2560" height="1279" alt="Image" src="https://github.com/user-attachments/assets/70a2ca58-e50c-4184-92f0-77ea1acc8ac1" />

PR #4980, opened 2026-09-18
fix(geometry): bound B-spline degree/control-points, memoize Cox-de Boor (#4901)

louistrue:
Part of #4901

## What changed

Bound `IfcBSplineCurveWithKnots.Degree` / `IfcBSplineSurfaceWithKnots.{U,V}Degree` and the total per-surface sampling work, and memoized the Cox-de Boor basis-function evaluation (`rust/geometry/src/processors/advanced_face/bspline.rs`):

- **`bspline_basis_table`** replaces the old naive-recursive `bspline_basis` call site (called once per control-point index, `O(2^degree)` per call, no caching across calls) with one bottom-up table build per sample point, reused across every index. Bit-identical to the old recursion — same recurrence, same `denom.abs() < 1e-10` guard — so memoizing a pure function cannot change legitimate output.
- **`bspline_budget.rs`** (new) adds deterministic, non-wall-clock caps: `MAX_BSPLINE_DEGREE = 64` and `MAX_BSPLINE_SURFACE_SAMPLE_WORK = 500_000_000` (bounds `(u_segments+1) * (v_segments+1) * n_u * n_v` — the actual per-surface cost driver, not raw control-point count; see the "Review" section, a flat control-point cap was tried first and reverted) plus `MAX_BSPLINE_CURVE_CONTROL_POINTS` for the 1-D curve path. A surface past a cap fails loudly (`process_bspline_face` returns `Err`, which `GeometryRouter::record_unsupported_item` already turns into a typed `GeometryDiagnostics.unsupported_items_by_type` entry — no new diagnostics plumbing needed). A curve past its cap degrades to its start vertex, the same fallback every other malformed-curve guard in that function already uses.
- Parsing helpers (`parse_control_points`, `expand_knots`, `parse_knot_vectors`, `parse_rational_weights`) split into `bspline_parse.rs`, and `surfaces_tests.rs` / `curves_tests.rs` added as sibling test files (the house `#[cfg(test)] #[path]` pattern), so `bspline.rs`, `surfaces.rs` and `curves.rs` all stay under the 400-line module-size ratchet.

## Why / root cause

`#4901` asked to audit and bound every unbounded loop on the IfcSlab → extruded-area-solid → profile → CDT → void-subtraction path. I enumerated it:

| Loop | File:line | Bound today |
|---|---|---|
| CSG exact-predicate escalations, per boolean and per element | `rust/geometry/src/kernel/budget.rs` | Already capped (500k / 100k, #1109) |
| CDT constraint-segment flip recovery | `rust/geometry/src/cdt.rs:1006` (`recover_segment`) | Already capped at 100k flips/segment |
| CDT refinement Steiner insertion | `rust/geometry/src/cdt.rs:1192` (`next_steiner`) | Already bounded: pops a monotonically-shrinking `BTreeSet` worklist, plus `MAX_REFINE_ITERS = 20_000` and `max_steiner` in `refine_to_fixpoint` |
| Composite-curve / mapped-item / trimmed-curve recursion | `rust/geometry/src/profiles/mod.rs`, `router/mapped_item.rs`, `processors/advanced_face/polyline.rs` | Already depth-capped per the "Bounding walks over file-supplied references" AGENTS.md rule (`MAX_CURVE_DEPTH`, `MAX_PROFILE_DEPTH`, `MAX_MAPPED_ITEM_DEPTH`, `MAX_BASIS_CURVE_DEPTH`) |
| Profile/steel-shape arc tessellation segment counts | `rust/geometry/src/tessellation.rs` (`profile_arc_segments`, `scale_segments`) | Already capped (`.min(...)`, quality-scaled) |
| **B-spline basis-function evaluation** | `rust/geometry/src/processors/advanced_face/bspline.rs` (`bspline_basis`) | **Was NOT bounded** — naive Cox-de Boor recursion, `O(2^degree)` per call, called once per control-point index with no memoization. A file-supplied degree in the tens (paired with a matching, equally file-suppliable knot-vector length) was already enough to make a single element's tessellation non-terminating in practice. |

That last loop is the fix in this PR. Production evidence (`hung_element_types` = `IFCSLAB`) points at extruded/void-cut slabs, but the actual defect is upstream of any slab-specific code: any element whose profile boundary or advanced-face surface resolves through an `IfcBSplineCurveWithKnots` / `IfcBSplineSurfaceWithKnots` (composite curves with NURBS segments, CATIA/Revit free-form facade or roof panels) could hit this. I could not reproduce the *specific* production model, but constructed pathological IFC fragments that exercise the exact code path (see tests below) and confirmed the old un-memoized recursion is exponential in degree; the new code closes it by construction, not by observed timing on a real file.

## Evidence

- `cargo test -p ifc-lite-geometry --lib advanced_face`: 20/20, including regression tests (pathological surface degree, pathological curve degree, pathological sample-work grid — each asserted `< 2s` and a typed failure/fallback) and one proving a legitimate degree-40 surface (exponential under the old recursion, well under the new cap) still tessellates within the same bound.
- `cargo test -p ifc-lite-geometry --test geometry_correctness_harness`: passes, byte-identical to the pinned snapshot — this is the fixture (`tests/models/issues/472_2222.ifc`, issue #472) that actually contains 145 B-spline surfaces / 503 B-spline curves, so it's the real correctness proof, not just an isolation check.
- `cargo test --workspace --exclude ifc-lite-wasm --no-fail-fast` (rerun cleanly, sequentially, after catching a concurrent-cargo-build false pass on an earlier attempt — see Review): green except the **known Windows/CRLF false-failure set** named in `BRIEF.md`/`AGENTS.md`: `census_golden::tests::the_checked_in_golden_carries_exactly_the_generated_header` (×3 binaries), `mesh_output_matches_pinned_manifest`, `symbolic_nan_sentinel_json_roundtrip::the_typescript_wire_fixture_matches_what_the_serializer_emits` — all CRLF header-drift, unrelated to this diff.
- `cargo clippy --workspace --exclude ifc-lite-wasm --all-targets -- -D warnings`: clean.
- `node scripts/check-module-size.mjs`: OK, 0 new files over 400, allowlist digest unchanged.
- `bash scripts/build-wasm.sh`: succeeds; `packages/wasm/pkg/ifc-lite.d.ts` has **no real diff** (only the pre-existing CRLF/LF working-tree warning) — `process_bspline_face`'s `Result` shape was already `Result`, so the wasm bindings are unchanged. Not committed.
- `pnpm test:wasm-contract`: 95 passed, 0 failed, 1 skipped (unrelated fixture).
- `turbo typecheck --filter=@ifc-lite/geometry --filter=@ifc-lite/wasm` and `turbo test --filter=@ifc-lite/geometry`: typecheck clean, 52 test files / 499 tests green.
- `turbo test --filter=@ifc-lite/wasm`: 9 pre-existing failures, all `ERR_UNSUPPORTED_ESM_URL_SCHEME` from `textures.test.mjs` / `type-only-geometry.test.mjs` dynamically importing an absolute Windows path instead of a `file://` URL — a pre-existing Windows-only test-harness bug in unrelated texture-loading tests (confirmed those files never touch B-spline code). Not fixed here — out of scope for #4901.
- **Perf verdict** (base `0d84727c5` vs branch, native `perf_probe`, best-of-5): AC20-FZK-Haus 14 ms both sides, byte-identical mesh output. ISSUE_129 CSG fixture 912 → 910 ms (noise), byte-identical. Neither carries a B-spline curve/surface, so that's an isolation check. The fixture that DOES (`tests/models/issues/472_2222.ifc`) is covered by the correctness harness instead: 53,479 triangles / 33,069 vertices, byte-identical base vs branch — that surface's worst B-spline patch (207x180 = 37,260 control points) measures ~23.3M against the new 500M work bound, ~20x headroom. Ledger entry (including the dead end below) added to `scripts/perf/README.md`.
- **Not run**: `node scripts/check-api-surface.mjs` (requires a full `pnpm build` across ~50 packages; no TypeScript public API surface changed — only Rust internals and existing `GeometryDiagnostics` wiring already exercised via `record_unsupported_item`), `node scripts/check-test-revert-oracle.mjs` (new tests are Rust, not covered by that TS-focused script).

## Review

Self-caught, before any external review: my first version capped `IfcBSplineSurfaceWithKnots.ControlPointsList` size directly (`MAX_BSPLINE_SURFACE_CONTROL_POINTS = 4096`), calibrated on a guess ("a few hundred control points is already a dense authored patch"). `cargo test -p ifc-lite-geometry --test geometry_correctness_harness` caught it immediately: the in-tree `tests/models/issues/472_2222.ifc` fixture legitimately carries a 207x180 (37,260-point) B-spline patch, and the flat cap silently dropped it — one tessellated face's worth of triangles vanished (53,479 → 52,327), a real output change on a legitimate model, exactly the failure mode #4901 says to avoid. Root-caused it (traced with a standalone Rust reproduction comparing the memoized table against the original recursion index-by-index — they agreed exactly, so the regression was the cap, not the math) and replaced the flat cap with a bound on the actual quantity `tessellate_bspline_surface`'s sampling loop multiplies out to (`samples * n_u * n_v`, `MAX_BSPLINE_SURFACE_SAMPLE_WORK`), which the #472 fixture clears with 20x headroom. Also separately caught, before pushing: a `cargo test --workspace` run I kicked off in the background produced a false "all green" because I started a concurrent `cargo build` (the wasm build) against the same `target/` directory while it was still running — reran it in isolation to get a trustworthy result. Both are recorded in the perf ledger / commit history rather than only here, so the next agent touching this code sees the dead end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards for unusually large or malformed B-spline geometry inputs.
  * Prevented excessive processing that could cause geometry operations to hang.
  * Over-limit curves and surfaces now fail individually while valid geometry continues to tessellate normally.
  * Improved handling of malformed knot data and irregular control-point layouts.

* **Documentation**
  * Documented B-spline processing limits and behavior for unsupported geometry.


louistrue:
Addressed the Macroscope / Codex findings in 499644393:
- Standalone `sample_curve_polyline` path (surface-of-revolution profiles) now re-checks the degree/control-point bound before its endpoint-refinement reparse, instead of treating the capped single-point fallback as "genuinely sampled".
- `process_bspline_face` now checks the sample-work bound against the RAW (undecoded) `ControlPointsList` dimensions before calling `parse_control_points`, so a hostile file no longer pays for the full decode before rejection.
- The work estimate now includes the per-axis basis-table-build cost (`degree * (n + degree)`), not just the weighted-sum term (`n_u * n_v`) — closes the "many near-empty rows" bypass where `n_v` staying small let `n_u` grow unbounded.
- `FaceBasedSurfaceModelProcessor` / `ShellBasedSurfaceModelProcessor`'s silent `Err(_) => continue` now at least traces via `diag_debug!` (these run outside `GeometryRouter`, so full structured-diagnostics wiring there is a larger follow-up, not done in this PR).

Four new regression tests added; `geometry_correctness_harness` still passes byte-identical against the real 145-B-spline-surface fixture.

louistrue:
Addressed the coordinator's round-2 follow-ups (499644393 + 847e26c18):

1. Confirmed already fixed (prior commit): the raw-dims pre-check is the ONLY control-point-count check in `process_bspline_face` now — no redundant post-decode check.
2. Curve path: added a thread-local relay (mirrors `kernel::budget`'s shape) so a capped degree/point-count curve is drained by the router once per item and reported as `IfcType::IfcBSplineCurveWithKnots` via the existing `record_unsupported_item` path — a typed diagnostic, not silent, without threading `Result` through the whole edge-loop call graph. New tests: degree 65 (one past the bound) rejected; a curve naming 1,000,000 control-point references (none of which exist in the file) rejected before any decode attempt.
3. `FaceBasedSurfaceModelProcessor` / `ShellBasedSurfaceModelProcessor` now check the drain after every `process_advanced_face` call (Ok or Err).
4. Added a revert-oracle witness in the pre-existing `processors/tests.rs` (not a new file, so the oracle's whole-file production revert doesn't remove its ability to run): a 250x250 grid at degree 2, safe to execute even under a revert (cheap un-memoized cost at that degree) while still observing the new sample-work bound flip `Ok`→`Err`. The `revert-oracle-exempt` label stays — the two new sibling test files still hit the documented structural limitation independently.

Macroscope then caught a real follow-on bug in my own fix for #2/#3: a capped curve inside an `IfcMappedItem` was invoked through `process_mapped_item_cached`'s own direct processor dispatch, which the router-level drain never saw — worse, the thread-local flag stayed set and leaked onto whichever unrelated item was processed next. Fixed by moving the drain to wrap the whole `process_representation_item` (covers the mapped-item-cached, dedup-cache-hit, and uncached paths in one place) and adding the same drain to `mapped_item.rs`'s own direct dispatch. That also surfaced that `surface_model.rs`'s per-face drain (added for point 3) was racing the router's drain for the same flag and always winning — removed it, since both surface-model processors are only ever reached through the router's processor registry (verified, no standalone call site exists). New regression test in `router/tests.rs` (pre-existing file) pins both halves: the mapped source's own diagnostic is recorded, and it is NOT inherited by an unrelated element processed afterward.

All required checks green: `cargo test -p ifc-lite-geometry` (1005/1005 lib + correctness-harness byte-identical), `clippy -D warnings` (native + wasm32), `module_size_ratchet`, full workspace test (green except the 5 named Windows/CRLF false failures), `build-wasm.sh` (.d.ts unchanged), `pnpm test:wasm-contract` (95/0/1).

PR #4992, opened 2026-09-19
fix(release): major-bump parser for EntityTable contract

louistrue:
Fixes the valid unresolved release review on #4954.

`@ifc-lite/data` 5.0.0 makes `EntityTable.getNameOrUndefined` a required member, so parser implementations expose a source-breaking structural contract. The existing parser patch/minor changesets would release that through the compatible 7.x range. This source changeset forces `@ifc-lite/parser` 8.0.0 and lets Changesets regenerate the release PR rather than hand-editing generated versions or changelogs.

Validation: `git diff --check` passes. The local changeset checker could not start because this checkout's shared dependency installation lacks `@changesets/parse`; CI is authoritative.


louistrue:
@codex review

Please focus on Changesets semantics: confirm this source changeset correctly forces `@ifc-lite/parser` 8.0.0 in the regenerated release PR without hand-editing generated release artifacts.

Issue #216, opened 2026-02-15
Basket issue when adding floor to it


IbrahimFahdah:
<img width="1195" height="686" alt="Image" src="https://github.com/user-attachments/assets/65096d22-85dd-4ffb-bf88-f1c38949812b" />

louistrue:
complete overhaul... let me know what you think 😉 @IbrahimFahdah

PR #5013, opened 2026-09-19
test(export): prove structural analysis STEP round-trip fidelity

louistrue:
Part of #4206 — the write / round-trip acceptance tranche. #4206 remains open for the explicitly listed remaining gaps below.

## Why this is the write layer

Structural analysis is currently read-only: there is no structural edit surface and therefore no edited read model for a domain serializer to consume. Same-schema IFC export already has one canonical write path for untouched data: `StepExporter` passes the original STEP records through. Adding a second structural serializer with no caller would create a drift-prone parallel write path.

This PR makes that existing production contract explicit and non-regressible against the real Constructivity structural-analysis export.

## What the proof covers

`structural_analysis_curve.ifc` is parsed, exported through the public `exportToStep(store)` API with no schema override, and reparsed. The test then proves both representations:

- every DATA entity is present under the same express id with the exact original STEP record text;
- `extractStructuralOnDemand` is deeply equal before and after export, covering the analysis model, members, connections, actions/reactions, loads, boundary conditions, load/result groups and their relationships;
- non-vacuous fixture guards pin 1 analysis model, 3 members, 4 connections, 10 activities, and more than 30 structural-domain records.

The header is deliberately outside the per-record diff because same-schema export may regenerate header framing; the IFC DATA records are the fidelity contract.

## Evidence

- Focused fixture test: 1/1 green.
- Full `@ifc-lite/export` suite: 109 files, 1,335 tests green (4 files / 37 tests skipped by their existing fixture guards).
- `@ifc-lite/export` test typecheck: 113 files, green.
- `pnpm turbo build --filter=@ifc-lite/export...`: 13/13 package builds green.
- Direct oxlint on the new file: green.
- `node scripts/check-module-size.mjs`: green, no new or raised allowlist entry.
- `git diff --check`: green.

The repository-wide lint wrapper reached its unrelated standalone-compile audit locally and reported the existing baseline set of packages that do not compile standalone in this partial worktree; the changed package's build, test typecheck, and direct lint are green. CI remains authoritative for the full repository lane.

No changeset: this is a regression proof for an existing production export contract and changes no published runtime or type surface.

## Remaining #4206 work

- source a second redistributable real-tool structural-analysis fixture;
- exercise/report IFC2X3 conversion losses;
- fill the deferred `IfcOrientedEdge`, `IfcEdgeCurve`, and `IfcStructuralSurfaceMember` geometry gaps.


louistrue:
Mutation evidence on the exact pushed head:

- Temporarily changed production `writeSourceEntityLines` to drop only `IFCSTRUCTURALANALYSISMODEL` records.
- The focused test went red: exported/reparsed entity count fell 188 → 187 and the exact record diff named missing `#216=IFCSTRUCTURALANALYSISMODEL(...)` at the new assertion.
- Reverted the mutation byte-for-byte; focused test is green again and the worktree is clean against `cba11bff8`.

This proves the fixture test observes the production STEP pass-through rather than only comparing two reader-side projections.

Issue #4885, opened 2026-09-17
Viewer never recovers from WebGPU device loss; post-loss createBuffer failures misreported as OOM


louistrue:
## Production signal (PostHog, 14 days to 2026-09-17)

- `GPU device lost (unknown): ... DXGI_ERROR_DEVICE_HUNG (0x887A0006)` (`ifc-lite:device_lost:unknown`): **17 events / 12 users** (#3 by users), Windows D3D12, many releases. Plus `A valid external Instance reference no longer exists.` (3 users).
- 6 of the losses have `ms_since_last_frame` of 6–43 **hours**: tabs left in the background, where the browser reclaimed the GPU. A reload toast is the wrong outcome there.
- One Edge user produced 5 device losses **and** 23 `ifc-lite:out_of_memory` events in 2 days: `createBuffer failed, size (672|5544|15176) is too large for the implementation when mappedAtCreation == true`, several `handled: false`, `gpu_resident_mb: 0`. Allocations that tiny are fallout from a lost or dying device, not real OOM.

## Mechanism (read from origin/main)

- `packages/renderer/src/device.ts` ~203 forwards `device.lost`. `index.ts` `handleDeviceLost` (~840) sets `deviceLost` and `render()` skips every frame. **Nothing re-requests the device or re-uploads the scene.** `apps/viewer/src/components/viewer/device-loss-report.ts` shows "Reload the page", so the viewer is dead until a reload.
- Uploads outside the guarded path keep using the device after loss, or before `lost` resolves: `Viewport.tsx` ~1246 `setSpaceOverlayMeshes`, `renderer.addMeshes` / `appendToBatches` (`index.ts` ~1179/1204). A `RangeError` there is not recognised as device loss.
- `load-errors.ts` ~152 classifies any `/createbuffer failed/` as `out_of_memory`. Unguarded throws hit PostHog uncapped on every occurrence.

## Fix plan (one defect class: "the viewer does not survive GPU device loss")

1. On a loss whose reason is not `'destroyed'`: request a new adapter/device, rebuild pipelines, and re-upload from CPU-side mesh data (the scene already retains it for eviction). Cap the attempts, then fall back to the reload toast.
2. Gate every upload on `isDeviceLost()` through one guarded helper. Treat a mapped-`createBuffer` `RangeError` as a device-loss probe, not a frame-only failure.
[...]

PR #5031, opened 2026-09-20
fix(renderer): free point-cloud assets by id and ignore duplicate loss reports

louistrue:
Follow-up to #4998 / #4885 — two bot findings (CodeRabbit, Macroscope) posted on that PR's final head after review had settled.

## What changed
- `Renderer.removePointCloudAsset` keys its stream-epoch guard by handle **id** (a `Map<number, number>` cleared whenever the epoch advances) instead of a `WeakMap` on handle identity. `usePointCloudLifecycle` rebuilds `{ id: handleId }` for cleanup, so the identity lookup missed and the GPU asset stayed resident after model removal.
- `handleDeviceLost` no longer advances `deviceLossSequence` for a duplicate report of the device already recorded as lost (late guarded-upload rejection, frame exception). Only the current device's `device.lost` signal (`fromDevicePromise`) advances it while a loss is latched, so an in-flight recovery is not aborted by its predecessor's stragglers while a real replacement loss is still observed.

## Evidence
- `packages/renderer`: `tsx --test src/*.test.ts` → 1585 pass / 0 fail (two new regressions: "removes a live point-cloud asset through a rebuilt { id } handle", "ignores a duplicate loss report from the dead device while recovery is in flight"; both fail on main).
- `turbo typecheck` renderer + viewer OK; `check-module-size` OK; changeset (patch).


BIMvoice:
Audited this post-merge at `e9c6eb9a8` because it is the follow-up that fixed the two defects found after #4998, and I wanted to check the fixes were complete rather than present. The loss-sequence half is complete. **The point-cloud half looks to me like it reintroduces the same stale-identity shape on the sibling path** — reasoned from code, not executed, with the four legs below each verified by reading.

## The epoch is stored and checked on one path, stored and ignored on the other

Same map, two readers. The append/end path compares the epoch:

```ts
private currentPointCloudStreamRenderer(handle: PointCloudAssetHandle): PointCloudRenderer {
    if (this.deviceLost || !this.pointCloudRenderer
        || this.pointCloudStreamEpochs.get(handle.id) !== this.pointCloudStreamEpoch) throw rendererDeviceLostError();
```

The removal path does not:

```ts
removePointCloudAsset(handle: PointCloudAssetHandle): void {
    if (!this.pointCloudStreamEpochs.delete(handle.id)) return;
    this.pointCloudRenderer?.removeAsset(handle);
```

It asks only "is this id present", never "is this id *mine*". The epoch value is written at `beginPointCloudStream` and then never consulted here.

## Why the id alone is not enough to tell a stale handle from a live one

- `PointCloudRenderer` allocates from a per-instance counter — `private nextHandleId = 1`, `const id = this.nextHandleId++`.
- `teardown()` does `this.pointCloudStreamEpoch++; this.pointCloudStreamEpochs.clear(); this.pointCloudRenderer?.clear(); this.pointCloudRenderer = null;` — so the next stream runs on a **new** `PointCloudRenderer`, whose counter restarts at 1.
- A stale handle is reachable from a real call path, not a contrived one: `apps/viewer/src/hooks/ingest/pointCloudIngest.ts` captures `handle` in a closure and calls `opts.renderer.removePointCloudAsset(handle)` from both `onError` and the surrounding `catch`, which fire after the async streaming work.

**Sequence:** a stream holds `{id: 1}` in epoch 0 → device loss, `teardown()` runs, epoch becomes 1, map cleared, renderer replaced → a new stream begins and is allocated `{id: 1}` in epoch 1 → the *original* stream's worker finally errors and its closure calls `removePointCloudAsset({id: 1})`. The delete finds the **new** entry, removes it, and calls `pointCloudRenderer.removeAsset({id: 1})` against the **current** renderer — destroying the live point cloud rather than no-op'ing on the dead one.

That is the same class this PR fixed, just inverted: the old bug leaked because cleanup silently missed, this one would remove the wrong asset because cleanup silently hits.

The fix looks like one line — ask the same question the streaming path already asks:

```ts
if (this.pointCloudStreamEpochs.get(handle.id) !== this.pointCloudStreamEpoch) return;
this.pointCloudStreamEpochs.delete(handle.id);
```

## Why the new tests do not catch it

`'removes a live point-cloud asset through a rebuilt { id } handle'` and the pre-existing `'stale cleanup must not remove a replacement asset with the same id'` both look like they cover this, but neither populates the map with a **live** entry at the colliding id before invoking removal with the stale handle — the replacement mock has no `beginAsset`, so the map is empty at the moment the stale call is checked. The id-collision case is the one that matters and it is the one not constructed.

**Scope, honestly:** I could not execute any of this. `renderer-device-recovery.test.ts` will not even load here — `ERR_MODULE_NOT_FOUND: Cannot find package '@ifc-lite/spatial'`, which needs a build I do not have. So this is a code trace whose four legs I each read at `e9c6eb9a8`, not an observed failure, and I have not shown the timing window is hit in practice. You can settle it faster than I can: a test that begins a stream, tears down, begins a second stream that gets the same id, then calls `removePointCloudAsset` with the first handle.

**The loss-sequence fix is complete.** `handleDeviceLost` advances the sequence while latched only under `options.fromDevicePromise`, and that flag is set at exactly one call site — inside `initializingDevice.onDeviceLost(...)`, itself gated on `this.device !== initializingDevice` — so only the current device's own `lost` promise can advance it. `reportUploadLoss` and `containFrameThrow` never pass it, so a late upload rejection or frame throw while latched is now a genuine no-op. That was the defect and it is closed.

BIMvoice:
Correcting myself: **the one-line fix I suggested above does not work.** I wrote that `removePointCloudAsset` should ask what the streaming path asks —

```ts
if (this.pointCloudStreamEpochs.get(handle.id) !== this.pointCloudStreamEpoch) return;
```

— and that is not sufficient for the collision I described in the same comment. The map is keyed by id alone, so once the replacement stream has registered under the reused id, `get(handle.id)` **is** the current epoch. A stale handle and a live one become indistinguishable at exactly the moment the check is supposed to separate them: the guard passes and the live asset is still destroyed. Confirmed by building it and watching the live asset die with that check in place.

So the epoch check alone only narrows the window; it does not close it. The finding stands, my remedy did not.

**What actually closes it is stopping the id from being reused.** `PointCloudRenderer` allocates from a per-instance `private nextHandleId = 1`, and `teardown()` replaces the instance, so the counter restarts. Carrying the outgoing instance's watermark across the teardown into its replacement means a handle id is never reissued, and a stale `{id: N}` then finds no entry at all — which the existing `Map` lookup already handles correctly.

I have that on a branch, based on `main`, with a test that constructs the collision your current tests do not: begin a stream, tear down, begin a second stream that receives the colliding id, then remove through the **first** handle and assert the live asset survives. It fails without the change and passes with it, and the existing `'removes a live point-cloud asset through a rebuilt { id } handle (#4885 review)'` still passes unchanged. I am holding it rather than offering it yet, because it currently puts `index.ts` and `point-cloud-renderer.ts` four and seven lines over their recorded budgets, and raising an allowlist entry to land a bug fix is not something I will do quietly — I would rather land it extracted than widen a ratchet.

The reason I am posting the correction before the branch is that the wrong remedy is worse than the plain finding: if you had taken the one-liner it would have read as fixed, the tests would have been green, and the collision would still be there.

Apologies for the noise on your thread. The underlying report — that removal asks "is this id present" where the streaming path asks "is this id mine" — is unchanged, and `pointCloudIngest.ts`'s `onError`/`catch` closures still supply the stale handle that reaches it.

BIMvoice:
Branch pushed: `fix-pointcloud-epoch-removal` (`275b0bb`), based on `main`, one commit, **no PR** — and it carries a module-size problem you should decide on, described at the end.

**The epoch check turned out to be unnecessary.** Having corrected myself once already, I had the redundancy tested rather than assumed: with the watermark in place, reverting the epoch check entirely leaves the collision test green. `removePointCloudAsset` is therefore **byte-identical to `main`** in this branch — `git diff upstream/main` shows no hunk for it. Once an id can never be reused, `Map.delete` returning `false` on a stale handle is already the right answer, and a second guard would have been decoration that reads like protection.

So the whole fix is: `PointCloudRenderer` takes its starting id from a new `PointCloudHandleIds` holder, and `teardown()` reads the outgoing instance's watermark before dropping it, seeding the replacement. An id is never reissued across a teardown, so a stale `{id: N}` finds no entry.

**The test constructs the collision your current ones do not.** `'stale cleanup must not remove a replacement asset with the same id'` and `'removes a live point-cloud asset through a rebuilt { id } handle'` both look like they cover this, but neither ever populates a live entry at the colliding id before removing through the stale handle — the replacement mock has no `beginAsset`, so the map is empty at the moment of the check. The new one begins a stream, tears down, begins a second stream that receives the colliding id, then removes through the **first** handle.

Verified on the pushed head rather than only in authoring: 19/19 in `renderer-device-recovery.test.ts`, and reverting just the watermark hand-off in `teardown()` fails exactly one test — the new one — with the other 18 still green.

**The problem, and it is yours to rule on.** Both touched files sit at *exactly* their recorded budgets with zero headroom, and the wiring puts them over:

```
packages/renderer/src/index.ts: 3733 lines, budget 3730
packages/renderer/src/pointcloud/point-cloud-renderer.ts: 478 lines, budget 475
```

I extracted the counter into a new `point-cloud-handle-ids.ts` (34 lines, no allowlist row) and dropped a wrapper getter, which took `point-cloud-renderer.ts` from +7 to +3 and `index.ts` from +4 to +3. What remains is three lines per file: the field holding the watermark, the line passing it into the replacement's constructor, and the line in `teardown()` reading it back. I could not remove more without deleting the rationale comments, and I did not touch `module-size-allowlist.txt` — raising an allowlist entry to land a bug fix is the ratchet-loosening I have been flagging on other PRs today, so it seemed wrong to do quietly here.

So this needs either a deliberate budget decision from you or a larger extraction from that area than a bug fix should drag in. I would rather hand you a branch that fails a gate honestly than one that passes because I widened the gate.

One thing the commit body discloses and I will repeat here: no test in this suite reaches the real `new PointCloudRenderer(...)` inside `initOnce()` — every device-recovery test replaces `initOnce` itself, as they did before this change. So mutating away the watermark argument at that one call site is not caught by anything. Pre-existing blind spot in how the suite covers point-cloud construction, not something this branch introduces or closes.

BIMvoice:
Two follow-up commits on that branch, and one of them is a correction to my own text.

**The changeset was wrong** (`45a7778`). It said *"`removePointCloudAsset` now checks the epoch the same way the streaming path already does"* — a sentence left over from the first iteration, before I had the redundancy tested and dropped the check. The shipped code leaves that method byte-identical to `main`, and the commit message beside it said so, so the release note contradicted its own commit. Rewritten to describe the watermark alone. Worth flagging rather than quietly fixing, since a changeset is user-facing release text and I would have caught this on someone else's PR.

**Partial coverage for the wiring gap** (`7ee1794`), with its limit stated rather than implied. I noted above that nothing reaches the real `new PointCloudRenderer(...)` inside `initOnce()`. The new test constructs a real `PointCloudRenderer` against a fake device exposing only the four `GPUDevice` methods `PointRenderPipeline`'s constructor touches, and asserts a distinctive `startHandleId` flows through to `handleIds.current()` and the next `allocate()` instead of restarting at 1. That closes the constructor→holder wiring, which was previously verified only by reading.

**It does not close the `initOnce()` call site**, and I checked rather than assumed: mutating that call site's argument to a literal `1` leaves the new test green. Reaching it for real needs the browser DOM — `RendererOverlays.init()` pulls in `SymbolicTextAtlas`, which wants `document` — plus a much larger WebGPU surface than this file's stubbing style covers. So the positional argument carrying the watermark is still unverified by any executing test, and I would rather say that plainly than let a test named after the wiring imply otherwise.

Suite unchanged at 1588 pass / 1 skip; `renderer-device-recovery.test.ts` 19 → 20.

The module-size question from my previous comment is unchanged and still yours — `index.ts` +3, `point-cloud-renderer.ts` +3, both from files that had zero headroom. I am looking at whether the whole point-cloud stream lifecycle can be extracted into its own module so the gate passes outright; if that turns out to need a budget raise rather than a clean extraction, I will say so instead of quietly widening the allowlist.

Issue #4918, opened 2026-09-17
Localize the remaining viewer UI strings (sweep charter, follow-up to #4785)


louistrue:
Follow-up to #4785 (closed: the localization mechanism and the translator path are complete).

## What already exists
- String catalogues with interpolation and plurals (#4810, #4831, #4860)
- One-file locale contribution: `apps/viewer/src/i18n/locales/<tag>.ts`
- Locale selection: `?lang=` (remembered), then the browser's languages, then English
- Missing or broken messages fall back to English, and a test fails on any such problem in a committed locale (#4883)
- Converted so far: MergeLayersBanner, appearance assignment list/members, section tool, ribbon toolbar (#4883)

## Charter for this sweep
**What is swept:** hardcoded user-facing strings (labels, tooltips, aria-labels, toasts, dialog text) in `apps/viewer/src/components/**`, moved into per-feature catalogues. Follow the existing pattern and add an `*.i18n.test.tsx` per converted feature.

**Size (2026-09-17):** 19 of 349 component `.tsx` files converted. A rough grep finds about 1,710 hardcoded literals in about 227 files. That is a lower bound: it misses strings built in code and the shared command lists.

**Suggested order:** one reviewable PR per slice (≲1,500 changed lines).
1. The classic `MainToolbar`
2. The shared command lists (camera, export, command palette), which also feed ribbon labels still left in English
3. The properties and hierarchy panels
4. Dialogs
5. The long tail, panel by panel

**What ends it:** a grep-based gate (a script in `scripts/`) that reports no hardcoded JSX text or `aria-label`/`title` literals under `apps/viewer/src/components`, with an explicit allowlist for IFC EXPRESS names and technical identifiers, wired into CI as a ratchet.

**Out of scope, decide separately:** shipping real translations, and an in-app language picker (only worth building once a second real locale exists).

louistrue:
Taking the next isolated sweep slice: `BulkPropertyEditor.tsx` only (plus its feature catalogue/test and literal baseline/module budget). This is coordinated to avoid the concurrent `PropertyEditor.tsx` and AddElement slices. I will use `Part of #4918`; this does not close the umbrella.

louistrue:
[...]

PR #5055, opened 2026-09-20
fix(viewer): localize BCF panels (#4918 slice N)

louistrue:
Part of #4918

## What changed

Converts the BCF collaboration panel's own chrome to the viewer's i18n catalogue system, following the established #4918 slice pattern:

- New `apps/viewer/src/i18n/catalogues/bcf.en.ts` (`bcfEn`, prefix `bcf.*`) covering all 9 in-scope files.
- Registered into `apps/viewer/src/i18n/en.ts`.
- Converted each of the 9 files to `useTranslation()` + `t('bcf.*')`:
  - `apps/viewer/src/components/viewer/bcf/BCFCreateTopicForm.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFServerConnectForm.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFServerDialog.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFTopicDetail.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFTopicList.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFViewpointCaptureButtons.tsx`
  - `apps/viewer/src/components/viewer/bcf/bcfHelpers.tsx`
  - `apps/viewer/src/components/viewer/bcf/BCFServerControl.tsx`
  - `apps/viewer/src/components/viewer/BCFPanel.tsx`
- Pluralized/interpolated states (selected-object count, comment count, the "replace N topics" warning, "Created by X on Y") are single full templated messages selected by count/condition in the component code — never assembled fragments, per the i18n README's house rules.
- Shared chrome duplicated across files (Close/Cancel/Save, the `your@email.com` author placeholder) lives under `bcf.shared.*` instead of being repeated per component.
- A handful of adjacent user-facing strings the AST gate doesn't scan (OAuth popup-blocked/timeout messages, the missing-client-id guidance, BCF import/export failure text, the file-dialog title/filters, the server-sync toast copy) were also localized for completeness, since they're immediately adjacent BCF-domain prose.
- Added 3 new `*.i18n.test.tsx` files (pseudo-locale oracle pattern): `BCFTopicDetail.i18n.test.tsx`, `BCFServerConnectForm.i18n.test.tsx`, `BCFPanel.i18n.test.tsx`.
- Added a coverage paragraph to `apps/viewer/src/i18n/README.md`.
- Regenerated `scripts/i18n-literals-baseline.json` (`--update`): the 9 rows for these files are removed (0 literals remain), no other file's count changed.

## Why

Continues the #4918 localization sweep: the BCF panel was one of the remaining uncatalogued viewer surfaces. BCF topic titles/descriptions/authors/GUIDs/comments/labels/dates remain runtime model content, and `TOPIC_TYPES`/`TOPIC_STATUSES`/`PRIORITIES` stay literal English because they are the actual `topicType`/`topicStatus`/`priority` field values this app round-trips into exported BCF files, not display-only labels — translating them would desync on-screen text from stored data.

## Evidence

- `node scripts/check-i18n-literals.mjs` before: 18/25/16/35/13/4/1/1/11 = 124 literals across the 9 files (per `scripts/i18n-literals-baseline.json`).
- `node scripts/check-i18n-literals.mjs --update` after: those 9 rows are gone from the baseline; `check-i18n-literals: OK (373 file(s) scanned, 140 with hardcoded literals, 2136 total)` on the plain run — no other file regressed or improved.
- `pnpm exec turbo typecheck --filter=@ifc-lite/viewer` — passes.
- New i18n tests + all pre-existing BCF tests (`BCFTopicDetail.author.test.tsx`, `BCFServerDialog.test.tsx` (12 cases), `BCFViewpointCaptureButtons.test.tsx`, `BCFPanel.*.test.tsx`) — all green.
- `apps/viewer/src/i18n/*.test.ts` (catalogue-problems, intlFormat, negotiate-locale) — all green.
- `node scripts/check-module-size.mjs` — `+0 added, ^0 raised, v0 lowered, -0 deleted` vs `origin/main`; pre-existing headroom notes are unrelated to this slice's files.
- `node scripts/check-test-revert-oracle.mjs --base origin/main --ci --json` — verdict `OBSERVED`, `restoration: "verified"`.

## Could not verify

- Full CI (will watch after opening).

## Summary by CodeRabbit

- **New Features**
  - Added localization support across the BCF panel, server connections, topic management, viewpoints, dialogs, statuses, errors, and accessibility labels.
  - Added English translations with support for dynamic values, pluralized messages, and runtime locale changes.

- **Tests**
  - Added coverage verifying English defaults and dynamic localization across BCF workflows.

- **Documentation**
  - Updated localization coverage documentation for BCF interface text.


BIMvoice:
Reviewed this together with #5056 and #5057, since three concurrent slices into one catalogue is where they interfere. **All three clean** — posting once rather than three times, and only because two of the facts are set-level and invisible from any single PR.

**The three baseline diffs are disjoint, so they merge in any order.** Against `main`'s `scripts/i18n-literals-baseline.json`, this PR changes 9 rows, #5056 changes 6, #5057 changes 3, and the pairwise intersections are all empty — `bcf/*`, `charts/*` and `document/*` rows respectively, no shared entry. That file is the one I would expect three simultaneous slices to collide on, and they do not.

**No key collisions anywhere.** 118 keys added here (the new `bcf.en.ts`), 92 in #5056, 71 in #5057; pairwise overlap across the three is zero, and none collides with the 3,323 keys already on `main`. No key is defined twice inside a single file either, which matters because an object literal would silently keep the last one rather than erroring.

Also checked on this PR specifically: `bcf.en.ts` is both imported and spread in `en.ts`, so its keys are actually reachable — the failure mode where a merge drops one of the two lines leaves every key in the module `undefined` at runtime with no error, and that is not present. Base is `main` for all three, with 49/47/47 check runs against 43 on a recent main commit, so none of them is in the zero-lane stacked state.

Nothing translated reaches a `data-testid`, `className`, storage key or network payload in any of the three. The one telemetry call touched, in #5057's `DocumentPanel`, sends only counts and static strings — its "Counts only — never the document's text or name" comment matches what the code does.

Interpolation placeholders match their call sites throughout. Two strings in #5057 are deliberately called with no params — `'Insert a {path} that reads the model'` and the `{IfcProject.Name}` placeholder text — and both render literally as intended: the interpolation regex only matches bare identifiers so the dotted one can never substitute, and `interpolate()` leaves an unmatched `{path}` as text rather than printing `undefined`. Worth noting only because it looks like a bug until you check the regex.

Scope: the viewer component tests need builds I do not have here, so the i18n gate output and the catalogue/baseline analysis are executed, and the component-level behaviour is read.

PR #5063, opened 2026-09-20
fix(viewer): refuse an inbound remote attribute write to a tombstoned entity

### LINKED ISSUE #5059: collab: a remote attribute write to a locally tombstoned entity is recorded, hidden, and resurrects on undo
# A remote attribute write survives a local tombstone and resurrects on undo

## Defect

The inbound remote-attribute path (`applyRemoteAttribute`, used by
collaboration to apply a peer's edits) applies a write unconditionally, even
when the target entity is **locally** tombstoned (deleted by the local user,
not yet synced/acked). The inbound **create** path already guards this exact
case; the attribute path does not. The write is recorded permanently in the
mutation history, only *hidden* from the derived change list while the
tombstone stands — so undoing the local delete (`restoreFromTombstone`)
makes the peer's write effective again, in a slot the local user never
touched.

### Sequence a user would hit

1. User A deletes an entity locally (tombstoned, not yet removed from the
   room's path registry — the code deliberately keeps that registration so
   outbound tombstoning can resolve it).
2. Before that delete propagates, user B's client sends an attribute write
   to that same entity (via its stable room path). `applyRemoteAttribute`
   applies it unconditionally: `MutablePropertyView.setPositionalAttribute`
   validates only that `index` is a non-negative integer — nothing about the
   entity's tombstone state — and records the write in
   `positionalAttrMutations` and `mutationHistory`.
3. `packages/mutations/src/effective-changes.ts` filters the resulting
   change out of the *derived* change list at read time
   (`snapshot.tombstones.has(c.entityId)`), so nothing appears to have
   happened while the entity stays deleted.
4. User A undoes the delete. `restoreFromTombstone` is a bare
   `this.tombstones.delete(expressId)` — it does not purge the recorded
   write. The tombstone is gone, so the read-time filter above no longer
   excludes the change: the restored entity now carries user B's value in a
   field user A never edited and has no record of accepting.

### Where (at `776bd0cae98cbe935ddf6d11d5acd62155f9fed4`, `upstream/main`)

- The inbound **create** path already guards this:
  `apps/viewer/src/lib/collab/remote-entity-create.ts:22-23`:
  ```ts
  const currentOwner = entityForPath(store, entityPath);
  if (currentOwner !== null) {
    if (!view.isDeleted(currentOwner)) return false;
    // A local delete keeps the path registration so outbound tombstoning can
    // resolve it. A later peer recreation of the same room path is a new live
    // entity; release the dead owner before rebinding the reverse map.
    unregisterEntityPath(store, currentOwner);
  }
  ```
  (the comment about "keeps the path registration" is exact — quoted from
  lines 24-26).
- The inbound **attribute** path has no equivalent check. Reference
  resolution in `apps/viewer/src/lib/collab/entity-reference-wire.ts:153`
  rejects only an unresolved path, not a tombstoned one:
  ```ts
  if (id === null) return { ok: false, reason: `unresolved room reference: ${path}` };
  ```
  and `entityForPath` still resolves a tombstoned entity's id (path
  registration is retained per the comment above), so a reference to a
  now-deleted entity resolves successfully.
- `MutablePropertyView.setPositionalAttribute`
  (`packages/mutations/src/mutable-property-view.ts:1030-1037`) validates
  only the index:
  ```ts
  if (!Number.isInteger(index) || index < 0) {
    throw new Error(`setPositionalAttribute: index must be a non-negative integer, got ${index}`);
  }
  ```
  — nothing checks `entityId` against `this.tombstones`.
- The crux — a read-time filter, not a purge —
  `packages/mutations/src/effective-changes.ts:360`:
  ```ts
  if (c.kind !== 'entity-deleted' && snapshot.tombstones.has(c.entityId)) return false;
  ```
  This runs inside `collectEffectiveChanges`, over the list `changes` already
  produced by `collectAttributeAndTypeChanges` etc. — it drops the record
  from what callers *see*, it does not touch
  `MutablePropertyView`'s underlying `positionalAttrMutations`/
  `mutationHistory` maps at all.
- `restoreFromTombstone` (`packages/mutations/src/mutable-property-view.ts:1303-1305`):
  ```ts
  restoreFromTombstone(expressId: number): boolean {
    return this.tombstones.delete(expressId);
  }
  ```
  Confirmed: it touches only the `tombstones` set. Once it runs, the
  `effective-changes.ts` filter above no longer suppresses the earlier
  write, and it becomes visible/exported again.

## What was verified (CONFIRMED — read/ran at `upstream/main` 776bd0cae98cbe935ddf6d11d5acd62155f9fed4, and at the fix branch tip below)

- CONFIRMED (read): `remote-entity-create.ts:22-28`, exact guard and comment
  as quoted.
- CONFIRMED (read): `entity-reference-wire.ts:153`, exact `id === null`
  check and no tombstone check anywhere else in `decodeRoomAttributeValue`.
- CONFIRMED (read): `mutable-property-view.ts:1030-1037`, exact index-only
  validation in `setPositionalAttribute`.
- CONFIRMED (read): `effective-changes.ts:358-362`, exact filter predicate
  including the `c.kind !== 'entity-deleted'` guard clause and that it runs
  over the already-collected `changes` array (a read-time projection), never
  reaching back into `positionalAttrMutations`.
- CONFIRMED (read): `restoreFromTombstone`
  (`mutable-property-view.ts:1303-1305`) is exactly `this.tombstones.delete(expressId)`
  with no interaction with `positionalAttrMutations`/`mutationHistory`.
- CONFIRMED (ran): on branch `fix-tombstoned-attribute-writes` @ `39d349aae`,
  `apps/viewer/src/lib/collab/mutation-bridge.tombstone.test.ts` run directly
  via `npx tsx --test` (the file uses `node:test`, not vitest, specifically
  to avoid `@ifc-lite/export`'s wasm-dependent import chain) —
  **4 passed, 0 failed**: refuses a write to a tombstoned entity; a
  different id's tombstone/restore does not leak into the guard; a live
  entity's write still applies; and a tombstone → write → restore cycle no
  longer resurrects the write.
- CONFIRMED (read): the fix itself
  (`apps/viewer/src/lib/collab/mutation-bridge.ts`, one line) adds
  `if (view.isDeleted(entityId)) return \`entity ${entityId} is locally deleted\`;`
  at the top of `applyRemoteAttribute`, reusing the existing
  `onAttribute -> rejectRemoteAttribute -> collabGeometryNotice` rejection
  plumbing already used for other `applyRemoteAttribute` rejections (e.g. an
  unresolved room reference) — no changes needed to that plumbing.
- CONFIRMED (read): the fix's commit message explicitly states
  `MutablePropertyView.setPositionalAttribute` itself was deliberately left
  unchanged, because `StoreEditor`/authoring paths and undo/redo replay of
  `UPDATE_POSITIONAL_ATTRIBUTE` mutations (`mutationSlice.ts`) call it
  unconditionally on ids that can be mid-tombstone during replay and do not
  expect a blanket refusal there — i.e., the fix is scoped to the
  collaboration ingress point, not the shared mutation primitive.

## Not claimed

- Not claimed: an end-to-end reproduction through two live collaborating
  clients — the sequence above is reconstructed from tracing the guard,
  the filter, and the undo path in isolation; the branch's own test
  exercises `applyRemoteAttribute` directly, not a full two-peer collab
  session.
- Not claimed: any UI-visible symptom was directly observed (e.g. a
  property-panel screenshot showing the resurrected value) — this is a
  data-layer defect traced through the mutation/effective-changes code, not
  something driven through the viewer UI in this session.
- Not claimed: exhaustive coverage of every inbound collab write path
  (e.g. named/root attribute writes vs. positional) — only the positional
  attribute path (`setPositionalAttribute`) and `applyRemoteAttribute`'s
  entry point were traced; a parallel named-attribute path was not
  independently re-verified for the same gap in this session.
- Not claimed: severity/frequency under real collaboration latency —
  no measurement of how often a peer write races a local delete in
  practice.

## Branch

`fix-tombstoned-attribute-writes` @ `39d349aae988a1859c1e891785e08e2221f0abef`,
based on `upstream/main`, merges clean (`git merge-tree` against current
`upstream/main` produced no conflict output). Single commit, two files
changed (`mutation-bridge.ts` +16 lines including the doc comment;
new companion test file `mutation-bridge.tombstone.test.ts`). Test run
confirmed 4/4 passing as above.

Offering to open a PR from this branch, or to bin it — your call.

BIMvoice:
Closes #5059.

The inbound remote-attribute path applied a write to a locally tombstoned entity. `effective-changes.ts` hides such a record at read time via `snapshot.tombstones.has(c.entityId)`, but that is a suppression over the derived change list, not a purge — the entry stays in `positionalAttrMutations` and history, so `restoreFromTombstone` makes it effective. Undoing a local delete brought the entity back carrying a peer's value in a slot the local user never edited.

One line, mirroring the guard the inbound **create** path already has:

```ts
if (view.isDeleted(entityId)) return `entity ${entityId} is locally deleted`;
```

at the top of `applyRemoteAttribute`, returning through the existing rejection-string convention that `collabSlice`'s `onAttribute` already forwards to `collabGeometryNotice` — the same path the "unresolved room reference" rejection uses. No change was needed in `collabSlice.ts`.

**Not in `MutablePropertyView.setPositionalAttribute`**, deliberately: undo/redo replay of `UPDATE_POSITIONAL_ATTRIBUTE` in `mutationSlice.ts` calls that primitive unconditionally and is not gated on `isDeleted`, so a blanket refusal there would risk breaking replay. The collab inbound path is the only route that should refuse.

**It fires only where intended.** `applyRemoteAttribute` has three call sites: `collabSlice.ts`'s inbound handler, and two in `remote-entity-create.ts` — one inside `if (sourceExpressId !== undefined && !view.isDeleted(sourceExpressId))`, the other on a freshly `addEntity`-created id. Neither can pass a tombstoned entity, so the create path is untouched.

## Verification

`mutation-bridge.tombstone.test.ts` (4 tests, wasm-free so it runs without the geometry chain): RED before the fix was the resurrection itself — `+ Map(1) { 3 => 'peer value' }` against `- null`. With the guard removed again, 3 of the 4 fail and the one that still passes is the live-entity control, which is the shape you want; restored, 4/4.

An adversarial pass over the patch separately confirmed: `isDeleted` is exactly `tombstones.has(id)` with no false positive for an id that never existed; the delete-then-recreate path is unaffected because `createRemoteOverlayEntity` unregisters the dead owner and rebinds before any attribute is applied; refusing changes nothing about what this peer publishes, since `onAttribute` never calls `mirrorAttributeEdit` and the observer skips local transactions; and mutating the *write* rather than the guard fails only the live-entity control, so that control is meaningful rather than vacuous.

## One decision worth your eye

Refusing drops a peer's concurrent write locally rather than queueing it, so if the delete is later undone that edit is gone. The alternative — hold it and replay on restore — converges differently. I went with refuse-and-notify because it matches what the create path does with a live occupant and what the wire layer does with an unresolved reference, so the posture is consistent rather than novel. If you would rather it queue, this guard is the wrong shape and I would bin it.

Scope: `mutation-bridge.test.ts` itself pulls in `StepExporter` → `@ifc-lite/geometry` → the wasm bridge, which I could not build, so the new test is a wasm-free companion exercising `applyRemoteAttribute` directly rather than a case added to that file. No changeset — `apps/viewer` is private.

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote attribute updates are no longer applied to entities that were deleted locally.
  * Prevented stale remote changes from resurfacing when a locally deleted entity is restored.
  * Remote updates continue to apply correctly to active entities.
* **Tests**
  * Added coverage for locally deleted entities, restoration scenarios, and valid updates to active entities.


louistrue (approved):
Reviewed the inbound mutation path, the tombstone lifecycle, and the focused test. Rejecting only remote writes to the tombstoned entity prevents a write from surviving into local delete undo, while preserving live writes and local/replay paths. Approved pending required CI.

louistrue (changes requested):
Required CI failure: `apps/viewer/src/lib/collab/mutation-bridge.ts` is now 416 lines, exceeding the 400-line module-size limit. Please move or condense the newly added 16-line implementation rationale (the PR description already preserves this context) so this module returns to ≤400 lines, then push and let CI rerun. The behavior and regression coverage otherwise look correct.

BIMvoice:
Fixed in `3c13aa4` — `mutation-bridge.ts` is now 399 lines and the ratchet is clear. But the arithmetic is worth flagging, because your suggested remedy could not have worked on its own.

**The file was already at exactly 400 before this PR.** `git show 531246a1a:apps/viewer/src/lib/collab/mutation-bridge.ts | wc -l` → `400`. So the one-line guard alone takes it to 401, and condensing only the rationale I added floors at 401 too. Something pre-existing had to give, or the file had to be split.

What I did, rather than the allowlist row you offered: moved the rationale into `mutation-bridge.tombstone.test.ts` as a module doc comment — it is not budgeted, and the reasoning is arguably more useful next to the tests that demonstrate it — left a two-line pointer above the guard, and tightened two **pre-existing** paragraphs for wording only (the `'$'` sentinel one 7→4 lines, the `removeAttributeMutation` one 3→2). No facts dropped from either. Flagging that last part explicitly since it goes beyond the lines this PR introduced.

I did not add a row to `scripts/module-size-allowlist.txt`. You offered it with a written justification, but widening a budget to land a bug fix is the thing I have been reporting on other PRs today, and it seemed wrong to take the exemption for my own.

**The guard is byte-identical** — `if (view.isDeleted(entityId)) return \`entity ${entityId} is locally deleted\`;` — and I re-ran the mutation check after the edit specifically because comment surgery near a one-line guard is how a guard gets clipped: deleting it fails 3 of the 4 tests with the live-entity control still passing, restoring it returns 4/4.

`node scripts/check-module-size.mjs --base upstream/main` now reports no "New source file(s) over 400 lines" block, no "grew PAST" block, and `+0 added, ^0 raised, v0 lowered, -0 deleted` on the allowlist.

Worth knowing for whatever lands in this file next: it is at 399 of 400, so the next line added to `mutation-bridge.ts` fails the gate again.

louistrue (approved):
Re-reviewed the corrective commit. It restores `mutation-bridge.ts` to the module-size limit without changing the tombstone guard, and keeps the behavioral rationale adjacent to the regression coverage. Approved pending the fresh CI run.

PR #5064, opened 2026-09-20
fix(viewer): localize ClashPanel (#4918 slice N)

louistrue:
Part of #4918

## What changed

Converted every hardcoded UI string in `apps/viewer/src/components/viewer/ClashPanel.tsx` to the i18n catalogue, following the pattern established by prior #4918 slices:

- Added `apps/viewer/src/i18n/catalogues/clash-panel.en.ts` (`clashPanelEn`), covering the header/help disclosure, detection controls (mode/tol/gap, run buttons, live progress), the result-summary toolbar (group-by/sort, review-status filters, on-select focus mode, bulk actions), the on-demand intersection-solid status line, the user's own exclusion list, every empty/no-match/no-comparison state, and the per-row exclusion and review-comment controls.
- Registered the catalogue in `apps/viewer/src/i18n/en.ts`.
- Converted `ClashPanel.tsx`: `useTranslation()` in the main component and in the module-level `ClashReviewControls`/`ExcludeAnyButton`/`ClashExclusionActions` components (each is a real function component, so each calls its own hook rather than threading `t` through props). The `SEVERITY`/`REVIEW_STATUS`/`SORT_LABEL`/`EXCLUSION_KIND` lookup tables moved to the `labelKey` pattern already used elsewhere in the sweep (`sectionConstants.ts`'s `AXIS_INFO`, slice 2's command registries).
- Added `ClashPanel.i18n.test.tsx`: a pseudo-locale oracle exercising empty/running/populated-result/no-match/intersection-solid states, plus interpolation/plural assertions for every templated key.
- Documented the catalogue's scope in `i18n/README.md`, including the deliberate exclusion of `describeClash()`'s plain-language finding description (reused verbatim as a BCF topic's persisted description — exported content, not view chrome).
- Regenerated `scripts/i18n-literals-baseline.json`.

## Why

Slice of the ongoing #4918 sweep to localize the remaining viewer UI strings. `ClashPanel.tsx` was the largest remaining file per the baseline (113 literals).

## Evidence

- `node scripts/check-i18n-literals.mjs` (before): `ClashPanel.tsx: 113`
- `node scripts/check-i18n-literals.mjs --update` (after): `ClashPanel.tsx` row removed entirely (0 remaining literals); `check-i18n-literals: OK (373 file(s) scanned, 148 with hardcoded literals, 2147 total)`
- `export PATH=$(echo "$PATH" | tr ':' '\n' | grep -vi 'Roaming/npm' | paste -sd:) && pnpm exec turbo typecheck --filter=@ifc-lite/viewer` — passes
- `npx tsx --import ./src/test/vite-module-hooks.mjs --test src/components/viewer/ClashPanel.i18n.test.tsx` — 16/16 pass
- `npx tsx --import ./src/test/vite-module-hooks.mjs --test src/i18n/catalogue-problems.test.ts src/i18n/useTranslation.test.tsx src/i18n/negotiate-locale.test.ts src/i18n/locale-activation.test.tsx` — 29/29 pass (no locale-fallback regressions)
- `node scripts/check-module-size.mjs` — `ClashPanel.tsx` stayed at exactly its existing 1303-line budget (compacted a few multi-line `t()` calls to stay under it; no budget was raised)
- `node scripts/check-test-wiring.mjs` — OK
- `node scripts/check-test-revert-oracle.mjs --base origin/main --ci --json`: reports `INCONCLUSIVE` (full revert deletes the brand-new catalogue file, so the test's guarded dynamic import cleanly `describe.skip`s — same shape as `Measure.i18n.test.tsx`'s own documented case). Scoping with `--only apps/viewer/src/components/viewer/ClashPanel.tsx` hits a Windows-only `git apply` CRLF failure reverse-applying the patch (known local false-failure class for this gate on Windows). Manually reverting `ClashPanel.tsx` to `origin/main`'s copy and re-running the suite directly confirms the real question the oracle asks: the test goes genuinely RED by assertion (`clashPanel.title: "Clash detection" must be translated, marked text not found`, etc.) across every affected test, not a load failure.

## Not verified

- CI's actual run of `check-test-revert-oracle` (Linux) — expected to report OBSERVED given the manual verification above, but not confirmed in this environment.

## Summary by CodeRabbit

- **New Features**
  - Added English localization for the Clash Detection panel, including controls, tooltips, help text, statuses, filters, grouping, exclusions, review actions, and empty/result states.
  - Added localized pluralized and interpolated messages for counts, tags, distances, volumes, statuses, and rule names.
  - Clash Detection labels now update when the active locale changes.

- **Tests**
  - Added comprehensive coverage for localized panel states, including running, filtered, excluded, review, intersection, and clean results.

- **Documentation**
  - Documented Clash Detection translation coverage and clarified which runtime model content remains untranslated.


BIMvoice:
`Lint` is red on `b83749c2`, and the cause is a baseline moved one step ahead of the code rather than anything wrong with the localization.

From the job log:

```
apps/viewer: 246 (baseline 245, +1)
apps/viewer-embed: 248 (baseline 247, +1)
node scripts/check-unused-locals.mjs --update   (and commit the baseline)
```

The PR lowers `scripts/unused-locals-baseline.json` from `main`'s `246`/`248` to `245`/`247`, but the measured counts on this head are still `246`/`248`. So the ratchet is comparing a real count against a baseline that was tightened before the corresponding local actually went away — a `+1` against the PR's own new floor.

Worth noting it is the mirror of what #4999 hit earlier today: there the sweep removed unused locals and the baseline was left *too loose*, so the gate failed for slack. Same gate, opposite direction, same one-command fix — `node scripts/check-unused-locals.mjs --update` on this head and commit it.

**One consequence that is easy to miss:** `lint` is an `&&` chain —

```
check-changesets && check-test-glob-coverage && check-unused-locals && check-i18n-literals && check-lint-ran
```

— so `check-i18n-literals` and `check-lint-ran` **never executed on this SHA**. The PR body's "0 hardcoded literals" is therefore a local result, not something CI has verified here. It may well be right; it just has not been checked by the lane that would catch it. Both will run once the baseline lands.

## Otherwise clean

Base is `main` with 46 distinct check-run names, in line with a normal main-based head — not one of the stacked-CI cases I flagged on #5069/#5070.

The `SEVERITY` / `REVIEW_STATUS` / `SORT_LABEL_KEY` / `EXCLUSION_KIND_LABEL_KEY` move to the `labelKey` pattern is consistent — every usage I grepped goes through `t(...)` on the key, with no stale string-literal lookup left behind. `clashPanelEn` has a single import and spread in `en.ts`, and its `clashPanel.*` namespace does not collide with #5062's `chat.*`/`chatByok.*` keys, so the two can land in either order.

`ClashPanel.tsx` is 1303 lines at this head, matching its unchanged allowlist row — verified by line count rather than from the PR body.

Scope: the viewer component tests here need builds I do not have, so `ClashPanel.i18n.test.tsx` is unrun on my side; the failing-gate analysis above is from CI's own log, and the baseline comparison is measured against `upstream/main`.

BIMvoice:
Stale on arrival — you fixed this at `4a37e2b` (13:59:25) and my comment landed 22 seconds later, still reading `b83749c2`. The baseline is back to `246`/`248` matching `main`, so the `+1` is gone. Sorry for the noise.

The one part that outlives it: because `lint` is an `&&` chain, `check-i18n-literals` and `check-lint-ran` never executed on the failing SHA, so the "0 hardcoded literals" line in the body had not been CI-verified at that point. It will be on this head once the lane finishes — worth a glance when it does, since that gate is the whole point of the slice.

PR #5065, opened 2026-09-20
test(geometry): magnitude-gate alt once a census host is already diverged

### LINKED ISSUE #5060: census golden: an alt regression on an already-diverged host cannot fail the gate
## Title
`diverged()` is boolean, so `alt` can regress on an already-diverged host with no gate firing

## Summary

The watertightness census golden compares the alternate-triangulator open-edge
count (`alt`) only through a boolean, `HostRow::diverged()`. Once a host's
`alt` already disagrees with its `open` count, `alt` is free to move to any
other value — including a much larger one — without the census flagging
anything, because the comparison only asks whether the *disagree/agree*
state changed, never by how much.

This is the identical shape to the known, already-documented `coll`
limitation on the same golden — but for `coll` the author left a comment
explaining the trade-off and its cost. No equivalent rationale exists for
`alt`. This issue is about `alt` only; it does not propose touching `coll`.

All line numbers are against `rust/geometry/tests/census_golden/mod.rs` at
`upstream/main` SHA `776bd0cae98cbe935ddf6d11d5acd62155f9fed4`.

## The defect

**CONFIRMED.** `diverged()`:

```rust
// mod.rs:201-204
pub fn diverged(&self) -> bool {
    self.alt != Some(self.open)
}
```

**CONFIRMED.** Every use of `.alt` and of `diverged()` in the file (full
grep, `rust/geometry/tests/census_golden/mod.rs`):

- `mod.rs:124` — field declaration, `pub alt: Option<usize>`.
- `mod.rs:203` — `diverged()`'s body, quoted above.
- `mod.rs:230` — doc comment noting `alt` and `pre` are signed readings.
- `mod.rs:330` — `t.non_invariant += usize::from(r.diverged());` (a summary
  counter, not a gate).
- `mod.rs:863-866` — the ONLY place `alt` (via `diverged()`) affects
  pass/fail, quoted in full below.
- `mod.rs:872, 919` — comments referencing `alt` as one of the fields a
  reclassification holds constant.
- `mod.rs:1017-1033, 1064, 1079-1081, 1189, 1220` — TSV header, the
  `--features triangulation-alt` doc line, serialization/deserialization of
  the column. No comparison logic.
- `mod.rs:1502, 1732-1739, 1803-1810, 1838, 1875` — test fixtures/comments,
  not production comparison logic.

The `reclassifications` function (`mod.rs:535`–~600) and `classify`'s other
branches never touch `.alt`. The pass/fail-relevant block is exactly:

```rust
// mod.rs:863-866
if r.diverged() && !g.diverged() {
    c.worse_counts.push("newly depends on the triangulator's diagonal choice".to_string());
} else if !r.diverged() && g.diverged() {
    c.better.push("no longer depends on the triangulator's diagonal choice".to_string());
}
```

If both sides are already diverged (`r.diverged() == g.diverged() == true`),
neither branch fires — regardless of how far `r.alt` moved from `g.alt`, or
from `r.open`.

**CONFIRMED — the asymmetry.** `open` and `strict`, the two other per-edge
counts on the same row, ARE magnitude-compared, not boolean-compared:

```rust
// mod.rs:720-723
if r.open > g.open {
    c.worse_counts.push(format!("open edges {} -> {}", g.open, r.open));
} else if r.open < g.open {
    c.better.push(format!("open edges {} -> {} ({})", g.open, r.open, fall_note(g, r)));
}
```

```rust
// mod.rs:738-745
if r.strict > g.strict {
    c.worse_counts.push(format!("strict directed-pair edges {} -> {}", g.strict, r.strict));
} else if r.strict < g.strict {
    c.better.push(format!(
        "strict directed-pair edges {} -> {} ({})",
        g.strict, r.strict, fall_note(g, r)
    ));
}
```

`alt` carries a count exactly like `open` and `strict` (`mod.rs:122-124`:
"Open boundary edges under the ALTERNATE triangulator"), but is the only one
of the three reduced to a boolean before comparison.

## The live instance: `various/rvt01.ifc #31323`

**CONFIRMED.** PR #5040 ("test(geometry): bless the census golden for
#5033's closed hosts; pin the rvt01 #10191 replay honestly") merged as
`upstream/main`'s current tip, `776bd0cae98cbe935ddf6d11d5acd62155f9fed4`
(single-parent commit on `main`, not a two-parent merge — verified via
`git log -1 776bd0cae...^1`, which resolves; `^2` does not exist). Its
parent is `47f39b850` ("feat(viewer): load LandXML 1.2 TIN surfaces
(#5010)").

Row for `various/rvt01.ifc` id `31323`, `rust/geometry/tests/manifests/watertightness_census.tsv`
(columns per the file's own header: `model id rep open tris coll far alt pre strict vol`):

Before (`47f39b850`):
```
various/rvt01.ifc	31323	Tessellation	8	198	0	0	6	0	8	-
```

After (`776bd0cae9`, current `upstream/main`):
```
various/rvt01.ifc	31323	Tessellation	8	198	0	0	7	0	8	-
```

`alt` moved 6 → 7 (worse: further from `open`), `open` held at 8. On both
sides `diverged()` is `true` (`6 != 8` and `7 != 8`), so `mod.rs:863-866`
took neither branch either before or after — the regression was blessed
with no gate firing, by construction, not by a decision anyone made about
this specific row.

## Scale

**CONFIRMED**, re-derived independently: counting non-comment, non-header
data rows in `rust/geometry/tests/manifests/watertightness_census.tsv` at
`upstream/main` where `alt != open` (excluding the header row, which the
naive "non-`#`-line" count includes):

- Total data rows: 1170.
- Rows with `alt != open` (frozen against further `alt` drift, exactly like
  the `rvt01 #31323` row above): **37**.
- Rows with `alt == "x"` (alternate pass failed to run at all): 0.

This matches the number given in the task (37) exactly.

## Fair to the author: `coll` has the same shape, but it's a documented, costed trade-off

**CONFIRMED.** `mod.rs:793-805` (inside the `open`-comparison commentary):

> A COLLAPSE is the fourth way the count can be unreal. It is guarded inside
> `open_is_comparable`, which the two clauses above call, and deliberately
> NOT inside `open_is_a_defect_count`, because that predicate also defines
> `is_torn_solid`, which the derived totals and `worse_gated` read and which
> has its own pinned test. `edge_stats` SKIPS degenerate triangles, so every
> one of them removes edges from `open`, and `HostRow.collapsed` is a
> BOOLEAN: on a host already at `coll=1`, going from 10 degenerate triangles
> to 200 moves no axis at all, and the entire open drop can be collapse
> rather than repair. 44 golden rows would otherwise be eligible and carry
> `coll=1`, so folding this into `open_is_comparable` takes the eligible
> population from 130 to 86. Carrying the degenerate COUNT instead of a
> flag would let the axis fire on an increase and is the better fix, but it
> changes the golden's schema and so needs its own re-bless.

This is exactly `alt`'s shape (a boolean standing in for a count, hiding a
worsening once already tripped) — but here the author named the cost (44/130
rows affected), named the correct fix (carry the count), and named why it's
deferred (schema change, needs its own re-bless pass).

**No equivalent comment exists for `alt` or `diverged()`.** Searched every
occurrence of `alt` and `diverged` in the file (list above); `mod.rs:196-200`,
the doc comment directly above `diverged()`, explains only what divergence
*means* semantically ("a failed alternate pass counts as divergence... the
old census counted it as one too") — it says nothing about the boolean
being a deliberate, costed simplification. Reported as: nothing found, not
merely unread.

This issue is scoped to `alt` only, precisely because that documented
rationale — and the deferral it justifies — does not extend to it.

## Why `alt`'s fix is cheaper than `coll`'s would be

Unlike `coll` (a `bool` field that would need to become a count, i.e. a
schema change), `alt` already carries `Option<usize>` — the count is already
in the TSV (`mod.rs:1025`, `# alt: open edges under the alternate
triangulator, or x if that pass failed.`). A magnitude comparison analogous
to `mod.rs:720-723` needs no new column and no reparse logic. It would,
however, need its own re-bless pass over the golden: all 37 rows currently
frozen with `alt != open` would need their `alt` values re-blessed as a
starting baseline before a magnitude gate could compare future runs against
them without instantly reporting every one as newly-worse.

## Not claimed

- I have not run the census suite (no `cargo` available in this
  environment). Every fact above is read statically from
  `rust/geometry/tests/census_golden/mod.rs`,
  `rust/geometry/tests/manifests/watertightness_census.tsv`, and the PR
  #5040 / commit history, all via `git show <ref>:<path>` against
  `upstream/main` at `776bd0cae98cbe935ddf6d11d5acd62155f9fed4`.
- I have not verified that `alt` moving 6→7 on `rvt01 #31323` reflects a
  real triangulator regression versus an artifact of whatever #5033/#5040
  changed elsewhere — only that the golden's own comparison logic could not
  have caught it either way, which is the defect being reported.
- I am not proposing any change to `coll`'s handling; that limitation is
  already documented and costed by its author.
- I have not audited whether `diverged()`'s counter use at `mod.rs:330`
  (`t.non_invariant`) is itself gated anywhere downstream of the
  `Totals`/`t` struct — only that it is a summary count, not a per-row
  pass/fail branch, based on what's visible in this file.

BIMvoice:
Closes #5060.

`HostRow::diverged()` is `self.alt != Some(self.open)` — a boolean — and it is the only thing in `classify` that reads `alt`. So once golden and run are **both** diverged for a host, `alt` can move to any larger value and neither existing clause fires: they watch the agree/disagree state change, not the count behind it.

Live instance: `various/rvt01.ifc` id `31323` moved `alt` 6 → 7 with `open` held at 8 (blessed in #5040). `diverged()` was true before (6 ≠ 8) and after (7 ≠ 8), so nothing could have caught it. 37 of ~1170 manifest rows are already in that frozen state.

The fix sits beside the existing boolean pair and mirrors `open`'s own magnitude check:

```rust
if let (Some(ga), Some(ra)) = (g.alt, r.alt) {
    if r.diverged() && g.diverged() && ra > ga {
        c.worse_counts.push(format!(
            "alternate-triangulator open edges {ga} -> {ra} (already diverged from `open`)"
        ));
    }
}
```

Gated on **both** sides already being diverged so it never overlaps the boolean clauses — on a non-diverged host `alt == Some(open)` by definition and `open`'s own check is the whole story. A `None` on either side (the alternate pass failed to run) is left alone: `diverged()` already counts a failed pass as divergence, and there is no count there to compare against.

**No re-bless, and that is load-bearing rather than incidental.** The comparison is golden vs. run, and the golden's `alt` values already are what running today reproduces, so `ra > ga` does not fire on any of the 37 frozen rows. This should be a no-op for the current manifest.

**`coll` is deliberately untouched.** It has the identical boolean-hides-a-count shape, but you documented and costed that one in this file — carrying the degenerate count "is the better fix" but "changes the golden's schema and so needs its own re-bless". No equivalent deferral is on record for `alt`, and `alt` needs no schema change since the count is already in the TSV. Fixing `coll` would be a separate PR with a bless pass.

## What is unverified

**I have no Rust toolchain here, so none of this has been compiled or run.** Please treat the test as unexecuted until CI or you say otherwise. The unit test is in the file's existing `#[cfg(test)] mod tests` harness, following the neighbouring `diverged()` tests, and uses only helpers already defined there — `row(model, id, open, tris)`, `swept(&[&str])`, `diff(&[HostRow], &[HostRow], &BTreeSet<String>)` and `Delta::reasons` — whose signatures I checked against their definitions in the same file. Types: `alt` is `pub alt: Option<usize>`, `usize` is `Copy`, so `if let (Some(ga), Some(ra)) = (g.alt, r.alt)` binds two `usize` out of `&HostRow` without moving, and `diverged(&self) -> bool` is available on both arguments of `classify(g: &HostRow, r: &HostRow)`.

The test covers two cases: both sides diverged with `alt` worsening and `open` fixed, which isolates the new clause; and a `None` on one side, asserting it stays silent rather than inventing a magnitude.

To verify: `cargo test -p ifc-lite-geometry --test triangulation_invariance` — the census golden's unit tests run in the default suite, without `--features triangulation-alt`.

No changeset: `ifc-lite-geometry` is a Rust crate, not an npm package. Module-size gate clean, and it does not track this file in any case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved geometry validation reporting to identify cases where alternate triangulation results worsen, including scenarios with existing discrepancies.
  - Preserved silence for comparisons where the alternate triangulation step is unavailable.
  - Added coverage to ensure these outcomes are reported consistently.


louistrue (approved):
Reviewed the classifier and its focused regression coverage. The new magnitude gate is correctly limited to hosts already diverged on both golden and run, so it closes the blind spot without duplicating the existing transition or `open` checks. Approved pending required CI.

louistrue (changes requested):
Required Rust CI failure: `a_clean_diff_can_never_let_a_derived_total_grow` now observes 56,448 clean pairs, not the pinned 63,648, because the new `alt` worsening branch correctly sends 7,200 formerly clean variant pairs through `requires_bless()`. Please update that intentional exhaustive-sweep expectation (and its nearby explanatory count prose if it now claims the old clean population), then push for a fresh CI run. The new classification behavior itself is correct.

louistrue:
One addition to the requested correction: please cover the reverse failed-alternate direction too (`golden.alt = Some(...)`, `run.alt = None`) alongside the existing `None → Some` case. The magnitude clause must stay silent whenever either side has no count, regardless of direction; pinning both directions makes that contract explicit while you update the exhaustive clean-pair count.

BIMvoice:
Pushed `c1699e0`. Your diagnosis was right and the classification is unchanged — but there is a **second stale pin in the same test**, which is the part worth your attention.

`checked` is now 56448. I did not simply copy CI's observed value: I transcribed `variants()`, `classify()`, `diff()` and `reclassifications()` into a throwaway script and re-ran the full 2304 × 2304 sweep with and without the new clause. It reproduces 63648 → 56448 exactly, and the 7,200 falls out of the sweep's own dimensions — pairs where `open` is held equal or improved, both sides are already `diverged()`, and `r.alt > g.alt`. `variants()`'s `alt in [None, Some(0), Some(3), Some(9)]` makes exactly 2 of the 4 values both `Some` and diverged at each `open`, which is where the factor comes from.

**`landed` also moves, and CI never saw it.** Rust aborts a test at its first failing assertion, so `checked` failing masked the one below it. A worsened count outranks a re-tessellation verdict in `diff`'s routing, so pairs that used to land in `retessellated` and now also trip the `alt` clause are re-routed to `regressed`. Same simulation gives 4368 → 3984.

**I have pinned 3984, and you should treat that number with more suspicion than 56448.** 56448 is CI-observed. 3984 is a prediction from a Python port of your Rust logic that no toolchain has checked — I have no cargo in this environment. The port reproducing 56448 exactly is reassuring about its fidelity, not proof of it. If 3984 is wrong, the next run fails on that line with the real value and it is a one-character fix; I would rather guess and be visibly wrong there than leave a known-stale pin for you to hit.

`detected` (135680) is untouched and should stay: it is tallied from `c.retessellated` above the bless-skip, and the new clause only appends to `c.worse_counts`. The simulation agrees.

All three numbers now carry a comment explaining what moved them and why — including that 3984 is a derivation rather than a measurement — since a freshly changed constant beside a stale explanation is the exact shape I have been reporting on other PRs today.

One correction to my own PR text while I am here: I wrote that this change "should be a no-op for the current state". That is true of the 37 frozen manifest rows, which is what I had checked, and **not** true of this synthetic sweep over the full classification space. I should have said which one I meant.

BIMvoice:
Added in `7aa1bca` — the reverse direction is pinned beside the existing one.

```rust
let g_had_alt = HostRow { alt: Some(6), ..row("a.ifc", 1, 8, 800) };
let r_lost_alt = HostRow { alt: None, ..row("a.ifc", 1, 8, 800) };
let d = diff(&[g_had_alt], &[r_lost_alt], &swept(&["a.ifc"]));
assert!(d.regressed.is_empty(), ...);
```

You are right that one direction does not establish the contract. The clause is gated on `if let (Some(ga), Some(ra)) = (g.alt, r.alt)`, which is symmetric in *which* side is `None` — but symmetric-by-construction is exactly the kind of claim that stops being true after someone restructures the match, and a single direction cannot catch that.

**On the strength of the assertion**, since this is the part I would want challenged: I pinned `d.regressed.is_empty()` rather than only checking that no `alternate-triangulator` reason appears. That is a stronger claim than the clause under test needs, so I walked every other branch in `classify` for this exact pair before making it — `open` 8/8, `strict` 8/8, `tris` 800/800, `collapsed` false/false, `far` false/false, `pre` identical, `vol` `None` on both since `open != 0` — and both `diverged()` boolean clauses stay silent because both sides are already diverged: the golden via `alt=6 != Some(8)`, the run via `None`, which `diverged()` counts as divergence in its own right. Nothing can push into any bucket, so the pair is genuinely clean rather than merely free of magnitude reasons. If you would rather the assertion be narrowed to the reason text so it cannot pass for an unrelated reason later, say so and I will tighten it.

The exhaustive sweep counts are unaffected: `variants()` builds its own synthetic cross-product and `a_clean_diff_can_never_let_a_derived_total_grow` runs only over that set, so a separate `#[test]` calling `diff()` on two hand-built rows cannot move `checked` or `landed`.

Still unverified locally — no Rust toolchain here, so this is type-checked and logic-traced by eye only. `alt` is `Option<usize>`, `None` is valid there, and `row`/`swept`/`diff`/`Delta::reasons` all match their existing call sites in the same file. And my `landed = 3984` prediction from the previous commit remains unconfirmed: the Rust lane has not finished on either head yet, so if that number is wrong this run will say so.

BIMvoice:
Closing the loop I opened: **`Rust tests` passed on `7aa1bca`** (job 106086462235, run 35513867899, 13:44:03Z), with `a_clean_diff_can_never_let_a_derived_total_grow` green. So `landed = 3984` was right.

I want to be precise about what that does and does not establish, since I published it as a guess. CI agreeing with an `assert_eq!` I wrote means the Rust implementation produces 3984 — the same answer my Python port of `classify`/`diff`/`variants` produced. Two independent derivations agreeing is good evidence the port was faithful, and it is the only check available without a toolchain here. It is not the same as having run the real test before pushing, and if the port had been subtly wrong in a way that happened to cancel out, this would look identical.

Both re-pinned counts are now CI-confirmed rather than one confirmed and one predicted, and `detected` (135680) stayed put as expected.

The reverse failed-alternate case you asked for is in `7aa1bca` as well, so both directions of the `None` contract are pinned. Nothing outstanding from your review on this one as far as I can tell — happy to be told otherwise.

BIMvoice:
Re-review request — the `CHANGES_REQUESTED` decision predates the current head, and GitHub does not clear it on push.

- Review submitted `2026-09-20T13:04:10Z`; follow-up asking for the reverse alt direction `13:20:19Z`.
- Current head `7aa1bca` pushed `2026-09-20T13:33:37Z`.

Both requests are on that head:

1. **Re-pinned counts.** `checked` 63648 → 56448 (CI-observed, not predicted) and `landed` 4368 → 3984 (predicted at the time, since CI-confirmed).
2. **Reverse `Some → None` direction**, which was the `13:20:19Z` ask. `rust/geometry/tests/census_golden/mod.rs` now carries the mirror of the existing `None → Some(7)` case: `g_had_alt = HostRow { alt: Some(6), .. }` against `r_lost_alt = HostRow { alt: None, .. }`, over the same held-fixed `open = 8` row. The clause is gated on `if let (Some(ga), Some(ra)) = (g.alt, r.alt)`, which is symmetric in which side is `None`, so this direction must stay just as silent — and the test asserts that rather than assuming it.

CI on `7aa1bca` is green: 55 distinct check-runs on the head SHA, no `failure`, `timed_out`, `action_required` or `cancelled`, nothing queued. The only non-success is `Vercel Agent Review` → `neutral`.

No action needed from me that I can see, but if I have mis-read either request I would rather fix it than have you merge around it.

louistrue (approved):
Re-reviewed the revised head. The reverse None-direction coverage is present, and the corrected derived totals are confirmed by the exact Rust regression test plus green CI. Approved.

PR #5067, opened 2026-09-20
fix(renderer): stop a stale point-cloud handle deleting the live asset that reused its id

### LINKED ISSUE #5052: renderer: a stale point-cloud handle can delete the live asset that reused its id after a device-loss teardown
A point-cloud stream handle held across a device-loss teardown can delete the **live** asset that inherited its id, because removal is keyed on the id alone while the id is not unique across a teardown.

## The asymmetry

`packages/renderer/src/index.ts` tracks streams in `pointCloudStreamEpochs: Map<number, number>` (handle id → epoch). The streaming path asks whether the entry is *current*:

```ts
private currentPointCloudStreamRenderer(handle: PointCloudAssetHandle): PointCloudRenderer {
    if (this.deviceLost || !this.pointCloudRenderer
        || this.pointCloudStreamEpochs.get(handle.id) !== this.pointCloudStreamEpoch) throw rendererDeviceLostError();
```

The removal path asks only whether it is *present*:

```ts
removePointCloudAsset(handle: PointCloudAssetHandle): void {
    if (!this.pointCloudStreamEpochs.delete(handle.id)) return;
    this.pointCloudRenderer?.removeAsset(handle);
```

## Why the id is not enough

`PointCloudRenderer` allocates from a per-instance counter — `private nextHandleId = 1` — and `teardown()` replaces the instance:

```ts
this.pointCloudStreamEpoch++; this.pointCloudStreamEpochs.clear(); this.pointCloudRenderer?.clear();
this.pointCloudRenderer = null;
```

so the replacement restarts at 1 and reissues ids the dead instance handed out.

A stale handle reaches removal from ordinary code, not a contrived path: `apps/viewer/src/hooks/ingest/pointCloudIngest.ts` captures `handle` in a closure and calls `opts.renderer.removePointCloudAsset(handle)` from both `onError` and the surrounding `catch`, after the async streaming work.

## Sequence

1. A stream holds `{id: 1}` in epoch 0.
2. Device loss → `teardown()` → epoch 1, map cleared, renderer replaced.
3. A new stream begins and is allocated `{id: 1}` in epoch 1.
4. The original stream's worker errors; its closure calls `removePointCloudAsset({id: 1})`.
5. The delete finds the **new** entry, removes it, and calls `pointCloudRenderer.removeAsset({id: 1})` against the **current** renderer — destroying the live point cloud.

## What the existing tests miss

`'stale cleanup must not remove a replacement asset with the same id'` and `'removes a live point-cloud asset through a rebuilt { id } handle (#4885 review)'` both look like they cover this, but neither populates a **live** entry at the colliding id before invoking removal with the stale handle — the replacement mock has no `beginAsset`, so the map is empty at the moment of the check. The collision, which is the only case that matters, is never constructed.

## Not claimed

I have not observed this in a running viewer. Every step above is read from the code at the SHA named, and the timing window — a worker error arriving after a teardown *and* a new stream having started — is reasoned, not measured. What is demonstrated is that nothing refuses it.

## Ready branch

`fix-pointcloud-epoch-removal` @ `a26ea813`, based on `main`, merges clean.

The fix is not the epoch check — I proposed that first and it does **not** work, because once the replacement registers under the reused id, `get(handle.id)` *is* the current epoch and the two cases become indistinguishable. What closes it is carrying the outgoing instance's handle-id watermark into the replacement so an id is never reissued; a stale handle then finds no entry and the existing lookup is already correct. `removePointCloudAsset` ends up byte-identical to `main`.

Verified on that head: `renderer-device-recovery.test.ts` 19 → 20 with a test that constructs the collision, full renderer suite 1588 pass / 1 skip unchanged, module-size gate clean with the allowlist untouched (`+0 added, ^0 raised`), and reverting just the `teardown()` watermark read fails exactly the new test with the other 18 green.

Happy to open a PR if you label this, or to bin the branch if you would rather solve it differently.

BIMvoice:
Closes #5052.

A point-cloud handle held across a device-loss teardown can delete the **live** asset that inherited its id. `removePointCloudAsset` keys on the id alone, and the id is not unique across a teardown: `PointCloudRenderer` allocates from a per-instance `nextHandleId = 1` and `teardown()` replaces the instance. A stale handle reaches removal from ordinary code — `pointCloudIngest.ts` captures it in a closure and calls `removePointCloudAsset` from both `onError` and the surrounding `catch`, after the async work.

## The fix is the watermark, and only the watermark

I first proposed adding an epoch check to `removePointCloudAsset`, mirroring what the streaming path does. **That does not work**, and it is worth stating because it looks like it should: once the replacement stream registers under the reused id, `get(handle.id)` *is* the current epoch, so a stale handle and a live one become indistinguishable exactly where the guard is meant to separate them. Built it, watched the live asset die with the check in place.

What closes it is stopping the id from being reused. `teardown()` now reads the outgoing instance's handle-id watermark and seeds the replacement, so an id is never reissued and a stale `{id: N}` finds no entry — which the existing `Map` lookup already handles correctly.

Having been wrong once, I had the redundancy tested rather than assumed: with the watermark in place, reverting the epoch check leaves the collision test green. So it is not here. **`removePointCloudAsset` is byte-identical to `main`** — `git diff upstream/main` shows no hunk for it. One mechanism that provably works, rather than two where one is decoration.

## The test constructs the collision the existing ones do not

`'stale cleanup must not remove a replacement asset with the same id'` and `'removes a live point-cloud asset through a rebuilt { id } handle (#4885 review)'` both read as though they cover this, but neither populates a live entry at the colliding id before removing through the stale handle — the replacement mock has no `beginAsset`, so the map is empty at the moment of the check. The new test begins a stream, tears down, begins a second stream that receives the colliding id, then removes through the **first** handle.

Verified on the pushed head, not only in authoring: `renderer-device-recovery.test.ts` 20/20, full renderer suite **1588 pass / 1 skip**, and reverting just the `teardown()` watermark read fails **exactly one** test — the new one — with the other 18 green.

## Module size, and why there is a refactor commit

Both touched files sat at exactly their recorded budgets with zero headroom, and the wiring put them over. Rather than raise an allowlist entry to land a bug fix — the ratchet-loosening I have been flagging on other PRs today — the lifecycle was extracted into two modules following the repo's own precedents: `point-cloud-stream-lifecycle.ts` uses the `scene.ts` → `scene-device-recovery.ts` host-interface shape, and `point-cloud-pick-sources.ts` mirrors `buildRayQuerySources` in `point-cloud-ray-transform.ts`.

The gate now passes outright: no "grew PAST" block, and the allowlist diff is `+0 added, ^0 raised, v0 lowered, -0 deleted`. Suite identical before and after the extraction, and the mutation check still bites afterwards — which is what proves the extraction did not move the watermark read out of the tested path.

## One gap I did not close

No test reaches the real `new PointCloudRenderer(...)` inside `initOnce()` — every device-recovery test replaces `initOnce` wholesale, as they did before this change. A later commit here adds a GPU-free test covering the constructor→`handleIds` wiring, but I checked rather than assumed: mutating that call site's argument to a literal `1` leaves it green. Reaching it for real needs a DOM shim (`RendererOverlays.init()` pulls in `SymbolicTextAtlas`, which wants `document`) plus a much larger WebGPU surface than that file's stubbing style covers, so the positional argument carrying the watermark is still unverified by anything executing.

Also in here: a changeset correction. The original text claimed `removePointCloudAsset` gained an epoch check — a sentence left over from the first iteration that the shipped code contradicts.


louistrue (approved):
Reviewed the stale-handle collision and recovery flow. Carrying the allocator watermark across teardown makes handle IDs unique across renderer replacement, so stale cleanup cannot alias a live replacement asset; it preserves the existing id-only cleanup contract. The changeset is appropriate and the focused collision regression is meaningful. Approved pending full CI.

PR #5068, opened 2026-09-20
ci(codegen): verify canonical Rust type artifacts

### LINKED ISSUE #5054: codegen: nothing regenerates rust/core/src/generated/schema.rs or type_ids.rs, so a hand-edit survives CI
`scripts/check-codegen-sync.mjs` regenerates and byte-diffs the per-schema Rust registries, but not the canonical type universe — so a hand-edit to `rust/core/src/generated/schema.rs` or `type_ids.rs` survives CI, while the same edit to their smaller siblings fails.

## What the ratchet covers

```js
const rustSchemaInputs = [
  ['IFC2X3_TC1.exp', 'ifc2x3'],
  ['IFC4_ADD2_TC1.exp', 'ifc4'],
];
for (const [schemaFile, registry] of rustSchemaInputs) {
  ...
  results.push({
    name: `rust/core/src/generated/${registry} (generated from ${schemaFile})`,
    ...diffDirs(rustOut, join(root, 'rust/core/src/generated', registry)),
  });
}
```

`schema.rs` and `type_ids.rs` sit at the top level of `rust/core/src/generated/`, not inside a `${registry}` subdirectory, so no `diffDirs` call reaches them. Grepping the whole script, those two registries are the only `rust/core/src/generated/...` paths it names.

## Why it matters more now

That is the artifact #5014 grew. Measured with the repo's own `scripts/lib/rust-schema-names.mjs` against the committed bytes either side of that merge:

```
canonicalGeneratedNames   pre 876   now 876
generatedNames            pre 876   now 1021
```

145 names added through the new supplemental merge path (`entityCatalogSchema('IFC4_FAMILY', ENTITIES_IFC4, …)` → `mergeTypeUniverse`), in the one generated artifact nothing regenerates. The generator that produces it also got materially more complex in the same change.

The gap predates #5014 — `scripts/check-generated.mjs`'s own header documents the general shape by name — so this is not a regression. It is that the uncovered surface grew by 145 names and a large generator diff on the same day, while the two covered siblings stayed small.

## Not claimed

I could not run `check-codegen-sync.mjs` itself: it shells out to a `pnpm --filter` build, which I am not running here. So the claim is from reading the script and the generated files, not from observing the gate pass over a mutated `schema.rs`. I also have not checked whether some *other* workflow regenerates these two files — I grepped `.github/workflows/*.yml` and `scripts/check-generated.mjs` for `generated/schema.rs` and `type_ids.rs` and found nothing, but that is an absence of evidence from two places, not an exhaustive sweep.

## Possible overlap

#4203 (`ready`, assigned) covers generating the Rust type universe per schema version. This is adjacent but not the same claim — that issue is about what the generator produces, this is about whether CI notices when the committed output stops matching it. If you would rather this were an acceptance criterion on #4203 than a separate issue, that seems reasonable and I have no attachment to it standing alone.

Extending the existing loop to the canonical output looks like the whole fix, though it would need the generator invoked the way that file was actually produced (the IFC4X3 `.exp` plus the supplemental `ENTITIES_IFC4` merge), which is more than a third tuple in the array.

louistrue:
Part of #4203; closes #5054.

Regenerates the public IFC4X3 Rust type universe with its production IFC4 and IFC2X3 supplements, rustfmt-finalizes the result, and byte-compares schema.rs and type_ids.rs. The gate now fails if either top-level artifact is hand-edited or stale.

## Summary by CodeRabbit

- **Tests**
  - Expanded code-generation freshness checks to include canonical Rust schema and type ID outputs.
  - Validation now covers production IFC4 and IFC2X3 supplemental schemas.
  - Checks report missing, extra, or mismatched generated files more clearly.

- **Chores**
  - Automated verification now formats generated Rust output before comparison.
  - The verification workflow uses the stable Rust toolchain with formatting support.


BIMvoice:
Reviewed this at `ce5ce9e3`. The gate itself reads right — regenerating and rustfmt-ing before the byte compare is the only way that comparison means anything, and normalising either side would let a hand edit through. **One wiring point looks off**, and your own comment on the new step is what made me check it.

## The toolchain step is in `revert-oracle`, not the job that calls `rustfmt`

The new call is unguarded — no try/catch, no existence check:

```js
if (rustDir && !options.cratePrivate) {
  ...
  execFileSync('rustfmt', ['+stable', join(rustDir, 'schema.rs'), join(rustDir, 'type_ids.rs')], {
```

so a missing binary is an ENOENT crash, not a staleness verdict.

Two steps reach it, and both are in **`node-tests`**:

- `node --test scripts/*.test.mjs ...` (the glob catch-all), which picks up this PR's new `describe('real repo — runAllTargets ...')` block. Its only skip gate is `existsSync('packages/codegen/schemas/IFC2X3_TC1.exp')` — nothing about a Rust toolchain.
- `pnpm run check:codegen-sync` → `node scripts/check-codegen-sync.mjs` → `runAllTargets` → `checkCanonicalRustCodegen`.

The toolchain step this PR adds sits in **`revert-oracle`** (its job block starts at `test.yml:573`; the step is at `:598`). Grepping the whole `node-tests` body for `dtolnay|rust-toolchain|components:` returns **0**.

The step's own comment is why I did not let this go as a style nit:

> Keep that check in this Node-owned lane, but make its formatter dependency explicit rather than relying on a runner image incidental.

That reads as intending to cover the lane that runs the freshness check — and `revert-oracle` never calls `check-codegen-sync.mjs`, the new lib, or `rustfmt`.

## Where it would actually bite, and where it would not

`node-tests` runs on `frontend == 'true' || rust == 'true'`, and the `frontend` filter includes `scripts/**`, `packages/**`, `apps/viewer/**`. So a PR touching only viewer code reaches this code with no Rust toolchain step.

Being fair about the blast radius: on the `ubuntu-latest` fallback (fork, draft, or `ci:hosted`) the hosted image ships a rustup toolchain, which would mask it. The exposure is the self-hosted `ifclite` path, where whether `rustfmt` is present depends on what previous jobs left behind — which is precisely the "runner image incidental" your comment says you are trying to stop relying on. Moving the step into `node-tests`, or adding a second one there, would close it.

I have not run any of this — no cargo or CI here, and I deliberately did not run `rustfmt` locally since this machine says nothing about the runner. The job placement and the absence of a toolchain step in `node-tests` are read from the workflow at that SHA; the self-hosted runner's actual state is the one thing I cannot determine from the repo.

## Otherwise

`--rust-supplemental-schema` and `--rust-crate-private` already exist in `packages/codegen/src/cli.ts` and `rust-generator.ts`, so the new lib is wiring existing flags rather than inventing an interface. Regenerating the canonical artifacts and diffing them is exactly the gap I raised in #5054 — with this, a hand edit to `schema.rs` or `type_ids.rs` stops surviving CI, which was the whole concern there.

louistrue:
Stack update for #4203: this foundational codegen/regeneration gate is fully green (including Build + WASM + Rust + Node, semver, typecheck, lint, and review signals). The downstream schema, unknown-keyword, and legacy-retirement PRs are ready behind it. A maintainer approval is the remaining merge prerequisite; please review when available.

BIMvoice:
Two follow-ups on my earlier comment, one closing a question I raised and one turning a guess into evidence.

**The gate is sound — I had not established that, and it matters more than the wiring point.** I asked myself whether the regeneration truly reproduces the committed artifact, because a regeneration path that differs from how the file was produced either fires on every run or normalises both sides into agreement, and the second kind examines nothing. It matches: `checkCanonicalRustCodegen` invokes the same primary schema and the same supplemental order as the documented production command in `packages/codegen/README.md`, and the rustfmt pass is applied only to the freshly generated side, so a hand edit still produces a byte diff rather than being normalised away. Your own comment says exactly that — *"comparing pre-format text would report formatting noise, while normalising either side would let a hand edit through"* — and the code does what the comment claims. The run on `b3ad9938e` prints `PASS` for both `schema.rs` and `type_ids.rs` with a populated regeneration, so it is a live check rather than a self-agreeing one. That closes #5054 properly.

**On the toolchain step, I can now replace my hedge with a fact.** I said the exposure was the self-hosted path and that I could not determine the runner's state from the repo. The run on this head settles it: the `node-tests` job's step list is `checkout → pnpm/action-setup → setup-node → pnpm install --frozen-lockfile → Download build artifacts → …` with **no Rust toolchain step at all**, and `Check codegen output freshness` still succeeded — so `rustfmt` was there incidentally. That is precisely the dependency your added step says it is removing:

> Keep that check in this Node-owned lane, but make its formatter dependency explicit rather than relying on a runner image incidental.

The step is in `revert-oracle`, which never calls `check-codegen-sync.mjs`. It can reach this code only when the revert-oracle happens to re-run this PR's own changed test — so it covers this PR and not the general case.

Worth adding: the failure mode is at least loud. An absent `rustfmt` throws out of `runCodegenCli`, past `runAllTargets`'s `try/finally` which has no `catch`, into the CLI's handler that prints `❌ check-codegen-sync: regeneration failed` and exits 1 — in a required step with no `continue-on-error`. So a runner without it blocks merge rather than reporting a silent pass, which is the right way round. It would just block on a tooling error rather than a staleness verdict, on a PR that may not have touched Rust at all.

PR #5069, opened 2026-09-20
feat(core): expose schema-local entity metadata

louistrue:
Part of #4203.\n\nBuilds on #5068. Exposes source-schema entity membership, direct parent, abstractness, and inheritance queries from the independently generated IFC2X3, IFC4, and IFC4X3 registries. Unknown names fail closed; no result is inferred from the merged public type universe.


BIMvoice:
Reviewed this alongside #5068. **The Rust changes read clean** — but this PR is getting no Rust CI at all, and the checks page does not show that.

## Base is `codex/issue-4203-per-schema-types`, so `test.yml` never triggers

`test.yml` filters `branches: [main]`, so a PR based on a feature branch runs none of its lanes. Counting check-run names on the two head SHAs rather than reading the checks tab:

- `ce5ce9e3` (#5068, base `main`): **45** distinct lanes
- `b6e46e56` (this PR, base `codex/issue-4203-per-schema-types`): **27**

The 26 present on #5068 and absent here:

```
AGENTS.md ratchet              Node tests
Build packages + WASM          Plato clash-math freshness
Changed tests observe production   Provenance tests
CSG accept gates (feature builds)  Rust crate semver
Detect changes                 Rust tests
Docs checks                    Rust tests (feature builds)
Docs site build                Typecheck
Geometry watertightness census Viewer E2E smoke
Lint                           Viewer tests (shard 0-7)
MPL license headers
```

**`Rust tests` and `Rust tests (feature builds)` are both in that list, on a diff that is entirely `rust/core/src/**`.** The 27 lanes that do run are `cost parity`, `wheel`/PyPI publish jobs, Vercel preview comments, Macroscope, Cursor Bugbot and similar — none of which compile or test the code this PR changes.

What makes it worth raising rather than shrugging at is that absent CI and passing CI render identically. The page fills with green ticks from the bots that *do* run, and nothing states that `cargo test` never executed on this head. If this rebases onto `main` after #5068 merges and lands on the strength of those ticks, `is_subtype_of` and `entity_info_for_schema` will have merged without their test lane ever having run.

I hit the same thing on #5025 earlier today — rebasing onto `main` once the parent merged was all it needed, and the lane count went 23 → 33.

## The diff itself

Additive as far as I can tell. `attribute_names_for_schema` is re-exported unchanged beside the new `entity_info_for_schema` / `is_subtype_of_for_schema` / `SchemaEntityInfo`, and the three per-schema `entity_info_*` functions call only pre-existing generated methods (`from_str`, `as_str`, `parent`, `is_abstract`, `declared_by_canonical_schema`), so nothing is invented and no existing path is displaced.

**One coverage note:** `generated_registries_answer_schema_local_inheritance_facts` exercises a depth-1 parent match — `IFCDOORSTYLE`'s direct parent is `IFCTYPEPRODUCT` — plus a fail-closed unknown-name case. It never walks more than one hop, so `is_subtype_of`'s `while current = current.parent` loop would still pass this test if it were reduced to a single parent check. A two-hop case would pin the loop itself. Not a break, just the depth the test actually reaches.

`is_subtype_of` has no cycle guard, which is inert for an EXPRESS-derived tree — noting it only so it is on the record rather than as a concern.

Source-level only: no Rust toolchain here, so none of the above was compiled or run, and the lane counts are the one part that is measured rather than read.

louistrue:
The revert oracle's whole-file Rust rollback removes the newly exported schema-registry API before `schema_registry_tests` can compile, producing its documented INCONCLUSIVE capability gap (not an unobserved test). The five behavior tests pass normally and directly cover the new per-schema metadata/inheritance behavior. Applying the maintainer-only `revert-oracle-exempt` label for this public API addition; no test coverage is being weakened.

PR #5070, opened 2026-09-20
feat(core): generate IFC4X1 and IFC4X2 registries

### LINKED ISSUE #4203: Rust IfcType: generate the type universe per schema version and retire the legacy tables

louistrue:
Part of #4203.

Pins the official IFC4X1 and IFC4X2 EXPRESS inputs, generates crate-private registries, and replaces the export path’s hand-written transitional positional layouts with source-schema metadata. Extends the regeneration gate to each added registry.

Does not close #4203: remaining work includes retiring the general legacy tables and preserving arbitrary unknown names.

## How it was verified

- `cargo test -p ifc-lite-core --lib -- schema_registry_tests::` — 6 passed, including IFC4X1/IFC4X2 label, attribute, and inheritance assertions.
- `node --test scripts/check-codegen-sync.test.mjs` — 10 passed, including every bundled Rust schema target against a clean checkout.
- Fresh main-based PR CI passed: build/WASM type synchronization, Rust, Node, typecheck, lint, viewer, and provenance lanes.


BIMvoice:
Reviewed this at `0139af7a`. **The registry work looks right, and it closes something I raised on #4996** — but the stack is now three deep and only its base is getting real CI.

## The fallback is genuinely gone, not left dead

`uses_ifc4_attribute_layout` and `ifc4x2_infrastructure_attribute_names` are deleted — `grep` for the former in `rust/` returns nothing at this head — and `render_attributes` is now a single direct lookup with no borrow to silently win over a real registry. That is the clean version of the two-literal allowlist I flagged on #4996.

I checked the thing that would have made it a regression: whether the new `ifc4x1`/`ifc4x2` registries actually carry what the deleted tables used to return. For the alignment family plus the three hand-pinned infrastructure layouts, the attribute lists match exactly, so a file declaring `IFC4X1` or `IFC4X2` gets byte-identical output before and after.

Worth recording for the #4996 thread: this does **not** change the plain-`IFC4` case, and correctly so. `IFC4_ADD2_TC1.exp` genuinely does not declare `IFCALIGNMENTCURVE`, so `attribute_names_for_schema("IFC4", …)` returns `None` either side of this PR. `legacy_attribute_names.rs` still tags those fifteen `// IFC4`, but that table is no longer reachable from the render path at all now that the fallback is gone — so the mislabeling I reported is now inert rather than latent. That is a better outcome than the fix I had imagined.

Adding `['IFC4X1.exp','ifc4x1']` and `['IFC4X2.exp','ifc4x2']` to `rustSchemaInputs` also means these two do not repeat the gap I filed as #5054 — they are regenerated and byte-diffed from the start.

## The stack is three deep, and the top two run no tests

- #5068 → base `main`: **45** distinct check-run names
- #5069 → base `codex/issue-4203-per-schema-types`: **27**
- #5070 (this) → base `codex/issue-4203-schema-metadata`: **27**

`test.yml` filters `branches: [main]`, so only the bottom of the stack triggers it. Missing here are `Rust tests`, `Rust tests (feature builds)`, `Rust crate semver`, `Build packages + WASM`, `Lint`, `Typecheck`, `Node tests`, `AGENTS.md ratchet`, `Provenance tests`, all eight viewer shards and `Viewer E2E smoke`. The 27 that do run are bots, release/wheel jobs and gates — none of which compile the generated Rust this PR adds.

So the registries, the `SchemaVersion` variants and the deleted fallback have not been compiled by CI on this head, and will not be until it sits on `main`. I raised the same thing on #5069; mentioning it here only because the tip of a three-deep stack is where it compounds — two of the three PRs are in that state simultaneously.

## One note on the new test

`generated_registries_use_ifc4x1_and_ifc4x2_not_transitional_tables` has one assertion that cannot fail for the right reason and one that can. The attribute-list check expects exactly what the deleted fallback already returned, so on its own it cannot distinguish "new registry wired up" from "old fallback still present". The `is_subtype_of_for_schema("IFC4X2", …)` assertion does carry the signal: before this PR there was no `Ifc4x2` variant, so `from_file_schema` returned `None` and that call was `false` unconditionally. Worth knowing which half of that test is doing the work if it ever needs trimming.

Source-level only — no Rust toolchain here, so the equivalence above is a byte comparison of the committed tables and registries, not a compile or a run. The lane counts are measured.

PR #5072, opened 2026-09-20
feat(core): preserve owned unknown IFC keywords

louistrue:
Part of #4203.

Makes unknown IFC keywords owned and exact through Rust, instead of reducing them to a CRC-only Unknown value. from_id now refuses hashes that cannot reconstruct a keyword.

## How it was verified

- cargo test -p ifc-lite-core --test crc32_spec_vectors — 6 passed, including owned keyword preservation and non-reconstructable IDs.
- cargo clippy -p ifc-lite-processing --test styling_parity -- -D warnings — passed after updating the non-Copy consumer.
- node scripts/check-rust-api-surface.mjs — passed with the regenerated 443-item Rust API manifest.
- pnpm --filter @ifc-lite/codegen test; node scripts/check-codegen-sync.mjs; node scripts/check-module-size.mjs; and Rust core, export, processing, and wasm library tests were completed on this implementation.
- Fresh main-based PR CI is running for the final head.

## Summary by CodeRabbit

- **New Features**
  - Unknown IFC types now preserve their normalized keyword names and CRC32 IDs.
  - Public Rust APIs expose unknown IFC types and related schema-query functionality.
  - Generated schemas now provide clearer public and crate-private layouts, including type IDs and schema metadata.

- **Bug Fixes**
  - Unknown type lookups now distinguish known values from unsupported values instead of treating unknown IDs as known types.
  - Improved handling preserves IFC type information across geometry, processing, and WebAssembly workflows.


BIMvoice:
Reviewed this at `de6fd440`. **The change itself reads clean** — and it makes one doc comment elsewhere false without touching it, which is the only thing I would ask you to fix.

## A comment in an untouched file becomes wrong

`rust/core/src/schema_helpers.rs` has zero diff in this PR, but says:

> A genuinely unknown name still becomes `IfcType::Unknown`, which stores a **CRC32 hash, not the name**, so the record is the only place its keyword still exists.

After this PR, `UnknownIfcType` stores `name: String` alongside `id: u32`, and `as_str()`/`name()` return the real keyword. So the record is no longer the only place it exists, and the emphasised clause is now the opposite of true.

Worth more than a doc nit because of what sits directly under it: `get_legacy_entity_info(decoded.as_str())`. With the old representation `as_str()` was always the literal `"UNKNOWN"`, so that lookup could never hit — it was a permanent no-op for unknown types. Your change makes it a real lookup. That is an improvement the comment now actively argues against, and the next reader reconciling the two will assume the code is wrong rather than the prose.

## Otherwise

`from_id` going `u32 -> Option<Self>` is the right call once a hash can no longer reconstruct a keyword, and it is safe here: grepping all of `rust/` for `IfcType::from_id` finds only two call sites, both tests. No production caller to break.

**No third source of truth introduced**, which was my main worry given #4203's purpose. `schema_helpers.rs` is untouched, the `legacy_entities.rs` and `lib.rs` diffs are a `Copy` derive fix and a re-export, and the deciders remain the generated registries plus `legacy_attribute_names.rs`. Nothing new decides whether a keyword is known.

**Empty and absent stay distinct** — `declared_by_canonical_schema()` still lists `Self::Unknown(_)` among its exclusions, so an unknown keyword reports "not declared" rather than "declared with no attributes", and the per-schema lookups guard on `Unknown` before reaching `known_as_str()`.

I also want to credit something, since it is the exact pattern I keep flagging: narrowing the `parent()` test to scope its substring search was necessary, not cosmetic. `known_as_str()` emits `Self::IfcRoot => Some("IFCROOT")`, which would have made the old unscoped `.not.toContain('Self::IfcRoot => Some(')` assertion pass for the wrong reason — a test that silently stops checking anything. Catching that while adding the method is the hard part.

The new test earns its place too: `as_str() == "IFC_VENDOR_WIDGET"` would have read `"UNKNOWN"` under the old representation, so it fails without the change rather than asserting what the code produces anyway.

**One thing I could not settle:** what happens to an owned unknown keyword across a schema conversion where it is unrepresentable. `rust/export/src/schema_convert.rs` has no `Unknown` handling at all (grep: 0 hits) and this PR does not touch the export path, so the round-trip question is neither introduced nor answered here — flagging it as out of scope rather than as a gap in this diff.

On CI: 20 distinct check-run names on this head, with `Rust tests`, `Build packages + WASM`, `Lint`, `Typecheck` and the viewer shards all absent — this is the fourth PR in the #5068 → #5069 → #5070 → #5072 chain, and three of the four are in that state. Same cause I noted on #5069 and #5070, stated here only because the count is the new part; I am not going to keep raising it.

Source-level only — no Rust toolchain here, so none of the above was compiled or run.

louistrue:
Addressed the review-noted stale unknown-keyword comment in c841c1a90. The same commit fixes the non-Copy consumer Clippy regression and updates the committed Rust API-surface manifest. Verified locally with  and .

louistrue:
Correction to the previous comment: local verification was cargo clippy -p ifc-lite-processing --test styling_parity -- -D warnings, plus node scripts/check-rust-api-surface.mjs. Both passed before c841c1a90 was pushed.

Issue #1496, opened 2026-07-02
Geometry and attribute passes disagree on what is a product — GLB nodes with no EntityRow (join-contract gap)


louistrue:
6 fixtures produce GLB nodes whose `expressId` has no `stream_export_model` row, so a viewer joining attributes on `expressId` gets orphaned, unpickable-metadata geometry. Pre-existing at `38bc966f` (pre-#1491), verified identical on `6d2cb21a`.

**Cause.** The geometry pass meshes any product with a representation, and `node_extras` falls back to `ifcType: "Unknown"` for keywords outside the known-type table. The attribute pass (`stream_export_model`, keyword pre-filter from #1465) only admits known product keywords — so these entities render but never yield an `EntityRow`.

**Concrete cases (all in `tests/models`):**
- `ifcopenshell/1030-sphere.ifc`, `ifcopenshell/1032-curve.ifc` — `IFCPROXY` (valid IFC2x3 occurrence, plausible in real Revit exports)
- `issues/860_solid_stratum.ifc` — `IFCSOLIDSTRATUM` (IFC4x3 geoscience)
- `buildingsmart/annex_e/tessellated-shape-with-style/tessellation-with-{blob,image,pixel}-texture.ifc` — `IFCBOILERTYPE` (type object with a meshed representation)

**Fix sketch.** Align the two passes' product criteria: (a) admit the missing occurrence keywords (`IFCPROXY`, geoscience strata, …) in the pre-filter and emit their real keyword in `node_extras` instead of "Unknown"; (b) decide the type-object case deliberately (either stop meshing bare type objects or emit a row for them); (c) optionally, as a downstream backstop, expose enough for wrappers to synthesize stub rows for orphan nodes. A test asserting `glb nodes ⊆ export-model rows` across the fixture corpus would lock the contract.

Issue #860, opened 2026-05-28
Unsupported parsing of IfcSolidStratum entity


xyzbety:
The parser fails to recognize or render IfcSolidStratum entities. When loading files containing such objects (typically used in geological or infrastructure models), the relevant data is missing or not displayed.

The attachment（UT_Tin_in_MGA_56） is a model of the IfcSolidStratum. Referring to the following comparison images.By the way, the spatial tree does not have a IFCFacility icon.

[UT_Tin_in_MGA_56.zip](https://github.com/user-attachments/files/28335136/UT_Tin_in_MGA_56.zip)

<img width="2550" height="1307" alt="Image" src="https://github.com/user-attachments/assets/a6c9845f-2b31-4348-9f41-3a0d4f9bdb21" />

<img width="2880" height="1716" alt="Image" src="https://github.com/user-attachments/assets/d860d831-7078-4aee-85d5-69390f24ff26" />

PR #5073, opened 2026-09-20
refactor(core): retire legacy IFC type tables

louistrue:
Part of #4203.

Replaces legacy nearest-modern mappings, attribute tables, and rooted tables with generated per-schema facts. Keeps only the three explicit non-EXPRESS stratum aliases as bounded compatibility policy.

## How it was verified

- cargo test -p ifc-lite-core --all-targets — passed.
- cargo test -p ifc-lite-export --all-targets — passed, including the exact-label geometry/export alias join.
- pnpm rust-api-surface:update followed by node scripts/check-rust-api-surface.mjs — passed; the retired surface is the generated 439-item manifest.
- Previous implementation validation covered codegen freshness across all targets, module-size, processing, WASM, and generated-file gates.
- Fresh main-based PR CI is running for this final rebased head.

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for exporter-specific stratum aliases: `IFCSOLIDSTRATUM`, `IFCVOIDSTRATUM`, and `IFCWATERSTRATUM`.
  * These aliases are now recognized during IFC type detection, geometry classification, rooted-type checks, and streaming export while preserving their original names.

* **Bug Fixes**
  * Improved handling of legacy and transitional IFC layouts by resolving supported keywords to their exact generated IFC types.
  * Unknown vendor-specific names continue to be rejected safely.

* **Refactor**
  * Replaced legacy entity lookup and rooted-type tables with generated schema metadata and bounded compatibility rules.

Final review remediation (4df3d9f): schema-local diagnostics aligned; exporter alias matching is allocation-free. Focused Rust validation: core schema helpers 7/7, exporter model 77/77.


BIMvoice:
Reviewed this at `3a4cfa64`. **The retirement itself looks right** — the twenty-two formerly-legacy keywords now resolve as exact generated variants, `legacy_attribute_names.rs` goes with its only caller, and I found no reader left behind for any deleted table. One narrow thing, and it is a comment rather than code.

## A comment that the three stratum aliases now falsify

`rust/processing/src/processor/mod.rs`:

> ```
> // Legacy-aware so a remapped entity (IfcProxy, IfcSolidStratum, …)
> // labels its node with the real base type, not "Unknown", and matches
> // the attribute pass's row type (#1496).
> let ifc_type = ifc_lite_core::ifc_type_from_keyword(type_name);
> ```

`ifc_type_from_keyword` is now just `IfcType::from_str(upper)`. `IFCSOLIDSTRATUM`, `IFCVOIDSTRATUM` and `IFCWATERSTRATUM` are not generated variants — zero matches in `schema.rs`, unlike the other twenty-two former-legacy names which each match exactly once — so those three land in `Unknown`. Before this stack they mapped to `IfcType::IfcGeotechnicalStratum` via the legacy table.

The change is deliberate and you re-pinned it: `issue_860_stratum_subtypes.rs` now asserts `ifc_type_from_keyword("IFCSOLIDSTRATUM").as_str() == "IFCSOLIDSTRATUM"`, which is the right expectation once `Unknown` carries its keyword (#5072). And `is_exporter_stratum_alias` covers the two places that would otherwise break: `compute_has_geometry` returns `true` so geometry is still scheduled, and `rooted_type.rs` special-cases them before `from_str` so GlobalId reconciliation is unaffected.

So the only thing actually wrong is the comment, which now states the opposite of what happens — it promises "the real base type, not Unknown" on the exact line that produces `Unknown` for these three. Worth fixing because the next person reconciling the two will assume the code regressed.

**The one substantive question it raises**, which I could not settle by reading: `default_color_for_type(ifc_type)` sits a few lines below, and `is_subtype_of` classification downstream (profile extraction, GPU batching, export) now sees `Unknown("IFCSOLIDSTRATUM")` rather than a subtype of `IfcProduct`. `has_geometry` is preserved by the alias check, so these render — but if colour or any subtype gate depended on `IfcGeotechnicalStratum`, that is a behaviour change the #860 test does not reach, since it asserts only the keyword round-trip. If you already know those paths do not care, ignore this.

## Checked and clean

The twenty-two non-stratum former-legacy keywords each resolve to exactly one generated variant. `legacy_attribute_names.rs` is deleted wholesale and I found no remaining reader — its consumer went in the same commit as its only caller. `declared_by_canonical_schema()` and `schema_registry.rs` are untouched, so "declared with no attributes" and "not declared here" stay distinct. The CI steps removed (`check-legacy-entity-coverage`, the rooted-type freshness check) match the generators they covered, with no orphaned step pointing at a deleted file.

Note for the record: an earlier pass of mine flagged four files still importing the deleted symbols — `issue_860_stratum_subtypes.rs`, `issue_3187_legacy_type_product_gate.rs`, `rooted_type_parity.rs` and `dump_rooted_type_sweep.rs`. All four are fixed at this head; that pass was reading `7757afb3`. Mentioning it only so the record is straight, not as an outstanding item.

Base `codex/issue-4203-owned-unknown`, 27 distinct check-run names on this head.

Source-level only — no Rust toolchain here, so none of the above was compiled.

BIMvoice:
Followed my earlier stratum note through to its blast radius by auditing the whole stack as one cumulative diff against `main`, which is a view no single PR review gets. **There is one real consequence, in attribute export**, and the code's own comment predicts it.

## `IFCSOLIDSTRATUM` now renders with no attribute row

`rust/export/src/model.rs`, pass 2:

```rust
// Filter on the STEP keyword *before* decoding, skipping the millions of
// non-product geometry primitives. `legacy_aware_ifc_type` (not a bare
// `from_str`) resolves removed/renamed keywords (IFCPROXY, IFCSOLIDSTRATUM,
// …) to their modern base type, matching what the geometry pass meshes —
// otherwise those products render as GLB nodes with no attribute row (#1496).
let ty = ifc_lite_core::ifc_type_from_keyword(type_name);
if !ty.is_subtype_of(IfcType::IfcProduct) {
    continue;
}
```

The comment names `IFCSOLIDSTRATUM` explicitly and names the failure mode. Traced at the tip (`3a4cfa64`):

- `ifc_type_from_keyword` is now `IfcType::from_str(upper)`, and the three stratum names have **zero** matches in the generated `schema.rs`, so they resolve to `Unknown`.
- `parent()` has no `Unknown` arm, so it returns `None` via the wildcard and `is_subtype_of`'s ancestor walk terminates on the first step — `Unknown.is_subtype_of(IfcProduct)` is `false`.
- So the `continue` fires and **no attribute row is emitted**.
- On `main` the same keyword resolved through `legacy_entities.rs` to `IfcGeotechnicalStratum`, whose chain is `IfcGeotechnicalStratum → IfcGeotechnicalElement → IfcElement → IfcProduct` — a product, so pass 2 emitted a row.

Meanwhile the geometry side is unaffected, because `compute_has_geometry` checks `is_exporter_stratum_alias(upper)` and returns `true` before it ever consults `from_str`. So the entity still meshes and still renders.

Rendered node, no attribute row — the #1496 shape the comment says this line exists to prevent.

**Nothing catches it.** I grepped every `.rs` occurrence of the three names: `issue_860_stratum_subtypes.rs` asserts the keyword round-trip and `has_geometry_by_name`, `rooted_type.rs`'s own test covers rootedness and GlobalId reconciliation, `merged/tests.rs` covers the reconciliation path. None reaches `model.rs`'s pass 2 for these keywords.

## What I checked that turned out fine

I expected wider fallout and did not find it. `default_color_for_type` has no `IfcGeotechnicalStratum` arm on either side, so both fall to the neutral default. `is_rooted_type` special-cases the aliases before `from_str`, so GlobalId export is unaffected. Profile extraction gates on `is_subtype_of(IfcFeatureElement)`, which was false for `IfcGeotechnicalStratum` too. GPU batching only feeds the colour path. So `model.rs` pass 2 appears to be the single consumer where `Unknown` and `IfcGeotechnicalStratum` give different answers.

`is_exporter_stratum_alias` is one-directional by construction — it can say "yes" where the registries say unknown, never the reverse — so adding it to that gate, or giving the three names a real variant, would both close it without reintroducing a second source of truth.

## The rest of the stack is cumulatively clean

Worth saying, since this was a five-PR diff and the interesting failure would have been a deletion in one step with a reader surviving in another. There is none: no `mod`/`use` for any deleted module, and every remaining mention of `legacy_entities`, `LEGACY_ROOTED_TYPES`, `LEGACY_ENTITY_NAMES`, `legacy_attribute_names` or the old `legacy_aware_*` names is inside a doc comment or test prose, not code.

The name universe is unchanged end to end — `generatedNames` 1021 and `canonicalGeneratedNames` 876 at both `main` and the tip, with no entity present on `main` and absent at the tip. And the thing #4203 set out to do is done: known-type and attribute-name resolution now route through one generated dispatcher per schema version rather than competing hand-kept tables.

Several comments still narrate `legacy_aware_ifc_type` and the legacy tables as current — `fast_parse.rs`, `batch.rs`, `rooted_type.rs`, `gltf_tests.rs`, `model.rs` — and `scripts/check-generated.mjs`'s header still lists the deleted rooted-types gate in its enumeration though the executable list is correctly trimmed. All cosmetic, except the `model.rs` one, which sits directly above the regression and describes the opposite of what now happens.

Source-level only — no Rust toolchain here, so the above is a code trace rather than an observed failing export.

louistrue:
Addressed the authoritative Node CI failure: #4203 retires the rooted-type generator, so its stale standalone freshness test and generated-gate inventory entry are now removed. The schema registry codegen sync check remains.

louistrue:
Addressed the latest exact-head Node failure: the WASM contract now verifies the generated registry preserves `IfcBeamStandardCase` exactly, replacing the retired legacy-table expectation that collapsed it to `IfcBeam`.

louistrue:
Addressed both current CodeRabbit findings in 4df3d9f: corrected stale legacy-table diagnostics across the resolver and GPU paths, and replaced the per-record uppercase allocation in the export filter with case-insensitive alias matching. Verified cargo test -p ifc-lite-core --lib schema_helpers (7 passed) and cargo test -p ifc-lite-export --lib model (77 passed).

PR #5079, opened 2026-09-20
fix(viewer): localize annotations and anonymized-export panels (#4918 slice N)

louistrue:
Part of #4918

## What changed
Converts the Annotate tool and the "Export anonymized subset" dialog to the i18n catalogue, following the established #4918 sweep pattern (23 prior slice PRs):

- `apps/viewer/src/components/viewer/annotations/{AnnotationDropInput,AnnotationLayer,AnnotationPin,AnnotationPopover}.tsx`
- `apps/viewer/src/components/viewer/anonymized-export/{AnonymizationOptionsPanel,AnonymizedExportDialog,RelatedEntityList,RelationTogglePanel,TypeCategoryBar}.tsx`
- New catalogues: `apps/viewer/src/i18n/catalogues/annotations.en.ts` (`annotations.*`) and `anonymized-export.en.ts` (`anonymizedExport.*`), registered in `apps/viewer/src/i18n/en.ts`
- New oracle tests: `AnnotationPopover.i18n.test.tsx`, `AnonymizedExportDialog.i18n.test.tsx`
- `apps/viewer/src/i18n/README.md` gains two coverage paragraphs
- `scripts/i18n-literals-baseline.json` regenerated (`--update`)

Annotation note TEXT, resolved entity TYPE NAMEs, and IFC class NAMEs (`TypeCategoryBar`'s chips) remain model/runtime content and stay out of the catalogue, per the README's house rules.

The two new oracle tests deliberately do **not** wrap their `describe` block in a catalogue-presence skip guard (a pattern some earlier slices used): with the catalogue module reverted, a missing key resolves to `undefined`, so `assert.equal(typeof value, 'string')` fails on its own — a real `ASSERTION_FAILURE` the revert-oracle can attribute — instead of a `describe.skip` collecting zero tests (`NO_TESTS`), which the oracle scores as `INCONCLUSIVE` regardless of other evidence in the run. This was confirmed against a full revert-oracle run before pushing (see Evidence).

## Why
Continues the #4918 localization sweep; these two directories share this slice only by directory proximity (two distinct features, two catalogues, one PR).

## Evidence
Before (per `scripts/i18n-literals-baseline.json`, all in this slice's scope):
```
 6  annotations/AnnotationDropInput.tsx
 1  annotations/AnnotationLayer.tsx
 2  annotations/AnnotationPin.tsx
11  annotations/AnnotationPopover.tsx
 7  anonymized-export/AnonymizationOptionsPanel.tsx
32  anonymized-export/AnonymizedExportDialog.tsx
 2  anonymized-export/RelatedEntityList.tsx
 3  anonymized-export/RelationTogglePanel.tsx
 7  anonymized-export/TypeCategoryBar.tsx
```
After: `node scripts/check-i18n-literals.mjs` reports **0** for all 9 files.

Commands run locally, all green:
```
pnpm exec turbo typecheck --filter=@ifc-lite/viewer
node scripts/check-i18n-literals.mjs           # OK, 0 for all 9 target files
node scripts/check-module-size.mjs             # OK, +0 ^0 v0 -0 vs merge-base
node scripts/check-test-wiring.mjs             # OK
node scripts/check-test-revert-oracle.mjs --base origin/main --ci --json   # OBSERVED
npx tsx --import ./src/test/vite-module-hooks.mjs --test <new *.i18n.test.tsx files>   # pass
npx tsx --import ./src/test/vite-module-hooks.mjs --test src/i18n/negotiate-locale.test.ts src/i18n/catalogue-problems.test.ts src/i18n/intlFormat.test.ts   # pass
```

`scripts/check-unused-locals.mjs` and `scripts/check-changesets.mjs` could not be run locally (known `spawnSync pnpm ENOENT` on this Windows shell, per prior slices) — reviewed the diff by hand for unused imports/locals instead.

## Summary by CodeRabbit

* **New Features**
  * Added localization for annotation tools, including drop input, pins, previews, popovers, editing controls, accessibility labels, and relative timestamps.
  * Added localization throughout anonymized export workflows, including dialogs, options, relationship controls, entity lists, type filters, statuses, warnings, and errors.
  * Added English translation catalogues for annotations and anonymized export.

* **Documentation**
  * Expanded localization coverage documentation for annotation and anonymized export interfaces.

* **Tests**
  * Added localization tests covering annotation and anonymized export dialogs, controls, placeholders, accessibility text, and locale switching.


BIMvoice:
One confirmed regression in the warning count's plural, at head `6fde998`.

**On `main`**, `AnonymizedExportDialog.tsx:156` carried two independent count-dependent fragments, and pluralized the warning one:

```ts
+ (warningCount > 0 ? ` (${warningCount} warning${warningCount === 1 ? '' : 's'})` : '');
```

**On this head**, both plural forms hardcode the plural word (`anonymized-export.en.ts:109-112`):

```ts
'anonymizedExport.dialog.exportedEntitiesWithWarnings': {
  one: 'Exported {count} entity ({warnings} warnings)',
  other: 'Exported {count} entities ({warnings} warnings)',
},
```

and the call site passes both counts as separate params (`AnonymizedExportDialog.tsx:158`):

```ts
t('anonymizedExport.dialog.exportedEntitiesWithWarnings', { count: result.stats.entityCount, warnings: warningCount })
```

`pluralForm` selects the form from `params.count` alone — `registry.ts:90`:

```ts
const count = Object.hasOwn(params, 'count') ? params.count : undefined;
```

So `warnings` never selects a form, and **one warning now renders as "(1 warnings)"**.

Worth being precise about the net change, because it is not purely a loss: `main` hardcoded `entities` and never pluralized the entity count, so this PR *fixes* that half. The trade is that the half `main` got right is now wrong. With one entity and one warning: `main` gave "Exported 1 entities (1 warning)", this head gives "Exported 1 entity (1 warnings)".

The single-count mechanism can't express two independent plurals in one message, so this needs either a separately pluralized warning fragment composed in, or the warning count split into its own message.

Note `clashTools.setFilter.unreadableWarning` in #5077 handles a genuine double-plural correctly — both fragments ("One rule…that rule" / "{count} rules…those rules") vary together off the same count, so one selector covers it. Here the two counts are independent, which is what breaks it.

---

Also checked, and clean, across #5075/#5076/#5077/#5079 together:

- **No catalogue key collisions.** 470 distinct keys extracted structurally via the TypeScript compiler API (not regex), including all six cross-PR pairings — the case that matters, since each PR is green alone and a collision would only appear once two merge, with the later spread silently overwriting the earlier. Zero keys defined twice with differing values. None of the new prefixes exist on `main`.
- **Every other converted interpolation preserves its parameter count** — ~85 sites, and each manual `count === 1 ? '' : 's'` maps 1:1 to a correct `{one, other}` pair.
- All four carry `apps/viewer/src/i18n/catalogues/reposition-panel.en.ts`, byte-identical across all four heads (`md5` = `e73fcaff44575bc059565ae79c648e10`). Harmless as add/add today, but if any one of the four edits that file before the others merge it becomes a real conflict.

One limitation I'd rather state than paper over: `Lint` was still `in_progress`/`queued` on #5076, #5077 and #5079 when I looked, so whether `check-i18n-literals` actually completed on those heads is **not determined**. `pnpm lint` is an `&&` chain, so a failure at any earlier gate means the i18n gate never runs — and an unrun gate looks identical to a passing one. Only #5075's `Lint` was confirmed green.

Issue #5142, opened 2026-09-21
Adding a List to the documentation page


gDonia87:
### The problem

A list table could be an important document to share in a report. We don´t  have, right now, the possibility to create a short table with the tabular information of the models and then add it to a report

### Proposed solution

Adding a list in a documentation page with a block.

<img width="382" height="201" alt="Image" src="https://github.com/user-attachments/assets/b0e4bcc3-aa6a-4fd9-b13c-c169a7d402bf" />

louistrue:
what should table contain? could you give an example?

louistrue:
Thanks — agreed, and this is going in. Rather than a new table-authoring UI, the block will reuse **Lists**: a **Table** block in the Document panel holds a copy of a list (saved or preset — entity types, conditions, columns, grouping) and prints it the way the list's own export does: unit-aware cells, nested groups with subtotals or the schedule/pivot view, sorted.

Shape of it:
- *Add block → Table*, pick a saved list or preset; "Edit in Lists" and "Update from saved list" keep the round trip.
- 50 rows by default (one A4 page), up to 500; longer tables paginate with the header repeated and a `… n more rows` line; a caption under the table.
- The block re-runs on the loaded models whenever the document is shown or printed, so a template opened on the next revision reads that revision.

Landing as three PRs: the list-execution extraction, the block model + PDF composer, then the editor/preview.

gDonia87:
> what should table contain? could you give an example?

It is a list. It is only a reproduction of the created list in the dedicated tool. It contains what we build in the list tool

louistrue:
Shipped on `main` in three parts: #5146 (list execution extracted from the Lists panel), #5148 (the Table block + PDF composer) and #5163 (editor, live preview, list runs). **Analyze → Document → Add block → Table (from a list)** takes a copy of a saved list or preset and prints it the way the list's own export does — unit-converted cells, group rows with count and sums or the schedule view, 50 rows by default (up to 500), longer tables continuing across pages with the header repeated. *Edit in Lists* / *Update from saved list* keep the round trip with the Lists panel. Docs: `docs/guide/documents.md` (Blocks table) and `docs/guide/lists.md` ("Lists in documents"). It is live on ifclite.dev after the next nightly production advance.

Issue #4317, opened 2026-09-09
Lists panel swallows execution failures into console.error, so a rejected name pattern reads as an empty result


BIMvoice:
## Finding

The Lists panel swallows every execution failure into `console.error` and leaves the user with no indication anything went wrong.

`apps/viewer/src/components/viewer/lists/ListPanel.tsx`:

```ts
        setView('results');
      } catch (err) {
        console.error('[Lists] Execution failed:', err);
      } finally {
        setListExecuting(false);
      }
```

There is no error state set, nothing rendered, and no toast. The spinner stops and the panel simply does not advance to results. The same shape appears again a few lines down for definition import:

```ts
    } catch (err) {
      console.error('[Lists] Import failed:', err);
    }
```

## Why this is worth fixing now

It has become reachable in a new way. `compileNameMatcher` (`packages/lists/src/name-pattern.ts`) previously never threw — an invalid pattern was logged and fell back to an exact compare. #4262 adds a ReDoS guard that makes it **throw** on a catastrophic-backtracking shape or an over-256-character pattern. A user typing such a pattern into a Lists column now gets a silent no-op rather than a slow query.

The sibling surface already does this correctly, which is the useful contrast: the advanced-filter path in `SearchModal.filter.tsx` wraps its run in a try/catch that calls `setSearchFilterError(...)`, and renders a `FilterErrorBox` — the modal stays open and the user sees why. The Lists panel has no equivalent.

So the guard is right and the panel is the gap: a rejected pattern should read as "this pattern was rejected", not as "no results".

## The class

A caught error with no user-visible consequence is a false clean — indistinguishable from a genuine empty result. That is the same defect shape as several issues fixed this cycle (a clash run reporting clean while examining zero geometry, an aggregate fabricating `0` for a group with no data). Here it is in the UI layer rather than the engine.

## Suggested direction, not prescribed

Mirror what `SearchModal.filter.tsx` already does: an error state on the store, surfaced in the panel. `unsafeNamePatternReason()` in `packages/lists/src/name-pattern.ts` is non-throwing and already exists for exactly this purpose — the pattern-preview path uses it — so a column editor could reject a bad pattern before execution rather than only at run time.

Whether the import handler deserves the same treatment is a separate judgement; it is included above because it is the same shape in the same file, not because this issue proposes fixing both.

## Not checked

[...]

Issue #1573, opened 2026-07-03
Enhancement Proposal: IFC Property Unit Display & Conversion


Zamgo:
### The problem

Currently, values are displayed without units in certain properties, such as volumetric flow rate.

[VZT.ifc.txt](https://github.com/user-attachments/files/29633071/VZT.ifc.txt)

<img width="1188" height="423" alt="Image" src="https://github.com/user-attachments/assets/53fe9ae9-f19a-4197-9a59-340e77e04213" />

### Proposed solution

First Proposal: Add unit visibility here, similar to how it is already handled for area ($m^2$) in quantity sets. For instance, in my HVAC use case, this would apply to IfcVolumetricFlowRateMeasure.

Second Proposal: Introduce a tool to convert default IFC units to specific user needs. For example, the default unit for IfcVolumetricFlowRateMeasure is $m^3/s$ (which, to my best knowledge, is the only format Revit exports). However, in practice, I need $m^3/h$. It would be highly beneficial if IFClite allowed users to modify this. In this specific case, the conversion is just a simple multiplication by 3600 with some rounding.

Thank you for considering this solution :)

PR #5146, opened 2026-09-21
refactor(lists): extract federated list execution from ListPanel (#5142, 1/3)

louistrue:
Part 1 of 3 for #5142 (table block in the Document panel, fed by Lists). Spec: a Document table block embeds a copy of a list and runs it against the loaded federation when the document is shown or printed. For that it needs the list-execution path `ListPanel` kept inline — extracted here, no behaviour change.

## What moved

- `apps/viewer/src/lib/lists/run-list.ts` — pure `runListFederated(definition, pairs, state)`: `scopeModelPairs` → `executeList` per model → flatten rows → `summariseListRows` → `mergeResultColumns`. Throws exactly what its parts throw (the #4317 contract); the caller chooses where the message is shown.
- `apps/viewer/src/components/viewer/lists/useListProviders.ts` — the `modelProviderPairs` / `modelUnits` memos (zone context, world-coordinate getter, per-model VOLUMEUNIT scale), moved verbatim.
- `ListPanel.tsx` now uses both: 389 → 284 lines. `handleExecuteList` keeps its rAF + try/catch + `ListErrorBox` surfacing.

## Tests

`run-list.test.ts` (node:test, stub providers): rows of two models merge in order with `totalCount` summed; the `quantityType` that only `executeList` resolves survives onto the result columns while the authoring definition stays unannotated (#1573 P0); grouping spans models; an unmatched model-tag scope throws the `scopeModelPairs` reason instead of returning an empty result.

Existing `ListPanel.wiring.test.tsx` / `ListPanel.modelTagScope.test.tsx` pass untouched (62/62 in the lists suites locally), viewer `pnpm typecheck` (turbo) green, `check-module-size` OK, unused-locals count unchanged vs base.

Next: #2 block model + composer + PDF seam, #3 editor/preview/hook.


louistrue:
Labelled `revert-oracle-exempt`: this is a pure extraction (no behaviour change), so no assertion can go red when production is reverted — the oracle reports REVERT-BROKE-BUILD because `run-list.test.ts` imports the new module. Parts 2 and 3 of the stack are OBSERVED by the oracle locally (`document.test.ts` and `BlockEditor.i18n.test.tsx` fail by assertion on revert).

PR #5150, opened 2026-09-21
fix(geometry): skip an un-triangulable holed face instead of aborting the whole IfcAdvancedBrep

### LINKED ISSUE #5053: geometry: one un-triangulable holed face aborts the whole IfcAdvancedBrep, and the two router routes disagree about recording it
One holed face that earcut cannot triangulate now aborts the whole `IfcAdvancedBrep`, where before it degraded to a hole-filled face. The narrowing that causes it is correct; the problem is that one of the two callers of the shared helper does not catch.

## The change in blast radius

`processors/advanced_face/surfaces.rs`'s `triangulate_planar_indices` restricts the historical fallback to the no-hole case:

```rust
Err(_) if holes_2d.is_empty() => { /* outer fan */ Ok(idx) }
Err(error) => Err(Error::geometry(format!(
    "planar face triangulation with holes failed: {error}"
))),
```

That reasoning is right — an outer-only fan silently closes authored openings. Before #5026 this arm returned `Ok` unconditionally, so the failure could not occur at all.

The two callers of `process_advanced_face` then diverge. `processors/brep/surface_model.rs` catches per face and keeps the rest of the solid:

```rust
let (positions, indices) = match process_advanced_face(&face, decoder, quality) {
    Ok(result) => result,
    Err(ref e) => { trace_capped_face(face.id, e); continue; }
};
```

`processors/advanced.rs`, in the `IfcAdvancedBrep`/`IfcAdvancedBrepWithVoids` face loop, does not — it is a bare `?`:

```rust
let (positions, indices) = process_advanced_face(&face, decoder, quality)?;
```

so one face takes the whole solid, and it also skips the `empty_faces` diagnostic in the `else` arm just below, which exists to report exactly this kind of per-face loss.

## A second asymmetry, one level up

`router/processing.rs` has two routes to the same processor and they disagree about whether the failure is recorded. One catches:

```rust
match self.process_representation_item(item, decoder) {
    Ok(mesh) => { if !mesh.is_empty() { sub_meshes.add(item.id, mesh); } }
    Err(_e) => {
        self.record_unsupported_item(item.ifc_type);
        crate::diag::diag_debug!(... "skipping unsupported geometry item" ...)
```

while `process_element`'s own item loop propagates:

```rust
self.process_representation_item(&item, decoder)?
```

as does `voids/mod.rs`'s `let wall_mesh = self.process_element(element, decoder)?;`. So whether a failed face is counted as an unsupported item or silently propagates depends on which route the element took.

## Reachability

The precondition — an `IfcAdvancedBrep` containing an `IfcAdvancedFace` with an `IfcFaceOuterBound` plus at least one further `IfcFaceBound` — is present in the fixture corpus, resolved per face rather than by substring match:

| fixture | holed advanced faces |
|---|---|
| `buildingsmart/annex_e/advanced-geometric-shape/basin-advanced-brep.ifc` | 1 of 5 |
| `various/issue-604-door.ifc` | 8 of 104 |
| `ara3d/FM_ARC_DigitalHub.ifc` | 242 of 10070 |
| `ara3d/ISSUE_159_kleine_Wohnung_R22.ifc` | 322 of 7126 |

`IFCADVANCEDBREPWITHVOIDS` appears in none of the 116 fixtures searched.

## Not claimed

I have **not** shown that `triangulate_polygon_with_holes` actually fails on any of those faces — that needs a run I could not do here. So the trigger is unproven; what is shown is that the shape the new `Err` arm exists for is present in four real files, and that if it fires on the `IfcAdvancedBrep` path the loss is the whole solid rather than one face.

## Shape of a fix, if wanted

Catching per face in `advanced.rs` and continuing, as `surface_model.rs` already does, is index-safe — `base_idx` is recomputed per face from `(all_positions.len() / 3)`, with no fixed-size or face-count-keyed array.

But catch alone would trade one bad face for a silent one: once the processor swallows the error it always returns `Ok`, so the router-level `record_unsupported_item` never fires and the face vanishes from an otherwise-successful solid with no record. That is why `surface_model.rs` carries its own local trace. So the fix is catch, continue **and** record — the `empty_faces` list is the natural place, though it currently only records faces that meshed to zero positions, not faces that errored.

No branch for this one; I would rather you rule on the posture first, since "abort the solid" may be the intended answer for a brep whose faces cannot be trusted.

BIMvoice:
Closes #5053

Implements the posture you ruled on in the issue: *"degrade to a per-face skip"*.

## The change

`rust/geometry/src/processors/advanced.rs`'s `IfcAdvancedBrep` / `IfcAdvancedBrepWithVoids` face loop called `process_advanced_face(...)?` with a bare `?`, so one face that failed to triangulate took the whole solid — while `surface_model.rs` already caught per face and continued. The loop now catches, records, and continues, mirroring `FaceBasedSurfaceModelProcessor`.

The failure still reaches the existing `empty_faces` diagnostic in the `else` arm below, which the bare `?` used to skip — that diagnostic exists to report exactly this per-face loss, so a skipped face is not silently lost. The new push sits under the same `#[cfg(any(feature = "debug_geometry", feature = "observability"))]` gate as the declaration and the pre-existing push, so no new asymmetry is introduced.

`triangulate_planar_indices`'s narrowing is **untouched**. That narrowing is correct — an outer-only fan silently closes authored openings — and the defect was always in the caller.

## The router asymmetry: investigated, deliberately not changed

You noted the router-level split is "the same question one level up". It was investigated and left alone, for a concrete reason rather than caution:

`rust/geometry/src/router/structural_tests.rs:661-672` (`oriented_edge_cycle_fails_loudly_instead_of_recursing`) asserts the opposite contract:

```rust
let error = router.process_element(&entity, &mut decoder).unwrap_err();
assert!(error.to_string().contains("cycle"), "{error}");
```

`process_element` is relied on elsewhere to fail loudly rather than drop an item, so a blanket catch-and-continue there would break that guarantee. The two routes also return different shapes — `Result<Mesh>` versus `Result<SubMeshCollection>` — across 13+ call sites. That is a larger change deserving its own issue and its own ruling, so it is reported rather than bundled here.

## Test

`rust/geometry/src/processors/tests.rs` gains `advanced_brep_survives_one_untriangulable_holed_face`. Two `IfcAdvancedFace`s in one `IfcClosedShell`, the poisoned holed face listed **first** so a bare `?` fails before the good face is ever reached. It uses the repo's existing `1.0E400` → `f64::INFINITY` technique (same as the `#4206` test in `advanced_face/surfaces_tests.rs`) to deterministically hit the `holes_2d`-non-empty `Err` arm.

Mutation-tested in three directions, each an actual run:

| Mutation | Result | Evidence the mutated line ran |
|---|---|---|
| none (as committed) | GREEN, 1 passed | stdout carries `Skipping unsupported advanced face #14 ... non-finite coordinate in polygon ring` — the new catch arm executed, not merely compiled |
| restore the bare `?` | RED | panics with `one un-triangulable holed face must not abort the whole IfcAdvancedBrep (#5053)` |
| permissive: make the good face's merge a no-op | RED | panics with `the good face must still produce geometry after the bad face is skipped` — so the test checks geometry actually arrives, not just "no crash" |

The third row matters: a test that only asserted "no error" would pass against a silently-empty mesh.

## Verification

- `cargo test -p ifc-lite-geometry --lib` — **1048 passed, 0 failed, 1 ignored**
- `cargo test -p ifc-lite-geometry --lib advanced` — 28 passed, 0 failed
- `cargo check -p ifc-lite-geometry` — clean, 0 warnings
- `check-module-size.mjs` — no "grew past budget" and no new over-400 file for either touched file. The one failure block is the pre-existing `rust/processing/tests/module_size_allowlist.txt` merge-base audit, reproduced byte-for-byte on a pristine `upstream/main` worktree, so it is unrelated to this change.
- `check-changesets.mjs` — **could not run** (`ERR_MODULE_NOT_FOUND: @changesets/parse`). Stating that rather than implying it passed. The changeset was written against the format and package of the most recent merged Rust-geometry-only fix (`.changeset/keep-grazing-slivers-in-shard-sweep.md`, also `"@ifc-lite/wasm": patch`).

One scoping note in the interest of accuracy: the issue's fixture table established that holed advanced faces are *present* in the corpus; it did not show `triangulate_polygon_with_holes` failing on those specific faces. The regression test therefore constructs the failure deterministically rather than relying on a corpus model to trigger it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Advanced BREP solids now continue processing when individual faces cannot be triangulated, preserving geometry from valid faces.
  * Non-triangulatable faces are skipped without aborting the entire solid.
  * Diagnostics are recorded for skipped faces, improving visibility into incomplete geometry processing.

* **Tests**
  * Added regression coverage confirming valid faces remain available when another face cannot be triangulated.
  * Verified skipped faces do not contribute invalid geometry.


louistrue (changes requested):
Required before approval: replace the new raw  face-skip output with the existing feature-gated diagnostic mechanism ( / equivalent). A recovery path must not emit unsolicited stderr in normal builds. The full Rust suite itself passed; the current lint and E2E failures are repository-wide unrelated paths, but they still prevent a green merge until rerun/baseline recovery.

louistrue:
Required before approval: replace the new raw eprintln! face-skip output with the existing feature-gated diagnostic mechanism (diag_debug! or equivalent). A recovery path must not emit unsolicited stderr in normal builds. The full Rust suite itself passed; the current lint and E2E failures are repository-wide unrelated paths, but they still prevent a green merge until rerun/baseline recovery.

BIMvoice:
Fixed in `5d4fe84`. The `eprintln!` fallback is gone — the recovery path is now silent on stderr in normal builds:

```rust
crate::diag::diag_debug!(
    { face_id, error = %error, "skipping unsupported advanced face in advanced_brep" }
    else {}
);
```

Used the empty `else {}` rather than inventing a mechanism, since `diag.rs:60` already sanctions it: *"pass an empty `else {}` to compile out entirely"*. With `observability` ON it still emits `tracing::debug!`; with it OFF the legacy branch compiles away.

Re-verified after the change, and specifically checked that the regression test does not depend on the stderr line I removed. The earlier mutation evidence quoted that line, so the test could have been passing for the wrong reason. It isn't: its assertions are on `mesh.is_empty()`, `positions.len() == 12` and `indices.len() == 6`, with no stderr capture anywhere in `tests.rs`. Re-proved it independently by reverting the recovery behaviour itself — restoring the bare `?` so the error propagates — and the test failed on its own `.expect()` message. Restored, green again.

- `cargo test -p ifc-lite-geometry --lib` — 1048 passed, 0 failed, 1 ignored
- `cargo check -p ifc-lite-geometry` — clean
- `check-module-size.mjs` — the only failure is the `rust/processing/tests/module_size_allowlist.txt` merge-base audit, reproduced verbatim on a pristine `upstream/main` worktree at `e18d2d76f` (identical text, same 11 rows), so unrelated to this branch.

Thanks for merging `main` in — I checked the merge didn't quietly drop anything in either direction: our three files are the only diff against `main`, and `main` had no commits touching `advanced.rs` or `tests.rs` in that window.

---

**One thing worth your ruling, not fixed here.**

This helper was modelled on the pre-existing `trace_capped_face` in `brep/surface_model.rs:19-27`, which still has the exact shape you objected to:

```rust
fn trace_capped_face(face_id: u32, error: &Error) {
    crate::diag::diag_debug!(
        { face_id, error = %error, "skipping unsupported advanced face in surface model" }
        else {
            #[cfg(debug_assertions)]
            eprintln!("[ifc-lite] Skipping unsupported advanced face #{face_id} in surface model: {error}");
        }
    );
}
```

Same situation — a recovery path, skipping a face and continuing — so if unsolicited stderr is wrong in `advanced_brep`, it is equally wrong there. I've left it alone because you scoped this to the new code and I'd rather not widen the diff, but the two functions are now inconsistent: identical behaviour, different diagnostic policy.

Happy to either fold the one-line change into this PR or file it separately, whichever you prefer. Flagging it rather than deciding, since a quick `grep` suggests this `else { #[cfg(debug_assertions)] eprintln! }` shape may exist in other recovery paths too and you may want one consistent policy rather than a piecemeal fix.

louistrue (approved):
Re-reviewed 5d4fe8405. The raw fallback is removed, the recovery diagnostic uses the established mechanism, and required CI is green. Approved.

PR #5155, opened 2026-09-21
fix(lint): bash ${VAR:-} in vercel-sourcemaps.test.mjs is not a JS template (#5143)

louistrue:
main's **Lint** job is red since #5143 (run 35610113267): `scripts/vercel-sourcemaps.test.mjs:32` carries `${VITE_SOURCEMAP:-}` inside a `bash -c` string, and `no-template-curly-in-string` is in the error tier of `.oxlintrc.json`. It is bash parameter expansion, not a forgotten template literal, so the line gets an `eslint-disable-next-line` with the reason above it (the directive form the house rule asks for).

Local: `pnpm exec oxlint --config .oxlintrc.json scripts` → 0 errors; `node --test scripts/vercel-sourcemaps.test.mjs` passes. Blocks the gate for every open PR (incl. the #5142 stack: #5146 / #5148 / #5154).


louistrue:
Superseded by #5145 (merged), which rewrites the expression as `"$VITE_SOURCEMAP"` instead of disabling the rule, and adds the unknown-memory fail-safe.

Issue #5167, opened 2026-09-21
Flow graphs: keyed-data node-graph runtime, editor, connectors, and graphs-as-tools


louistrue:
Tracking issue for the flow-graph effort: a Dynamo/Grasshopper-class node graph over the SDK that runs identically in the viewer and in `ifc-lite flow run`, writes into the change set, and learns from IFCflow's failures (untyped ports, deep-copied elements, whole-file rewrites, full re-execution, server-side Python).

Design in one line: one keyed level (`Item / List / Group<T>`) plus typed `Table`s; ports declare `item | list | group` access and the scheduler lifts by **key**, never by position; write nodes track their output under a user-visible tracking key with deterministic GlobalIds and remove orphans.

Plan (rev 2, adversarially reviewed before coding): `~/ci/plans/PLAN-flow-graph-2026-09-21.md`.

### Phase 1 — core — **merged** (#5168)
- [x] `@ifc-lite/flow`: values, tables, digests, documents + validation + migrations, registry, lifting/lacing, scheduler + memo, availability, tracking
- [x] `@ifc-lite/flow-nodes`: core, model read, table, viewer (no-op headless), setProperty/setAttribute (pset-level capability check), Script (QuickJS)
- [x] `ifc-lite flow run|describe|validate` + docs (`docs/guide/flow.md`, CLI guide)
- [x] Fixture graph + browser-vs-headless parity + real-model CLI test with independent read-back
- [x] **Added during review**: `validateFlowWiring` — registry-aware validation (`parseFlowDocument` is registry-free, so a declared output naming a missing port validated clean and produced nothing at run time)

### Phase 2 — editing — **merged** (#5170, `0eafae1cb`)
- [x] Undo batch tagging in the viewer mutate adapter (`batchBegin` was a TODO) + `bim.mutate.batchAsync` (a call while one is in flight joins it; refused inside a sync `batch`); headless has no history by design, so no mirror
- [x] `GlobalId?` on every in-store builder and the `bim.store.add*` params + deterministic GUID helper — **except MCP tools, which do not expose it yet**
- [x] Tracking engine: `update / replace / disabled`, tracked-set store (CLI sidecar / browser), create-update-keep-remove, failed-lane retry, collision error
[...]

PR #5168, opened 2026-09-21
feat(flow): keyed-data graph runtime, standard nodes, and ifc-lite flow command (#5167)

louistrue:
Phase 1 of #5167.

## What
- **`@ifc-lite/flow`** — the runtime. Values are `Item / List / Group<T>` (one keyed level) and typed `Table`s. A port declares `item | list | group` access; the scheduler lifts a node over lists by lacing (`shortest | longest | cross`, guarded) and over groups **by key**, never by position, reporting unmatched keys. Memoised evaluation with a per-model revision and a write generation; structured run log with per-lane errors; `*.flow.json` documents with hand validation and a migration chain; element tracking that derives GlobalIds from a user-visible tracking key and removes vanished lanes.
- **`@ifc-lite/flow-nodes`** — `core.*`, `model.*` reads (selector, properties, quantities, relationships, storey/type grouping), `table.*` with IFC value types and pset/prop bindings, `viewer.*` that are no-ops headlessly and pass entities through, `model.setProperty / setAttribute` gated on the *actual* pset (`model.mutate:<Pset>`), and `script.run` in the QuickJS sandbox.
- **`ifc-lite flow run | describe | validate`** over `HeadlessBackend`, with `--input nodeId.param=value` overrides and `--out` for the model with the run's mutations.
- Docs: `docs/guide/flow.md`, CLI guide section, package READMEs; changesets for all three packages.

## Why this shape
Reviewed against Grasshopper data trees / Dynamo lacing + List@Level / RIR tracking and against IFCflow's own source: untyped ports that sniff `Array.isArray(input) ? … : input.elements`, deep-copied elements, whole-file rewrites, full re-execution. Here nothing is copied (entities are handles), every port is typed, and the restructuring set is total (`groupBy / flatten / keys / lookup / wrap / filter / pivot`) with no path arithmetic.

## Evidence
- `packages/cli/src/__fixtures__/flows/fire-rating-audit.flow.json` on `apps/viewer/public/samples/building-architecture.ifc`: 4 walls without `Pset_WallCommon.FireRating` found, colorize reported `noop` headlessly, default rating written. The CLI test reads the written IFC back through a **fresh** headless context (`where FireRating = REI90` → 4; original → 0).
- Script node runs in QuickJS headlessly (`bim.query.byType('IfcWall').length * inputs.a` → 8). Gotcha fixed: `bim.sandbox` dynamically imports `@ifc-lite/sandbox` relative to the SDK, which a pnpm-isolated install cannot resolve, so the node imports `createSandbox` directly and skips the browser-only esbuild-wasm transpile.
- Lifting tests cover the review's scenarios: openings-per-wall as a group keyed by wall, grid × level cross lanes keeping their keys when a grid line is inserted, key-matched groups with missing keys reported, null-lane short-circuit.
- Tracking tests: re-run keeps GUIDs, changed lane updates under the same GUID, vanished lane is removed, `replace` removes the previous set with deterministic fresh GUIDs, sidecar pinned to another model state is refused.

## Local gates
`turbo typecheck test` for `cli`, `flow`, `flow-nodes`: flow 48/48, flow-nodes 8/8, cli typecheck OK; cli tests 80 passed with the 3 known Windows-only false failures (`geometry-report` locale, `delivery` `<script>` path, `diff-content-inputs` symlink EPERM — identical on untouched `origin/main`). `check-module-size` OK, oxlint 0 errors, api-surface and generated docs regenerated, doc snippets hand-typechecked.

## Summary by CodeRabbit

* **New Features**
  * Added flow graphs for IFC data processing, including structured values, tables, model queries, viewer actions, controlled mutations, and sandboxed scripts.
  * Added standard nodes for arithmetic, filtering, grouping, table transformations, model reads, and model updates.
  * Added `ifc-lite flow run`, `describe`, and `validate` commands for headless execution, input overrides, validation, reporting, and optional IFC export.
  * Added deterministic element tracking, memoized evaluation, capability-aware availability, and headless-safe viewer behavior.

* **Documentation**
  * Added flow graph guides, CLI references, package documentation, and API listings.

---

**Follow-up commits on this PR** (all review findings from both bots):
- `ee6aec1d4` — eight defects fixed with tests (capability gate on the Script node, cached sandbox rejection, duplicate lane keys dropping outputs, pivot key-column collision, `maxCross: NaN`, `validate --json` exiting 0, exporting a failed run, unchecked `--out`/`--input` operands).
- `8f42129fc` — `validateFlowWiring`: `parseFlowDocument` is registry-free, so a declared output naming a port no node has validated clean and produced nothing at run time. `flow validate` now also reports unknown node types, edge endpoints naming missing ports, type-incompatible edges, and unconnected required inputs, and exits 2.
- The fixture moved to `packages/cli/src/__fixtures__/flows/` so it can trigger the lane that runs it (`check-ci-path-coverage`); `sdk-canary.yml` `paths:` carries the two new packages.


louistrue:
Applied `revert-oracle-exempt`: the PR is new modules end to end (`packages/flow`, `packages/flow-nodes`, `commands/flow.ts`), so a whole-file revert deletes the seams the tests import through (`Cannot find module './flow.js'`) — the oracle's own REVERT-BROKE-BUILD case, same as #5144/#5157. The Windows-only failures and the 401-line `cli/src/index.ts` ratchet hit are fixed in the follow-up commit.

louistrue:
All review findings from both bots are fixed in `ee6aec1d4`, each with a test:

| Finding | What was actually wrong | Fix |
|---|---|---|
| `script-node.ts` capability | The node declares `model.read` but never called `requireCapability`, so a graph granted nothing still got a sandbox with `query`+`model` on — user code could read the model through a capability the graph never had. | Gate added; test runs a Script node with only `viewer.read` granted and expects `CapabilityDenied`. |
| Cached sandbox rejection | One transient wasm-load failure stayed in the per-context cache forever, so every later lane awaited the same rejection. | The pending promise removes itself on rejection; test makes `createSandbox` fail once and asserts the retry builds a new sandbox. |
| `assemble()` duplicate lane keys | Two lanes sharing a key (a driving list repeating an entity) **replaced** instead of appended — and only for list/group ports, so the loss depended on port access. | Every access appends; test drives `[W1, W1]` and asserts both outputs survive. |
| `pivot()` key-column collision | A long-format row whose pivot column is literally `GlobalId` overwrote the row's identity with its value, merging two distinct things. | Counted as `collisions`, skipped, and warned in the run log. |
| `maxCross: NaN` | `NaN < 1` is false, so it passed validation and then disabled the cross guard entirely (`size > NaN` is always false). | `Number.isFinite` first; test covers NaN, Infinity, 0, -1. |
| `flow validate --json` | Printed `ok: false` and exited **0**, so CI read an unrunnable graph as valid. | Report first, then exit 2 in both formats; test asserts exit 2 with `ok:false` in the JSON. |
| `flow run --out` on a failed run | Write nodes can run before a later node fails; the export wrote that half-applied state behind a green-looking file. | Export only when `result.ok`; `out` is reported `null` otherwise. Test: a graph that writes then divides by zero leaves no file. |
| `--out` / `--input` / `--tracking` operands | `--out --json` wrote a file named `--json`; `--input rating=REI90` (no node id) was silently dropped, so the graph ran on its **defaults** and reported success — in the audit flow that writes the wrong FireRating. | Operands are checked; unknown `--input` keys are refused up front and name the declared keys. |
| GlobalId index rebuild-on-miss | A tracked create node asks for every GlobalId it is about to mint, so a miss is the expected case and each one triggered a full model scan. | Explicit `rememberGlobalId` / `forgetGlobalId` / `invalidateGlobalIdIndex`; no rebuild on miss. |

**One finding I am not acting on**: CodeRabbit's `packages/flow/tsconfig.json` — "overrides the inherited `include` and re-enables `src/*.test.ts`, the build emits them into `dist`". `exclude` filters `include` in TypeScript, and the root `tsconfig.json` excludes `**/*.test.ts`; `ls packages/flow/dist | grep -c test` is **0**. The file is byte-identical in shape to `packages/regex-guard/tsconfig.json`, the house pattern.

Local: flow 51/51, flow-nodes 10/10, cli 7/7 flow tests (full CLI suite green apart from the 3 known Windows-only failures), typecheck-tests OK for all three, api-surface / module-size / pset-lookup / sdk-canary / i18n gates clean.

Issue #4946, opened 2026-09-18
Filtering the source of a single chart


gDonia87:
### The problem

The charts read the whole amount of objects present in the model. The analysis is not so granular.

### Proposed solution

Having a filter for the source data for every single chart, so that we can create an analysis for a specific amount of components or a specific model.

### Alternatives considered

It could also be useful, if we used as source of the chart a created list.
There we would have the possibility to filter what we want and then we can reuse this table for the chart.
Any other ideas?

gDonia87:
@louistrue 
Thanks for the implementation. It works correctly.
I have another idea about this implementation.
Would it be possible to have the function "add rule" also in the source filter (selector)?
The problem is that I would like to use the model tag option or other predefined flter rule options and it is difficult to find the right combination in the Source filter selector.

<img width="886" height="112" alt="Image" src="https://github.com/user-attachments/assets/2cb2413e-0584-4995-b49f-643aa5bdd7c2" />

Another option could be useful, if we could reconvert the rule from the advance filter into a selector, so that I can use it in the source filter of the charts without doubts of its correct structure.

This is the opposite of what this button does.

<img width="896" height="378" alt="Image" src="https://github.com/user-attachments/assets/1624ab68-42ec-4fa8-acd7-a912c20f180e" />

PR #5176, opened 2026-09-22
feat(charts): filter a clash chart to one detection rule

### LINKED ISSUE #5156: Choosing one clash test for populating charts
### The problem

I can't choose one of the run clash detection check, when I select this tool as source of my chart. All checks are counted in the chart

### Proposed solution

Having the possibility for creating a chart with only the information contained in one clash test

### Alternatives considered

_No response_

BIMvoice:
Closes #5156

A clash-sourced chart counted every detection rule together, with no way to chart just one. This adds that.

## What it turned out to be

No new "run/check" concept was needed. A clash run is one `ClashResult` containing several `ClashRule`s, each `Clash.rule` is a `ClashRule.id`, and `buildClashDataset` already writes that id into every row's `Rule` column. So "one clash test" is a rule id that is *already on the data* — the work was exposing it as a filter, not modelling it.

## The change

`ChartSourceFilter` (the per-chart filter #4946 built) gains an optional field:

```ts
clashRule?: string;
```

Absent means every rule, exactly as before this field existed. A rule id restricts to that rule. **Never an empty string** — an empty `clashRule` is indistinguishable from absent at every reader, which is the sentinel pattern rejected in review on #5151.

`applyClashRuleFilter(dataset, ruleId)` in `source-filter.ts` is synchronous; unlike the selector filter it needs no federation scan. It is applied in `ChartCard.tsx` and `useDocumentData.ts` alongside the existing `applyChartFilter`, and composes with it as AND.

The editor gains a "Clash rule" select, shown only for `source: 'clash'`, fed from `clashResult.rulesRun` — the same id/name pairs `ClashPanel.tsx` already uses for its "By rule" labels, so the names match what the user sees in the clash panel.

## Validation: one relaxation, one tightening

`filter.selector` previously had to be non-empty whenever `filter` existed. A chart can now filter by `clashRule` alone, so an empty selector is permitted — but only in that sense:

- a non-string selector is still rejected;
- a **whitespace-only** selector is still rejected, preserving the earlier review finding that `"   "` resolves to "no filter" at every consumer (they all trim) and strands the chart on "Resolving filter…" forever;
- `clashRule` must be a non-empty string when present, and is rejected outright on a non-`clash` source;
- **new:** a filter with neither a meaningful selector nor a `clashRule` is now rejected rather than silently accepted as an inert object.

Net effect on existing saved dashboards is nil: `filter: { selector: '' }` was invalid before (empty selector) and is still invalid now (nothing else to act on).

## Tests, and what has not been run

Added to `datasets.test.ts`, following the existing `applyChartFilter` tests:

- filtering by each rule id narrows to that rule's rows with a distinct fingerprint;
- an **unknown rule id yields 0 rows** and never falls back to "all" — the empty-vs-absent distinction, asserted rather than assumed;
- composes as AND with `applyChartFilter`;
- a dataset with no `Rule` column (`elements`) passes through unchanged;
- validation: `clashRule` alone passes, both together pass, neither is rejected, empty-string `clashRule` is rejected, `clashRule` on a non-clash source is rejected.

**These tests have not been executed.** `packages/charts` imports `@ifc-lite/parser`, which pulls the wasm build chain, and no workspace package has a `dist/` in this environment. The mutation analysis behind them is reasoning, not an observed RED/GREEN. CI is their first real run — saying so explicitly, because an unrun gate is indistinguishable from a passing one.

Gates that did run: `check-i18n-literals` OK (392 files, 0 literals), `check-changesets` OK, `check-module-size --base upstream/main` OK (0 new over budget; touched files are 99–314 lines). `check-unused-locals` could **not** run — every package, including ones untouched here, reports "does not compile standalone" because nothing has a `dist/`; that is environment-wide and pre-existing, not caused by this branch.

Changeset included: `@ifc-lite/charts` is published (v0.4.0, not private) and gained an additive optional field, so `minor`.

## Summary by CodeRabbit

- **New Features**
  - Added filtering of clash charts by a specific detection rule or run.
  - Added a “Clash rule” selector in the chart editor, including an “All rules” option.
  - Selected clash rules now appear in chart subtitles.
  - Clash-rule filters can be combined with selector filters.

- **Bug Fixes**
  - Added validation to prevent clash-rule filters on non-clash charts and reject invalid rule values.
  - Preserved existing behavior when no specific clash rule is selected.


louistrue (changes requested):
Holding this one on a single point, and it is the point you chose to make in the description.

You write, as an explicit promise:

> a **whitespace-only** selector is still rejected, preserving the earlier review finding that `"   "` resolves to "no filter" at every consumer (they all trim) and strands the chart on "Resolving filter…" forever

and the code comment at `packages/charts/src/validate.ts:116-124` says the same — *"Reported once, at `.selector`, whether the string is empty or whitespace-only."*

The condition does not do that:

```ts
} else if (chart.filter.selector.trim().length === 0 && !hasClashRule) {
```

For `{ selector: '   ', clashRule: 'str' }`, `hasClashRule` is true, so the whitespace-only selector is accepted. Both review bots found this independently.

**I do not think the resulting behaviour is wrong** — a whitespace selector alongside a real `clashRule` trims to "no filter" and the chart narrows by rule alone, which is fine and not the stranding case your finding was about. So this is not a correctness hold. It is a hold because the description and the comment both assert a rule the code does not implement, and in this repo that is the thing I cannot let through: the next person reads the comment, not the condition.

Two ways out, your call:

1. **Implement what you wrote** — reject a selector that is non-empty but whitespace-only regardless of `clashRule`, keeping the `&& !hasClashRule` relaxation for the genuinely empty `''` that `ChartEditor` writes. This is the one I would pick: `''` and `'   '` arrive from different places (one is the editor's unset sentinel, one is a human or an import), and treating them differently is defensible where treating them identically is what the old code did.
2. **Correct the prose and the comment** to say that a whitespace-only selector is accepted when `clashRule` narrows the chart, and say why it is harmless.

Either is a small diff. I would rather have (1) plus the test that pins it.

Separately, noted and not blocking: the tests were not executed locally and you said so, which is the right thing to do — CI is green on them now. And the `GID(41)`/`GID(42)` point on `datasets.test.ts:729` is worth taking while you are in here: as written, a no-op `applyChartFilter` still passes both assertions.

BIMvoice:
Took option (1). Pushed as `b7c0cd4c037091bc179482ff0c80ce7ebf5b14b4` on top of `84abb427d` — no rebase, no force-push.

You were right that the prose was the problem rather than the behaviour, and right that the two empty-ish values deserve different treatment. `packages/charts/src/validate.ts`:

```ts
} else if (chart.filter.selector.length === 0 ? !hasClashRule : chart.filter.selector.trim().length === 0) {
```

`''` keeps the `clashRule` relaxation; any other whitespace-only string is rejected unconditionally.

The comment now says what the condition does, including the reason for the asymmetry rather than just the rule:

> `selector` is a required STRING on the type but, since #5156, may be the EMPTY string when `clashRule` narrows the chart instead — a filter can now narrow by `clashRule` alone, with no selector text at all. `ChartEditor` always writes the key (possibly `''`) rather than omitting it, so that exact sentinel is relaxed when `clashRule` is present. A selector that is non-empty but WHITESPACE-only is still rejected regardless of `clashRule`: `''` and `'   '` arrive from different places — the editor's unset sentinel vs. a human or an import — so only the former gets the relaxation. Whitespace would otherwise trim to "no filter" at every consumer — they all trim — stranding the chart on "Resolving filter…" forever with no entry to find; the original review finding this preserves.

**Pinned, with the two halves independently exercised:**

| mutation | observed |
|---|---|
| revert to `&& !hasClashRule` | RED — the `{selector:'   ', clashRule:'x'}` pin fails, `0 !== 1` |
| reject `''` too (bare `trim().length === 0`) | RED on two *different* assertions — the clashRule-alone case and the editor-sentinel case — while the whitespace-with-rule assertion stays green |

Disjoint failures, so neither half is riding on the other's coverage.

**And the `GID(41)` fixture — you were right, it was vacuous.** The GID was present on both clash rows, so the rule filter alone determined the row counts and the assertions passed whatever `applyChartFilter` did. Switched to `GID(42)`, present on only the `str` row, and proved it the direct way: with `applyChartFilter` stubbed to `return dataset;` the suite now goes **21 pass / 3 fail**, including the new assertion at `734:12` (`1 !== 0`). Reverting the stub returns it to 24/24.

That check is worth more than the fix — the old fixture would have kept passing through any future regression in selector composition.

Tests: `datasets.test.ts` 24/24 under `tsx --test`; `packages/charts` 39/39 under vitest. Gates: module-size, changesets and test-glob coverage all OK; `tsc --noEmit` clean on `packages/charts`.

One caveat I'd rather state than paper over: the `apps/viewer` `tsc --noEmit` reports 327 pre-existing errors, all from unrelated packages whose `dist/` had to be sourced from another local worktree to make the wasm-dependent test chain resolvable without a full build. Zero of them are in any charts file — `validate.ts`, `source-filter.ts`, `datasets.test.ts`, `ChartCard.tsx` and `ChartEditor.tsx` each grep to zero hits. Environment workaround, not a code change.

**Description correction**, since the claim in the PR body was the thing you held on — please treat this as the amended text:

> The original condition let a whitespace-only selector through whenever `clashRule` was present, contradicting the promise above and the code comment. Fixed in a follow-up commit so `''` (the editor's unset sentinel) keeps the `clashRule` relaxation, while any other whitespace-only string is rejected unconditionally — matching what was originally described. Also fixed the `datasets.test.ts` selector-composition fixture flagged in review: it selected a GID present on both clash rows, so the assertions passed even with a no-op `applyChartFilter`; switched to a GID present on only one rule's row so the composition is actually exercised.

BIMvoice:
Conflict resolved — `4d0d2a69be59f00e16e2aea9dbe38820f47520c6`. GitHub now reports `MERGEABLE` again.

The conflict appeared because **our own merges caused it**: #5151 and #5153 landing reshaped `ChartEditor.tsx` underneath this branch. Resolved by **merging `main` in, not rebasing** — a force-push on a PR you are mid-review on would throw away your review anchors.

**The single conflicting hunk** was the form's `onSubmit` handler (~lines 218-233). Everything else in the file auto-merged clean — the clash-rule `<select>`, the `elementCount` handling, `draftToSpec`/`specToDraft`, the i18n calls.

- Our side built `{ selector, clashRule }` and called `onSave({ ...draft, … })` directly.
- Main's side (from #5151) called `onSave(draftToSpec({ ...draft, … }))`, with no `clashRule`.

Kept both: the resolved code computes `trimmedSelector`/`clashRule`/`filter` exactly as this branch did, then passes the result through `draftToSpec` — which is #5151's addition and the only place `elementCount`'s `dimension`/`measure` sentinel resolution happens. Dropping either side would have silently broken one of the two features.

**Both things you asked for survived, verified rather than assumed:**

`validate.ts` auto-merged with no conflict, and the condition is intact at `:127`:

```ts
} else if (chart.filter.selector.length === 0 ? !hasClashRule : chart.filter.selector.trim().length === 0) {
```

The `GID(42)` fixture also survived — and I re-proved it is still non-vacuous rather than trusting that it looks right. Stubbing `applyChartFilter` to `return dataset;` turns **3 subtests red**, including the `arcAndBeam.rows.length` assertion at `datasets.test.ts:734`. Restored afterwards, working tree clean.

That re-check was the point of the exercise: a merge is exactly the moment a fixture quietly reverts to the version that passes either way.

Tests, with the runner named because it matters here:
- `packages/charts` — `pnpm exec vitest run` — 44 tests across 3 files, all pass
- `apps/viewer` — **`tsx --test`, not vitest** — `datasets.test.ts` 24 tests, all pass, including the `#5156` clashRule validation and the `#5176` whitespace-rejection test by name

`mergeStateStatus` is `BLOCKED`, which is the ordinary awaiting-review state rather than a conflict — `mergeable` flipping `CONFLICTING` → `MERGEABLE` is the signal that the resolution took.

Nothing else changed. Over to you.

louistrue (changes requested):
Not ready to merge yet. Two small but real filter-state defects remain on 4d0d2a69:

1. `clashRule: "   "` is accepted because `hasClashRule` tests length rather than trimmed length. It produces a saved chart that silently filters to nothing. Reject whitespace-only rule IDs and add the validation pin.
2. `clashRuleLabel ?? clashRule` preserves an empty valid rule name, then `subtitleFor` treats it as falsy. The card omits the indicator that a rule filter is active. Fall back to the rule ID for an empty label, and cover it.

All required CI is green; these two corrections plus their tests should make this ready for re-review.

louistrue (changes requested):
The validation fix is exactly right, and the comment is better than the code change — it now states the asymmetry *and* its reason (`''` is the editor's unset sentinel, `'   '` came from a human or an import), so the next reader knows why two empty-looking values are treated differently instead of guessing it was sloppiness. That was the whole point of asking.

Resolving the conflict by **merging `main` in rather than rebasing**, explicitly to avoid throwing away review anchors on a PR under review, is the correct instinct and I want it noted. A force-push here would have cost more than it saved.

**One thing left, and it's one character.** `ChartCard.tsx:173`:

```ts
const subtitle = subtitleFor(spec, aggregation, filterSelector, filterState, clashRuleLabel ?? clashRule);
```

`clashRuleLabel` is `rulesRun.find(r => r.id === clashRule)?.name`, and `ClashRule.name` is a required `string` — which `''` satisfies. For a rule with an empty name, `??` passes `''` through (it only guards null/undefined), and `subtitleFor:67` does `clashRuleLabel ? \`rule: ${clashRuleLabel}\` : null` — so the rule indicator vanishes and **the user sees no sign that a clash-rule filter is active on the chart**.

Make it `||`. The intent is "fall back to the id when there's no usable label", and `||` is what says that; `??` says "when the label is absent", which is a different and weaker claim.

I'm asking for it rather than waving it through because it's the same class as the finding that started this review — a value that is present-but-empty reading as "nothing to show" — in code this PR introduces. Not worth a debate; worth the character.

Push that and I'll approve.

BIMvoice:
Done — `45ebb5b90`. `clashRuleLabel || clashRule`.

You were right that it's a different claim, and it turned out to be observable rather than stylistic: I pinned the case that distinguishes the two operators, a clash rule whose `name` is `''`. With `??` the subtitle read `'1 bucket · 1 element'` — the empty label passed through and the rule indicator vanished entirely. With `||` it reads `'… · rule: str'`.

So the weaker claim was also silently dropping the rule from the subtitle whenever a rule had a blank name. Test added rather than just the character, since a one-character change with no pin would be indistinguishable from a no-op to the next reader.

`tsx --test`, 55/55 in that file. Module-size, changesets, test-glob coverage and `tsc --noEmit` all clean.

Apologies for the delay on this one — your review sat unread because I was checking `reviewDecision` rather than the inline threads, which is the exact trap I've been warning about elsewhere.

louistrue (approved):
The requested empty-name fallback is fixed and pinned. I also corrected the missing-rule empty state on d21d351fc: a saved rule ID that no longer appears in a nonempty clash run now says no rows match the filter. Root pnpm typecheck passed, and the focused root pnpm test viewer shard containing ChartsPanel.test.tsx passed 181/181. Awaiting full CI and merge queue.

Issue #5138, opened 2026-09-21
A new more comprehensive Data Validation Tool


gDonia87:
### The problem

Today, we only have the possibility of checking the information in the models with an IDS-file or with a visual check of the information. or with the information palette. or with the lists.
As BIM-Coordinators we need something more flexible and rule-based.

### Proposed solution

Creating a brand new Tab "Data Validation" or "Information checker" which can join the capability of opening an IDS File with a dedicate button, or with another button we can create rule-based checks.
For example:
I imagine the new palette with two buttoms at he top, which could be implemented with other checks options in the future, one is IDS Validation and the other one Information Validation.
If I use IDS, I can open an external file and I perform a normal check, as it works now.
If I want to check specific information I can open the second button.
There I can write rule-based functions with the contained properties and entities in the loaded models.
I can choose one of them, or I can write freely using selector syntax rules. I can choose the model to be checked and I decide whether this property should:
equal;
not equal;
greater than;
at least;
less than;
at most;
contain;
exists;
between; 

When I create the sets of rules, I can get the result similar to the IDS check results:

<img width="595" height="720" alt="Image" src="https://github.com/user-attachments/assets/799c4d7b-38d5-4028-94fb-922dfa621611" />

I can navigate, in this case, the 3 rules and analyse the results.
At the end of this workflow, I can report these results as a List in my Documentation page or I can create charts. 

Of course, these rule-based checks should be saved as an external file and reused in different sessions.

louistrue:
Thanks — this is a good one, and it lines up with what buildingSMART itself lists as out of scope for IDS 1.0 (`Documentation/UserManual/specifications.md` § Limitations, plus #397/#398 uniqueness, #320 value comparison, #329 OR logic, #177/#323 exceptions, #346 case sensitivity, #379 conditional partOf in the IDS repo). Scoping it as follows; implementation starts now as a series of PRs.

**Shape**
- The `IDS validation` panel becomes **Data validation** with two entries: *IDS validation* (unchanged, still one model per run, still with audit + auto-correct) and *Information validation* (rule sets).
- Rules are authored either with the same chip builder as the Filter tab or as free text in the IfcOpenShell selector syntax the viewer already parses. Pset/property names are discovered from the loaded models; free text is accepted for names that are not present.
- Each rule = **applicability** (which elements) + **requirement**. Requirement kinds:
[...]

PR #5178, opened 2026-09-22
feat(viewer): add an IDS validation report block to the Documentation tab

### LINKED ISSUE #5125: Adding an IDS Report to a document for exporting
### The problem

We don´t have the possibility of including an IDS Report in the documentation tab.

### Proposed solution

Having the possibility of reporting an IDS Validation to a documentation page.

The strucure of the added IDS Validation should not be so different from the Report that we already have in the tab.

A resume´ of the checked, passed and failed objects.
A percentage of the failed/passed objects

List of the different checks with short and long description, number of the total check entities, passed and failed entities.
A child list with the different rules with the same informations of as the parent check.

Image

### Alternatives considered

_No response_

BIMvoice:
Closes #5125

Adds an **IDS validation report** block to the Documentation tab, so an IDS run can be included in an exported document.

## Modelled on the table block, as the issue asked

The issue says the structure "should not be so different from the Report that we already have in the tab", so this follows the table block (#5142) recipe rather than inventing one: a `DocumentBlock` variant holding a frozen snapshot, its own `compose-*.ts` layout module, a `*Preview.tsx` screen component, an editor block, and structural validation.

That extension point was anticipated — `types.ts:78` on `main` already earmarks `TableSource`'s discriminator as "where a validation-results table (#5138) plugs in".

The result model is reused, not recomputed: `ValidationReport`/`SpecificationResult` from `packages/ids/src/report-types.ts`, with `calculateSummary` from the ids package doing the pass/fail maths.

## What is in, and what is deliberately not

**In:** the summary (checked / passed / failed / pass rate, with the zero-checked case handled) and the **top-level check list** — one row per specification with id, short description (`name`), long description (`description`), checked/passed/failed and pass rate. Rendered in both the on-screen preview and the PDF export, added via an "IDS validation report" block with a "Refresh from current validation report" action.

**Not in: the nested per-rule child list the issue also asks for.** Two reasons, and I'd rather state them than ship a rollup with a silent caveat:

1. `RequirementSummary` carries only `id` / `label` / `optionality` — there is no separate short *and* long description per rule to populate those columns with. I did not invent the field.
2. Per-rule checked/passed/failed counts can only be aggregated from `entityResults`, which is complete only when the report was built with `includePassingEntities: true`. That holds for the viewer's own IDS panel run, but not for every producer — a failures-only run would silently undercount.

So the child list needs a small model change first. Happy to do that as a follow-up if you want it; it seemed wrong to fake it here.

## Tests, and how they were run

The viewer suite could not run: the dataset registry pulls `@ifc-lite/clash` → `@ifc-lite/geometry`/`@ifc-lite/wasm`, which needs `cargo`/`wasm-pack`.

So rather than assert coverage that was never observed, the real production functions were executed directly via `tsx` — `validateDocumentSpec`, `migrateDocumentSpec`, `composeDocument`/`layoutIdsReport`, and `idsReportBlockFromReport` against the built `@ifc-lite/ids`:

- zero checked entities validates and composes with no `NaN`/`Infinity` ever printed, and `passRate` is `100`, not `NaN`
- a `NaN`, out-of-range `passRate` or negative count is **rejected** by the structural validator
- a v3 document migrates to v4
- 60 checks paginate across 3 pages with every check printed exactly once
- a specification with no `description` carries `longDescription: undefined` rather than an invented value

**Mutation table — every row executed, reverted, and re-run, not reasoned:**

| Mutation | Result |
|---|---|
| `passRate <= 100` → `<= 1000` | red: out-of-range test failed |
| `count >= 0` removed | red: negative-count test failed |
| zero-checks branch disabled | red: empty-checks-line test failed |
| `passed`/`failed` swapped in the mapping | red: per-check-counts test failed |

## Notes for review

- `DOCUMENT_VERSION` 3 → 4 with the migration extended to accept v3. The bump is deliberate per the existing comment: an older viewer should refuse a file containing a block it cannot print rather than silently dropping it.
- Two existing tests asserted the document version literally and were updated; six call sites gained the new `idsValidationReport` prop.
- PDF text in `compose-ids-report.ts` is hardcoded English, matching this module's documented precedent — `generate-document-pdf.ts:58`: *"The English the PDF prints for a table block's rows, like every other string this module prints."* The preview component does go through the catalogue. The new plural key counts **checks**, which is a different quantity from the checked/passed/failed entity counts, so no single selector is doing double duty.
- No changeset: every touched file is under `apps/viewer`, which is `"private": true`.

Gates: `check-module-size --base upstream/main` OK (0 new over 400, `+0 ^0 v0 -0`), `check-i18n-literals` OK, `check-changesets` OK. `check-unused-locals` **not run** — it reports "these packages do not compile standalone" in an unbuilt worktree, which is environment-wide and not specific to this branch. `tsc --noEmit -p apps/viewer` (after building 14 leaf packages by hand) shows **zero errors in any file touched or added here**; the remainder are pre-existing in untouched files.

## Summary by CodeRabbit

- **New Features**
  - Added IDS validation report blocks to documents.
  - Add reports from the document block menu and refresh them when a live validation report is available.
  - View report summaries, pass rates, check results, descriptions, and source information in document previews and PDFs.
  - Added localized labels, status messages, and empty-state text for IDS reports.

- **Bug Fixes**
  - Improved handling and validation of IDS report blocks in saved documents.


BIMvoice:
**Correction to my own description above, plus the first real execution of this branch's tests.**

The tests now run, using the `@ifc-lite/wasm@9.2.0` artifact already pinned in `packages/wasm/package.json`, fetched from npm as local tooling (nothing git-tracked touched):

- `apps/viewer/src/lib/document/document.test.ts` — **22/22 pass**
- `apps/viewer/src/components/viewer/document/BlockEditor.i18n.test.tsx` — **7/7 pass**

Two things worth correcting.

**1. CI has never run either file.** Viewer test shards all report `pass`, but grepping all 8 shard logs for these two filenames returns zero hits. So the shard-level green says nothing about them — the same "a gate that passes while examining nothing" shape that has come up repeatedly. They pass, but until now nothing had observed that.

**2. My mutation table above overstated what is covered.** I wrote it as "every row executed, reverted, and re-run", and that was true — but it was executed against a throwaway `tsx` script driving the real functions, **not** against committed tests. The distinction matters and I blurred it: that proves the code was correct when written; it does not pin the behaviour against regression.

`idsReportBlockFromReport` is referenced only by `BlockEditor.tsx`, `DocumentPanel.tsx` and its own definition — **no test file calls it anywhere in the repo.** So:

- swapping `passed`/`failed` in the mapping (`ids-report.ts:37-38`) would turn no assertion red
- breaking the zero-checked-entities guard would turn no assertion red

`document.test.ts` never constructs an `'ids-report'` block, and `BlockEditor.i18n.test.tsx` only asserts label strings for a report block with no report loaded. The two rows of my table that claimed to exercise the mapping describe behaviour that is real but untested.

The `types.ts` validator rows (`passRate <= 100`, `count >= 0`) are genuinely covered by `document.test.ts` and do hold.

So the honest state of this PR: the composition, pagination, validation and migration paths are tested and green; the **snapshot mapping from `ValidationReport` to the block is not**. I'm adding that coverage rather than leaving the claim standing — will push it here shortly.

Apologies for the imprecision; "executed" should have said "executed against an ad-hoc harness, not pinned by the suite".

BIMvoice:
Coverage gap closed: `1bad10a`, a new `apps/viewer/src/lib/document/ids-report.test.ts`. `idsReportBlockFromReport` now has tests — it previously had none anywhere in the repo, which is what made my earlier mutation claim overstated.

Five tests, and the fixtures are deliberately asymmetric so a regression cannot hide in a coincidence:

- `passedCount: 3, failedCount: 7` — never equal, so swapping the fields is observable.
- One spec with a `description`, one without, asserting `longDescription === undefined` and explicitly `!== ''`.
- The two-spec summary totals 8 passed / 7 failed / 15 checked → `floor(8/15*100) = 53`, which is distinct from either spec's own rate (30, 100) **and** from their average. A summation bug can't land on 53 by accident.
- Zero checked entities asserts `passRate === 100` and `Number.isNaN(...) === false`.

**Mutation table — executed, each restored and re-verified:**

| mutation | test reddened | result |
|---|---|---|
| swap `passed`/`failed` in `ids-report.ts:37-38` | per-check mapping | RED, 4/5 |
| remove the zero-checked ternary in `calculateSummary` (yields `NaN`) | zero-checked | RED, 4/5 |
| `longDescription: … ?? ''` | missing-description | RED, 4/5 |

Each mutation reddens exactly one test, which is the property I wanted — they pin independent behaviours rather than all riding on one change.

Real counts, all executed: `ids-report.test.ts` 5/5, `document.test.ts` 22/22, `BlockEditor.i18n.test.tsx` 7/7.

`check-test-glob-coverage` reports OK (54 packages, 0 unrun test files), so the new file is genuinely reached by the viewer's test glob — worth confirming explicitly, given CI had never run the other two files in this PR despite the shards reporting green.

`check-module-size --base upstream/main` clean, `check-i18n-literals` OK (393 files, 0 literals). `check-unused-locals` still reports `apps/viewer` as not compiling standalone — pre-existing, with a baseline of 238 exceptions already recorded, unrelated to this file.

That settles the correction above: the mapping is now pinned by the suite, not just verified once by hand.

louistrue:
Merge-train pass: this one is out because it is **conflicting against `main`** (`mergeable: CONFLICTING`), so nothing here can be evaluated on its current head — CI last ran against a base that has since moved.

Please rebase onto `main` and push. Note that #5153 and #5151 are in the merge queue ahead of you and both touch viewer components, so rebase after those land rather than before, or you will do it twice.

No review of the substance yet; I will look once it is mergeable.

BIMvoice:
Conflict resolved — `2aef9649bc69ad7ead455744866120734ab06e8f`, now `MERGEABLE`. Merged `main` in; no rebase, no force-push.

**But the conflict was the smaller question. I checked first whether this PR is still wanted at all**, because #5166's commit message says it ports validation results into the table-block design *"instead of beside it"* — and this PR builds a separate block. That reads like two designs for one feature, and resolving a conflict on work that should be closed would be worse than leaving it conflicted.

**It isn't redundant. The two are complementary, and the evidence is concrete.**

#5166's `resolveValidationTableState` emits **one row per entity or per set**, and its `TableColumnId` union is a fixed eleven ids:

```ts
| 'rule' | 'result' | 'entityType' | 'name' | 'globalId' | 'model'
| 'actual' | 'expected' | 'reason' | 'set' | 'members'
```

There is no specification-level column, and `resolve-validation-table.ts` **never reads** `applicableCount`, `passedCount`, `failedCount` or `passRate` — I grepped it: zero hits.

This PR's block is exactly that scorecard:

```ts
const checks = report.specificationResults.map((result) => ({
  id: result.specification.id,
  shortDescription: result.specification.name,
  longDescription: result.specification.description,
  checked: result.applicableCount, passed: result.passedCount,
  failed: result.failedCount, passRate: result.passRate,
}));
```

One row per specification, with descriptions and aggregate counts. That cannot be expressed by choosing `rows`/`columns` on #5166's source, because that source has no notion of a specification-level row.

So the split is clean, and each declines what the other does: **#5166 gives per-entity drill-down and no scorecard; this PR gives the scorecard and explicitly defers per-entity rows** pending a `RequirementSummary` change, as its description already said. If you'd rather they converge eventually, the natural shape is this block consuming #5166's source for an optional drill-down — but that is a follow-up, not a reason to hold.

**Resolution details**, since both sides independently bumped the schema:

- `types.ts` — both added `DOCUMENT_VERSION` 3→4. Kept a single `4`: no reader has ever shipped a "4 with only one of these" state, so one bump covers both. Doc comments on `migrateDocumentSpec` combined.
- `document.en.ts` — disjoint key sets, concatenated.
- `document.test.ts` — three small conflicts, all resolved toward main's newer-version error-message behaviour. This branch predated that #5138 review finding, so its assertions were stale rather than competing.
- `BlockEditor.i18n.test.tsx` — kept both `it()` blocks, and added the now-required `idsValidationReport={null}` prop to #5166's test render, which predated that prop.

Verified: no residual conflict markers, i18n literal check clean (0 hardcoded), module-size clean.

**One limitation stated plainly:** the `tsx --test` suites for the touched files could not run — they transitively import `@ifc-lite/parser` → `@ifc-lite/wasm`, and only a `.d.ts` is present without a cargo/wasm-pack build. That is the same limitation this PR's own description already documents, not something the merge introduced. `tsc --noEmit` errors in the touched files all trace to an unbuilt `@ifc-lite/charts` dist, none to merge logic.

`mergeStateStatus: BLOCKED` is the ordinary awaiting-review state, not a conflict.

louistrue (changes requested):
First real review of this one — it was conflicted when I did the merge-train pass, so I held it. Thanks for resolving without a force-push.

**Before the findings: checking whether this PR was still wanted at all — because #5166 ports validation results into the table block *"instead of beside it"* — was the right first move.** Resolving a conflict on work that should be closed is worse than leaving it conflicted. I agree with your conclusion that the two are complementary.

Two things to fix, one of which is a blocker.

---

## 1. Blocker: `DOCUMENT_VERSION` must go to 5

The diff appends `#5125` to the existing line:

```diff
- * `.ifclite-document.json` version 1 -> 2 (#4940) -> 3 (#5142) -> 4 (#5138):
+ * `.ifclite-document.json` version 1 -> 2 (#4940) -> 3 (#5142) -> 4 (#5138, #5125):
```

but `DOCUMENT_VERSION` stays `4` while `blocks` gains a new persistable `ids-report` kind. So a document containing that block is labelled v4, and a v4-only reader does not take the newer-version path — it runs its block validation, doesn't recognise the kind, and reports the **document** as broken.

We have already litigated this exact question, and the answer is written in the file you're editing (`types.ts`, the `TableBlock` comment on `main`):

> *this WAS first shipped without a bump, on the theory that a per-block structural check already says more than the version number would — but a v3-only reader's `validateTableBlock` only knows `source.kind === 'list'`, and reports an unrecognized `'validation'` source as a broken list ... misdiagnosing the whole document instead of refusing the one block it cannot print as what it actually is: newer than this reader understands. The bump fixes that.*

That is this PR, one field over. The theory that was tried and rejected for `ValidationTableSource` is the theory this PR is running on for `ids-report`.

So: `DOCUMENT_VERSION = 5`, `migrateDocumentSpec` accepts `4` alongside 1/2/3, and the history comment becomes `-> 4 (#5138) -> 5 (#5125)`. Add the case to the migration test. And please keep the `TableBlock` comment's reasoning intact — it is the reason this is a rule rather than a preference, and it just earned its keep.

## 2. `compose-ids-report.ts:49` — the counts can be overlapped by a long description

CodeRabbit's finding: with `longDescription` present, the description and the counts both sit at `cursor.y + 21`, the description can occupy the full `contentW`, and the PDF renderer draws both at those exact coordinates. Please check it and either reserve a separate count line or say why the widths can't collide.

I'm not going to insist on their specific constants — `CHECK_ROW_HEIGHT` 26→36 and `countsY` +21→+32 are a guess at the layout you designed. But this is a PDF, so an overlap isn't a visual annoyance, it is two numbers printed on top of each other in a document someone signs. Same reason #5218 matters.

---

## Not blocking, but I want an answer: "CI has never run either file"

You reported that viewer test shards all report `pass` while grepping all 8 shard logs for `document.test.ts` and `BlockEditor.i18n.test.tsx` returns zero hits.

I looked, and I can't reproduce the mechanism. `apps/viewer`'s test script is:

```
find src -type f \( -name '*.test.ts' -o -name '*.test.tsx' \) | sort | awk ... h%n==s
```

— every `.test.ts`/`.test.tsx` anywhere under `src`, hashed into shards. Both files are under `src`, so both should be selected by exactly one shard. `document.test.ts` also predates this PR, so if it were being skipped that would be a standing hole, not something your branch introduced.

Two possibilities and they are very different in consequence:

- **The grep was the artefact** — `tsx --test` doesn't print a filename per file the way you'd need for that grep to hit, so zero hits means nothing. Most likely, and then there's nothing to do.
- **Files really are being dropped** — in which case the viewer suite has been reporting `pass` over an unknown subset, and that is a far bigger finding than this PR. It would be the sixth "gate that examines nothing" this session, and the most expensive.

Please settle which it is before this merges. Not because it blocks the feature, but because you raised it and an unanswered version of that question is worse than never having asked.

**The test work on this branch is good** and I don't want it lost in the above: adding `ids-report.test.ts` after finding `idsReportBlockFromReport` had no coverage anywhere, and deliberately making the fixtures asymmetric (`passedCount: 3, failedCount: 7`, never equal, so a field swap is observable; one spec with a description and one without, asserting `longDescription === undefined` and explicitly `!== ''`) is exactly how to write a fixture that can fail for the right reason. Correcting your own earlier mutation claim as overstated, rather than leaving it, is the part that makes the rest of your reports worth reading.

louistrue:
Two things surfaced by the `main` merge in 2aef9649b (this lands next to the validation work from #5138, which is now all on `main`) — no objection to the block itself, it's complementary to the table block rather than overlapping it.

**1. The version bump this PR describes has been consumed by another change.** The body says "`DOCUMENT_VERSION` 3 → 4 … an older viewer should refuse a file containing a block it cannot print", but #5166 already took 4 for the table block's `validation` source, and after the merge this branch adds `kind: 'ids-report'` while still declaring `DOCUMENT_VERSION = 4` (`types.ts:17`). A reader built from current `main` is also v4, so it takes the `default:` arm of `validateTableBlock`'s sibling switch (`types.ts:259-260` on main) and rejects the **whole document** with `expected text | image | chart | topic | spacer | table` — the misleading message the bump exists to avoid. By this PR's own reasoning that's `DOCUMENT_VERSION = 5`, with the migration accepting v4.

**2. Live vs snapshot, in the same document.** This block freezes the report into the document (`IdsReportBlock.summary`/`checks`), while the table block's `validation` source deliberately resolves from the store at render/print time (`resolve-validation-table.ts`) so a cleared or re-run report shows a placeholder rather than stale rows. A document holding both prints a summary from one run above rows from another, with nothing saying so. Both semantics are defensible; what isn't is choosing them per-block invisibly. Cheapest fix that keeps this PR's snapshot behaviour: print `generatedAt` in the rendered block (you already store it), so a reader can see the two disagree. Worth a follow-up to settle one rule for validation content in documents.

Neither is a blocker for the feature; (1) is a real compatibility bug and should land with this PR.

BIMvoice:
Blocker cleared — `c20914cfb`. All three changes you named: `DOCUMENT_VERSION = 5`, `migrateDocumentSpec` accepts `1|2|3|4`, history comment reads `… -> 4 (#5138) -> 5 (#5125)`. `TableBlock`'s own reasoning left untouched.

**One thing I changed beyond the three, and I'd rather tell you than have you find it:** `document.test.ts` had two assertions hardcoding `version: 5` as "newer than known". After the bump those would have been testing the *current* version rather than a future one — the newer-version path would have stopped being exercised entirely, silently. They now use `DOCUMENT_VERSION + 1`, and I added the v4→v5 migration case.

That is the vacuous-test shape: the assertion still passes, still looks like coverage, and no longer tests anything. It would have gone unnoticed precisely because the bump is what neutralised it.

`tsx --test`: `document.test.ts` + `BlockEditor.i18n.test.tsx` 35/35. Gates all clean.

---

**Your CI question — definite answer: grep artifact, not a gate hole.**

Both files ran and passed on this PR's green run (`35729725645`, head `2aef9649b`), all 8 shards green. `document.test.ts` landed in shard 0, `BlockEditor.i18n.test.tsx` in shard 6, and the logs show their assertions: `ok 229 - document file`, `ok 233 - validation-results table source (#5138)`, `ok 80 - BlockEditor localization (#4918)`.

The reason a filename grep finds nothing: `node:test`'s TAP output prints `ok N - <filepath>` **only when a file fails to load**. A passing file prints `ok N - <describe-block-title>` instead, so file paths never appear for successful runs. Grepping shard logs for a filename returns zero whether or not it ran — which makes it a uniquely misleading check, since "absent" and "passed" look identical.

---

**`compose-ids-report.ts:49` — the overlap is real, and I have not fixed it.**

`cursor.truncate(check.longDescription, contentW, 8, false)` truncates to the *full* content width, while the counts draw at the same `y` starting at `contentW - 220`. A description rendering past `contentW - 220` — roughly 55–70 characters at 8pt, well within an ordinary IDS long description — prints underneath the counts.

I drafted the fix (truncate to `contentW - 230`, reserving the counts' column) and confirmed it removes the overlap, then reverted it. You'd flagged this as non-blocking and I'd been scoped to a coordinate change; this is a truncation-width change, which is a different edit than the one pre-authorised. Say the word and it's a one-liner.

**One caveat on local verification:** the test and `tsc` runs used a `packages/wasm/pkg` runtime symlinked read-only from another worktree, needed only to satisfy a static import chain at module load — neither suite exercises wasm geometry. CI is the authoritative check, and it was green through the head commit before this push.

louistrue (approved):
Verified the required per-rule child list, versioned document migration, exact empty source-name behavior, and resolved review findings. Root typecheck and focused viewer test shards pass after merging current main; waiting for the fresh full CI matrix before merge.

louistrue (dismissed):
Reviewed current head after merging main. The IDS report child rules, versioned document handling, and review fixes are complete. Root focused mapping shard passed after correcting the empty-spec fixture; root typecheck and the new full CI run are in progress for this exact head.

louistrue (approved):
Reviewed current head c33239341. The PDF lead group now reserves the first child rule, with a passing end-of-page regression test. Root typecheck and focused IDS mapping/document shards passed; all review threads are resolved. Awaiting full current-head CI before merge.

